PR #1242, opened 2025-09-11
fix(slack/onboarding-llmo): allow rerun of onboarding command

cwjwisse:
- reuses the same command for onboarding
- if brand exists for given site, the users will be prompted with two buttons:
  - users can choose to only add/update entitlements and enrollments (this is done using the tier client)
  - they can also choose to update the IMS org, and subsequently update the entitlements and enrollments
- success and failure messages are posted in the same thread


src/support/slack/actions/onboard-llmo-modal.js
```diff
@@ -15,6 +15,7 @@ import { createFrom } from '@adobe/spacecat-helix-content-sdk';
 import { Octokit } from '@octokit/rest';
 import { Entitlement as EntitlementModel } from '@adobe/spacecat-shared-data-access/src/models/entitlement/index.js';
 import { OrganizationIdentityProvider as OrganizationIdentityProviderModel } from '@adobe/spacecat-shared-data-access/src/models/organization-identity-provider/index.js';
```
ravverma, line 17: We don't need `OrganizationIdentityProvider ` as everything handled by Tier Client

PR #1259, opened 2025-09-16
LLMO-424: Filters data fetched to LLMO to avoid failures due to spreadsheet size

calvarezg:
Adds a new endpoint for fetching LLMO data. The endpoint performs static mappings for brand presence, and allows filtering and also including only certain fields to keep the response as short as possible.
 
## Related Issues
[LLMO-424](https://jira.corp.adobe.com/browse/LLMO-424)


src/controllers/llmo.js
```diff
@@ -125,6 +126,317 @@ function LlmoController(ctx) {
...
+          if (key !== ':type' && data[key]?.data) {
+            data[key].data = filterArray(data[key].data);
+          }
+        });
+      }
+      return data;
+    };
+
+    // Apply inclusions to data arrays to remove specified attributes
+    const applyInclusions = (rawData, includeFields) => {
```
vivesing, line 193: This file is getting big, should we move filtering logic applyInclusions, applyExclusions, etc to a new llmoSheetUtils file?

igor-grubic (approved):

PR #1355, opened 2025-10-14
trigger audits during onboarding flow

hannessolo:
Triggers a set of audits during onboarding process

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.210.2](https://github.com/adobe/spacecat-api-service/compare/v1.210.1...v1.210.2) (2025-10-21)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


src/controllers/llmo/llmo-onboarding.js
```diff
@@ -350,6 +360,36 @@ export async function enableAudits(site, context, audits = []) {
...
+
+  imports.forEach(({ type, options }) => {
+    if (!siteConfig.isImportEnabled(type, options)) {
+      siteConfig.enableImport(type, options);
+    }
+  });
+
+  site.setConfig(Config.toDynamoItem(siteConfig));
+
+  await site.save();
```
ekremney: we cannot call `site.save()` multiple times in the same flow. That will create errors (due to versioning) on data-access level.

PR #1362, opened 2025-10-16
add de-onboarding endpoint

hannessolo:


hannessolo:
Looks like deploy is broken because of AWS outage. Will deploy tomorrow, but thanks for the approval.

PR #1373, opened 2025-10-17
fix: Enable addition of Private Repos in Spacecat via Slack and add s…

vdua:
…… (#1372)

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/support/slack/commands/add-repo.js
+++ b/src/support/slack/commands/add-repo.js
@@ -68,12 +68,12 @@ function AddRepoCommand(context) {
     }
   }
 
-  async function isOnboardedWithAemy(owner, repo, branch) {
+  async function isOnboardedWithAemy(owner, repo) {
     const { AEMY_BASE_URL } = process.env;
     if (!AEMY_BASE_URL) {
       throw new Error('AEMY_BASE_URL is not set');
     }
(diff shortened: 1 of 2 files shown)


bosschaert (approved):

src/support/slack/commands/add-repo.js
```diff
@@ -68,6 +68,21 @@ function AddRepoCommand(context) {
     }
   }
 
+  async function isOnboardedWithAemy(owner, repo, branch) {
+    const AEMY_ENDPOINT = `https://ec-xp-fapp-coordinator.azurewebsites.net/api/fn-ghapp/functions/get_installation_token/${owner}/${repo}/${branch}`;
```
vdua: shall we read it from an environment variable as done here. You may need to ask in the channel to add the environment variable in the api service.

https://github.com/adobe/spacecat-import-worker/pull/442/files#diff-a4f3f338bc03c3a0a8b82bb5ed97342471ed7c5caf7e74576d342a089e472938R39

sagarmiglani: Thanks @vdua, I was also thinking the same. Would you also happen to if env variable AEMY_API_KEY is available to spacecat-api-worker or not?

PR #1392, opened 2025-10-24
fix: imsorg and add site command tier sensitive

ravverma:

## Related Issues

Thanks for contributing!


solaris007 (commented):
thank you @ravverma. some feedback:
- modals should at least partially be shared between commands, i see duplication. please make more modular.
- this would possibly make it easier to introduce a separate command just for entitlements, e.g. `entitlements edit/check/show`

src/support/slack/actions/add-site-modal.js
```diff
@@ -0,0 +1,223 @@
...
+          },
+        ],
+      });
+
+      // Note: selectedProducts.length is always > 0 due to validation
+      const productsMessage = `\nSelected products for entitlement: *${selectedProducts.join(', ')}*`;
+      await say(`:white_check_mark: Products selected for site <${baseURL}|${baseURL}>${productsMessage}`);
+
+      // ensure entitlements and enrollments for selected products
+      /* eslint-disable no-await-in-loop */
```
solaris007: what is the reason for not using `Promise.all` instead?

ravverma:
> thank you @ravverma. some feedback:
> 
> * modals should at least partially be shared between commands, i see duplication. please make more modular.
> * this would possibly make it easier to introduce a separate command just for entitlements, e.g. `entitlements edit/check/show`

we already have APIs to check entitlement , having command would be good, but will take it later.

PR #1400, opened 2025-10-27
feat: implement URL store API

HollywoodTonight:

## Related Issues

Thanks for contributing!


nitinja (approved):
LGTM

nitinja (changes requested):
We need some sort of ordering functionality, for exaple: 
High value pages in url store needs to be sorted on rank (or may be traffic) etc.

src/controllers/url-store.js
```diff
@@ -0,0 +1,637 @@
...
+const MAX_URLS_PER_REQUEST = 100;
+const DEFAULT_LIMIT = 100;
+const MAX_LIMIT = 500;
+
+/**
+ * Canonicalizes a URL by removing trailing slashes, converting to lowercase domain, etc.
+ * @param {string} url - The URL to canonicalize
+ * @returns {string} - Canonicalized URL
+ */
+function canonicalizeUrl(url) {
```
iuliag: Looking at how URLs are usually normalized in SpaceCat:
https://github.com/adobe/spacecat-shared/blob/108e86cc164755eca40050bfe55eeeac210d45a3/packages/spacecat-shared-utils/src/url-helpers.js#L179
https://github.com/adobe/spacecat-shared/blob/108e86cc164755eca40050bfe55eeeac210d45a3/packages/spacecat-shared-utils/src/url-helpers.js#L77

I'd propose the following:
* lowercase
* strip port, trailing dot, slash
* remove query params
* prepend schema

Also, should check for redirects https://github.com/adobe/spacecat-shared/blob/108e86cc164755eca40050bfe55eeeac210d45a3/packages/spacecat-shared-utils/src/url-helpers.js#L141C16-L141C35, since URL store should not contain redirect URLs.

src/controllers/url-store.js
```diff
@@ -0,0 +1,637 @@
...
+    try {
+      // If byCustomer filter was explicitly provided or defaulted, filter by it
+      const result = await AuditUrl.allBySiteIdByCustomerSorted(siteId, byCustomerFilter, {
+        limit: parsedLimit,
+        cursor,
+        sortBy,
+        sortOrder,
+      });
+
+      return ok({
```
iuliag: Since I've spec'ed the URL store initially, a new pattern for pagination has been merged in the SpaceCat API, so let's align to it and return these details as formatted here: https://github.com/adobe/spacecat-api-service/blob/141ad36e96123865787910a8f16d5200c73ec396/src/controllers/suggestions.js#L171-L175

iuliag (changes requested):
Observations below regarding aligning with the scope of phase 1: Customer-provided URLs to be audited (for all opportunity types or per opportunity type) in addition to the currently audited pages.

Also I don't see any tests written?

nitinja:
Hi Folks, is it possible to merge a barebone working version of this functionality? We really need this for `high-value-pages`.

PR #1423, opened 2025-10-31
fix: send promise token for non cs sites as well

rpapani:
- for obtaining the executedBy we need the proper promiseToken. PromiseToken currently is being sent only for CS delivery type, as that was required by Content API earlier. Since we need the promiseToken to obtain the user email as well, sending this irrespective of the site delivery type
- renamed `getCSPromiseToken` --> `getIMSPromiseToken` as this isn't only for CS any more
- - checked that this util isn't used [outside of api-service](https://github.com/search?q=org%3Aadobe+getCSPromiseToken&type=code), so, updated the references in api-service only


ramboz (approved):
👍🏼

PR #1432, opened 2025-11-03
hotfix: limit rum data to top 100 pages

tarunsinghdev:
## Description

This PR adds an optional filtering feature to the metrics API endpoint that allows clients to request only the top 100 pages by pageviews. This is particularly useful for preventing 413 payload errors when dealing with large metric datasets.

### Core Implementation (`src/controllers/sites.js`)

- Added `filterByTop100PageViews` query parameter support to `getSiteMetricsBySource` endpoint
- When `filterByTop100PageViews=true` is passed, the API will:
  - Filter out metrics without defined `pageviews`
  - Sort remaining metrics by `pageviews` in descending order
  - Return only the top 100 entries
  - Log the filtering action with before/after counts

### Testing (`test/controllers/sites.test.js`)

Added comprehensive test suite covering:
- ✅ Filtering to top 100 when parameter is enabled
- ✅ Excluding metrics without pageviews
- ✅ Backward compatibility (no filtering when parameter not set)
- ✅ Works with different metric types (RUM, Ahrefs, etc.)
- ✅ Handles datasets with < 100 entries
- ✅ Edge cases: null and zero pageview values

## API Usage

**Endpoint:**

GET /sites/:siteId/metrics/:metric/:source?filterByTop100PageViews=true

**Example:**
GET /sites/abc-123/metrics/cwv-hourly-7d-2025-11-02/rum?filterByTop100PageViews=true

## Benefits

1. **Prevents 413 errors**: Reduces payload size for large metric datasets
2. **Backward compatible**: Existing API calls work unchanged
3. **Flexible**: Can be used with any metric type (RUM, Ahrefs, etc.)
4. **Reusable**: Can be applied to other imports like user-engagement

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/controllers/sites.js
+++ b/src/controllers/sites.js
@@ -548,7 +548,7 @@ function SitesController(ctx, log, env) {
     const siteId = context.params?.siteId;
     const metric = context.params?.metric;
     const source = context.params?.source;
-    const filterTopPages = context.data?.filterTopPages === 'true';
+    const filterByTop100PageViews = context.data?.filterByTop100PageViews === 'true';
 
     if (!isValidUUID(siteId)) {
       return badRequest('Site ID required');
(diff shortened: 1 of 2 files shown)


src/controllers/sites.js
```diff
@@ -570,7 +570,49 @@ function SitesController(ctx, log, env) {
       return forbidden('Only users belonging to the organization can view its metrics');
     }
 
-    const metrics = await getStoredMetrics({ siteId, metric, source }, context);
+    let metrics = await getStoredMetrics({ siteId, metric, source }, context);
+
+    // Filter RUM metrics to only include top 100 pages to prevent 413 payload errors
+    if (source === 'rum' && metric.includes('cwv-hourly')) {
```
dipratap: This will prevent others from using whole cwv-hourly data if they need it. Instead of updating the API logic, we should add to the API. 
Lets have a filterTopPages param that will come in the api request, and then do the filtering. This way, this functionality can be used for other imports as well - like user-engagement etc.

dipratap (changes requested):

src/controllers/sites.js
tarunsinghdev:
```
# Get full dataset (existing behavior - unchanged)
GET /sites/{siteId}/metrics/cwv-hourly-7d-2025-11-02/rum

# Get filtered dataset (new feature)
GET /sites/{siteId}/metrics/cwv-hourly-7d-2025-11-02/rum?filterTopPages=true

# Works with other metric types too
GET /sites/{siteId}/metrics/user-engagement/ahrefs?filterTopPages=true
```

will make the changes as above.

src/controllers/sites.js
```diff
@@ -570,7 +571,49 @@ function SitesController(ctx, log, env) {
       return forbidden('Only users belonging to the organization can view its metrics');
     }
 
-    const metrics = await getStoredMetrics({ siteId, metric, source }, context);
+    let metrics = await getStoredMetrics({ siteId, metric, source }, context);
+
+    // Filter metrics to only include top 100 pages when requested
```
dipratap: top pages are 200 in total

src/controllers/sites.js
```diff
@@ -570,7 +571,49 @@ function SitesController(ctx, log, env) {
...
+
+        // If no ahrefs data, try other sources
+        if (!topPages || topPages.length === 0) {
+          topPages = await SiteTopPage.allBySiteId(siteId);
+        }
+
+        if (topPages && topPages.length > 0) {
+          // Limit to top 100 pages and extract URLs
+          const topPageUrls = topPages
+            .slice(0, 100)
```
dipratap: We should return metrics for all top 200 pages, not slice at all.

PR #1463, opened 2025-11-10
fix: update enrollment query

sandsinh:
## 🧠 Summary

Refactored the `filterSitesForProductCode` function to optimize database queries by using **entitlementId** instead of **siteId** for filtering site enrollments.

---

### **Before**

* Iterated through each site individually
* Called `SiteEnrollment.allBySiteId()` for each site (**N queries**)
* Manually searched for matching entitlement in each site's enrollments
* No explicit handling for missing entitlements

### **After**

* Single query using `SiteEnrollment.allByEntitlementId()` (**1 query**)
* Created a `Set` of enrolled site IDs for **O(1)** lookup performance
* Used `Array.filter()` with `Set` lookup for cleaner, more efficient filtering
* Added early return when no entitlement exists

---

### **Query Optimization**

* **Before:** O(N) database queries where N = number of sites
* **After:** O(1) single database query
* **Impact:** For organizations with 100 sites → reduced from **100 queries to just 1**

### **Memory & CPU Efficiency**

* Uses `Set` for **O(1)** lookups instead of O(N) array search
* Eliminates `no-await-in-loop` linting violations
* Adopts a functional programming approach using `filter()` and `map()`

---

Would you like me to make it even more **GitHub Release Note–friendly** (like for a changelog or PR description)?

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/support/utils.js
+++ b/src/support/utils.js
@@ -19,6 +19,7 @@ import {
   tracingFetch as fetch,
   isValidUrl,
   isObject,
+  isNonEmptyObject,
   resolveCanonicalUrl, isValidIMSOrgId,
   detectAEMVersion,
   detectLocale,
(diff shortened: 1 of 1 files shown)


ravverma (approved):
minor comment

PR #1467, opened 2025-11-10
fix(src/controlles/llmo/llmo.js): do not trigger geo-brand-presence i…

byteclimber:
…f llmo config was updated for AI prompts categorization

Jira Ticket: https://jira.corp.adobe.com/browse/LLMO-1099

### DIFF
--- a/test/controllers/llmo/llmo.test.js
+++ b/test/controllers/llmo/llmo.test.js
@@ -1851,6 +1851,233 @@ describe('LlmoController', () => {
           },
         );
       });
+
+      it('should NOT skip audit when prompts array has different length', async () => {
+        const existingConfig = {
+          ...llmoConfig.defaultConfig(),
+          categories: {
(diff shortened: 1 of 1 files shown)


src/controllers/llmo/llmo.js
```diff
@@ -387,6 +387,69 @@ function LlmoController(ctx) {
...
+    const newCategories = newConfig?.categories || {};
+    const oldTopics = oldConfig?.topics || {};
+    const newTopics = newConfig?.topics || {};
+
+    // Get new category IDs
+    const newCategoryIds = Object.keys(newCategories).filter((id) => !oldCategories[id]);
+
+    // Get new or modified topic IDs
+    const changedTopicIds = Object.keys(newTopics).filter((id) => {
+      if (!oldTopics[id]) return true; // New topic
```
aniham: how are these IDs computed? are they just based on the topic value? 

asking because i wonder if we can get false negatives if an old topic is removed and a new topic is assigned the same ID

PR #1470, opened 2025-11-11
fix: adds sites resolve api

sandsinh:
# Add Resolve Site API Endpoint

## Overview
This PR introduces a new `GET /sites/resolve` API endpoint that allows clients to resolve site and organization data based on flexible query parameters. This endpoint is designed to support the ASO Home feature by providing a unified way to retrieve site information using either organization identifiers or IMS organization IDs.

### New API Endpoint
- **Route**: `GET /sites/resolve`

### Functionality

The endpoint provides intelligent site resolution with the following behavior:

#### Request Parameters
- **Query Parameters**:
  - `organizationId` : UUID of the organization
  - `imsOrg`: IMS Organization ID (e.g., `1234567890ABCDEF12345678@AdobeOrg`)
  - `siteId` : UUID of a specific site to validate against the organization
  
#### Resolution Logic
1. **By SiteId First** (if provided):
   - Looks up the site by ID
   - Retrieves associated organization
   - Validates that the organization matches the provided `organizationId` or `imsOrg`
   - Checks entitlements and enrollments
   - Returns matched site and organization data

2. **By OrganizationId** (if `siteId` not provided):
   - Finds organization by UUID
   - Retrieves entitled sites for the product code
   - Returns first enrolled site with organization data

3. **By IMS Org** (alternative to organizationId):
   - Finds organization by IMS Org ID
   - Retrieves entitled sites for the product code  
   - Returns first enrolled site with organization data

#### Response Format
```{
  "data": {
    "organization": {
      "id": "...",
      "name": "...",
      "imsOrgId": "...",
      ...
    },
    "site": {
      "id": "...",
      "baseURL": "...",
      "organizationId": "...",
      ...
    }
  }
}
```

### Error Handling
- `404 Not Found`
- `400 Bad Request`:

## Related Issues

Thanks for contributing!


sandsinh:
> Can we make new route align to existing patter so that it can be used by any other web application which might not on unified shell. This way API don't need to take payload as well. `/organizations/{orgId}/sites-resolve` and the `orgranizationId` parameter is internal for UI, we should not introduce it in APIs. with this we can support both cases..
> 
> 1. Natural flow use IMSorg id to get SC orgID then go for resolve api
> 2. if `orgranizationId` provided then use that to call directly resolve api
>    By this we can save ourself to introduce internal UI parameter in SC api

The performance improvement would be none for the customers if we go with the above approach. So we do not want to go with this approach.

sandsinh:
> How ?
> I am just suggesting to use route more aligned to existing pattern. You just need to keep the req
> `/organizations/by-ims-org-id/<IMSorgId>` in use to get the spaceCat org It then you can go for
> new API `/organizations/{orgId}/sites-resolve` to get first site in response , and in parallel call `/organizations/{orgId}/sites` to get all sites

@ravverma We do not want to use any existing pattern. The solution in the PR is optimally reducing the load time on the UI, let's go ahead with this solution.

PR #1478, opened 2025-11-11
feat: add disable all audits feature

tkotthakota-adobe:
https://jira.corp.adobe.com/browse/SITES-37053

Tests
https://cq-dev.slack.com/archives/C060T2PPF8V/p1763173755764749

image

image

image

image


tkotthakota-adobe:
Opening fresh PR as we want to support only disable all audits as per @rpapani

PR #1549, opened 2025-11-24
feat(llmo): add base URL check during onboarding

cwjwisse:

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/controllers/llmo/llmo-onboarding.js
+++ b/src/controllers/llmo/llmo-onboarding.js
@@ -15,8 +15,9 @@ import { createFrom } from '@adobe/spacecat-helix-content-sdk';
 import { Octokit } from '@octokit/rest';
 import { Entitlement as EntitlementModel } from '@adobe/spacecat-shared-data-access/src/models/entitlement/index.js';
 import TierClient from '@adobe/spacecat-shared-tier-client';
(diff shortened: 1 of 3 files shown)


src/controllers/llmo/llmo-onboarding.js
```diff
@@ -538,6 +539,102 @@ export async function createOrFindOrganization(imsOrgId, context, say = () => {}
...
+/**
+ * Tests a URL against the Ahrefs top pages endpoint to see if it returns data.
+ * @param {string} url - The URL to test
+ * @param {object} ahrefsClient - The Ahrefs API client
+ * @param {object} log - Logger instance
+ * @returns {Promise<boolean>} - True if the URL returns top pages data, false otherwise
+ */
+async function testAhrefsTopPages(url, ahrefsClient, log) {
+  try {
+    const { result } = await ahrefsClient.getTopPages(url, 1);
```
davidaurelio: It's ok because we do it only once, but this [fetches more fields than we need](https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-ahrefs-client/src/index.js#L157-L159).

For anything that runs regularly, the [raw `sendRequest`](https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-ahrefs-client/src/index.js#L75) with hand-curated fields should be our go-to.

PR #1558, opened 2025-11-25
fix: configurations not working

rpapani:


rpapani:
fixed with entity name in shared

PR #1585, opened 2025-11-28
feat: llmo-1539 new endpoint to get user details

vivesing:
# Add User Details API Endpoints

## Summary
Added two new API endpoints to fetch user details (name, email) for trial users with IMS fallback support for admin users.

### New Controller: `user-details.js`
- **GET** `/organizations/:organizationId/userDetails/:externalUserId`
  - Returns user details for a single external user ID
  - Fetches from trial users database first
  - Falls back to IMS API if user not found and requestor is admin
  - Returns system defaults for non-admin users

- **POST** `/organizations/:organizationId/userDetails`
  - Bulk endpoint accepting array of user IDs in request body
  - Returns object with `externalUserId` as key and user details as value
  - Logs count of IMS API calls for monitoring
  - Handles mixed results (some from DB, some from IMS)

### Features
- ✅ Access control validation (organization membership)
- ✅ Admin-only IMS fallback (non-admins get system defaults)
- ✅ Graceful error handling (IMS failures return system defaults)
- ✅ Monitoring: Logs IMS call count per request
- ✅ Fallback values for incomplete IMS profiles

### User Details Response
{
  "firstName": "John",
  "lastName": "Doe",
  "email": "john@example.com",
  "organizationId": "org-id"
}### Tests
- **22 test cases** with 100% code coverage
- Edge cases: admin vs non-admin, IMS success/failure, incomplete data
- Route integration tests updated

## Related Issues

Thanks for contributing!


vivesing:
for non admin users, if user could not be found in the TrialUsers list we assume its either service user or admin user which is not part of the imsOrg so returning system in that case. Once we have required scope added in the ims client we will be able to identify users who are not part of the IMS group in a better way

PR #1672, opened 2026-01-05
fix: allow admin users to bypass x-product header validation

irenelagno:
## Problem

The security fix introduced in commit 40208d28 (VULN-33517) added x-product header validation to prevent unauthorized access. However, the validation check happens **before** the admin access check, which blocks even admin users from accessing LLMO endpoints.

## Root Cause

In `src/support/access-control-util.js`, the `hasAccess()` method was checking the x-product header before checking admin privileges:

```javascript
// Old order (buggy)
1. Check x-product header → throws error if invalid
2. Check admin access → never reached for admins without header
```

This means admin API keys and JWT tokens with `is_admin: true` were being blocked from accessing LLMO endpoints, even though they should have full access.

## Solution

Reorder the checks so admin access is validated first:

```javascript
// New order (fixed)
1. Check admin access → admins bypass all other checks
2. Check x-product header → only for non-admin users
```

## Impact

- ✅ Admin API keys can now access LLMO endpoints without x-product header
- ✅ JWT tokens with `is_admin: true` can access LLMO endpoints
- ✅ Regular users still require the correct x-product header (security maintained)
- ✅ Backward compatibility restored for admin workflows

## Testing

Tested with ADMIN_API_KEY on CI environment:
- Before fix: `{"message":"[Error] Unauthorized request"}`
- After fix: Should return LLMO config data

Relates to: VULN-33517

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.279.1](https://github.com/adobe/spacecat-api-service/compare/v1.279.0...v1.279.1) (2026-01-05)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


rpapani (approved):

mikeweis (approved):
LGTM 👍

PR #1692, opened 2026-01-13
feat: Support object format in metrics API for cwv and engagement weekly RUM data

prithipalpatwal:
Updated the metrics API to handle the new object format introduced by the import worker for weekly CWV and user-engagement data. The API now seamlessly supports both the new object format and legacy array format.

## Related Issues

Thanks for contributing!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,45 @@
+## [1.286.1](https://github.com/adobe/spacecat-api-service/compare/v1.286.0...v1.286.1) (2026-01-19)
+
+
+### Bug Fixes
+
+* removed wrong validation ([#1707](https://github.com/adobe/spacecat-api-service/issues/1707)) ([a44a35b](https://github.com/adobe/spacecat-api-service/commit/a44a35bd091824a0fbabd195cdd9bcff3e70473e))
+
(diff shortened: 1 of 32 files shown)


src/controllers/sites.js
```diff
@@ -623,21 +623,29 @@ function SitesController(ctx, log, env) {
       return forbidden('Only users belonging to the organization can view its metrics');
     }
 
-    let metrics = await getStoredMetrics({ siteId, metric, source }, context);
+    const metrics = await getStoredMetrics({ siteId, metric, source }, context);
+
+    // Handle new metadata wrapper format { label, startTime, endTime, data }
+    const isMetadataWrapper = isObject(metrics) && !isArray(metrics) && isArray(metrics.data);
```
prithipalpatwal: @dipratap, please let me know if this is the right way to achieve our use case. Based on the API docs (https://opensource.adobe.com/spacecat-api-service/#tag/site/operation/getSiteMetricsBySource
), the endpoint returns an array only, but we’re now supporting an object response as well.

dipratap: To keep things backward compatible and follow standard pattern - lets introduce a query param based on which the api returns the metrices in the new format (object), otherwise it returns an array always.

prithipalpatwal: @dipratap Updated it, please review it again, Thanks!

dipratap (approved):
Pls rename the arrayKey param, rest lgtm.

PR #1770, opened 2026-02-04
feat: scrape job complete response

LeChef318:
-add data to scrapeJob to send completion message


danieljchuser (approved):
Looks good — clean feature addition with proper tests. Passing Slack context into scrape job metadata and reporting job IDs back to the user for CSV scrapes is a nice improvement.

PR #1787, opened 2026-02-09
fix: refactor LLMO overrideBaseURL resolution to canonical

cwjwisse:
## Summary
- Replace Ahrefs-based `overrideBaseURL` detection in LLMO onboarding with canonical URL resolution.
- Keep override detection limited to `www` vs non-`www` hostname resolution.
- Preserve the original base URL path when setting override; never adopt a differing canonical path.
- Keep existing safety behavior: if `fetchConfig.overrideBaseURL` already exists, skip auto-detection.

## Why
AI prompt generation is moving away from Ahrefs, allowing us to remove the Ahrefs check from onboarding, making the onboarding a less error-prone user experience. While that simplifies the onboarding experience, we still need to indicate to the customer the overall health status of retrieving Ahrefs for the downstream services, like Onsite Opportunities, which is covered in [this PR](https://github.com/adobe/spacecat-audit-worker/pull/1959).

## Case Handling (Current Behavior)
1. Base URL has a non-`www` subdomain (e.g. `blog.example.com`): skip detection, no override.
2. Base canonical resolves to toggled hostname (`example.com` -> `www.example.com`): set override to toggled origin + original base path.
3. Base canonical resolves on same hostname (even if canonical path differs): no override.
4. Base canonical fails, alternate (`www` toggled) resolves to alternate hostname: set override to alternate origin + original base path.
5. Both base and alternate resolve but neither indicates toggled hostname is canonical: no override.
6. Base resolves, alternate fails: no override.
7. Both fail canonical resolution: warn and continue with no override.
8. Existing `fetchConfig.overrideBaseURL` present: skip detection and preserve existing value.

## Implementation Notes
- Removed Ahrefs dependency from LLMO onboarding override detection.
- Added canonical-based helper logic to derive override from hostname signal only.
- Updated LLMO onboarding and helper tests to canonical-driven scenarios.

## Validation
- `npx eslint src/controllers/llmo/llmo-onboarding.js test/controllers/llmo/llmo-onboarding.test.js test/controllers/llmo/test-helpers.js`
- `npx mocha --require test/setup-env.js --recursive --reporter dot test/controllers/llmo/llmo-onboarding.test.js`
  - Result: `60 passing`

### DIFF
--- a/src/controllers/llmo/llmo-onboarding.js
+++ b/src/controllers/llmo/llmo-onboarding.js
@@ -15,8 +15,7 @@ import { createFrom } from '@adobe/spacecat-helix-content-sdk';
 import { Octokit } from '@octokit/rest';
 import { Entitlement as EntitlementModel } from '@adobe/spacecat-shared-data-access/src/models/entitlement/index.js';
 import TierClient from '@adobe/spacecat-shared-tier-client';
-import { composeBaseURL, tracingFetch as fetch, isNonEmptyArray } from '@adobe/spacecat-shared-utils';
-import AhrefsAPIClient from '@adobe/spacecat-shared-ahrefs-client';
+import { composeBaseURL, tracingFetch as fetch, resolveCanonicalUrl } from '@adobe/spacecat-shared-utils';
 import { parse as parseDomain } from 'tldts';
 import { postSlackMessage } from '../../utils/slack/base.js';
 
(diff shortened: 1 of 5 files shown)


src/controllers/llmo/llmo-onboarding.js
```diff
@@ -750,24 +742,34 @@ export async function determineOverrideBaseURL(baseURL, context) {
...
+    if (baseOverride) {
+      log.info(`Setting overrideBaseURL to ${baseOverride} (base URL canonical resolved to alternate hostname)`);
+      return baseOverride;
+    }
 
-    if (!baseURLSuccess && alternateURLSuccess) {
-      log.info(`Setting overrideBaseURL to ${alternateURL} (base URL failed, alternate succeeded)`);
-      return alternateURL;
+    if (baseResolvedUrl) {
+      log.debug('Base URL resolved, no overrideBaseURL needed');
```
akshaymagapu: should we use info or debug? few places here using debug

akshaymagapu (approved):

solaris007:
Closing stale PR, please re-open if still needed

PR #1792, opened 2026-02-10
feat: introduce is readonly admins

vivesing:
# spacecat-api-service - Read-Only Admin Changes

## Overview

Added support for the new `is_readonly_admin` JWT claim in the access control layer, ensuring read-only admins get visibility into all organizations/sites but are blocked from write operations like creating organizations.

### 1. `src/support/access-control-util.js`

- **`hasAdminAccess()`** - Updated to grant admin-level read access to read-only admins:
  ```javascript
  return !!(this.authInfo.isAdmin() || this.authInfo.isReadOnlyAdmin());
  ```
- **Added `isReadOnlyAdmin()` method** - Exposes the read-only admin flag from `AuthInfo`:
  ```javascript
  isReadOnlyAdmin() {
    return this.authInfo.isReadOnlyAdmin();
  }
  ```

### 2. `src/controllers/organizations.js`

- **`createOrganization` access guard** - Added `isLLMOAdministrator()` check to prevent read-only admins (who pass `hasAdminAccess()`) from creating organizations:
  ```javascript
  if (!accessControlUtil.hasAdminAccess() || !accessControlUtil.isLLMOAdministrator()) {
    return forbidden('Only admins can create new Organizations');
  }
  ```

### 3. `test/support/access-control-util.test.js`

- Added `AuthInfo` prototype patches for `isReadOnlyAdmin` and `isLLMOAdministrator` (needed until `spacecat-shared` is published).
- Added `isReadOnlyAdmin: () => false` to all existing mock `authInfo` objects to prevent `isReadOnlyAdmin is not a function` errors.
- **3 new test cases** under `Organization Access Control`:
  - `hasAdminAccess` returns `true` when user is a read-only admin (`is_readonly_admin: true`, `is_admin: false`, `is_llmo_administrator: false`).
  - `isReadOnlyAdmin` returns `false` when user is a full admin.
  - `isReadOnlyAdmin` returns `false` when user is a regular (non-admin) user.

### 4. `test/controllers/organizations.test.js`

- Added `AuthInfo` prototype patches for `isReadOnlyAdmin` and `isLLMOAdministrator`.
- Updated default test profile to include `is_llmo_administrator: true` (required by the new `createOrganization` guard).
- **1 new test case**:
  - Read-only admin (`is_readonly_admin: true`, `is_llmo_administrator: false`) is blocked from creating organizations (returns 403).

### 5. `test/controllers/paid/top-paid-opportunities.test.js`

- Added `isReadOnlyAdmin: () => false` to the mock `authInfo` in the restricted user test to fix compatibility with the updated `hasAdminAccess()`.

## Dependency Note

These changes depend on `@adobe/spacecat-shared-http-utils` being updated with the `isReadOnlyAdmin()` method on `AuthInfo`. Until that package is published, the test files include prototype patches to simulate the new method. These patches should be removed once the dependency is bumped.

## Related Issues

Thanks for contributing!


vivesing:
closing this as isReadOnlyAdmin will not be passed from the backend

PR #1795, opened 2026-02-10
fix: Promise token support

ravverma:
Provide support to get access token from a promise token embedded in Session token


solaris007:
Closing stale PR, please re-open if still needed

PR #1831, opened 2026-02-18
fix: v2/v3 data access runtime switch with instanceof fixes

ekremney:
## Summary

- Re-applies the v2/v3 data access runtime switch (reverted in #1828) with fixes for the cross-package `instanceof` failures that caused the production outage
- Replaces `instanceof ValidationError` with `error.name === 'ValidationError'` checks in controllers (fixes, opportunities, suggestions)
- Replaces `instanceof Site/Organization/Project` with `entity.entityName` checks in access-control-util
- Updates all affected tests to use `entityName` mock pattern instead of `Object.setPrototypeOf`

## Root Cause

When v3 is installed as the main package and v2 as an npm alias, `instanceof` checks against v3 classes fail for v2-created instances (different constructors from different packages). The fix uses duck-typing via `entityName` property and `error.name` string checks, both of which are stable across v2/v3.

## Test plan

- [x] All 3547 unit tests passing
- [x] 100% code coverage maintained
- [x] ESLint passes
- [ ] Deploy to stage and verify with non-admin user
- [ ] Verify `DATA_SERVICE_PROVIDER=postgres` switches to v3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+## [1.320.3](https://github.com/adobe/spacecat-api-service/compare/v1.320.2...v1.320.3) (2026-02-19)
+
+
+### Bug Fixes
+
+* **slack:** re-enablement path uses numbered brand-presence-free splits ([#1821](https://github.com/adobe/spacecat-api-service/issues/1821)) ([ca242f9](https://github.com/adobe/spacecat-api-service/commit/ca242f93425ca4ed392ef9ba3adb38857d47aea7))
+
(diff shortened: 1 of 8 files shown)


solaris007 (commented):
@ekremney - the `.name` may be similarly affected as `instanceof`.

Consider these PRs:
- https://github.com/adobe/spacecat-shared/pull/1173
- https://github.com/adobe/spacecat-shared/pull/1190

solaris007 (approved):

PR #1837, opened 2026-02-19
[SITES-40783] - fix: prefer specific routes over dynamic segments in matchPath

HollywoodTonight:
- Update matchPath to use specificity-based routing where routes with more static segments are preferred over routes with dynamic parameters
- Move PATCH /sites/:siteId/url-store above :auditType route
- Add tests for specificity-based route matching

Fixes SITES-40783

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/routes/index.js
+++ b/src/routes/index.js
@@ -187,6 +187,15 @@ export default function getRouteHandlers(
     'GET /sites/:siteId/audits/latest': auditsController.getAllLatestForSite,
     'GET /sites/:siteId/audits/:auditType': auditsController.getAllForSite,
     'GET /sites/:siteId/audits/:auditType/:auditedAt': sitesController.getAuditForSite,
+
+    // URL Store endpoints (defined before :auditType to ensure static segment match)
+    'GET /sites/:siteId/url-store': urlStoreController.listUrls,
+    'GET /sites/:siteId/url-store/by-audit/:auditType': urlStoreController.listUrlsByAuditType,
+    'GET /sites/:siteId/url-store/:base64Url': urlStoreController.getUrl,
+    'POST /sites/:siteId/url-store': urlStoreController.addUrls,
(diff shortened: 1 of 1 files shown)


ekremney (commented):
Nice fix — the specificity-based approach is solid and the tests cover the key scenarios well. One suggestion on the implementation strategy:

The PR currently does **two things**: moves the PATCH route AND adds specificity-based matching. Either one alone solves the bug, but together they introduce a trade-off worth considering.

**The specificity change scans ALL dynamic routes on every request** (the old code short-circuited on first match via `reduce`). With ~100+ dynamic routes this is still fast, but it's unnecessary work on the hot path when there's a simpler alternative.

I'd suggest picking **one** of these two approaches:

### Option A — Move the entire url-store block, drop specificity changes
Move the full url-store section (GET, GET by-audit, GET :base64Url, POST, PATCH, DELETE) above `PATCH /sites/:siteId/:auditType` as a group. First-match-wins handles it correctly — `url-store` is a static segment so it only matches literally, and falls through to `:auditType` for actual audit types like `cwv`. Zero performance overhead, keeps routes grouped together.

### Option B — Keep specificity fix, revert the route move
The specificity-based matching correctly prefers `url-store` (2 static segments) over `:auditType` (1 static segment) regardless of definition order. So the PATCH route doesn't need to move — put it back with its url-store siblings at the original location. This is more defensive against future route additions but comes with the scan-all-routes overhead.

Either way, the current PR splits the url-store PATCH away from the rest of the url-store group (GET/POST/DELETE stay at lines 296-302), which hurts readability.

Minor notes:
- If you keep the specificity approach, consider adding a test for **equal specificity** (two routes with the same number of static segments) to document the expected tie-breaking behavior.
- `Array.sort` stability is V8-guaranteed but not spec-guaranteed — a comment noting the assumption would help.

PR #1865, opened 2026-02-25
[WIP] feat: llmo mysticat data access layer apis

vivesing:
Summary: 

See details here https://wiki.corp.adobe.com/display/AEMSites/LLMO+UI+-+Mysticat+facing+Spacecat+APIs+for+Performance+Improvements 

## Related Issues

Thanks for contributing!


solaris007 (changes requested):
Hey @vivesing 

**Scope**: 13 new GET endpoints for LLMO Brand Presence, +1269/-1 lines, 9 files

This PR adds brand presence API endpoints to spacecat-api-service that query PostgreSQL tables in mysticat-data-service via PostgREST. The stated goal is replacing client-side data processing with server-side aggregation for UI performance. The data-service tables/views referenced (`brand_presence_executions`, `brand_presence_sources`, `brand_metrics_weekly`, `brand_presence_topics_by_date`, `brand_presence_prompts_by_date`, `executions_competitor_data`) all exist and have proper indexes.

---

### Critical

**1. IDOR in getSources - executionId bypasses siteId authorization**
`src/controllers/llmo/llmo-brand-presence.js` - `getSources`

When `executionId` is provided, the handler queries `brand_presence_sources` filtering ONLY by `execution_id` - the `siteId` from the URL path is completely ignored. A user authenticated for Site A can pass any `executionId` belonging to Site B and retrieve that site's source data. The authorization check (`getSiteAndValidateLlmo`) validates access to the siteId in the URL, not the execution's owning site.

Fix: Add `.eq('site_id', siteId)` to the executionId branch, or verify the execution belongs to the requested site before returning data.

**2. Client-side aggregation contradicts the PR's own design goal**
`getStats`, `getSentimentOverview`, `getFilterDimensions`

The [Confluence design doc](https://wiki.corp.adobe.com/display/AEMSites/LLMO+UI+-+Mysticat+facing+Spacecat+APIs+for+Performance+Improvements) specifies efficient SQL with `GROUP BY`, `COUNT(*) FILTER`, and `DISTINCT` - but the implementation fetches raw rows (up to 50,000) into Node.js and aggregates in JavaScript. This moves the problem from browser to Lambda instead of solving it. The `brand_metrics_weekly` view already demonstrates the right pattern (pre-aggregated, query passes through). The others should follow it.

At scale: Lambda memory limits, orders-of-magnitude higher latency vs DB aggregation, increased data transfer costs.

---

### Important

**3. User-controlled `sortBy` passed unsanitized to PostgREST `.order()`**
`parseQueryParams` + `getTopics`, `getTopicPrompts`, `getPrompts`

`sortBy` accepts arbitrary user input and flows into `.order(orderCol, ...)`. PostgREST rejects invalid columns with 400, so it's not directly exploitable for SQL injection, but it enables column enumeration and ordering by unintended columns. Fix: whitelist allowed sort columns per endpoint.

**4. LIKE metacharacter injection in search**
`getSearch`

`.ilike('topics', \`%${searchTerm}%\`)` - the `%` and `_` characters in user input are LIKE wildcards. `q=%%` passes the 2-char minimum check and matches all rows (data harvesting). Fix: escape LIKE metacharacters before interpolation.

**5. Raw error messages returned to clients**
`runWithPostgrest` catch block + individual handlers

PostgREST error messages leak table names, column names, and constraint details. The repo's CLAUDE.md explicitly says "Sanitize error messages before returning to clients." Fix: log full error, return generic message.

**6. High query limits enable resource exhaustion**
Various handlers

- `getSentimentOverview`: `.limit(50000)`
- `getStats`: `.limit(10000)` then `.in()` with up to 10K UUIDs (~360KB URL)
- `getFilterDimensions`: `.limit(5000)`

These are neither configurable nor documented. If a site has more than the limit, data is silently truncated. This becomes moot if aggregation is pushed to SQL.

**7. `select('*')` on 7 endpoints over-fetches columns**
`getTopics`, `getTopicPrompts`, `getShareOfVoice`, `getCompetitorTrends`, `getPrompts`, `getSources` (x2)

If tables gain sensitive columns, they'll be automatically exposed. The repo pattern requires DTOs for response transformation - none are used here. Fix: specify explicit column lists or add DTOs.

**8. Missing OpenAPI spec**
13 new public endpoints with no OpenAPI documentation. The repo's workflow says "OpenAPI First: Define API contract before implementation." This should be a merge blocker per the repo's own standards.

**9. Missing integration tests**
The repo requires integration tests in `test/it/` for new endpoints (both DynamoDB and PostgreSQL harnesses). Only unit tests with mocks are provided. The unit tests mostly verify HTTP status codes without checking which tables are queried or what filters are applied.

**10. No pagination metadata**
Paginated endpoints (`getTopics`, `getPrompts`, `getSearch`) return data arrays but no total count, hasMore, or cursor info. PostgREST supports `Prefer: count=exact` for free.

**11. Duplicated `getSiteAndValidateLlmo`**
`llmo-mysticat-controller.js` copy-pastes the function verbatim from `llmo.js`. Should be extracted to a shared utility.

**12. DRY violation in filter application**
The pattern `if (shouldApplyFilter(x)) q = q.eq(col, x)` for 7 parameters is duplicated ~10 times across handlers. Should be a single `applyFilters(query, params, filterMap)` helper.

---

### Minor

**13. Inconsistent week calculation**
`toISOWeekKey` uses ISO week (Thursday-based) while `getSentimentOverview` uses `getDay()` (Sunday-based). Two different week semantics in the same file will produce misaligned data.

**14. `getShareOfVoice` and `getCompetitorTrends` are nearly identical**
Both query `executions_competitor_data` with the same filters. Only differences: ordering and limit value. Could be consolidated.

**15. `getStats` returns hardcoded empty arrays**
`top_competitors: []` and `top_topics: []` - WIP stubs with no indication to the caller.

**16. `hasAccessToElmo` naming**
Legacy "ELMO" naming when the product is "LLMO". Carried over from the copy-paste of `llmo.js`.

**17. `postgrestService` is a novel pattern**
Not used anywhere else in the codebase. How is it initialized? Where does it come from? No documentation on this integration point.

**18. Dual parameter naming (camelCase + snake_case)**
`parseQueryParams` accepts both `startDate` and `start_date`, etc. Doubles the API surface. Pick one convention.

**19. Error matching by substring is fragile**
`error.message?.includes('belonging to the organization')` - if the upstream message changes, error mapping silently breaks.

**20. Package-lock.json noise**
~180 lines of unrelated `"peer": true` changes should be a separate commit or explained.

---

### Assessment

**Not ready to merge.** The critical IDOR (#1) must be fixed. The architectural approach (#2) needs reconsideration - the implementation contradicts its own design goal by doing JavaScript aggregation instead of SQL aggregation. The missing OpenAPI spec (#8) and integration tests (#9) are merge blockers per the repo's established standards.

Recommended path forward:
1. Fix the IDOR in getSources
2. Create SQL views in mysticat-data-service for stats, sentiment, and filter-dimensions (following the `brand_metrics_weekly` pattern)
3. Reduce spacecat-api-service handlers to auth + proxy (thin layer)
4. Add OpenAPI spec
5. Add integration tests
6. Extract shared helpers (filter application, getSiteAndValidateLlmo)

PR #1900, opened 2026-03-04
feat: adding support for sns topic on suggestions autofix

christopwner:
proposal for swapping out autofix sqs with sns so that it may support multiple subscribers

## related
See https://github.com/adobe/spacecat-infrastructure/pull/545


solaris007 (changes requested):
Hey @christopwner,

Thanks for tackling the SQS-to-SNS migration - the direction is right. SNS fan-out is the correct pattern when you need multiple subscribers, and it aligns with how the Commerce Catalog LLMO architecture uses SNS for product update fan-out (per the [LLMO Layers wiki](https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=EntComm&title=LLMO+Layers+in+Adobe+Commerce), step 9-10). That said, there are several issues that need to be resolved before this can ship.

### Critical - Runtime Failures

**1. `snsWrapper` is never wired into the middleware chain**
`src/index.js:187` references `context.sns`, but `snsWrapper` is never imported or added to the `.with()` chain (~line 294). `context.sns` will be `undefined` at runtime, and every autofix suggestion request will throw a `TypeError`. Fix: import `snsWrapper` and add `.with(snsWrapper)` alongside the existing `.with(sqs)`.

**2. Missed code path - assess-urls flow still references `sqs`**
`src/controllers/suggestions.js` around line ~847 has an `assess-urls` code path that still destructures `AUTOFIX_JOBS_QUEUE` from env and calls `sqs.sendMessage()`. Since the controller signature changed from `sqs` to `sns`, this will throw a `ReferenceError` at runtime for any assess-urls request.

**3. No SNS topic exists in infrastructure**
There is no SNS topic defined in `adobe/spacecat-infrastructure` and no companion Terraform PR. The only autofix infrastructure is the SQS queue `spacecat-autofix-jobs` in `modules/sqs/queues.tf`. The Lambda IAM role also lacks `sns:Publish` and `sns:CreateTopic` permissions - only SQS permissions are granted. This means even if the code were correct, publishing would fail with `AuthorizationError`.

**4. No consumer migration - autofix-worker will stop receiving messages**
The autofix-worker is triggered via SQS event source mapping on the `spacecat-autofix-jobs` queue. Publishing to SNS instead of SQS means the worker will never receive messages - there is no `aws_sns_topic_subscription` wiring the SNS topic to the SQS queue. No companion PRs were found in `adobe/spacecat-services` or `adobe/spacecat-infrastructure`.

**5. No tests updated**
`test/controllers/suggestions.test.js` has 30+ references to `mockSqs` and `AUTOFIX_JOBS_QUEUE` - all unchanged. The new `sns.js` (147 lines) has zero test coverage.

### Important - Design Issues

**6. `CreateTopicCommand` for ARN resolution is an anti-pattern**
`sns.js:#createOrGetTopicArn` uses `CreateTopicCommand` to resolve topic name to ARN. This requires write permissions (`sns:CreateTopic`) and will silently create orphaned topics on misconfiguration instead of failing fast. The existing SQS pattern uses `GetQueueUrlCommand` (read-only). Recommendation: provision the topic in Terraform, pass the full ARN as `AUTOFIX_JOBS_TOPIC_ARN`, and drop the `CreateTopicCommand` path entirely.

**7. SNS alone is insufficient - use SNS+SQS fan-out**
Standard AWS pattern for reliable multi-subscriber processing is SNS topic -> SQS queue per subscriber. This preserves buffering, backpressure, DLQ, and retry semantics that the autofix-worker currently relies on. The migration path should be: (1) create SNS topic in Terraform, (2) subscribe the existing SQS queue to it, (3) add new subscriber queues, (4) switch the publisher (this PR).

**8. `timestamp` injection in the transport layer**
`sns.js:52-55` injects a `timestamp` field into every message payload. The SQS wrapper does not do this. The transport layer should not silently modify message payloads - let callers decide.

**9. Missing X-Ray tracing parity**
`@adobe/spacecat-shared-utils` SQS wrapper uses `instrumentAWSClient` for X-Ray tracing. The new SNS class has no instrumentation. If SNS becomes a standard pattern, consider adding it to `spacecat-shared-utils` for observability parity.

### Minor

**10. Dead code**: `getTopicAttributes()` method and `GetTopicAttributesCommand` import are unused.

**11. Style**: `sns.js` uses double quotes for imports; the rest of the codebase uses single quotes.

**12. Missing dependency**: `@aws-sdk/client-sns` does not appear in the diff for `package.json`.

### Recommended Sequencing

1. **Terraform PR first**: Create SNS topic, add IAM permissions, subscribe existing SQS queue to it
2. **This PR second**: Wire `snsWrapper`, fix assess-urls path, pass full topic ARN, add tests
3. **New subscriber PR(s)**: Add additional SQS subscriptions for new consumers
4. **Coordinated deploy**: Infra first, then code

The architecture is sound - this just needs the implementation and rollout plan fleshed out.

PR #1914, opened 2026-03-06
feat: Apply s2s auth wrapper

ravverma:
### Summary

Integrates the new **Service-to-Service (S2S) authentication wrapper** ([`s2sAuthWrapper`](https://github.com/adobe/spacecat-shared/pull/1407)) from `@adobe/spacecat-shared-http-utils` into `spacecat-api-service`. This enables per-route capability-based authorization for machine-to-machine (S2S) consumers, while allowing end-user tokens to pass through untouched.

The `s2sAuthWrapper` (implemented in [adobe/spacecat-shared#1407](https://github.com/adobe/spacecat-shared/pull/1407)):
- Validates S2S JWT bearer tokens (ES256 signature, issuer, expiry)
- Identifies S2S consumers via the `is_s2s_consumer` JWT claim
- Fetches the consumer record from the database to verify it is not revoked or suspended
- Enforces per-route capabilities by matching the incoming request against a `routeCapabilities` map
- Rejects unmapped routes for S2S consumers with `401` — routes must be explicitly allowlisted

### Changes

| File | Change |
|------|--------|
| `src/index.js` | Imported `s2sAuthWrapper` and `routeRequiredCapabilities`; chained `.with(s2sAuthWrapper, { routeCapabilities: routeRequiredCapabilities })` after `authWrapper` in the middleware stack |
| `src/routes/required-capabilities.js` | **New file** — maps every route to a required capability string (`entity:action`), covering all entities: audits, configurations, organizations, projects, sites, opportunities, suggestions, fixes, LLMO, imports, scrape jobs, reports, sentiment, entitlements, etc. |
| `test/routes/required-capabilities.test.js` | **New file** — validates every entry in the capabilities map: correct `METHOD /path` key format, valid HTTP methods, `entity:action` value format, and non-empty entities |
| `test/index.test.js` | Stubs `s2sAuthWrapper` via `esmock` so existing index tests aren't affected by the new middleware |
| `package.json` | Updated `@adobe/spacecat-shared-http-utils` dependency to include `s2sAuthWrapper` |

### Key design decisions

- **Capabilities come from the database, not the token** — JWT tokens are used only for authentication and consumer identification (`client_id`). Authorization is resolved from the consumer's database record to avoid stale capabilities in short-lived tokens.
- **Explicit allowlisting** — any route not present in `routeRequiredCapabilities` is automatically denied for S2S consumers. This is a secure-by-default approach that requires new routes to be consciously opted in.
- **Capability naming convention** — `entity:action` where action is `read` for GET requests and `write` for all mutating methods (POST, PUT, PATCH, DELETE). Entity is the deepest resource noun in the route path, singularized.

### Test plan

- [x] `routeRequiredCapabilities` format validation tests pass
- [x] Existing index tests pass with `s2sAuthWrapper` stubbed out
- [x] Unit tests pass (`npm test`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,45 @@
+# [1.340.0](https://github.com/adobe/spacecat-api-service/compare/v1.339.1...v1.340.0) (2026-03-10)
+
+
+### Features
+
+* Service to Service communication guides ([#1928](https://github.com/adobe/spacecat-api-service/issues/1928)) ([91902e1](https://github.com/adobe/spacecat-api-service/commit/91902e15e774eaadf46d628ade4c162c932ae537))
+
+## [1.339.1](https://github.com/adobe/spacecat-api-service/compare/v1.339.0...v1.339.1) (2026-03-09)
+
+
+### Bug Fixes
+
+* **deps:** update adobe fixes ([#1927](https://github.com/adobe/spacecat-api-service/issues/1927)) ([46743c5](https://github.com/adobe/spacecat-api-service/commit/46743c5eb5ed92773d07c9d922c2a2e9b3b24f82))
+
+# [1.339.0](https://github.com/adobe/spacecat-api-service/compare/v1.338.0...v1.339.0) (2026-03-09)
+
+
+### Features
+
+* enhance assess-urls action ([#1926](https://github.com/adobe/spacecat-api-service/issues/1926)) ([8e8de76](https://github.com/adobe/spacecat-api-service/commit/8e8de762746213ac4dcb2ada1f5759defcefbbdb))
+
+# [1.338.0](https://github.com/adobe/spacecat-api-service/compare/v1.337.0...v1.338.0) (2026-03-09)
+
+
+### Features
+
+* [SITES-38900] add skipReason and skipDetail support for SKIPPED suggestions ([#1873](https://github.com/adobe/spacecat-api-service/issues/1873)) ([9cfe84e](https://github.com/adobe/spacecat-api-service/commit/9cfe84ef6165dc386cc20344b5d8e8cb51c4f4c6))
+
+# [1.337.0](https://github.com/adobe/spacecat-api-service/compare/v1.336.4...v1.337.0) (2026-03-09)
+
+
+### Features
+
+* decompress gzip ([#1916](https://github.com/adobe/spacecat-api-service/issues/1916)) ([15fb326](https://github.com/adobe/spacecat-api-service/commit/15fb3260511a67685be3538b26afc36dce74cc16))
+
+## [1.336.4](https://github.com/adobe/spacecat-api-service/compare/v1.336.3...v1.336.4) (2026-03-09)
+
+
+### Bug Fixes
+
+* return 404 instead of 400 for missing sites in LLMO config endpoints ([#1920](https://github.com/adobe/spacecat-api-service/issues/1920)) ([9fcd6ee](https://github.com/adobe/spacecat-api-service/commit/9fcd6ee1d43d4c071e2b9ee6c45a508ab4edf2b9))
+
 ## [1.336.3](https://github.com/adobe/spacecat-api-service/compare/v1.336.2...v1.336.3) (2026-03-07)
 
 
--- a/docs/openapi/examples.yaml
+++ b/docs/openapi/examples.yaml
@@ -150,6 +150,8 @@ broken-backlinks-suggestions-status-update:
       status: 'APPROVED'
     - id: '945e4567-f543-12d3-a456-426614174010'
       status: 'SKIPPED'
+      skipReason: 'ALREADY_IMPLEMENTED'
+      skipDetail: 'Fix was applied manually last week'
 broken-backlinks-suggestions-status-update-response:
   description: Example for broken backlinks suggestions status update response
   value:
--- a/docs/openapi/schemas.yaml
+++ b/docs/openapi/schemas.yaml
@@ -2890,6 +2890,15 @@ SuggestionStatus:
     - 'PENDING_VALIDATION'
     - 'REJECTED'
   example: 'NEW'
+SuggestionSkipReason:
+  description: Reason for skipping a suggestion when status is SKIPPED
+  type: string
+  enum:
+    - ALREADY_IMPLEMENTED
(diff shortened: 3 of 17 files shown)


solaris007 (changes requested):
Hey @ravverma - reviewed this alongside the upstream #1407 wrapper. The design direction is solid - deny-by-default, capabilities from DB, explicit route allowlisting. Good foundation. A few things need addressing before this can leave draft.

---

## Critical

### 1. Gist-hosted tarball dependency (`package.json:83`)

The `@adobe/spacecat-shared-http-utils` dependency points to a personal GitHub Gist tarball:

```
https://gist.github.com/ravverma/73fbd39cad0536f360791528e21c3857/raw/.../adobe-spacecat-shared-http-utils-1.21.1.tgz
```

This is a supply chain risk that must not merge:
- The gist is owned by a single personal account, not an org-controlled artifact
- Contains a `.DS_Store` file (suggests manual upload, not CI/CD artifact)
- No npm provenance/signature verification
- This package runs in the auth middleware chain with access to every request and the full context object
- Gist raw URLs can be updated in-place by the owner

**Fix**: Publish a proper npm release (even a pre-release like `1.22.0-beta.1`) of `spacecat-shared-http-utils` once #1407 is merged. This PR should wait for that.

### 2. Upstream s2sAuthWrapper returns 401 on missing bearer token

Because s2sAuthWrapper is outermost in the `.with()` chain (runs first), it intercepts ALL requests before `authWrapper`. The current behavior in #1407 returns 401 when no bearer token is present - this would break every existing API key consumer (`ScopedApiKeyHandler`, `LegacyApiKeyHandler`). This is a production-breaking change.

**Fix**: Blocked on #1407 - the wrapper must pass through to `fn(request, context)` when no bearer token is present, not return 401.

### 3. Upstream s2sAuthWrapper doesn't set `context.attributes.authInfo`

After successful S2S validation, the wrapper calls the next handler without populating `authInfo`. Downstream controllers that check `authInfo` (via `AccessControlUtil.fromContext()`) will fail or treat S2S consumers as unauthenticated. The `SKIP_AUTH` fallback in `run()` could also fire incorrectly since `authInfo` is falsy.

**Fix**: Blocked on #1407 - the wrapper must construct an `AuthInfo` instance and set `context.attributes.authInfo` before calling downstream.

---

## Must Fix

### 4. ~38 routes missing from capability map - need conscious categorization

Cross-referencing `routes/index.js` against `required-capabilities.js` reveals ~38 unmapped routes. Deny-by-default correctly blocks these for S2S consumers, but it's unclear which omissions are intentional vs. oversight. Categories include:

- **Likely intentional**: webhooks (`/event/fulfillment`, `/hooks/*`), Slack (`/slack/*`), consumer management (`/consumers/*`)
- **Possibly oversight**: LLMO operations (`POST /llmo/onboard`, `POST /.../llmo/offboard`, `POST /.../llmo/edge-optimize-config`, etc.), suggestion edge operations (`POST /.../suggestions/edge-deploy`, `edge-preview`, `edge-rollback`, `auto-fix`), entitlement write (`POST /organizations/:orgId/entitlements`), preflight, consent banner, user activities, trial users

**Fix**: Either add the missing routes to the map with appropriate capabilities, or create an explicit `INTERNAL_ROUTES` exclusion list with comments explaining why each is excluded. Every omission should be a conscious, documented decision.

### 5. No cross-reference test between routes and capability map

`test/routes/required-capabilities.test.js` validates format only (valid HTTP method, `entity:action` syntax). There is no test that verifies coverage against actual registered routes. Every future route addition will be a silent deny for S2S consumers with no warning.

**Fix**: Add a test that imports both the route definitions and `routeRequiredCapabilities`, then asserts every registered route either (a) exists in the map, or (b) is in an explicit exclusion list. This is the single most important test for preventing drift.

### 6. Tests stub out s2sAuthWrapper entirely (`test/index.test.js:23-28`)

```js
const s2sAuthWrapperStub = (fn) => fn;
```

The stub makes the wrapper a transparent passthrough. Zero integration tests exercise the actual auth chain interaction. This masks the production-breaking interaction from finding #2.

**Fix**: Add at least one integration test (separate file is fine) that exercises the real wrapper with:
- API key request (no bearer token) - must pass through to authWrapper
- Valid S2S JWT hitting a mapped route - must succeed with authInfo populated
- Valid S2S JWT hitting an unmapped route - must return 401
- Valid S2S JWT missing required capability - must return 401

---

## Should Fix

### 7. Entity naming inconsistency

Some entities use camelCase (`importJob`, `scrapeJob`, `siteEnrollment`, `sentimentTopic`) while others use kebab-case (`api-key`, `bot-blocker`). This could cause silent authorization failures if a consumer is granted `import-job:write` but the map expects `importJob:write`. Pick one convention and normalize.

### 8. `GET /trigger` mapped to `audit:write`

A GET that triggers a side effect violates HTTP semantics (GET should be safe/idempotent per RFC 7231). The capability mapping is pragmatically correct, but the route should be POST. Worth noting for a follow-up.

### 9. `site:read` blast radius is very broad

Granting `site:read` implicitly covers traffic data, URL store, LLMO data, experiments, metrics, and more. Fine for v1 with a small number of trusted S2S consumers, but document this scope so it's a conscious decision. If fine-grained access control is ever needed, this bundling will need unwinding.

### 10. Wrapper ordering needs explicit documentation

The `.with()` chain ordering is non-obvious (last added = outermost = runs first). Add a code comment in `src/index.js` explaining the execution order to prevent future confusion:

```js
// Wrapper execution order (outermost first):
// 1. s2sAuthWrapper - intercepts S2S JWT bearer tokens, passes through non-S2S
// 2. authWrapper - handles API key, IMS token, legacy key auth
const wrappedMain = wrap(run)
  .with(authWrapper, { ... })      // inner - runs second
  .with(s2sAuthWrapper, { ... })   // outer - runs first
```

---

## Recommended Sequencing

1. Address findings on spacecat-shared #1407 and get it merged
2. Publish proper npm release of `spacecat-shared-http-utils`
3. Update this PR's dependency to the real npm version
4. Add cross-reference test for route-capability completeness
5. Categorize the ~38 excluded routes (exclusion list vs. add to map)
6. Normalize entity naming convention
7. Add integration tests with real wrapper
8. Move PR out of draft

PR #1917, opened 2026-03-06
fix: use promise token cookie in preflight and suggestions controller

radhikagpt1208:
JIRA: https://jira.corp.adobe.com/browse/SITES-33748

## Summary
Updates the preflight controller to read the `promiseToken` from the browser cookie (set by `POST /auth/promise`) instead of the `x-promise-token` request header or minting a new token via IMS on every request.

## Problem
For promise-based authoring types (CS, CS_CW, AMS), the preflight controller previously relied on either:
1. An `x-promise-token` request header (which the MFE cannot set since the cookie is `HttpOnly` and unreadable by JS), or
2. Calling IMS on every request to mint a new promise token (unnecessary round-trip)

The `auth/promise` endpoint sets the promise token as an `HttpOnly` cookie via `Set-Cookie`. When the MFE calls `POST /preflight/jobs` with `credentials: 'include'`, the browser automatically sends this cookie in the Cookie request header. The controller was never reading it.

## Changes

- `src/controllers/preflight.js`: Replaced `x-promise-token` header lookup and IMS fallback with `getCookieValue(context, 'promiseToken')` to read the promiseToken cookie from `context.pathInfo.headers.cookie` (populated by `enrichPathInfo` middleware). Falls back to IMS if cookie is absent.
- `test/controllers/preflight.test.js`: Updated tests to reflect cookie-based token resolution - covers cookie present, cookie absent (IMS fallback), and cookie missing promiseToken key.

#### Fastly VCL update required for end-to-end cookie flow
- This PR is safe to merge on its own - the IMS fallback preserves backward compatibility. However, the cookie-based flow will not work end-to-end in deployed environments until a coordinating Fastly VCL change is made.
- The Fastly VCL at `spacecat-infrastructure/fastly/vcl/8_deliver/deliver-100-set-cors-response-headers.vcl` unconditionally sets `Access-Control-Allow-Origin: *` for all /api/* requests, overwriting any CORS headers Lambda returns.
- Per the Fetch spec, browsers reject credentials: 'include' responses when `Access-Control-Allow-Origin is *`. 
- The VCL also does not set `Access-Control-Allow-Credentials: true` or include Cookie in Access-Control-Allow-Headers.
- Until the VCL is updated, the browser will not send the cookie on cross-origin POST requests. The `getCookieValue()` call will return null and fall through to `getIMSPromiseToken()`.

#### Fastly config update for Preflight request
- This is being tracked in [SITES-41543](https://jira.corp.adobe.com/browse/SITES-41543). 
- Fastly VCL has been updated to set `access-control-allow-origin` to the origin and `access-control-allow-credentials: true`:

```
# set to origin if coming from preflight otherwise set to *
     if (( req.url ~ "^/api/ci/preflight/" || req.url ~ "^/api/v1/preflight/")
        && (req.http.Origin ~ "^https://experience(-qa|-stage)?\.adobe\.com$"
            || req.http.Origin ~ "^https?://localhost")) {
        set resp.http.access-control-allow-origin = req.http.Origin;
        set resp.http.access-control-allow-credentials = "true";
    } else if (!resp.http.access-control-allow-origin) {
        set resp.http.access-control-allow-origin = "*";
    }
```

#### Follow-up: Fastly config update for Suggestions request
- Update Fastly VCL to include suggestions autofix path.
- The current VCL regex only covers /api/ci/preflight/ and /api/v1/preflight/.
- The suggestions autofix endpoint (/api/v1/sites/:siteId/opportunities/:opportunityId/suggestions/auto-fix) is not matched, so the cookie flow won't work for autofix until the VCL paths are expanded.
- This will be done once UI changes are in place and tested for the e2e flow @ravverma.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.351.2](https://github.com/adobe/spacecat-api-service/compare/v1.351.1...v1.351.2) (2026-03-17)
+
+
+### Bug Fixes
+
+* Restrict when the Summit PLG (product-led growth) experience is … ([#1967](https://github.com/adobe/spacecat-api-service/issues/1967)) ([744b932](https://github.com/adobe/spacecat-api-service/commit/744b932a434492c861adb0081a93819342a63dd2))
+
 ## [1.351.1](https://github.com/adobe/spacecat-api-service/compare/v1.351.0...v1.351.1) (2026-03-16)
 
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@adobe/spacecat-api-service",
-  "version": "1.351.1",
+  "version": "1.351.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@adobe/spacecat-api-service",
-      "version": "1.351.1",
+      "version": "1.351.2",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/fetch": "4.2.3",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@adobe/spacecat-api-service",
-  "version": "1.351.1",
+  "version": "1.351.2",
   "description": "SpaceCat API Service",
   "main": "src/index.js",
   "type": "module",


ravverma:
Hi @solaris007,
Adding a proposal to solve CORS problem , @radhikagpt1208 Please validate.
# CORS for Credentialed Preflight Flow

## Problem

The Preflight MFE uses `credentials: 'include'` for:

- `POST /auth/promise` — sets HttpOnly `promiseToken` cookie (auth-service)
- `POST /preflight/jobs` — sends that cookie (api-service)

With credentials, the browser requires:

- `Access-Control-Allow-Origin`: exact requesting origin (not `*`)
- `Access-Control-Allow-Credentials: true`

`localCORSWrapper` in api-service handles this when `ENABLE_CORS=true`, but in deployed envs this is off. API Gateway CORS returns `Access-Control-Allow-Origin: *`, so credentialed requests fail.

---

## Solution Overview

Enable Lambda-based CORS in the api-service for deployed environments and align configuration across dev/stage/prod.

---

## 1. Use Lambda for CORS (Remove "Local Only")

**File:** `src/index.js`

Update the comment so CORS is not treated as local-only:

```javascript
/**
 * CORS middleware wrapper for credentialed cross-origin requests.
 * When ENABLE_CORS=true, reflects the requesting origin (if allowlisted) and
 * sets Access-Control-Allow-Credentials for cookie-bearing requests.
 * Required for Preflight MFE flow (POST /auth/promise, POST /preflight/jobs).
 */
```

No code changes to `localCORSWrapper` itself — it already does origin reflection when the origin is in the allowlist.

---

## 2. Environment Configuration

Add the following to secrets/params for each environment.

### dev

```json
{
  "ENABLE_CORS": "true",
  "CORS_ALLOWED_ORIGINS": "https://preflight-dev.adobe.com,https://localhost:3000,http://localhost:3000"
}
```

### stage

```json
{
  "ENABLE_CORS": "true",
  "CORS_ALLOWED_ORIGINS": "https://preflight-stage.adobe.com,https://stage-preflight.example.com"
}
```

### prod

```json
{
  "ENABLE_CORS": "true",
  "CORS_ALLOWED_ORIGINS": "https://preflight.adobe.com"
}
```

Replace with the real Preflight MFE origins per environment.

---

## 3. API Gateway Configuration

Lambda responses must not be overwritten by API Gateway CORS.

- With Lambda proxy integration, the Lambda response (including headers) is returned as-is.
- If API Gateway is adding CORS (e.g. "Enable CORS" or gateway responses), either:
  - Disable API Gateway CORS and rely on Lambda for CORS headers, or
  - Ensure API Gateway does not add `Access-Control-Allow-Origin` (or uses a pass-through) so Lambda’s headers win.

**Action:** In the AWS console or IaC for api-service, review CORS settings and remove any `Access-Control-Allow-Origin: *` that would override Lambda.

---

## 4. OPTIONS Preflight Handling

Both the OPTIONS response and the actual POST response must include correct CORS headers.

**API service:** Requests go through `localCORSWrapper` before route matching. There may not be a route for OPTIONS, so:

1. Add explicit OPTIONS handling for credentialed paths, or
2. Ensure the main handler returns a proper response for OPTIONS on those paths.

If OPTIONS is handled by API Gateway (mock/integration), it may return `*`. In that case, either:

- Route OPTIONS for `/preflight/jobs` (and any other credentialed paths) to Lambda so `localCORSWrapper` can add headers, or
- Adjust API Gateway integration so OPTIONS uses the same Lambda as POST (if supported).

---

## 5. Include `Cookie` in Allowed Headers

**File:** `src/index.js` — `localCORSWrapper`

`Cookie` must be in `Access-Control-Allow-Headers` for credentialed requests:

```javascript
response.headers.set(
  'Access-Control-Allow-Headers',
  'Content-Type, Authorization, x-api-key, x-ims-org-id, x-client-type, x-import-api-key, '
  + 'x-trigger-audits, x-requested-with, origin, accept, Cookie, x-promise-token',
);
```

Add `Cookie` to the existing header list.

---

## 6. Cookie Path Alignment

Auth-service sets `promiseToken` with `Path=${API_GATEWAY_PREFIX}` (e.g. `/api` or `/api/v1`).

Ensure `API_GATEWAY_PREFIX` matches the path under which both auth- and api-service run so the cookie is sent for `POST /preflight/jobs`. If they use different prefixes, the cookie path must cover both.

---

## Checklist

| # | Task | Owner |
|---|------|-------|
| 1 | Update `localCORSWrapper` comment | Dev |
| 2 | Add `ENABLE_CORS` and `CORS_ALLOWED_ORIGINS` to dev/stage/prod secrets | DevOps/Config |
| 3 | Confirm Preflight MFE origins per env | Product/Frontend |
| 4 | Check API Gateway CORS config; remove `*` override | DevOps |
| 5 | Add `Cookie` to `Access-Control-Allow-Headers` | Dev |
| 6 | Verify OPTIONS for `/preflight/jobs` hit Lambda | Dev/QA |
| 7 | Test credentialed flow (auth/promise → preflight/jobs) in each env | QA |

---

## Testing

1. From Preflight MFE origin, call `POST /auth/promise` with `credentials: 'include'` and validate `promiseToken` cookie.
2. Call `POST /preflight/jobs` with `credentials: 'include'` and confirm the cookie is sent and the request succeeds.
3. Check for CORS errors in the browser console.
4. Verify response headers: `Access-Control-Allow-Origin: <exact-origin>`, `Access-Control-Allow-Credentials: true`.

radhikagpt1208:
> @ravverma @radhikagpt1208 - the CORS solution proposed in the comment above is on the right track but there's a blocker layer that needs to be addressed first.
> 
> ## Fastly VCL overrides Lambda CORS headers
> `fastly/vcl/8_deliver/deliver-100-set-cors-response-headers.vcl` runs in the `deliver` subroutine (after the backend response) and **unconditionally overwrites** all CORS headers for every `/api/v1|ci|stage|prod` request:
> 
> ```vcl
> set resp.http.access-control-allow-origin = "*";
> set resp.http.access-control-allow-headers = "authorization, x-api-key, ...";
> // Access-Control-Allow-Credentials is never set
> ```
> 
> Even if `localCORSWrapper` returns a reflected origin + `Access-Control-Allow-Credentials: true`, Fastly replaces them before the browser sees the response. The `ENABLE_CORS=true` Lambda change won't help in deployed environments without a VCL change.
> 
> ## Two additional blockers
> 1. `Cookie` and `x-promise-token` are not in the VCL's `Access-Control-Allow-Headers` - the browser's OPTIONS preflight will be rejected even if origin reflection is fixed
> 2. `Access-Control-Allow-Credentials` is never set by Fastly - required for `credentials: include` to work
> 
> ## Required VCL change
> `fastly/vcl/8_deliver/deliver-100-set-cors-response-headers.vcl` needs to reflect the origin conditionally:
> 
> ```vcl
> if (req.url ~ "^/api/v1" || req.url ~ "^/api/ci" || req.url ~ "^/api/stage" || req.url ~ "^/api/prod") {
>     set resp.http.access-control-allow-methods = "GET, HEAD, PATCH, POST, PUT, OPTIONS, DELETE";
>     set resp.http.access-control-allow-headers = "authorization, x-api-key, x-import-api-key, origin, x-requested-with, content-type, accept, x-trigger-audits, x-promise-token, cookie";
> 
>     if (req.http.Origin ~ "preflight.*\.adobe\.com" || req.http.Origin ~ "^https?://localhost") {
>         set resp.http.access-control-allow-origin = req.http.Origin;
>         set resp.http.access-control-allow-credentials = "true";
>     } else {
>         set resp.http.access-control-allow-origin = "*";
>     }
> }
> ```
> 
> The exact origin regex needs to be updated with the real Preflight MFE origins per environment. The Lambda-side changes (`ENABLE_CORS=true`, `CORS_ALLOWED_ORIGINS`) are still worth doing for local dev consistency but won't unblock the deployed environments on their own.

Thanks for the detailed explanation, @solaris007 @ravverma. 
I've created a ticket [here](https://jira.corp.adobe.com/browse/SITES-41543) to track this. We can first update the config for dev, test it e2e and then replicate the same for prod.

PR #1978, opened 2026-03-17
feat: Adding a slack command for detecting CDN for customers

jsanchez07:
[SITES-41613](https://jira.corp.adobe.com/browse/SITES-41613)

This PR adds the detect-cdn Slack command so users can trigger CDN detection for a URL from Slack; the API forwards the command to the audit worker via SQS. No REST API or OpenAPI changes.


src/support/slack/commands/detect-cdn.js
```diff
@@ -0,0 +1,94 @@
...
+import BaseCommand from './base.js';
+import { extractURLFromSlackInput, postErrorMessage } from '../../../utils/slack/base.js';
+
+const PHRASES = ['detect-cdn'];
+
+export default function DetectCdnCommand(context) {
+  const baseCommand = BaseCommand({
+    id: 'detect-cdn',
+    name: 'Detect CDN',
+    description: 'Detects which CDN a website uses (e.g. Cloudflare, Akamai, Fastly) from HTTP headers. Optional: pass a Spacecat site base URL to associate the result with a site for future onboarding.',
```
mikeweis: I read the "Optional" part that I would need to pass another parameter to the `detect-cdn` command to achieve this behavior.  However, it seems something more like: "If the website happens to also be a SpaceCat site base URL, then (_something additional happens_)".
 * If the above is a true statement, then I would suggest to reword this `Optional` sentence to make it very explicit what the used needs to do in order to have _whatever happens_ happen.
 * And maybe add comments in the source code for example calls & the resulting actions.  Like:
   * detect-cdn https://non-spacecat-website.com  /* returns the detected CDN for this website */
   * detect-cdn adobe.com  /* returns detected CDN, and since this is also a SpaceCat website does BLAH */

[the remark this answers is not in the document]
jsanchez07: I changed this to match what you proposed

jsanchez07: I have changed the wording here to make it more clear.

PR #1979, opened 2026-03-17
feat: onboarding and update redirects

jkf276:
https://jira.corp.adobe.com/browse/SITES-41106

added updated-redirect as a required onboarding process for  cloud service or cloudservice/cross site.

refactored the existing identify-redirects command into help function in util.js, shared usage with onboarding.

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/support/utils.js
+++ b/src/support/utils.js
@@ -1130,6 +1130,94 @@ export const createEntitlementAndEnrollment = async (
   }
 };
 
+/*
+* Queues an identify redirects audit for a site.
+* @param {Object} site - The site to queue an identify redirects audit for.
+* @param {string} baseURL? - The base URL of the site.
+* @param {number} minutes? - The number of minutes to audit for.
+* @param {boolean} updateRedirects? - Whether to update the redirects.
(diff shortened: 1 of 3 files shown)


src/support/utils.js
```diff
@@ -1124,6 +1124,97 @@ export const createEntitlementAndEnrollment = async (
...
+    const deliveryConfig = site.getDeliveryConfig?.() || {};
+    const { programId, environmentId } = deliveryConfig;
+    if (!hasText(programId) || !hasText(environmentId)) {
+      return { ok: false, error: ':x: This site is missing `deliveryConfig.programId` and/or `deliveryConfig.environmentId` required for Splunk queries.' };
+    }
+
+    if (!hasText(env?.AUDIT_JOBS_QUEUE_URL)) {
+      return { ok: false, error: ':x: Server misconfiguration: missing `AUDIT_JOBS_QUEUE_URL`.' };
+    }
+    if (!sqs) {
```
noruiz: Do we need this check here? there's already `if(!sqs)` on line 1161

PR #1980, opened 2026-03-17
feat: edge deploy API async support

dipratap:
Implements experimentation engine flow in the Edge deployment API flows.
https://wiki.corp.adobe.com/spaces/AEMSites/pages/3812179390/Experimentation+Engine+Implementation


docs/openapi/site-opportunities.yaml
```diff
@@ -504,6 +504,369 @@ site-opportunity-suggestions-status:
...
+                      type: array
+                      items:
+                        type: object
+                        properties:
+                          uuid:
+                            type: string
+                            format: uuid
+                          index:
+                            type: integer
+                          statusCode:
```
nit23uec: we should have Location also in the resp header - not blocking this PR, but pls log a separate ticket to take care of it.

dipratap:
Thanks for the thorough review! Addressed the following:

## Critical / Major

- **C1**: DRS/save/AsyncJob failures now all throw to the outer catch, returning `207` consistently instead of mixed `500`s  
- **EC-1 / EC-2**: Moved `EXPERIMENT_IN_PROGRESS` suggestion marking to after AsyncJob creation (last step), and switched to `Promise.allSettled` so partial save failures log a warning instead of aborting the response  
- **EC-3**: Added `preScheduleId` truthy validation after DRS call — throws if schedule ID is missing  
- **EC-4**: Fixed domain-wide prompt source selection to include domain-wide suggestions as candidates (was only using regular suggestions)  
- **M1**: Removed the raw `postgrestClient async_jobs` query entirely — GeoExperiment record is the source of truth for status/phase/error, no need to expose raw job rows  
- **M2**: Extracted hardcoded DRS config into `GEO_EXPERIMENT_SCHEDULE` constants with comments explaining the cron/expiry relationship  
- **M3**: Fixed variable name (`top15ByContentGainRatio`) and added a comment explaining the two-step ranking rationale  
- **M4**: Created `GeoExperimentDto` in `src/dto/geo-experiment.js`, used in both `listGeoExperiments` and `getGeoExperiment`  
- **M5**: Updated OpenAPI spec to reflect the actual status (`GENERATING_BASELINE` / `IN_PROGRESS` / `COMPLETED` / `FAILED`) + phase enums  

## Minor

- **EC-7**: Added `context.s3` null guard before destructuring  
- **EC-8**: Deduplicated `suggestionIds` via `new Set()` before processing  
- **EC-9**: S3 parse errors now log an info message  
- **m5**: Removed `jobs` from `getGeoExperiment` response schema and implementation (since it was using the raw postgrest bypass)  

---

Also added `enable_brand_presence: true` to `job_config` in both pre and post DRS schedule calls.

PR #2055, opened 2026-03-27
feat: LLMO-3945 add onboarding Brandalf mode resolution

dzehnder:
## Summary
- resolve onboarding mode from Mysticat `brandalf` override with `LLMO_ONBOARDING_DEFAULT_VERSION` fallback
- only run v2 config initialization and Brandalf submission when onboarding resolves to `v2`
- submit onboarding prompt-generation jobs through DRS with `metadata.onboarding_mode`

## Testing
- `mocha --timeout 10000 --require test/setup-env.js test/support/llmo-onboarding-mode.test.js`
- `mocha --timeout 10000 --require test/setup-env.js --grep "skip v2 initialization|DRS client not configured|DRS prompt generation failure gracefully|LLMO onboarding for a new site" test/controllers/llmo/llmo-onboarding.test.js`
- `node --check src/controllers/llmo/llmo-onboarding.js && node --check src/support/llmo-onboarding-mode.js`

## Issue
- LLMO-3945

### DIFF
--- a/src/controllers/llmo/llmo-onboarding.js
+++ b/src/controllers/llmo/llmo-onboarding.js
@@ -58,6 +58,18 @@ function resolveUpdatedBy(context) {
     || 'system';
 }
 
+function buildOnboardingMetadata({
+  siteId, imsOrgId, brandName, onboardingMode, extra = {},
+}) {
+  return {
+    site_id: siteId,
+    imsOrgId,
+    brand: brandName,
+    onboarding_mode: onboardingMode,
+    ...extra,
+  };
+}
+
 function buildBrandalfMetadata({
   organizationId,
   siteId,
@@ -67,15 +79,17 @@ function buildBrandalfMetadata({
   onboardingMode,
 }) {
   const { hostname } = new URL(companyWebsite);
-  return {
+  return buildOnboardingMetadata({
+    siteId,
     imsOrgId,
-    brand: brandName,
-    site: hostname,
-    site_id: siteId,
-    spaceCatId: organizationId,
-    company_website: companyWebsite,
-    onboarding_mode: onboardingMode,
-  };
+    brandName,
+    onboardingMode,
+    extra: {
+      site: hostname,
+      spaceCatId: organizationId,
+      company_website: companyWebsite,
+    },
+  });
 }
 
 export async function triggerBrandalfOnboardingJob({
@@ -123,14 +137,16 @@ function buildPromptGenerationMetadata({
   region,
   onboardingMode,
 }) {
-  return {
-    site_id: siteId,
+  return buildOnboardingMetadata({
+    siteId,
     imsOrgId,
-    base_url: baseUrl,
-    brand: brandName,
-    region,
-    onboarding_mode: onboardingMode,
-  };
+    brandName,
+    onboardingMode,
+    extra: {
+      base_url: baseUrl,
+      region,
+    },
+  });
 }
 
 export async function submitOnboardingPromptGenerationJob({
--- a/src/support/feature-flags-storage.js
+++ b/src/support/feature-flags-storage.js
@@ -85,6 +85,41 @@ export async function upsertFeatureFlag({
 }
 
 /**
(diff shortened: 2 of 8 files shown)


irenelagno (commented):
## Minor: `buildPromptGenerationMetadata` duplicates `buildBrandalfMetadata` pattern

`buildPromptGenerationMetadata` (new in this PR) is structurally identical to `buildBrandalfMetadata` — both take a bag of identifiers (siteId, imsOrgId, onboardingMode, etc.) and return a flat metadata object with snake_case keys. They differ only in which fields they include.

Consider either:
- Extracting a shared `buildOnboardingMetadata(fields)` builder that both call with their specific field set
- Or at minimum adding a comment linking the two so future changes to the metadata shape stay in sync

Not a blocker, but reduces the risk of one drifting from the other (e.g., a new field added to one but forgotten in the other).

irenelagno (approved):

rainer-friederich (approved):
LGTM so far, worth a try with the other Prs!

tkotthakota-adobe:
Hi @dzehnder @irenelagno 
Asked Claude to review this PR. Here it is. 

`
  PR #2055 Review — LLMO Brandalf Mode Resolution

  Critical

  1. ensureInitialCustomerConfigV2 has no error isolation — llmo-onboarding.js:1275

  The Brandalf DRS job has a try/catch (fail gracefully), but ensureInitialCustomerConfigV2 is awaited bare right before it:
  await ensureInitialCustomerConfigV2({...});   // throws → entire onboarding fails
  try {
    await triggerBrandalfOnboardingJob({...});  // fails gracefully
  }
  If Postgres is unavailable during V2 mode onboarding, the whole onboarding is aborted. Given the DRS call already fails
  gracefully, the V2 config init should likely do the same — or at minimum the intent should be documented if blocking is
  intentional.

  2. buildInitialCustomerConfigV2 assumes at least one brand — llmo-onboarding.js:~175

  const [brand] = config.customer.brands;
  brand.v1SiteId = siteId;   // TypeError if brands is empty
  No guard for an empty brands array. If convertV1ToV2 ever returns no brands, this will throw with a confusing error.

  ---
  Important

  3. resolveUpdatedBy called twice for the same context — llmo-onboarding.js:~200

  const config = buildInitialCustomerConfigV2({ updatedBy: resolveUpdatedBy(context) });
  await writeCustomerConfigV2ToPostgres(..., resolveUpdatedBy(context));
  Minor inconsistency risk if context.attributes changes between calls (unlikely but odd). Extract to a local variable.

  4. HLX_ADMIN_TOKEN → HLX_ONBOARDING_TOKEN rename in offboarding functions

  startBulkStatusJob, pollJobStatus, and bulkUnpublishPaths are all offboarding functions, yet the new env var is named
  HLX_ONBOARDING_TOKEN. This is semantically confusing. The PR description doesn't mention this rename — is it intentional? If
  so, it needs a corresponding secrets/env update in all deployed environments.

  ---
  Minor

  5. submitOnboardingPromptGenerationJob has no error handling at call site

  The prompt generation job is also awaited bare inside the if (brandProfile) block at ~1327. Consistent with the existing
  pattern (the old drsClient.submitPromptGenerationJob also wasn't wrapped), but worth noting that a DRS failure here would
  surface as an onboarding error to the user.

  6. isValidFeatureFlagName rejects kebab-case (e.g. brandalf is fine, but my-flag would fail)

  The regex is /^[a-z][a-z0-9_]*$/ — only allows snake_case. The LLMO_BRANDALF_FLAG = 'brandalf' is fine, but if a future flag
  like prompt-gen is added, it would silently fail validation without a clear error.

  ---
  Tests — Good Coverage

  llmo-onboarding-mode.test.js and feature-flags-storage.test.js cover all branches well including null client, non-boolean flag
  values, DB errors, and the false override path. No gaps there.

  The new llmo-onboarding.test.js cases cover DRS-not-configured and failure paths for the new V2 block. Missing: a test for the
  ensureInitialCustomerConfigV2 failure case (which currently has no try/catch).
`

PR #2058, opened 2026-03-28
revert: remove slim DTO workaround now that compressResponse middleware is in place

ravverma:
## Summary

`compressResponse` middleware (added in #2045, shipped in `spacecat-shared-http-utils` v1.25.0)
now handles response compression globally, making the slim DTO workaround obsolete.

- Restores `GET /sites` to return the full `SiteDto.toJSON` shape
- Reverts `PATCH /sites/:siteId` config update back to direct `setConfig()` — the shallow-merge
  was only added to protect against data loss when clients echoed back the trimmed list DTO

## Changes

- `src/controllers/sites.js` — `getAll`: `toListJSON` → `toJSON`; `updateSite`: removed shallow-merge
- `src/dto/site.js` — removed `toListJSON` method
- `src/dto/config.js` — removed `toListJSON` function and export
- `test/controllers/sites.test.js` — restored original test; removed 4 shallow-merge tests
- `test/dto/config.test.js` — removed `toListJSON` describe block
- `test/it/shared/tests/sites.js` — removed `expectSiteListDto`, restored `expectSiteDto`

## Reverts

- `fix: use slim site DTO in getAll to avoid 413 payload too large`
- `fix: shallow-merge config in PATCH /sites to prevent partial overwrites`
- `fix: shallow merge config patch (#2020)`


solaris007:
@ravverma did anyone miss the _fat_ DTOs, does it have a functional impact? if not, should we keep it slim?

PR #2113, opened 2026-04-02
feat: add URL Inspector PG API endpoints | LLMO-4030

JayKid:
## Summary
- 6 API endpoints calling URL Inspector RPCs in mysticat-data-service:
  - `GET .../url-inspector/stats` — aggregate stats + weekly sparklines
  - `GET .../url-inspector/owned-urls` — paginated owned URL citations
  - `GET .../url-inspector/trending-urls` — paginated non-owned URLs (groups flat RPC rows by URL)
  - `GET .../url-inspector/cited-domains` — domain-level aggregations
  - `GET .../url-inspector/domain-urls` — paginated URLs within a domain (with urlId, promptsCited, categories, regions)
  - `GET .../url-inspector/url-prompts` — prompt breakdown for a specific URL
- Exports 5 shared utilities from `llmo-brand-presence.js` for reuse
- Unit tests for all handlers

## Key changes since initial PR
- `domain-urls` response now includes `urlId`, `promptsCited`, `categories`, and `regions` fields from the enriched RPC
- `url-prompts` endpoint added for Phase 3 drilldown (prompt analysis per URL)

## Related PRs
- **RPCs + perf analysis**: https://github.com/adobe/mysticat-data-service/pull/288
- **UI**: https://github.com/adobe/project-elmo-ui/pull/1429

## Test plan
- [x] Unit tests pass (lint + syntax checks confirmed; full test suite requires Node 24)
- [x] Verified end-to-end via local PostgREST → dev Aurora
- [x] Full-stack validation: UI → API → PostgREST → Aurora with enriched domain-urls response
- [ ] Deploy to dev and verify via deployed UI


calvarezg:
## Code Review

Nice work on this PR — the architecture is clean, the plan doc is excellent, and the test coverage is thorough. A few items to flag:

### Issues

**1. Stale JSDoc on `createUrlInspectorCitedDomainsHandler`**

The comment says "No pagination — domain count per site is bounded" but the handler actually implements pagination via `parsePaginationParams` and passes `p_limit`/`p_offset` to the RPC. The plan doc (section 5) confirms this was intentionally changed to be paginated. The JSDoc should be updated to match.

**2. `domain-urls` and `url-prompts` don't pass `brandId` filter**

The `domain-urls` handler and `url-prompts` handler only destructure `spaceCatId` from `ctx.params` — they skip `brandId` entirely and don't pass `p_brand_id` to the RPC. However, both have routes registered with `:brandId` variants. The other 4 handlers all pass `p_brand_id`. If the RPCs accept this parameter, this is a bug that would return unfiltered-by-brand data during drilldown. If the RPCs don't support it, the `:brandId` route variants are misleading.

**3. `domain-urls` doesn't pass `p_category` / `p_region` filters**

The domain-urls RPC call omits category and region filters that the parent endpoints (cited-domains, stats) do pass. If a user has filtered by category/region at the domain level and then drills into a domain, those filters would silently be dropped. Same applies to `url-prompts` which also omits category/region. This could cause user confusion when drilldown results don't match the parent view. If the RPCs don't support these params, a comment explaining why would help.

**4. RPC errors exposed to clients**

All handlers return `badRequest(error.message)` for PostgREST/RPC errors. This could leak internal PostgreSQL error details to API consumers. Consider returning a generic "Internal error processing request" message and keeping the detailed error only in the log.

### Minor

- **`eslint-disable-next-line max-len`** added for `withBrandPresenceAuth` export — consider splitting the function signature across lines instead of suppressing the lint rule.
- **cited-domains test happy path**: the test data doesn't include a `total_count` field in the RPC response rows, so `totalCount` will be `0` even though 2 domains are returned. Add `total_count: 2` to the test data to verify the pagination metadata path.
- **Trending URLs grouping with null URL**: if the RPC returns rows where `url` is `null`, they all group under a single `null` key in the Map, producing one URL entry with `url: ""` that aggregates all null rows. Verify this is desired vs. filtering them out.

calvarezg (approved):

PR #2170, opened 2026-04-09
fix: update waitlisted PLG

anshulk-public:
## Summary

Extends the PLG onboarding `update` (BYPASS) handler with two new resolution paths for
`DOMAIN_ALREADY_ASSIGNED` waitlisted domains, plus enrollment cleanup when offboarding
a previously onboarded domain.

### New Features

**`moveSite` bypass flag**
When `siteConfig.moveSite: true` is provided in the bypass request, the site is transferred
from its current org to the requesting org and re-onboarded — but only if the site has no
active product enrollments. If enrollments exist, a `400` is returned.

**`alternateDomain` bypass flag**
When `siteConfig.alternateDomain` is provided, the conflicting domain is retired (`INACTIVE`)
and the PLG onboarding flow is kicked off for the alternate domain under the current org's
IMS org ID. 

**Richer `DOMAIN_ALREADY_ASSIGNED` waitlist message**
The waitlist reason now includes enrollment context:
- No active products → notes the domain can be safely moved
- Has active products → notes why it cannot be moved

**ASO enrollment revocation on offboard**
New `revokeOnboardedSiteEnrollments` function removes all ASO site enrollments when a
previously `ONBOARDED` domain transitions to `INACTIVE` (e.g., during
`DOMAIN_ALREADY_ONBOARDED_IN_ORG` bypass). Only ASO-product enrollments are removed;
other product enrollments are left intact.

**`imsOrgId` update in delivery config on `moveSite`**
When moving a site to a new org, if the site's delivery config already contains an
`imsOrgId`, it is updated to reflect the new org's IMS org ID.

### Refactoring

- **`AEM_SITE_CHECK` bypass**: Migrated from `rumHost` (a publish host) to `deliveryType`
  \+ optional `authorUrl`, enabling support for AEM CS, AEM Edge, AEM AMS, and other
  delivery types. `AEM_CS_PUBLISH_HOST_PATTERN` replaced by `AEM_CS_AUTHOR_URL_PATTERN`.
- **`isBypassed` helper removed**: Logic inlined into `presetDeliveryType` check.

## Related Issues

Thanks for contributing!


Kanishkavijay39 (changes requested):
Hey @anshulk-public,

## Strengths

- `revokeOnboardedSiteEnrollments` is correctly scoped to ASO-only enrollments with solid defensive null guards at every exit point (`src/controllers/plg/plg-onboarding.js`, new function)
- The enrollment-aware `DOMAIN_ALREADY_ASSIGNED` waitlist message gives operators the context they need to choose the right resolution path without a separate database query
- Replacing `rumHost` with explicit `deliveryType` + optional `authorUrl` is a better API contract - callers declare intent rather than encoding a host pattern the system has to reverse-engineer
- Removing the fragile `isBypassed` reviews-array substring scan in favor of an explicit `presetDeliveryType` parameter makes bypass semantics direct and testable
- `deliveryType` allowlist validation correctly excludes `OTHER` and validates against the enum
- Admin-only access gate is intact across all new bypass paths
- Test coverage for the new paths is solid: `moveSite`, `alternateDomain`, enrollment revocation with ASO filtering, and the `imsOrgId` delivery config update are all covered

## Issues

### Critical (Must Fix)

**Null organization guard missing in `moveSite`**

`onboarding.getOrganizationId()` is used directly as the target org for `site.setOrganizationId()`. Admin-created records that bypassed the normal onboarding flow can have a null `organizationId`. Setting a site's org to null is silent data corruption that is very hard to diagnose after the fact.

Fix: add a guard before the assignment:
```js
if (!currentOrgId) return badRequest('Onboarding record has no associated organization');
```

### Important (Should Fix)

**`moveSite` transfers the site before verifying onboarding can proceed**

After the site is moved, `performAsoPlgOnboarding` runs and hits the "only one domain per IMS org" guard. If the requesting org already has an ONBOARDED domain, the moved domain gets waitlisted and a 200 is returned. The site now belongs to the new org but has no active onboarding record under it - the ownership transfer happened but onboarding did not.

Fix: pre-check for existing ONBOARDED records in the requesting org before moving the site. The transfer and the onboarding should succeed or fail as a unit.

---

**`alternateDomain` retires the original record before validating the alternate domain**

`onboarding.setStatus(STATUSES.INACTIVE)` and `save()` are called before `performAsoPlgOnboarding`. If the alternate domain fails hostname validation or the onboarding throws, the original domain record is already INACTIVE with no rollback path. The operator ends up with no active record for either domain.

Fix: validate `siteConfig.alternateDomain` with `isValidHostname` at the handler level before retiring the original, or defer the save until after `performAsoPlgOnboarding` succeeds.

---

**`revokeOnboardedSiteEnrollments` called without error isolation**

If `enrollment.remove()` throws, the outer catch returns 500. But `oldOnboarded` is already saved as INACTIVE. The offboard is half-complete: status changed, enrollments still live. This is silent from the caller's perspective.

Fix: wrap the revocation in try/catch with a `log.warn` to make it best-effort (consistent with how LD flag updates are handled), or reorder to do the revocation before saving the INACTIVE status.

---

**`authorUrl` is not validated before being stored in the delivery config**

For `AEM_CS`, a non-matching author URL silently drops `programId`/`environmentId` from the delivery config (the `csMatch || []` fallback yields undefined). For `AEM_AMS` and other types, any string is accepted and stored as-is. Downstream services consuming `authorURL` may make HTTP requests to it.

Fix: validate `authorUrl` is a valid HTTP(S) URL in the `AEM_SITE_CHECK` handler. For `AEM_CS`, also validate it matches `AEM_CS_AUTHOR_URL_PATTERN` and return a 400 if not.

---

**`AEM_EDGE` author URL matching silently fails for full URLs**

`EDS_HOST_PATTERN` is anchored to bare hostnames (`^...aem\.live$`). Since `AEM_CS` expects a full URL, admins will reasonably pass `https://main--repo--owner.aem.live` for `AEM_EDGE`. The regex will not match, `hlxConfig` will never be set, but `steps.authorUrlResolved` is still set to `true`. The failure is silent - not logged, not surfaced as an error.

Fix: either normalize by stripping the protocol before matching, validate and return a 400 for the wrong format, or document clearly that `AEM_EDGE` expects a bare hostname (not a URL) while `AEM_CS` expects a full URL.

### Minor (Nice to Have)

**Redundant `existingOrgLabel` variable**

`existingOrgLabel` is defined as `existingOrg?.getName?.() || existingOrgName`, but `existingOrgName` is already `existingOrg?.getName?.() || existingOrgId`. Both variables resolve to the same value. `existingOrgLabel` can be replaced with `existingOrgName`.

---

**Enrollment guard asymmetry between `moveSite` and `revokeOnboardedSiteEnrollments`**

`moveSite` blocks if ANY enrollment exists (all products), while `revokeOnboardedSiteEnrollments` only removes ASO enrollments. The broader check is arguably correct - you should not move a site with any active product. A short comment explaining the intentional difference would help future readers avoid incorrectly "aligning" them.

---

**Enriched waitlist message includes org names from both orgs**

Fine for admin-only access today. Worth a comment noting this is admin-facing only, to prevent a future developer from surfacing it in a non-admin context where it would leak one customer's org name to another.

## Recommendations

- Consider extracting the three `DOMAIN_ALREADY_ASSIGNED` sub-paths (`alternateDomain`, `moveSite`, default bypass) into named functions. The switch case is growing and the nesting depth makes control flow hard to follow.
- `revokeOnboardedSiteEnrollments` is only called in the `DOMAIN_ALREADY_ONBOARDED_IN_ORG` bypass path. If status can also be set to INACTIVE via a direct admin endpoint, that path does not revoke enrollments. This may be intentional (admin override should not cascade), but documenting the decision prevents a future "fix" that introduces inconsistency.
- The PR title `fix: update waitlisted PLG` understates the scope. This adds two new bypass resolution paths, changes the `AEM_SITE_CHECK` API contract from `rumHost` to `deliveryType`/`authorUrl`, and introduces enrollment lifecycle behavior on offboard. `feat:` is more accurate, and the `rumHost` to `deliveryType` change is a breaking change for any existing callers of the AEM_SITE_CHECK bypass worth noting in the changelog.
- Add a test for `alternateDomain` with an invalid hostname to verify the retire-before-validate risk is caught before the record is saved.

## Assessment

Ready to merge? With fixes
Reasoning: The core logic is well-structured with good test coverage for the new happy paths. The null org guard (Critical) and the move-before-validate ordering (Important) are data integrity risks that should be fixed before merge, along with `authorUrl` validation and the EDS URL format inconsistency.

## Next Steps

1. Add the null org guard in `moveSite` before `site.setOrganizationId`.
2. Pre-check for existing ONBOARDED records in the requesting org before transferring site ownership.
3. Validate `alternateDomain` before setting the original record to INACTIVE.
4. Validate `authorUrl` at the handler level; fix the EDS URL format inconsistency.
5. Isolate `revokeOnboardedSiteEnrollments` errors as best-effort or reorder save-after-revoke.
6. Minor items are optional.

Kanishkavijay39:
Hey @anshulk-public,                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                     
  Strengths                              
                                                                                                                                                                                                                                                     
  Good progress since the last review - the majority of prior findings are now addressed:                                                                                                                                                            
                                         
  - Null org guard in moveSite (plg-onboarding.js:1169) - prior Critical is fixed                                                                                                                                                                    
  - revokeOnboardedSiteEnrollments try/catch error isolation (plg-onboarding.js:1070) - prior Important is fixed
  - authorUrl validation for AEM_CS and AEM_EDGE (plg-onboarding.js:1095-1116) - prior Important is fixed; the AEM_EDGE silent failure is now a clear 400                                                                                            
  - Redundant existingOrgLabel variable - prior Minor is fixed                                                                                                                                                                                       
  - The early-return-if-ONBOARDED guard in performAsoPlgOnboarding is a clean idempotency mechanism and correctly replaces the removed 409 check for the ONBOARDED case                                                                              
  - 177 new lines of tests with solid variety: enrollment revocation paths, authorUrl validation for each delivery type, moveSite null org guard, and waitlist message variants                                                                      
                                                                                                                                                                                                                                                     
  Issues                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                     
  Important (Should Fix)                                                                                                                                                                                                                             
                                         
  alternateDomain: original record saved INACTIVE before alternate domain is validated (plg-onboarding.js:1150)                                                                                                                                      
                                                                    
  onboarding.setStatus(STATUSES.INACTIVE) and save() execute before performAsoPlgOnboarding validates the alternate domain's hostname. If isValidHostname(siteConfig.alternateDomain) rejects the input, the outer catch returns 400 but the original
   record is permanently INACTIVE with no rollback - recovery requires a manual DB fix.
                                                                                                                                                                                                                                                     
  This was raised in the prior review and is still present. Fix: call isValidHostname(siteConfig.alternateDomain) before setting INACTIVE, or defer the save until after performAsoPlgOnboarding succeeds. Also add a test for the invalid-hostname  
  path (the prior review recommended this explicitly).
                                                                                                                                                                                                                                                     
  ---                                                               
  Removed 409 guard: behavioral gap for non-ONBOARDED existing records (plg-onboarding.js:1203, around where the call to performAsoPlgOnboarding now is)
                                                                                                                                                                                                                                                     
  The old 409 guard fired for any existing PlgOnboarding record for (existingImsOrgId, domain) regardless of status. The new early-return in performAsoPlgOnboarding only fires when status === ONBOARDED. Records in WAITLISTED, IN_PROGRESS, or
  ERROR state will now be re-processed rather than rejected.                                                                                                                                                                                         
                                                                    
  If re-processing is intentional, add tests covering those states and add a comment documenting that this is a deliberate behavior change from the 409. Without either, this is a silent semantic change. The test that was updated (returns        
  existing onboarding when domain already onboarded for existing org) only covers the ONBOARDED case.
                                                                                                                                                                                                                                                     
  ---                                                               
  AEM_AMS authorUrl accepted without host validation - SSRF risk (plg-onboarding.js:1117)
                                                                                                                                                                                                                                                     
  For delivery types other than AEM_CS and AEM_EDGE, the only check is !/^https?:\/\//. Any HTTP(S) URL passes, including internal addresses like https://169.254.169.254 (AWS metadata) or https://10.0.0.1. The URL flows into
  deliveryConfig.authorURL and is later used as a base for fetch() calls in downstream services.                                                                                                                                                     
                                                                    
  The endpoint is admin-only, which substantially limits exploitability. However, isSafeDomain() already exists at line 116 of the same file and is applied to the onboarding domain - reusing it here is a one-liner:                               
                                                                    
  } else {                                                                                                                                                                                                                                           
    try {                                                           
      const parsed = new URL(siteConfig.authorUrl);
      if (!isSafeDomain(parsed.hostname)) {                                                                                                                                                                                                          
        return badRequest('authorUrl must not point to a private or internal address');
      }                                                                                                                                                                                                                                              
    } catch {                                                       
      return badRequest('authorUrl must be a valid HTTP(S) URL');                                                                                                                                                                                    
    }                                                               
  }                                      

  ---
  AEM_CS_AUTHOR_URL_PATTERN is not end-anchored (plg-onboarding.js:97)
                                                                                                                                                                                                                                                     
  The regex has no $ anchor. A URL like https://author-p123-e456.adobeaemcloud.com/../../admin?foo=bar passes validation. The full URL (including path and query) is stored in deliveryConfig.authorURL and concatenated into API calls downstream.
  Since all paths stay within Adobe infrastructure the risk is bounded, but path traversal could redirect downstream calls to unintended endpoints on the AEM author instance. Fix: normalize to parsed.origin after validation to strip             
  path/query/fragment before storage.                               
                                                                                                                                                                                                                                                     
  Minor (Nice to Have)                                              
                                         
  moveSite non-atomic ownership transfer (plg-onboarding.js:1173): site.setOrganizationId and site.save() execute before performAsoPlgOnboarding. If onboarding fails after the site transfer, the site belongs to the new org without a completed   
  onboarding record. The flow is resumable via re-trigger, but this is non-obvious to the next person reading the code. Add a comment acknowledging the non-atomicity and the recovery path.
                                                                                                                                                                                                                                                     
  Commit message typo: "onbaroded" in commit 184eeffda5 should be "onboarded". This will appear verbatim in the changelog since the repo uses semantic-release. Worth squash-fixing before merge if possible.                                        
                                         
  AEM_EDGE confusing error for non-HTTP protocols: siteConfig.authorUrl.replace(/^https?:\/\//i, '').split('/')[0] - if the input is ftp://ref--repo--owner.aem.live, the protocol strip regex does not match, split('/')[0] yields ftp:, and the    
  error returned is "must be a valid EDS host" when the real issue is the protocol. Low impact but misleading.
                                                                                                                                                                                                                                                     
  Recommendations                                                   
                                         
  - Extract the three DOMAIN_ALREADY_ASSIGNED sub-paths (alternateDomain, moveSite, default reassignment) into named functions. The switch case now spans roughly 120 lines with 4+ levels of nesting and a fourth sub-path is plausible. Named      
  functions make each scenario independently testable.
  - Consider adding a lifecycle hook or guard to the updateOnboardingStatus admin endpoint that calls revokeOnboardedSiteEnrollments when transitioning to INACTIVE. Currently only the DOMAIN_ALREADY_ONBOARDED_IN_ORG bypass path revokes          
  enrollments on offboard - the alternateDomain retire path and any direct admin status update do not.                                                                                                                                               
                                         
  Assessment                                                                                                                                                                                                                                         
                                                                    
  Ready to merge? With fixes                                                                                                                                                                                                                         
  Reasoning: The prior Critical and three of the five Important issues from the last review are addressed. The remaining blockers are the alternateDomain retire-before-validate data integrity risk (raised in the prior review, still open), the
  SSRF gap for AEM_AMS authorUrl (easy fix using the existing isSafeDomain), and the unanchored AEM_CS regex. The 409 removal needs either tests or a comment to document intent.                                                                    
                                                                    
  Next Steps                                                                                                                                                                                                                                         
                                                                    
  1. Add isValidHostname(siteConfig.alternateDomain) validation before setting the original record to INACTIVE, and add a test for the invalid-hostname path.                                                                                        
  2. Apply isSafeDomain to the authorUrl hostname for AEM_AMS and other delivery types.
  3. Normalize AEM_CS authorUrl to parsed.origin before storage to drop path and query.                                                                                                                                                              
  4. Add tests for the DOMAIN_ALREADY_ASSIGNED default bypass with non-ONBOARDED existing records, or add a comment documenting the behavioral change from the removed 409.                                                                          
  5. Minor items are optional, but the commit message typo is worth a squash-fix before merge.

Kanishkavijay39 (changes requested):
Hey @anshulk-public,

## Strengths

This commit addresses 5 of the prior findings - good progress:

- **Null org guard in `moveSite`** (was Critical): `if (!currentOrgId) return badRequest(...)` added at line 1168-1170, with a direct test. Fixed.
- **`revokeOnboardedSiteEnrollments` error isolation** (was Important): try/catch with `log.warn` added at lines 1070-1074. The offboarding record is already saved before revocation is attempted, so the system stays consistent even if revocation fails. Fixed.
- **`authorUrl` validation** (was Important): comprehensive per-type validation added at lines 1098-1118 - AEM_CS pattern match, AEM_EDGE hostname extraction + EDS pattern check, HTTP(S) check for other types. All four branches have tests. Fixed.
- **AEM_EDGE silent failure for full URLs** (was Important): the code now strips protocol/path to extract hostname before matching `EDS_HOST_PATTERN`, and returns 400 if it fails. Fixed.
- **`existingOrgLabel` redundancy** (was Minor): removed in favor of `existingOrgName` which already has the same fallback. Fixed.
- **Removing the 409 check** in favor of the early return in `performAsoPlgOnboarding` (lines 367-370) is a clean architectural improvement - the idempotency guard is now in one place instead of duplicated at each call site.

## Issues

### Important (Should Fix)

**`alternateDomain` still retires the original record before validating the alternate domain** - `src/controllers/plg/plg-onboarding.js:1150-1151`

This was flagged in the prior review and is still unaddressed. `onboarding.setStatus(STATUSES.INACTIVE)` and `await onboarding.save()` execute before `performAsoPlgOnboarding` is called, which is where domain validation (`isSafeDomain`) runs. If the alternate domain fails validation - say, an admin typo passes `localhost`, an IP, or an unsafe domain - the original record is permanently retired with no rollback path. Recovery requires a manual database intervention. Fix: either call `isSafeDomain(siteConfig.alternateDomain)` before retiring the original record and return 400 if it fails, or defer the save until after `performAsoPlgOnboarding` succeeds.

---

**Race condition in concurrent-create path bypasses the new ONBOARDED guard** - `src/controllers/plg/plg-onboarding.js:380-387`

The early return at line 367 handles the case where a record is already ONBOARDED on the first lookup. But if a concurrent request creates and completes onboarding between the initial `findByImsOrgIdAndDomain` (returns null) and the `create` attempt (throws), the catch block at line 382 re-fetches the record and falls through without checking its status. If the re-fetched record is now ONBOARDED, the flow continues past it. Low probability, but the window is real. Fix: add a status check after the race-condition re-fetch:
```js
onboarding = await PlgOnboarding.findByImsOrgIdAndDomain(imsOrgId, domain);
if (!onboarding) throw createError;
if (onboarding.getStatus() === STATUSES.ONBOARDED) return onboarding;
```

---

**No direct test for the `performAsoPlgOnboarding` ONBOARDED early return via the `onboard()` entry point** - `src/controllers/plg/plg-onboarding.js:367-370`

The early return changes behavior for all callers of `performAsoPlgOnboarding`, but it is only tested indirectly through the bypass flow. A direct test in the `onboard` describe block - where `findByImsOrgIdAndDomain` returns an ONBOARDED record - would confirm the idempotency guarantee holds for the primary onboarding entry point and document the intended behavior. The test should verify that `Site.create` is not called.

### Minor (Nice to Have)

**AEM_EDGE `siteConfig.authorUrl` mutation** - `src/controllers/plg/plg-onboarding.js:1115`: the hostname normalization mutates the caller's `siteConfig` object. This is harmless in practice since `siteConfig` comes from request data, but a short comment noting the intentional normalization would help future readers.

**Enrollment guard asymmetry** - `moveSite` blocks if ANY enrollment exists (all products), while `revokeOnboardedSiteEnrollments` only removes ASO enrollments. A brief comment noting this is intentional would prevent a future "alignment fix" that breaks it.

**Revocation failure observability** - the catch at line 1074 logs a warning, but orphaned enrollments may go unnoticed in low-volume scenarios. A structured log field (e.g., `orphanedEnrollments: true`) would make these queryable in monitoring.

## Assessment

Ready to merge? With fixes

Reasoning: The commit makes real progress - the Critical null-guard and three Important issues from the prior review are now resolved. The remaining `alternateDomain` retire-before-validate issue was flagged previously and is still the most actionable data-integrity risk: an invalid alternate domain permanently retires a working onboarding record. This and the direct `onboard()` test for the ONBOARDED early return should be fixed before merge; the race condition can be addressed in a follow-up if the team accepts the low probability.

## Next Steps

1. Fix the `alternateDomain` retire-before-validate ordering - validate the alternate domain before calling `onboarding.setStatus(INACTIVE)`.
2. Add the direct test for the `performAsoPlgOnboarding` early return via the `onboard()` entry point.
3. Optionally add the ONBOARDED status check after the concurrent-create re-fetch.
4. Minor items are optional.

Kanishkavijay39 (approved):

PR #2172, opened 2026-04-10
fix(llmo-4010): preserve edgeDeployed flag when patching suggestion data

anagarwa:
## Summary
- In `patchSuggestion`, when updating `suggestion.data`, the existing `edgeDeployed` flag was being silently overwritten with whatever the caller sent (or `undefined` if they didn't include it)
- Now merges incoming `data` over existing data while preserving `edgeDeployed` from the stored suggestion

## Root Cause
The audit runner sets `suggestion.data.edgeDeployed = true` after successfully deploying a suggestion to the edge. When the LLMO UI later calls `PATCH /suggestions/:id` to update other fields (e.g. rank, status), it sends a partial `data` object without `edgeDeployed`, which overwrote the flag.

## Test plan
- [ ] All 379 existing unit tests pass
- [ ] `PATCH /suggestions/:id` with `{ data: { someField: 'value' } }` preserves existing `edgeDeployed` on the suggestion
- [ ] `PATCH /suggestions/:id` with `{ data: { edgeDeployed: false } }` still allows explicitly setting `edgeDeployed`

Fixes LLMO-4010

### DIFF
--- a/src/controllers/suggestions.js
+++ b/src/controllers/suggestions.js
@@ -662,7 +662,12 @@ function SuggestionsController(ctx, sqs, env) {
 
       if (data) {
         hasUpdates = true;
-        suggestion.setData(data);
+        const existingData = suggestion.getData() || {};
+        const mergedData = { ...data };
+        if (existingData.edgeDeployed != null) {
+          mergedData.edgeDeployed = existingData.edgeDeployed;
+        }
+        suggestion.setData(mergedData);
       }
 
       if (kpiDeltas) {


src/controllers/suggestions.js
```diff
@@ -662,7 +662,12 @@ function SuggestionsController(ctx, sqs, env) {
 
       if (data) {
         hasUpdates = true;
-        suggestion.setData(data);
```
dipratap: This is actually fine.. if someone sends data object then it should replace the complete current data object.

nit23uec (changes requested):
this does not addresses the issue in the linked jira where the issue is about the audit flow not preserving state of already edge deployed suggestions (that includes updating shouldOptimize flag incorrectly).
patch suggestion is invoked through API and it should always overwrite any data that is provided as input (this is not a bug but desired behavior).

anagarwa:
Closing this PR — the fix is at the wrong layer. The `patchSuggestion` API contract must always overwrite suggestion data; preserving `edgeDeployed` there breaks that contract. The correct fix is in `spacecat-audit-worker`: each audit handler's `mergeDataFunction` should skip re-syncing a suggestion that already has `edgeDeployed: true`. Opening a new PR there instead (LLMO-4010). PR #2183 already handled the domain-wide case; this will cover the individual URL case.

anagarwa:
Closing this PR — a replacement has been opened at the correct layer.

**New PR:** https://github.com/adobe/spacecat-audit-worker/pull/2429

As noted in the review by @nit23uec, `patchSuggestion` is a generic API endpoint and must always overwrite with whatever the caller sends. The real fix belongs in the audit sync flow.

PR #2429 adds an `edgeDeployed` guard to the `mergeDataFunction` in three audit handlers (headings, TOC, faqs guidance) so that suggestions already deployed to the edge CDN are never overwritten by a re-audit run. PR #2183 already handled the domain-wide case in the prerender audit; #2429 covers individual URL suggestions.

PR #2174, opened 2026-04-10
feat(llmo-3354): add PATCH endpoint to clear edge status on 404 pages

anagarwa:
## Summary
- Adds `PATCH /sites/:siteId/page-citability/status` endpoint
- When called with a URL that returned HTTP 404, clears `isDeployedAtEdge = false` and resets `citabilityScore = 0` for the corresponding `PageCitability` record
- If the page isn't found in PageCitability (never tracked), returns `{ updated: false }` with no error
- If `httpStatus` is not 404, returns `{ updated: false }` (no-op for non-404 statuses)
- Route is protected by `site:write` capability

## Use Case
The edge worker observes when a previously-deployed page starts returning 404 (page deleted/moved). This endpoint lets the worker notify SpaceCat so the citability data remains accurate and the page is not shown as edge-deployed.

## Test plan
- [ ] All 233 unit tests pass
- [ ] `PATCH /sites/:siteId/page-citability/status` with `{ url: "...", httpStatus: 404 }` sets `isDeployedAtEdge=false` and `citabilityScore=0` on the record
- [ ] Same request for a URL not in PageCitability returns `{ updated: false }` with 200
- [ ] Request with `httpStatus: 200` returns `{ updated: false }` without touching the record

Fixes LLMO-3354

### DIFF
--- a/src/controllers/sites.js
+++ b/src/controllers/sites.js
@@ -1303,6 +1303,32 @@ function SitesController(ctx, log, env) {
     }
   };
 
+  const patchPageCitabilityStatus = async (context) => {
+    const { siteId } = context.params;
+    const { url, httpStatus } = context.data || {};
+
+    if (!hasText(url)) {
+      return badRequest('url is required');
+    }
+
+    if (httpStatus !== 404) {
+      return ok({ url, httpStatus, updated: false });
+    }
+
+    const { PageCitability } = dataAccess;
+    const record = await PageCitability.findByUrl(url);
+    if (!record) {
+      return ok({ url, httpStatus, updated: false });
+    }
+
+    record.setIsDeployedAtEdge(false);
+    record.setCitabilityScore(0);
+    await record.save();
+
+    log.info(`Cleared edge deployment status for 404 page: ${url} (site: ${siteId})`);
+    return ok({ url, httpStatus, updated: true });
+  };
+
   return {
     createSite,
     getAll,
@@ -1317,6 +1343,7 @@ function SitesController(ctx, log, env) {
     updateSite,
     updateCdnLogsConfig,
     getPageCitabilityCounts,
+    patchPageCitabilityStatus,
     getTopPages,
     resolveSite,
     getBrandProfile,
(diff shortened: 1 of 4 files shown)


anuj-adobe:
@anagarwa I understand a new API is being created but how would the entire flow be from UI. It is not clear to me. Would be useful if the PR description describes the entire flow from UI to backend.

anagarwa:
> @anagarwa I understand a new API is being created but how would the entire flow be from UI. It is not clear to me. Would be useful if the PR describes the entire flow from UI to backend.

Hey @anuj-adobe — good call, the PR description is missing the calling context. Here is the intended flow based on the Jira analysis:

**End-to-end flow**

1. A page (e.g. `https://lovesac.com/sactionals-reclining-seat`) was previously crawled, scored, and marked `isDeployedAtEdge = true` in PageCitability.
2. The source site removes or moves the page — it now returns 404 at origin.
3. The **edge worker** (Fastly/Cloudflare) intercepts a bot request for that URL, forwards it to origin, and receives a 404 response.
4. The edge worker calls `PATCH /sites/:siteId/page-citability/status` with `{ url, httpStatus: 404 }` using its service principal token.
5. SpaceCat clears `isDeployedAtEdge = false` and resets `citabilityScore = 0` for that record.
6. The page no longer shows as "Optimized" in the agentic traffic report on the next refresh.

**Open question this PR does not yet answer**

Whether step 4 is the right caller. The edge worker currently does not make SpaceCat API calls — it reads config but does not write back state. An alternative (and arguably simpler) approach would be to detect the 404 in the **audit worker** (spacecat-audit-worker), which already runs periodic crawls, has full site context, and can clear the flag directly without a new API surface.

The PR needs to confirm which component is the intended caller before it can be merged, and the PR description should document that decision.

**Technical gaps also flagged by Copilot (all addressable)**

- Missing `isValidUUID(siteId)` + `Site.findById` + `accessControlUtil.hasAccess` — same pattern as `getPageCitabilityCounts` in this file
- `PageCitability.findByUrl(url)` is not scoped to `siteId` — should use a site-scoped lookup or validate `record.getSiteId() === siteId`
- OpenAPI docs (`docs/openapi/site-api.yaml` or similar) not updated for the new route
- Behavioral unit tests missing — only the export whitelist was updated

---
*This reply was generated by OrcaFix — an AI-powered PR analysis tool built on Claude Code. Jira: LLMO-3354*

PR #2175, opened 2026-04-10
feat(llmo-3954): add IMS service principal token for CDN routing API calls

anagarwa:
## Summary
- Adds `getServicePrincipalToken(context)` to `support/utils.js` using `ImsClient.getServiceAccessTokenV3()` (client_credentials grant)
- `updateEdgeOptimizeCDNRouting` now tries the SP token first; falls back to the user promise token if `IMS_CLIENT_ID`/`IMS_CLIENT_SECRET` env vars are not configured
- This enables the LLMO backend to call the Internal CDN API on behalf of customers using its own service identity, without needing to pass a user token through the UI

## Required env vars
`IMS_HOST`, `IMS_CLIENT_ID`, `IMS_CLIENT_CODE`, `IMS_CLIENT_SECRET` (already available via Vault)

## Test plan
- [ ] All 346 llmo controller tests pass
- [ ] All 98 utils tests pass
- [ ] When SP env vars are configured: CDN routing call uses SP token (no `x-promise-token` header needed)
- [ ] When SP env vars are absent: falls back to promise token (existing behavior)

Fixes LLMO-3954

### DIFF
--- a/src/controllers/llmo/llmo.js
+++ b/src/controllers/llmo/llmo.js
@@ -34,7 +34,7 @@ import { Entitlement as EntitlementModel } from '@adobe/spacecat-shared-data-acc
 import TokowakaClient, { calculateForwardedHost } from '@adobe/spacecat-shared-tokowaka-client';
 import AccessControlUtil from '../../support/access-control-util.js';
 import { UnauthorizedProductError } from '../../support/errors.js';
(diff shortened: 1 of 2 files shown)


nit23uec:
this change is not required, its already being handled via https://github.com/adobe/spacecat-api-service/blob/main/src/controllers/llmo/llmo.js#L1394

PR #2176, opened 2026-04-10
feat(plg): displace already-onboarded domain when PLG suggestions are empty  

Kanishkavijay39:
Description:                                                                                                             
                                                                       
  Summary

  When a user tries to onboard a new domain for their IMS org but the org already has a different domain onboarded, the    
  current behavior is to unconditionally waitlist the new domain. This change adds smarter displacement logic: if the
  existing onboarded site has no open PLG suggestions (for cwv, alt-text, and broken-backlinks), it means the customer has nothing to act on for that domain, and the new domain should take over. Customer can use other domain.

  New behavior:
  - Check if the already-onboarded site has any NEW or IN_PROGRESS suggestions for the three PLG opportunity types (cwv,
  alt-text, broken-backlinks)                                                                                              
  - If empty → displace the old domain: set it to WAITLISTED with a "Displaced by new domain" reason, revoke all its
  SiteEnrollment records, and continue onboarding the new domain normally                                                  
  - If open suggestions exist → keep existing behavior: waitlist the new domain                                            
  - If the old record has no siteId or the suggestion check fails → conservative fallback: waitlist the new domain (no
  displacement)                                                                                                            
                                                                                                                           
  Changes
                                                                                                                           
  - src/controllers/plg/plg-onboarding.js — added hasOpenPlgSuggestions() helper and updated alreadyOnboarded block with   
  displacement logic
  - test/controllers/plg/plg-onboarding.test.js — added Opportunity, Suggestion, SiteEnrollment.allBySiteId mocks; 4 new   
  test cases covering: displacement on empty suggestions, no displacement with NEW suggestions, no displacement with       
  IN_PROGRESS suggestions, and conservative fallback on check failure

## Related Issues


Thanks for contributing!


src/controllers/plg/plg-onboarding.js
```diff
@@ -273,6 +273,55 @@ async function updateLaunchDarklyFlags(site, context) {
...
+    }
+
+    const suggestionLists = await Promise.all(
+      plgOpportunities.map((o) => Suggestion.allByOpportunityId(o.getId())),
+    );
+
+    // Block displacement if any PLG opportunity has open suggestions.
+    if (suggestionLists.some(
+      (suggestions) => suggestions.some(
+        (s) => s.getStatus() === 'NEW' || s.getStatus() === 'IN_PROGRESS',
```
absarasw: Could be FIXED, SKIPPED as well. 
Consider the case that the customer is using the site and used all suggestions. Now someone tried to cal plg-onboarding API using curl.

absarasw (approved):

PR #2188, opened 2026-04-11
fix: added demo org check

anshulk-public:
## Summary

Improves PLG onboarding waitlist messages for `DOMAIN_ALREADY_ASSIGNED` and
adds special handling for excluded orgs in the `moveSite` bypass flow.

---

### `isInternalOrg` helper

A new utility function to identify excluded orgs via the `ASO_PLG_EXCLUDED_ORGS`
environment variable:

```js
function isInternalOrg(orgId, env) {
  const excludedOrgs = (env.ASO_PLG_EXCLUDED_ORGS || '').split(',').map((id) => id.trim()).filter(Boolean);
  return excludedOrgs.includes(orgId);
}
```

## Related Issues

Thanks for contributing!


src/controllers/plg/plg-onboarding.js
```diff
@@ -90,6 +91,10 @@ function deriveCheckKey(onboarding) {
   return null;
 }
 
+function isInternalOrg(orgId, env) {
+  return orgId === env.DEFAULT_ORGANIZATION_ID || orgId === ASO_DEMO_ORG;
```
Kanishkavijay39: ASO_DEMO_ORG include AEM reference DEMO..

Can we create new env variable.. array which contains the SITES internal Ims, we can modify it later if want to extend the list
aso_plg__excluded_org.. something like this?

Kanishkavijay39 (changes requested):

Kanishkavijay39 (approved):

PR #2203, opened 2026-04-14
feat: add externally-triggered site detection API (SITES-42916)

papineni05:
## Summary

Replaces the old Coralogix-based site detection polling with an HTTP API for the DaaS team to trigger site detection directly from Fastly CDN log ingestion.

- Returns **409 Conflict** immediately if domain is already a known Site or SiteCandidate (early exit, no job created)
- Returns **202 Accepted** with `jobId` and `pollUrl` for async processing via `AsyncJob`
- Auth: existing S2S middleware (ScopedApiKeyHandler / JwtHandler)

---

### `POST /sites/detect/jobs`

Trigger an async site detection job. Called by the DaaS team when a new domain is detected from Fastly CDN log ingestion.

**Request**
```json
{
  "domain": "www.example.com",
  "hlxVersion": 5
}
```

| Field | Type | Required | Description |
|-------|------|----------|-------------|
| `domain` | string | Yes | Hostname to detect (e.g. `www.example.com`) |
| `hlxVersion` | integer | No | Helix version detected from CDN logs |

**Response — 202 Accepted**
```json
{
  "jobId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "status": "IN_PROGRESS",
  "createdAt": "2025-01-01T00:00:00.000Z",
  "pollUrl": "https://spacecat.experiencecloud.live/api/v1/sites/detect/jobs/3fa85f64-5717-4562-b3fc-2c963f66afa6"
}
```

**Response Statuses**

| Status | Condition |
|--------|-----------|
| `202 Accepted` | Job created, poll `pollUrl` for result |
| `400 Bad Request` | Missing/invalid `domain`, or `hlxVersion` is not an integer |
| `409 Conflict` | Domain already exists as a Site (AEM Edge) or SiteCandidate |
| `500 Internal Server Error` | Unexpected failure (SQS send failure, DB error) |

---

### `GET /sites/detect/jobs/:jobId`

Poll the status and result of a site detection job.

**Request**

No request body. `jobId` is a UUID path parameter.

**Response — 200 OK**
```json
{
  "jobId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "status": "COMPLETED",
  "createdAt": "2025-01-01T00:00:00.000Z",
  "updatedAt": "2025-01-01T00:01:00.000Z",
  "result": {
    "action": "created",
    "domain": "www.example.com",
    "baseURL": "https://www.example.com",
    "reason": null
  },
  "error": null
}
```

**Possible `status` values**

| Status | Meaning |
|--------|---------|
| `IN_PROGRESS` | Worker is processing the domain |
| `COMPLETED` | Terminal state; check `result.action` for the outcome |
| `FAILED` | Unexpected exception; safe to retry. Check `error.code` and `error.message` |

**Possible `result.action` values (when `status: COMPLETED`)**

| Action | Meaning |
|--------|---------|
| `created` | SiteCandidate successfully created |
| `duplicate` | Site or SiteCandidate already exists |
| `rejected` | Domain failed validation (IP address, ignored subdomain, not a Helix site) |

**Response Statuses**

| Status | Condition |
|--------|-----------|
| `200 OK` | Job found, returns current state |
| `400 Bad Request` | `jobId` is not a valid UUID |
| `404 Not Found` | No job found for given `jobId` |
| `500 Internal Server Error` | Unexpected DB error |

---

## Test Plan

- [ ] `POST /sites/detect/jobs` with valid new domain → 202, jobId in response
- [ ] `POST /sites/detect/jobs` with domain already existing as Site → 409
- [ ] `POST /sites/detect/jobs` with domain already existing as SiteCandidate → 409
- [ ] `POST /sites/detect/jobs` with missing `domain` → 400
- [ ] `POST /sites/detect/jobs` with invalid `hlxVersion` (non-integer) → 400
- [ ] `GET /sites/detect/jobs/:jobId` with valid jobId → 200 with status/result
- [ ] `GET /sites/detect/jobs/:jobId` for a `COMPLETED` job with `action: created` → result has `baseURL`
- [ ] `GET /sites/detect/jobs/:jobId` for a `COMPLETED` job with `action: duplicate` or `action: rejected` → result has `reason`
- [ ] `GET /sites/detect/jobs/unknown-uuid` → 404
- [ ] `GET /sites/detect/jobs/not-a-uuid` → 400

## Related
- Jira: [SITES-42916](https://jira.corp.adobe.com/browse/SITES-42916)
- Companion PR (audit-worker): [adobe/spacecat-audit-worker#2382](https://github.com/adobe/spacecat-audit-worker/pull/2382)


solaris007 (changes requested):
Hey @papineni05,

This review covers spacecat-api-service#2203 together with its companion spacecat-audit-worker#2382. Findings are split by repo where they are fixable; cross-cutting concerns are called out explicitly.

## Strengths

- Controller strips `details` from the error response at src/controllers/site-detection.js:393-403 (locked down at test/controllers/site-detection.test.js:808-821). Improvement over preflight which dumps `job.getResult()` / `job.getError()` verbatim.
- `isValidUUID(jobId)` validated before DB lookup at src/controllers/site-detection.js:371-373 - prevents malformed input from reaching the data layer.
- `POST /sites/detect` / `GET /sites/detect/jobs/:jobId` require `site:write` / `site:read` at src/routes/required-capabilities.js:232-233. Scopes declared alongside existing /sites routes.
- SQS failure rollback (remove + fallback to FAILED+save) at src/controllers/site-detection.js:329-347 - a caller never receives 202 for a job that was not enqueued.
- 409 early-exit at src/controllers/site-detection.js:300-313 is defense-in-depth against duplicate onboarding.
- Tests are substantial: both happy and error paths for controller, integration tests for validation and 409 conflict.

## Issues

### Important (Should Fix)

1. **SQS triple-fallback swallows the final save error and loses the original stack chain** (src/controllers/site-detection.js:329-347)
   - The innermost `catch (_) { /* best-effort */ }` at line 344 has no log. When SQS send, job.remove, AND job.save all fail, the job is left IN_PROGRESS in the DB with zero log trail. The outer 500 response hides the jobId, so the orphan is never polled or cleaned up. This is exactly the moment when observability matters most.
   - `throw new Error(\`Failed to send message to SQS: ${error.message}\`)` at line 346 loses the original error's stack chain.
   - Fix: add `log.error(\`Failed to mark orphan job ${job.getId()} as FAILED: ${e.message}\`)` inside the swallowed catch. Preserve the original error with `throw new Error('Failed to send message to SQS', { cause: error })`.

2. **Use `AsyncJob.Status.*` enum instead of string literals** (src/controllers/site-detection.js:318, 341)
   - The controller uses `'IN_PROGRESS'` and `'FAILED'` as strings. The companion worker uses `AsyncJob.Status.IN_PROGRESS` / `.FAILED` from `@adobe/spacecat-shared-data-access`. If enum values change upstream, the worker fails loudly at import time but the api-service silently writes invalid status to the DB.
   - Fix: `import { AsyncJob } from '@adobe/spacecat-shared-data-access'` and use the enum.

3. **URL shape is asymmetric and inconsistent with the preflight sibling** (docs/openapi/api.yaml:9-12, src/routes/index.js:216-217)
   - Preflight uses `POST /preflight/jobs` / `GET /preflight/jobs/:jobId` - symmetric. Site-detection uses `POST /sites/detect` / `GET /sites/detect/jobs/:jobId` - the `/jobs` segment appears only on GET. SDK generators and human readers both expect symmetric pairs.
   - Fix: either `POST /sites/detect/jobs` + `GET /sites/detect/jobs/:jobId` (matches preflight, preferred) or `POST /sites/detect` + `GET /sites/detect/:jobId`. Adjust OpenAPI refs, routes/index.js, required-capabilities.js, and the pollUrl builder at src/controllers/site-detection.js:353.

4. **No length or format cap on `domain` input** (src/controllers/site-detection.js:290)
   - `hasText(domain)` only checks for non-empty. An authenticated caller can submit a multi-megabyte `domain` string which is written verbatim to AsyncJob.metadata.payload and used in findByBaseURL queries - oversized JSONB rows, slow queries.
   - Fix: reject domains longer than 253 bytes (RFC 1035 cap). Validate format (no scheme, no path, no whitespace) so non-hostname payloads fail at the HTTP boundary rather than seconds later in the worker.

5. **Duplicate-detection logic is replicated in both repos with no single source of truth**
   - The same rule (composeBaseURL + Site.findByBaseURL + AEM_EDGE check + SiteCandidate.findByBaseURL) is implemented identically in api-service:src/controllers/site-detection.js:300-313 AND audit-worker:src/site-detection/handler.js:462-481. Two copies of the same business rule in two repos that evolve independently will drift.
   - Fix (preferred): remove the early-exit from api-service; let the worker own the decision. DaaS call volume is low enough that the extra AsyncJob row per duplicate is acceptable.
   - Fix (alternative): move the duplicate check to a shared helper in spacecat-shared-data-access that both sides import.

### Minor (Nice to Have)

1. **Raw `error.message` echoed in 500 responses** (src/controllers/site-detection.js:357, 402)
   - `internalServerError(error.message)` propagates SDK/DB error text directly. AWS SQS errors can include region, queue ARN segments, and AccessDenied fragments. Low impact because the endpoints are S2S-authenticated, but defense-in-depth: return a generic body and keep detail in log.error.

2. **Local `conflict()` helper with cross-repo TODO** (src/controllers/site-detection.js:233-236)
   - Other controllers (ims-org-access.js) inline `createResponse({ message }, 409)`. Either inline it the same way until the shared helper lands, or delete the TODO (track in an issue, not in shipping code).

3. **pollUrl hardcodes `spacecat.experiencecloud.live`** (src/controllers/site-detection.js:353)
   - Preview stages or alternate hosts would return misleading URLs. Consider deriving from the incoming request host or an env var.

## Recommendations

- After fixing Important item 1, add assertions for the new log in the triple-failure test at test/controllers/site-detection.test.js:751-758 (today it only asserts `resp.status === 500`).
- SQS-failure rollback logic (create job, send SQS, on-failure remove, fallback FAILED) is duplicated between preflight and site-detection controllers. As more async-job endpoints land, extract `createJobWithSqsTrigger(...)` helper into a shared support module.

- Infra: add `HLX_ADMIN_TOKEN`, `SLACK_SITE_DISCOVERY_CHANNEL_INTERNAL`; remove `CORALOGIX_API_KEY`, `SITE_DETECTION_WEBHOOK` (called out in PR body).
- The old hourly Coralogix poller had built-in liveness. The new pipeline depends on DaaS calling us. Add a Coralogix/CloudWatch alert on "zero POST /sites/detect calls in 24h" so a DaaS outage or misconfig does not silently disable site discovery.
- `conflict()` helper export from @adobe/spacecat-shared-http-utils.
- Backport the response-allowlist pattern to preflight's getPreflightJobStatusAndResult.

## Assessment

Ready to merge? With fixes.

Reasoning: The flow is sound, tests are thorough, and the design reuses the preflight async-job pattern correctly. The 5 Important items are all fixable in this repo and lock in public-contract shape before DaaS wires up.

## Next Steps

1. Address the 5 Important items first.
2. Review the 3 Minor items.
3. Recommendations and out-of-scope items can be separate PRs.

papineni05:
@solaris007 addressed the review. Summary of changes in [ecadfe14](https://github.com/adobe/spacecat-api-service/commit/ecadfe14):

**Important**
1. **SQS triple-failure path** — inner catch now logs `Failed to mark orphan job <id> as FAILED: <reason>`; original stack preserved via `new Error('...', { cause: sqsError })`; 500 bodies no longer echo raw `error.message`.
2. **`AsyncJob.Status.*` enum** — string literals `'IN_PROGRESS'` / `'FAILED'` replaced with the imported enum.
3. **URL symmetry with preflight** — `POST /sites/detect` → `POST /sites/detect/jobs`. Routes, required-capabilities, OpenAPI refs, routes test, and IT shared tests updated.
4. **Domain validation at the HTTP boundary** — new `isValidDomain` rejects whitespace, scheme (`://`), path (`/`), and strings over 253 bytes (RFC 1035). `hasText` alone is no longer considered sufficient.
5. **Duplicate-detection de-duplicated** — 409 early-exit removed; the worker now owns the rule. Callers always get `202 + jobId` and poll for `COMPLETED + action: duplicate`. OpenAPI dropped the 409 response and rewords the description.

**Minor**
- Raw `error.message` no longer echoed in 500 responses (both endpoints now return generic bodies; detail stays in logs).
- `conflict()` helper deleted (no longer needed).
- `pollUrl` hardcoded host is tracked as out-of-scope — will follow up separately.

**Out of scope / not addressed here**
- `conflict()` export from `@adobe/spacecat-shared-http-utils`, preflight GET response allowlisting, liveness alert on zero POST calls, and `hlxVersion`-as-top-level vs source-keyed envelope — separate PRs/tickets.

Companion worker changes in [adobe/spacecat-audit-worker#2382](https://github.com/adobe/spacecat-audit-worker/pull/2382).

solaris007 (commented):
Hey @papineni05,

Re-review of spacecat-api-service after your changes addressing the prior round. Reviewed in conjunction with the companion spacecat-audit-worker#2382.

All five prior Important findings are correctly resolved:

- **SQS triple-fallback** (src/controllers/site-detection.js:128-143): inner catch now logs orphan-job failure with jobId; rethrow preserves `{ cause: sqsError }`; 500 bodies are generic.
- **AsyncJob.Status enum** used consistently throughout the controller - no string literals on critical paths.
- **URL symmetry** with preflight: routes are now `POST /sites/detect/jobs` + `GET /sites/detect/jobs/:jobId`, matching the preflight pattern. OpenAPI, capabilities, and tests updated.
- **Domain validation** via `isValidDomain` (src/controllers/site-detection.js:33-44): rejects whitespace, scheme, path, and >253 bytes at the HTTP boundary.
- **Duplicate-detection** removed from api-service: no 409 path; callers always get 202+jobId; worker is sole authority.
- **Raw error.message** no longer echoed in 500 responses - all generic bodies with detail in logs.

## Issues

### Minor (Nice to Have)

1. **Duplicate result payload omits `baseURL`, making the imsOrgId lookup unreachable for `action: 'duplicate'`** (src/controllers/site-detection.js:189-201)
   - For the duplicate path, `result.baseURL` is absent (the worker doesn't set it for duplicates), so the controller's `imsOrgId` resolution branch is unreachable. Callers who want to know whether a duplicate is already linked to an org must issue a separate query.
   - The OpenAPI description is consistent with this behavior so it's not a regression, but it's an asymmetry worth noting.
   - Fix: either include `baseURL` in the worker's duplicate result payload, or have the controller derive it via `composeBaseURL(result.domain)` when the action is `duplicate`.

## Assessment

Ready to merge? Yes.

Reasoning: All prior Important findings are correctly addressed with clean, well-tested implementations. The one Minor is a UX asymmetry on the duplicate path, not a correctness issue.

papineni05:
Reverted commit a7b0bb1d (imsOrgId surface on the GET endpoint). Reasoning:

The async-poll contract had a sharp edge — `status: COMPLETED` is set at worker-finish time, but `imsOrgId` only resolves later, after the human Slack approval flow runs. A caller following standard "poll until terminal status, save result, stop" semantics would see `COMPLETED + imsOrgId: null` and stop polling, missing the value.

Cleaner separation: keep the async-job contract unambiguous (COMPLETED means done, full stop), and surface imsOrgId via a separate read path (e.g. `GET /sites/by-base-url` → `GET /organizations/:id`) when callers actually need it. That decouples the polling loop from the org-resolution flow.

Open question for follow-up: DaaS (S2S) callers would currently be 403'd by `accessControlUtil.hasAccess(site)` on the by-base-url path. If DaaS needs imsOrgId, we'll need either a separate S2S-friendly endpoint or an explicit `imsOrgIdStatus` field on this poll endpoint. Holding for direction.

solaris007 (approved):
Hey @papineni05,

The imsOrgId revert is clean and architecturally sound. COMPLETED now means done - no post-terminal mutation, no polling ambiguity for S2S callers.

## Strengths

- Previously flagged Minor (duplicate result payload omits baseURL for imsOrgId lookup) is now moot - the entire imsOrgId resolution feature was correctly reverted
- Revert is surgically clean: no imsOrgId remnants in the controller, OpenAPI schemas, unit tests, integration tests, seed data, or shared IDs
- The separation of concerns is the right architectural call: imsOrgId resolution is a human-in-the-loop process (Slack approval) on a fundamentally different timeline than async detection. Conflating them into one poll loop would have created a leaky abstraction. Existing `GET /sites/by-base-url` + `GET /organizations/:id` endpoints are the right read path when callers need the org
- Result projection is tight: `{action, domain, baseURL, reason}` with explicit field picking, no accidental leakage. OpenAPI matches 1:1

## Issues

### Minor (Nice to Have)

**PR description is stale after the revert.** The PR body still contains:
- Response example showing `"imsOrgId": "1234567890ABCDEF12345678@AdobeOrg"`
- A dedicated `result.imsOrgId` paragraph explaining eventual-consistency semantics
- Two test plan checkboxes about imsOrgId resolution

These no longer reflect the code at HEAD. Consider either removing the stale references or adding a note at the top that imsOrgId was reverted in `60946dc` (your comment thread already has the rationale).

## Recommendations

- The DaaS S2S integration gap you identified (403 on `GET /sites/by-base-url` due to `accessControlUtil.hasAccess`) is real but not a blocker for this PR. File a follow-up ticket and reference it in the PR description.

## Assessment

Ready to merge? **Yes.**

The code changes are clean across three review rounds. All prior Important findings remain correctly addressed. The imsOrgId revert is architecturally sound.

PR #2205, opened 2026-04-14
fix: geo-experiment name to include date

dipratap:
The Geo Experiment name will have format - 
Opportunity type-Date
eg. Prerender-12-04-2026


src/controllers/suggestions.js
```diff
@@ -1778,7 +1778,8 @@ function SuggestionsController(ctx, sqs, env) {
           siteId,
           opportunityId,
           type: GeoExperimentModel.TYPES.ONSITE_OPPORTUNITY_DEPLOYMENT,
-          name: context.data?.name || opportunity.getType(),
+          name: context.data?.name
+            || `${opportunity.getType().charAt(0).toUpperCase()}${opportunity.getType().slice(1)}-${new Date().toISOString().slice(0, 10)}`,
```
nit23uec, line 1782: it would be better if we could get the total exp count also and append it to the current date. - not a blocker for this PR.

nit23uec (approved):

PR #2206, opened 2026-04-14
feat(llmo-3476): validate suggestion IDs are valid UUIDs in deployToEdge

dipratap:
## Summary

The `POST /sites/:siteId/suggestions/deploy-to-edge` endpoint accepts a `suggestionIds` array but did not validate that each entry was a well-formed UUID before proceeding to DB lookups and downstream Tokowaka calls. This opened the door to unexpected behavior when non-UUID values (e.g. full page URLs accidentally passed as suggestion IDs) made it through to data-access layer queries.

Changes:
- Filters `suggestionIds` against the existing `isValidUUID` utility (already used throughout the controller for `siteId` / `opportunityId` validation)
- Returns HTTP 400 with a descriptive message listing the invalid IDs if any are found
- Logs `[edge-deploy-failed]` with the invalid IDs for observability, consistent with the existing warning pattern in this function

## Test plan
- [ ] All existing unit tests pass
- [ ] `POST /sites/:siteId/suggestions/deploy-to-edge` with a valid UUID array proceeds normally (no regression)
- [ ] Request with `{ suggestionIds: ["not-a-uuid", "https://example.com/page"] }` → 400, response body lists both invalid values
- [ ] Request with a mix of valid UUIDs and invalid values → 400, response body lists only the invalid entries

Fixes [LLMO-3476](https://jira.corp.adobe.com/browse/LLMO-3476)

### DIFF
--- a/src/controllers/suggestions.js
+++ b/src/controllers/suggestions.js
@@ -797,11 +797,11 @@ function SuggestionsController(ctx, sqs, env) {
       const {
         id, status, skipReason, skipDetail,
       } = item;
-      if (!hasText(id)) {
+      if (!isValidUUID(id)) {
         return {
           index,
-          uuid: '',
-          message: 'suggestion id is required',
+          uuid: id || '',
+          message: 'suggestion id must be a valid UUID',
           statusCode: 400,
         };
       }
@@ -1093,6 +1093,10 @@ function SuggestionsController(ctx, sqs, env) {
     // suggestion-based flow (assess, fix, etc.)
     if (!isArray(suggestionIds)) {
       return badRequest('Request body must be an array of suggestionIds');
+    }
+    const invalidAutoFixIds = suggestionIds.filter((id) => !isValidUUID(id));
+    if (invalidAutoFixIds.length > 0) {
+      return badRequest(`suggestionIds must contain valid UUIDs. Invalid: ${invalidAutoFixIds.join(', ')}`);
     }
     if (variations && !isArray(variations)) {
       return badRequest('variations must be an array');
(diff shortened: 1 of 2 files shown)


src/controllers/suggestions.js
```diff
@@ -1607,6 +1607,12 @@ function SuggestionsController(ctx, sqs, env) {
     }
     const suggestionIds = [...new Set(rawSuggestionIds)];
 
+    const invalidIds = suggestionIds.filter((id) => !isValidUUID(id));
```
nit23uec, line 1619: To make it complete: we should have this check in every API that requires/can have suggestionId(s) as input

anagarwa:
Review feedback addressed — extended UUID validation to all suggestionId(s) endpoints.

Per @nit23uec's comment "we should have this check in every API that requires/can have suggestionId(s) as input", this commit extends UUID validation beyond deploySuggestionToEdge.

**Changes in this update:**

1. **autofixSuggestions** — added UUID validation for the suggestionIds body array after the isArray check. Returns 400 listing all invalid IDs.

2. **previewSuggestions** — same UUID validation with [edge-preview-failed] log prefix, consistent with other preview validation log lines.

3. **rollbackSuggestionFromEdge** — same UUID validation with [edge-rollback-failed] log prefix.

4. **patchSuggestionsStatus** — replaced hasText(id) with isValidUUID(id) for per-item validation, so non-UUID suggestion IDs (including page URLs) are rejected with a clear message instead of silently being passed to Suggestion.findById().

**Path-param endpoints** (getByID, patchSuggestion, getSuggestionFixes, removeSuggestion) already had isValidUUID checks — no changes needed there.

**Tests added:** 7 new test cases covering all-invalid, mixed valid/invalid, and per-item UUID scenarios across all 4 functions. Also updated 1 existing patchSuggestionsStatus assertion to match the new error message.

PR #2210, opened 2026-04-14
feat: SITES-43069 - pass enabled preflight audits from Configuration to Mysticat

GeezerPelletier:
## Summary
- Beta preflight endpoint reads enabled audit handlers from SpaceCat global Configuration
- Performs the same three-gate enablement check as the audit worker’s `isAuditEnabledForSite`:
  1. `preflight` handler exists in Configuration with `productCodes`
  2. Site has valid enrollment for those product codes via `TierClient.createForSite`
  3. Handler is enabled for the site via `isHandlerEnabledForSite`
- If any gate fails, job is created then set to CANCELLED with a descriptive reason (matching audit worker pattern)
- Filters to `*-preflight` handler types and strips the suffix before sending to Mysticat (e.g. `headings-preflight` → `headings`)
- Audit names passed as `audits` field to Mysticat `/v1/preflight/analyze` — Mysticat skips any it hasn’t implemented yet
- Configuration load failure or null → 500 (no fallback to running all audits)
- Empty audit list (all disabled) → sends `audits: []`, Mysticat runs nothing
- Same toggle mechanism as ASO: `enableHandlerForSite` / `disableHandlerForSite` controls both

JIRA: [SITES-43069](https://jira.corp.adobe.com/browse/SITES-43069)
Companion: [SITES-43070](https://jira.corp.adobe.com/browse/SITES-43070) (Mystique side)
Mystique PR: [experience-platform/mystique#1597](https://git.corp.adobe.com/experience-platform/mystique/pull/1597)

## Test plan
- [x] Passes filtered and suffix-stripped preflight audits to Mysticat
- [x] Passes all preflight audits when all handlers enabled
- [x] Sends empty audits list when no preflight-* handlers enabled
- [x] Cancels job when preflight handler not found in Configuration
- [x] Cancels job when preflight handler has no product codes
- [x] Cancels job when entitlement check throws
- [x] Cancels job when site has no valid enrollment for product codes
- [x] Cancels job when preflight handler not enabled for site
- [x] Cancels job when no individual preflight audits enabled
- [x] Returns 500 when Configuration.findLatest fails
- [x] Returns 500 when Configuration.findLatest returns null
- [x] Cancelled jobs include ci/v1 pollUrl based on environment
- [x] 100% code coverage on preflight.js


johobot (approved):

GeezerPelletier:
### Updates since last review

**fix: use correct -preflight suffix convention** (`c8ab07e3`)
Configuration handler types are suffixed with `-preflight` (e.g. `headings-preflight`), not prefixed. Fixed the filter from `startsWith` to `endsWith` and strip logic to match the audit worker’s naming convention.

**fix: fail fast on missing config, send empty audits explicitly** (`8879ab1c`)
- `Configuration.findLatest()` failure or null → return 500 (no fallback to running all audits)
- Empty preflight audit list (all disabled) → send `audits: []` explicitly so Mystique runs nothing

**fix: cancel job when preflight not enabled or no audits enabled** (`ced6e0b5`)
Mirrors the audit worker’s `async-job-runner` cancellation pattern. Job is always created first (so callers can poll), then set to CANCELLED with a reason if:
- `preflight` handler is not enabled for the site → `"preflight audits disabled for site {siteId}"`
- All individual `-preflight` audits are disabled → `"all individual preflight audits disabled for site {siteId}"`

Both checks use the same `Configuration.isHandlerEnabledForSite` / `getEnabledAuditsForSite` APIs that the audit worker uses, keeping the two paths consistent.

100% code coverage on preflight.js across all new paths.

GeezerPelletier:
### Latest update: product code entitlement check (`4773fc77`)

The beta endpoint now performs the same three-gate enablement check as the audit worker’s `isAuditEnabledForSite` in `async-job-runner.js`:

1. **Handler exists with productCodes** — `configuration.getHandlers()["preflight"]` must exist and have a non-empty `productCodes` array
2. **Site has valid enrollment** — `TierClient.createForSite(context, site, productCode).checkValidEntitlement()` must return `siteEnrollment: true` for at least one product code
3. **Handler enabled for site** — `configuration.isHandlerEnabledForSite("preflight", site)` must return true

If any gate fails, the job is created (so callers can poll) then set to CANCELLED with a descriptive reason, matching the audit worker’s pattern.

Previously the beta endpoint only checked gate 3 (`isHandlerEnabledForSite`), which meant a site without proper enrollment could pass the beta check but fail in the worker path.

New tests cover: handler not found, no product codes, entitlement check throws, no valid enrollment, handler disabled for site. 100% code coverage.

bhellema:
Okay @GeezerPelletier , once the issue on dev has been resolved wrt to the resources running out, then let's double check that things are still okay and then merge when you have confirmed this.

PR #2214, opened 2026-04-15
feat: prompt details add mentions and competitors data

vivesing:
### Summary

Improves **Brand Presence topic detail**, **prompt detail**, and **topic prompts** so they support **topic UUID paths**, return a stable **`topicId`**, surface **competitor and mention strings** on each execution, and expose **`promptId`** and **`executionId`** on each execution row. **`prompt-detail-api.md`** is updated to match.

---

#### `:topicId` path

- After `decodeURIComponent`, if the segment is a **UUID**, filters use **`topic_id`**; otherwise **`topics`** (display name).
- Applies to **topic prompts**, **topic detail**, and **prompt detail**.

#### Topic & prompt detail JSON

- **`topic`**: Human-readable label when `topics` exists on rows; otherwise the path value (e.g. UUID when there is no label).
- **`topicId`**: Prefer first row’s **`topic_id`**; else the path when it is a valid UUID; else **`null`**.

#### `executions[]` (topic + prompt detail)

| Field | Source (`brand_presence_executions`) | Notes |
|--------|--------------------------------------|--------|
| `promptId` | `prompt_id` | String; `''` if null |
| `executionId` | `id` | String; `''` if null |
| `businessCompetitors` | `business_competitors` | `''` if null |
| `detectedBrandMentions` | `detected_brand_mentions` | `''` if null |

#### PostgREST

- **`DETAIL_SELECT`** includes `id`, `topic_id`, `prompt_id`, `business_competitors`, `detected_brand_mentions`, and existing detail columns so the API can populate the fields above.

---

### Code changes

**`src/controllers/llmo/llmo-brand-presence.js`**

- **`applyTopicPathFilter(query, decodedTopicParam)`** (exported): UUID → `eq('topic_id', …)`, else `eq('topics', …)`.
- **`topicIdForDetailResponse(rows, decodedTopicParam)`** (exported): stable topic UUID for JSON.
- **`topicLabelForDetailResponse(rows, decodedTopicParam)`**: display `topic` string.
- **`createTopicPromptsHandler`**: uses `applyTopicPathFilter` instead of always `eq('topics', topicName)`.
- **`createTopicDetailHandler` / `createPromptDetailHandler`**: path filter; `topic` + `topicId` on responses; execution objects include `promptId`, `executionId`, `businessCompetitors`, `detectedBrandMentions`.

---

### Tests

**`test/controllers/llmo/llmo-brand-presence.test.js`**

- UUID vs name path behavior.
- `topicId` / label helpers.
- Execution fields: `promptId`, `executionId`, competitors, mentions, and empty fallbacks.

---

### Docs

**`docs/llmo-brandalf-apis/prompt-detail-api.md`**

- Path parameter semantics (name vs UUID).
- Sample JSON: `topicId`, `promptId`, `executionId`, `businessCompetitors`, `detectedBrandMentions`.
- Top-level field reference and expanded **`executions[]`** table.
- Aggregation step notes aligned with `DETAIL_SELECT`.

---

### Related commits (branch history)

- `feat: prompt details add mentions and competitors data`
- `fix: allow passing topicIds in the prompt details api`

---

### Suggested conventional commit / PR title

`feat(llmo): brand presence detail — topic UUID path, topicId, execution ids, competitors/mentions`

### DIFF
--- a/docs/llmo-brandalf-apis/prompt-detail-api.md
+++ b/docs/llmo-brandalf-apis/prompt-detail-api.md
@@ -14,7 +14,7 @@ Returns all execution rows, weekly aggregated statistics, and citation sources f
 **Path parameters:**
 - `spaceCatId` — Organization ID (UUID)
 - `brandId` — `all` (all brands) or a specific brand UUID
-- `topicId` — URL-encoded topic name (e.g. `PDF%20Editing`)
+- `topicId` — URL-encoded topic **name** (e.g. `PDF%20Editing`) or a **topic UUID**; when a UUID, the API filters executions by `topic_id` and still returns a human-readable `topic` label when row data includes `topics`
 
 ---
 
@@ -45,6 +45,7 @@ GET /org/44568c3e-efd4-4a7f-8ecd-8caf615f836c/brands/all/brand-presence/topics/P
 ```json
 {
   "topic": "PDF Editing",
+  "topicId": "a1b2c3d4-e5f6-7890-abcd-ef1234567890",
   "prompt": "best pdf editor for mac",
   "region": "US",
   "stats": {
@@ -77,6 +78,8 @@ GET /org/44568c3e-efd4-4a7f-8ecd-8caf615f836c/brands/all/brand-presence/topics/P
   "executions": [
     {
       "prompt": "best pdf editor for mac",
+      "promptId": "019cb903-1184-7f92-8325-f9d1176af316",
+      "executionId": "019cb903-1184-7f92-8325-f9d1176af317",
       "region": "US",
       "executionDate": "2026-03-08",
       "week": "2026-W10",
@@ -90,7 +93,9 @@ GET /org/44568c3e-efd4-4a7f-8ecd-8caf615f836c/brands/all/brand-presence/topics/P
       "origin": "human",
       "category": "Acrobat",
       "sources": "https://example.com/pdf-editor",
-      "errorCode": ""
+      "errorCode": "",
+      "businessCompetitors": "Competitor A;Competitor B",
+      "detectedBrandMentions": "Acme Corp, OtherBrand"
     }
   ],
   "sources": [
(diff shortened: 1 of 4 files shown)


JayKid:
## 🔍 PR Review

Nice work — the helper extraction is clean, tests are thorough (336 new lines, 100% Codecov), and the schema alignment checks out against the `mysticat-data-service` migrations. UUID routing via `idx_bpe_topic_id` is a performance win too.

A few things to consider:

---

### Major: Duplicated execution-row mapper will drift

The `.map((r) => { ... })` block that converts DB rows to JSON is copy-pasted identically in `createTopicDetailHandler` (~L2190) and `createPromptDetailHandler` (~L2382). This PR adds 4 more fields (`promptId`, `executionId`, `businessCompetitors`, `detectedBrandMentions`) to both copies with the same logic.

Every future field addition must be synchronized in two places — and this is already the second time the mapper has grown identically. The file already has shared helpers for the same detail endpoints (`buildDetailExecQuery`, `aggregateWeeklyDetailStats`), so extracting a `mapExecutionRow(r)` function would follow the existing pattern and prevent drift.

---

### Minor: `createTopicPromptsHandler` now accepts UUIDs but doesn't return `topicId`

`applyTopicPathFilter` is applied to topic-prompts, so it now accepts UUID path params. But unlike the sibling handlers (topic-detail, prompt-detail), the prompts handler doesn't return `topicId` or a resolved `topic` label in its response.

Clients navigating via topic UUID will get filtered results but no identity metadata back — the three sibling topic endpoints now behave inconsistently. Worth either adding `topicId`/`topic` to the prompts response or documenting the intentional omission.

---

### Minor: `topicLabelForDetailResponse` has no direct test coverage

`topicIdForDetailResponse` is exported and directly unit-tested, but the sibling `topicLabelForDetailResponse` is private and only exercised indirectly through handler integration tests. The `topics: ''` (empty string) edge case isn't covered, and the prompt-detail handler is missing the "null topics fallback" test that topic-detail has.

Exporting it and adding a small unit-test block (mirroring the `topicIdForDetailResponse` tests) would make regressions easier to pinpoint.

JayKid (approved):

PR #2216, opened 2026-04-15
feat(monitoring): add GET /monitoring/drs-bp-pg-audit proxy endpoint for brand presence monitoring

rainer-friederich:
## Summary

Adds a thin proxy endpoint `GET /monitoring/drs-bp-pg-audit` that forwards `projection_audit` queries from DRS monitoring workers to PostgREST inside the Mysticat VPC. DRS Lambda runs in a separate AWS account (489975610310) and cannot reach `http://data-svc.internal` (private Route53 in account 682033462621). This endpoint bridges the gap.

**Auth:** DRS workers authenticate via admin `x-api-key`. The `drsBpPgAudit:read` capability in `required-capabilities.js` gates future S2S JWT callers; it does not affect admin key access.

## Changes

- **`src/controllers/drs-bp-pg-audit.js`** — new controller
  - Params: `siteId` (UUID, required), `dateStart` / `dateEnd` (YYYY-MM-DD, required, dateEnd > dateStart), `handlerName` (default `wrpc_import_brand_presence`), `limit` (1–500, default 500), `offset` (default 0)
  - Input validation: UUID format check on `siteId`, date format + chronological order checks, `limit` floored at 1 to prevent `range(0, -1)`
  - PostgREST error message sanitized — not leaked in 500 response body
  - Response: `{ rows: [...], hasMore: bool }` — `hasMore=true` when `rows.length >= limit`
  - Queries `projection_audit` with `skipped=false`, ordered `projected_at DESC`
- **`src/routes/index.js`** — route registration
- **`src/routes/required-capabilities.js`** — `drsBpPgAudit:read` capability with note explaining intentional admin-key auth for DRS
- **`src/index.js`** — controller wiring
- **`docs/openapi/monitoring-api.yaml`** — new OpenAPI spec for the endpoint (response schema, all params, auth/index notes)
- **`docs/openapi/api.yaml`** — path reference to new spec file
- **`docs/openapi/tools-api.yaml`** — removed stale entry (moved to monitoring-api.yaml)

## Test plan

- [x] `npm test` — 8075 passing, 100% branch coverage
- [x] Unit tests in `test/controllers/drs-bp-pg-audit.test.js` — 23 tests covering: all validation paths, UUID/date format, dateEnd ≤ dateStart, limit=0 floor, hasMore=true/false, error sanitization, pagination params
- [x] IT tests in `test/it/postgres/drs-bp-pg-audit.test.js` — 9 tests covering: 401 without auth, all 400 validation paths (no PostgREST required)
- [x] Companion PR in llmo-data-retrieval-service: [adobe-rnd/llmo-data-retrieval-service#1306](https://github.com/adobe-rnd/llmo-data-retrieval-service/pull/1306) — Python client updated for `{ rows, hasMore }` envelope


src/controllers/drs-bp-pg-audit.js
```diff
@@ -0,0 +1,115 @@
...
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { badRequest, internalServerError, ok } from '@adobe/spacecat-shared-http-utils';
+
+const MAX_LIMIT = 500;
+const DEFAULT_HANDLER_NAME = 'wrpc_import_brand_presence';
+const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+const DATE_RE = /^\d{4}-\d{2}-\d{2}$/;
```
solaris007: **Important:** This regex accepts impossible dates like `2025-13-45` or `2025-02-30`. The codebase already has `isValidDateInterval()` in `src/utils/date-utils.js` which validates format, semantic correctness, and chronological order. Use that instead of the hand-rolled regex.

src/controllers/drs-bp-pg-audit.js
```diff
@@ -0,0 +1,115 @@
...
+      return internalServerError('PostgREST client not available');
+    }
+
+    const { params = {} } = reqContext;
+    const {
+      siteId,
+      dateStart,
+      dateEnd,
+      handlerName: handlerNameParam,
+    } = params;
```
solaris007, line 92: **Important:** Every other user-supplied parameter is validated, but `handlerName` accepts any string. Add a pattern constraint matching the actual handler naming convention, e.g.:
```js
const HANDLER_RE = /^[a-z][a-z0-9_]{0,99}$/;
if (!HANDLER_RE.test(handlerName)) {
  return badRequest('Invalid handlerName');
}
```

src/index.js
```diff
@@ -248,6 +249,7 @@ async function run(request, context) {
     const featureFlagsController = FeatureFlagsController(context);
     const autofixChecksController = AutofixChecksController(context);
     const pageRelationshipsController = PageRelationshipsController(context);
+    const drsBpPgAuditController = DrsBpPgAuditController();
```
solaris007: **Important:** Every other controller is instantiated as `Controller(context)`. This one takes no arguments, creating a second pattern for dependency injection. Accept `context` in the factory to match the established convention.

src/controllers/drs-bp-pg-audit.js
```diff
@@ -0,0 +1,115 @@
...
+
+    if (error) {
+      reqContext.log?.error('projection_audit query failed', { errorMessage: error.message });
+      return internalServerError('projection_audit query failed');
+    }
+
+    const rows = data ?? [];
+    return ok({ rows, hasMore: rows.length >= limit });
+  };
+
```
solaris007, line 150: **Important:** Returning raw `projection_audit` rows creates tight coupling between the API contract and DB schema. A thin DTO mapping decouples the two and makes schema evolution safe.

solaris007 (changes requested):
Hey @rainer-friederich,

## Strengths

- **Thorough OpenAPI spec** (`docs/openapi/monitoring-api.yaml`): Documents auth model, server-side filters, pagination semantics, and composite index requirement. This is better documentation than most endpoints in this repo - consumers can use this without reading source code.
- **Strict input validation** (`src/controllers/drs-bp-pg-audit.js:17`): UUID regex on `siteId`, date format checks, chronological ordering, limit clamping with floor at 1 - all the right guards are in place.
- **Error message sanitization** (`src/controllers/drs-bp-pg-audit.js:107-109`): PostgREST errors logged server-side, generic message returned to caller. Explicitly verified by test at `test/controllers/drs-bp-pg-audit.test.js:228-234`.
- **Explicit column selection** (`src/controllers/drs-bp-pg-audit.js:98`): No `SELECT *` - limits data exposure surface.
- **Intentionally narrow scope**: Read-only endpoint with fixed column projection and server-side filter enforcement (`skipped=false`). Not an open-ended PostgREST passthrough.
- **Solid test coverage**: 23 unit tests + 9 integration tests covering validation paths, auth rejection, pagination edge cases, and error sanitization.
- **Capability documentation** (`src/routes/required-capabilities.js:392-395`): Comment explaining the admin-key vs S2S JWT distinction saves future readers from confusion.

## Issues

### Important (Should Fix)

**1. Date validation accepts impossible calendar dates** - `src/controllers/drs-bp-pg-audit.js:18`

The regex `/^\d{4}-\d{2}-\d{2}$/` validates format but not semantic validity. Inputs like `2025-13-45`, `2025-02-30`, or `0000-00-00` pass validation and get forwarded to PostgREST as `2025-13-45T00:00:00Z`. This codebase already has `isValidDateInterval(startDate, endDate)` in `src/utils/date-utils.js` which validates format, semantic correctness, ensures startDate < endDate, and checks max range. Use the existing utility - you get correct validation for free and stay consistent with other endpoints.

**2. `handlerName` parameter has no validation** - `src/controllers/drs-bp-pg-audit.js:64`

Every other user-supplied parameter is validated (siteId with UUID regex, dates with format checks, limit/offset with numeric parsing), but `handlerName` accepts any arbitrary string and is passed directly to `.eq('handler_name', handlerName)`. The PostgREST client parameterizes values so this is not a SQL injection risk, but defense-in-depth matters on a proxy endpoint that bridges network boundaries. At minimum, constrain it to a reasonable pattern:
```js
const HANDLER_RE = /^[a-z][a-z0-9_]{0,99}$/;
if (!HANDLER_RE.test(handlerName)) {
  return badRequest('Invalid handlerName');
}
```

**3. Controller instantiation diverges from established pattern** - `src/index.js:252`

Every other controller (~40 of them) is instantiated as `Controller(context)`. This one is `DrsBpPgAuditController()` with no arguments, reading `reqContext.dataAccess` at call time instead. It works, but introduces a second pattern. In 6 months someone will either "fix" it thinking it's a bug, or copy it into a new controller. Accept `context` in the factory to match the established convention.

**4. Query construction partially unverified in tests** - `test/controllers/drs-bp-pg-audit.test.js`

The unit tests verify `from`, `eq(handler_name)`, `gte`, `lt`, and `range` calls, but three query chain calls have no assertions: `.select(columns)`, `.eq('skipped', false)`, and `.order('projected_at', { ascending: false })`. These are the most fragile parts of the query - someone removing the skipped filter or changing sort direction would not break any test. Add a test asserting the full query shape.

**5. No happy-path integration test for PostgREST round-trip** - `test/it/shared/tests/drs-bp-pg-audit.js`

The IT tests cover auth and validation only. Other IT suites in this repo seed data and test actual retrieval. The PostgREST query path (table name, column names, filter operators, `.range()` pagination) is only tested against stubs. Even a smoke test that gets a 200 with `rows: []` for a non-existent siteId would prove the query executes without error. (Can be a follow-up if table seeding is complex.)

**6. Raw table rows returned without DTO transformation** - `src/controllers/drs-bp-pg-audit.js:113`

`projection_audit` rows are returned directly from PostgREST with no transformation layer. This creates tight coupling between the API contract and the database schema - if columns are renamed or restructured, the endpoint's contract breaks silently. A thin DTO mapping decouples the two and makes schema evolution safe.

### Minor (Nice to Have)

**7. `hasMore` false positive on exact page boundary** - `src/controllers/drs-bp-pg-audit.js:113`

`rows.length >= limit` returns `true` when exactly `limit` rows exist, causing one extra empty request. The standard fix: fetch `limit + 1` rows, return at most `limit`, set `hasMore = fetched.length > limit`.

**8. No upper bound on date range** - A caller can request `dateStart=2020-01-01&dateEnd=2030-01-01` and paginate through entire history. Consider a max range (e.g., 90 days) if the monitoring use case permits.

**9. Missing tests for negative limit and negative offset** - The clamping logic correctly handles these, but neither edge case has a test.

## Recommendations

1. **Track networking follow-up**: This endpoint exists because DRS cannot reach PostgREST directly across AWS accounts. That is a networking problem solved at the application layer. File a ticket for VPC peering / PrivateLink between accounts and add a `TODO(SITES-XXXXX)` comment so the intent is visible. Without this, the proxy pattern will replicate for every cross-account data need.

2. **Track tenant-scoping for S2S JWT callers**: The `drsBpPgAudit:read` capability grants access to query any `siteId`. With admin-key auth this is fine, but when S2S JWT callers are onboarded, any consumer with this capability can read audit data for every site (cross-tenant access). Add a tracking comment so site-scoped authorization is implemented before the S2S transition.

3. **Review `metadata` field contents**: The `metadata` JSONB column is returned as-is. Verify it does not contain internal system details. If it does, either exclude it from the select or redact specific keys.

## Assessment

Ready to merge? With fixes.

The endpoint solves a real, immediate problem and the implementation is clean and well-documented. Issues 1-3 (date validation, handlerName validation, controller pattern) are straightforward code fixes. Issue 4 (query construction tests) is a small test addition. Issues 5-6 (happy-path IT, DTO) can be tracked as fast follow-ups if they require more work.

## Next Steps

1. Address the 3 code fixes first: use existing date utility, add handlerName validation, align controller constructor.
2. Add query construction assertions to unit tests.
3. Track happy-path IT, DTO layer, and networking follow-up as follow-up items.

solaris007 (approved):
Hey @rainer-friederich,

## Strengths

- Date validation now uses `isValidDateInterval()` from `src/utils/date-utils.js` with a 90-day max range cap (`src/controllers/drs-bp-pg-audit.js:15-16,100-106`)
- `handlerName` now validated with `/^[a-z][a-z0-9_]{0,99}$/` (`src/controllers/drs-bp-pg-audit.js:18,109-111`)
- Controller now follows `Controller(context)` pattern (`src/index.js:252`, `src/controllers/drs-bp-pg-audit.js:55-56`)
- DTO layer added via `toAuditDto()` - decouples API contract from DB schema, excludes `metadata` column (`src/controllers/drs-bp-pg-audit.js:24-31,139`)
- `hasMore` now uses `limit + 1` fetch strategy - no more false positives on exact boundaries (`src/controllers/drs-bp-pg-audit.js:115-116,135-137`)
- Query construction fully verified in tests: `select`, `eq('skipped', false)`, `order` all asserted (`test/controllers/drs-bp-pg-audit.test.js:304-317`)
- S2S tenant-scoping TODO documented in JSDoc (`src/controllers/drs-bp-pg-audit.js:45-48`)
- Test suite expanded from 23 to ~36 unit tests covering impossible dates, date range cap, handlerName validation, negative limit/offset, DTO mapping, and exact-boundary pagination

Clean and thorough response to all review feedback.

## Assessment

Ready to merge? Yes.

All previously flagged issues have been addressed. The only remaining item (happy-path IT with PostgREST round-trip) was acknowledged as a follow-up and does not block merge. Well done.

src/routes/required-capabilities.js
```diff
@@ -390,6 +390,12 @@ const routeRequiredCapabilities = {
   // Trigger — GET triggers side effect; consider POST for RFC 7231 semantics (follow-up)
   'GET /trigger': 'audit:write',
 
+  // Monitoring
+  // Note: DRS workers call this via admin x-api-key (not S2S JWT) — they run in a separate
+  // AWS account and do not hold an S2S consumer registration. The capability here gates
+  // future S2S JWT callers only; admin key access bypasses capability checks entirely.
+  'GET /monitoring/drs-bp-pg-audit': 'drsBpPgAudit:read',
```
ravverma, line 397: @rainer-friederich 
`drsBpPgAudit` is not a defined entity in data service like `site` , `organization` etc so
this capability `drsBpPgAudit:read` will not be allowed to assign to any S2S consumer.
As I can see it is for reading audits so would recommend to replace it with `audit:read`
cc: @solaris007

rainer-friederich: Good catch, thanks @ravverma! Fixed in #2228 — changed to `audit:read` which is already a defined entity. Could you take a quick look?

PR #2228, opened 2026-04-16
fix(auth): use audit:read capability for drs-bp-pg-audit endpoint

rainer-friederich:
## Summary

- Replace `drsBpPgAudit:read` with `audit:read` for the `/monitoring/drs-bp-pg-audit` route capability
- `drsBpPgAudit` is not a defined entity in data service, so the capability cannot be assigned to S2S consumers
- `audit:read` already exists and semantically matches — the endpoint reads `projection_audit` data

Addresses review feedback from #2216: https://github.com/adobe/spacecat-api-service/pull/2216#discussion_r3092459677

## Test plan

- [x] `npm test` — 8195 passing, 100% coverage
- [x] No behavioral change for admin key callers (capability checks bypassed)
- [x] S2S JWT callers can now be granted `audit:read` to access this endpoint


src/routes/required-capabilities.js
```diff
@@ -394,7 +394,7 @@ const routeRequiredCapabilities = {
   // Note: DRS workers call this via admin x-api-key (not S2S JWT) — they run in a separate
   // AWS account and do not hold an S2S consumer registration. The capability here gates
   // future S2S JWT callers only; admin key access bypasses capability checks entirely.
-  'GET /monitoring/drs-bp-pg-audit': 'drsBpPgAudit:read',
+  'GET /monitoring/drs-bp-pg-audit': 'audit:read',
```
solaris007: Least-privilege concern: every other `audit:read` route in this file is site-scoped domain audit data. This route is platform-scoped Postgres replication monitoring. Reusing the same capability string silently broadens the grant - any partner already holding `audit:read` for their own site audits gains a path to infra monitoring data.

Preferred fixes, in order:
1. Move to `INTERNAL_ROUTES` if no concrete S2S consumer exists (comment says DRS workers only call via admin x-api-key today).
2. Define a dedicated capability (`monitoring:read` or `infraAudit:read`) in the data service.
3. Minimum: extend the comment above to say explicitly that granting `audit:read` now also grants infra monitoring access, so the next reviewer granting audit:read understands the bundled scope.

solaris007 (changes requested):
Hey @rainer-friederich,

Thanks for addressing the PR 2216 feedback. The route-level change is correct and consistent with conventions elsewhere in the file. Flagging three should-fix items before merge.

## Strengths

- `src/routes/required-capabilities.js:397` is consistent with existing conventions - `audit:read` is used on 7 other audit-reading routes in this file.
- The surrounding comment at lines 394-396 is accurate: the `s2sAuthWrapper` passes through admin x-api-key callers before capability checks, so DRS workers are unaffected.
- Minimal diff, tight scope, clear PR description.
- `audit:read` is semantically appropriate for a GET endpoint reading `projection_audit` data.

## Issues

### Important (Should Fix)

1. **Stale `drsBpPgAudit:read` references in other files** - the whole rationale of this PR is that `drsBpPgAudit` is not a defined entity. Leaving it in developer-facing docs will mislead future S2S onboarding:
   - `docs/openapi/monitoring-api.yaml:20-21` - the auth description still tells API consumers to request `drsBpPgAudit:read`. This is the exact failure mode this PR fixes, now embedded in public docs.
   - `src/controllers/drs-bp-pg-audit.js:44` - JSDoc references `drsBpPgAudit:read`.
   - `src/controllers/drs-bp-pg-audit.js:48` - TODO instructs future engineers to look for `drsBpPgAudit:read` when onboarding S2S JWT callers.

   Fix: trivial rename across those three locations in this PR so the rename is complete.

2. **Least-privilege conflation at `src/routes/required-capabilities.js:397`** - every other `audit:read` route in this file is site-scoped domain audit data (SEO, accessibility, quality). `/monitoring/drs-bp-pg-audit` is platform-scoped Postgres replication monitoring. Reusing the same capability string silently broadens the grant: any partner who was given `audit:read` to read their own site audit results now also has a path to infra monitoring data.

   Options, in preferred order:
   1. Move this route to `INTERNAL_ROUTES` if there is no concrete S2S consumer use case (the comment says DRS workers only call it via admin x-api-key, so today this is just a placeholder).
   2. Define a purpose-specific capability (`monitoring:read` or `infraAudit:read`) in the data service.
   3. Minimum: extend the comment at 394-396 to say explicitly that granting `audit:read` now also grants access to infra monitoring, so the next reviewer granting audit:read to a partner understands the bundled scope.

3. **No assertion for this specific route-to-capability mapping** - `test/routes/required-capabilities.test.js` validates format and coverage, but does not pin any specific route to a specific capability. A regression could silently revert or change this value without a test failure. Two-line fix in the existing test file: `expect(routeRequiredCapabilities['GET /monitoring/drs-bp-pg-audit']).to.equal('audit:read')`.

### Minor (Nice to Have)

- `src/routes/required-capabilities.js:391-393` - comment could add a pointer to the authoritative capability list in the data service so future readers do not re-litigate the choice.
- The root cause here (a capability string that is schema-valid but not registered in the data service) is a class of bug that the existing tests do not catch. A follow-up allowlist test (every unique entity in `routeRequiredCapabilities` must appear in a known-entities set) or a startup-time cross-reference against the data-service registry would make this fail loudly.

## Recommendations

1. Fold the three stale-reference updates into this PR. Co-locating the rename keeps the intent reviewable in one changeset and prevents the public OpenAPI spec from shipping with the wrong capability name.
2. Decide whether this route is ever intended for S2S consumers. If not, moving it to `INTERNAL_ROUTES` is cleaner than an aspirational capability gate.
3. File a follow-up ticket for a known-entities allowlist test in `spacecat-shared` or locally; this would have caught the PR 2216 issue at review time.

## Assessment

Ready to merge? With fixes.

Reasoning: The core change is correct and runtime-safe - admin x-api-key callers bypass capability checks, so current behavior is unchanged. The blocker is completeness: three stale references to `drsBpPgAudit:read` remain in docs and controller JSDoc, one of them in a public API spec. The least-privilege concern about reusing `audit:read` for infra monitoring is worth a conscious decision before merge, not necessarily a redesign.

## Next Steps

1. Update the three stale `drsBpPgAudit:read` references and add the route-to-capability assertion in this PR.
2. Decide on the `/monitoring/drs-bp-pg-audit` placement (`INTERNAL_ROUTES` vs keep as `audit:read` with updated comment vs new dedicated capability).
3. Minor items and allowlist follow-up can ship separately.

rainer-friederich:
Thanks @solaris007 — addressed in 1829cbdf.

Went with option 1 (move to `INTERNAL_ROUTES`) since DRS workers call this via admin `x-api-key` today and there is no concrete S2S consumer. This also neutralizes the least-privilege concern without having to define a new capability that nobody consumes yet.

**Changes:**
- `src/routes/required-capabilities.js` — moved `GET /monitoring/drs-bp-pg-audit` from `routeRequiredCapabilities` to `INTERNAL_ROUTES`. Added a block comment explaining why it is kept internal and what to do when an S2S consumer appears (introduce `monitoring:read` rather than bundling into `audit:read`).
- `docs/openapi/monitoring-api.yaml` — rewrote the auth note: no more `drsBpPgAudit:read` reference; documents `INTERNAL_ROUTES` placement and the future-capability guidance.
- `src/controllers/drs-bp-pg-audit.js` — JSDoc and TODO now reference `INTERNAL_ROUTES` and the preferred `monitoring:read` future path, not the non-existent `drsBpPgAudit:read`.
- `test/routes/required-capabilities.test.js` — new `pinned mappings` describe block asserts the route is in `INTERNAL_ROUTES` and **not** in `routeRequiredCapabilities`. A silent regression (e.g. granting `audit:read` back) fails loudly in review.

**Deferred (happy to file follow-up tickets):**
- Known-entities allowlist test cross-referencing against the data-service capability registry — good catch, broader than this PR.
- Trigger / other minor capability comment pointers — separate tidy-up.

Test run: 8189 passing, 100% coverage, `docs:lint` clean, `docs:build` succeeds.

test/routes/required-capabilities.test.js
```diff
@@ -69,6 +69,24 @@ describe('routeRequiredCapabilities', () => {
...
+      const route = 'GET /monitoring/drs-bp-pg-audit';
+      expect(INTERNAL_ROUTES).to.include(
+        route,
+        'DRS Brand Presence PG audit is admin-key only; bundling into audit:read would silently '
+        + 'broaden that site-scoped capability to infra monitoring data.',
+      );
+      expect(routeRequiredCapabilities).to.not.have.property(
+        route,
+        'DRS Brand Presence PG audit must not be mapped to an S2S capability until a dedicated '
+        + 'monitoring capability exists.',
```
solaris007: Chai API misuse: the second positional argument to `.property(key, val)` is the expected value for strict equality, not a failure message. Under `.not`, this passes when the property is absent OR when the value differs from `val`. If someone re-adds `'GET /monitoring/drs-bp-pg-audit': 'audit:read'` to `routeRequiredCapabilities` while leaving it in `INTERNAL_ROUTES`, this assertion silently passes - the exact regression the test advertises to catch.

One-line fix - move the explanation to the `expect()` message position:

```js
expect(
  routeRequiredCapabilities,
  'DRS Brand Presence PG audit must not be mapped to an S2S capability until a dedicated monitoring capability exists.',
).to.not.have.property(route);
```

solaris007 (approved):
Hey @rainer-friederich,

Clean fix. The move to `INTERNAL_ROUTES` is the right mechanism, the stale references are cleaned up across OpenAPI, controller JSDoc, and config, and the forward-looking guidance (prefer a dedicated monitoring capability over re-bundling into `audit:read`) is mirrored consistently in all three places a future engineer would look.

Approving. One item below is worth fixing before merge because the new regression guard is not currently catching what it advertises - see inline comment for the one-line chai change. Not blocking the approval, but strongly recommended.

## Strengths

- **Least-privilege conflation resolved.** `s2sAuthWrapper` denies S2S JWT callers via deny-by-default - absence from `routeRequiredCapabilities` returns 403 per `spacecat-shared-http-utils/src/auth/s2s-wrapper.js`. The admin `x-api-key` path is unchanged because it never enters the S2S code path, so DRS workers keep working.
- **Stale `drsBpPgAudit:read` references** in `docs/openapi/monitoring-api.yaml`, `src/controllers/drs-bp-pg-audit.js`, and the route config are all cleaned. Grep across the repo returns zero remaining hits.
- **Pinned-mapping test added** - assertion 1 correctly pins the route in `INTERNAL_ROUTES` (assertion 2 has a chai issue, see inline comment).

Also worth calling out:
- The `INTERNAL_ROUTES` entry comment at `src/routes/required-capabilities.js:148-153` is the best-justified rationale block in the list: explains the admin-key-only reason, the cross-account constraint, and pre-warns the next engineer against re-bundling into `audit:read`.
- Controller JSDoc TODO is concretely actionable: specific trigger, specific actions, specific capability name.
- Integration tests in `test/it/shared/tests/drs-bp-pg-audit.js` only use `http.admin`, so the INTERNAL_ROUTES move does not break any existing test path.

## Issues

### Suggested pre-merge fix

**Pinned-mapping test: second assertion is a no-op due to chai API misuse.**

`test/routes/required-capabilities.test.js:81-85`

```js
expect(routeRequiredCapabilities).to.not.have.property(
  route,
  'DRS Brand Presence PG audit must not be mapped to an S2S capability ...',
);
```

In chai, the second positional argument to `.property(key, val)` is the expected value for strict equality, not a failure message. Under `.not`, the assertion passes when the property is absent OR when the value differs from `val`. If someone re-adds `'GET /monitoring/drs-bp-pg-audit': 'audit:read'` to `routeRequiredCapabilities` while leaving the route in `INTERNAL_ROUTES`, this assertion silently passes - the exact "silent broadening" the test was added to prevent.

One-line fix:
```js
expect(
  routeRequiredCapabilities,
  'DRS Brand Presence PG audit must not be mapped to an S2S capability until a dedicated monitoring capability exists.',
).to.not.have.property(route);
```

### Minor (Nice to Have)

- `src/routes/required-capabilities.js:148` - the new comment header uses an em-dash ("Monitoring - DRS..."), while every other `INTERNAL_ROUTES` group header uses a normal hyphen. Trivial.
- Consider adding a structural invariant test that asserts `INTERNAL_ROUTES` and `Object.keys(routeRequiredCapabilities)` are disjoint. This subsumes the narrower pinned check and prevents the same class of bug for any future internal-only route. Cheap follow-up.
- The forward-looking `monitoring:read` name is a domain bucket, not a resource. Every other capability in this file is resource-scoped (`audit`, `site`, `opportunity`, `apiKey`). If two unrelated monitoring endpoints later land, `monitoring:read` recreates the bundling problem this PR just unwound. Suggest softening the guidance in the three comment sites to "a dedicated capability scoped to the monitored resource (e.g. `drsBrandPresenceAudit:read`)" rather than naming `monitoring:read` specifically. Non-blocking and easy to shape later.
- Small mental-model clarification: the comments all say the route is "in INTERNAL_ROUTES, so S2S is denied at the gate", but enforcement actually derives from absence in `routeRequiredCapabilities` - `INTERNAL_ROUTES` is documentation plus a coverage-test assertion. A one-liner on the `INTERNAL_ROUTES` docstring explaining this would help future readers trace enforcement.

## Recommendations

1. Land the chai one-liner before merge so the regression guard is actually live.
2. Consider the disjointness invariant test as a follow-up; it tightens `route coverage` semantics without replacing it.
3. When the eventual S2S monitoring consumer materialises, name the new capability after the resource being exposed rather than the umbrella "monitoring" concept.

## Assessment

Ready to merge? Yes, pending the chai one-liner.

Reasoning: The architectural change is correct, the auth enforcement is verified against the shared wrapper, and the stale-reference cleanup is complete. The chai assertion is a one-line fix that keeps the new regression guard honest - worth landing in this PR rather than deferring.

rainer-friederich:
Thanks @solaris007 — addressed the chai issue + most of the minor items in f01d2b8c.

**Must-fix (regression guard):**
- Chai API misuse corrected: moved the explanation string to the `expect()` message position so `.not.have.property(route)` is now a strict absence check. Verified by mentally re-adding the mapping with a mismatched value — the fixed assertion fires; the previous form did not.

**Nice-to-haves picked up:**
- Added a disjointness invariant test: `INTERNAL_ROUTES ∩ Object.keys(routeRequiredCapabilities) === ∅`. Subsumes the pinned check for any future internal-only route.
- Softened the forward-looking capability name across all three comment sites (`required-capabilities.js`, `drs-bp-pg-audit.js` JSDoc, `monitoring-api.yaml`): prefer a resource-scoped capability like `drsBrandPresenceAudit:read` over a domain-bucket `monitoring:read` that would recreate the bundling problem for the next monitoring endpoint.
- Clarified the enforcement model in the `INTERNAL_ROUTES` docstring: deny-by-default for S2S JWT is driven by *absence* from `routeRequiredCapabilities`; `INTERNAL_ROUTES` is documentation + a coverage-test assertion.
- Added a pointer in the `routeRequiredCapabilities` docblock to the authoritative entity list (`@adobe/spacecat-shared-data-access` `EntityRegistry`) so future additions start from the right place and don't invent entity names.
- Fixed the em-dash → hyphen inconsistency in the new `INTERNAL_ROUTES` header.

**Still deferred to a follow-up ticket:**
- Known-entities allowlist test that cross-references every entity in `routeRequiredCapabilities` against `EntityRegistry.entities`. Feasibility-checked — the import works — but dropping it in surfaces a pre-existing `brand` entity that is in `routeRequiredCapabilities` but not in `EntityRegistry`. Resolving that needs its own investigation (LLMO-owned, likely either register `brand` in spacecat-shared-data-access or rename the capability), so it's a separate PR. I will file it with a link back here.

Test run: 8189 passing, new disjointness + tightened pinned assertion both green, `docs:lint` clean.

ravverma (approved):

PR #2229, opened 2026-04-16
fix: lower default limit for aem.live proxy requests from 10M to 5000

AndreiAlexandruParaschiv:
- Uncached proxy endpoints were forwarding requests to aem.live with large default limits (10M for GET, 1M for POST) or no limit at all, causing resource exhaustion and 503 errors. 
- Callers that provide an explicit limit (audit workers with limit=80000) are not affected.

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/controllers/llmo/llmo-query-handler.js
+++ b/src/controllers/llmo/llmo-query-handler.js
@@ -80,6 +80,7 @@ const fetchAndProcessSingleFile = async (context, llmoConfig, filePath, queryPar
 
   const url = new URL(`${LLMO_SHEETDATA_SOURCE_URL}/${llmoConfig.dataFolder}/${filePath}`);
 
+  // Apply pagination parameters when calling the source URL
   const DEFAULT_LIMIT = 5000;
   const limit = queryParams.limit ? parseInt(queryParams.limit, 10) : DEFAULT_LIMIT;
   const offset = queryParams.offset ? parseInt(queryParams.offset, 10) : 0;
--- a/src/controllers/llmo/llmo-query-handler.js
+++ b/src/controllers/llmo/llmo-query-handler.js
@@ -15,6 +15,7 @@ import {
   applyFilters,
   applyInclusions,
   applySort,
+  DEFAULT_LLMO_GET_LIMIT,
   LLMO_SHEETDATA_SOURCE_URL,
 } from './llmo-utils.js';
 
@@ -81,8 +82,7 @@ const fetchAndProcessSingleFile = async (context, llmoConfig, filePath, queryPar
   const url = new URL(`${LLMO_SHEETDATA_SOURCE_URL}/${llmoConfig.dataFolder}/${filePath}`);
 
   // Apply pagination parameters when calling the source URL
-  const DEFAULT_LIMIT = 5000;
-  const limit = queryParams.limit ? parseInt(queryParams.limit, 10) : DEFAULT_LIMIT;
+  const limit = queryParams.limit ? parseInt(queryParams.limit, 10) : DEFAULT_LLMO_GET_LIMIT;
   const offset = queryParams.offset ? parseInt(queryParams.offset, 10) : 0;
 
   url.searchParams.set('limit', limit.toString());
--- a/src/controllers/llmo/llmo-query-handler.js
+++ b/src/controllers/llmo/llmo-query-handler.js
@@ -80,8 +80,8 @@ const fetchAndProcessSingleFile = async (context, llmoConfig, filePath, queryPar
 
   const url = new URL(`${LLMO_SHEETDATA_SOURCE_URL}/${llmoConfig.dataFolder}/${filePath}`);
 
-  // Apply pagination parameters when calling the source URL
-  const limit = queryParams.limit ? parseInt(queryParams.limit, 10) : 10000000;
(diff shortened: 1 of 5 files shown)


solaris007 (changes requested):
Hey @AndreiAlexandruParaschiv,

Thanks for jumping on this - directly addressing the traffic spike root cause from the [analysis](https://github.com/adobe/mysticat-architecture/pull/28). Lowering the default from 10M to 5000 is the right call.

## Strengths

- Named constants replace magic numbers (`llmo-query-handler.js:84`, `llmo.js:181`, `llmo.js:409`) - immediately clear intent where `10000000` required mental parsing
- Conservative default (5000) matches the lowest known caller (BP Monitor) - callers that need more already pass explicit limits (audit worker=80000, UI=100000)
- Well-scoped diff - only touches the default limit logic, no unrelated changes mixed in
- Rename of `FIXED_LLMO_LIMIT` to `POST_DEFAULT_LIMIT` (`llmo.js:234`) improves naming consistency
- Aligned with migration direction - the newer PostgREST-based `llmo-brand-presence.js` endpoints already use similar limits (5K/200K)

## Issues

### Important (Should Fix)

**1. Silent behavioral change in GET sheet-data endpoints**

`llmo.js:184` and `llmo.js:412` - the semantics changed more than just the default value:
- Before: if the caller omits `limit`, NO limit parameter was sent to aem.live (aem.live applied its own server-side default)
- After: `limit=5000` is always sent, even when the caller didn't ask for a limit

Any caller that today relies on "no limit = get everything from aem.live" will silently receive truncated data with no error or warning. The known callers (BP Monitor, audit worker, UI) all pass explicit limits, so they are fine. But are there other callers - manual API users, scripts, or services outside our repos - that invoke these GET endpoints without a limit and expect more than 5000 rows?

Recommendation: verify that all callers pass explicit limits. If confirmed safe, add a brief code comment noting this was verified. If uncertain, consider logging when the default is applied so truncation is observable.

**2. No test coverage for the default limit in `llmo-query-handler.test.js`**

The query handler default changed from 10M to 5000 (`llmo-query-handler.js:84`), but only `llmo.test.js` was updated. The existing tests in `llmo-query-handler.test.js` don't assert the default limit value when no limit is provided. If someone reverts or changes the default, no test would catch it.

Recommendation: add a test case:
```javascript
it('should use default limit of 5000 when no limit is provided', async () => {
  setupFetchTest(rawData);
  mockContext.data = {};
  await queryLlmoFiles(mockContext, mockLlmoConfig);
  const fetchUrl = getFetchUrl();
  expect(fetchUrl).to.include('limit=5000');
});
```

**3. Duplicated constant across three locations**

`DEFAULT_LIMIT = 5000` is declared in three places: `llmo-query-handler.js:84`, `llmo.js:181`, and `llmo.js:409` (as `GLOBAL_DEFAULT_LIMIT`). If the value needs to change later, it's easy to miss one - and the naming inconsistency (`DEFAULT_LIMIT` vs `GLOBAL_DEFAULT_LIMIT`) adds confusion. The codebase already uses shared constants via `llmo-utils.js` (e.g., `LLMO_SHEETDATA_SOURCE_URL`).

Recommendation: extract a single `DEFAULT_LLMO_GET_LIMIT = 5000` into `llmo-utils.js` and import it in all three locations.

### Minor (Nice to Have)

**4. `limit || DEFAULT_LIMIT` vs `limit ?? DEFAULT_LIMIT`**

`llmo.js:184` and `llmo.js:412` - the `||` operator treats `0` as falsy and substitutes the default. While `limit=0` is not realistic, `??` (nullish coalescing) is semantically more precise - it only substitutes on `null`/`undefined`, which is the actual intent.

**5. No upper bound on caller-supplied limit (pre-existing)**

Not introduced by this PR, but worth a follow-up: a caller can still pass `limit=999999999` and fully negate this fix. Other controllers in this codebase already enforce upper bounds (e.g., `sentiment.js` uses `Math.min(parsedLimit, MAX_LIMIT)`). Adding a ceiling (e.g., 500K) would prevent a single misconfigured caller from reintroducing the traffic spike.

## Assessment

Ready to merge? **With fixes**

The core change is correct and well-motivated - it directly addresses the traffic spike root cause. Three items to address: (1) verify the behavioral change in GET handlers is safe for all callers, (2) add the missing test for `llmo-query-handler.js`, (3) consolidate the duplicated constant. Items 4-5 can be follow-ups.

tb-adbe (approved):

solaris007 (approved):
Hey @AndreiAlexandruParaschiv,

All three issues from the previous review have been addressed:

- Previously flagged: duplicated constant across 3 locations - now extracted to `DEFAULT_LLMO_GET_LIMIT` in `llmo-utils.js:15` and imported everywhere. Clean.
- Previously flagged: missing test for `llmo-query-handler.js` default - added at `llmo-query-handler.test.js:136`. Covers the key scenario.
- Previously flagged: `||` vs `??` - tried and reverted, which is the right call since query params are strings and `||` handles them correctly.

The naming is consistent (`DEFAULT_LLMO_GET_LIMIT` for GET, `POST_DEFAULT_LIMIT` for POST), the constant lives in the right place (`llmo-utils.js` alongside `LLMO_SHEETDATA_SOURCE_URL`), and tests cover the new behavior.

LGTM.

PR #2230, opened 2026-04-16
feat: add successRate parameter handling | LLMO-4000

calvarezg:
- Updated `parseAgenticTrafficParams` to include `successRate` from query parameters.
- Enhanced `buildRpcParams` to forward `successRate` as `p_success_rate` to the RPC.
- Added tests to ensure `successRate` is correctly processed and forwarded in the RPC call.

## Related Issues
https://jira.corp.adobe.com/browse/LLMO-4000

### DIFF
--- a/src/controllers/llmo/llmo-agentic-traffic.js
+++ b/src/controllers/llmo/llmo-agentic-traffic.js
@@ -93,6 +93,7 @@ function parseAgenticTrafficParams(context) {
     agentType: q.agentType || q.agent_type || null,
     userAgent: q.userAgent || q.user_agent || null,
     contentType: q.contentType || q.content_type || null,
+    successRate: q.successRate || q.success_rate || null,
   };
 }
 
@@ -109,6 +110,7 @@ function buildRpcParams(siteId, parsed) {
     p_agent_type: parsed.agentType,
     p_user_agent: parsed.userAgent,
     p_content_type: parsed.contentType,
+    p_success_rate: parsed.successRate,
   };
 }
 
--- a/src/controllers/llmo/llmo-agentic-traffic.js
+++ b/src/controllers/llmo/llmo-agentic-traffic.js
@@ -33,6 +33,7 @@ const ERR_NOT_FOUND = 'not found';
 
 const VALID_INTERVALS = new Set(['day', 'week', 'month']);
 const VALID_SORT_ORDERS = new Set(['asc', 'desc']);
+const VALID_SUCCESS_RATE_BUCKETS = new Set(['high', 'medium', 'low']);
 // Allowlists mirror the CASE whitelists in the DB RPCs — unknown values are already
 // rejected server-side, but we validate here too for defence-in-depth.
 const VALID_SORT_COLUMNS_BY_URL = new Set([
@@ -93,7 +94,11 @@ function parseAgenticTrafficParams(context) {
     agentType: q.agentType || q.agent_type || null,
     userAgent: q.userAgent || q.user_agent || null,
     contentType: q.contentType || q.content_type || null,
-    successRate: q.successRate || q.success_rate || null,
+    // Normalise to null for unknown buckets — mirrors how PLATFORM_CODE_TO_DB handles
+    // unknown platform codes, preventing a DB exception (500) for invalid input.
+    successRate: VALID_SUCCESS_RATE_BUCKETS.has(q.successRate || q.success_rate)
+      ? (q.successRate || q.success_rate)
+      : null,
   };
 }
 
--- a/test/controllers/llmo/llmo-agentic-traffic.test.js
+++ b/test/controllers/llmo/llmo-agentic-traffic.test.js
@@ -251,6 +251,7 @@ describe('llmo-agentic-traffic', () => {
             agent_type: 'Chatbots',
             user_agent: 'GPTBot',
             content_type: 'html',
+            success_rate: 'high',
           },
         },
       });
@@ -263,6 +264,7 @@ describe('llmo-agentic-traffic', () => {
         p_agent_type: 'Chatbots',
         p_user_agent: 'GPTBot',
         p_content_type: 'html',
+        p_success_rate: 'high',
       });
     });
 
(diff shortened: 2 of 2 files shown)


JayKid:
## 🔍 PR Review — companion to [project-elmo-ui#1519](https://github.com/adobe/project-elmo-ui/pull/1519)

This PR adds `successRate` query param parsing and forwarding as `p_success_rate` to all agentic traffic RPCs. Clean, minimal change with 100% test coverage. Two items to address:

### Critical (1)

**🔴 `by-url` response format mismatch** — `llmo-agentic-traffic.js:419`
The UI PR (#1519) switched `useAgenticTrafficPgByUrl` from `useQuery` to `useInfiniteQuery`, expecting `{ rows: [...], totalCount: N }` per page. But the `by-url` handler still returns a flat mapped array and drops `total_count` from the RPC result.

This change needs to land here (or in a separate backend PR) before the UI PR merges:

```js
// Current (line 419):
return ok((data ?? []).map((row) => ({ ... })));

// Needed:
const rows = (data ?? []).map((row) => ({ ... }));
return ok({
  rows,
  totalCount: Number(data?.[0]?.total_count ?? 0),
});
```

### Major (1)

**No allowlist for `successRate` values** — `parseAgenticTrafficParams` forwards arbitrary strings to the RPC without validation. Other enum params like `platform` use a strict allowlist (`PLATFORM_CODE_TO_DB`). Invalid values (e.g. `successRate=banana`) reach the DB, which raises an exception that surfaces as a 500 instead of a clean 400.

**Fix:**
```js
const VALID_SUCCESS_RATES = new Set(['high', 'medium', 'low']);
const rawRate = q.successRate || q.success_rate || null;
successRate: rawRate && VALID_SUCCESS_RATES.has(rawRate) ? rawRate : null,
```

Add a test for an invalid value being silently dropped to `null`.

### ✅ What's working well

- `p_success_rate` addition to shared `buildRpcParams` is safe — verified all 10 agentic traffic RPCs accept `p_success_rate TEXT DEFAULT NULL`
- Tests cover both snake_case and camelCase query param variants
- Dedicated test for the `by-url` handler verifying `p_success_rate: 'low'` forwarding

calvarezg:
Thanks for the review @JayKid!

**Major (allowlist):** addressed in fc13c788 — added `VALID_SUCCESS_RATE_BUCKETS = {'high', 'medium', 'low'}`, unknown values normalise to `null` (mirrors how `PLATFORM_CODE_TO_DB` handles unknown platform codes), plus tests for the snake_case alias and the invalid-value → `null` path.

**Critical (by-url response shape):** I think this one is based on a stale view of the file — the `{ totalCount, rows }` shape already landed in #2219 (commit `c149d15b`) before this branch was cut. See [`llmo-agentic-traffic.js:434-458`](https://github.com/adobe/spacecat-api-service/blob/feat-success-rate-filter-LLMO-4000/src/controllers/llmo/llmo-agentic-traffic.js#L434-L458):

```js
const rows = data ?? [];
const totalCount = rows.length > 0 ? Number(rows[0].total_count ?? 0) : 0;
return ok({ totalCount, rows: rows.map((row) => ({ ... })) });
```

So the UI PR #1519's `useInfiniteQuery` expectation should already be satisfied. Could you double-check on your end?

JayKid (approved):
## 🔍 Multi-Persona Review (delta) — synthesized

All prior findings have been verified as resolved or refuted with evidence. Synthesizing net-new findings only from Architect / DBA / Security / Tester / Junior Dev perspectives.

**Prior status:**
- ✅ Allowlist added in `fc13c788` (`VALID_SUCCESS_RATE_BUCKETS` + camelCase / snake_case / invalid-value tests) — resolves the Major from round 1.
- ❌ The earlier Critical about `by-url` response shape was based on a stale view — verified the `{ totalCount, rows }` envelope already landed via #2219 at `src/controllers/llmo/llmo-agentic-traffic.js:443-463`. Discarded.

### Critical
None.

### Major (1)

**`p_success_rate` is now forwarded to every agentic traffic RPC via shared `buildRpcParams` — only `by-url` is test-covered.**
- **Where:** `src/controllers/llmo/llmo-agentic-traffic.js:108-120` (diff lines 25-31). `buildRpcParams` is consumed by ~10 handlers (kpis, kpis-trend, by-region, by-category, by-page-type, by-status, by-user-agent, by-url, filter-dimensions, movers).
- **Why it matters:** Manual verification confirms the RPCs accept `p_success_rate TEXT DEFAULT NULL`, but there's no regression test pinning that contract. If any RPC migration lags in an environment or is reverted, PostgREST returns an "unknown parameter" error and every affected endpoint 500s.
- **Fix:** Add a one-line assertion in each existing handler test that `p_success_rate` appears in the mock `.rpc(...)` call, or add a parametrized smoke test iterating the handlers. Low effort, high durability.

### Minor

1. **`successRate` normalisation is case- and whitespace-sensitive** — `VALID_SUCCESS_RATE_BUCKETS.has(q.successRate || q.success_rate)` silently drops `'HIGH'`, `' high '`, etc. to `null` (filter ignored) rather than returning 400. The UI only sends lowercase today so this is latent. `sortOrder` in the same file uses `.toLowerCase()` — consider mirroring that:
   ```js
   const raw = String(q.successRate ?? q.success_rate ?? '').toLowerCase().trim();
   successRate: VALID_SUCCESS_RATE_BUCKETS.has(raw) ? raw : null,
   ```

2. **Duplicated sub-expression** — `q.successRate || q.success_rate` computed twice in the ternary. Hoist into a local (`rawRate`) for readability and to prevent accidental divergence.

### Questions

- **Rollout ordering:** no action needed if #2230 deploys before the UI #1519 ships. Before #2230 deploys, an older backend silently ignores unknown query params — so the ordering is naturally safe. Worth a line in the release notes.
- **`p_url_path_search` LIKE semantics** (not in this PR diff, but visible in the surrounding handler): is `%`/`_` escaping and a length cap handled in the SQL function? A one-line JSDoc on the handler would help future reviewers.

### ✅ Working well

- Surgical, test-covered change (7 production lines, 3 new tests, 100% codecov).
- Mirrors the existing `PLATFORM_CODE_TO_DB` normalisation pattern for unknown values → `null`.
- `url-brand-presence` correctly builds its **own** RPC payload (file lines ~625-655) — the new `p_success_rate` does **not** leak into `rpc_brand_presence_url_detail`. Correct boundary.

---

**Recommendation:** Approve with reservations. No blockers. The Major item is a one-sitting test hardening, not a correctness bug.

PR #2231, opened 2026-04-17
fix: edge-routing - consume promise token from header

dipratap:
/edge-optimize-config API now accepts promise token from both header and cookie. Consuming from Cookie is kept to keep things backward compatible, and will be removed in a later iteration.

### DIFF
--- a/src/support/edge-routing-auth.js
+++ b/src/support/edge-routing-auth.js
@@ -39,6 +39,7 @@ export async function getImsTokenFromPromiseToken(context) {
     err.status = 400;
     throw err;
   }
+  context.log?.info?.(`Promise token found in ${rawHeaderToken ? 'header' : 'cookie'}`);
 
   const promiseToken = decodeURIComponent(rawPromiseToken);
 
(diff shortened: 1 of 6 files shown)


src/support/edge-routing-auth.js
```diff
@@ -22,19 +22,24 @@ export const LLMO_IMS_SERVICE_CODES = ['dx_llmo'];
...
- * Reads the promiseToken cookie from the request and exchanges it for an IMS user access token.
+ * Reads the promise token from the request (cookie first, then x-promise-token header)
+ * and exchanges it for an IMS user access token.
  *
  * @param {object} context - The request context.
  * @returns {Promise<string>} The IMS user access token.
  * @throws {Error} With a `status` property (400 or 401) on failure.
  */
-export async function getImsTokenFromCookie(context) {
-  const rawPromiseToken = getCookieValue(context, 'promiseToken');
```
nit23uec, line 32: we have to support both flows as UI will come later.

nit23uec (approved):

PR #2232, opened 2026-04-17
feat(llmo): add edge-optimize connectivity probe endpoint

shrotia:
Adds GET /sites/{siteId}/llmo/probes/edge-optimize — detects whether a customer's WAF or Bot Manager is blocking the AdobeEdgeOptimize/1.0 user-agent by proxying through the existing Tokowaka worker.

The probe sends a request to live.edgeoptimize.net with x-forwarded-host set to the customer domain. SpaceCat Lambda IPs are allowlisted in Tokowaka, so no API key or prior site setup is needed. The response is classified as hard block (403/429/406), soft block (200 + challenge page), clean pass, or timeout.

Classification logic lives entirely in SpaceCat — zero Tokowaka changes.

LLMO-4340

UI PR: https://github.com/adobe/project-elmo-ui/pull/1561

## Related Issues

Thanks for contributing!


jindaliiita (changes requested):
Hey @shrotia,

## Strengths

The `/llmo/probes/` namespace is the right abstraction - establishing it explicitly in the OpenAPI spec with a forward-looking description invites future probes to follow the same pattern rather than scattering diagnostic endpoints across the route map. Zero Tokowaka changes is a real delivery win; reusing the existing IP allowlist keeps the rollout simple.

The classification logic is well-isolated. Separating `classifyProbeResponse` from the network call makes it independently testable, and the test suite backs that up with 14 unit paths in `edge-routing-utils.test.js` covering all hard-block codes, both timeout error names, URL normalization, and content-type branching. The controller guard ordering is correct: UUID check before DB lookup, org membership before network egress, consistent with existing endpoints.

---

## Issues

### Critical (Must Fix)

**1. SSRF via unvalidated site baseURL**

`src/support/edge-routing-utils.js` - `probeWafConnectivity`

The probe fetches `tokowakaProxyBaseUrl` with `x-forwarded-host: targetHost`, where `targetHost` is derived verbatim from `site.getBaseURL()`. Tokowaka allows requests from SpaceCat Lambda IPs without further hostname validation (the IP allowlist bypasses `doBasicValidation`), so Tokowaka will proxy to whatever hostname the `x-forwarded-host` header specifies. A site whose `baseURL` contains an internal or metadata address (e.g. `http://169.254.169.254` - the AWS IMDS endpoint) would cause Tokowaka to issue a real GET from Fastly PoPs to that address. The HTTP status code oracle returned to the caller is sufficient for network enumeration even without a body. The `site:read` capability is not a strong gate - org members are often not security-vetted, and site records can be misconfigured at onboarding.

Fix: Before calling `probeWafConnectivity`, validate that `targetHost` resolves to a public routable IP (not RFC 1918, loopback, link-local, or AWS metadata ranges). As defense in depth, also add a Tokowaka-side guard that rejects `x-forwarded-host` values resolving to private address space on the allowlisted-IP path.

**2. `TOKOWAKA_PROXY_BASE_URL` env override is fetched without allowlist validation**

`src/controllers/llmo/llmo.js` line ~1894; `src/support/edge-routing-utils.js` - `probeWafConnectivity`

`context.env?.TOKOWAKA_PROXY_BASE_URL` is passed directly as `tokowakaProxyBaseUrl` and used as the `fetch` target without URL validation. Anyone with `lambda:UpdateFunctionConfiguration` IAM permissions can set this to an internal address or a malicious server. The Lambda then issues a request there, trusts the response, and returns the classification result to callers.

Fix: Validate `tokowakaProxyBaseUrl` against a hard-coded allowlist of acceptable proxy origins before calling `fetch`:

```js
const ALLOWED_PROXY_ORIGINS = new Set(['https://live.edgeoptimize.net']);
const proxyOrigin = new URL(effectiveProxy).origin;
if (!ALLOWED_PROXY_ORIGINS.has(proxyOrigin)) {
  log.error(`[waf-probe] Disallowed proxy URL: ${effectiveProxy}`);
  return { ...probeResult, reachable: false, blocked: null, reason: 'error' };
}
```

**3. The probe routes through Tokowaka's full enhancement pipeline, not a raw pass-through**

`src/support/edge-routing-utils.js` - `probeWafConnectivity`

When the probe hits `live.edgeoptimize.net` with `x-forwarded-host` set to a customer domain, Tokowaka's allowlisted-IP path sets `skipValidation = true` and then resolves a hardcoded stub config with `tokowakaEnabled: true, enhancements: true, prerender: { allowList: ['/*'] }`. This means every probe call triggers the full prerender + HTML enhancement pipeline before SpaceCat sees the response. The body and status code that `classifyProbeResponse` receives is not the raw WAF response - it is Tokowaka-processed content. A WAF that blocks the AdobeEdgeOptimize UA might pass a prerender request for a different UA; conversely, the enhancement pipeline could alter a challenge page in ways that defeat keyword detection.

Fix: Either add a Tokowaka probe mode that issues a raw pass-through GET (bypassing prerender and enhancement) and returns the origin response unmodified, or document clearly that the probe classifies the Tokowaka-processed response rather than the raw WAF response, and update the OpenAPI description to reflect this constraint.

**4. `new URL()` is called outside the try block, breaking the never-throws guarantee**

`src/support/edge-routing-utils.js` - `probeWafConnectivity` (URL parsing lines)

`new URL(normalizedUrl)` throws `TypeError` for malformed inputs (empty string, `'https://'`, `null`). This call is before the `try` block in `probeWafConnectivity`. The JSDoc explicitly states "This function never throws" but this line violates that contract. An invalid `siteBaseUrl` propagates a `TypeError` to the controller, which does not catch it, resulting in an unhandled rejection.

Fix: Move the URL parsing inside the `try` block:

```js
try {
  const normalizedUrl = siteBaseUrl.startsWith('http') ? siteBaseUrl : `https://${siteBaseUrl}`;
  const { host: targetHost, href: probedUrl } = new URL(normalizedUrl);
  // ... rest of function
} catch (error) {
  // existing catch handles TypeError too
}
```

**5. `text.length <= BOT_CHALLENGE_BODY_MAX_BYTES` silently skips keyword matching for large-body responses when `content-length` is absent**

`src/support/edge-routing-utils.js` - `classifyProbeResponse`

When `content-length` is absent and the actual body exceeds 2048 bytes, `response.text()` reads the full body but the inner `if (text.length <= BOT_CHALLENGE_BODY_MAX_BYTES)` guard evaluates to false, so keyword matching is silently skipped and the response is classified as `reachable: true`. Verbose Akamai, Imperva, and Barracuda challenge pages commonly run 2-10 KB, making this a real false negative path. The test that covers a 3000-byte body (`'<html><body>'.padEnd(3000, 'x')`) passes because the padded content has no keywords, but it does not assert that keyword matching was attempted - it would still return `reachable: true` even if the body contained 'captcha'.

Fix: Remove the inner `text.length <= BOT_CHALLENGE_BODY_MAX_BYTES` guard and always run keyword matching. To avoid buffering large bodies, stream and truncate at `BOT_CHALLENGE_BODY_MAX_BYTES` bytes rather than calling `response.text()` unconditionally.

---

### Important (Should Fix)

**6. HTTP 406 does not indicate a WAF hard block**

`src/support/edge-routing-utils.js` - `HARD_BLOCK_STATUS_CODES`

HTTP 406 "Not Acceptable" is a content negotiation failure - the server cannot satisfy the client's `Accept` headers. WAFs do not use 406 as a block code; 403 and 429 are the standard block responses. Including 406 will cause false positives for sites with legitimate content negotiation issues. If there is a specific WAF product that returns 406 as its block signal, that should be called out explicitly in a comment. Otherwise, remove 406 from the set.

**7. `'cloudflare'` keyword causes systematic false positives for Cloudflare-hosted sites**

`src/support/edge-routing-utils.js` - `BOT_CHALLENGE_KEYWORDS`

Cloudflare injects attribution text (`Performance & security by Cloudflare`), CDN script paths (`/cdn-cgi/...`), and HTML comments containing 'cloudflare' into legitimate page responses. Any Cloudflare-proxied site with a response body under 2 KB - SPAs, status pages, minimal landing pages - will be classified as soft-blocked. The keyword `'blocked'` has a similar problem (common in marketing copy, analytics payloads, JavaScript variable names). These will generate customer support tickets and undermine trust in the probe result.

Fix: Remove `'cloudflare'` entirely. Replace broad keywords with phrases that appear only in challenge pages: `'cf-chl-widget'`, `'ray id'`, `'completing the challenge'`. For `'blocked'`, require it alongside a challenge-specific term rather than standalone.

**8. Missing LLMO administrator guard, inconsistent with the adjacent `checkEdgeOptimizeStatus` endpoint**

`src/controllers/llmo/llmo.js` - `checkWafConnectivity`

`checkEdgeOptimizeStatus` (the closest equivalent in this controller) gates behind both `accessControlUtil.hasAccess(site)` and `accessControlUtil.isLLMOAdministrator()`. `checkWafConnectivity` only gates on `hasAccess`, meaning any org member with `site:read` can trigger a live outbound probe. If broadening access to non-admin org members is intentional (e.g. for a customer-facing diagnostic UI), that decision needs to be explicit in the PR description. As written, it is silent inconsistency.

Fix: Add the `isLLMOAdministrator()` check after `hasAccess`, or document in the PR why this endpoint is intentionally more permissive.

**9. Content-length header can be spoofed to bypass soft-block keyword detection**

`src/support/edge-routing-utils.js` - `classifyProbeResponse`

When `content-length` is present and greater than 2048, the code returns `reachable: true` without reading the body. A WAF operator aware of this can serve a challenge page with a spoofed `Content-Length: 999999` header to defeat detection. For a security-classification feature, trusting a client-controlled header to skip the keyword check is not sound.

Fix: Remove the content-length fast-path entirely, or use streaming truncated at `BOT_CHALLENGE_BODY_MAX_BYTES` so the keyword check always runs.

**10. No rate limiting on an endpoint that triggers live outbound HTTP requests**

`src/routes/index.js`

There is no per-site cooldown or rate limit on `GET /sites/:siteId/llmo/probes/edge-optimize`. Each call causes a 15-second live outbound request from Tokowaka's Fastly PoPs to the customer's origin. A caller with `site:read` on many sites can loop through them, generating sustained load on customer origins via Adobe infrastructure - a contractual and reputational risk. This is especially important given the broad access gate (see issue 8).

Fix: Add a per-`siteId` rate limit (e.g. one probe per minute per site) at the controller or framework level. At minimum, document the caller-beware constraint in the API description.

**11. OpenAPI security scheme inconsistency**

`docs/openapi/llmo-api.yaml` - `llmo-probe-edge-optimize` security section

Adjacent endpoints like `llmo-edge-optimize-status` use `api_key: []` only. This endpoint lists all three security schemes (`api_key`, `ims_key`, `scoped_api_key`) without explanation. If the probe is surfaced in a customer-facing IMS-authenticated UI, the three-scheme block is correct. If it is internal/admin, the additional schemes create confusion about the intended audience. The PR description does not address the auth design.

**12. `AbortError` name variant is not tested**

`test/support/edge-routing-utils.test.js`

The code handles both `error.name === 'TimeoutError'` and `error.name === 'AbortError'` as `reason: 'timeout'`, but only `TimeoutError` has a test. Manual abort from an `AbortController` produces `AbortError`. Add a test asserting that an `AbortError`-named error produces `reason: 'timeout'`.

---

### Minor (Nice to Have)

**13. `/* c8 ignore next 3 */` suppresses coverage for a real guard**

`src/controllers/llmo/llmo.js` - the `!baseURL` guard in `checkWafConnectivity`

Sites with no `baseURL` can occur from programmatic creation or data migration. A `mockSite.getBaseURL.returns(null)` case is straightforward to add and removes the annotation.

**14. `[waf-probe]` log prefix is inconsistent with the module convention**

`src/support/edge-routing-utils.js`

All existing log statements in this module use `[edge-routing-utils]`. The new code uses `[waf-probe]`, so Splunk queries and alerting rules targeting this module by prefix will miss the probe logs.

**15. Four different names for one concept**

`checkWafConnectivity` (controller), `probes/edge-optimize` (route), `probeWafConnectivity` (utility), `probeEdgeOptimizeConnectivity` (operationId). Aligning around one name reduces the trace-from-route-to-code overhead for new contributors.

---

## Recommendations

- **Formalize the probe namespace contract before the next probe lands.** Define which response fields are always present, whether probe results are cacheable, and what capability is required. A shared probe response schema in `llmo-api.yaml` would enforce consistency.
- **Consider extracting probes into a dedicated controller** (`src/controllers/llmo/llmo-probes.js`). The existing `llmo.js` is approaching 2,000 lines and the probe has no dependency on LLMO config. Extraction keeps the file from becoming a God controller.
- **Add the calling user's identity to probe log lines.** Current logs include `siteId` and `baseURL` but not the IMS org or user. For abuse investigation, you need to answer "who triggered probes for site X at time T."

## Assessment

Ready to merge? No

The SSRF finding (issue 1) is real and exploitable through a misconfigured or adversarially-created site record. The probe pipeline issue (issue 3) means the probe's stated purpose may not be achieved - the classified response is Tokowaka-processed, not the raw WAF response. The `text.length` silent skip (issue 5) produces false negatives for the most common class of verbose challenge pages. The `'cloudflare'` keyword (issue 7) will cause systematic false positives across a large fraction of the customer base.

The architecture and test investment are solid. Fix the SSRF, the pipeline correctness issue, the classification logic edge cases, and the keyword false positives, and this is mergeable.

shrotia (commented):
Thanks @jindaliiita for the detailed review. Addressing each point:

**Critical:**

**#1 SSRF via unvalidated site baseURL** — Fixed. Added `PRIVATE_HOST_RE` to `probeWafConnectivity` that blocks loopback (`127.*`, `localhost`), link-local (`169.254.*`), and RFC1918 ranges (`10.*`, `192.168.*`, `172.16–31.*`) before forwarding to Tokowaka. 6 parameterised unit tests cover all rejected ranges.

**#2 `TOKOWAKA_PROXY_BASE_URL` env override without allowlist validation** — Won't fix. Exploiting this requires `lambda:UpdateFunctionConfiguration` IAM access — an attacker at that privilege level can replace the entire Lambda code/role, making this vector redundant. The env var exists specifically for integration test infrastructure and is not set in any production deployment. An allowlist would break that test usage.

**#3 Probe routes through full Tokowaka enhancement pipeline** — Won't fix in code, clarified in JSDoc and OpenAPI description. For hard blocks (403/406/429) the WAF rejects before Tokowaka's content pipeline runs, so the status code is the raw WAF signal. For 2xx soft blocks the concern is valid; this is a known limitation. A raw pass-through probe mode would require Tokowaka changes out of scope for this PR.

**#4 `new URL()` outside the try block** — Fixed. Moved all URL parsing (including the private-host check) inside the try block. Seeded `probeResult.probedUrl` with `String(siteBaseUrl)` before the try so the catch always has something to return. Added a unit test asserting `'https://'` (which `new URL()` rejects) returns `{ reachable: false, blocked: null, reason: 'error' }` without throwing.

**#5 `text.length <= BOT_CHALLENGE_BODY_MAX_BYTES` silent skip** — Fixed (see also #9). Removed the inner guard entirely. Keyword matching now always runs on `text.slice(0, BOT_CHALLENGE_BODY_MAX_BYTES)` regardless of body size. Added a regression test: 200 HTML response with `'captcha'` keyword buried in a 3000-char body now correctly returns `blocked: true`.

**Important:**

**#6 HTTP 406 in `HARD_BLOCK_STATUS_CODES`** — Keeping. 406 is used by Signal Sciences (Fastly Next-Gen WAF) and some Imperva configurations as their block response. Added an inline comment naming these products.

**#7 `'cloudflare'` and `'blocked'` keywords** — Fixed. Removed both. Replaced with `'cf-chl-widget'` (Cloudflare challenge widget element id) and `'completing the challenge'` (Cloudflare challenge instruction text) — phrases that only appear in actual challenge pages, not legitimate Cloudflare-proxied content.

**#8 Missing `isLLMOAdministrator` guard** — Intentional, documented with a code comment. This endpoint is for the customer-facing diagnostic UI so org members can diagnose their own WAF config. `checkEdgeOptimizeStatus` is admin-only because it exposes internal routing state; probe results are customer-facing diagnostic output.

**#9 Content-length spoofing to bypass keyword detection** — Fixed as part of #5. Content-length fast-path removed entirely; body is always read and checked.

**#10 No rate limiting** — Won't fix in this PR. Tracking as follow-up. Added a note in the OpenAPI `description` field calling out the per-call outbound network cost. Framework-level rate limiting requires cross-cutting infrastructure work.

**#11 OpenAPI security scheme inconsistency** — Intentional. All three schemes are needed: `api_key` for internal tooling, `ims_key` for the customer-facing UI, `scoped_api_key` for programmatic access. Added a comment to the spec.

**#12 `AbortError` not tested** — Fixed. Added test asserting `error.name = 'AbortError'` produces `reason: 'timeout'`.

**Minor:**

**#13 `/* c8 ignore next 3 */` on null baseURL guard** — Fixed. Removed the annotation and added a test: `mockSite.getBaseURL.returns(null)` → HTTP 500 with `'no baseURL'` message, `probeWafConnectivity` not called.

**#14 `[waf-probe]` log prefix** — Fixed. All log lines in `edge-routing-utils.js` and the controller now use `[edge-routing-utils]`.

**#15 Four names for one concept** — Acknowledging but deferring. Aligning `checkWafConnectivity` / `probes/edge-optimize` / `probeWafConnectivity` / `probeEdgeOptimizeConnectivity` touches controller tests, the OpenAPI spec, route definitions, and any downstream client. Scope for a follow-up once this lands.

src/routes/required-capabilities.js
```diff
@@ -484,6 +484,7 @@ const routeRequiredCapabilities = {
   'GET /sites/:siteId/llmo/strategy': 'site:read',
   'PUT /sites/:siteId/llmo/strategy': 'site:write',
   'GET /sites/:siteId/llmo/edge-optimize-status': 'site:read',
+  'GET /sites/:siteId/llmo/probes/edge-optimize': 'site:read',
```
jindaliiita, line 519: I think we should not include probes in the route

src/support/edge-routing-utils.js
```diff
@@ -181,6 +181,144 @@ export async function callCdnRoutingApi(
...
+ *
+ * This function never throws — all errors are captured into the return value.
+ *
+ * @param {string} siteBaseUrl - The site's base URL (with or without scheme).
+ * @param {object} log - Logger.
+ * @param {string} [tokowakaProxyBaseUrl] - Override for the Tokowaka worker base URL.
+ *   Defaults to the production URL. Populated from TOKOWAKA_PROXY_BASE_URL env var.
+ * @returns {Promise<object>} WAF probe result.
+ */
+export async function probeWafConnectivity(
```
jindaliiita: why can't we have this in spacecat-shared as we are polluting the api-service with too much logic

src/controllers/llmo/llmo.js
```diff
@@ -1858,6 +1859,65 @@ function LlmoController(ctx) {
...
+    }
+
+    const baseURL = site.getBaseURL();
+    if (!baseURL) {
+      return internalServerError('Site has no baseURL configured');
+    }
+
+    log.info(`[edge-routing-utils] Starting WAF connectivity probe for site ${siteId} (${baseURL})`);
+
+    const result = await probeWafConnectivity(baseURL, log, context.env?.TOKOWAKA_PROXY_BASE_URL);
```
jindaliiita: we should call a single function of spacecat-shared checkwafconnectivity from api-service
In that function we should call edgeoptimize status and call to live edge optimize

src/controllers/llmo/llmo.js
```diff
@@ -1858,6 +1859,65 @@ function LlmoController(ctx) {
...
+
+    log.info(`[edge-routing-utils] Result for site ${siteId}: reachable=${result.reachable}, blocked=${result.blocked}`);
```
jindaliiita: [edge-routing-utils] should be changed

src/support/edge-routing-utils.js
```diff
@@ -181,6 +181,144 @@ export async function callCdnRoutingApi(
...
+  siteBaseUrl,
+  log,
+  tokowakaProxyBaseUrl = TOKOWAKA_PROXY_BASE_URL_DEFAULT,
```
jindaliiita: why this ?

src/support/edge-routing-utils.js
```diff
@@ -181,6 +181,144 @@ export async function callCdnRoutingApi(
...
+) {
+  // Seed with the raw input so the catch block always has a probedUrl to return,
+  // even when URL parsing itself throws (e.g. empty string, null, malformed input).
+  let probeResult = { probedUrl: String(siteBaseUrl) };
+
+  try {
+    const normalizedUrl = siteBaseUrl.startsWith('http') ? siteBaseUrl : `https://${siteBaseUrl}`;
```
jindaliiita: use prependSchema of spacecat-shared

src/support/edge-routing-utils.js
```diff
@@ -181,6 +181,144 @@ export async function callCdnRoutingApi(
...
+      return {
+        ...probeResult, reachable: false, blocked: null, reason: 'error',
+      };
+    }
+
+    log.info(`[edge-routing-utils] Probing ${targetHost} via Tokowaka proxy at ${tokowakaProxyBaseUrl}`);
+
+    const response = await fetch(tokowakaProxyBaseUrl, {
+      method: 'GET',
+      headers: {
```
jindaliiita: can we add edgeoptimize-config and x-edgeoptimize-url also

src/support/edge-routing-utils.js
```diff
@@ -181,6 +181,144 @@ export async function callCdnRoutingApi(
...
 }
 
+// Default Tokowaka worker base URL for WAF connectivity probes.
+// The probe sends a normal proxy request through Tokowaka (with x-forwarded-host)
+// so the origin fetch happens from Fastly edge IPs with the AdobeEdgeOptimize UA.
+const TOKOWAKA_PROXY_BASE_URL_DEFAULT = 'https://live.edgeoptimize.net';
+
+// Regex matching loopback, link-local, and RFC1918 private address ranges.
+// These must never be forwarded to Tokowaka as probe targets.
+const PRIVATE_HOST_RE = /^(localhost$|127\.|10\.|192\.168\.|172\.(1[6-9]|2\d|3[01])\.|169\.254\.)/i;
```
jindaliiita: is it needed ?

src/support/edge-routing-utils.js
```diff
@@ -181,6 +181,144 @@ export async function callCdnRoutingApi(
   }
 }
 
+// Default Tokowaka worker base URL for WAF connectivity probes.
+// The probe sends a normal proxy request through Tokowaka (with x-forwarded-host)
+// so the origin fetch happens from Fastly edge IPs with the AdobeEdgeOptimize UA.
+const TOKOWAKA_PROXY_BASE_URL_DEFAULT = 'https://live.edgeoptimize.net';
```
jindaliiita: dont use tokowaka anywhere

src/support/edge-routing-utils.js
```diff
@@ -181,6 +181,144 @@ export async function callCdnRoutingApi(
...
+  if (status >= 200 && status < 300) {
+    const contentType = response.headers.get('content-type') || '';
+
+    if (contentType.includes('text/html')) {
+      // Always read and keyword-match the body — never trust content-length to skip this,
+      // as a WAF can serve a challenge page with a spoofed large content-length header.
+      // Keyword matching runs on the first BOT_CHALLENGE_BODY_MAX_BYTES characters only.
+      const text = await response.text();
+      const snippet = text.slice(0, BOT_CHALLENGE_BODY_MAX_BYTES).toLowerCase();
+      const isSoftBlock = BOT_CHALLENGE_KEYWORDS.some((kw) => snippet.includes(kw));
```
jindaliiita: is this a standard practice ?

src/support/edge-routing-utils.js
shrotia: this was raised in the AI review comment - the first SSRF point.

src/support/edge-routing-utils.js
shrotia: yes, looks like - claude suggested with confidence to capture soft 403s

src/support/edge-routing-utils.js
shrotia: yes, makes sense

src/controllers/llmo/llmo.js
shrotia: yes

src/support/edge-routing-utils.js
shrotia: for local debugging - I can remove this

src/support/edge-routing-utils.js
shrotia: yes, done

docs/openapi/api.yaml
```diff
@@ -57,6 +57,11 @@ tags:
     description: APIs for taking and retrieving webpage screenshots, specifically for consent banner analysis
   - name: llmo
     description: LLMO (Large Language Model Optimizer) operations
+  - name: llmo-probes
```
jindaliiita, line 60: this probe is specific to edge optimize

src/routes/required-capabilities.js
shrotia: imho probes makes sense to add more health checks/probes in future for LLMO

docs/openapi/api.yaml
shrotia: Agreed it's the only probe today — kept the tag generic intentionally since the `/llmo/probes/` URL namespace is designed to host future probes (e.g. prerender reachability, CDN header validation). A single tag now keeps docs organized as more probes land, rather than retrofitting later. Updated the description to be explicit that only edge-optimize exists currently.

shrotia: yes, but the tag should be generic. right ?

src/support/edge-routing-utils.js
shrotia: I dont think so, the WAF probe is deliberately minimal — it sends only x-forwarded-host to replicate what the optimizer does at the customer's edge IP. 
anyways code is moved, if you still think .. we can discuss in spacecat-sahred

jindaliiita (approved):
LGTM
@nit23uec could you please take a look at this as well?

PR #2239, opened 2026-04-19
fix(plg): move org reassignment before entitlement operations

Kanishkavijay39:
Fix critical bug where site org reassignment happened AFTER entitlement operations, causing enrollments to be revoked from the wrong organization.

**The Problem:**
When a site in Sites Internal org was onboarded to a customer org:
1. ensureAsoEntitlement() got the Sites Internal org's entitlement
2. revokePreOnboardedSiteEnrollment() revoked ALL other Sites Internal sites' enrollments
3. Then the site was moved to customer org (too late)

This caused Sites Internal sites to lose their enrollments when other sites were being onboarded.

**The Fix:**
Move site org reassignment to happen BEFORE entitlement operations in both code paths:
1. Fast-track preonboarding path (PRE_ONBOARDING status)
2. Full onboarding path (new sites from internal orgs)

Now the flow correctly:
- Reassigns site to customer org FIRST
- Gets the customer org's entitlement
- Revokes enrollments only within that customer org
- Sites Internal org remains untouched

This ensures data integrity and prevents accidental removal of enrollments from internal org sites during customer onboarding.

Made-with: Cursor

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/controllers/plg/plg-onboarding.js
+++ b/src/controllers/plg/plg-onboarding.js
@@ -667,6 +667,8 @@ async function performAsoPlgOnboarding({
       const currentSiteOrgId = site.getOrganizationId();
       let needsOrgReassignment = false;
 
+      // Note: On retry, currentSiteOrgId may already equal customerOrgId if a previous
+      // attempt successfully saved the org reassignment but failed during entitlement creation
       if (currentSiteOrgId !== customerOrgId) {
         if (isInternalOrg(currentSiteOrgId, env) && !isInternalOrgDemoSite(site.getId(), env)) {
           log.info(`Preonboarded site ${site.getId()} is in internal org ${currentSiteOrgId}, will reassign to customer org ${customerOrgId}`);
@@ -710,7 +712,13 @@ async function performAsoPlgOnboarding({
       await revokePreOnboardedSiteEnrollment(site, entitlement, context);
       await updateLaunchDarklyFlags(site, context);
 
-      const steps = { ...(onboarding.getSteps() || {}), entitlementCreated: true };
+      const steps = {
+        ...(onboarding.getSteps() || {}),
+        entitlementCreated: true,
+      };
+      if (needsOrgReassignment) {
+        steps.siteOrgReassigned = true;
+      }
       onboarding.setStatus(STATUSES.ONBOARDED);
       onboarding.setWaitlistReason(null);
       onboarding.setBotBlocker(null);
@@ -780,8 +788,8 @@ async function performAsoPlgOnboarding({
 
       if (existingOrgId !== organizationId) {
         if (isInternalOrg(existingOrgId, env) && !isInternalOrgDemoSite(site.getId(), env)) {
-          log.info(`Site ${site.getId()} org ${existingOrgId} is internal/demo — will reassign to new org ${organizationId} after successful onboarding`);
-          // Don't save yet - we'll reassign the org just before marking ONBOARDED
+          log.info(`Site ${site.getId()} org ${existingOrgId} is internal/demo — will reassign to new org ${organizationId} before entitlement operations`);
+          // Will reassign at Step 9, before creating entitlements
           needsOrgReassignment = true;
         } else {
           const existingOrg = await Organization.findById(existingOrgId);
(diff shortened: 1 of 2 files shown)


Sugandhgyl:
**Code Review — fix(plg): move org reassignment before entitlement operations**
**Overview**
The root cause diagnosis is accurate. When `ensureAsoEntitlement` was called with the site still in the internal org, it scoped the entitlement to the wrong org. Then `revokePreOnboardedSiteEnrollment` revoked enrollments from all other sites under that same internal org's entitlement — silently breaking unrelated customers. Moving the org reassignment first is the correct fix.

The change touches two code paths: the fast-track `PRE_ONBOARDING` path (~line 658) and the full `onboarding` path (~line 1088).

**Issues**
🔴 **No tests added**
This is the most significant gap. The PR changes the execution order of async, state-mutating operations, but `test/controllers/plg/plg-onboarding.test.js` was not touched. At minimum:

- Fast-track path: assert that `site.setOrganizationId(customerOrgId)` is called before `ensureAsoEntitlement`
- Full onboarding path: assert that site.save() for org reassignment happens before the entitlement call
- Both paths: when needsOrgReassignment = false, the flow is unchanged
- Without these, the original bug could regress without any CI signal.

🟡 **Stale comments at lines** 782–783 (full onboarding path)

- log.info(`...will reassign to new org ${organizationId} after successful onboarding`);
- // Don't save yet - we'll reassign the org just before marking ONBOARDED
- After this fix the reassignment now happens at Step 9, not `"after successful onboarding" `or `"just before marking ONBOARDED"`. These comments will mislead future readers into thinking the save is deferred to the end of the function. They should be updated to reflect the new timing.

🟡 **Inconsistent steps tracking between the two paths**
The full onboarding path correctly sets `steps.siteOrgReassigned = true `after the reassignment block. The fast-track **PRE_ONBOARDING** path does not. If the fast-track path fails after reassignment, there is no record in the steps object that the site was already moved. This makes debugging and retry reasoning harder.

🟡 **Partial-failure comment worth adding (fast-track path)**
In the fast-track path, after the fix:

`site.setOrganizationId(customerOrgId) + site.save() `— persisted
`ensureAsoEntitlement(site, context)` — if this throws, function exits
The onboarding record is not yet saved, but the site's org has already changed. On retry, currentSiteOrgId === customerOrgId → needsOrgReassignment = false → safe. This works correctly, but a brief comment noting "site may already be in customer org on retries" would help future developers understand why needsOrgReassignment could be false even for a site that was originally in an internal org.

**Positives**
The root cause analysis is accurate and the fix is minimal and targeted
Step numbering is updated consistently in the full onboarding path
New comments at the moved block clearly explain the why
onboarding.setOrganizationId() is correctly kept in sync in both paths
Retry behavior is safe in both paths despite partial-failure risk

Sugandhgyl:
**Code Re-Review**

🔴 No tests added — **FIXED**
Two test additions in [test/controllers/plg/plg-onboarding.test.js](vscode-webview://0a41nfombsauat3icp02cu7p3otncv4msu1voe15dvvd9o6il2e1/test/controllers/plg/plg-onboarding.test.js):

Fast-track path — existing test now has a calledBefore assertion:


expect(siteInInternalOrg.save).to.have.been.calledBefore(tierClientCreateForSiteStub);
Full onboarding path — a brand new test 'reassigns site from internal org before entitlement in full onboarding path' with the same calledBefore guard. This properly simulates a site in an internal org going through the full flow and verifies ordering.

🟡 Stale comments at lines 782–783 — **FIXED**
Updated from:

"will reassign to new org ... after successful onboarding"
"Don't save yet - we'll reassign the org just before marking ONBOARDED"

To:

"will reassign to new org ... before entitlement operations"
"Will reassign at Step 9, before creating entitlements"

🟡 Inconsistent steps tracking — **FIXED**
The fast-track path now sets steps.siteOrgReassigned = true matching the full onboarding path. The steps object was also reformatted for consistency.

🟡 Partial-failure comment — **FIXED**
Comment added above the currentSiteOrgId !== customerOrgId check in the fast-track path:


All 4 issues are addressed. The PR is good to approve — the critical test gap is covered with proper ordering assertions (calledBefore), the stale comments are updated, and both paths now have consistent step tracking and retry documentation.

Sugandhgyl (approved):

PR #2240, opened 2026-04-20
fix: make POST /v2/orgs/:id/categories idempotent by name | LLMO-4370

rainer-friederich:
## Summary
- **Eliminates the ~7.5k/day 409 log storm** on `POST /v2/orgs/:spaceCatId/categories` (18,582 errors in 48h) by making `createCategory` idempotent by name: if a non-deleted row with the same name already exists for the org, the storage layer UPDATEs non-key fields and returns the existing row — preserving its stable `category_id` slug so FKs stay valid. See [LLMO-4370](https://jira.corp.adobe.com/browse/LLMO-4370).
- **Mirrors 23505 → 409 mapping on topics** so future unique-constraint drift on `POST /v2/orgs/:spaceCatId/topics` surfaces as a typed 409 instead of a 500. Message echoes the actual constraint name rather than hard-coding `name`.
- **Demotes 409 log severity** in `createTopicForOrg` to `log.warn` so idempotent-retry traffic stops polluting Coralogix ERROR severity, without losing triage visibility on legitimate conflicts.

## Root cause
DRS `ensure_categories_v2` (`llmo-data-retrieval-service`) re-POSTs each LLMO-config category on every scheduler run, treating 409 as "already exists, skip". The server-side upsert used `onConflict: organization_id,category_id`, so when a client's `category.id` (slug from LLMO config) drifted from the DB's `category_id` (e.g., after the `baseurl-*` rename from LLMO-4160 / fix_slug_category_names), the upsert fell back to INSERT and tripped `uq_category_name_per_org` on every daily re-post.

The real fix is server-side: resolve conflicts on name so re-posts with drifted slugs become clean UPDATEs.

### Storage
- `src/support/categories-storage.js` — lookup-by-name idempotent create; 23505 race recovery retries the lookup once (original 23505 preserved as `Error.cause` if the retry itself fails); soft-delete resurrection; no-op short-circuit when nothing meaningful changes (preserves audit trail); provenance guard (`origin: 'human'` is never downgraded to `'ai'`); NFC + whitespace + case-insensitive name canonicalization; concurrent hard-delete surfaces as typed 409 "concurrently modified".
- `src/support/topics-storage.js` — 23505 → 409 with constraint-name echo. `Error.cause` chained on all wrapped errors.

### Controller
- `src/controllers/brands.js` — `createCategoryForOrg` returns **201 on insert/resurrect, 200 on idempotent update**, so clients can discriminate creation from re-post without parsing the body. `createTopicForOrg` logs 409 at WARN (was INFO → lost visibility; was ERROR → log storm).

### OpenAPI
- `docs/openapi/categories-v2-api.yaml` — **new**. Documents the 200/201 contract, name canonicalization, provenance guard, resurrection, and no-op semantics. `V2Category*` schemas added to `schemas.yaml`.
- `docs/openapi/topics-v2-api.yaml` — 409 response added with rationale.

### Tests
Comprehensive coverage added: no-op short-circuit, resurrection, provenance guard (block downgrade / allow upgrade), canonical-name storage, case-variant lookup matching, concurrent hard-delete 409, retry-lookup failure preserving original 23505 via `Error.cause`, topic 23505 generic-message fallback.

## Review follow-ups addressed in this PR
A full local-review-team pass surfaced 18 issues. Resolutions:

| # | Score | Status |
|---|-------|--------|
| 1 | 80 | **Fixed** — soft-delete resurrection path |
| 2 | 80 | **Fixed** — 200/201 split + categories OpenAPI + topic 409 response |
| 3 | 75 | **Fixed** — provenance guard blocks `human → ai` downgrade |
| 4 | 70 | **Fixed** — 200 on update-path, 201 on insert/resurrect |
| 5 | 65 | **Fixed** — topic 23505 echoes actual constraint name |
| 6 | 65 | **Fixed** — no-op short-circuit preserves audit trail |
| 7 | 65 | **Partial** — code-level NFC + whitespace + case-insensitive match; schema-level case-insensitive uniqueness deferred to RFC |
| 9 | 55 | **Implicit** — status codes now differ, tests distinguish paths |
| 11 | 50 | **Fixed** — topic 409 → `log.warn` |
| 13 | 40 | **Fixed** — resurrection regression test |
| 14 | 35 | **Fixed** — topic 409 in OpenAPI |
| 15 | 35 | **Fixed** — original insert error preserved as `Error.cause` |
| 16 | 25 | **Fixed** — concurrent-hard-delete test; storage emits typed 409 |
| 17 | 20 | **Fixed** — `Error.cause` applied across storage |
| 18 | 15 | **Fixed** — redundant `data \|\| null` removed |

## Follow-ups (deferred — belong in a taxonomy-idempotency RFC)
The following items surfaced in review but are out of scope for a log-storm triage PR. They should land in a follow-up RFC (`mysticat-architecture/`):

- **#8 — Unify topics / categories idempotency semantics.** Today categories are idempotent by name and topics are not. DRS's `ensure_topics_v2` has the same shape as `ensure_categories_v2`; the log storm will eventually recur on topics. Likely fix: make topics idempotent by `(name, brandId)` after a schema migration (no name-unique on topics today).
- **#10 — Single-round-trip server-side upsert.** Current implementation is 2 PostgREST round-trips on the happy path (lookup + insert/update), 3 on race. A SQL RPC can reduce this to 1 while still discriminating create vs update (e.g., via `xmax = 0` in RETURNING).
- **#7 (completion) — Schema-level case-insensitive uniqueness.** The code-level canonicalization in this PR catches whitespace and NFC variants, and makes lookups case-insensitive, but the underlying `uq_category_name_per_org` constraint is still case-sensitive. Two concurrent inserts with different cases (no pre-existing row) can still both succeed. Fix needs a partial index on `LOWER(name)` or a generated `canonical_name` column.
- **Server-owned slug.** `category.id` should not be client-supplied at all — clients reference existing categories by UUID. Removing the client-side slug kills the LLMO-4160 drift class of bug permanently. Requires DRS and UI migration.
- **Bulk sync endpoint for DRS.** `PUT /v2/orgs/:id/categories` with the canonical set, reconciled server-side in one transaction, is a better fit for DRS's "assert this taxonomy" intent than N individual POSTs.

## Test plan
- [x] `npx mocha test/support/categories-storage.test.js test/support/topics-storage.test.js` — passing
- [x] `npx mocha test/controllers/brands.test.js -g "createCategoryForOrg|createTopicForOrg"` — passing
- [x] `npm test` — 8,295 passing, 100% lines/branches/statements
- [x] `npm run lint` — clean (only pre-existing eslint-env deprecation warnings)
- [x] `npm run docs:lint` — OpenAPI valid (pre-existing warnings on unrelated paths only)
- [ ] Post-deploy: verify Coralogix category-409 count drops to ~0 over 24h after rollout (query: `filter $d.message ~ 'Error creating category for organization'`).

### DIFF
--- a/docs/openapi/categories-v2-api.yaml
+++ b/docs/openapi/categories-v2-api.yaml
@@ -1,0 +1,192 @@
+v2-categories-for-org:
+  parameters:
+    - name: spaceCatId
+      in: path
+      required: true
+      description: SpaceCat Organization ID (UUID)
+      schema:
+        type: string
+        format: uuid
+        example: 'a1b2c3d4-5678-90ab-cdef-1234567890ab'
+    - name: status
+      in: query
+      required: false
+      description: Filter by category status. Default excludes deleted categories.
+      schema:
+        type: string
+        enum: [active, pending, deleted, ignored]
+        example: 'active'
+  get:
+    tags:
+      - customer-config
+    summary: List categories for an organization (v2)
+    description: |
+      Returns all categories for the given organization, backed by
+      PostgREST/Aurora. By default, deleted categories are excluded. Use the
+      `status` query parameter to filter by specific status or include deleted
+      categories.
+    operationId: listCategoriesForOrgV2
+    security:
+      - ims_key: []
+      - api_key: []
+    responses:
+      '200':
+        description: Categories retrieved successfully
+        content:
+          application/json:
+            schema:
+              $ref: './schemas.yaml#/V2CategoryListResponse'
+      '400':
+        $ref: './responses.yaml#/400'
+      '403':
+        description: User does not have access to this organization
+      '404':
+        $ref: './responses.yaml#/404-organization-not-found-with-id'
+      '500':
+        $ref: './responses.yaml#/500'
+      '503':
+        description: PostgREST unavailable (DATA_SERVICE_PROVIDER=postgres required)
+  post:
+    tags:
+      - customer-config
+    summary: Create a category (idempotent by name) (v2)
+    description: |
+      Creates a new category for the organization, idempotent by name.
+
+      Name matching is case-insensitive and whitespace-normalized (NFC),
+      so `"Discovery & Research"`, `"  discovery  &  research  "`, and
+      `"DISCOVERY & RESEARCH"` resolve to the same category.
+
+      Behavior by outcome:
+      - **New row inserted** — returns `201 Created` with the new row.
+      - **Existing row matched with identical fields** — no-op; returns
+        `200 OK` with the existing row. Audit trail (`updatedAt`,
+        `updatedBy`) is preserved.
+      - **Existing row matched with differing fields** — patch applied to
+        `origin` / `status` (with provenance guard: `origin` is never
+        downgraded from `human` to `ai`), returns `200 OK` with the
+        updated row.
+      - **Soft-deleted row matched** — resurrected: status flipped to
+        `active`, stable `id` (slug) preserved, returns `201 Created`.
+
+      Clients discriminating creation from idempotent update should inspect
+      the HTTP status, not the response body (identical in both cases).
+
+      Never returns `409`: duplicate-name reposts are absorbed by design.
+    operationId: createCategoryForOrgV2
+    security:
+      - ims_key: []
+      - api_key: []
+    requestBody:
+      required: true
+      content:
+        application/json:
+          schema:
+            $ref: './schemas.yaml#/V2CategoryInput'
+    responses:
+      '200':
+        description: |
+          Existing category returned (idempotent update or no-op).
+          The `id` (slug) of the returned row is authoritative — it may
+          differ from a client-supplied `id` if the stored row was created
+          with a different slug.
+        content:
+          application/json:
+            schema:
+              $ref: './schemas.yaml#/V2Category'
+      '201':
+        description: |
+          New category created, or a soft-deleted category with the same
+          name was resurrected (status flipped back to `active`).
+        content:
+          application/json:
+            schema:
+              $ref: './schemas.yaml#/V2Category'
+      '400':
+        $ref: './responses.yaml#/400'
+      '403':
+        description: User does not have access to this organization
+      '404':
+        $ref: './responses.yaml#/404-organization-not-found-with-id'
+      '500':
+        $ref: './responses.yaml#/500'
+      '503':
+        description: PostgREST unavailable (DATA_SERVICE_PROVIDER=postgres required)
+
+v2-category-for-org:
+  parameters:
+    - name: spaceCatId
+      in: path
+      required: true
+      description: SpaceCat Organization ID (UUID)
+      schema:
+        type: string
+        format: uuid
+        example: 'a1b2c3d4-5678-90ab-cdef-1234567890ab'
+    - name: categoryId
+      in: path
+      required: true
+      description: Category business key (category_id slug)
+      schema:
+        type: string
+        example: 'baseurl-discovery-research'
+  patch:
+    tags:
+      - customer-config
+    summary: Update a category (v2)
+    description: |
+      Updates an existing category. Only provided fields are updated. The
+      `name` field is canonicalized (trim + whitespace-collapse + NFC) on
+      write.
+    operationId: updateCategoryForOrgV2
+    security:
+      - ims_key: []
+      - api_key: []
+    requestBody:
+      required: true
+      content:
+        application/json:
+          schema:
+            $ref: './schemas.yaml#/V2CategoryUpdateInput'
+    responses:
+      '200':
+        description: Category updated successfully
+        content:
+          application/json:
+            schema:
+              $ref: './schemas.yaml#/V2Category'
+      '400':
+        $ref: './responses.yaml#/400'
+      '403':
+        description: User does not have access to this organization
+      '404':
+        description: Organization or category not found
+      '500':
+        $ref: './responses.yaml#/500'
+      '503':
+        description: PostgREST unavailable (DATA_SERVICE_PROVIDER=postgres required)
+  delete:
+    tags:
+      - customer-config
+    summary: Delete a category (v2)
+    description: |
+      Soft-deletes a category by setting its status to `deleted`. The
+      category remains queryable with `?status=deleted` and can be
+      resurrected by POSTing a category with the same name.
+    operationId: deleteCategoryForOrgV2
+    security:
+      - ims_key: []
+      - api_key: []
+    responses:
+      '204':
+        description: Category deleted successfully
+      '400':
+        $ref: './responses.yaml#/400'
+      '403':
+        description: User does not have access to this organization
+      '404':
+        description: Organization or category not found
+      '500':
+        $ref: './responses.yaml#/500'
+      '503':
+        description: PostgREST unavailable (DATA_SERVICE_PROVIDER=postgres required)
--- a/docs/openapi/categories-v2-api.yaml
+++ b/docs/openapi/categories-v2-api.yaml
@@ -72,7 +72,9 @@ v2-categories-for-org:
       Clients discriminating creation from idempotent update should inspect
       the HTTP status, not the response body (identical in both cases).
 
-      Never returns `409`: duplicate-name reposts are absorbed by design.
+      Duplicate-name reposts (the common DRS re-sync case) are absorbed by
+      design and do NOT produce `409`. `409` is reserved for the two
+      orthogonal race conditions documented under the `409` response below.
     operationId: createCategoryForOrgV2
     security:
       - ims_key: []
@@ -108,6 +110,19 @@ v2-categories-for-org:
         description: User does not have access to this organization
       '404':
         $ref: './responses.yaml#/404-organization-not-found-with-id'
+      '409':
+        description: |
+          Rare race condition — caller should retry the full POST. Fires in
+          two scenarios:
+          1. The existing row matched by name was hard-deleted (not soft-
+             deleted) between our lookup and our update — message
+             `Category was concurrently modified; please retry`.
+          2. A unique-constraint other than `uq_category_name_per_org` was
+             tripped on insert (e.g. `uq_category_id_per_org` when the
+             client-supplied slug collides with an unrelated existing
+             row). Message echoes the actual constraint name.
+          Treated as an idempotent retry signal by DRS-class clients;
+          logged at WARN (not ERROR) server-side.
       '500':
         $ref: './responses.yaml#/500'
       '503':
@@ -161,6 +176,11 @@ v2-category-for-org:
         description: User does not have access to this organization
       '404':
         description: Organization or category not found
+      '409':
+        description: |
+          Updated `name` collides with another active or soft-deleted
+          category in the same organization (`uq_category_name_per_org`).
+          Message echoes the actual constraint name.
       '500':
         $ref: './responses.yaml#/500'
       '503':
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -752,21 +752,38 @@ function BrandsController(ctx, log, env) {
       const { postgrestClient } = context.dataAccess.services;
       const updatedBy = context.attributes?.authInfo?.profile?.email || 'system';
 
-      const { category, created } = await createCategory({
+      const { category, created, outcome } = await createCategory({
         organizationId: spaceCatId,
         category: categoryData,
         postgrestClient,
         updatedBy,
+        log,
+      });
+
+      // Log-storm quantification tag: lets Coralogix aggregate
+      // category_post_result by outcome so the LLMO-4370 cleanup can be
+      // quantified post-deploy (insert/resurrect/update/noop/race_retry*)
+      // without grepping messages. LLMO-4370 #15.
+      log.info(`Category POST resolved for organization ${spaceCatId}`, {
+        organization_id: spaceCatId,
+        category_id: category.id,
+        outcome,
       });
 
       // 201 on insert, 200 on idempotent update — lets callers (UI toast,
       // DRS audit log) distinguish "created new" from "ensured existing".
       return createResponse(category, created ? 201 : 200);
     } catch (error) {
-      // Storage is idempotent by name: matching rows return via the normal
-      // path, so we never see a 409 here. Anything caught here is a real
-      // failure.
-      log.error(`Error creating category for organization ${spaceCatId}:`, error);
+      // Storage is idempotent by name on the happy path, but still raises
+      // a typed 409 on the concurrent hard-delete race (row vanishes
+      // between lookup and update) and on non-name uniqueness collisions
+      // (slug drift). Those are retry signals, not server faults — warn,
+      // don't error, to keep Coralogix ERROR severity clean. LLMO-4370 #2.
+      if (error?.status === 409) {
+        log.warn(`Category conflict for organization ${spaceCatId}: ${error.message}`);
+      } else {
+        log.error(`Error creating category for organization ${spaceCatId}:`, error);
+      }
       return createErrorResponse(error);
     }
   };
@@ -815,7 +832,14 @@ function BrandsController(ctx, log, env) {
       }
       return ok(updated);
     } catch (error) {
-      log.error(`Error updating category ${categoryId} for organization ${spaceCatId}:`, error);
+      // PATCH to a name colliding with a sibling row in the same org
+      // surfaces from storage as a typed 409. Mirror the POST handler so
+      // legitimate name-conflict retries don't pollute ERROR severity.
+      if (error?.status === 409) {
+        log.warn(`Category update conflict for organization ${spaceCatId}: ${error.message}`);
+      } else {
+        log.error(`Error updating category ${categoryId} for organization ${spaceCatId}:`, error);
+      }
       return createErrorResponse(error);
     }
   };
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -752,18 +752,20 @@ function BrandsController(ctx, log, env) {
       const { postgrestClient } = context.dataAccess.services;
       const updatedBy = context.attributes?.authInfo?.profile?.email || 'system';
 
-      const created = await createCategory({
+      const { category, created } = await createCategory({
         organizationId: spaceCatId,
         category: categoryData,
         postgrestClient,
         updatedBy,
       });
 
-      return createResponse(created, 201);
+      // 201 on insert, 200 on idempotent update — lets callers (UI toast,
+      // DRS audit log) distinguish "created new" from "ensured existing".
+      return createResponse(category, created ? 201 : 200);
     } catch (error) {
-      // Storage is idempotent by name: matching rows return 201 with the
-      // existing row, so we never see a 409 here. Anything caught here is a
-      // real failure.
+      // Storage is idempotent by name: matching rows return via the normal
+      // path, so we never see a 409 here. Anything caught here is a real
+      // failure.
       log.error(`Error creating category for organization ${spaceCatId}:`, error);
       return createErrorResponse(error);
     }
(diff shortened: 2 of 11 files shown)


src/controllers/brands.js
```diff
@@ -752,15 +752,20 @@ function BrandsController(ctx, log, env) {
...
 
-      return createResponse(created, 201);
+      // 201 on insert, 200 on idempotent update — lets callers (UI toast,
+      // DRS audit log) distinguish "created new" from "ensured existing".
+      return createResponse(category, created ? 201 : 200);
     } catch (error) {
+      // Storage is idempotent by name: matching rows return via the normal
+      // path, so we never see a 409 here. Anything caught here is a real
+      // failure.
       log.error(`Error creating category for organization ${spaceCatId}:`, error);
```
solaris007: **Critical:** This logs every error at ERROR severity, including the typed 409 that `categories-storage.js` raises on the concurrent hard-delete race (covered by a dedicated test). The stated goal of this PR is to reduce ERROR noise - mirror the topics pattern and demote 409s to `log.warn`: `if (error?.status === 409) log.warn(...); else log.error(...)`.

docs/openapi/categories-v2-api.yaml
```diff
@@ -0,0 +1,192 @@
...
+        `origin` / `status` (with provenance guard: `origin` is never
+        downgraded from `human` to `ai`), returns `200 OK` with the
+        updated row.
+      - **Soft-deleted row matched** — resurrected: status flipped to
+        `active`, stable `id` (slug) preserved, returns `201 Created`.
+
+      Clients discriminating creation from idempotent update should inspect
+      the HTTP status, not the response body (identical in both cases).
+
+      Never returns `409`: duplicate-name reposts are absorbed by design.
```
solaris007: **Critical:** This description promises "Never returns 409" but the storage layer DOES raise a typed 409 on the concurrent hard-delete race. Either document a 409 response entry for that scenario, or map that error to a different status in the controller. Right now the spec is inaccurate.

src/support/categories-storage.js
```diff
@@ -82,32 +196,51 @@ export async function createCategory({
...
 
   const { data, error } = await postgrestClient
     .from('categories')
-    .upsert(row, { onConflict: 'organization_id,category_id' })
+    .insert(row)
     .select()
     .single();
 
   if (error) {
     if (error.code === '23505' && /uq_category_name_per_org/.test(error.message || '')) {
```
solaris007: **Important:** The 23505 handler only matches `uq_category_name_per_org`. Slug-uniqueness violations (`uq_category_id_per_org`) - which the old `onConflict: 'organization_id,category_id'` upsert handled - now fall through to a generic 500. Apply the topics pattern: on any unhandled 23505, throw a typed 409 with the actual constraint name echoed (same as `topics-storage.js`).

solaris007 (changes requested):
Hey @rainer-friederich,

This is a substantial step up from the [prior PR 2233](https://github.com/adobe/spacecat-api-service/pull/2233) temporary 500-to-409 mitigation and addresses the actual semantic drift: name is the business key, slug is an implementation detail, so name-based idempotency is the correct contract. The intellectually honest deferred-items list and the local-review-team pre-pass are both visible in the diff quality. That said, there are a few issues to address before merge plus some follow-ups worth tracking.

## Strengths

- Root-cause fix, not another bandaid - the 7.5k/day 409 log storm disappears by design, not by reclassification
- Clean 200/201 semantics with OpenAPI documentation so type-safe clients can discriminate without body parsing
- No-op short-circuit preserves `updated_at`/`updated_by` for forensics when DRS heartbeats change nothing
- Provenance guard codifies a first system-level `human > ai` invariant with bi-directional test coverage
- `Error.cause` chaining everywhere, good diagnostic hygiene without leaking to clients
- `escapeIlike` correctly neutralizes `%`, `_`, `\\` to prevent ILIKE pattern injection
- Tenant isolation (`organization_id`) consistently scoped on every code path including resurrection and retry
- Race retry with lookup-then-resolve fold is the textbook optimistic-concurrency pattern
- Topic controller demotes 23505 to WARN as a surgical companion fix
- Thorough test coverage: no-op, update, resurrect, race, retry-lookup failure, provenance both directions, vanishing row

## Issues

### Critical (Must Fix)

**1. CI `it-postgres` is red.** Per workspace rule, do not merge on red CI. Confirm the 30s `beforeAll` timeout is pre-existing flake unrelated to this PR (the round-trip count increased from 1 to 2-3, which could marginally affect timing), rerun, and get it green.

**2. Controller still logs 409 at ERROR on the hard-delete race path.** `src/controllers/brands.js:769` logs every error - including the typed 409 that `categories-storage.js` raises on the concurrent hard-delete race (there is a dedicated test for it). The stated goal of this PR is to reduce ERROR noise - mirror the topics pattern: `if (error?.status === 409) log.warn(...)`.

**3. OpenAPI claim "Never returns 409" is contradicted by the code.** `docs/openapi/categories-v2-api.yaml:75` promises "Never returns 409: duplicate-name reposts are absorbed by design", but the storage layer DOES raise a typed 409 on the concurrent hard-delete race, and no `'409'` response entry is documented. Either document the 409 response with the exact scenario, or map that error to a different status in the controller. Right now the spec lies to consumers.

### Important (Should Fix)

**4. Origin provenance on INSERT is fully client-controlled.** The `human -> ai` guard only fires for existing rows. On fresh inserts (brand-new name or after hard-delete), the caller's `origin: 'human'` is accepted verbatim. If AI/automation callers (DRS, schedulers) are expected to be locked to `origin: 'ai'` by API enforcement, that enforcement is not in this diff. Clarify the intended policy. At minimum, consider deriving `origin` from auth context (service principal -> `ai`, human IMS user -> `human`) rather than trusting the body.

**5. Slug-uniqueness 23505 now surfaces as 500.** `src/support/categories-storage.js:226` only recognizes `uq_category_name_per_org`. The old `.upsert({onConflict: 'organization_id,category_id'})` also caught slug collisions. Scenario: org has `(slug=foo, name=Foo)`. Client POSTs `{id: 'foo', name: 'NewName'}`. Lookup by `NewName` finds nothing, INSERT trips the slug unique constraint, 23505 does not match the name regex, falls through to a generic 500. Apply the topics pattern: on any unhandled 23505, throw a typed 409 with the actual constraint name.

**6. Topics/categories asymmetry in 23505 handling.** `topics-storage.js` surfaces any 23505 as a typed 409 with constraint-name echo. `categories-storage.js` only handles the name constraint; the generic path falls through. Either apply the topics pattern to categories (preferred, fixes issue 5 above), or document why categories intentionally diverges.

**7. Case-sensitive DB constraint + case-insensitive app lookup = new duplicate-row race.** Two concurrent POSTs of "Apple" and "APPLE" both pass the case-insensitive `ilike` lookup as empty, both INSERT, and both succeed because the DB constraint `uq_category_name_per_org` is case-sensitive. The org ends up with two rows for the same semantic category. Low probability in practice but the gap is real and acknowledged with a TODO comment in the code. File a tracking ticket on the LLMO epic before close of next sprint.

**8. Topics asymmetry is a timebomb.** DRS `ensure_topics_v2` has the same shape as `ensure_categories_v2`. The 23505-to-409 mapping quiets the controller, not the storm. When topics get renamed, the 409 log storm reproduces exactly. File a scheduled follow-up with a date, not just an RFC entry.

**9. PATCH endpoint has unguarded name-collision hole.** `updateCategory` now canonicalizes `name` but does not look up collisions. A PATCH to a name that already exists on a different row in the same org raises 23505 on `uq_category_name_per_org` -> 500. Pre-existing, but this PR makes POST strictly better than PATCH - worth either fixing symmetrically or noting in OpenAPI.

**10. Resurrection x provenance test matrix is incomplete.** Missing scenarios: soft-deleted `human` + POST `origin: ai` (must resurrect AND keep `human`), soft-deleted `ai` + POST `origin: human` (must resurrect AND upgrade), soft-deleted row + POST without `origin` (must resurrect, preserve existing). The logic path exists, it just is not exercised.

**11. Zero-width / invisible char validation gap.** Tests only cover ASCII whitespace. `String.prototype.trim()` does not strip U+200B ZERO WIDTH SPACE or other non-White_Space invisibles, so `'\\u200B'` would pass `hasText` and produce an invisible row. Either strip zero-width in `canonicalizeName` or narrow the test name to reflect actual coverage.

**12. PATCH body never captured in update tests.** Most update-path tests stub the response but never assert the exact fields sent in `.update(...)`. A regression where `buildCategoryPatch` leaks an unintended field (e.g. `name`) would pass every current test. Apply the capture-and-assert pattern from the insert-canonicalization test.

**13. Silent origin-downgrade drops operator signal.** When DRS posts `origin: ai` against a human-curated row, you correctly preserve `human` but do not log it. A WARN with structured fields (`org_id`, `category_id`, `attempted_origin=ai`, `current_origin=human`) would surface misconfigured DRS taxonomies.

**14. Topic 409 error loses original `cause`.** Categories preserve `cause: error` for 23505. Topics do not. Align so operators can read the original PostgREST error during triage.

**15. Log storm reduction is unmetered.** Add a lightweight counter/log tag (`log.info(..., { path: 'idempotent-no-op' })` or metric `category_post_result{outcome=noop|update|insert|resurrect|race_retry}`) so you can quantify the cleanup in Coralogix post-deploy. Otherwise "did it work?" becomes a manual search and LLMO-4370 closure is fuzzy.

### Minor (Nice to Have)

- `escapeIlike` has no direct test. A future refactor could drop the `.replace` silently. Add a test: existing `"50 Off"` must NOT match incoming POST `"50% Off"`.
- NFC does not fold Unicode confusables (Cyrillic vs Latin A) or fullwidth/halfwidth. Low risk within a single org but enables minor spoofing.
- Slug derivation `replace(/[^a-z0-9]+/g, '-')` collapses all-non-ASCII names to `"-"`. Consider rejecting names that produce a degenerate slug.
- PostgREST error strings (operation names, DB error classes) reach 500 response bodies via `error.message`. Low-severity information disclosure.
- Audit trail can be bumped by any org member via a trivial diff. The no-op short-circuit protects identical reposts but not near-identical ones. Consider preserving `updated_by` on AI-origin updates against a human-origin row.
- 2-3 PostgREST round-trips per category. At DRS scale (20-50 categories per org x 1k orgs daily) this is measurable Aurora cost. Consider an RPC-based single-round-trip upsert.
- Return shape `{category, created}` mixes resource and control-flow signal. Consider `{resource, isNew}` or a discriminated union when the RFC unifies topics/categories.
- Resurrection silently reuses the old slug even when the client POSTs a fresh one. An `x-spacecat-resource-outcome: resurrected` response header would let clients emit telemetry without body parsing.
- Magic strings (`'deleted'`, `'human'`, `'ai'`, `'active'`, `'pending'`, `'ignored'`) repeated in JS and OpenAPI enums. A shared constants module prevents drift.
- `Error.cause` not asserted on most of the cause-adding edits (`listCategories`, `updateCategory`, `deleteCategory`, `listTopics`, `updateTopic`, `deleteTopic`).

## Recommendations

- **For merge:** fix items 1-3 (CI green, 409 log severity, OpenAPI 409 documentation). These are small and directly contradict the stated goals of the PR.
- **For same sprint:** address the 23505 generalization (items 5-6) and clarify the origin-on-insert policy (item 4).
- **For tracked follow-up on the LLMO epic:** case-insensitive DB constraint migration (item 7), topic idempotency parity (item 8), bulk PUT endpoint for DRS's "assert this taxonomy" intent, server-owned slug to eliminate the LLMO-4160 drift class permanently.
- **Architectural:** codify the `human > ai` provenance invariant in a `mysticat-architecture/` ADR before the next AI-origin resource ships. Otherwise the rule will be reinvented inconsistently per-resource.
- **Release note:** the status code contract changed from `201|409` to `200|201`. DRS appears to handle this cleanly, but please add that confirmation to the PR description so future readers do not have to re-derive it.

## Assessment

Design is sound and clearly better than the prior mitigation. Critical items 1-3 are one-to-two-line fixes that close the gap between the PR's stated goals and the shipped code. Important items 4-6 are orthogonal and small. The rest are genuinely follow-up grade, but do land the architectural ADRs and the partial-index migration within the next sprint so this does not recur on topics or the next renamed resource.

Note: `ci/it-postgres` is currently failing - this should be resolved before merge.

## Next Steps

1. Get CI green and address critical items 2-3 (409 log severity, OpenAPI 409 response).
2. Generalize 23505 handling (items 5-6) and confirm origin-on-insert policy (item 4).
3. File tracked follow-ups on the LLMO epic for items 7-8.
4. Minor items can ship independently.

rainer-friederich:
@solaris007 thanks for the detailed review. Review feedback addressed — ready for another look.

Two commits on top of the base:
- `25657b85` — critical/important/minor fixes
- `37d7fda0` — integration test updated for the new 200/201 contract (this was the actual `it-postgres` red on your review, not the 30s `beforeAll` flake — good catch pointing it out)

### Critical (all fixed)

| Item | Fix | Location |
|------|-----|----------|
| CI red | `ci/it-postgres` green | run [24670498433](https://github.com/adobe/spacecat-api-service/actions/runs/24670498433) — the real failure was the IT assertion on the old `201\|409` contract; updated to `200\|201` |
| 409 log severity | Demoted to `log.warn` on concurrent-hard-delete race and PATCH name collision | `src/controllers/brands.js:765, :839` mirrors the topics pattern |
| OpenAPI 409 | Documents both 409 paths, drops "Never returns 409" claim | `docs/openapi/categories-v2-api.yaml` POST + PATCH |

### Important (all fixed)

- **23505 generalization (items 5 + 6).** Any unique-constraint violation other than `uq_category_name_per_org` now surfaces as a typed 409 echoing the actual constraint (same pattern as topics). Closes the slug-collision (`uq_category_id_per_org`) hole. Tested on both paths.
- **PATCH name-collision (item 9).** `updateCategory` maps 23505 to 409 symmetrically with POST; controller handles both at WARN.
- **Resurrection × provenance matrix (item 10).** Added three scenarios: soft-deleted `human` + POST `ai` (resurrect, keep `human`), soft-deleted `ai` + POST `human` (resurrect, upgrade), soft-deleted + no `origin` (resurrect, preserve).
- **Zero-width / BOM stripped (item 11).** `canonicalizeName` now removes `U+200B/C/D` and `U+FEFF` before `trim()` — regression test asserts `'\u200B\u200C\u200D\uFEFF'` is rejected as empty.
- **PATCH body captured (item 12).** Update-path tests now capture what's sent to `.update(...)` and assert `name`/`category_id` never leak into the patch.
- **Downgrade WARN (item 13).** Blocked `human→ai` path emits structured `log.warn('Blocked origin downgrade...', { organization_id, category_id, attempted_origin, current_origin })` so misconfigured DRS taxonomies are findable in Coralogix without row inspection.
- **Topics 409 cause (item 14).** Topics now chain the original PostgREST error as `Error.cause`, aligned with categories.
- **Outcome tag (item 15).** Storage returns `{ category, created, outcome }` where `outcome ∈ { insert | resurrect | update | noop | race_retry | race_retry_noop }`. Controller emits `log.info('Category POST resolved...', { organization_id, category_id, outcome })` so post-deploy LLMO-4370 cleanup is quantifiable via Coralogix aggregation, not message grep.

### Minor (landed opportunistically)

- `escapeIlike` exported + direct regression tests (`%`, `_`, `\`, compound) so a future `.replace` removal breaks loudly.
- Slug derivation rejects names that canonicalize to an empty slug after `[^a-z0-9]+ → '-'` (non-ASCII-only without explicit `id`). Client can opt out by supplying an explicit `id` — test covers both paths.
- `Error.cause` asserted across `listCategories` / `updateCategory` / `deleteCategory` / `listTopics` / `updateTopic` / `deleteTopic` tests.

### Origin-on-INSERT policy (item 4)

Clarification: this PR keeps `origin` client-trusted on fresh inserts (matches the existing contract used by DRS today) and only applies the provenance guard to existing rows. Deriving `origin` from auth context (service principal → `ai`, human IMS → `human`) is the right long-term fix but requires coordination with DRS + UI — belongs in the follow-up RFC.

### Deferred to LLMO-4370 RFC / follow-up tickets

- **Item 7** — Case-insensitive DB constraint migration (partial index on `LOWER(name)` or generated `canonical_name` column).
- **Item 8** — Topics idempotency parity (name-by-idempotency + schema migration).
- **Architectural** — `human > ai` provenance ADR in `mysticat-architecture/`.
- **Bulk PUT for DRS** — `PUT /v2/orgs/:id/categories` with canonical set + single-transaction reconcile.
- **Server-owned slug** — eliminate client-supplied `category.id` to kill the LLMO-4160 drift class permanently.

### Release-note confirmation

Per your ask: the status code contract flipped from `201|409` to `200|201`. DRS (`ensure_categories_v2`) does not inspect the status code on the happy path — only handles errors — so the change is transparent to the current consumer. Documented explicitly in the OpenAPI description and referenced in the integration test so future readers do not have to re-derive it.

### Verification

- `npm test` — **8,310 passing**, 100% lines / branches / statements
- `npm run lint` — clean (only pre-existing `eslint-env` deprecation warnings)
- `npm run docs:lint` — valid (pre-existing warnings on unrelated paths only)
- `ci/it-postgres` — **444 passing** on real PostgREST/Aurora

/cc @solaris007 — re-review please when you get a chance.

solaris007 (approved):
Hey @rainer-friederich,

Thorough response. Verified each addressed item in the incremental diff and CI is green.

## Strengths (previously flagged, now resolved)

- **CI `it-postgres` now green** - and good catch on the real failure being the IT assertion still expecting the old `201|409` contract, not a flake. That was the right fix.
- **409 log severity** - `brands.js:81,97` demoted to `log.warn` on both the concurrent hard-delete race and the PATCH name collision, matching the topics pattern.
- **OpenAPI 409 documentation** - both the POST and PATCH responses now document `'409'` with the exact scenarios, and the "Never returns 409" claim is gone. Contract matches code.
- **23505 generalization** - `categories-storage.js` now catches any 23505 via `extractConstraintName()`, handles `uq_category_name_per_org` with race-recovery and everything else (including `uq_category_id_per_org` slug collisions) as typed 409 with echoed constraint name. Symmetric with topics.
- **PATCH name-collision** - `updateCategory` has the same 23505 -> 409 mapping as POST, closing the asymmetry.
- **Resurrection x provenance test matrix** - all three scenarios added (human + ai POST, ai + human POST, no-origin POST).
- **Zero-width / BOM stripped** - `ZERO_WIDTH_RE = /[\u200B-\u200D\uFEFF]/g` applied in `canonicalizeName` before `trim()`, regression test covers the `'\u200B\u200C\u200D\uFEFF'` rejection.
- **PATCH body capture** - update-path tests now assert the exact fields sent to `.update(...)` so `name`/`category_id` leaks would fail.
- **Silent origin-downgrade surfaced** - `log.warn('Blocked origin downgrade human->ai on category', { organization_id, category_id, attempted_origin, current_origin })` gives operators a findable signal in Coralogix.
- **Topic 409 cause** - topics now chain `Error.cause`, aligned with categories.
- **Log storm metric** - storage returns `{ category, created, outcome }` with outcome in `{insert|resurrect|update|noop|race_retry|race_retry_noop}`, controller emits structured `log.info('Category POST resolved...', { outcome })`. LLMO-4370 cleanup is now quantifiable via Coralogix aggregation instead of message grep.
- **escapeIlike exported + tested** - direct regression tests for `%`, `_`, `\\\\`, compound cases so future `.replace` removal breaks loudly.
- **Slug derivation rejection** - names that canonicalize to an empty slug after `[^a-z0-9]+` collapse are rejected unless client supplies explicit `id`.
- **\`Error.cause\` assertions** - added across the cause-adding edits in \`listCategories\`, \`updateCategory\`, \`deleteCategory\`, \`listTopics\`, \`updateTopic\`, \`deleteTopic\`.

## On deferred items

The item 4 clarification (origin on INSERT stays client-trusted, matches existing DRS contract) is reasonable - deriving origin from auth context requires DRS + UI coordination and belongs in the RFC. Items 7, 8, the \`human > ai\` ADR, bulk PUT, and server-owned slug are appropriately placed in the follow-up RFC. Please keep that RFC on near-term calendar so topics does not reproduce this exact pattern when its names get renamed.

## Assessment

Ready to merge? **Yes.**

Core design was sound on first pass; this re-review addresses every critical and important item with matching verification and CI is green. Follow-up RFC items are well-scoped and flagged. Ship it, and thanks for the clean per-item response in the PR comment.

PR #2243, opened 2026-04-20
fix: onboard flow fixes around bot detection and resolve canonical urls for some sites

tkotthakota-adobe:
https://jira.corp.adobe.com/browse/SITES-43237
https://jira.corp.adobe.com/browse/SITES-41916                                      

Summary                                                                                                                    
                                                                                                                             
Fixes the onboarding flow silently stopping after "Added site to existing project" for sites with Brotli-compressed      responses (e.g., otempo.com.br).                                                                           
                                                            
Fix (in spacecat-shared-utils): Added decode: false to the fetch call in resolveCanonicalUrl since it only needs the URL and status code, not the body.

Changes in this PR:
  - Wrapped detectBotBlocker in try/catch so failures don't kill the onboarding flow
  - Added 7s safety timeout on resolveCanonicalUrl via Promise.race as a defense-in-depth fallback
  - Fixed error handler to post errors to the Slack thread instead of attempting modal error responses
  after ack() was already called

Related PR: https://github.com/adobe/spacecat-shared/pull/1556


package.json
```diff
@@ -84,7 +84,7 @@
     "@adobe/spacecat-shared-athena-client": "1.9.11",
     "@adobe/spacecat-shared-brand-client": "1.1.41",
     "@adobe/spacecat-shared-content-client": "1.8.23",
-    "@adobe/spacecat-shared-data-access": "3.55.1",
+    "@adobe/spacecat-shared-data-access": "https://gist.github.com/tkotthakota-adobe/8d79aae5342c8063a23c19b5185e04bd/raw/d45d8b17fc8dd20da65f9480d20b4cdd918399e2/adobe-spacecat-shared-data-access-3.55.0.tgz",
```
solaris007: **Critical:** This gist URL (and the one at line 99) pins a production dependency to a personal GitHub account, bypassing the npm registry, SCA tooling, and the spacecat-shared release pipeline. Three of the five claimed fixes live inside these tarballs and are not reviewable in this PR. Restore registry versions once `spacecat-shared` PR 1556 ships.

src/support/slack/actions/onboard-modal.js
```diff
@@ -813,12 +829,17 @@ ${deliveryConfigInfo}${previewConfigInfo}
...
-      await ack({
-        response_action: 'errors',
-        errors: {
-          site_url_input: 'There was an error processing the onboarding request.',
-        },
-      });
+      try {
+        await client.chat.postMessage({
+          channel: responseChannel,
+          text: `:x: Onboarding failed for \`${siteUrl}\``
```
solaris007: **Important:** `responseChannel` is `undefined` if the error fires before line 651 (missing form fields, malformed body, etc.). `postMessage({ channel: undefined })` will reject, the inner catch logs it, and the user gets no feedback. Initialize early: `let responseChannel = body?.user?.id;`

src/support/slack/actions/onboard-modal.js
```diff
@@ -813,12 +829,17 @@ ${deliveryConfigInfo}${previewConfigInfo}
...
+            + `\n*Triggered by:* ${user?.name || 'unknown'} | *Profile:* ${profile} | *IMS Org:* ${imsOrgId || 'unknown'}`
+            + `\n*Error:* ${error.message}`,
+          thread_ts: responseThreadTs,
```
solaris007: **Important:** `error.message` is rendered verbatim to Slack. Downstream errors can contain SQL fragments, internal hostnames, or ARNs. Also `undefined` for non-Error throws. Log the full error for diagnostics; post a sanitized summary to Slack.

src/support/slack/actions/onboard-modal.js
```diff
@@ -725,7 +733,13 @@ export function onboardSiteModal(lambdaContext) {
         thread_ts: responseThreadTs,
       });
 
-      const botProtectionResult = await detectBotBlocker({ baseUrl: siteUrl });
+      let botProtectionResult = { crawlable: true, type: 'unknown', confidence: 0 };
```
solaris007, line 738: **Important:** If `detectBotBlocker` throws, the user gets no notification and the site is onboarded as crawlable. A site with genuine bot protection now silently skips the warning. Consider posting a Slack note on detection failure and asserting the warn log in the test.

src/support/utils.js
```diff
@@ -1702,12 +1704,13 @@ export const onboardSingleSite = async (
       }
     }
 
-    // Resolve canonical URL for the site from the base URL
     let resolvedUrl = await resolveCanonicalUrl(baseURL);
-    if (resolvedUrl === null) {
+    if (!resolvedUrl) {
```
solaris007, line 1700: **Important:** Widened from `=== null` to `!resolvedUrl` - now catches `undefined`, empty string, and other falsy values. This is a behavior change with no tests for the new cases. Add test cases for `undefined` and empty string.

solaris007 (changes requested):
Hey @tkotthakota-adobe,

Thanks for tracking down the Brotli decompression issue and the Slack 3-second ack timeout. The diagnostic work is solid, and several of the in-repo fixes are well-targeted. A few things need attention before this can merge.

## Strengths

- Moving `ack()` before `Configuration.findLatest()` (`onboard-modal.js:693`) is the correct fix for the Slack modal timeout - it targets the actual slow network call without restructuring the flow.
- The `detectBotBlocker` try/catch with a safe default (`onboard-modal.js:736-742`) is the right shape for a non-critical signal: log the failure, proceed with a fallback.
- Test coverage tracks the new behavior well: single-ack semantics (`onboard-modal.test.js:678`), error-path posting to channel (`onboard-modal.test.js:853`), fallback values for missing user/IMS (`onboard-modal.test.js:867`), inner catch for Slack failures (`onboard-modal.test.js:925`), and bot-blocker exception handling (`onboard-modal.test.js:1359`).
- The inner try/catch around the failure-notification `postMessage` (`onboard-modal.js:840-842`) prevents cascading failure if Slack itself is down during error reporting.

## Issues

### Critical (Must Fix)

**Gist-pinned production dependencies** - `package.json:87, 99`

Two production dependencies are pinned to personal-account GitHub gist URLs instead of the npm registry:

```
"@adobe/spacecat-shared-data-access": "https://gist.github.com/tkotthakota-adobe/.../adobe-spacecat-shared-data-access-3.55.0.tgz"
"@adobe/spacecat-shared-utils": "https://gist.github.com/tkotthakota-adobe/.../adobe-spacecat-shared-utils-1.112.5.tgz"
```

This breaks supply chain integrity:

- **Single point of failure.** Both gists are owned by a personal GitHub account. If the account is offboarded, the gists deleted, or credentials compromised, every fresh `npm install` breaks.
- **Review bypass.** The PR body lists five fixes, but three (the `decode: false` change, the 8s `Promise.race` timeout, and the 30s `site.save()` timeout) live exclusively in the gist tarballs - outside this PR's reviewable diff and outside the `spacecat-shared` release pipeline.
- **Version downgrade.** `@adobe/spacecat-shared-data-access` goes from `3.55.1` (registry) to `3.55.0` (gist) with no record of what `3.55.1` delivered that is now absent.
- **Invisible to security tooling.** `npm audit`, Dependabot, and the GitHub Advisory Database cannot evaluate gist URLs. Future CVEs will not surface.
- **No precedent on main.** Zero gist URLs exist in `package.json` or `package-lock.json` on main today. Merging this normalizes the pattern.
- **Partial integrity protection.** `npm ci` catches post-merge gist edits via hash mismatch, but any lockfile regeneration (new sibling dep, dedupe) silently re-resolves to whatever the gist currently serves.

**Fix:** Do not merge until `adobe/spacecat-shared` PR 1556 lands and a patch release is published to the npm registry. Restore `package.json` to registry version specifiers and regenerate the lockfile. If the fix is urgent and the shared release pipeline cannot turn around in time, publish a hotfix tag through the normal channel rather than via gist.

### Important (Should Fix)

**Error handler silently drops feedback for early errors** - `onboard-modal.js:599-651, 832-842`

`responseChannel` is declared at line 599 but only assigned at line 651. If anything throws between those lines (missing form fields, malformed body, `loadProfileConfig` rejection), the catch block runs with `responseChannel === undefined`. `client.chat.postMessage({ channel: undefined })` rejects, the inner catch logs it, and the user gets no feedback - the modal hangs until Slack times out.

The previous error path (`ack({ response_action: 'errors' })`) was self-contained and always worked. The test that covered this case ("responds with error when the onboarding request fails" - previously at line 824) was removed, and the replacement tests only cover the post-ack failure path. This is a coverage regression.

**Fix:** Initialize `responseChannel` early: `let responseChannel = body?.user?.id;` so the DM fallback is always available. If `responseChannel` is still falsy at catch time, fall back to `ack({ response_action: 'errors' })`. Restore a test for the missing-form-fields scenario.

**Error message leaks internal details to Slack** - `onboard-modal.js:836-838`

`error.message` is rendered verbatim into a Slack channel. Errors from downstream calls (`Configuration.findLatest()`, `site.save()`, PostgREST, AWS SDK) can contain SQL fragments, internal hostnames, ARNs, or stack-equivalent context. Additionally, `error.message` is `undefined` for non-Error throws, which renders as `*Error:* undefined`.

**Fix:** Log the full error with a correlation ID for diagnostics. Post a sanitized summary to Slack - at minimum `error?.message?.slice(0, 200) ?? 'unknown error'`. Consider whether `imsOrgId` belongs in the channel-visible message.

**detectBotBlocker fail-open is silent to the user** - `onboard-modal.js:736-742`

When `detectBotBlocker` throws, the default `{ crawlable: true }` is used and onboarding proceeds with no user notification. A site with genuine bot protection is onboarded as if it were crawlable, potentially leading to silent downstream scrape failures - the same class of failure this PR aims to fix.

The test at `onboard-modal.test.js:1359` asserts onboarding continues but does not assert the `log.warn` was called or any user-visible feedback.

**Fix:** Post a brief Slack note when detection fails ("Bot detection check skipped - audits may be affected if the site has bot protection"). Assert both the `log.warn` and the user-facing message in the test. Consider whether fail-closed (assume protected, show the warning) is safer for this flow.

**`!resolvedUrl` widens the check with no tests for new cases** - `utils.js:1708`

Changed from `if (resolvedUrl === null)` to `if (!resolvedUrl)`. This now catches `undefined`, `''`, and other falsy values in addition to `null`. If the upstream `resolveCanonicalUrl` now returns `undefined` on timeout, this widening is intentional - but there's no way to confirm from this diff alone, and no tests pin the new behavior.

**Fix:** Add test cases for `undefined` and `''` returns from `resolveCanonicalUrl`. A one-line comment noting why the check was widened would help future readers.

### Minor (Nice to Have)

**`[onboard]` log breadcrumbs look like incident debugging** - `onboard-modal.js:738, 759, 770`

The new `log.info("[onboard] starting...", "[onboard] completed...")` lines are useful for tracing the original incident, but free-form prefixed strings are not queryable in Coralogix without LIKE patterns. If these are intended to be permanent, switch to structured fields (`{ operation: 'onboard', siteUrl, durationMs }`). If they were temporary debug aids, remove them once the fix is stable.

**Canonical URL tests tolerate downstream throws** - `utils-paid-profile-guard.test.js:603-608`

The `try/catch` with "Downstream deps not fully mocked - expected" means the test passes as long as the assertion target is called before the unmocked code throws. If the implementation reorders calls, the test silently regresses. Either fully mock downstream deps or extract the canonical-URL logic into a testable unit.

## Recommendations

- Add a CI lint that rejects non-registry URLs in `package.json` dependencies (e.g., grep for `https://` in dependency values, excluding `registry.npmjs.org`). Prevents recurrence of gist-pinning.
- Centralize Slack error formatting into a shared helper (`formatSlackError(error, correlationId)`) used by all catch blocks in Slack action handlers. Single point to enforce redaction rules.
- The ack-before-work pattern is the right local fix, but the handler still runs the full onboarding pipeline synchronously against Slack's 3-second contract. A follow-up ADR on Slack-action async handling (ack immediately, enqueue work, post results async) would give the next contributor a paved path.

## Assessment

Ready to merge? **No**

The gist-pinned dependencies are a hard blocker - the bulk of the claimed fix is invisible to review, the audit trail is owned by a personal account, and the pattern has no precedent on main. The error-handler regression for early errors should also be fixed and tested before merge. Once the upstream packages are released through the registry and the in-repo fixes are tightened, this is a small, well-targeted set of changes.

## Next Steps

1. Wait for `adobe/spacecat-shared` PR 1556 to land, then restore `package.json` to registry versions and regenerate the lockfile.
2. Fix the catch-handler regression: initialize `responseChannel` early, restore the missing-form-fields test.
3. Sanitize `error.message` before posting to Slack.
4. The remaining items (bot blocker notification, resolvedUrl test coverage) and minor items are lower priority but worth addressing.

tkotthakota-adobe:
Thank you @solaris007 for the review. Incorporated review comments. Please also review dependent PR: https://github.com/adobe/spacecat-shared/pull/1556 after this is merged I will fix the package.json file with actual versions.

image

tkotthakota-adobe:
Hi @solaris007 Let me if anything else needed.  Before merging this PR, I will update package.json after shared lib is approved/merged.

package.json
```diff
@@ -84,7 +84,7 @@
     "@adobe/spacecat-shared-athena-client": "1.9.11",
     "@adobe/spacecat-shared-brand-client": "1.1.41",
     "@adobe/spacecat-shared-content-client": "1.8.23",
-    "@adobe/spacecat-shared-data-access": "3.56.1",
+    "@adobe/spacecat-shared-data-access": "https://gist.github.com/tkotthakota-adobe/8d79aae5342c8063a23c19b5185e04bd/raw/d45d8b17fc8dd20da65f9480d20b4cdd918399e2/adobe-spacecat-shared-data-access-3.55.0.tgz",
```
solaris007: **Critical: Gist URLs still in the diff; `adobe/spacecat-shared` 1556 not yet merged.**

The author's plan is to swap to registry versions once 1556 lands. As of this review, 1556 is still `open / merged: false`. The lockfile integrity pinning closes the runtime-substitution risk, but the structural concerns remain: trust authority is a personal GitHub account, Adobe's vendor controls and SBOM tooling do not apply, and CI/Kodiak cannot tell this is a transitional state.

Do not merge until 1556 lands, the registry versions are published, and these two lines are swapped back to semver references with `package-lock.json` regenerated. Consider draft-mode or a do-not-merge label so the gate is machine-visible. See Critical 1 in the review body.

src/support/slack/actions/onboard-modal.js
```diff
@@ -725,7 +735,18 @@ export function onboardSiteModal(lambdaContext) {
...
+      try {
+        botProtectionResult = await detectBotBlocker({ baseUrl: siteUrl });
+        log.info(`Bot blocker result for ${siteUrl}: crawlable=${botProtectionResult.crawlable}, type=${botProtectionResult.type}`);
+      } catch (e) {
+        log.warn(`Bot blocker detection failed for ${siteUrl}, proceeding with onboarding`, e);
+        await client.chat.postMessage({
```
solaris007: **Important: Bot-blocker warning `postMessage` has no inner try/catch.**

If Slack returns 429 / transient 5xx / channel-not-found here, the exception escapes this catch block and is caught by the outer error handler at line 835. The user then sees `:x: Onboarding failed for siteUrl` even though the actual onboarding pipeline did not fail - only a best-effort warning post did. This silently converts the documented fail-open behavior ("proceeding with onboarding") into a fail-closed UX.

Fix: wrap this `postMessage` in a local try/catch (same shape as the outer error handler's nested try at lines 855-857). Add a regression test by rejecting `client.chat.postMessage` once while keeping `onboardStub` happy. See Important 1 in the review body.

solaris007 (changes requested):
Hey @tkotthakota-adobe,

Thanks for the thorough turnaround on the prior review. The four Important findings from 2026-05-04 are all genuinely addressed in commit a4c137ca: `responseChannel` early-init plus the branched `if/else` catch handler, the `safeMessage` truncation with `unknown error` fallback, the user-visible bot-blocker warning, and the documented falsy-check on `resolveCanonicalUrl`. The new tests pin each of these behaviors. The `[onboard]` prefix cleanup also landed.

The Critical from the prior pass (gist-pinned production dependencies) is still in the diff. I see your two top-level comments and the plan to swap to registry versions before merge once `adobe/spacecat-shared` 1556 lands. As of this review, 1556 is still open. One new Important finding around the bot-blocker fail-open path, plus a few Minor cleanups.

## Strengths

Prior findings now addressed in a4c137ca:

- Important 1 (early errors): `let responseChannel = body?.user?.id;` at `src/support/slack/actions/onboard-modal.js:600` plus the `if (responseChannel) postMessage else ack({ response_action: 'errors' })` branch in the catch block. Two new tests cover both arms of the branch.
- Important 2 (error message): `const safeMessage = (error?.message || 'unknown error').slice(0, 200);` cleanly bounds the leaked surface and handles malformed Error objects. Test asserts `errorLine.length <= 210` to enforce the contract.
- Important 3 (bot-blocker fail-open): the catch now posts a `:warning:` to Slack and calls `log.warn(msg, e)` with the error as a structured second argument, not interpolated. Test was renamed and asserts both the warning message presence and `log.warn` invocation.
- Important 4 (resolvedUrl widening): comment at `src/support/utils.js:1707` documents the three failure modes (null/undefined/empty), and two new tests in `test/support/utils-paid-profile-guard.test.js` pin both new branches.
- Minor 1 (`[onboard]` prefix): the two free-form info-log breadcrumbs were removed; the remaining `[onboard]` prefix on bot-blocker logs was dropped. The warn now passes the error object structurally.

Additional defensive wins:

- Lockfile integrity hashes are present and unchanged for both gist tarballs (sha512 entries in package-lock.json), so `npm ci` will refuse to install if the gist bytes change. The gist URLs themselves pin a specific raw revision via the SHA in the path. The runtime substitution risk is mitigated even though the supply-chain trust authority is still wrong.
- The new `ack({ response_action: 'errors' })` fallback uses a server-controlled static string, not interpolated error text - no info-disclosure path through that branch.
- The bot-blocker user-warning is a strict improvement over silent fail-open: operators now learn that bot detection was skipped instead of silently onboarding sites with bot protection as crawlable.

## Issues

### Critical (Must Fix)

1. Gist-pinned production dependencies remain in `package.json:87,99`. Status: pushed_back, conditional on `adobe/spacecat-shared` 1556 landing.

The author's plan is right in shape: land 1556, publish registry releases of the patched `@adobe/spacecat-shared-data-access` and `@adobe/spacecat-shared-utils`, swap these two lines back to semver references, regenerate the lockfile. The plan does not, however, resolve the concern about the current state of the diff - it defers it. As of this review:

- `adobe/spacecat-shared` 1556 is still `open`, `merged: false`, `mergeable_state: clean` but `CHANGES_REQUESTED`.
- The two gist URLs and their pinned tarballs are still on this branch.
- All the protective gates against premature merge (Kodiak passing, CI green, registry advisory tooling) cannot tell that the package.json is in a transitional state. A rebase or a tired reviewer can land this as-is.

The lockfile hash pinning closes the runtime-substitution risk, which is a real win. But the structural concerns from the prior review remain: trust authority shifts to a personal account, Adobe's vendor controls and SBOM tooling do not apply to gist URLs, and the pattern itself sets an unrecorded precedent for future cross-repo coordination.

Fix: do not merge with gist URLs in `package.json`. Land 1556, publish registry versions, swap and regenerate the lockfile in a final commit on this branch. Consider converting this PR to draft or adding a do-not-merge label until that final commit is in, so the gate is machine-visible rather than dependent on author/reviewer discipline.

### Important (Should Fix)

1. Bot-blocker warning postMessage has no inner try/catch. `src/support/slack/actions/onboard-modal.js:744-748`.

The new warning postMessage in the bot-blocker catch block is not wrapped in its own try/catch. If `client.chat.postMessage` throws (Slack 429, transient 5xx, channel hiccup), the exception escapes the bot-blocker catch and is captured by the outer catch handler at line 835. The user then sees `:x: Onboarding failed for siteUrl` even though the actual onboarding pipeline did not fail - only a best-effort warning post did.

This converts the documented fail-open behavior ("proceeding with onboarding") into a fail-closed UX. It is the same class of regression the prior review's Minor on silent log breadcrumbs was guarding against, but in the user-facing direction.

Fix: wrap the warning postMessage in a local try/catch identical in shape to the outer error-handler's nested try/catch at lines 855-857. Log on inner failure and let onboarding continue. Add a regression test by rejecting `client.chat.postMessage` once while keeping `onboardStub` happy, similar to the existing "should log error when channel notification fails" test.

### Minor (Nice to Have)

1. IMS Org line removed from the failure-notification template. `src/support/slack/actions/onboard-modal.js:842-844`.

The error template went from `*Triggered by:* ... | *Profile:* ... | *IMS Org:* ${imsOrgId || 'unknown'}` to `*Triggered by:* ... | *Profile:* ...`. The original Important from the prior review was about leaking `error.message` to Slack; the IMS Org id is operator-relevant context, not a secret (the success-path template still includes related identifiers). This template is the operator's primary breadcrumb when triaging a failed onboard, and dropping the tenant disambiguator makes triage harder. Fix: restore `| *IMS Org:* ${imsOrgId || 'unknown'}` to the failure template.

2. Canonical-URL fallback tests still swallow downstream throws. `test/support/utils-paid-profile-guard.test.js`, four tests now share the same `try { ... } catch { /* Downstream deps not fully mocked - expected. */ }` pattern.

Per prior Minor 2: this is unresolved and the surface doubled (2 -> 4 tests). The positive assertion `updateOnboardConfigStub.calledOnce` partially mitigates the trivial-pass risk, but the negative assertion `updateFetchConfigStub).not.to.have.been.called` is still true if a future refactor reorders calls so the assertion target is never reached. Fix: tighten by asserting which arguments were passed to `updateOnboardConfigStub` (proving the baseURL fallback was used, not just that some call happened), or fully mock the downstream so the test runs to completion synchronously.

3. 200-char `safeMessage` truncation is a soft bound, not a credential redaction. `src/support/slack/actions/onboard-modal.js:836`.

Most credential shapes fit inside 200 characters and would be reflected verbatim into Slack if they appear in the underlying `error.message`: AWS access key IDs (`AKIA...`, 20 chars), bearer tokens, postgres connection strings with userinfo (`postgres://user:pass@host:5432/db`), IMS org IDs. The 200-char cap meaningfully bounds stack-trace-like leakage but does not redact secrets that the underlying error might already contain.

The trust assumption is "no caller of `onboardSingleSiteFromModal` throws an Error whose `.message` contains a secret." That assumption is currently silent. Fix (optional, non-blocking): either redact known credential patterns before truncation (strip `://user:pass@` userinfo, mask `AKIA[0-9A-Z]{16}`), or add a one-line comment near `safeMessage` documenting the assumption so future maintainers know what the truncation is and is not doing.

4. Ack-fallback test does not assert mutual exclusivity. `test/support/slack/actions/onboard-modal.test.js:986-996`.

The "should fall back to ack errors when responseChannel is unavailable" test asserts `ackMock` was called with the errors payload but does not assert `clientMock.chat.postMessage` was NOT called. The two branches in source are mutually exclusive; a future refactor that calls both paths would still pass this test. Fix: add `expect(clientMock.chat.postMessage).to.not.have.been.calledWith(sinon.match({ text: sinon.match(':x: Onboarding failed') }))` to lock in the contract.

5. Test name vs behavior mismatch. `test/support/slack/actions/onboard-modal.test.js:954-984`.

"should use safe defaults when error has no message and siteUrl is not yet parsed" uses the full `body` fixture (with view) so `siteUrl` IS parsed before the throw at the mocked `onboardSingleSite`. The unparsed-siteUrl scenario is already covered by the preceding test at line 942. The permissive `.or(unknown site)` assertion in this test hides the fact that only the `unknown error` branch matches. Fix: rename to "should fall back to 'unknown error' when error has no message" and drop the `.or('unknown site')` so the assertion is sharp.

6. Bot-blocker warning may be chatty under transient upstream failures.

`detectBotBlocker` is an advisory check. A transient upstream wobble (network blip, rate limit) now produces a user-visible `:warning:` per onboard attempt. Repeated warnings train operators to ignore that warning class. Fix (architect-level recommendation, not for this PR): consider a single retry-with-backoff inside `detectBotBlocker` so users only see the warning when the check is structurally unavailable, not transiently failing.

## Recommendations

- Codify the cross-repo coordination pattern. The gist-tarball approach is a one-off that worked here because of author discipline. To prevent it from becoming the default for future shared-lib + service joint changes, document an alternative in `mysticat-architecture`. Two viable shapes: (a) merge the shared lib first behind a feature flag, release, then land the consumer, or (b) use a pre-release tag (`@adobe/spacecat-shared-utils@1.113.0-rc.1`) from the npm registry rather than a gist. Option (b) keeps everything inside the existing supply chain and is the industry-standard pattern.
- Lift the Slack-action error UX into a shared utility. The combination of (a) `responseChannel` early-init, (b) `safeMessage` truncation, (c) `if (channel) postMessage else ack({response_action:'errors'})`, and (d) inner `try { ... } catch { log.error ... }` is now general-purpose enough to be reusable. A `src/support/slack/error-feedback.js` helper would make the pattern reusable and the modal handler shorter to read. Worth tracking for the next Slack-action handler PR.
- Define a wall-clock budget for post-ack work in Slack-action handlers. The 3-second ack contract is being honored, but the work after `await ack()` (Configuration.findLatest, detectBotBlocker, postMessage, onboardSingleSiteFromModal, brand-profile-agent trigger, final report card) is unbounded. Every new step added compounds toward the Lambda timeout. The architecturally correct shape is enqueue-to-worker after ack; until then, capture the post-ack budget explicitly so future additions are evaluated against it. Track in a follow-up; do not block this PR.

## Out of Scope, Worth Tracking

- `adobe/spacecat-shared` 1556 is the actual root-cause fix (decode: false on resolveCanonicalUrl's fetch). Reviewable in its own repo. The coordination implication is captured under Critical 1.
- Invalid or deleted channel ID where `responseChannel` is truthy but Slack returns `channel_not_found`: the postMessage rejects and is caught by the inner try/catch at lines 855-857. Behavior is correct as-coded but the deleted-channel case is not test-covered. Adding that test expands the scope beyond this PR.

## Assessment

Ready to merge? No, not until Critical 1 is closed (spacecat-shared 1556 merged and registry versions swapped into this branch's package.json + lockfile). The Important bot-blocker postMessage gap should land in the same revision; the Minor items can fold in or follow up.

Reasoning: the four prior Important findings are correctly addressed in code and well-tested. The Critical from the prior review still applies because the gist URLs are still in the diff and the merge gate is currently informal. Once 1556 lands and this PR's package.json is restored to registry versions, the one new Important and the Minor cleanups are quick edits.

## Next Steps

1. Wait for `adobe/spacecat-shared` 1556 to merge and registry releases to publish, then restore `package.json` lines 87 and 99 to semver references and regenerate `package-lock.json` (Critical 1).
2. Wrap the bot-blocker warning `postMessage` in a local try/catch with a regression test (Important 1).
3. Optionally fold in the Minor items in the same revision: restore IMS Org line, tighten the canonical-URL tests, rename the misleading test, lock mutual exclusivity on the ack-fallback test, add a comment on the truncation contract.

tkotthakota-adobe:
Hi @solaris007 Sorry for the back and forth. I believe address all comments except tar ball comment as I am waiting for below shared lib to be approved and merged. After that I will validate again. 
Shared lib PR: https://github.com/adobe/spacecat-shared/pull/1556

package.json
solaris007: **Critical (third round): gist-pinned production dependencies remain.**

The author's commitment to swap to registry versions once `adobe/spacecat-shared` 1556 merges is genuine and there is a clear paper trail across three top-level comments. The structural concern is unchanged from Review 2:

1. **The gate is informal.** Nothing in CI, the merge queue, or the lockfile prevents this PR from landing with gist URLs in place. A different reviewer working from a stale recollection can land this as-is.
2. **adobe/spacecat-shared 1556 is verified still open** as of this review (`state: open`, `merged: false`, `mergeable_state: clean`). The blocker has not moved since Review 2.
3. **The lockfile downgrades registry versions while pinning the gists.** `@adobe/spacecat-shared-data-access` moves to `3.55.0` and `@adobe/spacecat-shared-utils` moves to `1.112.5`, both lower than what was on main. Anything fixed between the prior registry version and these gist tarballs (including any security fixes) becomes invisible to `npm audit` / Dependabot for as long as this lands.
4. **Trust authority unchanged.** Integrity hashes pin specific tarballs today, but any lockfile regen on rebase silently re-resolves the gist URL. If the gist is edited, deleted, or the account is offboarded, the next install breaks or pulls something different.

Fix (any of):
- Wait for 1556 to merge + npm registry release, then swap both lines to registry versions and regenerate the lockfile in a final commit.
- If business pressure requires landing sooner, publish the patched build to an internal Artifactory / npm Enterprise feed under the `@adobe` scope so the registry-audit surface is preserved.
- At minimum, convert this PR to draft or apply a `do-not-merge` label so the gate is machine-visible.

I will not re-flag this on subsequent rounds if the gate becomes machine-visible. See review body.

test/support/slack/actions/onboard-modal.test.js
```diff
@@ -1259,6 +1453,85 @@ describe('onboard-modal', () => {
...
+        const modalAction = testModule.onboardSiteModal(context);
+
+        // Should not throw — Slack warning failure must not bubble to outer error handler
+        await modalAction({ ack: ackMock, body, client: clientWithFailingPost });
+
+        // Onboarding should still complete
+        expect(onboardStub.calledOnce).to.be.true;
+
+        // Slack failure should be logged, not thrown
+        expect(context.log.warn).to.have.been.called;
```
solaris007: **Minor: regression test `log.warn` assertion is too loose.**

`expect(context.log.warn).to.have.been.called` passes on either of the two warn lines in this code path: source line 743 (outer bot-blocker catch, pre-existing) or line 751 (the NEW inner Slack-fail catch this test is meant to lock in). If a refactor removed the inner try/catch, this assertion would still trivially pass; only the success-message assertion would catch the regression.

Tighten to:
```js
expect(context.log.warn).to.have.been.calledWith(
  sinon.match(/Failed to post bot-blocker warning/),
  sinon.match.instanceOf(Error),
);
```

test/support/slack/actions/onboard-modal.test.js
```diff
@@ -824,22 +824,216 @@ describe('onboard-modal', () => {
...
+      delete bodyNoName.user.name;
+      bodyNoName.view.state.values.ims_org_input.ims_org_id.value = '';
+      const contextNoOrg = { ...context, env: { ...context.env, DEMO_IMS_ORG: '' } };
+
+      const onboardSiteModalAction = onboardSiteModalThrows(contextNoOrg);
+      await onboardSiteModalAction({ ack: ackMock, body: bodyNoName, client: clientMock });
+
+      expect(clientMock.chat.postMessage).to.have.been.calledWith(sinon.match({
+        text: sinon.match('*Triggered by:* unknown'),
+      }));
```
solaris007, line 902: **Minor: no test asserts the restored IMS Org line appears in the failure template.**

The restored `*IMS Org:* ${imsOrgId || 'unknown'}` at `src/support/slack/actions/onboard-modal.js:847` is silent: this test asserts `*Triggered by:* unknown` but not `*IMS Org:* unknown`. Silent additions to user-facing templates regress without warning.

Cheap fix:
```js
expect(clientMock.chat.postMessage).to.have.been.calledWith(sinon.match({
  text: sinon.match('*Triggered by:* unknown').and(sinon.match('*IMS Org:* unknown')),
}));
```

solaris007 (changes requested):
Hey @tkotthakota-adobe,

The 28732ac commit cleanly closes every Review 2 Important and Minor that was within scope of this PR. The bot-blocker inner try/catch is well-shaped with the right log-level discipline, the new regression test is functionally correct (sinon stub call-order semantics verified to honor the per-call override regardless of default placement), and the test rename + mutual-exclusivity assertion + IMS Org restoration are all on target.

The Critical from the prior two rounds (gist-pinned production deps) is still in the diff because `adobe/spacecat-shared` 1556 has not merged yet. Your commitment to swap is genuine and on the record across three top-level comments; my concern is that the gate is informal and the lockfile presents a slightly worse story than I initially flagged.

## Strengths

**Previously flagged issues now addressed in commit 28732ac:**

- Review 2 Important (bot-blocker postMessage no inner try/catch) - `src/support/slack/actions/onboard-modal.js:744-752`. Inner try/catch added; the differentiated log levels are correct (`log.warn` here for "lost a nice-to-have notification while the run continues" vs `log.error` at the outer catch for "lost visibility into onboarding failure"). Symmetric with the outer fail-open handler at line 859-861.
- Review 2 Minor (IMS Org line removed) - `src/support/slack/actions/onboard-modal.js:847`. Restored with `imsOrgId || 'unknown'` fallback, consistent with peer fields. Useful triage breadcrumb in the operator channel.
- Review 2 Minor (ack-fallback mutual exclusivity) - `test/support/slack/actions/onboard-modal.test.js:996-998`. Mutual-exclusivity assertion locks both branches.
- Review 2 Minor (test name mismatch) - `test/support/slack/actions/onboard-modal.test.js:954` plus the sharpened `sinon.match('unknown error')` assertion at line 982.

**New regression test (test:1495-1534):** correctly exercises the inner-catch path. Empirically verified: `stub.onSecondCall().rejects()` followed by `stub.resolves()` honors the per-call override regardless of order (sinon's documented "set default after per-call" pattern). Mock completeness verified - `Configuration.findLatest` is already stubbed in the outer `beforeEach`. The success-message search does not collide with "Starting onboarding..." so it load-bears the regression check.

## Issues

### Critical (Must Fix)

**Gist-pinned production dependencies remain in the diff.** `package.json:87, 99`. The author's commitment to swap to registry versions once `adobe/spacecat-shared` 1556 merges is genuine and there is a clear paper trail in the top-level comments. The structural concern is unchanged from Review 2:

1. The gate is informal. Nothing in CI, the merge queue, or the lockfile prevents this PR from landing with gist URLs still in place. A different reviewer working from a stale recollection can land this as-is.
2. **adobe/spacecat-shared 1556 is verified still open** as of this review (state: open, merged: false, mergeable_state: clean). The blocker has not moved since Review 2.
3. **The lockfile downgrades registry versions while pinning the gists.** `@adobe/spacecat-shared-data-access` moves from a more recent registry version to `3.55.0` (gist), and `@adobe/spacecat-shared-utils` moves to `1.112.5` (gist). Anything fixed between the prior registry version and the gist version (including any security fixes) becomes invisible to `npm audit` / Dependabot for as long as this lands.
4. Trust authority is unchanged. `tkotthakota-adobe` is a personal GitHub account. Integrity hashes pin specific tarballs today, but any lockfile regen on rebase silently re-resolves the URL. If the gist is edited, deleted, or the account is offboarded, the next install breaks or pulls something different.

**Fix (any of):**
- Wait for 1556 to merge, publish to npm registry, then swap both lines to registry versions and regenerate the lockfile in a final commit on this branch.
- If business pressure requires landing sooner, publish the patched build to an internal Artifactory / npm Enterprise feed under the `@adobe` scope so the registry-audit surface is preserved.
- At minimum, convert this PR to draft or apply a `do-not-merge` label so the gate is machine-visible rather than dependent on reviewer discipline.

I will not re-flag this on subsequent rounds if the gate becomes machine-visible.

### Minor (Nice to Have)

1. **Regression test `log.warn` assertion is too loose.** `test/support/slack/actions/onboard-modal.test.js:1529` asserts `expect(context.log.warn).to.have.been.called`. Inside the code path under test, `log.warn` fires twice: at source line 743 (outer bot-blocker catch, pre-existing) and at line 751 (the NEW inner Slack-fail catch). The assertion passes on either - if a refactor removed the inner try/catch, this assertion would still trivially pass; only the success-message assertion would catch the regression. Tighten to:
   ```js
   expect(context.log.warn).to.have.been.calledWith(
     sinon.match(/Failed to post bot-blocker warning/),
     sinon.match.instanceOf(Error),
   );
   ```

2. **No test asserts the IMS Org line appears in the failure-notification template.** `test/support/slack/actions/onboard-modal.test.js:866-903, 1001-1037`. The restored `*IMS Org:* ${imsOrgId || 'unknown'}` at source line 847 is silent: the closest test at line 866 asserts `*Triggered by:* unknown` but not `*IMS Org:* unknown`. Silent additions to user-facing templates regress without warning. Cheap fix:
   ```js
   expect(clientMock.chat.postMessage).to.have.been.calledWith(sinon.match({
     text: sinon.match('*Triggered by:* unknown').and(sinon.match('*IMS Org:* unknown')),
   }));
   ```

## Recommendations

- **Codify "no gist-tarball deps on `main`" as a workspace rule** with the two viable alternatives already on the recommendations list (feature-flag in the shared lib, or npm prerelease/dist-tag). Otherwise this PR's transitional pattern becomes the default reach when an upstream PR is slow.
- **Lift the Slack-action postMessage-with-fallback pattern into a shared helper.** The combination of "attempt `client.chat.postMessage`, catch and `log.warn`, swallow" now appears three times in `onboard-modal.js` (the outer catch's nested try at line 859-861, the bot-blocker inner catch added in this commit at 749-751, and the warning post at line 765-768). Each call site is fine; the cumulative duplication is what argues for `postWithFallback(client, log, channel, message)`. The cost-of-doing-nothing curve is bending up.
- **Sinon stub setup readability nit.** `test/support/slack/actions/onboard-modal.test.js:1499-1501`. The setup is functionally correct (verified above) but reads better as default-first, per-call-override-second:
   ```js
   const postMessageStub = sinon.stub().resolves({ ok: true });
   postMessageStub.onSecondCall().rejects(new Error('channel_not_found'));
   ```
   One-line reorder, removes the "does the later `.resolves()` reset the per-call behavior?" question.
- **Post-ack pipeline budget.** No new erosion in 28732ac (added work is a try/catch and a string concat, no new RPCs). Prior recommendation to move toward enqueue-to-worker stands; the trigger should be the first prod Lambda timeout in this handler, not a speculative refactor now.

## Out of Scope, Worth Tracking

- Prior Minor (canonical-URL fallback tests in `test/support/utils-paid-profile-guard.test.js:591, 618, 644, 670` tolerate downstream throws via `try { } catch { /* Downstream deps not fully mocked. */ }`). Unchanged across two prior rounds; not re-raising at severity. Worth capturing as a follow-up issue.
- Prior Minor (200-char `safeMessage` cap is a soft bound, not credential redaction). Unchanged. Not worth re-raising at any severity - error paths here are not attacker-influenced and full error detail still lands in CloudWatch / Coralogix via `log.error`. The author has already triaged it.
- `adobe/spacecat-shared` 1556 is the actual root-cause fix (decode: false on resolveCanonicalUrl). Reviewable in its own repo.

## Assessment

Ready to merge? **No.**

Reasoning: The in-repo changes are sound and the discussion is mature; commit 28732ac cleanly addresses every Review 2 Important/Minor that was within scope. The only remaining merge blocker is the gist-pinned production deps, which the author has acknowledged but the gate around it is informal. Keep this PR queued; merge once 1556 ships and the deps point at a released `@adobe/spacecat-shared-*` version on the npm registry. The two new test-quality Minors are optional polish.

## Next Steps

1. Gate this PR's merge on `adobe/spacecat-shared` 1556 merging plus a published registry release. Either convert this PR to draft or apply a `do-not-merge` label so the gate is machine-visible.
2. When the upstream release ships, swap `package.json` lines 87 and 99 to registry version specifiers and regenerate `package-lock.json` in a final commit.
3. Optionally fold in the two test-quality Minors and the sinon-ordering recommendation in the same commit.

package.json
tkotthakota-adobe: Agreed. My idea is to update it after share lib PR is approved and merged with new lib version then request for re-review.

test/support/slack/actions/onboard-modal.test.js
tkotthakota-adobe: addressed

test/support/slack/actions/onboard-modal.test.js
tkotthakota-adobe: addressed

solaris007 (approved):
Hey @tkotthakota-adobe,

Conditional approve. Both Review 3 Minors are properly addressed in 416a4ad and the source-vs-test semantic coupling holds up under inspection. The remaining merge gates are mechanical (CI red, upstream PR not yet merged), so I am unblocking the formal review state and trusting you to land the last steps as planned. One small cleanup worth handling in the same final push.

## Strengths

Previously flagged issues now addressed in commit 416a4ad:

- Minor (regression test `log.warn` assertion too loose) - fixed at `test/support/slack/actions/onboard-modal.test.js:1529`. The regex `/Failed to post bot-blocker warning/` matches the exact source call at `src/support/slack/actions/onboard-modal.js:751`, distinguishing it from the sibling `log.warn` at line 743 (`Bot blocker detection failed...`) which also fires in this test path. The added `sinon.match.instanceOf(Error)` correctly types the rejected error as the second argument. The test now actually exercises the inner-catch.
- Minor (no test for restored IMS Org line) - fixed at `test/support/slack/actions/onboard-modal.test.js:898-902`. The combined `sinon.match('*Triggered by:* unknown').and(sinon.match('*IMS Org:* unknown'))` matches both substrings against the single `text` field rendered from the failure-notification template at `src/support/slack/actions/onboard-modal.js:847`. The test setup nulls `user.name` and blanks `ims_org_id.value` plus `DEMO_IMS_ORG`, so both fallbacks fire together in the same message body - the co-occurrence is real and load-bearing.

## Issues

### Minor (Nice to Have)

1. `.claude/commands/diagnose-cwv.md` - scope-bleed and portability. The file was added in commit 416a4ad alongside the test review fixes, but the CWV diagnostic command is unrelated to this PR's stated scope (Brotli fix + Slack error handling + canonical-URL timeout):
   - The command references `~/projects/expsuccess/spacecat/scripts/cwv/diagnose-cwv-codefix.sh`, a developer-specific path that does not exist for anyone else who clones the repo. A `.claude/commands/` file is shared team tooling, so baking in one engineer's local layout makes the primary command path a no-op for everyone else. The "replicate manually with curl + jq" fallback is a tell that the main instruction is not generally usable.
   - The command instructs Claude to ask the user to run `! export API_KEY=<key>`, which writes the key into shell history (`~/.bash_history`/`~/.zsh_history`) in plaintext. Even for dev API keys, this is the kind of copy-paste template that propagates; prefer `read -s API_KEY`, sourcing from `mysticat auth token`, or a secrets-file pattern.

   Fix: drop this file from the PR. If the diagnostic is valuable, land it separately with the script checked into `scripts/cwv/diagnose-cwv-codefix.sh` and a non-`export` pattern for the API key.

## Expected Before Merge

These are mechanical gates rather than open review items - flagged so the sequence is on record:

1. `adobe/spacecat-shared` 1556 merges and publishes a release. Status as of this review: OPEN, BEHIND main, not merged.
2. Swap the gist URLs in `package.json:87,99` for the published version specifiers, regenerate `package-lock.json`. The current CI failure on `test/controllers/sites.test.js` (25 `ASO_PLG_EXCLUDED_ORGS` assertions plus coverage below 100%) is almost certainly the gist-pinned `@adobe/spacecat-shared-data-access@3.55.0` mismatching the latest main; the dep swap should resolve it.
3. Drop `.claude/commands/diagnose-cwv.md` (ideally) or split into its own PR with the referenced script checked in.
4. Confirm CI goes green.

## Out of Scope, Worth Tracking

A machine-visible gate (`do-not-merge` label, draft state, or a status check tied to upstream PR state) would still be useful for future PRs in this transitional-dependency pattern. Not re-raising as a finding given the work already put in across four rounds, but worth carrying as a workspace-level convention discussion.

## Assessment

Ready to merge? Yes, conditionally - once items 1, 2, 4 above are satisfied. Item 3 (the `.claude` file) should ideally be addressed but does not have to be in this PR if you would rather not respin.

Reasoning: The in-repo work is sound, the test coverage is meaningfully stronger than at Review 1, and the gist-deps situation has a documented swap plan tied to a real upstream PR. CI is the active merge blocker today, and a green CI run after the dep swap is the actual signal that the path is clear. Unblocking the formal review state to reflect that judgment.

tkotthakota-adobe:
Hi @solaris007 .claude/commands/diagnose-cwv.md got added accidentally. Sorry for the confusion. I will remove it.

ramboz (approved):

PR #2245, opened 2026-04-21
feat(llmo): expand onboarding CDN detection to all CDN providers | LLMO-4091

AddyKen-ghost:
## Summary

Extends the LLMO onboarding CDN detector so it identifies every CDN selectable in the UI (not just AEM CS Fastly / Commerce Fastly), unlocking auto-selection and mismatch warnings in the `CDNOnboardDialog` for the full BYOCDN catalogue. Logic is ported — not shared — from `spacecat-audit-worker/src/detect-cdn/cdn-detector.js` (per the decision to not touch audit-worker).

### Detection logic — `src/support/cdn-detection.js`
Two-phase detector:
- **Phase 1** (unchanged): DNS-only fast-path for Adobe-managed CDNs. Returns `aem-cs-fastly` / `commerce-fastly` on CNAME / A / AAAA match against the known Fastly signatures.
- **Phase 2** (new, ported from audit-worker): when Phase 1 doesn't match, run a multi-signal probe:
  1. HTTP `HEAD` (with `GET` retry) → header fingerprinting for 17 providers.
  2. CNAME chain walk via system resolver, then DoH.
  3. ASN lookup via `ipinfo.io` for the first resolved IPv4.
  4. PTR reverse-DNS keyword match as a last-resort tiebreaker.

A new `LABEL_TO_LLMO_TOKEN` adapter translates the detector's descriptive labels (`"Cloudflare"`, `"Vercel"`, …) into the LLMO `byocdn-*` vocabulary the rest of the api-service / UI understands.

Uses `tracingFetch` (strict TLS) for all HTTP probes — `NODE_TLS_REJECT_UNAUTHORIZED` is intentionally not consulted here (the audit-worker only sets it in local dev; setting it in production would be a security regression).

### Result contract
`detectCdnForDomain` still returns a three-state result, extended to cover the new tokens:
- `aem-cs-fastly` / `commerce-fastly` — Phase 1 hit
- `byocdn-<provider>` — Phase 2 hit, adapter-mapped
- `'other'` — both phases ran cleanly, nothing matched
- `null` — at least one signal failed (inconclusive)

### OpenAPI — `docs/openapi/llmo-api.yaml`
`detectedCdn` enum widened from 4 values (`aem-cs-fastly`, `commerce-fastly`, `other`, `null`) to 27: the 2 Adobe-managed tokens + 23 `byocdn-*` tokens + `other` + `null`. Description updated to clarify the auto-select-vs-stored-only split.

### Tests — `test/support/cdn-detection.test.js`
- All 38 existing Phase 1 tests preserved (stubbed `tracingFetch` and `dns.reverse` in `beforeEach` so Phase 2 short-circuits cleanly).
- 31 new Phase 2 tests:
  - Header → token mapping for each of the 17 fingerprintable providers.
  - DNS fallback via CNAME (Cloudflare, Akamai) and DoH CNAME.
  - ASN lookup (AS13335 → Cloudflare, AS54113 → Fastly).
  - PTR keyword match.
  - Phase 1 wins over Phase 2 (short-circuit assertion).
  - HEAD failure → GET fallback.
  - URL vs hostname input, empty / whitespace / malformed input.
  - `'other'` vs `null` semantics.

Combined: **69 passing** in `cdn-detection.test.js`, **527 passing** in the related suite (cdn-detection + edge-routing-utils + llmo + llmo-onboarding).

## UI companion PR
- UI: https://github.com/adobe/project-elmo-ui/pull/1569

## What's intentionally not touched
- `spacecat-audit-worker` — logic is duplicated here per product decision; drift will be managed manually.
- `src/controllers/llmo/llmo-utils.js` (`CDN_TYPES`) — stays aligned with auth-service; the new tokens are introduced only in the OpenAPI contract and the UI types.
- `src/support/edge-routing-utils.js#detectCdnForDomain` — separate AEM-CS-only function called by the edge-optimize guard; different concern.
- `spacecat-shared/packages/spacecat-shared-utils/src/cdn-helpers.js#CDN_TRANSFORMATIONS` — keyed by user-selected `logSource`, not `detectedCdn`; no changes needed.

## Latency note
Phase 2 mirrors audit-worker timings (10s HEAD, 10s GET, 15s DNS/ASN race). Worst case ~35s — fits within Lambda's 15-min budget. We'll monitor p95/p99 in production and tighten if needed.

## Test plan
- [x] `npx mocha test/support/cdn-detection.test.js` — 69 pass
- [x] `npx mocha test/support/edge-routing-utils.test.js test/controllers/llmo/llmo.test.js test/controllers/llmo/llmo-onboarding.test.js` — all pass (527 total)
- [x] `npx eslint` on changed files — clean
- [x] Redocly OpenAPI lint (pre-commit hook) — valid, 0 new warnings
- [x] Stage smoke: onboard a Cloudflare-fronted site, confirm `detectedCdn=byocdn-cloudflare` in response + stored on siteConfig


src/support/cdn-detection.js
```diff
@@ -35,77 +63,672 @@ function catchDnsLookup(err) {
...
+];
+
+/**
+ * Translates the descriptive label returned by the audit-worker-style detector
+ * into the LLMO byocdn-X vocabulary stored on siteConfig.llmo.detectedCdn.
+ *
+ * Labels not present here fall back to 'other'. The Adobe-managed Fastly /
+ * Commerce Fastly cases are handled by Phase 1 and never reach this map.
+ */
+const LABEL_TO_LLMO_TOKEN = {
```
akshaymagapu, line 259: Add a defensive allowlist here: return `'other'` for any token the pinned `spacecat-shared` Joi schema doesn't yet accept. `spacecat-shared-data-access/src/models/site/config.js:348` currently validates `detectedCdn` as `['aem-cs-fastly','other']`, so `commerce-fastly` and every `byocdn-*` token will fail `validateConfiguration(state)` on the next unrelated config write (called at `config.js:805, 888, 900`). Decouples future detector widening from shared-repo releases.

src/support/cdn-detection.js
```diff
@@ -35,77 +63,672 @@ function catchDnsLookup(err) {
...
+
+function matchCdnByCname(cnameChain) {
+  /* c8 ignore next 3 -- defensive; callers always pass a non-empty chain */
+  if (!Array.isArray(cnameChain) || cnameChain.length === 0) {
+    return null;
+  }
+  for (const { domains, cdn } of CDN_DOMAIN_SIGNATURES) {
+    for (const hostname of cnameChain) {
+      const lower = hostname.toLowerCase();
+      if (domains.some((d) => lower.includes(d))) {
```
akshaymagapu: Reopens the attacker-suffix bug Phase 1 was tightened to fix. `cnameMatches` at line ~101 switched Phase 1 from `includes()` to suffix-anchored; Phase 2 here still does substring match, so `www.fastly.net.attacker.com` classifies as `byocdn-fastly`. Reuse `cnameMatches` (or equivalent `=== d || endsWith('.' + d)`).

src/support/cdn-detection.js
```diff
@@ -35,77 +63,672 @@ function catchDnsLookup(err) {
...
+    timeout = 10000,
+    fallbackTimeout = 15000,
+    userAgent = SPACECAT_USER_AGENT,
+    log,
+  } = options;
+
+  const controller = new AbortController();
+  const id = setTimeout(() => controller.abort(), timeout);
+  const fetchOptions = {
+    redirect: 'follow',
```
akshaymagapu, line 680: SSRF: `redirect: 'follow'` with no hop cap, and `detectCdnForDomain` accepts `http://` input at line ~697. Attacker-controlled DNS can publish an A record → `169.254.169.254` (AWS IMDS) or RFC1918, and the probe will follow. Mitigations: (1) reject non-`https` schemes at the entry parser; (2) set `follow: 3` on fetch options; (3) after DNS resolution, block resolved IPs in `127.0.0.0/8`, `169.254.0.0/16`, `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`, `100.64.0.0/10`, `fe80::/10`, `fc00::/7`, `::1` — re-check on each redirect hop.

src/support/cdn-detection.js
```diff
@@ -35,77 +63,672 @@ function catchDnsLookup(err) {
...
+      const message = getError?.message || String(getError);
+      result = { cdn: 'unknown', error: message };
+    }
+  }
+
+  if (result.cdn === 'unknown') {
+    const fallback = await Promise.race([
+      detectCdnFromDnsFallback(url, { log }),
+      new Promise((resolve) => {
+        setTimeout(() => resolve({ cdn: 'unknown' }), fallbackTimeout);
```
akshaymagapu: Timer leak: this `setTimeout` is never cleared when the other racer wins `Promise.race`, so it keeps the event loop alive for up to 15s after `detectCdnFromUrl` returns — delays Lambda freeze and continues billing. Capture the id and `clearTimeout` on win. Also: when the timeout wins, in-flight DoH/ipinfo calls inside `detectCdnFromDnsFallback` keep running and log against an already-returned request. Plumb a parent `AbortSignal` into `detectCdnFromDnsFallback` so the race timeout actually cancels the inner work.

src/support/cdn-detection.js
```diff
@@ -35,77 +63,672 @@ function catchDnsLookup(err) {
...
+    log?.info?.(`[cdn-detection] Detecting CDN for domain ${bareDomain}`);
+
+    const phase1 = await detectAdobeManagedCdn(bareDomain, log);
+    if (phase1 === 'aem-cs-fastly' || phase1 === 'commerce-fastly') {
+      return phase1;
     }
 
-    if (wwwResult === null || bareResult === null) {
-      return null;
+    const phase2Token = await detectGenericCdnToken(url, log);
```
akshaymagapu: Worst-case 40-45s sync latency on `POST /llmo/onboard` (HEAD 10s + GET retry 10s + fallback race 15s + unbounded Phase 1 DNS). Adobe API Gateway default is 30s → user sees 504 while Lambda continues to completion, writing partial state that the retry then rejects via `validateSiteNotOnboarded` (`llmo.js:990`). Before this PR the detector was ~2s. Fix options: (1) async-ify Phase 2 — mirror `triggerBrandProfileAgent` at `llmo.js:1012`, return the onboarding response immediately, write `detectedCdn` via followup; or (2) tighten: HEAD 3s, GET 3s, race 5s, add a 3s `Promise.race` timeout on each Phase 1 DNS call. Also gate Phase 2 behind a feature flag — infra already imported at `llmo-onboarding.js:35` — so there's a kill switch for signature regressions without a revert deploy.

akshaymagapu (dismissed):
Hey @AddyKen-ghost,

Thanks, please check below comments:

- widen spacecat-shared Joi schema (`config.js:348` accepts only `['aem-cs-fastly','other']`) 
- Before merging, manually onboard a real Cloudflare-fronted site and a real Vercel-fronted site in stage (the two unchecked boxes in this PR description). All 94 tests stub the network — we have zero proof that the Lambda VPC can actually reach `ipinfo.io` and `dns.google`. If egress is blocked, Phase 2 silently returns `null` for every non-Adobe site in prod.

nit23uec (changes requested):
## Review Summary

**Verdict: Request changes — needs rework on token-vocabulary consistency and a commitment on the duplication debt before merge.**

The core detection logic is correct and thoroughly tested, but the token vocabulary diverges from LLMO's source of truth, and the "port instead of share" decision creates ~566 lines of guaranteed-to-drift duplication with `spacecat-audit-worker`.

---

## Strengths

- **Backward-compatible contract.** Phase 1 result semantics (`aem-cs-fastly` / `other` / `null`) are preserved; existing callers in `llmo-onboarding.js` and `llmo.js` keep working.
- **Security tightening worth calling out.** Moving CNAME matching from `includes()` to suffix/FQDN-aware `cnameMatches` (`src/support/cdn-detection.js`) closes an attacker-prefix loophole (`evil.cdn.adobeaemcloud.com.attacker.com` used to match). The dedicated regression test is exactly the right coverage.
- **Commerce Fastly signatures are sourced and documented.** Link to the official Commerce Launch Checklist, IPv6 included, all four anycast A-records enumerated.
- **Disciplined test scaffolding.** Phase-1 tests short-circuit cleanly via a no-op `fetchStub` default; 94 passing with 100% branch coverage, and `/* c8 ignore */` comments justify each ignored branch.
- **Strict-TLS on probes.** Using `tracingFetch` from `@adobe/spacecat-shared-utils` and explicitly declining to read `NODE_TLS_REJECT_UNAUTHORIZED` matches the desired security posture.

---

## Critical issues (blocking)

### C1. CDN token vocabulary is inconsistent with the rest of LLMO

`src/controllers/llmo/llmo-utils.js` defines the authoritative enum (described in-file as "aligned with auth-service"):

```
BYOCDN_FASTLY, BYOCDN_AKAMAI, BYOCDN_CLOUDFRONT, BYOCDN_CLOUDFLARE,
BYOCDN_IMPERVA, BYOCDN_OTHER, AMS_CLOUDFRONT, AMS_FRONTDOOR,
AEM_CS_FASTLY, COMMERCE_FASTLY
```

This PR adds 18 new BYOCDN tokens **only in the OpenAPI enum and the detector** — `llmo-utils.js` is explicitly left untouched. Three concrete mismatches:

1. **Azure Front Door & CloudFront.** The PR maps `Azure Front Door` → `byocdn-frontdoor` and all CloudFront hits → `byocdn-cloudfront`, but the canonical tokens in `CDN_TYPES` are `ams-frontdoor` and `ams-cloudfront`. The detector cannot distinguish AMS vs BYOCDN tenancy from DNS/headers alone — that's a product decision — so emitting a BYOCDN token commits every detected site to that classification, disagreeing with the existing taxonomy.
2. **`byocdn-other` is missing.** `CDN_TYPES.BYOCDN_OTHER` exists and has a `CDN_TRANSFORMATIONS` entry in `spacecat-shared/packages/spacecat-shared-utils/src/cdn-helpers.js`. The PR emits bare `'other'` instead. Downstream consumers (log forwarding, UI) that key off `byocdn-other` will silently miss these sites.
3. **16 new tokens aren't in `CDN_TYPES`.** `byocdn-vercel`, `byocdn-netlify`, `byocdn-bunny`, `byocdn-keycdn`, `byocdn-limelight`, `byocdn-cdnetworks`, `byocdn-stackpath`, `byocdn-sucuri`, `byocdn-alibaba`, `byocdn-clever`, `byocdn-airee`, `byocdn-cachefly`, `byocdn-edgecast`, `byocdn-beluga`, `byocdn-myra`, `byocdn-google`, `byocdn-azure` are not in `CDN_TYPES`. A site saved with `siteConfig.llmo.detectedCdn = 'byocdn-vercel'` is a string no other part of the codebase recognizes — log helpers, edge-routing guards, `CDN_TYPES`-keyed lookups all fall through. The PR description frames these as "stored verbatim for analytics," but the OpenAPI enum presents them as first-class values.

**Fix:** Extend `CDN_TYPES` in `llmo-utils.js` in lock-step with the detector, coordinating with auth-service's `cdn-logs-infrastructure/common.js`. Or, if the position is "detector-only stored strings," drop them from the OpenAPI enum and make the contract honest.

### C2. 566-line near-identical port of audit-worker code; commit to extraction

`spacecat-audit-worker/src/detect-cdn/cdn-detector.js` already has the same signature tables, 10-hop DoH loop, `ipinfo.io/${ip}/json` ASN parse, and 17-provider header fingerprint. The PR description defends the duplication as a "product decision" with "drift will be managed manually." That's unworkable for security-sensitive code where header regexes track CDN header changes, ASN tables drift as CDNs reorganize, and a single label typo (`"Cloudflare"` vs `"cloudflare"`) silently corrupts analytics aggregation.

Duplication can be accepted **for this PR only**, on the explicit condition that:

- A follow-up JIRA is filed to move the detector into `spacecat-shared-utils/src/cdn-detection.js`; audit-worker and api-service both consume the shared module; api-service keeps only the `LABEL_TO_LLMO_TOKEN` adapter and the Adobe-managed Phase 1 fast-path.
- Both copies carry a `TODO(shared): track in <JIRA>` marker pointing at that ticket so neither file can be edited without the reader seeing the debt.
- Target extraction within one or two sprints — the longer both copies exist, the more drift accumulates, especially if anyone lands a signature tweak on only one side in the meantime.

### C3. Name shadow between two `detectCdnForDomain` exports

`src/support/cdn-detection.js` and `src/support/edge-routing-utils.js` both export a function called `detectCdnForDomain` with different return contracts. Post-this-PR the shadow is worse: one returns up to 27 string values, the other still returns `aem-cs-fastly | null`. Rename one (e.g. the edge-routing-utils one to `detectAemCsFastlyForDomain`) or consolidate on the new detector and have edge-routing import it.

### M4 → promoted to blocking. Duplicate AEM CS Fastly signatures

`AEM_CS_FASTLY_CNAME_PATTERNS` and `AEM_CS_FASTLY_IPS` exist in both `src/support/cdn-detection.js` and `src/support/edge-routing-utils.js`, and they have **already drifted** (edge-routing-utils lacks `adobe-aem.map.fastly.net`). Once the BYOCDN expansion lands, the mismatch will manifest as "edge-optimize routing succeeds but detection says it's not AEM CS" or vice versa. Consolidate before merging.

---

## Major issues (non-blocking for this PR)

### M1. Latency budget during onboarding

Worst case is roughly 35s on the happy path, but a 10-hop DoH chain adversarially configured can push higher. Detection runs inline inside the onboarding POST. Either (a) impose a hard outer cap on the Phase 2 block, (b) move detection to an async post-onboarding job, or (c) document the latency SLO explicitly.

### M3. No integration test

Per CLAUDE.md, new/modified endpoints need IT coverage under `test/it/`. Onboarding uses the detector but no wiring validates the end-to-end persistence of the new tokens.

### M5. `'other'` vs `null` ambiguity

In the new code, `'other'` can mean "DNS + Phase 2 ran cleanly, no match" OR "HEAD/GET both failed and DNS fallback missed." The second case is effectively inconclusive but returns `'other'`, not `null`. Either propagate probe failures into `null`, or rename `'other'` → `'unknown'` to stop pretending we had signal.

---

## Minor issues

- `detectCdnForDomain(input, log)` — `input` now accepts URL or hostname; function name still says "ForDomain." Rename or document.
- `log?.info?.` vs `log?.info` — mixed optional-chain styles across the file. Pick one.
- `CDN_KEYWORD_SIGNATURES` ordering is non-obvious (`'clever-cloud'` before `'cloudflare'` is correct; `'limelight'` before `'fastly'` is worth a comment).
- `detectCdnFromHeaders` returns `"unknown"` as a sentinel string — mixes two vocabularies. A typed result (`{ cdn: string | null }`) would be cleaner.
- `getCnameChain` and `getCnameChainDoh` differ only in the resolver primitive — a `walkCnames(resolver, hostname)` helper would remove ~30 lines.
- DoH is hard-coded to Google (`dns.google`) with no Cloudflare `1.1.1.1` fallback. One provider = one outage from the path being dead.
- `matchCdnByKeywords` substring-matches a concatenated header blob — `server: "Apache/2.4 (limelight-build)"` would misclassify. Port has the same bug; worth fixing at the shared-extraction step if not here.
- OpenAPI enum lists `null` as a string alongside `type: [string, 'null']`. Drop from the enum.

---

## Remediation plan

**Blocking (address in this PR):**
1. **C1** — align detector tokens with `CDN_TYPES` in `llmo-utils.js` / auth-service.
2. **C2** — file the shared-extraction follow-up JIRA; add `TODO(shared):` markers in both copies pointing at the ticket.
3. **C3** — rename one of the two `detectCdnForDomain` functions.
4. **M4** — single source for AEM CS Fastly CNAME / IP signatures.

**Post-merge follow-up (the C2 ticket):**
- Extract detector + signature tables to `spacecat-shared-utils/src/cdn-detection.js`.
- audit-worker and api-service both re-export from the shared module; delete duplicated tables from both services.
- Target within one or two sprints.

**Non-blocking:** M1 (latency cap / async), M3 (integration test), M5 (`other` vs `unknown` semantics), minors.

src/support/cdn-detection.js
AddyKen-ghost: Resolved by aligning the schema rather than masking drift in the detector. @adobe/spacecat-shared-data-access@3.53.0 (adobe/spacecat-shared#1559) widens siteConfig.llmo.detectedCdn to the detector's full emit set:

['aem-cs-fastly','commerce-fastly','byocdn-fastly','byocdn-akamai',
 'byocdn-cloudflare','byocdn-imperva','byocdn-other','other']
This PR bumps the dependency to 3.53.0 and un-skips the round-trip block in test/controllers/llmo/llmo-onboarding.test.js — every emitted token now goes through validateConfiguration as an active CI gate, plus a negative test that an unknown token (byocdn-cloudfront, deliberately outside the emit set) is rejected.

Intentionally not adding a runtime "downgrade to 'other' if shared doesn't recognise it" guard: that silently masks contract drift the round-trip test is designed to surface. If the detector grows a token without a matching schema bump, unit tests fail before the next config write hits prod.

src/support/cdn-detection.js
AddyKen-ghost: Fixed in 5fbf6689. matchCdnByCname now uses suffix-anchored matching (lower === d || lower.endsWith('.' + d)), identical to cnameMatches in edge-routing-utils.js. Added a Phase-2 regression in test/support/cdn-detection.test.js (edge.azureedge.net.attacker.example → null) so a future revert can't silently reintroduce the substring path. Same fix mirrored in edge-routing-utils.js's local helper for parity.

src/support/cdn-detection.js
AddyKen-ghost: Partial mitigations landed here; rest is filed for follow-up.

In this PR (`5fbf6689`):
- `detectCdnForDomain` rejects any input with a non-`https` scheme; bare hostnames are normalised to `https://` before any probe runs.
- Every outbound `fetch` (HEAD probe, GET fallback probe, ipinfo ASN lookup) carries `redirect: 'follow', follow: 5`, capping redirect chains.

Deferred to **https://jira.corp.adobe.com/browse/LLMO-4564 — Harden CDN detector outbound fetch against SSRF (resolved-IP blocklist + AbortSignal)**:
- Resolved-IP blocklist for `127.0.0.0/8`, `169.254.0.0/16`, `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`, `100.64.0.0/10`, `fe80::/10`, `fc00::/7`, `::1`, re-checked on every redirect hop.
- Same blocklist applied to ipinfo's response IP before trusting the ASN/org string.

The blocklist needs lookup-aware fetch options (or a custom `Agent`) that `@adobe/fetch` doesn't surface directly — it's a meaningful refactor rather than a one-line guard, so splitting it keeps this PR scoped to the auto-select correctness fix. The new JIRA captures the full hardening alongside the AbortSignal plumbing in your other comment, since both touch the same outbound-probe machinery.

src/support/cdn-detection.js
AddyKen-ghost: Timer leak fixed in `5fbf6689`: `detectCdnFromUrl`'s `Promise.race` now captures the `setTimeout` id and `clearTimeout`s it whenever the racer wins, so the event loop drains immediately on completion. Every probe inside the race has its own bounded timeout (HEAD 5s, GET 5s, fallback 10s, DoH 3s with the Cloudflare fallback also at 3s, ASN 3s, per-Phase-1 DNS lookup 3s) — no silent unbounded waits.

Plumbing a parent `AbortSignal` end-to-end through `detectCdnFromDnsFallback` (so the race timeout actually cancels nested DoH/ASN/PTR work) is the right next step but every `fetch`/`dns` site has to honour the same signal — captured in **LLMO-4564** alongside the SSRF blocklist since they share the same outbound-fetch refactor.

src/support/cdn-detection.js
AddyKen-ghost: Latency tightened in `5fbf6689`. New worst case ≈ 24s (HEAD 5s + GET 5s + fallback 10s with inner DoH/ASN bounded at 3s + Phase 1 DNS 3s), down from ~45s and comfortably inside the 30s API Gateway ceiling. All numbers are runtime-tunable via the constants at the top of `cdn-detection.js` if we need to come down further on hot paths.

Holding off on async-ifying / a dedicated Phase-2 feature flag in this PR:

1. Async-ifying changes the observable contract of `POST /llmo/onboard` — the response goes from "here's your auto-select hint" to "hint follows via a separate channel" — which adobe/project-elmo-ui#1569 is actively consuming for the radio auto-select. Splitting that out blocks the auto-select feature on a contract migration.
2. The detector is already gated by the same onboarding rollout flag that fires `triggerBrandProfileAgent` (`llmo-onboarding.js:35` call site), so a problematic signature change can be killed per-tenant without redeploy. A Phase-2-scoped flag is something we can layer on reactively if the shared flag proves too coarse.

If post-merge p99s show the 24s ceiling biting real traffic, async-ification + a Phase-2-only flag is the natural follow-up — happy to file with concrete latency data at that point rather than speculatively now.

Separately: also filed **LLMO-4545 — Extract CDN detector and signature tables into spacecat-shared** so api-service and audit-worker stop carrying parallel signature lists. Marked with a `// TODO(LLMO-4545)` at the top of the Phase 2 section in `cdn-detection.js`.

AddyKen-ghost:
Hi @akshaymagapu, 

I have widened the schema, for testing the ipinfo and google dns will do quick checks today before merging thanks!

AddyKen-ghost:
Hi @nit23uec, 

Thanks for the detailed review — most of the structural concerns landed in the rework. Going through your blockers and the major/minor list:

---

## C1 — Token vocabulary inconsistency

Realigned, but the resolution is the inverse of "expand detector to match `CDN_TYPES`": we collapsed the detector emit set to only the tokens we can emit confidently from network signals, and treat everything else as `byocdn-other`. New emit set is exactly:

```
'aem-cs-fastly' | 'commerce-fastly' | 'byocdn-fastly' | 'byocdn-akamai'
  | 'byocdn-cloudflare' | 'byocdn-imperva' | 'byocdn-other' | 'other' | null
```

Mirrored in three places now (single source: detector → schema → UI):
- Detector emit set (`src/support/cdn-detection.js#LABEL_TO_LLMO_TOKEN`)
- OpenAPI enum (`docs/openapi/llmo-api.yaml`)
- Joi schema (`@adobe/spacecat-shared-data-access@3.53.0`)
- UI `DetectedCdn` type (adobe/project-elmo-ui#1569)

Concrete responses to your three mismatches:

1. **CloudFront / Azure Front Door.** You're right that emitting `byocdn-cloudfront` / `byocdn-frontdoor` was wrong — the detector cannot disambiguate AMS vs BYOCDN tenancy from network signals. Both now collapse to `byocdn-other` (`LABEL_TO_LLMO_TOKEN`). `byocdn-cloudfront` is deliberately removed from the emit set entirely; once a future detector revision ships AMS-aware CloudFront/Azure signatures we can split them back out, and the schema/enum/UI move together.

2. **`byocdn-other` missing.** Added — both as a detector-emit token and (since it's covered by `CDN_TYPES.BYOCDN_OTHER` already) it now flows into the existing `cdn-helpers.js#CDN_TRANSFORMATIONS` path without further work.

3. **16 stray BYOCDN tokens.** Dropped from the OpenAPI enum and the detector. Every label that used to map to `byocdn-vercel` / `byocdn-netlify` / `byocdn-bunny` / `byocdn-google` / `byocdn-azure` / etc. now maps to `byocdn-other` (`LABEL_TO_LLMO_TOKEN`). The contract is honest now: the detector emits 8 tokens + `null`, and that's the entire enum.

Note that this is the *detector* vocabulary — it's intentionally a subset of `llmo-utils.js#CDN_TYPES`, which models the full set of CDN providers a user can pick in the radio. The detector tells the UI "this is what we could identify from probes", and the UI uses that to drive auto-select where unambiguous (`aem-cs-fastly`, `commerce-fastly`, four byocdn-*). Where we can't disambiguate (`byocdn-other`, `other`, `null`), the UI leaves all radios neutral and the user picks. This is enforced by `AUTO_SELECT_DETECTED_CDNS` in `CDNOnboardDialog.tsx`.

---

## C2 — 566-line duplication; commit to extraction

Filed **LLMO-4545 — Extract CDN detector and signature tables into spacecat-shared**. Added `// TODO(LLMO-4545)` marker at the top of the Phase 2 section in `cdn-detection.js` so the reader can't miss the debt. Audit-worker carries the same marker as part of that ticket's scope (we explicitly didn't touch audit-worker in this PR, by agreement upstream). Target one to two sprints.

---

## C3 — Name shadow between two `detectCdnForDomain` exports

Renamed `edge-routing-utils.js#detectCdnForDomain` → `detectAemCsFastlyForDomain` to match its scope (AEM-CS-Fastly-only DNS check). Callers in `src/controllers/llmo/llmo.js` and the test file (`test/support/edge-routing-utils.test.js`) updated. The broader detector keeps the `detectCdnForDomain` name in `cdn-detection.js`.

---

## M4 (promoted) — Duplicate AEM CS Fastly signatures

Consolidated. `AEM_CS_FASTLY_CNAME_PATTERNS` and `AEM_CS_FASTLY_IPS` are now exported from `src/support/cdn-detection.js`; `edge-routing-utils.js` imports them. Single source — `adobe-aem.map.fastly.net` and the rest of the patterns are now in lock-step across the two callers, and any future signature add lands in one place.

---

## M1 — Latency budget

Tightened all timeouts inline rather than async-ifying:

- HEAD probe: 5s (was 10s)
- GET fallback: 5s (was 10s)
- Outer fallback race: 10s (was 15s)
- DoH per provider: 3s
- ASN lookup: 3s
- Per Phase-1 DNS lookup: 3s (was unbounded)

New worst-case ≈ 24s vs the original ~45s, comfortably under the 30s API Gateway ceiling. All numbers are runtime-tunable via constants at the top of `cdn-detection.js`.

Skipping async-ification for this PR because it would change the response contract of `POST /llmo/onboard` — the UI radio auto-select in adobe/project-elmo-ui#1569 consumes `detectedCdn` from the synchronous response. Splitting Phase 2 to async would block that feature on a contract migration. If post-merge p99s show the 24s ceiling biting real traffic I'll file an async-ification + Phase-2-only feature flag follow-up with concrete latency data.

---

## M3 — No integration test

Acknowledged. Detection is mock-heavy in this PR by design: real DNS / ipinfo / DoH calls are flaky in CI and gated by Lambda VPC egress (which is exactly what akshay's manual-stage-onboarding ask in his review verifies end-to-end). The cleanest home for a true IT lives on the shared module after **LLMO-4545** — that PR will add `test/it/cdn-detection.test.js` against a small set of live targets so both api-service and audit-worker share the IT.

---

## M5 — `'other'` vs `null` ambiguity

Disambiguated. Threaded a `probeSucceeded` flag through `detectCdnFromUrl` and `detectGenericCdnToken`:

- `'other'` now strictly means "Phase 1 + Phase 2 ran, at least one probe produced data, no signature matched" (clean miss).
- `null` means "at least one probe was inconclusive (HTTP failed AND DoH/ASN/PTR all yielded nothing)" — auto-select stays neutral.

Test expectations updated (`returns null when domain does not exist (ENOTFOUND) for all lookups`), OpenAPI description rewritten to spell out the four states. Decided to keep the token name `'other'` rather than `'unknown'` because (a) it matches the existing `CDN_TYPES.OTHER` token elsewhere in LLMO and (b) renaming it would be a breaking enum change.

---

## Minors

- **`ForDomain` naming.** Function still does logical "domain" work; the URL-vs-hostname normalisation is one line at the top. JSDoc clarifies. Worth revisiting at the C2 extraction.
- **`log?.info?` vs `log?.info`.** Normalised to `log?.info?.` everywhere.
- **`CDN_KEYWORD_SIGNATURES` ordering.** Added a comment block explaining the longest-prefix-wins ordering (`clever-cloud` before `cloud`, `limelight` before `light`, etc.) so the order isn't accidentally re-sorted by a future contributor.
- **`detectCdnFromHeaders` `"unknown"` sentinel.** Agreed it's ugly; the typed-result refactor (`{ cdn: string | null, probeSucceeded: boolean }`) is the natural next step at the C2 extraction since it touches every call site.
- **`getCnameChain` / `getCnameChainDoh` duplication.** `walkCnames(resolver, hostname)` belongs in the shared module — captured in LLMO-4545.
- **DoH single-provider.** Added Cloudflare `https://cloudflare-dns.com/dns-query` as a fallback (3s timeout each). One outage from path-dead is fixed; both providers down is still the failure mode but the log now reads `[cdn-detection] DoH query failed (both providers)` so it's diagnostic.
- **`matchCdnByKeywords` substring header match.** Real bug — `server: "Apache/2.4 (limelight-build)"` would misclassify. Inheriting from the audit-worker port; tracked under LLMO-4545's hardening scope rather than touching it here, since fixing it cleanly means moving to a tokenised-header check that's the same effort either way.
- **OpenAPI `null` listed as a string alongside `type: [string, 'null']`.** Cleaned up — bare `'null'` string entry removed from the enum; `type: [string, 'null']` is the only nullability declaration now.

---

In summary, every blocker is addressed in this PR (`5fbf6689` + the dep bump in `50ecb81d`), the duplication-debt items have a tracked extraction (LLMO-4545), and the latency / IT items have explicit follow-up plans gated on real-traffic data and the shared module respectively. Re-requesting review.

docs/openapi/llmo-api.yaml
```diff
@@ -916,8 +916,35 @@ llmo-onboard:
                   type:
                     - string
                     - 'null'
-                  enum: [aem-cs-fastly, other, null]
-                  description: CDN provider detected via DNS during onboarding. null when detection is inconclusive.
+                  enum:
+                    - aem-cs-fastly
```
nit23uec, line 920: we should align them with https://github.com/adobe/spacecat-api-service/blob/e4e7791df1a26a6dddaa3ead1c618677f3ca899a/src/controllers/llmo/llmo-utils.js#L17

src/support/cdn-detection.js
```diff
@@ -35,77 +65,792 @@ function catchDnsLookup(err) {
...
+    return null;
+  }
+  if (ipv6.length > 0) {
+    log?.info?.(`[cdn-detection] Detected IPv6 for domain ${host}: ${ipv6}`);
+  }
+  if (ipv6.some((ip) => COMMERCE_FASTLY_IPV6.has(ip))) {
+    return 'commerce-fastly';
+  }
+
+  return 'other';
```
nit23uec: it should be byocdn-other

docs/openapi/llmo-api.yaml
```diff
@@ -916,8 +916,35 @@ llmo-onboard:
...
+                        BYOCDN matches from the multi-signal probe (HTTP
+                        headers, CNAME chain, ASN, PTR).
+                      - `byocdn-other` — catch-all for any other detected
+                        third-party CDN (CloudFront, Azure Front Door / Azure
+                        CDN, Vercel, Netlify, Bunny, Sucuri, Alibaba, etc.).
+                        Used for CloudFront/Azure today because the detector
+                        cannot disambiguate AMS vs BYOCDN tenancy from
+                        network signals; a future detector revision can split
+                        these out once AMS-aware signatures land.
+                      - `other` — both detection phases ran cleanly and no
```
nit23uec: we can remove `other`. byocdn-other is the catch all.

docs/openapi/llmo-api.yaml
AddyKen-ghost: Aligned in `ba466569` (OpenAPI + detector) and `56f25b36` (UI). The `detectedCdn` enum is now exactly the 10-token `CDN_TYPES` set:

```
aem-cs-fastly, commerce-fastly, byocdn-fastly, byocdn-akamai,
byocdn-cloudfront, byocdn-cloudflare, byocdn-imperva, byocdn-other,
ams-cloudfront, ams-frontdoor
```

(plus `null` for inconclusive probes). Contract now agrees across four layers:

- **Detector** (`src/support/cdn-detection.js`) emits a 7-token subset today.
- **OpenAPI enum** (this file) — full 10 tokens + `null`.
- **Joi schema** (`@adobe/spacecat-shared-data-access@3.54.0`, adobe/spacecat-shared#1563, merged) — full 10 tokens + legacy `'other'` + `null`.
- **UI type** (`project-elmo-ui/src/types/site.ts` in adobe/project-elmo-ui#1569) — aliased to `CDNLogSource | null`, same 10 tokens.

The three CDN_TYPES tokens the detector cannot emit today (`byocdn-cloudfront`, `ams-cloudfront`, `ams-frontdoor`) are now pre-wired in all four layers: OpenAPI lists them, Joi accepts them, the UI type and `AUTO_SELECT_DETECTED_CDNS` handle them. Once detector revisions add AMS-aware signatures, they flip from "reserved" to "emitted" with no coupled schema / type / enum change needed. Description rewritten to spell this out.

src/support/cdn-detection.js
AddyKen-ghost: Done in `ba466569`. All three `return 'other'` sites in the detector flipped to `'byocdn-other'`:

- `checkHost` (line 148 — the Phase-1 fall-through after CNAME / A / AAAA lookups succeed but don't match AMS)
- `detectAdobeManagedCdn` (line 170 — aggregation over `www.{domain}` and `{domain}`)
- terminal path in `detectCdnForDomain` (line 845 — both phases ran cleanly, no specific match)

JSDoc updated to reflect the new sentinel; description reads "the result is always a `CDN_TYPES` value; `byocdn-other` is the sole catch-all." 19 test expectations in `test/support/cdn-detection.test.js` flipped to match. `cdn-detection.js` still reports 100/100/100/100 coverage.

`'other'` no longer appears anywhere in the detector source or test expectations. It remains in the Joi schema in `3.54.0` purely as a legacy sentinel — so any record from the original Phase-1-only detector (pre-this-PR) still validates on its next config write — but nothing in the api-service emits it now.

docs/openapi/llmo-api.yaml
AddyKen-ghost: Removed in `ba466569`. The `'other'` entry is gone from the enum; `byocdn-other` is now the sole catch-all.

The enum description also rewritten to make the semantics explicit:

> `byocdn-other` — catch-all. Emitted when Phase 1 + Phase 2 both run cleanly and no specific signature matches, OR when the probe identifies a third-party CDN (CloudFront, Azure Front Door / Azure CDN, Vercel, Netlify, Bunny, Sucuri, Alibaba, etc.) that LLMO does not model as a first-class BYOCDN. CloudFront and Azure Front Door land here today because the detector cannot disambiguate AMS vs BYO tenancy from network signals alone.

UI (`project-elmo-ui/src/components/customer-configuration/CDNOnboardDialog.tsx` via adobe/project-elmo-ui#1569) gates auto-select on `AUTO_SELECT_DETECTED_CDNS`, which includes every CDN_TYPES token *except* `byocdn-other` — so this value leaves the radios neutral and surfaces a "could not auto-detect" hint, which is the correct UX for a catch-all.

nit23uec (approved):

PR #2246, opened 2026-04-21
fix: use www-resolved domain for RUM AEM check in PLG onboarding

Kanishkavijay39:
## Summary

The PLG AEM verification step was calling retrieveDomainkey(domain) with the bare domain only. If a customer's RUM data is keyed on www.domain.com (or vice versa), the lookup failed and the domain got waitlisted as "not an AEM site" even though it is a valid AEM site.
Wired in resolveWwwUrl (already used in autoResolveAuthorUrl and SHM) into the AEM verification step so both domain.com and www.domain.com are tried against RUM before falling back to delivery type detection.
When no site record exists yet, a proxy object satisfying the getBaseURL() interface is used — safe because overrideBaseURL only applies to existing sites.

## Related Issues

Thanks for contributing!


src/controllers/plg/plg-onboarding.js
```diff
@@ -755,7 +756,9 @@ async function performAsoPlgOnboarding({
     const rumApiClient = RUMAPIClient.createFrom(context);
     let cachedDeliveryType = null;
     try {
-      await rumApiClient.retrieveDomainkey(domain);
+      const siteProxy = site ?? { getBaseURL: () => baseURL };
```
prithipalpatwal: @Kanishkavijay39 Why do we need a siteProxy? What does it mean if we don’t have a site in this flow?

Kanishkavijay39: No, we don't have site till now.. Will check the domain type is aem_cs or eds  from the rum result..
Then will proceed further..

prithipalpatwal: Then why are we even making a GET call for the site at line 755?

`let site = await Site.findByBaseURL(baseURL);`

Kanishkavijay39: If exists.. Like there can be possiblity that domain customer is trying to onboard is already there in system.
But we will not have this always. In some cases we will have in some not. So if we would have then use site object otherwise proxy by adding base url

prithipalpatwal (approved):

PR #2250, opened 2026-04-22
fix(brands): reject brand URLs that don't match a site in the org | LLMO-4435

jjenscodee:
## Summary
Fixes LLMO-4435 (Brandalf silent save failure). Previously, submitting a brand URL whose normalized base did not match a `sites` row in the brand's IMS organization was silently dropped by `syncBrandSites`: the PATCH returned 200 with `urls: []`, leaving the UI to show a misleading "saved" state. Users reported adding `yahoo.com` (and similar) to Adobe Acrobat and never seeing it appear — with no error surfaced.

### What changes
- Extract resolution into `resolveBrandSiteUrls` in `src/support/brands-storage.js`. It groups submitted URLs by normalized base URL, queries `sites` scoped to the org, and if any base URL can't be resolved, throws an `Error` with `status: 400`, `code: 'BRAND_URLS_UNRESOLVED'`, and a message listing every unresolved URL.
- Invoke the resolver at the top of both `upsertBrand` (create) and `updateBrand` (edit), BEFORE any `brands` or child-table writes — so an invalid request leaves all brand data untouched. The controller's existing `try/catch → createErrorResponse` path propagates the thrown error to a 400 response.
- `syncBrandSites` now takes the pre-resolved `{ sites, pathsByBase, typeByBase }` struct instead of raw URLs, removing the silent early-return behaviour on unresolved sites.

### Tests
- New: `throws BRAND_URLS_UNRESOLVED when a submitted URL has no matching site in the org` — verifies `status`, `code`, error message content, and that neither `brands` nor `brand_sites` were touched.
- New: `lists every unresolved URL in the error message when multiple are missing` — pluralization / partial-mismatch case.
- New: `throws when the sites resolution query fails` — covers the resolver's own error path.
- New: `throws BRAND_URLS_UNRESOLVED without touching the brand when a submitted URL has no matching site` (`updateBrand` mirror).
- New: `treats null urls the same as an empty array (no validation, brand_sites cleared)` — edge case.
- New: `still allows other field updates when updates.urls is not provided` — confirms non-URL patches don't trigger a needless sites query.
- Removed: the two obsolete tests that encoded the silent-drop behaviour (`skips syncBrandSites when urls resolve to no matching sites` and `uses empty paths array when site base_url is not in pathsByBase map`).
- Updated: `falls back to base URL when URL string is invalid in syncBrandSites` — mock `sites.base_url` aligned with `composeBaseURL` normalization (`not-a-valid-url` → `https://not-a-valid-url`).

Full suite: 8313 passing. Coverage: 100% / 100% branches.

## Jira
https://jira.corp.adobe.com/browse/LLMO-4435

## How to Test
1. In stage, pick any brand in an IMS org where `yahoo.com` is not registered as a `sites` row. Open the brand's URLs tab in Brandalf.
2. Add URL `https://yahoo.com` → Save.
3. Expected: Network tab shows PATCH `/v2/orgs/:orgId/brands/:brandId` returning **400** with a body like `The following URL(s) could not be saved because no matching site exists in this organization: https://yahoo.com`. The UI shows a negative toast surfacing that message. Re-opening the brand shows its URL list **unchanged** from before the save.
4. Happy path: add a URL that IS a registered site → 200, success toast, URL appears. No regression.
5. Edge case: submit a mix of valid + invalid URLs in one save. Expected: 400 listing only the invalid ones; the valid URLs are NOT persisted (the whole request is rejected — no partial writes).

## Rollback
Additive pre-validation. Reverting the commit restores the current silent-drop behavior. Low risk, no schema changes.

## Related
Replaces PR [project-elmo-ui#1580](https://github.com/adobe/project-elmo-ui/pull/1580) which attempted a UI-layer workaround; that approach was rejected in favor of this backend fix.

### DIFF
--- a/src/support/brands-storage.js
+++ b/src/support/brands-storage.js
@@ -128,10 +128,68 @@ async function replaceChildRows(table, brandId, rows, onConflict, postgrestClien
 }
 
 /**
- * Fully replaces brand_sites for a brand. Groups submitted URLs by normalized base URL
- * (via composeBaseURL) so that multiple paths under the same site share one brand_sites row.
+ * Groups submitted URLs by normalized base URL (via composeBaseURL) and resolves each
+ * base against the sites table in the brand's organization.
+ *
+ * Throws a 400 BRAND_URLS_UNRESOLVED error if any submitted base URL has no matching
+ * `sites` row. Callers invoke this BEFORE any destructive brand writes so invalid
+ * requests leave the brand untouched. See LLMO-4435.
+ *
+ * Returns the matched `sites` rows plus the `pathsByBase`/`typeByBase` maps so
+ * `syncBrandSites` can upsert without re-querying.
  */
-async function syncBrandSites(organizationId, brandId, urls, postgrestClient, updatedBy) {
+async function resolveBrandSiteUrls(organizationId, urls, postgrestClient) {
+  const pathsByBase = new Map();
+  const typeByBase = new Map();
+  (urls || []).forEach((u) => {
+    const value = typeof u === 'string' ? u : u?.value;
+    if (!hasText(value)) {
+      return;
+    }
+    const { base, path } = parseUrlParts(value);
+    const normalizedBase = composeBaseURL(base);
+    if (!pathsByBase.has(normalizedBase)) {
+      pathsByBase.set(normalizedBase, []);
+    }
+    pathsByBase.get(normalizedBase).push(path || '/');
+    // First URL with a type wins for a given base URL — prevents silent overwrite
+    // when multiple paths under the same domain carry different types.
+    if (typeof u === 'object' && hasText(u?.type) && !typeByBase.has(normalizedBase)) {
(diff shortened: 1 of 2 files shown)


jjenscodee:
Closing — the LLMO-4435 requirement has inverted. We're no longer rejecting unresolved brand URLs with a 400; instead we'll persist every submitted URL and annotate its onboarded status in the response.

Replacement work is in progress:
- Schema: adobe/mysticat-data-service#427 (new \`brand_urls\` child table)
- API: a fresh PR on branch \`llmo-4435-accept-all-brand-urls\` (coming next) — removes the rejection logic here, adds \`syncBrandUrls\`, unions \`brand_urls\` + \`brand_sites\` in the V2 response with \`onboarded\` / \`siteId\` flags.

Deleting this branch as part of cleanup.

jjenscodee:
Replacement PR now open: https://github.com/adobe/spacecat-api-service/pull/2252

PR #2252, opened 2026-04-22
feat: persist all brand URLs and annotate onboarded status | LLMO-4435

jjenscodee:
## Summary

Customers can now submit any URL to a brand — including URLs whose base does not resolve to a site registered with the org. Every submitted URL is persisted to a new \`brand_urls\` child table. The V2 brand response is extended with \`onboarded\` and \`siteId\` fields per URL so the UI can distinguish URLs that feed CDN / Search Console configuration from ones stored but external.

## Changes

- **\`src/support/brands-storage.js\`**
  - New \`syncBrandUrls\` mirroring \`syncSocialAccounts\` / \`syncAliases\` (full-replace child-table sync into \`brand_urls\`, normalizing via \`composeBaseURL\`).
  - Called alongside \`syncBrandSites\` on both \`upsertBrand\` and \`updateBrand\`.
  - \`mapDbBrandToV2\` unions \`brand_urls\` (raw user submissions) with \`brand_sites\` (sites join): each response URL carries \`onboarded\`; onboarded entries also carry \`siteId\`. Legacy brands with no \`brand_urls\` rows fall back to the \`brand_sites\` expansion (all implicitly onboarded).
- **\`docs/openapi/schemas.yaml\`** — \`V2BrandUrl\` documents \`onboarded\`, \`siteId\`.
- **Tests** — keep every existing happy-path; add coverage for persisted-unresolved, mixed onboarded/external, and the new sync error path. Branch coverage: 100%.

## Dependency

Depends on the new \`brand_urls\` table in adobe/mysticat-data-service#427 — merge/apply that migration before this.

## Test plan

- [ ] Unit suite passes with 100% branch coverage (verified locally — 8315 passing)
- [ ] Stage: PATCH a brand with \`{value: 'https://yahoo.com'}\` (not an org site) → 200, response carries \`onboarded: false\`, row in \`brand_urls\`, no row in \`brand_sites\`
- [ ] Stage: PATCH with a URL whose base IS a site → 200, response carries \`onboarded: true\` + \`siteId\`, rows in both tables
- [ ] Stage: PATCH with a mix → both appear in the response, each flagged correctly


calvarezg (commented):
## Code Review

### [Major] Docker image `v1.67.8` predates the `brand_urls` migration — integration tests will fail

`test/it/postgres/docker-compose.yml` is bumped to `v1.67.8`, but the `brand_urls` migration was just merged to `mysticat-data-service` main today. The current release there is `v1.68.0` (which doesn't include `brand_urls` either — that migration landed after `1.68.0`). The next release will be `v1.69.0` or similar. Until this is updated to a version that actually contains the `brand_urls` table, any integration test that touches `brand_urls` will fail with `relation does not exist`. This image reference needs a second update once the migration image is published.

### [Minor] Parameter order of `syncBrandUrls` is reversed from `syncBrandSites`

```js
// syncBrandSites signature:
async function syncBrandSites(organizationId, brandId, urls, ...)

// syncBrandUrls signature:
async function syncBrandUrls(brandId, organizationId, urls, ...)
```

Not a bug — both call sites pass the right values — but it's an inconsistency that's a future footgun for anyone adding a third call site or refactoring. Worth flipping `syncBrandUrls` to match the `(organizationId, brandId, ...)` convention.

---

### Looks good

- `mapDbBrandToV2` logic is correct: builds `siteByBase` map from `brand_sites`, annotates `brand_urls` entries with `onboarded`/`siteId`, propagates `brand_sites.type` for legacy read compatibility, and falls back to `brandSitesUrls` only when `brand_urls` is empty.
- Legacy fallback edge case is handled correctly: `urls: []` clears both tables, both expansion arrays are empty, no phantom entries surface.
- `syncBrandUrls` deduplication is solid: the `seen` Set prevents duplicates before the upsert, consistent with how `syncSocialAccounts` / `syncAliases` work.
- `Promise.all` for parallel sync is consistent with the existing pattern for other child-table syncs and is an acceptable trade-off given PostgREST's lack of cross-table transactions.
- Test coverage is thorough: unresolved URL path, mixed onboarded/external, `type` propagation, error path, all updated happy-path assertions.
- OpenAPI schema additions are well-documented with clear `Response-only` / `Ignored on input` callouts.

calvarezg (commented):
Parameter order on `syncBrandUrls` is fixed — matches `(organizationId, brandId, ...)` convention now. Thanks.

The docker image in `test/it/postgres/docker-compose.yml` is still at `v1.67.8`, which predates the `brand_urls` migration. This is likely blocked on the mysticat-data-service release pipeline — once the new image is published (should be `v1.69.0` given the `feat:` bump from the migration merge today), this needs one more update before the integration tests can pass. Everything else looks good to merge once that image reference is current.

calvarezg (approved):

PR #2254, opened 2026-04-22
feat: remove answers from brand presence topic details api

vivesing:
### Goal

For **`GET .../brand-presence/topics/:topicId/detail` only**, stop loading the LLM **`answer`** from PostgREST and **omit the `answer` field** from each object in **`executions[]`** (no optional flag to bring it back).

related to bug/ 413 response code https://jira.corp.adobe.com/browse/LLMO-4319 

### Backend (`spacecat-api-service`)

| Area | Change |
|------|--------|
| **PostgREST select** | Introduced **`TOPIC_DETAIL_SELECT`**: same columns as **`DETAIL_SELECT`** except **`answer`**. Topic detail uses this list so the DB/API path does not fetch answer text for this route. |
| **Query builder** | **`buildDetailExecQuery`** now accepts an optional 6th argument **`selectColumns`** (default **`DETAIL_SELECT`**). **`createTopicDetailHandler`** passes **`TOPIC_DETAIL_SELECT`**; **`createPromptDetailHandler`** unchanged (still **`DETAIL_SELECT`**). |
| **Response mapping** | Added **`mapBrandPresenceTopicDetailExecutionRow`** — same execution shape as before **without** an **`answer`** key. Topic detail maps with this; prompt detail keeps **`mapBrandPresenceDetailExecutionRow`** (still includes **`answer`**). |

### Documentation

- **`docs/llmo-brandalf-apis/topic-detail-api.md`**: Removed **`answer`** from the sample payload and field table; noted answer is not on this endpoint; updated aggregation description to reference **`TOPIC_DETAIL_SELECT`**.

### Tests

- **`test/controllers/llmo/llmo-brand-presence.test.js`**: Assertions that topic-detail **`executions[]`** items **do not** have **`answer`**; new test that the executions **`select`** string includes **`execution_date`** and does **not** include **`answer`**.

### Consumer impact

- **Breaking** for any client that still reads **`executions[].answer`** from **topic detail** only; property is **absent** (not empty string).
- **Prompt detail** and other routes: **unchanged** for this work.
- Smaller payloads for topic detail when answers were large.


calvarezg:
Two issues worth addressing before merge:

**1. Mapper function duplication**

`mapBrandPresenceTopicDetailExecutionRow` is a near-exact copy of `mapBrandPresenceDetailExecutionRow` with only the `answer` line dropped. If a field is added to the shared mapper in the future it won't automatically appear in the topic version — silent drift risk.

Consider composing instead:

```js
function mapBrandPresenceDetailExecutionRow(r, { includeAnswer = true } = {}) {
  const base = { /* all shared fields */ };
  if (includeAnswer) base.answer = r.answer || '';
  return base;
}

function mapBrandPresenceTopicDetailExecutionRow(r) {
  return mapBrandPresenceDetailExecutionRow(r, { includeAnswer: false });
}
```

Or extract a `mapBaseExecutionFields` helper and have both mappers call it.

---

**2. `TOPIC_DETAIL_SELECT` is a hand-edited copy of `DETAIL_SELECT`**

Both are long `max-len`-suppressed strings. A new column added to `DETAIL_SELECT` won't automatically appear in `TOPIC_DETAIL_SELECT`. Deriving it removes the drift risk:

```js
const TOPIC_DETAIL_SELECT = DETAIL_SELECT
  .split(', ')
  .filter((col) => col !== 'answer')
  .join(', ');
```

calvarezg:
Both issues addressed — thanks for the quick turnaround. LGTM 👍

calvarezg (approved):

PR #2256, opened 2026-04-22
feat: notify  LLMO team when a customer opted in for optimise at edge

khushboo5723:
**Summary**

Fires an internal notification email to the LLMO team when a customer opts into edge optimization for the first time (`isNewlyOpted = true`).

Email is written in customer-facing tone so it can be forwarded directly to the customer. Includes CDN-specific documentation link and guidance note per CDN type. Includes comma-separated list of org members in the email body. Email is fire-and-forget — never blocks the opt-in API response.

**Changes**

- `src/controllers/llmo/cdn-opt-in-notification.js` (new) — notification helper:
  - Maps `cdnProvider` → display name, doc link, CDN-specific note
  - Reads CDN type from `llmo.cdnBucketConfig.cdnProvider` (populated by `llmo-config-wrapper` in auth-service during provisioning)
  - Fetches org members via `TrialUser.allByOrganizationId`, filters blocked/deleted, deduplicates
  - Validates recipients from `OPT_IN_NOTIFICATION_RECIPIENTS` env (only `@adobe.com` addresses allowed)
  - Sends email via `llmo_cdn_opt_in_notification` CMC template
- `src/controllers/llmo/llmo.js` — calls `notifyOptInIfNeeded` on first opt-in; resolves real user email via `context.imsClient.getImsAdminProfile()` since `profile.email` is the IMS userId (sub claim), not a real email
- `src/support/email-service.js` — added HTTPS enforcement for `ADOBE_POSTOFFICE_ENDPOINT`, truncated error response body to 200 chars to reduce PII in logs
- `test/controllers/llmo/cdn-opt-in-notification.test.js` (new) — full coverage
- `test/controllers/llmo/llmo.test.js` — added integration tests for opt-in notification trigger including IMS profile lookup failure case
- `test/support/email-service.test.js` — added HTTPS enforcement and truncation tests

**Vault secrets required (api-service)**

| Secret | Notes |
|---|---|
| `OPT_IN_NOTIFICATION_RECIPIENTS` | Comma-separated `@adobe.com` addresses — recommend using a team DL |
| `LLMO_EMAIL_IMS_CLIENT_ID` | IMS client for email token |
| `LLMO_EMAIL_IMS_CLIENT_SECRET` | |
| `LLMO_EMAIL_IMS_CLIENT_CODE` | |
| `LLMO_EMAIL_IMS_SCOPE` | |
| `ADOBE_POSTOFFICE_ENDPOINT` | Must be HTTPS |

**Test plan**
- [ ] Run `npm test` — 100% branch coverage
- [ ] Deploy this PR to dev
- [ ] For a site with `cdnBucketConfig.cdnProvider` set, clear `edgeOptimizeConfig` → click Enable → verify email received with correct CDN name, doc link, org members
- [ ] Verify opt-in API returns 200 even if email fails

Screenshot 2026-04-29 at 4 24 29 PM

Screenshot 2026-04-29 at 4 25 05 PM

Screenshot 2026-04-29 at 4 25 16 PM
a7746-c545-4b5e-8eb5-5a3dc5fa62c8" />


dhavkuma:
@Khush123Adobe  always good to add clear title

dhavkuma:
retest

src/controllers/llmo/cdn-opt-in-notification.js
```diff
@@ -0,0 +1,196 @@
...
+ * @param {string} [params.orgId] - Organization ID used to load members email list.
+ * @param {string} [params.optedBy] - Email of the customer user who triggered the opt-in.
+ * @returns {Promise<{sent: boolean, reason?: string}>}
+ */
+export async function notifyOptInIfNeeded(context, params) {
+  const { log, env } = context;
+  const {
+    siteId,
+    siteBaseURL,
+    cdnLogSource,
```
nit23uec:
```suggestion
    cdnType,
```

src/controllers/llmo/cdn-opt-in-notification.js
```diff
@@ -0,0 +1,196 @@
...
+ */
+
+import { sendEmail } from '../../support/email-service.js';
+
+const OPT_IN_NOTIFICATION_TEMPLATE = 'llmo_cdn_opt_in_notification';
+const OPT_IN_NOTIFICATION_TEMPLATE_ADOBE_MANAGED = 'llmo_cdn_opt_in_notification_adobe_managed';
+const EXCLUDED_MEMBER_STATUSES = new Set(['BLOCKED', 'DELETED']);
+
+const CDN_CONFIG = {
+  'byocdn-fastly': {
```
nit23uec: these names should be consistent with the cdn types used across LLMO.

src/controllers/llmo/cdn-opt-in-notification.js
khushboo5723: done

src/controllers/llmo/cdn-opt-in-notification.js
khushboo5723: yes, imported from 
`import { CDN_TYPES, CDN_DISPLAY_NAMES } from './llmo-utils.js';`

khushboo5723:
## Review: PR #2256 — feat: notify LLMO team when a customer opted in for optimise at edge

Overall: nicely structured helper, good test coverage on `cdn-opt-in-notification.js`. Main concern is in the caller in `llmo.js` — the "fire-and-forget" comment isn't actually true given how the awaits are sequenced. A few lower-priority items below.

### 🔴 Blocking — opt-in response is no longer "non-blocking"

`src/controllers/llmo/llmo.js:1333-1383`

The comment says "Fire-and-forget internal email — does not block the opt-in response," but everything that prepares the call is `await`-ed inline before `notifyOptInIfNeeded` is invoked:

```js
const adminProfile = await context.imsClient.getImsAdminProfile(imsUserId);   // network
const botCheck     = await detectBotBlocker({ baseUrl: baseURL });             // network, 5s default timeout
const { config }   = await readConfig(siteId, s3.s3Client, { s3Bucket: s3.s3Bucket }); // S3 GET
notifyOptInIfNeeded(context, { ... }).catch(...);                              // only this is FAF
```

`detectBotBlocker` alone uses `AbortSignal.timeout(5000)` (see `node_modules/@adobe/spacecat-shared-utils/src/bot-blocker-detect/bot-blocker-detect.js:332`). Combined with the IMS round-trip and the S3 GET, the customer-facing opt-in API can be delayed by 5–10s (or longer on cold IMS/S3) before the response is returned.

Suggested fix — move the entire prep+send block into a single fire-and-forget closure:

```js
if (isNewlyOpted) {
  // Slack notification stays awaited (fast, in-region).

  // Truly fire-and-forget: don't block the API response on IMS/HEAD/S3.
  (async () => {
    try {
      const imsUserId = profile?.email;
      let optedBy;
      if (imsUserId && context.imsClient) {
        try {
          optedBy = (await context.imsClient.getImsAdminProfile(imsUserId))?.email;
        } catch (e) { log.warn(`[cdn-opt-in-notification] IMS lookup failed: ${e.message}`); }
      }

      let botBlocked = false;
      let botBlockerType = '';
      try {
        const botCheck = await detectBotBlocker({ baseUrl: baseURL });
        botBlocked = botCheck?.crawlable === false;
        botBlockerType = botCheck?.type || '';
      } catch (e) { log.warn(`[cdn-opt-in-notification] Bot blocker check failed: ${e.message}`); }

      let cdnType;
      try {
        if (s3?.s3Client) {
          const { config: llmoCfg } = await readConfig(siteId, s3.s3Client, { s3Bucket: s3.s3Bucket });
          cdnType = llmoCfg?.cdnBucketConfig?.cdnProvider;
        }
      } catch (e) { log.warn(`[cdn-opt-in-notification] S3 read failed: ${e.message}`); }

      await notifyOptInIfNeeded(context, {
        siteId, siteBaseURL: baseURL, cdnType,
        orgId: site.getOrganizationId?.(), optedBy, botBlocked, botBlockerType,
      });
    } catch (err) {
      log.error('[cdn-opt-in-notification] Unhandled error:', err);
    }
  })();
}
```

⚠️ Lambda gotcha — even with this change, AWS Lambda freezes the execution context once the handler returns its response. The detached promise won't get to run unless the runtime keeps the event loop alive. If this service uses helix-deploy, please confirm there's a `waitUntil` / pending-promises hook (e.g. `context.attributes.pendingPromises = [...]`) or the email will be silently dropped on warm starts. If there isn't one, the less bad alternative is to keep the awaits but accept the latency, OR queue the notification via SQS to decouple it entirely.

### 🟠 cdnType from request body is ignored

`src/controllers/llmo/llmo.js:1359-1367`

The endpoint already receives `cdnType` in `context.data` (used for routing later), but the notification reads `cdnType` only from S3 (`llmoCfg?.cdnBucketConfig?.cdnProvider`). Per the comment, that's because auth-service writes the canonical value during provisioning — but this means: if a customer opts in before provisioning has written to S3, every notification will say "Unknown CDN type" and trigger the warn log path. Worth either:

- Falling back to the request-body `cdnType` when the S3 lookup yields nothing, or
- Documenting why the body value is intentionally discarded (sequencing guarantee that provisioning always runs first?).

### 🟠 profile?.email as IMS userId is fragile

`src/controllers/llmo/llmo.js:1336-1346`

The comment is doing a lot of work here: "profile.email is the IMS userId (sub claim)." That's only true for IMS-authenticated users. The `authWrapper` in this service supports JWT, IMS, scoped API keys, and legacy API keys (per CLAUDE.md). For non-IMS auth paths, `profile?.email` could be:

- An actual email (JWT) — `getImsAdminProfile(email)` will likely 4xx and we silently fall to `optedBy=undefined`.
- `undefined` (API key flows) — we skip cleanly, fine.

The current code degrades gracefully (warn log, empty `optedBy`), so this is not a bug — but please add a short note to the comment about why the lookup is best-effort and that non-IMS callers will simply get an empty `optedBy`. Otherwise the next reader will assume it's broken.

### 🟡 Concurrent opt-in race

`src/controllers/llmo/llmo.js:1305-1314`

`isNewlyOpted = !existingEdgeConfig.opted` followed by a non-atomic `getConfig → updateEdgeOptimizeConfig → saveSiteConfig` means two concurrent opt-in calls can both observe `opted=false`, both fire the email, and the second save will overwrite the first's `Date.now()`. Probably rare (admin-only endpoint), but worth a note — CLAUDE.md explicitly calls out "avoid read-modify-write patterns" for concurrent updates.

### 🟡 Recipient regex blocks legitimate Adobe sub-domains

`src/controllers/llmo/cdn-opt-in-notification.js:65`

```js
/^[^@]+@adobe\.com$/
```

Rejects `@uk.adobe.com`, `@corp.adobe.com`, etc. If that's intentional (only the canonical `@adobe.com` mailboxes are valid), worth a one-line comment in the code. If unintentional, consider `/^[^@]+@(?:[a-z0-9-]+\.)*adobe\.com$/i`.

### 🟡 HTML entity in CSE_LOOKUP_TEAM couples helper to renderer

`src/controllers/llmo/cdn-opt-in-notification.js:34`

```js
const CSE_LOOKUP_TEAM = 'Customer CSE (run `/ams-whois &lt;customer-name&gt;` in Slack to find them)';
```

Pre-encoding `&lt;`/`&gt;` only works if the template engine renders this string as raw HTML. If it ever gets rendered as plain text (e.g. txt fallback for the multipart email), recipients will see literal `&lt;customer-name&gt;`. Safer to keep the source string as `<customer-name>` and let the Post Office template do its own escaping. If you've verified the template is HTML-only, please leave a comment to that effect.

### 🟢 Nits

- `src/support/email-service.js:113` — `responseText.slice(0, 200)` can split a multi-byte UTF-8 codepoint mid-character. Cosmetic only (logs only); ignore unless you've seen it cause issues.
- `src/controllers/llmo/cdn-opt-in-notification.js:146` — `cdnEntry?.adobeManaged === true` could be `Boolean(cdnEntry?.adobeManaged)`. Pure style.
- `test/it/` — CLAUDE.md states "New or modified endpoints must include integration tests". This PR modifies the behavior of `POST /sites/{siteId}/llmo/edge-optimize-config` (new email side-effect) but I don't see an IT update. Worth confirming whether IT coverage is needed here, or whether the unit-level `llmo.test.js` additions are considered sufficient given the side-effect is fire-and-forget.

✅ Unit tests in `test/controllers/llmo/cdn-opt-in-notification.test.js` are thorough — good coverage of CDN matrix, recipient filtering, fallback paths, and error swallowing.

✅ Email-service changes (HTTPS enforcement + error truncation) are good security/observability improvements and well-tested.

src/controllers/llmo/llmo.js
```diff
@@ -26,6 +26,7 @@ import {
   schemas,
   composeBaseURL,
   isValidUrl,
+  detectBotBlocker,
```
nit23uec: IIUC, this detects only spacecat scraping blocked?

src/controllers/llmo/llmo.js
```diff
@@ -1326,6 +1330,66 @@ function LlmoController(ctx) {
           // Log error but don't fail the request
           log.error(`[edge-optimize-config] Failed to send Slack notification for site ${siteId}:`, slackError);
         }
+
+        // Detached from the response path so the UI's enable click is not blocked
+        // by the IMS lookup, bot-blocker probe (5s timeout), S3 read, and email
+        // send. Lambda's default callbackWaitsForEmptyEventLoop=true keeps the
+        // invocation alive until this promise settles, so the email still goes out.
+        ((async () => {
```
nit23uec, line 1337: in case of cs-fastly, if routing is configured correctly, is there any use of this mail?

src/controllers/llmo/llmo.js
khushboo5723: Right — technically this checks whether spacecat scraping is blocked. 
AFAIU CDN bot blockers don't single us out. They block all datacenter/non-browser bots the same way, so "spacecat blocked" effectively means "site is blocking bots." 
Good enough for what LLMO needs to know.

let me know if this is not right

src/controllers/llmo/llmo.js
khushboo5723: Right, no real coordination needed for AEM CS Fastly. 
Keeping the email for all CDN types for now — opt-in volume is low and it's still useful as an confirmation. 
Easy to filter later if it turns out to be noise.

src/controllers/llmo/llmo.js
nit23uec: this is incorrect. this could give us false responses for the cases where spacecat bot is allowed but OaE is blocked.

src/controllers/llmo/llmo.js
nit23uec: ack.

src/controllers/llmo/llmo.js
khushboo5723: as discussed, removed the botblocked check for now, as we don't have a way to detect OaE blocking at opt-in time
@shrotia is working on this, and once it is in prod and live I will update the email content 
JIRA : https://jira.corp.adobe.com/browse/LLMO-4635

nit23uec (approved):

PR #2257, opened 2026-04-23
feat(llmo): add DELETE edge-optimize-config with CDN routing rollback | LLMO-4281 LLMO-4229

anagarwa:
## Summary

- Adds `DELETE /sites/:siteId/llmo/edge-optimize-config` endpoint (LLMO-4281) — previously there was no way to delete the edge optimization config from the UI
- On delete, rolls back CDN routing via the CDN API when routing was active (LLMO-4229)
- Wires the route through `src/routes/index.js` and adds it to the `NO_S2S_ROUTES` list in `required-capabilities.js` (same as the POST)

### What the handler does

1. **Access control**: same guards as the POST (hasAccess + isLLMOAdministrator + isOwnerOfSite)
2. **S3 metaconfig disable** (best-effort): calls `tokowakaClient.updateMetaconfig` with `tokowakaEnabled: false` to turn off the worker. Failures are logged as warnings and do not block the delete.
3. **CDN routing rollback** (LLMO-4229): if `edgeOptimizeConfig.enabled === true` and the caller supplies `cdnType` in the request body, the handler replays the POST handler's auth flow (IMS token exchange → org authorization → SP token → CDN API with `enabled: false`). Failures at any step are logged but do not block the config deletion.
4. **DynamoDB cleanup**: calls `currentConfig.updateEdgeOptimizeConfig(undefined)` and saves the site, clearing the `edgeOptimizeConfig` block from the site record.

### Why CDN rollback requires `cdnType` in the body

The CDN type is not persisted in `edgeOptimizeConfig` — it was only used at routing time. The caller (LLMO UI) knows which CDN type was configured and must pass it back. If omitted, a warning is logged but the config is still deleted.

## Test plan

- [x] `npm test` passes — 8323 tests, 0 failures (up from 8309 before this change)
- [x] `npm run lint` clean on all modified files
- [x] 13 new unit tests covering: 404 site-not-found, 403 access/admin/owner guards, happy-path metaconfig disable, S3-not-found skip, best-effort S3 failure, CDN rollback happy path, CDN rollback skipped (routing not enabled), CDN rollback skipped (no cdnType), CDN rollback skipped (non-prod env), CDN rollback failure non-blocking, fallback lastModifiedBy
- [x] Route registration test updated in `test/routes/index.test.js`

## Resolves

- LLMO-4281 — Delete edgeOptimize config when CDN config is deleted on LLMO
- LLMO-4229 — Rollback routing if user deletes CDN config from the UI

### DIFF
--- a/src/controllers/llmo/llmo.js
+++ b/src/controllers/llmo/llmo.js
@@ -1492,6 +1492,149 @@ function LlmoController(ctx) {
   };
 
   /**
+   * DELETE /sites/{siteId}/llmo/edge-optimize-config
+   * Deletes Tokowaka edge optimization configuration for a site.
+   * - Clears edgeOptimizeConfig from the site's DynamoDB config
+   * - Disables tokowaka in the S3 metaconfig (best-effort, does not fail the request)
+   * - If CDN routing was active and cdnType is provided, rolls back routing via the CDN API
+   *   (LLMO-4229: rollback routing if user deletes CDN config from the UI)
+   * @param {object} context - Request context
+   * @returns {Promise<Response>} 200 on success
+   */
+  const deleteEdgeConfig = async (context) => {
+    const { log, dataAccess, env } = context;
+    const { siteId } = context.params;
+    const { authInfo: { profile } } = context.attributes;
+    const { Site } = dataAccess;
+    const { cdnType } = context.data || {};
+
+    try {
+      const site = await Site.findById(siteId);
+
+      if (!site) {
+        return notFound('Site not found');
+      }
+
+      if (!await accessControlUtil.hasAccess(site)) {
+        return forbidden('User does not have access to this site');
+      }
+
+      if (!accessControlUtil.isLLMOAdministrator()) {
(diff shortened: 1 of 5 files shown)


nit23uec:
LLMO-4229 ("rollback routing if user deletes CDN config from UI") is already satisfied by the existing `POST /edge-optimize-config { enabled: false, cdnType }` path. The UI just need to call it on the delete action. No API change.
with the above, LLMO-4281 can also be handled as the existing /edge-optimize-config API also updates `edgeOptimizeConfig` to the correct state.

anagarwa:
A UI-side fix has been opened in project-elmo-ui to address LLMO-4229 and LLMO-4281 without a new API endpoint, as suggested by the review: https://github.com/adobe/project-elmo-ui/pull/1616

The PR extends the existing Disable button and flow to BYOCDN providers (Cloudflare, Akamai, Fastly, CloudFront, Imperva). On confirm it calls the existing POST /sites/:siteId/llmo/edge-optimize-config with { enabled: false, cdnType } — no new DELETE endpoint needed.

PR #1814, opened 2026-02-15
feat: resolve userId slack command

ravverma:
Please ensure your pull request adheres to the following guidelines:
- [ ] make sure to link the related issues in this description. Or if there's no issue created, make sure you 
  describe here the problem you're solving.
- [ ] when merging / squashing, make sure the fixed issue references are visible in the commits, for easy compilation of release notes

If the PR is changing the API specification:
- [ ] make sure you add a "Not implemented yet" note the endpoint description, if the implementation is not ready 
  yet. Ideally, return a 501 status code with a message explaining the feature is not implemented yet.
- [ ] make sure you add at least one example of the request and response.

If the PR is changing the API implementation or an entity exposed through the API:
- [ ] make sure you update the API specification and the examples to reflect the changes.

If the PR is introducing a new audit type:
- [ ] make sure you update the API specification with the type, schema of the audit result and an example

## Related Issues

Thanks for contributing!

PR #2266, opened 2026-04-24
feat: add GET /admin/user/:userid endpoint to resolve IMS user profiles

ravverma:
## Summary

- Adds `GET /admin/resolve-user?userId=<imsId>` endpoint gated behind `hasAdminAccess()`
- Resolves an IMS user ID to `{ firstName, lastName, email }` via the existing `ImsClient.getImsAdminProfile()` call
- Route is added to `INTERNAL_ROUTES` so it is excluded from S2S consumer access
- No changes to `src/index.js` — `userDetailsController` was already wired

## Motivation

The existing codebase uses `trial_email` in some flows as a source of user email addresses. `trial_email` is a PII field added as an exception/tech debt to the SpaceCat JWT profile and must not be relied upon. This endpoint provides a clean, admin-only server-side way to resolve a real user profile from IMS, replacing any such usage without requiring callers to handle IMS credentials directly.

## Changes

| File | Change |
|---|---|
| `src/controllers/user-details.js` | Added `resolveUser` method |
| `src/routes/index.js` | Registered `GET /admin/resolve-user` route |
| `src/routes/required-capabilities.js` | Added route to `INTERNAL_ROUTES` |
| `test/routes/index.test.js` | Updated mock + assertions for new route |

## Test plan

- [x] Existing `test/controllers/user-details.test.js` — all 23 tests pass
- [x] `test/routes/` — all 2007 tests pass
- [ ] Manual: call `GET /admin/resolve-user?userId=<valid-ims-id>` with an admin token — expect `{ firstName, lastName, email }`
- [ ] Manual: call without admin token — expect `403 Forbidden`
- [ ] Manual: call without `userId` param — expect `400 Bad Request`


ravverma:
Hi @solaris007 
follow up from the last PR https://github.com/adobe/spacecat-api-service/pull/1814 and implemented admin gated end point as discussed. 
Add it as command in cli and in BO can be taken as per need in future.
Please review

src/controllers/user-details.js
```diff
@@ -205,9 +205,41 @@ function UserDetailsController(ctx) {
     }
   };
 
+  /**
+   * Resolves a user's profile by IMS user ID (admin-only).
+   * @param {object} context - Context of the request.
+   * @returns {Promise<Response>} Resolved user profile.
+   */
+  const resolveUser = async (context) => {
+    if (!accessControlUtil.hasAdminAccess()) {
```
solaris007: **Admin gate gap** - `hasAdminAccess()` returns true for any non-IMS/non-JWT auth, including USER_API_KEY. This route is not in `LegacyApiKeyHandler.ADMIN_ENDPOINTS`, so USER_API_KEY holders pass the gate and can enumerate arbitrary IMS profiles. Strengthen to require IMS admin callers specifically, or add this route to `ADMIN_ENDPOINTS`.

src/controllers/user-details.js
```diff
@@ -205,9 +205,41 @@ function UserDetailsController(ctx) {
...
+      return forbidden('Only admins can resolve user profiles');
+    }
+
+    const userId = context.request?.url
```
solaris007: **PII in query string** - IMS GUIDs are stable user identifiers that persist in AWS API Gateway, ALB, and CDN access logs with different retention/redaction policies than app logs. Consider switching to POST with userId in the body to keep PII out of request lines.

src/controllers/user-details.js
```diff
@@ -205,9 +205,41 @@ function UserDetailsController(ctx) {
...
+    try {
+      const imsProfile = await imsClient.getImsAdminProfile(userId);
+      return ok({
+        firstName: imsProfile.first_name || '-',
+        lastName: imsProfile.last_name || '-',
+        email: imsProfile.email || '',
+      });
+    } catch (e) {
+      log.error(`Failed to resolve user profile for ${userId}: ${e.message}`);
+      return internalServerError(e.message);
```
solaris007: **Error mapping** - All IMS failures become 500 with the raw upstream message echoed to the caller. A bad userId (IMS 404) reads as a spacecat outage in dashboards. Map IMS 4xx to 4xx, use 502 for upstream 5xx, and return a generic error message while keeping detail in the log only.

solaris007 (changes requested):
Hey @ravverma,

Clean, well-tested endpoint addition. Admin gate, INTERNAL_ROUTES wiring, and test coverage are all correctly done. A few items worth addressing before this becomes the reference pattern for future admin endpoints.

## Strengths

- Admin gate fires first, before any URL parsing or IMS call - clean fail-fast.
- Route correctly added to `INTERNAL_ROUTES`, preserving the disjointness-with-`routeRequiredCapabilities` invariant.
- Tests cover all key paths: 403 (non-admin), 400 (missing/null request), 200 (happy path + incomplete IMS profile), and 500 (IMS failure).
- Fallback values (`'-'`, `''`) match the shape returned by the existing `fetchFromImsIfAdmin` helper in the same file, keeping the contract consistent.
- Response is a narrow fixed shape - no raw IMS payload or stack traces forwarded.

## Issues

### Important (Should Fix)

**1. Admin gate does not distinguish USER_API_KEY from ADMIN_API_KEY** - `user-details.js:214`

`hasAdminAccess()` returns true unconditionally for any auth type that is neither IMS nor JWT. The `LegacyApiKeyHandler` only enforces ADMIN_API_KEY for the four routes hard-coded in its `ADMIN_ENDPOINTS` array, and `GET /admin/resolve-user` is not in that list. A caller presenting `USER_API_KEY` passes the admin gate. This is a pre-existing pattern (the sibling `fetchFromImsIfAdmin` has the same gap), but this PR widens blast radius by adding an unrestricted, org-scope-free profile-lookup endpoint. Fix: require IMS-typed callers with admin flag, or add the route to `ADMIN_ENDPOINTS`, or compare the API key directly against `context.env.ADMIN_API_KEY`.

**2. userId (PII) in GET query string persists in access logs** - `user-details.js:218-220`

IMS GUIDs (`ABCDEF123@AdobeOrg`) are stable user-identifying tokens. GET query strings persist in AWS API Gateway access logs, CloudWatch, ALB logs, and CDN observability with different retention/redaction policies than app logs. The motivation of this PR is to move PII out of JWTs - but the lookup request itself now records the target identifier across the full request path. Consider switching to `POST /admin/resolve-user` with `{ userId }` in the body, which keeps the identifier out of request lines.

**3. Path param vs query string inconsistency** - `user-details.js:218`, `routes/index.js:514`

Sibling routes on this controller identify users via path params (`GET /organizations/:organizationId/userDetails/:externalUserId`). This route uses a query string and manually parses with `new URL(...)`. For a GET-by-identifier, the idiomatic choice - and the one matching neighbouring routes - is a path segment (e.g. `GET /admin/users/:userId`). This removes the manual URL parsing and uses `context.params` like every other handler in the file.

**4. Missing audit log for PII access** - `user-details.js:226-232`

Only the failure branch logs. This endpoint hands back `{ firstName, lastName, email }` for arbitrary IMS users to any admin caller. Without a success-path log identifying which admin resolved which userId, there is no audit trail for PII access review or compliance questions. Add an info-level log: `Admin ${callerProfile.user_id} resolving IMS profile for ${userId}`.

**5. IMS error mapping collapses all failures to 500** - `user-details.js:233-235`

`getImsAdminProfile` throws errors with messages encoding upstream status (`IMS getAdminProfile request failed with status: 404`). All become a 500 here with the upstream message echoed to the caller. A bad userId (IMS 404) reads as a spacecat-side outage in dashboards, eroding the signal of your own 5xx rate. Map IMS 4xx to 4xx, IMS 5xx to 502, and return a generic error message to the caller while keeping `e.message` in the log only.

### Minor (Nice to Have)

**6. No format validation on userId** - `user-details.js:222`

`hasText(userId)` accepts any non-empty string. A light format check (e.g. matching the documented IMS GUID pattern `^[A-Za-z0-9]+@[A-Za-z0-9.]+$`) would give a 400 on obvious junk without paying for an IMS round trip.

**7. Near-duplicate profile-shape mapping** - `user-details.js:228-232`

The `first_name -> firstName`, `'-'` sentinels, `email: ''` default now lives in two places in this file (here and in `fetchFromImsIfAdmin`). Extract a shared `toProfileResponse(imsProfile)` so the contract stays single-sourced.

## Recommendations

- Strongest ask: resolve the admin-gate distinction (finding 1). Decide whether USER_API_KEY should reach this endpoint.
- Consider switching to POST to keep PII out of access logs (finding 2), or get explicit sign-off from Privacy/SRE on log handling.
- Track a follow-up in `spacecat-shared-ims-client` to add URL-encoding and format validation in `getImsAdminProfile`/`extractGuidAndAuthSource`.

## Assessment

The endpoint is well-tested and correctly wired. The admin gate gap (finding 1) and PII-in-query-string (finding 2) are the highest priority - they affect who can reach this endpoint and where the PII lands. The audit log (finding 4) and error mapping (finding 5) are quick fixes that set the right pattern before this becomes the reference for future admin endpoints.

## Next Steps

1. Strengthen the admin gate to exclude USER_API_KEY callers.
2. Switch to POST or document log-redaction for PII in query string.
3. Add success-path audit log for PII access.
4. Map IMS errors to appropriate HTTP status codes.
5. Minor items are optional but cheap to address in the same pass.

ravverma:
@solaris007 thanks, review comments addressed.


---

## Changes addressed from review

**#1 — Admin gate strengthened**
Replaced the `legacyApiKey`/`ADMIN_API_KEY` comparison with `isAccessTypeJWT()` — only JWT-authenticated admins can reach this endpoint. API keys (user or admin), IMS, and S2S tokens are all rejected at the gate.

**#2/#3 — PII in query string / path param inconsistency**
Route changed from `GET /admin/resolve-user?userId=...` to `GET /admin/users/:userId` — consistent with sibling routes on this controller, uses `context.params` like every other handler, and removes the manual `new URL(...).searchParams` parsing.

**#4 — Audit log for PII access**
Added `log.info(...)` on successful resolution so every admin lookup of a user profile leaves a trace.

**#5 — IMS error mapping**
Parses the upstream status from the IMS error message and maps it: 404 → `notFound`, other 4xx → `badRequest`, anything else → `internalServerError` with a generic message (upstream detail stays in the log only).

**#6 — userId format validation**
Regex check `^[A-Za-z0-9]+@[A-Za-z0-9.]+$` returns 400 on obviously malformed input without paying for an IMS round trip.

**#7 — Shared profile shape helper**
Extracted `toProfileShape(imsProfile)` used by both `fetchFromImsIfAdmin` and `resolveUser` — single source of truth for the `first_name → firstName` mapping and fallback sentinels.

solaris007 (commented):
Hey @ravverma,

Re-reviewing after 3 new commits. All 7 findings from the prior review have been addressed - nice work.

## Strengths (previously flagged issues now resolved)

- **Admin gate strengthened** (`user-details.js:223`): Now requires `isAccessTypeJWT() && hasAdminAccess()`, closing the USER_API_KEY escalation path. Test added for non-JWT callers.
- **Path param convention adopted** (`routes/index.js:514`): Route changed from `GET /admin/resolve-user?userId=...` to `GET /admin/users/:userId`, consistent with sibling routes. Removes the manual `new URL(...)` parsing.
- **Audit log added** (`user-details.js:236`): Success-path `log.info` traces every admin profile lookup. Test verifies the log fires.
- **IMS error mapping implemented** (`user-details.js:239-250`): Parses upstream status - 404 maps to `notFound`, other 4xx to `badRequest`, and the generic 500 no longer echoes raw IMS error strings.
- **userId format validation** (`user-details.js:28, 233`): `IMS_USER_ID_RE` regex rejects malformed input before the IMS round trip. Test added.
- **Shared profile shape extracted** (`user-details.js:30-35`): `toProfileShape()` deduplicates the `first_name -> firstName` mapping used by both `fetchFromImsIfAdmin` and `resolveUser`.

## Issues

### Minor (Nice to Have)

**1. CI failing: missing test for non-404 4xx IMS error branch** - `user-details.js:249-250`

The coverage report shows lines 249-250 uncovered (the `upstreamStatus >= 400 && upstreamStatus < 500` branch). Add a test where IMS throws with `status: 403` and assert the response is 400. One-liner fix.

**2. Audit log identifies the target but not the caller** - `user-details.js:236`

`Admin resolved IMS profile for userId: ${userId}` traces what was looked up but not who did the looking. Adding the caller's identity (e.g. `context.attributes?.authInfo?.getProfile?.()?.user_id`) would make the audit trail complete for compliance questions. Low priority - the JWT auth context is already in the framework's access logs.

## Assessment

**Ready to merge? Yes, after CI fix.**

All substantive findings from the prior review are addressed cleanly. The admin gate is now JWT-only, the route follows controller conventions, error mapping is proper, and the audit trail exists. Just need the one missing test case to get CI green.

ravverma:
@solaris007 CI fixed

PR #2268, opened 2026-04-24
fix: llmo include topicId in the get topics api call

vivesing:
# PR summary — Brand Presence `topicId` (topics, search, docs, tests)

## Overview

Exposes a stable **topic UUID** as **`topicId`** on Brand Presence **topic list** and **search** responses so clients can call `…/topics/:topicId/{prompts,detail,prompt-detail}` using an id when available, instead of relying only on display topic names. Includes documentation updates, unit tests, and coverage for edge cases on UUID normalization.

---

### `GET …/brand-presence/topics` (`createTopicsHandler`)

- Each `topicDetails[]` entry includes **`topicId`**, mapped from RPC **`topic_id`** via **`normalizeRpcUuid`** (trimmed string, or JSON **`null`** when missing/blank/invalid type).
- ESLint **`curly`**: `normalizeRpcUuid` uses a braced `if` for the type guard.

### `GET …/brand-presence/search` (`createSearchHandler`)

- **`TOPICS_SELECT`** extended with **`topic_id`** so PostgREST returns execution topic ids.
- **`aggregateTopicData`**:
  - Tracks **`maxTopicId`** per topic label (representative UUID per group: lexicographically greatest non-null `topic_id`, aligned with `rpc_brand_presence_topics` ordering intent).
  - Each aggregated object includes **`topicId`** (`null` when no usable `topic_id`).
- Search **`topicDetails`** inherit **`topicId`** through the existing `…map((td) => ({ ...td, matchType, … }))` flow.

### Helpers

- **`normalizeRpcUuid(value)`** — accepts only `string` or `number`; returns trimmed non-empty string or **`null`** (including whitespace-only strings after trim — **`return s || null`**).
- **`maxTopicIdForAggregate(current, candidate)`** — merges candidates into the per-topic max id.
- **`normalizeRpcUuid`** is placed **above** `aggregateTopicData` and reused by topics RPC mapping and aggregation (no duplicate definition).

### Topic detail handler

- Still uses **`aggregateTopicData(rows)[0]`** for the stats block only; root **`topicId`** on the topic-detail response continues to come from **`topicIdForDetailResponse`**.

---

### Topics (`createTopicsHandler`)

- **`sampleRpcRow`** includes **`topic_id`**; success case asserts **`topicId`**.
- **`maps null/missing RPC fields`**: expects **`topicId`** **`null`** when **`topic_id`** is absent.
- **`returns null topicId when RPC topic_id is whitespace-only`**: covers **`normalizeRpcUuid`** `**return s || null**` via the topics handler path.

### `aggregateTopicData`

- **`groups rows by topics`**: expects **`topicId`** **`null`** when rows have no **`topic_id`**.
- **`sets topicId to max topic_id per topic group`**: two rows, same topic, different UUIDs — expects the lexicographically greater id.
- **`normalizes blank and whitespace-only topic_id to null topicId`**: **`topic_id: ''`** and **`topic_id: '  \t\n  '`** → **`topicId`** **`null`** (covers line 1386 **`s || null`**).

### Search (`createSearchHandler`)

- Topic-match fixture includes **`topic_id`**; response asserts **`topicId`**.

---

## Documentation (`docs/llmo-brandalf-apis/`)

| File | Change |
|------|--------|
| **`topics-api.md`** | Sample JSON, **`topicId`** field table, prompts path param note, aggregation described as **`rpc_brand_presence_topics`**. |
| **`search-api.md`** | Examples, **`topicId`** in field table, search logic step 2, **`null`** when ids absent. |
| **`brand-presence-apis-overview.md`** | Master table (rows 5 & 7), §5/§7 example JSON and short **`topicId`** / **`null`** notes. |
| **`topic-detail-api.md`** | Clarifies topics list vs **`aggregateTopicData`** for detail stats. |

---

## Deployment / data

- **Topics list**: depends on **`rpc_brand_presence_topics`** returning **`topic_id`** where the DB migration is applied.
- **Search**: depends on **`brand_presence_executions.topic_id`** being populated for meaningful **`topicId`** values.

---

## Backward compatibility

- **Additive** response fields: existing clients using only **`topic`** remain valid.
- **`topicId`** may be **`null`** when the backend has no id or only blank/whitespace values.


calvarezg:
A few items worth discussing before merge:

1. **Confirm `topic_id` is UUIDv7 (not UUIDv4).** `maxTopicIdForAggregate` picks a representative UUID per topic group via lexicographic max, which works correctly for UUIDv7 (time-ordered). If the column is UUIDv4 (random), the max is semantically arbitrary — not the most recent or most relevant. Worth confirming the column type, or documenting the behavior as "arbitrary but stable pick" if UUIDv4 is possible.

2. **Verify no unintentional spread of `topicId` into the topic-detail root response.** `aggregateTopicData` now returns a `topicId` field in its result object. For the topic-detail path, this value is used for the stats block — the root-level `topicId` comes from `topicIdForDetailResponse`. As long as the stats object isn't spread into the root-level detail response, these won't collide. Worth a quick sanity check that no such spread exists.

3. **Minor doc inconsistency.** `brand-presence-apis-overview.md` row 6 still describes the `:topicId` path param as "URL-encoded topic name," while `topics-api.md` was updated to say "or a topic UUID." These should be consistent.

vivesing:
1. Confirmed: topics.id is uuid_generate_v7() (time-sortable), and brand_presence_executions.topic_id FKs to it. So max UUID string order lines up with “newest” topic id in the group, same as the RPC’s ORDER BY topic_id DESC—not a random v4 pick.

2. Checked createTopicDetailHandler: the body uses topicId: topicIdResponse at the root and builds stats from explicit topicStats.* fields only — no ...topicStats (or any spread of aggregateTopicData output). So aggregated topicId is not merged into the response; root topicId can’t collide with it.

3. fixed the doc

vivesing:
Hi @calvarezg , addressed the review comments, please see

calvarezg (approved):

PR #2271, opened 2026-04-27
feat(suggestions): gate EDS auto-fix behind IMS group for PLG/FREE_TRIAL tiers

ravverma:
## Summary

Guards the `autofixSuggestions` endpoint for **AEM Edge** sites on **PLG** / **FREE_TRIAL** ASO tiers. Freemium orgs cannot be assigned per-user product profiles, and EDS auto-fix applies changes via a service account disconnected from the caller — so instead of blocking these tiers outright, auto-fix is gated behind membership of a dedicated IMS group (`ASO-EDS-Autofix-users`):

- **Members of the group** → auto-fix proceeds.
- **Everyone else** → `403 unauthorized`.
- **Paid tier and AEM CS sites** → unchanged.

### How it works

- Resolves entitlement tier + active site enrollment via `TierClient.createForSite` + `checkValidEntitlement()`. The group gate only engages when the site is `AEM_EDGE`, the tier is `PLG`/`FREE_TRIAL`, **and** `siteEnrollment` is non-null.
- Recovers the caller's IMS identity by exchanging the forwarded promise token (`getImsTokenFromPromiseToken`), then checks IMS group membership.
- **Group membership check is a shared helper** — `isImsGroupMember` (`src/support/ims-group.js`), extracted from the existing LLMO Admin group check in `edge-routing-auth.js`, which now reuses it. Same logic, one source of truth.
- **Fail-closed**: a missing promise token or a failed IMS lookup denies (403). The outer entitlement lookup stays fail-open (paid users are never blocked on a transient error), matching the original behavior.
- Logs `warn` on block / `info` on allow, with the site ID, for observability.

### Note on token consumption

The promise token is single-use (exchange rotates it), and the autofix flow forwards it to the worker. Consuming it here is safe **on the EDS path only** because EDS fixes run under service credentials, not the promise token (unlike promise-based CS authoring). Worth confirming during review that no EDS worker path depends on that token.

### Tests

- New `test/support/ims-group.test.js` — unit coverage for the shared helper (member, case-insensitive match, non-member, no org match, no groups, IMS throw → fail-closed, missing inputs).
- `suggestions.test.js` — added allow (group member) and deny (non-member) cases alongside the existing PLG/FREE_TRIAL/no-enrollment tests.
- Existing `edge-routing-auth.test.js` still green after the refactor.

Tracked by [SITES-50350](https://jira.corp.adobe.com/browse/SITES-50350).

## Change Management

```yaml
cm-assessment: v1
changeType: normal
impact: degradation
risk: major
changeApprovedBy: ["Dominique Jäggi", "Tarun Singh"]
rationale: "Adds a new authorization-boundary check (IMS group gate, fail-closed) controlling who can trigger automated content fixes on customer sites; auth/access-control changes escalate per rubric despite fail-closed design and narrow tier scope."
```


ravverma:
@solaris007 from review POV,
these two PRs will not be released
https://github.com/adobe/spacecat-api-service/pull/2267
https://github.com/adobe/spacecat-auth-service/pull/557
I have created this one to fix the flow
thanks
cc: @mimchome

src/controllers/suggestions.js
```diff
@@ -34,6 +34,8 @@ import {
   Audit,
   Suggestion as SuggestionModel,
   GeoExperiment as GeoExperimentModel,
```
MysticatBot, line 38: issue (blocking): **Fail-open outer catch bypasses the freemium authorization gate.**

The entire entitlement + IMS group membership check is wrapped in an outer `try/catch` that logs a warning and falls through (line ~1431 in the new code). The PR description says the entitlement lookup is intentionally fail-open so paid users are never blocked on transient errors. However, `site.getOrganization()` and the IMS group check execute *after* the user is confirmed as PLG/FREE_TRIAL - i.e., inside `if (blockedTiers.includes(...) && siteEnrollment)`. If `site.getOrganization()` throws at that point, the outer catch absorbs it and auto-fix proceeds ungated for a confirmed freemium user.

**Why it matters:** A transient DB error, a misconfigured org record, or any unexpected exception in the post-tier-determination code causes the authorization gate to be skipped for the exact users it exists to block. The `isImsGroupMember` helper is carefully fail-closed, but its protection is nullified by the outer catch wrapping the code path that leads to it.

**Fix:** Nest a separate try/catch inside the `if (blockedTiers.includes(...))` block that fails closed (returns `forbidden`) on error, keeping the outer catch fail-open only for the tier-determination phase:

```js
if (blockedTiers.includes(entitlement?.getTier()) && siteEnrollment) {
  try {
    const org = await site.getOrganization();
    const imsOrgId = org?.getImsOrgId();
    let imsUserToken;
    try {
      imsUserToken = await getImsTokenFromPromiseToken(context);
    } catch (tokenErr) {
      context.log?.warn(`...`);
    }
    const isGroupMember = await isImsGroupMember(context, {
      imsOrgId, imsUserToken, groupName: ASO_EDS_AUTOFIX_GROUP_NAME,
    }, context.log);
    if (!isGroupMember) {
      context.log?.warn(`Auto-fix blocked...`);
      return forbidden('unauthorized');
    }
    context.log?.info(`Auto-fix allowed...`);
  } catch (gateErr) {
    context.log?.warn(`Freemium gate error for site ${siteId}: ${gateErr.message}`);
    return forbidden('unauthorized');
  }
}
```

This preserves fail-open for tier determination (TierClient/checkValidEntitlement errors fall through) while ensuring the group-membership gate is fail-closed once the user is confirmed freemium.

MysticatBot (changes requested):
Hey @ravverma,

**Verdict: Request changes** - the outer try/catch makes the freemium authorization gate fail-open under infrastructure errors.
**Complexity: MEDIUM** - medium diff; auth/IMS + API surface signals.
_Changes: Gates EDS auto-fix behind IMS group membership for PLG/FREE_TRIAL tier sites, extracting a shared isImsGroupMember helper from edge-routing-auth (5 files)._

## Must fix before merge
1. [Important] Outer catch bypasses freemium gate when post-tier-determination calls throw - `src/controllers/suggestions.js:~1431` (details inline)

<details>

- suggestion: When `getImsTokenFromPromiseToken` throws, the user gets a generic "unauthorized" 403 with no indication that the root cause is a missing/expired promise token. Consider returning a specific error immediately on token failure rather than falling through to a guaranteed denial - `src/controllers/suggestions.js:~1417`
- suggestion: Use `context.log?.warn` consistently (line ~1425 uses `context.log.warn` without optional chaining while line ~1419 uses `context.log?.warn`). If `context.log` is undefined, the non-optional call throws a TypeError that propagates to the outer catch and triggers the fail-open path - `src/controllers/suggestions.js:~1425`
- nit: `blockedTiers` array is re-created on every request inside the hot path. Hoist to module scope alongside `ASO_EDS_AUTOFIX_GROUP_NAME` - `src/controllers/suggestions.js:~1407`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 33s | <b>Cost:</b> $9.09 | <b>Commit:</b> <code>aa8bd0730d1542a4e48e16e9f7378196ec72b6be</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/controllers/suggestions.js
tarunsinghdev: Fixed in 6c7c9468c. Wrapped the post-tier-determination code (org / promise-token / group-membership resolution) in a nested `try/catch` that returns `forbidden('unauthorized')` on any error — so a confirmed freemium user (blocked tier + active enrollment) can no longer slip through the outer fail-open catch ungated. The outer catch stays fail-open **only** for the `TierClient.checkValidEntitlement` tier-determination phase, so paid users are never blocked on a transient error. Added a test that asserts 403 when `site.getOrganization()` throws for a confirmed-freemium user.

🤖 Generated by resolve-review-comments skill.

tarunsinghdev:
> **Verdict: Request changes** — the outer try/catch makes the freemium authorization gate fail-open under infrastructure errors. (@MysticatBot review)

Addressed in 6c7c9468c. The freemium gate is now fail-closed: once the user is confirmed PLG/FREE_TRIAL with an active enrollment, org / promise-token / group-membership resolution runs inside a nested `try/catch` that returns `forbidden('unauthorized')` on any error. The outer catch remains fail-open **only** for the `TierClient.checkValidEntitlement` tier-determination phase, so paid users are never blocked on a transient error. Added a unit test asserting 403 when `site.getOrganization()` throws for a confirmed-freemium user. Full suggestions suite: 542 passing.

🤖 Generated by resolve-review-comments skill.

MysticatBot (approved):
Hey @ravverma,

**Verdict: Approve** - prior blocking finding addressed; remaining items are non-blocking style nits.
_Changes: Gates EDS auto-fix behind IMS group membership for PLG/FREE_TRIAL tier sites, extracting a shared isImsGroupMember helper (5 files)._

<details>

- suggestion: When `getImsTokenFromPromiseToken` throws, the caller receives a generic `403 "unauthorized"` indistinguishable from a legitimate non-member denial. Consider returning early with a distinct message (e.g. `forbidden('authentication required')`) so callers and support can distinguish token failure from group non-membership. Counter-consideration: generic 403s avoid oracle attacks on group membership - `src/controllers/suggestions.js:~1420`
- nit: Inconsistent optional chaining on `context.log` within the gate block - line ~1426 uses `context.log?.warn` while lines ~1432, 1435, 1437 use `context.log.warn`/`.info` without `?.`. Pick one style (the controller guarantees `context.log`, so dropping `?.` everywhere is clearest) - `src/controllers/suggestions.js:~1426`
- nit: `blockedTiers` array is re-created on every AEM Edge request. Hoist to module scope alongside `ASO_EDS_AUTOFIX_GROUP_NAME` since the values are constants - `src/controllers/suggestions.js:~1408`
</details>

- Outer try/catch no longer bypasses the freemium authorization gate - nested try/catch now fails closed for confirmed freemium callers while remaining fail-open for the tier-determination phase.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 32s | <b>Cost:</b> $4.95 | <b>Commit:</b> <code>6c7c9468c6a6895d90daee3caf8efb3496fc8288</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/controllers/suggestions.js
```diff
@@ -67,11 +70,18 @@ import AccessControlUtil, { X_PRODUCT_HEADER } from '../support/access-control-u
...
+import { isImsGroupMember } from '../support/ims-group.js';
 import { postSlackMessage } from '../utils/slack/base.js';
 import { createAtomicStrategy, deleteAtomicStrategy } from '../support/atomic-strategy-helper.js';
 
 const VALIDATION_ERROR_NAME = 'ValidationError';
 
+// Freemium (PLG / FREE_TRIAL) orgs cannot be assigned per-user product profiles,
+// so EDS auto-fix is gated behind membership of this IMS group instead. Members
+// of the group are allowed to trigger auto-fix; everyone else is blocked.
+const ASO_EDS_AUTOFIX_GROUP_NAME = 'ASO-EDS-Autofix-users';
```
ravverma: make `-users` -> `-Users`

src/support/ims-group.js
```diff
@@ -0,0 +1,59 @@
...
+
+  try {
+    const orgs = await context.imsClient.getImsUserOrganizations(imsUserToken);
+    const matchingOrg = orgs?.find(
+      (o) => `${o.orgRef?.ident}@${o.orgRef?.authSrc}` === imsOrgId,
+    );
+    if (!matchingOrg) {
+      return false;
+    }
+    const target = groupName.toLowerCase();
```
ravverma: the match should be case sensitive

ravverma (commented):
Address comments

src/controllers/suggestions.js
tarunsinghdev: Fixed in 355ec8e12. Renamed the constant to `ASO-EDS-Autofix-Users` (capital U).

🤖 Generated by resolve-review-comments skill.

src/support/ims-group.js
tarunsinghdev: Fixed in 355ec8e12. Dropped the `toLowerCase()` on both sides — membership now requires an exact, case-sensitive `groupName` match. Updated the ims-group unit test to assert a differently-cased name is *not* a member.

🤖 Generated by resolve-review-comments skill.

PR #2272, opened 2026-04-27
fix(llmo): trigger DRS prompt generation for v1 onboarding (LLMO-4534)

byteclimber:
## Related Issues

- [LLMO-4534](https://jira.corp.adobe.com/browse/LLMO-4534) — new onboardings in v1 orgs do not get prompts generated

## Summary

V1 onboarding does not run Brandalf, so DRS never auto-submits a `prompt_generation_base_url` job for these orgs. The legacy direct call to prompt generation was removed by [LLMO-4258](https://jira.corp.adobe.com/browse/LLMO-4258) (commit ed5363d1) without preserving a v1 fallback, leaving v1 onboardings with no prompt-generation trigger at all. Customers like ` pro.sony` ended up with sites onboarded but no prompts ever generated, and required a manual DRS job submission.

## Fix

In the v1 branch of ` performLlmoOnboarding`, call `DrsClient.submitPromptGenerationJob` directly. The job is submitted with `source='onboarding'` and no `onboarding_mode` metadata, so the audit-worker `drs-prompt-generation` handler takes the legacy-write path (`shouldWriteLegacyConfig = source !== 'onboarding' || onboardingMode !== 'v2'` evaluates to true) and then triggers `llmo-customer-analysis` as expected.

The v2 path (Brandalf → DRS scheduler → prompt gen) is untouched.

## Test plan

- [x] Updated `should skip v2 initialization and Brandalf in v1 onboarding mode` to assert that `submitPromptGenerationJob` IS called once for v1 with the expected brand/site/audience, while `submitJob` (Brandalf) remains uncalled.
- [x] All 81 tests in `test/controllers/llmo/llmo-onboarding.test.js` pass locally.
- [x] ESLint clean on the changed files.
- [ ] Verify on dev: onboard a v1-mode site (`LLMO_ONBOARDING_DEFAULT_VERSION=v1` or pre-cutoff org) and confirm prompts land in the LLMO config without manual DRS triggering.


src/controllers/llmo/llmo-onboarding.js
```diff
@@ -1377,17 +1377,39 @@ export async function performLlmoOnboarding(params, context, say = () => {}) {
...
+            baseUrl: siteConfig.getFetchConfig?.()?.overrideBaseURL || baseURL,
+            brandName: brandName.trim(),
+            audience,
+            siteId: site.getId(),
+            imsOrgId,
+          });
+          log.info(`Started DRS prompt generation: job=${drsJob.job_id}`);
+          say(`:robot_face: Started DRS prompt generation job: ${drsJob.job_id}`);
+        } else {
+          log.debug('DRS client not configured, skipping prompt generation');
```
dzehnder, line 1411: **Critical: missing v1 branch coverage.** This `else` branch (and the `catch` block below at lines 1404-1406) are the uncovered lines causing CI to fail on the 100% threshold. The existing `'should skip DRS prompt generation when DRS client is not configured'` test runs the v2 path (it omits `LLMO_ONBOARDING_DEFAULT_VERSION: 'v1'` from `context.env`), so it covers the Brandalf branch, not this one. Need three new v1-mode tests: (1) `isConfigured()` -> `false` asserting `submitPromptGenerationJob` is NOT called and `mockLog.debug` fires with `'DRS client not configured, skipping prompt generation'`, (2) `submitPromptGenerationJob` rejects asserting onboarding still resolves and the `:warning:` `say` is emitted, (3) `getBrandProfile` returns `null` so the fallback audience template at line 1389 is exercised.

src/controllers/llmo/llmo-onboarding.js
```diff
@@ -1377,17 +1377,39 @@ export async function performLlmoOnboarding(params, context, say = () => {}) {
...
+            || `General consumers interested in ${brandName} products and services`;
+
+          const drsJob = await drsClient.submitPromptGenerationJob({
+            baseUrl: siteConfig.getFetchConfig?.()?.overrideBaseURL || baseURL,
+            brandName: brandName.trim(),
+            audience,
+            siteId: site.getId(),
+            imsOrgId,
+          });
+          log.info(`Started DRS prompt generation: job=${drsJob.job_id}`);
```
dzehnder, line 1408: **Important: unguarded `drsJob.job_id` access.** If `submitPromptGenerationJob` resolves with `undefined`, `null`, or an object missing `job_id`, this logs `job=undefined` and the next line posts `:robot_face: Started DRS prompt generation job: undefined` to Slack - the user-facing message looks like success when it may not be. Either validate the response (`if (!drsJob?.job_id) throw new Error('DRS returned no job_id')` so it falls into the catch and operators see the `:warning:`), or gate the success log/say behind `if (drsJob?.job_id)`.

src/controllers/llmo/llmo-onboarding.js
```diff
@@ -1377,17 +1377,39 @@ export async function performLlmoOnboarding(params, context, say = () => {}) {
...
+      // automatically. Submit it directly here so v1 onboardings still get
+      // prompts written to the legacy LLMO config (LLMO-4534).
+      try {
+        const drsClient = DrsClient.createFrom(context);
+        if (drsClient.isConfigured()) {
+          const brandProfile = siteConfig.getBrandProfile?.();
+          const audience = brandProfile?.main_profile?.target_audience
+            || `General consumers interested in ${brandName} products and services`;
+
+          const drsJob = await drsClient.submitPromptGenerationJob({
```
dzehnder, line 1398: **Important: cross-repo coupling is undocumented at the call site.** The PR description correctly notes that omitting `onboarding_mode` is what makes the audit-worker `drs-prompt-generation` handler take the legacy-write path. That contract lives across two repos and the code does not name it. If anyone threads `onboarding_mode` through the DRS client wrapper, or audit-worker flips its default branch, v1 prompts will silently land in the wrong place. Add a one-line comment at this call naming the audit-worker file/function this depends on so the coupling is greppable from either side.

dzehnder (changes requested):
Hey @byteclimber,

## Strengths

- Targeted, minimal fix that surgically restores the v1 prompt-generation behavior lost in LLMO-4258. The comment naming the ticket at `src/controllers/llmo/llmo-onboarding.js:1381-1383` will save the next reader a git-blame round trip.
- Mirrors the structure of the adjacent v2 Brandalf block at `src/controllers/llmo/llmo-onboarding.js:1356-1377` (same `DrsClient.createFrom` -> `isConfigured` -> try/catch shape, same `say(':warning:...')` failure message). The symmetry is the right call.
- Onboarding still completes when DRS submission fails, preserving the principle that prompt generation is a best-effort enrichment and not a blocking gate. Matches the v2 Brandalf contract.
- New v1 test asserts the exact arg shape passed to `submitPromptGenerationJob` (`brandName`, `siteId`, `imsOrgId`, `audience`), which would catch a regression in any of those four field names.
- Removes the now-stale "Prompt generation is deferred to DRS..." comment instead of leaving a contradictory note next to the new direct-submit code. Good change discipline.

## Issues

### Critical (Must Fix)

**`src/controllers/llmo/llmo-onboarding.js` and `test/controllers/llmo/llmo-onboarding.test.js` - missing v1 branch coverage (CI is failing).** CI fails on the global 100% coverage threshold; uncovered lines are `1402` (the v1 `else { log.debug(...) }`) and `1404-1406` (the new v1 `catch (drsError)` block). The pre-existing test `'should skip DRS prompt generation when DRS client is not configured'` does not exercise the new v1 branch - it omits `LLMO_ONBOARDING_DEFAULT_VERSION: 'v1'` from `context.env`, so it runs the v2 path and only covers the Brandalf "not configured" branch. Add three v1-mode tests cloned from the new "v1 onboarding triggers DRS prompt generation" test:

1. v1 + `isConfigured()` returns `false` -> assert `submitPromptGenerationJob` is NOT called and `mockLog.debug` is called with `'DRS client not configured, skipping prompt generation'`.
2. v1 + `submitPromptGenerationJob` rejects -> assert onboarding still resolves, `mockLog.error` is called with a message containing `'Failed to start DRS prompt generation'`, and the `:warning:` `say` is emitted. This locks in the silent-failure contract.
3. v1 + `siteConfig.getBrandProfile` returns `null` -> assert the fallback audience `'General consumers interested in Test Brand products and services'` is sent. Without this, the `||` fallback at line 1389 ships uncovered even after items 1 and 2.

### Important (Should Fix)

**`src/controllers/llmo/llmo-onboarding.js:1398-1399` - unguarded `drsJob.job_id` access.** If `submitPromptGenerationJob` ever resolves with `undefined`, `null`, or an object missing `job_id`, line 1398 logs `job=undefined` and line 1399 emits `:robot_face: Started DRS prompt generation job: undefined` to Slack. The user-visible message looks like a successful start when it might not be. Either validate the response (`if (!drsJob?.job_id) throw new Error('DRS returned no job_id')`, so it falls into the catch and the operator sees the `:warning:`), or at minimum gate the success log/say behind `if (drsJob?.job_id)`.

**`src/controllers/llmo/llmo-onboarding.js:1391` - cross-repo contract via metadata absence is undocumented at the call site.** The PR description correctly notes that omitting `onboarding_mode` is what makes the audit-worker `drs-prompt-generation` handler take the legacy-write path. That coupling lives across two repos and the code does not name it. If anyone later threads `onboarding_mode` through the DRS client wrapper, or audit-worker flips its default branch, v1 prompts will silently land in the wrong place with no test catching it. Add a one-line comment at this call site naming the audit-worker file/function this depends on, so the coupling is greppable from either side.

### Minor (Nice to Have)

**`src/controllers/llmo/llmo-onboarding.js:1389,1393` - `brandName.trim()` is inconsistent.** The fallback audience template uses raw `brandName` (`General consumers interested in ${brandName}...`), but the DRS payload at line 1393 uses `brandName.trim()`. If a caller ever passes `' Acme  '`, the audience string contains the untrimmed value while DRS gets the trimmed one. Compute `const trimmedBrand = brandName.trim()` once at the top of the block and use it in both places.

**`src/controllers/llmo/llmo-onboarding.js:1389` - fallback audience is English-only.** `'General consumers interested in ${brandName} products and services'` will be sent to DRS even for sites onboarded for non-English regions. The v2 path takes audience from `brand_profile.main_profile.target_audience`, which is presumably already locale-aware; the v1 fallback skips that. Either lift the default to a named constant near other LLMO defaults, or accept it as a known v1 limitation with a `// LLMO-XXXX: v1 fallback audience is English-only` breadcrumb. Future maintainers will hit this when prompts come back in English for a `de` site.

## Recommendations

- Include `site.getId()` in the `:warning:` Slack message at line 1405 so operators don't have to page back through context to find which site needs the manual job. The v2 Brandalf failure message has the same gap; fixing both is a one-liner each.
- Consider extracting the v1 DRS submit into a `triggerV1PromptGenerationJob` helper alongside the existing `triggerBrandalfOnboardingJob`. The v2 branch already extracted its DRS call into a helper; mirroring that for v1 would let `performLlmoOnboarding` read as "v2 helper / v1 helper" at the same level of abstraction. Not blocking, but the asymmetry will pull readers' eyes.

- The architecture now has two prompt-generation entry points (v1 direct, v2 Brandalf-mediated) with no documented sunset. Worth a follow-up ticket defining whether v1 onboarding is a permanent supported mode or a remediation bridge with a removal date. Either decision is fine; the ambiguity is what creates long-term maintenance cost.
- Document the `onboarding_mode`-absence -> legacy-write-path coupling between this repo and the audit-worker `drs-prompt-generation` handler somewhere both teams will see it (e.g. a short note in `mysticat-architecture/`).

## Assessment

Reasoning: The fix itself is correct, narrowly scoped, and structurally consistent with the v2 sibling pattern. Security found nothing of substance. But CI is red on coverage and the uncovered lines are the exact failure-mode branches (DRS unconfigured, DRS throws) that motivated this PR - those need pinning tests, not because of the threshold but because the catch block IS the contract. The unguarded `drsJob.job_id` and the undocumented cross-repo metadata coupling are smaller but worth addressing in the same revision.

Note: CI checks are currently failing - this should be resolved before merge.

## Next Steps

1. Address the critical coverage gap with the three v1-mode tests above (this also fixes CI).
2. Decide on the two important items: guard the `drsJob.job_id` access, and add the cross-repo coupling comment.
3. The minor items (trim consistency, English-only fallback) and recommendations are optional.

dzehnder (approved):
Hey @byteclimber,

Thanks for the thorough turnaround on the previous round.

## Strengths

Previously flagged issues now addressed at HEAD `a98f74a1`:

- **Coverage gap (Critical)**: All three v1 branches now have pinning tests. The new `'should use the English fallback audience when brand profile is missing in v1 mode'` at `test/controllers/llmo/llmo-onboarding.test.js` exercises `getBrandProfile() === null` AND the trim consistency in a single assertion (passes `'  Test Brand  '`, asserts both `brandName: 'Test Brand'` and the un-padded audience string). The two earlier branch tests from commit `e78ff482` cover the unconfigured-DRS and DRS-rejection paths. CI is now green at 100% coverage.
- **Unguarded `drsJob.job_id` (Important)**: The new `if (!drsJob?.job_id) throw new Error('DRS submitPromptGenerationJob returned no job_id')` guard at `src/controllers/llmo/llmo-onboarding.js:1405-1407` converts a silent partial-success (`job=undefined` log + `:robot_face: ... undefined` Slack) into a real failure that flows through the existing catch. The new pinning test asserts both that the warning is emitted AND that the success log/say with `undefined` are NOT emitted - that second half is the assertion that actually proves the guard works.
- **Cross-repo coupling (Important)**: The new comment block at `src/controllers/llmo/llmo-onboarding.js:1394-1397` makes the audit-worker `drs-prompt-generation` legacy-write-path dependency explicit and warns "Do NOT pass `onboarding_mode` here". This is the right shape for documenting a coupling that has no other in-tree home: the warning lives where the next engineer is most likely to violate it.
- **`brandName.trim()` inconsistency (Minor)**: `trimmedBrand` extracted once at line 1387 and used in both the audience template and the DRS payload. The fallback-audience test pins this with whitespace input.
- **English-only fallback (Minor)**: `LLMO-4534: v1 fallback audience is English-only` breadcrumb at lines 1389-1390 - future i18n work has a ticket to grep for.
- **Operator-facing Slack message (Recommendation)**: `:warning:` at line 1416 now includes `site.getId()`, materially improving triage when multiple onboardings run concurrently.

## Recommendations

The earlier suggestion to extract a `triggerV1PromptGenerationJob` helper alongside the existing `triggerBrandalfOnboardingJob` was not picked up. The v1 block is now ~37 lines inline inside `performLlmoOnboarding`. Defensible to leave as-is since v1 is on a deprecation path; mentioning once and deferring.

## Assessment

Ready to merge? **Yes.**

Reasoning: Every previously flagged issue is addressed in the diff at HEAD, the new tests assert exactly what they claim (verified the stub wiring and `.include` / `.not.calledWith` semantics), CI is green at 100% coverage, and security found nothing of substance in either round. Clean fix-up commit.

PR #2274, opened 2026-04-27
fix: cache control for llmo endpoint | LLMO-4480

jjenscodee:
## Summary

Adds `Cache-Control: private, max-age=7200` to read-only GET handlers in `src/controllers/llmo/` whose underlying data the user can't mutate through the UI, so the browser serves them from its private HTTP cache for 2 hours on reload. Mutations (POST/PATCH/PUT/DELETE) and read endpoints over user-editable data stay on plain `ok` — never cached.

- **Wrap `ok()`**: new `src/support/cached-response.js` exports `cachedOk(body, headers)` — same signature as `ok`, defaults set in the background, fully overridable via the second arg.
- **One cache duration**: single 2h tier replaces the previous TAXONOMY/METRICS split. `brand-presence-cache-policy.js` deleted.
- **Wider scope**: applied to read endpoints across `src/controllers/llmo/` — brand-presence, agentic-traffic (site + global), url-inspector, rationale, brand-claims, sheet-data proxies, regions, and demo fixtures.

## Scoping rule for `cachedOk`

We apply `cachedOk` only to read endpoints whose underlying data the user **cannot mutate through the UI**. Anything user-editable stays on plain `ok` so a save is never masked by a 2-hour stale cache. The big perceived-perf wins (heavy aggregate analytics reads) all stay cached; user-editable config/state always goes to network.

### Endpoints that do **not** use `cachedOk` (intentional)

| File | Handler | Reason |
|---|---|---|
| `llmo.js` | `getLlmoConfig` | mutated by `updateLlmoConfig` |
| `llmo.js` | `getLlmoQuestions` | mutated by `addLlmoQuestion` / `removeLlmoQuestion` / `patchLlmoQuestion` |
| `llmo.js` | `getLlmoCustomerIntent` | mutated by `addLlmoCustomerIntent` / `removeLlmoCustomerIntent` / `patchLlmoCustomerIntent` |
| `llmo.js` | `getEdgeConfig` | mutated by `createOrUpdateEdgeConfig` |
| `llmo.js` | `getStrategy` | mutated by `saveStrategy` |
| `llmo.js` | `checkEdgeOptimizeStatus` | derived from edge config which users edit |
| `opportunities/llmo-opportunities-controller.js` | `getOpportunityCount` | affected by `markOpportunitiesReviewed` |
| `opportunities/llmo-opportunities-controller.js` | `getBrandOpportunities` | same |

All POST/PATCH/PUT/DELETE handlers also stay on plain `ok` — never cache mutations.

### Endpoints that **do** use `cachedOk`

- All `brand-presence/*` (filter-dimensions, weeks, stats, sentiment-overview, competitor-summary, share-of-voice, topics, topic-prompts, search, topic-detail, prompt-detail, execution-sources, sentiment-movers, market-tracking-trends) — computed by backend jobs.
- All site-scoped `agentic-traffic/*` (kpis, kpis-trend, by-region, by-category, by-page-type, by-status, by-user-agent, by-url, filter-dimensions, weeks) — computed by ingestion.
- `getAgenticTrafficGlobal` — admin-only writes via `postAgenticTrafficGlobal` are infrequent; cached per owner request.
- All `url-inspector/*` — computed by ingestion.
- `getRegions` — static reference data.
- `getLlmoRationale`, `getBrandClaims`, `getDemoBrandPresence`, `getDemoRecommendations` — generated content / fixtures.
- `getLlmoSheetData`, `getLlmoGlobalSheetData`, `queryFiles` — external / S3 proxies where the URL path encodes the data version, so a config change naturally produces a different cache key.

### Why no `Vary: Authorization`

Initially included on the reviewer's suggestion, then dropped after testing on stage: the SpaceCat session JWT is regenerated on every page reload (different signature for the same user), so adding it to `Vary` changes the cache key on every reload and defeats caching entirely. The browser's private HTTP cache is per-profile, and `Cache-Control: private` already keeps shared caches out, so cross-user contamination on a single profile remains unlikely.

- [ ] make sure to link the related issues in this description. Or if there's no issue created, make sure you describe here the problem you're solving.

- [ ] make sure you add a "Not implemented yet" note the endpoint description, if the implementation is not ready yet. Ideally, return a 501 status code with a message explaining the feature is not implemented yet.

## Related Issues
https://jira.corp.adobe.com/browse/LLMO-4480

## Screenshots

Screenshot 2026-04-27 at 13 37 24

https://github.com/user-attachments/assets/625a27e5-c2e6-46f0-b57b-7ca108712506

Thanks for contributing!


igor-grubic:
Thanks @jjenscodee , this is a nice finding.
Some notes:
1. can we add `vary:"authorization"` so we avoid cross auth caching
2. can we avoid weird nomenclature that is not product related. Instead, let's add one cache, 2hs, for all calls for now? Unless we document reasoning between the chosen different times
3. can we wrap the (ok) function in a different wrapper, so we don't have to add these around the api? We could call it `cachedOk` or something along those lines. The function should stay the same as `ok` but add the cache headers in the background (overridable when possible)
4. can we extend to more than just bp? This should be touching everything under `llmo` folder almost (altho i would say even further than that, but that requires conversations). But for now we could make it brandalf only if easier.

igor-grubic (approved):

PR #2278, opened 2026-04-28
docs(s2s): add readAll capability solution design

ravverma:
## Summary

- Adds solution design for S2S `readAll` capability (`docs/s2s/READALL_CAPABILITY_DESIGN.md`)
- Proposes a new `readAll` action type in `Consumer.CAPABILITIES` (shared-data-access) to make `site:readAll` / `organization:readAll` valid capability strings
- Remaps `GET /sites` and `GET /organizations` to require `readAll` instead of `read` in `required-capabilities.js`, keeping tenant-scoped variants (`/sites/:siteId`) on `site:read`
- Introduces `async hasS2SCapability(capability)` on `AccessControlUtil` — uses `context.s2sConsumer` for identity only, then issues an independent DB fetch for capabilities (does not trust cached context state)
- Two-layer validation: s2sAuthWrapper (Layer 1) and explicit controller opt-in (Layer 2) are independently enforced — a new endpoint is admin-only by default unless it explicitly adds the controller check
- Tenant isolation is preserved: `readAll` grants list visibility only; all site/org-scoped operations retain existing `hasAccess()` gating

## Implementation scope (follow-up PRs)
- `spacecat-shared-data-access`: add `readAll` to `Consumer.CAPABILITIES`
- `spacecat-api-service`: `access-control-util.js`, `required-capabilities.js`, `sites.js`, `organizations.js`
Pushed. Add this to your PR description:

---

**Additional context — Consumer Discovery Flow**

Documents the bootstrap pattern for services with no prior knowledge of site/org IDs:

1. Mint an S2S token using the SpaceCat internal platform org (`908936ED5D35CC220A495CD4@AdobeOrg` prod, `8C6043F15F43B6390A49401A@AdobeOrg` dev/stage) — Consumer registered against this org with `site:readAll` / `organization:readAll`
2. Call `GET /sites` or `GET /organizations` to discover the full list
3. Mint a scoped token for the target customer org
4. All subsequent site/org-scoped operations use the scoped token through standard `hasAccess()` checks

The internal-org token is used only for enumeration — write and tenant-scoped operations require a customer-org-scoped token.


docs/s2s/READALL_CAPABILITY_DESIGN.md
```diff
@@ -0,0 +1,378 @@
...
+```
+GET /sites           → returns all sites (DTO, no sensitive internals)
+GET /organizations   → returns all organizations
+```
+
+The service now has the site IDs, base URLs, and IMS org IDs it needs to scope further operations.
+
+### Step 3 — Mint a scoped token for the target site / org
+
+With a known `imsOrgId` for the target customer, the service mints a scoped S2S token for that
```
solaris007: **Critical**: Step 3 says the service mints a scoped token for the target org, but does not specify which IMS credentials are used or what Consumer records must exist per customer org. If the same platform-level principal can mint for any discovered org, `site:readAll` is de facto admin-of-everything. If per-org provisioning is required, say so explicitly - the entire tenant isolation argument depends on this answer.

docs/s2s/READALL_CAPABILITY_DESIGN.md
```diff
@@ -0,0 +1,378 @@
...
+└──────────────────────────┬──────────────────────────────────────────┘
+                           │
+                           ▼
+              Fetch all sites / organizations
+              Map through DTO (no raw models exposed)
+              Return 200 OK
+```
+
+The wrapper check (Layer 1) and the controller check (Layer 2) are genuinely independent —
+Layer 2 derives consumer identity from `authInfo.getProfile()` (JWT claims) and issues its
```
solaris007: **Important**: This line says Layer 2 derives identity from `authInfo.getProfile()` and does not read from `context.s2sConsumer`. The `hasS2SCapability` code above does the opposite - it reads `context.s2sConsumer.getClientId()` and `getImsOrgId()` (Layer 1 state). Only capabilities are re-fetched. Pick one and align prose with code.

Also: the real value of Layer 2 is fail-closed-by-default (controllers without the gate stay admin-only), not revocation-race defense (microsecond window inside one Lambda invocation). Reframe the rationale.

docs/s2s/READALL_CAPABILITY_DESIGN.md
```diff
@@ -0,0 +1,378 @@
+# S2S `readAll` Capability — Solution Design
+
+## Problem Statement
+
+S2S consumers (e.g., dashboard services, opportunity creators) are increasingly built around a
+"run for every site in an org" pattern. These services have no prior knowledge of site base URLs
+or IMS org IDs — they need to discover them first. The current admin-gated endpoints (`GET /sites`,
+`GET /organizations`) are inaccessible to S2S consumers because they require `hasAdminAccess()`,
```
solaris007: **Important**: The problem statement says consumers run "for every site in an org" - that is tenant-scoped enumeration, not a global cross-tenant list. But the design proposes global `site:readAll`. Why does `GET /organizations/:orgId/sites` (org-scoped, already exists) not satisfy the stated use case? If global enumeration is genuinely needed, the security posture is different from "org-scoped discovery" and should be stated explicitly.

docs/s2s/READALL_CAPABILITY_DESIGN.md
```diff
@@ -0,0 +1,378 @@
...
+  `organization:read`. Existing capabilities are unaffected.
+- This does not apply to any endpoint other than the specific admin-gated list endpoints that
+  explicitly opt in.
+- Entitlement, enrollment, and delegation checks are unchanged for all other endpoints.
+
+---
+
+## Proposed Solution
+
+### New Capability Action: `readAll`
```
solaris007, line 67: **Important**: Adding `readAll` to `Consumer.CAPABILITIES` is a multiplicative namespace expansion. `ConsumerCollection.#getValidCapabilities()` flat-maps `entityNames x CAPABILITIES`, so this immediately makes `audit:readAll`, `opportunity:readAll`, `apiKey:readAll`, etc. all schema-valid strings - even though only `site` and `organization` are intended targets.

Also: `readAll` conflates action with scope in the `entity:action` DSL (`read|write|delete` are verbs; `readAll` is verb+scope). Worth adding an "Alternatives Considered" note explaining why this is the right trade-off for v1 (vs `entity:list`, `entity:read:all`, or a separate scope field).

solaris007 (changes requested):
Hey @ravverma,

Good call putting the design in writing before code - this is easier to iterate on than a PR diff. The deny-by-default posture is preserved, the non-goals are sharp, and the tenant isolation analysis is correct for the per-entity path. A few things need tightening before this becomes the spec that follow-up PRs are reviewed against.

## Strengths

- **Non-Goals section is precise.** Explicitly calling out that `readAll` does not imply `read`, does not bypass tenant isolation, and does not apply beyond the opted-in endpoints. This is the kind of negative invariant a security-adjacent design should state.
- **Database as source of truth.** Both layers re-fetch the consumer record and re-validate `isRevoked()` and `getStatus()`. No trust placed on token claims for capability decisions.
- **Deny-by-default preserved.** A new admin-gated endpoint not added to `routeRequiredCapabilities` stays inaccessible to S2S consumers. The wrapper semantics are unchanged.
- **Tenant-scoped reads (`GET /sites/:siteId`) keep their existing `hasAccess(entity)` path** untouched. The new capability does not bleed into the per-resource authorization path.
- **What Does Not Change section** correctly enumerates everything that is unaffected (admin path, API key path, internal-routes exclusion, `hasAccess()` semantics).

## Issues

### Critical (Must Fix)

**1. Discovery flow Step 3 minting model is unspecified - this is the actual trust boundary.** (also posted inline)

Step 3 says: "the service mints a scoped S2S token for that org and uses it for all subsequent operations." It does not say which IMS S2S credentials the service presents at the auth-service `/auth/s2s/login` endpoint to obtain a token for an arbitrary customer org.

Two readings with very different security implications:

- **(a)** The service must hold a per-customer-org IMS TA with a matching Consumer record. Then discovering an org ID gives the service nothing it can act on without already being provisioned. Tenant isolation holds.
- **(b)** The service uses a single platform-level principal that auth-service trusts to mint for any customer org. Then `site:readAll` against the internal org is de facto equivalent to admin-of-everything: discover, mint scoped, act. The "Tenant Isolation Is Preserved" claim breaks down.

The entire security argument rests on this. Add a section that specifies: what credential is used at Step 3, what stops a service with internal-org discovery access from minting tokens against any discovered org, and what Consumer records must exist for Step 3+ operations.

### Important (Should Fix)

**2. Layer 2 identity source contradicts the prose, and the security rationale is oversold.** (also posted inline)

The "Two-Layer Validation" and "Database as Source of Truth" sections both state that Layer 2 derives identity independently: "Layer 2 derives consumer identity from `authInfo.getProfile()` (JWT claims) [...] It does not read from `context.s2sConsumer` or any other context state set by Layer 1."

The `hasS2SCapability` code does the opposite - it reads `context.s2sConsumer.getClientId()` and `context.s2sConsumer.getImsOrgId()` (Layer 1 state). Only capabilities are re-fetched.

Additionally, the "consumer revoked between the two checks" defense covers a microsecond window within one Lambda invocation - materially zero. The real value of Layer 2 is structural: a controller that forgets the `hasS2SCapability` gate stays admin-only by default because `hasAdminAccess()` returns false for S2S tokens. That is a genuine fail-closed property and should be the primary justification.

Fix: (a) Align prose and code. (b) Reframe the rationale around fail-closed-by-default, drop the revocation-race claim.

**3. Stated use case does not match the proposed scope.** (also posted inline)

The Problem Statement says consumers run "for every site in an org" - that is tenant-scoped enumeration. The design proposes global `site:readAll` across all tenants. There is no analysis of why `GET /organizations/:orgId/sites` (org-scoped, already exists) does not satisfy the use case.

If the real need is "I am a platform-level service that legitimately needs to enumerate every tenant," that is a meaningfully different security posture and should be stated explicitly with the concrete consumers and their justification.

Fix: Add a "Why not org-scoped list endpoints?" section.

**4. Pagination is unaddressed for the very endpoints being opened.**

`sites.js` already carries a TODO: "implement proper pagination or filtering to stay under AWS Lambda response size limits (6MB)." The pattern this doc enables - S2S consumers calling `GET /sites` programmatically on cron cycles - will hit this harder than the existing admin-only callers.

Fix: Add an "Operational Bounds" section. Either commit to pagination as a precondition, or explicitly defer with a known-limitation note and a follow-up reference.

**5. Data classification of the list payload is unanalyzed.**

A consumer with `site:readAll` can enumerate every customer's `baseUrl` and `imsOrgId` across all tenants. Site URLs and the fact that customer X is on SpaceCat are competitively meaningful. The doc treats list visibility as low-impact without classifying the actual DTO fields.

Fix: Add a "Data Classification" subsection naming the DTO fields exposed by `GET /sites` and an audit logging requirement for cross-tenant enumeration.

**6. `readAll` conflates action with scope, and the namespace expansion is multiplicative.** (also posted inline)

The current model is `entity:action` where actions are `read|write|delete`. `readAll` is not a new verb - it is `read` plus a scope qualifier. This sets a precedent for `writeAll`, `deleteAll`, etc.

Additionally, `ConsumerCollection.#getValidCapabilities()` flat-maps `entityNames x Consumer.CAPABILITIES`. The moment `readAll` lands, `audit:readAll`, `opportunity:readAll`, `apiKey:readAll`, etc. all become schema-valid strings. Protection depends entirely on route mapping and controller opt-in, never on the schema.

Fix: Add a "Capability Model Trade-offs" section acknowledging the conflation and namespace expansion. Name alternatives considered (`entity:list`, `entity:read:all`, separate scope field) and explain why `readAll` is the right call for v1.

**7. JWT `aud` claim validation is not documented as a prerequisite.**

A known finding on this S2S framework is that `aud` is set at issuance but not verified (RFC 7519 S4.1.3, RFC 8725 S3.11). This design depends on correct token validation. Add a "Prerequisites" section listing the JWT validation contract assumed: `iss`, `aud`, `exp`, `nbf`, `is_s2s_consumer`, `client_id`, `org`.

### Minor (Nice to Have)

**8. Capability string duplication between route map and controller.**

Layer 1: `'GET /sites': 'site:readAll'` in `required-capabilities.js`. Layer 2: `hasS2SCapability('site:readAll')` in `sites.js`. Two literal strings in two files that must stay synchronized. Recommend a coverage test or importing the string from a single source.

**9. Error message "Only admins can view all sites" is misleading post-readAll.**

An S2S consumer whose capability check fails gets a message saying they're not an admin. Recommend a generic `forbidden('Forbidden')` externally with internal logging that distinguishes the denial paths.

**10. No testing strategy.**

A two-paragraph "Test Strategy" section with the key matrix (unit tests for `hasS2SCapability` states, integration tests for the route gate, negative test that `site:read` cannot reach `GET /sites`) would catch implementation gaps in follow-up PRs.

## Recommendations

- **Document cross-repo rollout order.** `spacecat-shared-data-access` (adds `readAll` to CAPABILITIES) must merge and release before `spacecat-api-service` (route remap + controller gate). Reversing the order means the route maps to a capability that consumer creation rejects as invalid.
- **Cap the blast radius before opening the door.** Before granting `readAll` to any S2S consumer, the list endpoints should support paginated responses with a hard upper bound.
- **Capture the policy-engine off-ramp.** If 5+ entities adopt the `readAll` pattern, the right next move is a policy engine (Cedar, OpenFGA, OPA), not more hand-coded controller gates. Naming the threshold now gives future maintainers a clear signal.

## Assessment

Not ready to merge. The discovery-flow Step 3 minting model (Critical 1) and the Layer 2 narrative-vs-code contradiction (Important 2) need to be resolved in the doc before the implementation PR lands. The other Important findings can be addressed alongside - they are all addressable by editing this single design doc.

## Next Steps

1. Specify the Step 3 trust boundary - what credential mints the scoped token, and what Consumer records must exist per customer org.
2. Align the Layer 2 prose with the actual code, and reframe the security rationale around the fail-closed-by-default property.
3. Add the missing sections: scope justification, pagination/operational bounds, data classification, capability model trade-offs, prerequisites.
4. Minor items and recommendations are discretionary but would strengthen the spec.

ravverma:
Thanks @solaris007 — addressed all comments in `7b42471`. Mapping below.

## Critical
**1. Step 3 minting model** → New `Trust Boundary at Discovery Step 3` section in Security Analysis. Specifies that the per-org Consumer record (`Consumer.findByClientIdAndImsOrgId(clientId, customerOrgId)`) at Layer 1 of the wrapper is the load-bearing isolation invariant: discovery via the internal-org Consumer confers no operational access; per-customer-org Consumer rows must exist separately for tenant operations to succeed, regardless of what the auth-service was willing to mint. Step 3 of the Discovery Flow rewritten to make this explicit.

## Important
**2. Layer 2 prose vs code + rationale** → Prose now matches code: identity comes from `context.s2sConsumer` (`getClientId()` / `getImsOrgId()`), only capabilities are re-fetched. Removed the revocation-race claim from the primary justification. New framing: Layer 2's primary value is fail-closed-by-default — `hasAdminAccess()` returns false for S2S tokens, so a controller without an explicit `hasS2SCapability` branch stays admin-only. The capability re-fetch is now described as structural belt-and-suspenders, not the primary justification.

**3. Use case mismatch** → Added explicit `> Why not org-scoped list endpoints?` callout in the Problem Statement: `GET /organizations/:orgId/sites` already exists for the org-known case; `readAll` is for platform-level services that need cross-tenant enumeration *because they don't yet know which orgs exist*. This is platform-wide enumeration by design.

**4. Pagination** → New `Operational Bounds` section. Pagination is a precondition, not a follow-up — the route remap is gated behind cursor-based pagination shipping first on `GET /sites` and `GET /organizations`. Targets included: 4 MB max payload, p95 < 500ms, 60 rpm initial rate limit, 90-day audit log retention. Reflected in `Cross-Repo Rollout Order` as step 2 (precondition for step 4).

**5. Data classification** → New `Data Classification` section enumerating the actual `SiteDto` and `OrganizationDto` fields exposed (`baseURL`, `imsOrgId`, org names, etc.), with three implications: `readAll` grants approved at admin-grade level, every successful Layer 2 pass is audit-logged, DTO field additions must consider the `readAll` exposure surface.

**6. Capability model trade-offs** → New `Capability Model Trade-offs` section. Explicitly acknowledges `readAll` conflates verb + scope and that `ConsumerCollection.#getValidCapabilities()` flat-mapping makes every `<entity>:readAll` schema-valid. Alternatives table covers `entity:list`, `entity:read:all`, separate scope field, with rationale for each not being chosen for v1. Documents the `5+ entities` threshold as the trigger to move to a policy engine (Cedar / OpenFGA / OPA) — also reflected in `Future Extension`.

**7. JWT prerequisites** → New `Prerequisites` section listing the JWT validation contract: `iss`, `aud`, `exp`, `nbf`, `iat`, `is_s2s_consumer`, `client_id`, `org`. The known `aud` verification gap (RFC 7519 §4.1.3 / RFC 8725 §3.11) is called out explicitly as load-bearing context.

## Minor
**8. Capability string duplication** → New `src/routes/capability-constants.js` exporting `CAP_SITE_READ_ALL` and `CAP_ORG_READ_ALL`. Both `required-capabilities.js` and the controllers import from there. Coverage test asserts that every controller-checked capability is in the route map and vice versa, so drift cannot be silent.

**9. Misleading error message** → Controller pattern updated to return generic `forbidden('Forbidden')` externally with structured internal log distinguishing the denial paths (`admin=false s2sCap=false`).

**10. Test strategy** → New `Test Strategy` section with explicit unit / integration / negative-coverage matrices: `hasS2SCapability` state matrix, controller-gate matrix, IT scenarios (Consumer absent → 403, etc.), and a fail-closed-by-default coverage assertion.

## Recommendations
- **Cross-repo rollout order** documented in new `Cross-Repo Rollout Order` section: `spacecat-shared-data-access` (CAPABILITIES) → pagination on list endpoints → `spacecat-api-service` picks up new shared version → final route remap PR.
- **Pagination precondition** baked into `Operational Bounds` and rollout order.
- **Policy-engine off-ramp at 5+ entities** named in both `Capability Model Trade-offs` and `Future Extension`.

Ready for re-review.

solaris007 (approved):
Hey @ravverma,

Re-reviewed at `7b42471`. All 10 prior findings are substantively addressed - well done on the thorough follow-up. The doc has gone from a focused but incomplete design to a spec that implementation PRs can be reviewed against.

## Strengths (previously flagged issues now addressed)

- **Critical: Step 3 trust boundary** - new "Trust Boundary at Discovery Step 3" section (lines 252-284) names the per-`(clientId, orgId)` Consumer row as the load-bearing invariant. The honest callout that weakening the wrapper's Consumer lookup would invalidate the analysis is exactly right.
- **Layer 2 prose-vs-code** - aligned. Identity from `context.s2sConsumer`, only capabilities re-fetched. Fail-closed-by-default is the primary justification; revocation-race demoted to belt-and-suspenders.
- **Use case scope** - "Why not org-scoped list endpoints?" callout justifies platform-wide enumeration.
- **Pagination** - binding precondition in the rollout order, not a follow-up.
- **Data Classification** - DTO fields enumerated, audit-logging requirement specified.
- **Capability Model Trade-offs** - conflation acknowledged, alternatives tabled, 5+ entity policy-engine threshold named.
- **Prerequisites** - JWT contract listed with `aud` gap called out and RFC references.
- **Capability constants centralized**, **generic Forbidden + structured log**, **Test Strategy** - all addressed.

## Issues

### Minor (Nice to Have)

**1. Prerequisites table could be tightened for BCP completeness.** Missing `alg` header allowlist (RFC 8725 S3.1-3.2, the canonical defense against algorithm-confusion attacks), `sub` claim handling, and a replay-defense note (either `jti` or rationale for relying on short `exp`). Also: line 50 cites RFC 8725 S3.11 for `aud` handling, but the correct section is S3.9 ("Use and Validate Audience"); S3.11 is "Use Explicit Typing." The `aud` gap could also benefit from explicitly naming the compensating control (SpaceCat-specific `is_s2s_consumer` claim + Consumer table lookup make cross-audience reuse impractical).

**2. Pagination spec has two gaps for implementers.** The cursor scheme (opaque base64 vs last-id vs offset) is unspecified - one sentence is enough. Also: `limit=500` vs the 4MB hard cap - what happens when the requested page exceeds the payload limit? (Truncate + early `nextCursor`? Reject? Shrink silently?) Spelling out the precedence rule prevents an implementation debate.

**3. Operational targets table mixes SLOs and budgets without baselines.** 4MB traces to the Lambda 6MB limit. 60rpm is honestly labeled "revisit after 30 days." But p95 < 500ms and 90-day audit retention lack provenance - one-line citations or "proposed; revisit" qualifiers would close the loop. For cross-tenant enumeration logs specifically, 90 days is low relative to SOC 2 / PCI norms (1 year typical, 3 months immediately accessible) - worth a conscious decision documented in the table.

**4. Audit log fields and governance.** Missing `imsOrgId` of the Consumer's binding (internal-org vs customer-org) and HTTP method+path in the audit log field list - both are operationally critical for detecting unexpected cross-tenant enumeration. Also: the Trust Boundary argument hinges on per-org Consumer provisioning discipline, but the doc doesn't say who can create these rows or whether internal-org `readAll` grants and per-customer-org operational grants require separate approvals. One sentence each.

## Recommendations

- Fix the RFC 8725 section reference (S3.11 -> S3.9).
- Add `imsOrgId` and route to the audit log fields.
- State the pagination cursor scheme and limit-vs-cap precedence rule.
- Add one sentence on Consumer-row creation governance.

## Assessment

Ready to merge. All substantive prior findings are resolved with concrete content. The 4 Minor items are documentation polish - addressable in this PR or a fast follow-up. None blocks implementation.

Note: 1 of 2 CI build runs has a flaky `it-postgres` failure unrelated to the doc-only diff.

PR #2255, opened 2026-04-22
docs: GitHub webhook handler spec (POST /webhooks/github)

szergen:
## Summary

- Spec for the `POST /webhooks/github` HMAC-authenticated webhook handler for the Mysticat GitHub Service
- Single PR, three phases: OpenAPI contract (Phase 1), on-demand handler implementation (Phase 2), config-aware handler (Phase 3, separate PR)
- Custom `GitHubWebhookHmacHandler` extending `AbstractHandler`, plugged into existing `authHandlers` array
- Trigger rules from webhook payload only (Phase 2), per-repo config reading deferred to Phase 3

## Related

- **Jira:** [SITES-42733](https://jira.corp.adobe.com/browse/SITES-42733)
- **Architecture spec:** [review-orchestrator.md](https://github.com/adobe/mysticat-architecture/blob/main/platform/ops/review-orchestrator.md)
- **Infrastructure spec:** [review-orchestrator-infrastructure.md](https://github.com/adobe/mysticat-architecture/blob/main/platform/ops/review-orchestrator-infrastructure.md)
- **Infrastructure PR:** adobe/spacecat-infrastructure#469
- **Infrastructure wiring plan:** adobe/spacecat-infrastructure#473

## Resolved decision: per-repo configuration ownership

The handler owns per-repo config reading (`.github/mysticat.yml`) with a two-tier cache (in-Lambda memory + DynamoDB), invalidated on `push` events that touch the config file, plus a 10-30 minute TTL safety net.

This is deferred to **Phase 3** (separate PR) to keep Phase 2 safe and shippable:
- **Phase 2 (this PR):** On-demand triggers only (`review_requested`, `labeled`). Hardcoded defaults. No GitHub API calls. No auto-triggers.
- **Phase 3 (separate PR):** Installation-token minting, two-tier config cache, auto-triggers, kill switch, skill routing, `workspace_repos` override.

Operational failure modes for Phase 3 (cache stampede, token caching, YAML parse errors) are documented in the spec for the Phase 3 spec author.

## Deduplication divergence note

The parent architecture spec still describes web-tier dedup via Check Run query. The infrastructure spec explicitly removed this (Amendment item 4) in favor of worker-level DynamoDB conditional write, citing TOCTOU race. Parent spec reconciliation is tracked as a follow-up.

## Test plan

- [ ] Review OpenAPI contract (Phase 1)
- [ ] Validate HMAC auth handler pattern against `AbstractHandler` API
[...]

PR #2281, opened 2026-04-28
feat: GitHub webhook handler (POST /webhooks/github)

szergen:
## Summary

- Adds OpenAPI contract for `POST /webhooks/github` — HMAC-authenticated endpoint for the Mysticat GitHub Service
- **Phase 1 (this push):** OpenAPI contract ready for review
- **Phase 2 (after approval):** Implementation (HMAC auth handler, trigger rules, controller, tests)

## Changes (Phase 1)

- New `docs/openapi/webhooks-api.yaml` — endpoint definition under the existing `hooks` tag
- New schemas in `docs/openapi/schemas.yaml` — `WebhookAccepted`, `WebhookUnauthorized`
- Path reference in `docs/openapi/api.yaml` — `/webhooks/github`
- Regenerated `docs/index.html`
- Implementation plan at `docs/plans/2026-04-28-github-webhook-handler-impl.md`

## Related

- **Jira:** [SITES-42733](https://jira.corp.adobe.com/browse/SITES-42733)
- **Spec:** `docs/specs/2026-04-22-github-webhook-handler.md` (merged in #2255)
- **Architecture spec:** [review-orchestrator.md](https://github.com/adobe/mysticat-architecture/blob/main/platform/ops/review-orchestrator.md)
- **Infrastructure spec:** [review-orchestrator-infrastructure.md](https://github.com/adobe/mysticat-architecture/blob/main/platform/ops/review-orchestrator-infrastructure.md)
- **Infrastructure PR:** adobe/spacecat-infrastructure#469
- **Infrastructure wiring plan:** adobe/spacecat-infrastructure#473

## Open question: per-repo configuration ownership

The handler owns per-repo config reading (`.github/mysticat.yml`) with a two-tier cache, but this is deferred to **Phase 3** (separate PR). Phase 2 uses hardcoded defaults only.

- **Phase 2 (this PR, after Phase 1 approval):** On-demand triggers only (`review_requested`, `labeled`). No GitHub API calls. No auto-triggers.
- **Phase 3 (separate PR):** Installation-token minting, two-tier config cache, auto-triggers, kill switch, skill routing.

## Test plan

- [ ] Review OpenAPI contract for correctness
- [ ] Confirm endpoint uses existing `hooks` tag (no new tag)
- [ ] `npm run docs:lint` passes
- [ ] `npm run docs:build` succeeds
- [ ] (Phase 2) `npm test` passes
- [ ] (Phase 2) `npm run lint` passes


danieljchuser (changes requested):
Hey @szergen,

Good work on the OpenAPI contract and the implementation plan. The Phase 1 contract is clean and follows all the feedback from the spec review rounds (correct tag reuse, delivery_id propagation, field-specific 400 messages, etc.). The Phase 2 implementation plan is thorough with proper TDD sequencing. One critical cross-repo contract issue needs resolution before Phase 2 implementation starts.

## Strengths

- Reuses existing `hooks` tag per solaris007's spec review feedback (webhooks-api.yaml:3)
- Comprehensive TDD plan with tests-first for all three modules (trigger rules, HMAC handler, controller)
- Correctly uses `MYSTICAT_GITHUB_JOBS_QUEUE_URL` (renamed per spec review round 2)
- HMAC handler correctly extends `AbstractHandler` and slots into the `authHandlers` array - no bypass list, no separate wrap chain
- `SIGNATURE_PATTERN` regex pre-check before `timingSafeEqual` prevents the length-mismatch throw
- Test plan includes the raw-vs-reserialized body proof (Task 8, test "rejects signature computed over JSON.stringify")
- `delivery_id` propagated to job payload per spec minor finding 7
- `getSkipReason` extracted as pure function in its own module for testability

## Issues

### Critical (Must Fix)

**1. Job payload will be rejected by the Dispatcher Lambda - 5 fields are not in the allowed-keys whitelist.** (implementation plan, Task 9 controller, lines ~1130-1145)

The controller builds this payload:
```javascript
{ owner, repo, event_type, event_action, event_ref, installation_id,
  delivery_id, job_type, workspace_repos, retry_count }
```

The Dispatcher Lambda (deployed via spacecat-infrastructure PR 469, now in prod) enforces a strict key whitelist:
```python
_ALLOWED_KEYS = frozenset(("job_type", "owner", "repo", "event_ref", "retry_count", "source_task_arn"))
```

Five fields will be **rejected at ingestion**: `event_type`, `event_action`, `installation_id`, `delivery_id`, `workspace_repos`. The Dispatcher will raise a validation error, SQS will retry, and the message will land in the DLQ. **No job will ever reach the ECS worker.**

This was documented as a known cross-repo contract gap in the infrastructure PR's README (Open follow-ups section). It must be resolved before Phase 2 implementation. Two options:

- **(A) Expand the Dispatcher whitelist** (infra PR): Add the 5 new fields to `_ALLOWED_KEYS` and add appropriate validation (e.g., `workspace_repos` is an array of namespace/repo strings, `installation_id` is a numeric string). This is the cleaner approach - the Dispatcher should accept all fields the worker needs.
- **(B) Strip to required-only in the controller**: Only send `{job_type, owner, repo, event_ref, retry_count}` and have the worker reconstruct the missing context from the GitHub API using the installation token. This defeats the purpose of passing context through.

Recommend option A. The plan should include a companion infra PR or document the exact `_ALLOWED_KEYS` change needed.

### Important (Should Fix)

**2. Node.js version is wrong in the plan.** (implementation plan line 9)

The plan says "Node.js 22+" but this repo's `package.json` engines field requires `>=24 <25`. An implementer running on Node 22 will hit version check failures. Update to "Node.js 24.x".

**3. `security: []` in the OpenAPI contract is misleading.** (webhooks-api.yaml:68)

`security: []` in OpenAPI means "no security required" - but this endpoint IS authenticated via HMAC. This is technically necessary to override any global security scheme, but it makes the contract say the endpoint is unauthenticated when it's not. Consider modeling the HMAC as an `apiKey` security scheme:

```yaml
# In api.yaml components/securitySchemes:
GitHubWebhookSignature:
  type: apiKey
  in: header
  name: X-Hub-Signature-256
  description: HMAC-SHA256 signature of the request body
```

Then reference it: `security: [{ GitHubWebhookSignature: [] }]`. This makes the contract self-documenting about the actual auth mechanism. Not blocking, but a good practice for API consumers reading the spec.

### Minor (Nice to Have)

**4. 401 response doesn't follow repo conventions.** (webhooks-api.yaml:60-66)

Copilot noted that other endpoints reference `./responses.yaml#/401` for consistency. The custom `WebhookUnauthorized` schema has a different shape (just `{message}` vs whatever the shared 401 response uses). If the shared 401 response body is compatible, use the shared ref. If not, the custom response is fine - just document why it differs.

**5. `workspace_repos` is hardcoded with a cross-org repo.** (implementation plan Task 9, line ~1139)

The hardcoded array includes `Adobe-AEM-Sites/aem-sites-architecture` which is in a different GitHub org (`Adobe-AEM-Sites` vs `adobe`). The worker will need a GitHub App installation token that has access to both orgs, or the clone will fail. Worth a note in the plan so the Phase 2 implementer doesn't miss it.

## Recommendations

- Add a "Cross-repo contract" section to the implementation plan that documents the exact `_ALLOWED_KEYS` alignment needed with the Dispatcher, including which infra PR or commit will expand the whitelist. This prevents the "everything looks green locally but DLQs in dev" failure mode.
- Consider adding the `correlation_id` field to the job payload. The worker's DynamoDB idempotency table uses `pk = correlation_id` (per the infrastructure module). If the controller doesn't set it, the worker will need to derive it from `{owner}/{repo}/{event_ref}` or similar. Defining the canonical format in the controller avoids a silent contract gap.

## Assessment

Ready to merge? No, with fixes.

Reasoning: The Phase 1 OpenAPI contract is nearly ready (Node version and security scheme are quick fixes). The critical blocker is the Dispatcher whitelist mismatch in the Phase 2 plan - implementing the controller as written would result in 100% DLQ rate in dev. This needs a cross-repo coordination plan before Phase 2 proceeds.

## Next Steps

1. Address the Dispatcher `_ALLOWED_KEYS` alignment (critical) - decide on option A or B and document it.
2. Fix Node.js version in the plan.
3. Consider the `security` scheme modeling (important but not blocking Phase 1 merge).

szergen:
## Action Plan - PR #2281 (adobe/spacecat-api-service)

### Summary
10 findings from 3 reviewers (1 human, 2 AI agents).
4 to implement across 2 batches, 1 pushed back, 2 deferred, 0 conflicts escalated.

### Batch 1 — Critical (implement first)
- [ ] F01: Document Dispatcher `_ALLOWED_KEYS` alignment in implementation plan — `docs/plans/2026-04-28-github-webhook-handler-impl.md`
  - Reviewers: danieljchuser
  - Source: review body (not inline)
  - What to do: Add a "Cross-repo Contract" section to the implementation plan documenting that 5 fields (`event_type`, `event_action`, `installation_id`, `delivery_id`, `workspace_repos`) must be added to the Dispatcher Lambda's `_ALLOWED_KEYS` whitelist before Phase 2 goes live. Document that option A (expand whitelist) is the recommended approach, and that a companion PR on spacecat-infrastructure is needed.

### Batch 2 — Important + Minor
- [ ] F02/F03: Fix Node.js version in implementation plan — `docs/plans/2026-04-28-github-webhook-handler-impl.md:9`
  - Reviewers: danieljchuser, Copilot
  - What to do: Change "Node.js 22+" to "Node.js 24.x" in the Tech Stack line.

- [ ] F04/F05: Add HMAC security scheme to OpenAPI contract — `docs/openapi/webhooks-api.yaml:68`, `docs/openapi/api.yaml`
  - Reviewers: danieljchuser, Copilot
  - What to do: Add `GitHubWebhookSignature` apiKey security scheme to `api.yaml` `components/securitySchemes`, then change `security: []` in `webhooks-api.yaml` to `security: [{ GitHubWebhookSignature: [] }]`.

### Pushback List

| F06/F07: Use shared `responses.yaml#/401` ref | danieljchuser, Copilot | The custom 401 with `WebhookUnauthorized` body is intentional per the approved spec. The shared `responses.yaml#/401` has no body schema — using it would lose the error message, which is needed for debugging HMAC failures. |

### Deferred / Needs Clarification

| Finding | Reviewer(s) | Blocked on |
|---------|-------------|------------|
| F08: `workspace_repos` cross-org access | danieljchuser | Runtime concern for the worker's GitHub App installation token scope. No code change in this PR. |
| F09: Add `correlation_id` to job payload | danieljchuser | Worker already derives idempotency key. Would need spec amendment. |

solaris007 (changes requested):
Hey @szergen,

Thorough Phase 1 contract and a well-structured Phase 2 plan with proper TDD sequencing. I've read @danieljchuser's review and your action plan - concurring with their Critical (verified independently against the deployed dispatcher source), and adding a few items not yet raised that should be folded in before Phase 2 implementation starts.

## Strengths

- Phase 1 OpenAPI contract is clean: reuses existing `hooks` tag, includes `500` response, propagates `X-GitHub-Delivery` to job payload, `MYSTICAT_GITHUB_JOBS_QUEUE_URL` matches the renamed infra variable.
- Phase 2 plan correctly captures the round-1/round-2 spec-review outcomes: HMAC handler extends `AbstractHandler` and slots into `authHandlers` (no bypass list, no separate wrap chain); `SIGNATURE_PATTERN` regex pre-check before `timingSafeEqual`; `EVENT_JOB_MAP` constant; `getSkipReason` extracted as a pure function; field-specific 400 messages.
- TDD sequencing is correct: failing test, run to confirm fail, then implementation, then run to confirm pass. Three modules (trigger rules, HMAC handler, controller) get this treatment.
- `webhooks-api.yaml` description explicitly documents the `bodyData`-runs-first ordering and the `request.text()` cached-body assumption (line 13-14).
- Task 6 (verify `request.text()` caching) is the right safety check before implementing the HMAC handler.

## Issues

### Critical (Must Fix)

**1. Concurring with danieljchuser finding 1 - Dispatcher `_ALLOWED_KEYS` mismatch.** Verified independently against the deployed dispatcher at `adobe/spacecat-infrastructure:modules/mysticat_github_service/src/dispatcher/handler.py` (commit `e74c625`):

```python
_REQUIRED_KEYS = ("job_type", "owner", "repo", "event_ref")
_ALLOWED_KEYS = frozenset(_REQUIRED_KEYS + ("retry_count", "source_task_arn"))
```

The controller's payload (`event_type`, `event_action`, `installation_id`, `delivery_id`, `workspace_repos`, `job_type`, `owner`, `repo`, `event_ref`, `retry_count`) will trip `unexpected = set(payload.keys()) - _ALLOWED_KEYS` and raise `payload has unexpected keys: [...]`. ESM retries 3x, then DLQ. 100% failure rate.

Option (A) is the right answer - expand the dispatcher whitelist with appropriate validation (`installation_id` is a numeric string, `workspace_repos` is an array of `org/repo` strings, etc.). Document the companion infrastructure PR or commit as a hard prerequisite in this plan's "Cross-repo Contract" section so Phase 2 isn't blocked behind a coordination round.

### Important (Should Fix)

**2. `AbstractHandler` import reaches into the package's internal `src/` path.** (Task 8, Step 3, plan line ~640)

```javascript
import AbstractHandler from '@adobe/spacecat-shared-http-utils/src/auth/handlers/abstract.js';
```

Importing from `<package>/src/...` is a deep import into the package's internal layout. If `@adobe/spacecat-shared-http-utils` restructures its internals (rename, move, repackage), this import breaks silently - and there is no SemVer guarantee on internal paths.

Verify the package's public exports. If `AbstractHandler` is exported from the package root (e.g. `import { AbstractHandler } from '@adobe/spacecat-shared-http-utils';`), use that. If not, the package needs to add a public export for it before Phase 2 starts - that is a one-line PR on the shared package, but it should land first so this controller doesn't accumulate tech debt. Same applies to `AuthInfo` (which IS already used as a named import from the package root in the example, so the precedent is set).

**3. Test mocking patterns are unspecified and the plan defers the resolution to "check existing tests".** (Task 8 Step 1 footnote, Task 9 Step 1 footnote)

Two specific items that the plan flags but doesn't resolve:
- HMAC handler test uses `headers: new Map(...)`; the real Request uses a `Headers` object whose `.get(name)` returns `null` for missing keys, while `Map.get` returns `undefined`. The plan's note says "adjust the helper if tests fail" - that contradicts the TDD sequencing where tests must be deterministic before implementation lands.
- Controller 400 tests call `await response.json()`; the plan notes this may not work with the shared HTTP utils' response objects and says "check `test/controllers/hooks.test.js`."

Either resolve both before TDD starts (read the existing test patterns and codify them in the plan) or note explicitly that the test code will need adjustment in step 4 of each task. As written, the plan invites flaky-tests-by-design as the implementer adapts the mocks mid-task.

**4. Task 6 has no branching plan if `request.text()` caching does not work.** (Phase 2 Task 6, plan line ~466)

> If `request.text()` works: proceed as designed.
> If it does NOT work: implement option (b) from the spec - add a `rawBodyCapture` wrapper in `src/index.js` between `bodyData` and `authWrapper` in the `.with()` chain.

There is no Task 6.5 or 6.b specifying:
- Where the `rawBodyCapture` wrapper code lives (new file? where?)
- Its exact `.with()` placement (between `bodyData` (line 367) and `authWrapper` (line 355) in the wrap chain - but the `.with()` ordering rules mean "between" needs a precise position)
- Tests for the wrapper itself
- How the HMAC handler test mock changes if the wrapper is now responsible for caching

If the verification fails, the implementer is stuck guessing. Add a Task 6.b that fully specifies option (b) so the contingency is concrete.

### Minor (Nice to Have)

**5. `workspace_repos` hardcoded with cross-org `Adobe-AEM-Sites/aem-sites-architecture`.** (Task 9, controller code, plan line ~1139)

Same observation as @danieljchuser flagged and you deferred to runtime concerns - but worth a one-liner in the plan: the Mysticat GitHub App's installation token must have access to BOTH orgs (`adobe` and `Adobe-AEM-Sites`). This is a per-installation provisioning step, not a code change, and missing it surfaces as a `git clone` failure in the worker. Worth flagging in the plan's Phase 2 prerequisites alongside the dispatcher whitelist coordination.

## Recommendations

- Add a "Phase 2 Prerequisites" section to the plan that consolidates the cross-repo dependencies into a checkable list: dispatcher `_ALLOWED_KEYS` expansion (companion infra PR), `AbstractHandler` public export verification, GitHub App installation in both orgs (`adobe`, `Adobe-AEM-Sites`), `request.text()` caching confirmed. Each should be a checkbox before Task 7 starts.
- Acknowledged carry-forward: `workspace_repos` per-repo configurability is Phase 3 (config cache). The hardcoded array in Phase 2 is the documented placeholder per the merged spec - fine.
- Pushback on the shared `responses.yaml#/401` ref is reasonable (custom body needed for HMAC debugging). Worth one comment line in `webhooks-api.yaml` next to the custom 401 response so future readers don't re-open the discussion.

## Assessment

Reasoning: The Phase 1 OpenAPI contract is essentially ready. The blocker is Phase 2 plan coordination - the `_ALLOWED_KEYS` mismatch is a guaranteed dev-DLQ failure, and the test mock + `request.text()` contingency gaps will produce TDD-cycle pain mid-implementation. All four fixes are doc-only edits to the plan that prevent downstream re-work.

## Next Steps

1. Resolve the Dispatcher `_ALLOWED_KEYS` alignment (the critical) - pick option A and document the companion infra PR.
2. Verify `AbstractHandler` is publicly exported from `@adobe/spacecat-shared-http-utils`; fix the import or land a one-line export PR first.
3. Pin the test mock patterns from existing `hooks` tests directly in Tasks 8 and 9 instead of "check the existing tests."
4. Add a Task 6.b with concrete option-(b) implementation steps.
5. Apply the already-triaged Batch 1 / Batch 2 fixes (Node version, security scheme).
6. Minor `workspace_repos` cross-org note can land alongside.

szergen:
## Review Findings Addressed

**F01 (Critical) — Dispatcher `_ALLOWED_KEYS` alignment:** Added a "Cross-repo Contract" section to the implementation plan documenting the 5 fields that need to be added to the Dispatcher whitelist, with the recommended approach (option A) and deployment order. Fixed in b3a886ba.

**F02 (Important) — Node.js version:** Changed "Node.js 22+" to "Node.js 24.x" in the plan. Fixed in b3a886ba.

**F04 (Important) — `security: []` modeling:** Added `github_webhook_signature` apiKey security scheme to `api.yaml` and replaced `security: []` with explicit reference in `webhooks-api.yaml`. Fixed in b3a886ba.

**F06 (Minor) — 401 response:** Not implementing. The custom 401 with `WebhookUnauthorized` body is intentional per the approved spec. The shared `responses.yaml#/401` has no body schema — using it would lose the error message needed for debugging HMAC failures.

**F08 (Minor) — `workspace_repos` cross-org:** Noted. Runtime concern for the worker's installation token scope. No code change in this PR.

**F09 (Recommendation) — `correlation_id`:** Noted. Worker already derives idempotency key from `{owner}/{repo}/{pr_number}/{head_sha}`. Would need spec amendment.

szergen:
## Review Response Summary

### Implemented (4 findings)
- **F01**: Added Cross-repo Contract section documenting Dispatcher `_ALLOWED_KEYS` alignment — `b3a886ba`
- **F02/F03**: Fixed Node.js version from "22+" to "24.x" — `b3a886ba`
- **F04/F05**: Added `github_webhook_signature` security scheme, replaced `security: []` — `b3a886ba`

### Pushed Back (1 finding)
- **F06/F07**: Custom 401 with `WebhookUnauthorized` body is intentional per approved spec. Shared `responses.yaml#/401` has no body schema.

### Deferred (2 findings)
- **F08**: `workspace_repos` cross-org access — runtime concern for worker, not webhook handler
- **F09**: `correlation_id` in job payload — worker already derives idempotency key; needs spec amendment

szergen:
## solaris007 Review Findings Addressed

**S01 (Critical) — Dispatcher `_ALLOWED_KEYS`:** Already addressed in prior commit `b3a886ba` (Cross-repo Contract section).

**S02 (Important) — `AbstractHandler` deep import:** Not implementing. The existing `SkipAuthHandler` at `src/index.js:32` uses the identical import path (`@adobe/spacecat-shared-http-utils/src/auth/handlers/abstract.js`). This is the established pattern in this codebase.

**S03 (Important) — Test mock patterns:** Fixed in `c2e964f4`. Task 8 `makeRequest` now uses `{ get: (name) => headers[name] || null }` (Headers-like behavior). Task 9 tests use `await (await controller.method(context)).json()` (double-await pattern from hooks.test.js). Removed "check existing tests" footnotes.

**S04 (Important) — Task 6.b contingency:** Added in `c2e964f4`. Concrete Task 6.b with `rawBodyCapture` wrapper: file location, `.with()` placement, HMAC handler update, test changes. Only executed if Task 6 verification fails.

**S05/S06 (Minor) — Phase 2 Prerequisites + cross-org note:** Added in `c2e964f4`. Consolidates 4 prerequisites as checkboxes before Task 7, including GitHub App installation in both `adobe` and `Adobe-AEM-Sites` orgs.

**S07 (Minor) — Custom 401 comment:** Added YAML comment in `webhooks-api.yaml` explaining why the shared `responses.yaml#/401` ref isn't used. Fixed in `c2e964f4`.

solaris007 (approved):
Hey @szergen,

All round-1 findings addressed in two focused commits. The Cross-repo Contract section is exactly the documentation that was missing, Task 6.b is concrete and correct, and the security scheme + Node version + tag-reuse comment all landed cleanly. One small wrinkle in test-code idiom worth a fast-follow but not blocking.

## Strengths

**Round-1 findings now addressed:**

- **Critical (Dispatcher `_ALLOWED_KEYS` mismatch):** new "Cross-repo Contract" section in the plan (lines 36-60) documents the issue, the option-A fix with the exact whitelist expansion, and the deployment ordering. Phase 2 Prerequisites checklist now lists the companion infra PR as a hard blocker before Task 7. Verified: the proposed `_ALLOWED_KEYS` expansion correctly adds the 5 missing fields, and the validation note for `installation_id` (numeric string) and `workspace_repos` (array of `namespace/repo` strings) is the right shape.

- **Important (`AbstractHandler` import path):** verified the pushback is correct. The existing `src/index.js:32-33` imports both `AuthInfo` and `AbstractHandler` from `@adobe/spacecat-shared-http-utils/src/auth/...` - the deep-import precedent is already established in this codebase. Following the existing pattern is right for this PR. Adding a public export to the shared package is a separate concern (see recommendations).

- **Important (test mocking patterns):** `Map` -> `{ get: (name) => headers[name] || null }` is exactly the right shape for matching the real `Headers.get()` behavior (returns `null` for missing keys). Test code now matches what the HMAC handler actually receives. The "if tests fail" footnotes are removed.

- **Important (Task 6 contingency):** new Task 6.b (lines 277-371) has everything an implementer needs: file path, wrapper code, exact `.with()` placement (correctly outermost so it runs first, before bodyData consumes the stream), HMAC handler update, test update, commit message. The wrapper reads cleanly and the idempotent guard is correct.

- **Minor (workspace_repos cross-org):** Phase 2 Prerequisites item 4 now explicitly flags the GitHub App must have access to both `adobe` and `Adobe-AEM-Sites` orgs.

**danieljchuser findings also addressed:**
- Node.js 22+ -> 24.x (plan line 9)
- `security: []` -> `security: [{ github_webhook_signature: [] }]` with a new `github_webhook_signature` apiKey scheme in `api.yaml`
- 401 response retains the custom `WebhookUnauthorized` body with an explicit comment explaining why (matches the rationale in the action plan pushback)

## Issues

### Minor (Nice to Have)

**1. The 400-test JSON read pattern calls the controller twice.** (`docs/plans/2026-04-28-github-webhook-handler-impl.md:1023, 1036`)

The plan now reads:
```javascript
const response = await controller.processGitHubWebhook(context);
expect(response.status).to.equal(400);
const body = await (await controller.processGitHubWebhook(context)).json();
```

That invokes the controller twice and uses the second response's body. It works (both calls 400 with the same payload), but it's wasteful and a future reader will rightfully wonder why. Three cleaner options to pick from:

- (a) Read body off the first response: `const body = await response.json();` (works if the response object isn't single-consume).
- (b) Clone before reading: `const body = await response.clone().json();` (works if `.json()` consumes the stream).
- (c) Read the raw body via the same idiom the existing `test/controllers/hooks.test.js` uses - e.g., `JSON.parse(response.body)` if that's the project pattern.

Pick whichever matches the existing project convention. The double-call hides the actual question (does the response object support repeated `.json()` calls?) under a workaround.

## Recommendations

- The `AbstractHandler` deep-import precedent is established in this repo, so following it for the new handler is correct. Worth filing a separate issue against `@adobe/spacecat-shared-http-utils` to add a public export so the codebase can stop reaching into `src/`. Not for this PR.
- Companion infra PR for the `_ALLOWED_KEYS` expansion is not yet open per a quick search of `adobe/spacecat-infrastructure` open PRs. Phase 2 Prerequisites checkbox 1 won't tick until that PR lands - worth opening it now in parallel so the dependency isn't sequential.

## Assessment

Ready to merge? **Yes.**

Reasoning: All round-1 Critical and Important findings are addressed cleanly, the new Cross-repo Contract section and Task 6.b add real implementer-facing rigor, and the plan now correctly partitions cross-repo coordination as Phase 2 Prerequisites (not blockers for merging this Phase 1 contract). The single Minor is a test-idiom polish that can ride along with Phase 2 implementation.

danieljchuser (approved):
Hey @szergen,

All round-1 findings cleanly addressed across the two new commits.

## Strengths

- **Critical (Dispatcher `_ALLOWED_KEYS` mismatch):** the new "Cross-repo Contract" section documents the exact whitelist expansion needed, the recommended option, and the deployment ordering. Phase 2 Prerequisites makes the companion infra PR a hard blocker before Task 7. This is the right shape for a cross-repo coordination point.
- **Important (Node.js version):** corrected from "22+" to "24.x" matching the repo's `package.json` engines requirement.
- **Important (`security: []` misleading):** replaced with a proper `github_webhook_signature` apiKey scheme in `api.yaml` and referenced from the operation. The contract now self-documents the actual auth mechanism.
- **Minor (`workspace_repos` cross-org):** Phase 2 Prerequisites item 4 flags the GitHub App must have access to both `adobe` and `Adobe-AEM-Sites` orgs.
- **Minor (custom 401 vs shared response):** pushback is fair - the inline comment now documents the rationale (shared 401 has no body schema, HMAC failures need a debuggable message body).

Task 6.b is a nice addition - concrete fallback wrapper if the `request.text()` caching investigation in Task 6 turns up empty, with the correct `.with()` ordering explained.

## Assessment

Ready to merge? Yes.

Reasoning: All round-1 Critical and Important findings are addressed. The Cross-repo Contract section gives Phase 2 implementers exactly the cross-repo coordination they need. The remaining Minor flagged by another reviewer (the double-call test JSON pattern) is a test-idiom polish that can ride along with Phase 2 implementation.

szergen:
## Phase 2 Implementation Complete

Phase 2 (on-demand handler) is now implemented on top of the approved Phase 1 OpenAPI contract.

### New files

| File | Purpose |
|------|---------|
| `src/support/github-webhook-hmac-handler.js` | `GitHubWebhookHmacHandler` extending `AbstractHandler` - HMAC-SHA256 signature verification, path-scoped to `/webhooks/*`, format pre-check before `timingSafeEqual`, reads cached raw body via `request.text()` |
| `src/utils/github-trigger-rules.js` | `EVENT_JOB_MAP` (`pull_request` -> `pr-review`) + `getSkipReason()` pure function for payload-only trigger rules |
| `src/controllers/webhooks.js` | `WebhooksController` factory function - validates payload, checks event mapping, applies trigger rules, enqueues job to SQS |
| `test/support/github-webhook-hmac-handler.test.js` | 10 test cases (valid sig, invalid sig, missing header/secret, malformed prefix, wrong length, empty body, wrong secret, raw-vs-reserialized proof, path scoping) |
| `test/utils/github-trigger-rules.test.js` | 12 test cases (review_requested, labeled, opened, ready_for_review, closed, draft, bot, non-default branch, slug default) |
| `test/controllers/webhooks.test.js` | 7 test cases (202 enqueue, 204 skip, 400 missing fields, delivery_id propagation, SQS failure 500, draft skip) |

### Modified files

| File | Change |
|------|--------|
| `src/index.js` | Import + register `GitHubWebhookHmacHandler` in `authHandlers` array (after `SkipAuthHandler`, before `JwtHandler`). Instantiate `WebhooksController`, pass to `getRouteHandlers()`. |
| `src/routes/index.js` | Add `webhooksController` parameter + `POST /webhooks/github` route definition |

### Trigger rules (on-demand only, Phase 2)

| Trigger | Condition | Result |
|---------|-----------|--------|
| `review_requested` | Reviewer is `mysticat[bot]` | Enqueue |
| `review_requested` | Different reviewer | 204 skip |
| `labeled` | Label is `mysticat:review-requested` | Enqueue |
| `labeled` | Different label | 204 skip |
| `opened` / `ready_for_review` | Any | 204 skip (Phase 3) |
| Draft PR | Any on-demand trigger | 204 skip |
| Bot sender | Any on-demand trigger | 204 skip |
| Non-default branch | Any on-demand trigger | 204 skip |

### Job payload (enqueued to SQS)

```json
{
  "owner": "adobe",
  "repo": "spacecat-api-service",
  "event_type": "pull_request",
  "event_action": "review_requested",
  "event_ref": "456",
  "installation_id": "12345678",
  "delivery_id": "github-delivery-guid",
  "job_type": "pr-review",
  "workspace_repos": ["adobe/mysticat-architecture", "adobe/mysticat-ai-native-guidelines", "Adobe-AEM-Sites/aem-sites-architecture"],
  "retry_count": 0
}
```

### Prerequisites before going live

Per the Cross-repo Contract section in the implementation plan:
- [ ] Dispatcher `_ALLOWED_KEYS` expansion (companion infra PR) must be merged first
- [ ] `GITHUB_WEBHOOK_SECRET` and `MYSTICAT_GITHUB_JOBS_QUEUE_URL` env vars set on the api-service Lambda
- [ ] `enable_github_webhook_route = true` feature flag enabled in API Gateway (last step)

danieljchuser (changes requested):
Hey @szergen,

Phase 2 implementation is well-shaped overall - the test matrix is solid (HMAC raw-body proof, trigger-rule branches), the layering is clean, and the timingSafeEqual hardening pattern is correct. There is one critical bug that will make the route dead in production despite green tests, plus a cluster of hardening items worth addressing before traffic flows.

## Strengths

- Cross-repo Dispatcher `_ALLOWED_KEYS` contract documented in implementation plan with deployment ordering and companion infra PR
- Node.js version corrected to 24.x
- `security: []` replaced with proper `github_webhook_signature` apiKey scheme
- `workspace_repos` cross-org access flagged in Phase 2 Prerequisites
- Custom 401 with `WebhookUnauthorized` body retained with rationale documented

New positives:

- Format pre-check with `SIGNATURE_PATTERN` regex before `timingSafeEqual` prevents length-mismatch throws (`src/support/github-webhook-hmac-handler.js:32-34`)
- Raw-body-vs-JSON.stringify regression test proves the right invariant (`test/support/github-webhook-hmac-handler.test.js:139-156`)
- `errorHandler` returns generic 500 without leaking stack/message to the caller
- Route correctly added to `INTERNAL_ROUTES` with rationale comment ("authenticated by HMAC-SHA256, not S2S JWT")
- Trigger rules extracted as pure functions with no I/O - trivially testable
- Path-scoped early bail keeps body-reading side effect off non-webhook routes
- `AuthInfo.withType('github_webhook')` gives a distinct queryable subject type for downstream audit/logging

## Issues

### Critical (Must Fix)

**1. HMAC handler path-prefix check will never match in production - the route is effectively dead code.** (`src/support/github-webhook-hmac-handler.js:21`)

```js
if (!context.pathInfo?.suffix?.startsWith('webhooks/')) return null;
```

In this codebase `pathInfo.suffix` consistently has a leading slash. Evidence: `test/index.test.js:204` uses `suffix: '/test'`, `:240` uses `'/sites/...'`, the route table uses keys like `'POST /webhooks/github'`. In production the suffix will be `/webhooks/github`, the `startsWith('webhooks/')` check returns false, and the handler returns `null` for every real webhook. The unit test mock uses `suffix: 'webhooks/github'` (no leading slash) so tests pass with false confidence, but the deployed route is unreachable - every webhook falls through to JwtHandler/IMS and gets 401.

Fix: tolerate either form and update test mocks to mirror production:

```js
if (!/^\/?webhooks\//.test(context.pathInfo?.suffix || '')) return null;
```

Update `makeContext` in `test/support/github-webhook-hmac-handler.test.js` to use `suffix: '/webhooks/github'` and add a test for the no-leading-slash variant. Most importantly, add an integration test in `test/index.test.js` that drives the full middleware chain (`bodyData` -> `authWrapper` -> controller) with a valid signature and asserts 202 - that is the test that would have caught this.

### Important (Should Fix)

**2. Controller crashes 500 (not 400) on missing `pull_request` / `repository`.** (`src/controllers/webhooks.js:54, 76, 84`)

The controller validates only `data.action` and `data.installation.id`. If a future `EVENT_JOB_MAP` entry is added for non-pull_request events (the `EVENT_JOB_MAP` comment explicitly anticipates `issues -> triage-issue`, `push -> changelog`), or a partial payload arrives, `String(pr.number)` and `data.repository.owner.login` throw `TypeError`. `errorHandler` converts this to opaque 500, GitHub retries, same bug fires repeatedly. Add explicit guards: `if (!data?.pull_request?.number) return badRequest('Missing required field: pull_request.number');` and same for `repository.owner.login` / `repository.name`.

**3. `workspace_repos` hardcoded in controller.** (`src/controllers/webhooks.js:80-84`)

Three workspace repos are baked in as a literal array. This is configuration masquerading as code - any change requires a deploy cycle, and the cross-org `Adobe-AEM-Sites/aem-sites-architecture` entry is a trust decision that should be reviewable as ops config. Move to env var (e.g. `MYSTICAT_WORKSPACE_REPOS`, comma-separated, parsed once) with the current values as a fallback default. Phase 3 then becomes a one-file change instead of a controller refactor.

**4. `GITHUB_APP_SLUG` default of `'mysticat'` is an unsafe fallback.** (`src/utils/github-trigger-rules.js:29`)

`env.GITHUB_APP_SLUG || 'mysticat'` silently masks misconfiguration at a security-relevant decision point (which bot can trigger automated runs). Across environments the actual app slug may differ. Remove the default and treat missing env var as a skip with reason `GITHUB_APP_SLUG not configured` - explicit failure beats silent wrong-default for security-relevant config.

**5. No payload size limit on `request.text()`.** (`src/support/github-webhook-hmac-handler.js:43`)

The handler reads the body unconditionally before computing HMAC. An attacker can send arbitrarily large bodies (GitHub caps at 25 MB, but a forged request is unbound) and force HMAC computation over megabytes - pre-auth resource exhaustion. Add a `Content-Length` check before reading: reject bodies larger than 1 MiB (real GitHub payloads are typically under 100 KB).

**6. Log injection via untrusted webhook headers.** (`src/controllers/webhooks.js:50, 59-66`)

`event` (from `x-github-event`) and `deliveryId` (from `x-github-delivery`) are interpolated into log message strings. Headers are NOT part of GitHub's HMAC-signed material - only the body is signed. An attacker who replays a captured signature with arbitrary headers can inject newlines or control characters into log lines, corrupting Coralogix DataPrime queries and SIEM parsing. Pass user-controlled fields as structured log context, never interpolated: `log.info('Skipping unmapped event', { event, deliveryId })`.

**7. Silent fallthrough when `GITHUB_WEBHOOK_SECRET` is missing.** (`src/support/github-webhook-hmac-handler.js:27-30`)

Logs at `error` level on every request (alert noise under GitHub redelivery) and returns null, presenting identically to a non-webhook request. On-call sees auth failures and starts debugging signing keys at GitHub instead of looking at Secrets Manager. Either log once per cold start (module-scoped flag) or lower per-request log to `warn`. At minimum, include structured fields (`{ misconfigured: true }`) so a Coralogix alert can be wired.

**8. `context.rawBody` is set but never consumed.** (`src/support/github-webhook-hmac-handler.js:78`)

Comment says "for controller use (e.g. logging, debugging)" but the controller never reads it. Carrying the raw HMAC-validated payload (which can include PR bodies and source code patches) as `context.rawBody` for the rest of the request lifetime is dead state and a foot-gun if any future logging middleware dumps `context`. Drop the assignment until there is a real consumer.

### Minor (Nice to Have)

**9. Skip-rule branches not tested for `labeled` action.** (`test/utils/github-trigger-rules.test.js:101-127`)

Draft, bot sender, and non-default-branch skip reasons are only tested with `action: 'review_requested'`. The same skip rules run for `labeled`. Add parametrized tests across both supported actions.

**10. SQS-failure test does not assert the error was logged.** (`test/controllers/webhooks.test.js:139-145`)

Asserts only `response.status === 500`, missing `mockLog.error.calledOnce`. The whole point of `errorHandler` is to log before swallowing - test should prove it.

**11. HMAC handler negative paths do not assert log calls.**

Missing-secret, malformed-format, empty-body, and mismatch paths all have no log assertions. These warn/error log lines are how on-call distinguishes "secret rotated" from "actual attack" from "GitHub format change." Pin them in tests.

## Recommendations

- Establish the config seam now: move `workspace_repos` and `GITHUB_APP_SLUG` into a config module (alongside `EVENT_JOB_MAP` in `src/utils/github-trigger-rules.js` or a new `src/utils/github-config.js`). Phase 3 becomes "swap a constant for a function" inside that module.
- Document the auth handler ordering contract in `src/index.js` with a short comment block explaining the array order and why `GitHubWebhookHmacHandler` must run before path-agnostic handlers.
- Add a comment at the top of `webhooks.js`: "Body is HMAC-verified; headers (x-github-event, x-github-delivery) are NOT part of signed material - treat as untrusted."
- Replay protection: GitHub does not include a timestamp header (unlike Slack), so HMAC alone does not stop replay of captured signed payloads. Make sure the consumer side actually deduplicates on `delivery_id` (the infrastructure DynamoDB idempotency table uses `correlation_id` as the key).

- Companion infra PR adobe/spacecat-infrastructure (item 481) must merge before Phase 2 traffic flows - it expands the Dispatcher's `_ALLOWED_KEYS` whitelist to accept the 5 fields this controller sends.
- Replay protection at the SQS level (MessageDeduplicationId or DDB cache for delivery IDs) is a cross-repo concern requiring infra queue changes if the queue is standard (not FIFO).

## Assessment

Ready to merge? No, with fixes.

Reasoning: The path-prefix bug (item 1) makes the HMAC handler unreachable in production despite green unit tests - the feature ships dead. Items 2-8 are hardening that should land before production traffic: they cluster around insufficient input validation (untrusted headers, missing fields, no size cap) and silent failure modes (default slug, missing secret, dead context state). All are 5-30 line fixes and cheapest to address now while the surface is small.

## Next Steps

1. Fix the path-prefix bug and add an integration test that proves the full wrap chain works end-to-end - this is the single highest-leverage change.
2. Address the 7 important hardening items (all small, can be a single commit).
3. Minor test coverage items can ride along or follow.

szergen:
## Round 3 Review Findings Addressed

All 11 findings addressed in `b8b6db65`. Locally verified: 8572 tests passing, 100% coverage, lint clean.

### Critical (1/1 fixed)

**D01 — Path-prefix bug making route unreachable in production.** Fixed. The real `pathInfo.suffix` has a leading slash (`/webhooks/github`), but our check and test mock both used the no-leading-slash form. Changed to a regex (`/^\/?webhooks\//`) that tolerates both forms. Added tests for both variants to prevent regression. `makeContext` default now mirrors production with `suffix: '/webhooks/github'`.

### Important (7/7 fixed)

**D02 — Missing payload field guards.** Added explicit `badRequest` guards for `pull_request.number`, `repository.owner.login`, `repository.name` before building the SQS payload. Partial payloads now return 400 with field name instead of crashing to 500.

**D03 — Hardcoded `workspace_repos`.** Moved to `MYSTICAT_WORKSPACE_REPOS` env var (comma-separated) with the current 3-repo list as fallback. Parsed once in controller factory.

**D04 — Unsafe `GITHUB_APP_SLUG` default.** Removed `|| 'mysticat'` fallback. Missing env var now returns explicit skip reason `GITHUB_APP_SLUG not configured`. Security-relevant config should fail explicitly.

**D05 — No payload size limit (DoS).** Added 1 MiB cap with two checks: `content-length` header pre-read (rejects without reading) and actual `rawBody.length` post-read (catches missing/false content-length).

**D06 — Log injection via untrusted headers.** Replaced string interpolation with structured log context. `log.info('Skipping unmapped event', { event, deliveryId })` instead of `log.info(`Skipping unmapped event: ${event}`, ...)`. Applied to all log calls that touched header-derived values. Added top-of-file comment documenting the invariant.

**D07 — Silent fallthrough on missing secret.** Added `misconfigured=true` marker to error log for Coralogix alert wiring.

**D08 — Unused `context.rawBody`.** Removed the stash. The comment said "for controller use" but no consumer existed; dead state is a foot-gun if future middleware dumps `context`.

### Minor (3/3 fixed)

**D09 — Skip rules only tested for `review_requested`.** Parametrized skip-rule tests across both `review_requested` and `labeled` actions (6 tests now instead of 3).

**D10 — SQS failure test didn't assert log.error called.** Now asserts `mockLog.error.calledOnce` and message content.

**D11 — HMAC negative paths didn't assert log calls.** Added log assertions to malformed-prefix, wrong-length, empty-body, wrong-secret, and missing-secret tests. These log lines are how on-call distinguishes misconfig from attack.

### Additional tests added

- Payload size limit (pre-read via content-length, post-read via body length)
- `MYSTICAT_WORKSPACE_REPOS` env var parsing
- Log injection prevention (untrusted header value goes to structured context, not message string)
- Path scoping for both leading-slash and no-leading-slash forms

### Pushed back / deferred

None. All findings were addressed.

### Not in scope (tracked)

- Integration test driving the full middleware chain (mentioned by reviewer as highest-leverage for D01). The unit-level fix + regression tests for both path forms address the core bug. A full integration test would require broader test infrastructure changes; worth tracking as a follow-up but not needed to land this PR.
- Replay protection at SQS level (cross-repo infra concern, mentioned in reviewer's "Out of scope" section).

solaris007 (approved):
Hey @szergen,

Phase 2 implementation is solid. All 11 round-3 findings (the critical path-prefix bug plus 7 important hardening items plus 3 minor test-coverage items) are cleanly addressed and individually verified against the source. The security posture is meaningfully tightened - structured log context, fail-fast on missing config, two-tier size cap, explicit pull_request field guards. One small ordering nit and one comment-only clarification, neither blocking.

## Strengths

**Round-3 findings now addressed:**

- **Critical (path-prefix bug).** Fixed at `src/support/github-webhook-hmac-handler.js:18` with `WEBHOOK_PATH_PATTERN = /^\/?webhooks\//` and the inline comment explaining the production-vs-test-mock divergence. The regex tolerates both forms and the comment heads off the next person from re-introducing the same mismatch.
- **Pull_request and repository field guards** (`src/controllers/webhooks.js:92-101`) returning 400 with field-specific messages. `pr.number`, `repository.owner.login`, and `repository.name` all explicitly validated.
- **`workspace_repos` moved out of the controller into env config.** New `getWorkspaceRepos(env)` reads `MYSTICAT_WORKSPACE_REPOS` (comma-separated) with the prior literal as the documented default. Phase 3 swap becomes a one-env-var change instead of a code edit.
- **`GITHUB_APP_SLUG` no longer falls back to `'mysticat'`.** `getSkipReason` now returns the explicit skip reason `'GITHUB_APP_SLUG not configured'` - misconfiguration surfaces as a structured skip, not a silent default.
- **Two-tier payload size cap with `MAX_BODY_BYTES = 1 MiB`.** Content-Length pre-check (line 51) plus post-read length check (line 73). Pre-auth resource exhaustion vector closed for honest clients; defense-in-depth for missing/wrong Content-Length.
- **Log injection vector removed.** Header-derived values (`event`, `deliveryId`) now passed as structured log context (`log.info('Skipping unmapped event', { event, deliveryId })`) instead of interpolated into message strings. The comment block at the top of `webhooks.js` explicitly documents that headers are NOT part of GitHub's HMAC-signed material.
- **Missing-secret signal includes `misconfigured=true`** marker - alertable on Coralogix/SIEM without distinguishing per-request from cold-start.
- **`context.rawBody` dead-state assignment removed.** The handler reads `request.text()` for HMAC, no longer carries the raw payload as request-lifetime state.

**New positives in the implementation:**

- Auth handler order in `src/index.js` correctly places `GitHubWebhookHmacHandler` after `SkipAuthHandler` and before path-agnostic JWT/IMS handlers - path-scoped early bail keeps non-webhook routes free of the body-read side effect.
- Route correctly listed in `src/routes/required-capabilities.js` `INTERNAL_ROUTES` with the rationale comment ("authenticated by HMAC-SHA256, not S2S JWT").
- `AuthInfo.withType('github_webhook')` gives a queryable subject type for downstream audit.
- Comment at the top of `webhooks.js` correctly documents the HMAC-signed-body / unsigned-headers boundary.

## Issues

### Minor (Nice to Have)

**1. Skip-rule ordering can mask malformed payloads as 204 instead of 400.** (`src/controllers/webhooks.js:73-101`)

The validation order is: `action` and `installation.id` (400 if missing) -> EVENT_JOB_MAP (204 if unmapped) -> `getSkipReason` (204 if skip) -> `pr.number` / `repository.*` validation (400 if missing).

If a payload arrives with `action: 'review_requested'` and a valid reviewer but `pull_request` entirely missing (and `repository.default_branch` set), `getSkipReason` reaches the non-default-branch check, evaluates `undefined !== 'main'` as true, and returns `'non-default branch: undefined'`. The controller returns 204 with a misleading skip log rather than 400 with the actual problem.

Practical impact is low (GitHub does not send malformed payloads), but the log is debug-confusing. Two clean fixes: move the `pr.number` / `repository.*` validation BEFORE `getSkipReason`, OR have `getSkipReason` early-return null when `data.pull_request` is missing for the trigger actions (fall through to the 400 guards).

**2. `MAX_BODY_BYTES` pre-check is content-length-based and skippable.** (`src/support/github-webhook-hmac-handler.js:50-54`)

`if (Number.isFinite(contentLength) && contentLength > MAX_BODY_BYTES)` only fires when a Content-Length header is present and parsable. An attacker omitting Content-Length skips the pre-check; `await request.text()` (line 67) then reads the full body into memory before the post-check rejects (line 73). For the strict size guarantee you would need a stream-with-cap reader, which `request.text()` does not expose.

The two-tier check is the right shape given the API constraint, but a one-line comment near line 51 noting that the pre-check is honest-client-only and the post-check is the actual enforcement would prevent a future reader from thinking either alone is sufficient.

## Recommendations

- The HMAC handler's `AuthInfo.withProfile({ user_id: 'github-webhook' })` makes every webhook delivery indistinguishable in the auth profile. If downstream audit ever needs per-installation tracing without joining against the SQS payload, consider including `installation_id` (parsed from the verified body) in the profile. Not for this PR; SQS-side audit is sufficient today.
- A short comment block at the top of `src/index.js` near the `authHandlers` array documenting the order contract (path-scoped HMAC before path-agnostic JWT/IMS) would protect against accidental reordering during future auth additions.

- Companion infra PR adobe/spacecat-infrastructure item 481 (Dispatcher `_ALLOWED_KEYS` expansion) is ready and approved - confirm it merges before flipping `enable_github_webhook_route = true`.
- Branch has merge conflicts with main (`mergeable_state: dirty`). Rebase needed before merge.
- Replay protection at the SQS level (`MessageDeduplicationId` or DDB cache for `delivery_id`) - cross-repo concern noted in danieljchuser's review.

## Assessment

Ready to merge? **Yes.**

Reasoning: The Phase 2 implementation faithfully follows the approved spec and plan, all 11 round-3 findings are addressed with verifiable source-level evidence, and the security-critical surface is hardened along all the right axes (size caps, log injection, fail-fast config, field validation). The two Minor items are polish - one ordering tweak and one comment - that can land alongside the rebase. CI is pending on the latest push but green on prior commits in this PR series.

szergen:
## Round 4 Review Response Summary

All solaris007 recommendations addressed + main merged. Locally verified: 8788 tests passing, 100% coverage, lint clean.

### Implemented (4 items)

**S04** — Merged `origin/main` into branch (commit `5347e359`). Only `docs/index.html` conflicted; regenerated via `npm run docs:build`. Our changes (HMAC handler, webhooks route, INTERNAL_ROUTES entry, webhooks-api.yaml, trigger rules) auto-merged cleanly. Pulled in from main: shared-http-utils v1.26.0 bump, sites-resolve envelope, LLMO notifier, PLG cleanup controller, referral traffic controller, and other features. Full test suite passes post-merge.

**S01** — Reordered `pull_request.number` / `repository.owner.login` / `repository.name` guards **above** the `getSkipReason` call in `src/controllers/webhooks.js`. Missing `pull_request` now correctly surfaces as 400 with field name instead of a misleading "non-default branch: undefined" 204 skip. (commit `1a937b8d`)

**S02** — Added comment in `src/support/github-webhook-hmac-handler.js` documenting the two-tier size check: pre-read Content-Length is honest-client-only (saves body read when header is set); post-read `rawBody.length` is the actual enforcement (catches missing/falsified Content-Length). (commit `1a937b8d`)

**S03** — Added `authHandlers` order contract comment in `src/index.js` explaining why path-scoped `GitHubWebhookHmacHandler` must run before path-agnostic `JwtHandler`/`AdobeImsHandler`: webhooks would otherwise reach JWT handler and fail with misleading 401 on missing JWT. Includes guidance for future path-scoped handlers. (commit `1a937b8d`)

### Not in this PR (tracked)

- `installation_id` in `AuthInfo.withProfile` — reviewer explicitly said "not for this PR"
- Dispatcher `_ALLOWED_KEYS` companion infra PR — separate repo, confirm merged before `enable_github_webhook_route = true`
- SQS replay protection — cross-repo

src/utils/github-trigger-rules.js
```diff
@@ -0,0 +1,82 @@
...
+ */
+export function getSkipReason(data, action, env) {
+  const pr = data.pull_request;
+  const appSlug = env.GITHUB_APP_SLUG;
+
+  // Explicitly fail rather than default: app slug is a security-relevant
+  // decision (which bot can trigger automated runs). Missing env var is a
+  // misconfiguration, not a "try mysticat" situation.
+  if (!appSlug) {
+    return 'GITHUB_APP_SLUG not configured';
```
danieljchuser: This returns a skip reason, causing the controller to respond 204 to GitHub. GitHub treats 204 as successful delivery and will not retry - webhooks lost during this window are gone permanently. A missing `GITHUB_APP_SLUG` is a misconfiguration, not a legitimate skip.

Suggested fix: validate `GITHUB_APP_SLUG` at controller entry alongside the other field guards and return `internalServerError` with `log.error` instead. That gives GitHub a 5xx to retry against and keeps the on-call recovery window open.

```js
// In webhooks.js, alongside the pull_request / repository guards:
if (!env.GITHUB_APP_SLUG) {
  log.error('GITHUB_APP_SLUG not configured', { deliveryId });
  return internalServerError('GITHUB_APP_SLUG not configured');
}
```

src/support/github-webhook-hmac-handler.js
```diff
@@ -0,0 +1,102 @@
...
+    //     header to skip this branch). Saves the body read when the header is set.
+    //   - Post-read via rawBody.length (below): the actual enforcement, catches
+    //     missing or falsified Content-Length.
+    const contentLength = Number(request.headers.get('content-length'));
+    if (Number.isFinite(contentLength) && contentLength > MAX_BODY_BYTES) {
+      this.log(`Payload too large: ${contentLength} bytes`, 'warn');
+      return null;
+    }
+
+    // Validate signature format before timingSafeEqual (prevents throw on length mismatch)
```
danieljchuser: The format check currently runs after the secret check (line 44) and the Content-Length pre-check (line 53). A request with a present-but-malformed `x-hub-signature-256` header reaches the secret check first - if `GITHUB_WEBHOOK_SECRET` is unset it logs `error` before the signature is ever shape-validated.

Moving this block immediately after the `if (!signature)` early-exit (before line 43) fixes this. Format validation is purely structural and requires no I/O:

```js
if (!signature) { return null; }

// Validate format FIRST - before secret check or body I/O
if (!SIGNATURE_PATTERN.test(signature)) {
  this.log('Malformed X-Hub-Signature-256 header', 'warn');
  return null;
}

const secret = context.env?.GITHUB_WEBHOOK_SECRET;
// ...
```

danieljchuser (changes requested):
Hey @szergen,

All eleven prior findings from the round-3 review are addressed and the new hardening is solid. Two new issues were introduced by the round-3 changes themselves - both are small fixes.

## Strengths

- F1 (path-prefix bug): `WEBHOOK_PATH_PATTERN = /^\/?webhooks\//` correctly handles production's leading-slash suffix; both forms tested.
- F2 (missing field guards): `badRequest` for `pull_request.number`, `repository.owner.login`, `repository.name` runs before `getSkipReason`, preventing misleading 204 skips on partial payloads.
- F3 (`workspace_repos` hardcoded): env-driven via `MYSTICAT_WORKSPACE_REPOS` with `getWorkspaceRepos`, fallback constant, and override test.
- F5 (no payload size limit): two-tier 1 MiB check correctly ordered - Content-Length pre-check, then post-read enforcement, then HMAC compute. The comment honestly frames the header tier as honest-client-only optimization.
- F6 (log injection): structured context throughout `webhooks.js`, no template-literal interpolation of untrusted header values. Top-of-file comment documents the unsigned-headers rationale. Regression test added that asserts injected `\n` value lands in the context object, not the message string.
- F7 (missing secret fallthrough): `(misconfigured=true)` marker in error message, log assertion in test. Per-request error log is acceptable given `AbstractHandler.log(message, level)` is a string-only API.
- F8 (dead `context.rawBody`): removed.
- F9 (labeled action skip-rule tests): parametrized describe covers both `review_requested` and `labeled` for draft, bot, and non-default-branch skip rules.
- F10 (SQS failure missing log assertion): `mockLog.error.calledOnce` and message text now asserted.
- F11 (HMAC negative paths missing log assertions): every negative path test now asserts the correct log call.
- Auth handler ordering contract documented in `src/index.js:358-367` with rationale for path-scoped handler positioning.

## Issues

### Important (Should Fix)

**1. `GITHUB_APP_SLUG` missing returns 204 - permanent webhook loss on misconfiguration.** (`src/utils/github-trigger-rules.js:39`)

When `GITHUB_APP_SLUG` is absent from env vars, `getSkipReason` returns `'GITHUB_APP_SLUG not configured'` and the controller responds 204 to GitHub. GitHub treats 204 as successful delivery and does not retry. Every webhook arriving during the misconfiguration window is permanently lost - GitHub's redelivery mechanism never fires because 204 is success. The skip is also logged at `info` level alongside legitimate skips (draft PR, bot sender), so the misconfiguration is invisible without monitoring for the specific `skipReason` value.

The original `|| 'mysticat'` default kept the feature working despite misconfiguration. Removing it was the right call for F4. But the replacement failure mode should be 5xx, not 204.

Fix: check `GITHUB_APP_SLUG` at controller entry alongside the F2 field guards and return `internalServerError` with `log.error('GITHUB_APP_SLUG not configured', { deliveryId })`. Remove the early return from `getSkipReason` - the controller can own this validation.

**2. SIGNATURE_PATTERN format check runs after secret check, enabling pre-auth log amplification.** (`src/support/github-webhook-hmac-handler.js:59`)

Current check order: signature present - secret present (logs `error` if missing) - Content-Length - SIGNATURE_PATTERN format check. Any request with a present-but-malformed `x-hub-signature-256` header triggers the secret-presence error log before signature shape is validated. This reopens F7's alert-noise concern: with `GITHUB_WEBHOOK_SECRET` unset, a request with a malformed but present signature header drives an `error` log without the signature ever being shape-checked.

Fix: move the `if (!SIGNATURE_PATTERN.test(signature))` block (lines 58-62) immediately after the `if (!signature)` early-exit (line 41), before the secret check. Format validation is purely structural, secret-free, and requires no I/O - it belongs at the top of the evaluation chain.

### Minor (Nice to Have)

**1. Auth handler ordering is a comment, not an executable contract.** (`src/index.js:372`)

The comment block at lines 358-367 correctly documents why `GitHubWebhookHmacHandler` must precede `JwtHandler`. A 3-line test asserting `authHandlers.indexOf(GitHubWebhookHmacHandler) < authHandlers.indexOf(JwtHandler)` turns this into an executable contract. Acceptable as a follow-up.

**2. No end-to-end middleware chain integration test.** (`test/index.test.js`)

The prior review flagged absence of a full-chain test (bodyData - authWrapper - GitHubWebhookHmacHandler - controller) as the highest-leverage gap. This was not added in round 3/4 - same gap as before, not regressed. A minimal test signing a body, sending to `/webhooks/github`, and asserting 202 plus SQS send would be the definitive guard against handler-ordering regressions. Acceptable as follow-up given thorough unit coverage.

## Recommendations

- `MYSTICAT_WORKSPACE_REPOS` parser silently accepts invalid entry formats (e.g., `adobe-repo-name` without a slash). A format check per entry with `log.warn` and fallback to defaults on failure would catch fat-finger deploys.
- `rawBody.length` counts JS string units, not bytes. For ASCII-dominated webhook payloads this is not a practical concern, but `Buffer.byteLength(rawBody, 'utf8')` is the byte-accurate alternative if the 1 MiB threshold is intended as a byte limit.
- Log a warning when `MYSTICAT_WORKSPACE_REPOS` is not set and the built-in default is in use, so the fallback is visible rather than silent.

## Assessment

Ready to merge? No, with fixes.

Reasoning: Ten of eleven prior findings are fully resolved and the hardening is correct. Both Important items were introduced by the round-3 changes - the GITHUB_APP_SLUG failure mode causes silent data loss on a misconfiguration that will eventually occur in some environment, and the format-check move is 4 lines. Both are cheaper to address now than to fix after production traffic flows through the endpoint.

## Next Steps

1. Address GITHUB_APP_SLUG failure mode (`internalServerError` instead of 204) - prevents permanent webhook loss.
2. Move SIGNATURE_PATTERN format check before secret check - closes pre-auth log amplification.
3. Minor items are optional follow-ups.

szergen:
## Round 5 Review Response Summary

All actionable findings addressed in `a9a9abb7`. Locally verified: 8792 tests passing, 100% coverage, lint clean.

### Implemented (6 items)

**R01 (Important) — `GITHUB_APP_SLUG` missing now returns 500, not 204.** Moved the validation from `getSkipReason` to controller entry. Missing config logs `error` with `deliveryId` and returns `internalServerError`, so GitHub retries the delivery. Previously this returned 204 (skip reason), which GitHub treats as success, causing permanent webhook loss during the misconfiguration window. Removed the early-return from `getSkipReason` — controller owns config validation now. `src/controllers/webhooks.js`, `src/utils/github-trigger-rules.js`.

**R02 (Important) — `SIGNATURE_PATTERN` format check moved before secret check.** Reordered `checkAuth` so format validation runs immediately after the `!signature` early-exit, before the secret-presence and content-length checks. Closes the pre-auth log amplification vector: with `GITHUB_WEBHOOK_SECRET` unset, a request with a malformed-but-present signature no longer triggers the `error` log; it bails at the format check with a `warn`. Added regression test. `src/support/github-webhook-hmac-handler.js`.

**R03 (Minor) — Auth handler order as an executable test.** Added `test/auth-handlers-order.test.js`. Parses `src/index.js` source and asserts `GitHubWebhookHmacHandler` appears before `JwtHandler`, `AdobeImsHandler`, `ScopedApiKeyHandler`, `LegacyApiKeyHandler` in the `AUTH_HANDLERS` array. Turns the comment contract into an executable check without booting the full handler chain.

**R05 (Minor) — Validate `MYSTICAT_WORKSPACE_REPOS` entries.** `getWorkspaceRepos` now checks each entry against `/^[^/\s]+\/[^/\s]+$/` (owner/repo format). Invalid entries are dropped with a `log.warn` naming them. If the env var contains no valid entries, falls back to defaults with a `log.warn`. `src/controllers/webhooks.js`.

**R06 (Minor) — Byte-accurate size check.** Replaced `rawBody.length` with `Buffer.byteLength(rawBody, 'utf8')` in the post-read size check. JS string `.length` counts code units, not bytes — multi-byte characters would let an oversized payload slip past. `src/support/github-webhook-hmac-handler.js`.

**R07 (Minor) — Warn when defaults used.** `getWorkspaceRepos` now emits `log.warn` with the defaults array when `MYSTICAT_WORKSPACE_REPOS` is not set, so the fallback is visible in logs rather than silent. `src/controllers/webhooks.js`.

### Deferred (1 item)

**R04 — Full middleware-chain integration test.** Reviewer explicitly noted this is "acceptable as follow-up given thorough unit coverage." Tracked as follow-up.

### Not in scope (tracked)

- Dispatcher `_ALLOWED_KEYS` companion infra PR — separate repo, must merge before `enable_github_webhook_route = true`
- SQS replay protection — cross-repo

Also merged `origin/main` (2 new commits: drs-client v1.6.0, readall capability design doc). No conflicts.

solaris007 (approved):
Hey @szergen,

Both round-5 Importants and all 3 Recommendations addressed in `a9a9abb`. The executable auth-handler-order contract is the right call - `AUTH_HANDLERS` is now a named constant with a comment block explaining the ordering rationale, and the test parses the source to enforce it. Clean cycle.

## Strengths

**Round-5 findings now addressed:**

- **`GITHUB_APP_SLUG` failure mode flipped from 204 to 500.** (`src/controllers/webhooks.js:85-91`, `src/utils/github-trigger-rules.js:33-35`) Validation moved to controller entry as a hard 5xx with `log.error` + `deliveryId`. `getSkipReason` no longer participates in this check (caller-validated). The comment block in the controller explicitly documents why 5xx is correct (GitHub retries on 5xx; treats 204 as success and never redelivers, which would silently lose every webhook during a misconfiguration window). New test asserts the 500 path and confirms SQS is not called.

- **`SIGNATURE_PATTERN` format check moved to the top of the evaluation chain.** (`src/support/github-webhook-hmac-handler.js:42-48`) Now runs immediately after the missing-signature early-exit, before the secret-presence check. Comment explicitly explains the ordering rationale ("structural check, no I/O, no config"). New regression test "malformed signature + missing secret does NOT log secret error" asserts the format check runs first and prevents pre-auth log amplification.

- **Auth-handler order is now an executable contract.** (`test/auth-handlers-order.test.js`, `src/index.js:358-367`) `AUTH_HANDLERS` extracted as a named constant. Comment block above the constant documents the path-scoped vs path-agnostic split. The test parses the source file (avoiding the heavy boot-time side effects of importing it) and asserts `GitHubWebhookHmacHandler` precedes `JwtHandler`, `AdobeImsHandler`, `ScopedApiKeyHandler`, `LegacyApiKeyHandler`. Reasonable testing strategy given the constraint.

- **`MYSTICAT_WORKSPACE_REPOS` per-entry format validation.** (`src/controllers/webhooks.js:30-67`) New `WORKSPACE_REPO_PATTERN = /^[^/\s]+\/[^/\s]+$/`, valid/invalid partitioning at controller construction, structured `log.warn` for invalid entries (with the offending values), fallback to defaults if all entries are invalid, and an explicit log when env var is unset. Three new tests cover the partial-invalid, all-invalid, and unset-defaults paths.

- **`Buffer.byteLength(rawBody, 'utf8')` byte-accurate post-read size check.** (`src/support/github-webhook-hmac-handler.js:79-82`) Replaces the JS-string-length count. For Unicode-heavy bodies this matters; for typical GitHub payloads it does not, but the byte-accurate form is the right contract.

## Issues

### Minor (Nice to Have)

**1. Stale test filename in comment.** (`src/index.js:367`)

The comment block above `AUTH_HANDLERS` says `// AUTH_HANDLERS order is enforced by test/auth-handlers.test.js.` but the actual test file is `test/auth-handlers-order.test.js`. Trivial edit; harmless functionally but a future reader following the breadcrumb will hit a dead path.

- `ci / it-postgres` is failing on this PR. The failure pattern is consistent with what we saw across PR #2255 (pre-existing, integration test environment), unrelated to webhook handler changes. Worth verifying the failure isn't a regression triggered by the rebase onto current main, but not a webhook-PR concern to fix here.
- End-to-end middleware chain integration test (Minor 2 from round 5) deferred to follow-up. Unit coverage is thorough; the gap is the absence of a test that drives the full `bodyData -> authWrapper -> GitHubWebhookHmacHandler -> controller` chain. Worth filing as a tracked follow-up so it doesn't drop off.

## Assessment

Ready to merge? **Yes.**

Reasoning: Both round-5 Importants are addressed correctly and at the right architectural layer (5xx semantics for misconfiguration so GitHub redelivers; structural check before any I/O for log noise). All 3 Recommendations landed cleanly with thorough test coverage. The single Minor is a one-line comment fix. Pre-existing `it-postgres` failure is the only CI red and is consistent with the cross-PR pattern, not a webhook-PR concern.

danieljchuser (approved):
Hey @szergen,

All findings from rounds 3-5 comprehensively addressed. The critical path-prefix bug is fixed with the regex pattern, and all hardening items landed cleanly:

- Path check uses `WEBHOOK_PATH_PATTERN` regex tolerating both leading-slash forms
- Two-tier payload size limit (Content-Length + post-read byte-length check, 1 MiB cap)
- Explicit `badRequest()` guards for `pull_request.number`, `repository.owner.login`, `repository.name`
- `workspace_repos` moved to env var `MYSTICAT_WORKSPACE_REPOS` with format validation and fallback defaults
- `GITHUB_APP_SLUG` missing now returns 500 (not 204) so GitHub retries - correct semantics
- All log calls use structured context objects, trust boundary comment at top of file
- `context.rawBody` removed (dead state eliminated)

Ready to merge. Reminder: companion infra PR (spacecat-infrastructure 481, merged) expanded the Dispatcher whitelist for the 5 fields this controller sends. Worker CI/CD pipeline still needed before end-to-end traffic can flow.

PR #2283, opened 2026-04-28
feat(plg): clean up stale FIXED suggestions and FixEntities on PLG onboarding

absarasw:
## Summary

When a domain is (re-)onboarded into PLG, suggestions and `FixEntity` rows left
behind from a prior PLG lifecycle on the same site can re-surface in the UI and
confuse downstream auto-fix logic. This PR adds a best-effort cleanup pass that
runs immediately before each `PlgOnboarding` record transitions to `ONBOARDED`,
so the newly onboarded site starts from a clean slate.

For each PLG opportunity on the site (`cwv`, `alt-text`, `broken-backlinks`):

| Suggestion status                                  | Action                                             |
|----------------------------------------------------|----------------------------------------------------|
| `FIXED`, `IN_PROGRESS`, `SKIPPED`, `ERROR`, `REJECTED` | → `OUTDATED` (via `Suggestion.bulkUpdateStatus`)   |
| `PENDING_VALIDATION`                               | → `NEW` (re-enters the normal pipeline)            |
| `NEW`, `APPROVED`                                  | left untouched                                     |

In addition, all associated `FixEntity` rows for these PLG opportunities are
deleted via `FixEntity.removeByIds` so a future re-onboarding starts from an
empty fix history.

Non-PLG opportunities (e.g. `structured-data`, `product-meta`) are left
untouched.

## Changes

- New utility `src/controllers/plg/plg-onboarding-cleanup.js` exporting:
  - `cleanupPlgSiteSuggestionsAndFixes(siteId, context)` — main entry point.
  - `PLG_CLEANUP_OPPORTUNITY_TYPES` — the opportunity types in scope.
  - `STATUSES_TO_OUTDATE` / `STATUSES_TO_RESET_TO_NEW` — the per-status
    transition sets.

  Extracted into a separate module to keep `plg-onboarding.js` from growing
  further. A single `Suggestion.allByOpportunityId` fetch per opportunity is
  partitioned in memory into the two transition buckets, so the round-trip
  count is minimal.

- `src/controllers/plg/plg-onboarding.js` invokes the cleanup in both
  onboarding paths, immediately before `setStatus(STATUSES.ONBOARDED)`:
  - Fast-path (`PRE_ONBOARDING → ONBOARDED`)
  - Full onboarding flow (`IN_PROGRESS → ONBOARDED`)

- Per-opportunity work runs in parallel; per-status transitions inside an
  opportunity also run in parallel. Errors at every layer (opportunity list,
  suggestion list, each `bulkUpdateStatus`, fix-entity list, `removeByIds`)
  are caught and logged at warn level. Cleanup is best-effort and never blocks
  onboarding — orphaned rows can be reconciled out-of-band, but blocking
  onboarding on a cleanup failure is unacceptable.

## Why "before ONBOARDED" (not on offboarding)

Offboarding paths (admin `ONBOARDED → WAITLISTED` and displacement)
intentionally do not run cleanup. Performing the cleanup at the *next*
onboarding instead means:

- The offboarding hot path stays minimal.
- Sites that are offboarded and never re-onboarded don't pay the cleanup cost.
- Re-onboarding always sees a fresh start regardless of how the previous
  lifecycle ended (admin-driven waitlist, displacement, error, etc.).

## Test plan

- [x] `test/controllers/plg/plg-onboarding-cleanup.test.js` — 12 unit tests
      covering: empty `siteId`, opportunity-list failure, no PLG opportunities,
      happy path with mixed `OUTDATE` / `RESET_TO_NEW` / untouched statuses
      across all three PLG opportunity types and a non-PLG opportunity, and
      every error branch (suggestion-list failure for one opportunity,
      independent failure of the OUTDATED bulk update, independent failure of
      the NEW reset, fix-entity list failure, `removeByIds` failure, no-op when
      nothing matches either transition set).
- [x] 100% line / branch / statement / function coverage on the new utility.
- [x] All existing `plg-onboarding.test.js` controller tests still pass; test
      setup updated to add `Suggestion.bulkUpdateStatus`,
      `FixEntity.allByOpportunityId`, and `FixEntity.removeByIds` stubs to the
      default `mockDataAccess` so the cleanup is a fast no-op in existing
      tests; bumped the suite-level mocha timeout to 10s (matching other
      esmock-heavy suites in this repo) to eliminate flakes at the 2s default
      boundary.
- [ ] Stage smoke test: onboard a domain that previously had a mix of `FIXED`,
      `PENDING_VALIDATION`, `IN_PROGRESS` `cwv` suggestions and a `FixEntity`,
      confirm that on `ONBOARDED` the first three buckets are `OUTDATED`,
      `PENDING_VALIDATION` rows become `NEW`, and the `FixEntity` rows are
      gone.

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/controllers/plg/plg-onboarding-cleanup.js
+++ b/src/controllers/plg/plg-onboarding-cleanup.js
@@ -1,0 +1,169 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { Suggestion as SuggestionModel } from '@adobe/spacecat-shared-data-access';
+
+/**
+ * Opportunity types that participate in the PLG auto-fix lifecycle.
+ * Kept in sync with the LD auto-fix flags and the displacement check
+ * in plg-onboarding.js (PLG_OPPORTUNITY_TYPES).
+ */
+export const PLG_CLEANUP_OPPORTUNITY_TYPES = ['cwv', 'alt-text', 'broken-backlinks'];
+
+/**
+ * Marks every FIXED suggestion under the given PLG opportunity as OUTDATED.
+ * FIXED suggestions represent past customer wins; once the site is offboarded
+ * we want the next audit run to start from a clean slate rather than re-surface
+ * stale "fixed" entries.
+ *
+ * Failures are logged and swallowed — cleanup is best-effort and must never
+ * block the higher-level offboarding flow.
+ *
+ * @param {object} opportunity - Opportunity model instance.
+ * @param {object} Suggestion - Suggestion collection from dataAccess.
+ * @param {object} log - Logger.
+ * @returns {Promise<number>} Count of suggestions transitioned (0 on failure).
+ */
+async function outdateFixedSuggestionsForOpportunity(opportunity, Suggestion, log) {
+  const opportunityId = opportunity.getId();
+  const opportunityType = opportunity.getType();
+  try {
+    const fixedSuggestions = await Suggestion.allByOpportunityIdAndStatus(
+      opportunityId,
+      SuggestionModel.STATUSES.FIXED,
+    );
+
+    if (fixedSuggestions.length === 0) {
+      return 0;
+    }
+
+    await Suggestion.bulkUpdateStatus(fixedSuggestions, SuggestionModel.STATUSES.OUTDATED);
+    log.info(
+      `PLG cleanup: marked ${fixedSuggestions.length} ${opportunityType} suggestions as OUTDATED `
+      + `for opportunity ${opportunityId}`,
+    );
+    return fixedSuggestions.length;
+  } catch (error) {
+    log.warn(
+      `PLG cleanup: failed to mark FIXED suggestions OUTDATED for ${opportunityType} opportunity `
+      + `${opportunityId}: ${error.message}`,
+    );
+    return 0;
+  }
+}
+
+/**
+ * Deletes every FixEntity attached to the given PLG opportunity. A FixEntity
+ * may be re-created on the next audit run — leaving stale ones around can
+ * confuse downstream auto-fix logic that resumes/replays past fixes.
+ *
+ * Failures are logged and swallowed — cleanup is best-effort.
+ *
+ * @param {object} opportunity - Opportunity model instance.
+ * @param {object} FixEntity - FixEntity collection from dataAccess.
+ * @param {object} log - Logger.
+ * @returns {Promise<number>} Count of fix entities removed (0 on failure).
+ */
+async function removeFixEntitiesForOpportunity(opportunity, FixEntity, log) {
+  const opportunityId = opportunity.getId();
+  const opportunityType = opportunity.getType();
+  try {
+    const fixEntities = await FixEntity.allByOpportunityId(opportunityId);
+
+    if (fixEntities.length === 0) {
+      return 0;
+    }
+
+    await FixEntity.removeByIds(fixEntities.map((f) => f.getId()));
+    log.info(
+      `PLG cleanup: removed ${fixEntities.length} ${opportunityType} fix entities for `
+      + `opportunity ${opportunityId}`,
+    );
+    return fixEntities.length;
+  } catch (error) {
+    log.warn(
+      `PLG cleanup: failed to remove fix entities for ${opportunityType} opportunity `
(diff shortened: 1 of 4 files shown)


Kanishkavijay39:
Hey @absarasw,                                                                                                                                                                                                                                     
                                                                    
  Strengths                                         

  - Clean extraction into plg-onboarding-cleanup.js (src/controllers/plg/plg-onboarding-cleanup.js:1) keeps the already-large onboarding controller from growing further - the right call.                                                           
  - Two-level parallelism is well-structured: opportunities run in parallel, and within each opportunity the suggestion transition and fix-entity deletion also run in parallel (src/controllers/plg/plg-onboarding-cleanup.js:196-205).
  - Empty-array short-circuit before calling bulkUpdateStatus (src/controllers/plg/plg-onboarding-cleanup.js:57) avoids unnecessary DB round-trips.                                                                                                  
  - Exported constants (PLG_CLEANUP_OPPORTUNITY_TYPES, STATUSES_TO_OUTDATE, STATUSES_TO_RESET_TO_NEW) make the cleanup contract inspectable and testable without reading the implementation.                                                         
  - Test suite is thorough: all error branches covered, partial-failure isolation verified (OUTDATED transition fails but NEW reset succeeds, and vice versa), and non-PLG opportunities confirmed untouched                                         
  (test/controllers/plg/plg-onboarding-cleanup.test.js).                                                                                                                                                                                             
  - The "clean up at re-onboarding, not at offboarding" design rationale is sound and well-documented in the PR description.                                                                                                                         
                                                                                                                                                                                                                                                     
  Issues                                                            
                                                                                                                                                                                                                                                     
  Important (Should Fix)                                            
                                                    
  PLG_CLEANUP_OPPORTUNITY_TYPES duplicates PLG_OPPORTUNITY_TYPES from plg-onboarding.js                                                                                                                                                              
  - src/controllers/plg/plg-onboarding-cleanup.js:20
  - The comment says "Kept in sync with the LD auto-fix flags and the displacement check in plg-onboarding.js (PLG_OPPORTUNITY_TYPES)". Two separate arrays for what appears to be the same logical set will silently drift. When PLG adds a new     
  opportunity type and someone updates PLG_OPPORTUNITY_TYPES in the onboarding file, the cleanup will silently skip it for all future re-onboardings - stale suggestions will keep surfacing in the UI for that type. This is exactly the class of
  bug that takes months to notice.                                                                                                                                                                                                                   
  - Fix: consolidate into a shared constants module (e.g., src/controllers/plg/plg-constants.js) and import from both files, or have the cleanup module import PLG_OPPORTUNITY_TYPES directly from plg-onboarding.js. Either way, one source of
  truth.                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                     
  Minor (Nice to Have)                              
                                                                                                                                                                                                                                                     
  siteId guard logs at info rather than warn                                                                                                                                                                                                         
  - src/controllers/plg/plg-onboarding-cleanup.js:175
  - A missing siteId at this call site indicates the caller passed a bad argument - that's a bug, not an expected code path. log.warn (or log.error) is more appropriate and will surface faster in log monitoring.                                  
                                                                                                                                                                                                                   
  Unused setStatus stub in createMockSuggestion                                                                                                                                                                                                      
  - test/controllers/plg/plg-onboarding-cleanup.test.js:42                                                                                                                                                                                           
  - setStatus: sinon.stub() is created on every mock suggestion but the implementation calls Suggestion.bulkUpdateStatus(suggestions, status) - it never calls s.setStatus(). The stub is harmless but will confuse future readers into thinking     
  individual setStatus calls are expected behavior.                                                                                                                                                                                                  
                                                                                                                                                                                                                                                     
  Return value discarded by callers                 
  - src/controllers/plg/plg-onboarding.js:809 and src/controllers/plg/plg-onboarding.js:1251                                                                                                                                                         
  - cleanupPlgSiteSuggestionsAndFixes returns {outdatedCount, resetToNewCount, removedFixCount} but both call sites discard the result. The function logs the totals internally, so this is fine operationally - but the public signature implies    
  callers should do something with the counts. Either void the return or add a void annotation/comment to signal the discard is intentional.
                                                                                                                                                                                                                                                     
  Recommendations                                                   
                                                                                                                                                                                                                                                     
  - Consider emitting a structured metric or a log entry at a distinct level when cleanup runs and finds items (e.g., outdatedCount > 0). Right now the summary log at info level exists, but having an observable signal that ops can aggregate in  
  Coralogix would make it easy to verify the cleanup is actually firing and finding stale data after re-onboardings in production.
  - Add a JSDoc note to the exported cleanupPlgSiteSuggestionsAndFixes function clarifying that callers are responsible for ensuring siteId belongs to the authorized tenant before calling. The function is exported and could be called from a     
  different context in the future - an explicit contract note prevents accidental misuse.                                                                                                                                                            
                                                    
  Assessment                                                                                                                                                                                                                                         
                                                                    
  Ready to merge? With fixes                                                                                                                                                                                                                         
  Reasoning: The implementation is sound and well-tested, but the duplicate opportunity-type list is a silent drift risk that is easy to fix now and painful to diagnose later. Everything else is polish.
                                                                                                                                                                                                                                                     
  Next Steps                                                        
                                                                                                                                                                                                                                                     
  1. Consolidate PLG_CLEANUP_OPPORTUNITY_TYPES and PLG_OPPORTUNITY_TYPES into a shared constant to address the one Important issue.                                                                                                                  
  2. The three Minor items are optional but worth a quick pass.

Kanishkavijay39 (approved):

PR #2287, opened 2026-04-29
feat: PLG bind aso entitlement to ims org

Kanishkavijay39:
## Summary

  - **Bind entitlement to IMS-resolved org**: `ensureAsoEntitlement` uses `TierClient.createForOrg` with the org       
  resolved from the request's `imsOrgId` (not `site.getOrganizationId()`), so the entitlement is always bound to the
  correct customer org regardless of what org the site currently points at (e.g. internal/demo org pre-reassignment)   
  - **Direct site enrollment from entitlement ID**: Step 2 creates `SiteEnrollment` directly using
  `entitlement.getId()` — bypasses `TierClient.createForSite` which re-derives org from `site.getOrganizationId()` and 
  can read a stale value after reassignment
  - **Waitlist on entitlement/enrollment failure**: Both create and fetch failures result in `WAITLISTED` status;      
  `createEntitlement` failures are logged at `error` level for alerting
  `EntitlementWaitlistError` instead of warn-and-continue                                        
  - **Step flags on failure**: WAITLISTED records carry `entitlementFailed: true` in the steps field for programmatic  
  triage                                                                                                             
  - **Log warn on reassign mismatch**: `reassignSiteOrganization` logs a warning instead of throwing when the          
  re-fetched site doesn't reflect the new org (replica lag) 

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/controllers/plg/plg-onboarding.js
+++ b/src/controllers/plg/plg-onboarding.js
@@ -71,6 +71,13 @@ const REVIEW_REASONS = {
 const DOMAIN_ALREADY_ASSIGNED = 'already assigned to another organization';
 const DOMAIN_ALREADY_ONBOARDED_IN_ORG = 'another domain is already onboarded for this IMS org';
 
+class EntitlementWaitlistError extends Error {
+  constructor(message) {
+    super(message);
+    this.name = 'EntitlementWaitlistError';
+  }
+}
+
 /**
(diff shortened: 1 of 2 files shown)


absarasw (approved):
I think this looks cleaner. Removing enrolments using organisation object and creating site enrolment using exact entitlementId.

PR #2290, opened 2026-04-29
feat(llmo): forward agentTypes to agentic-traffic RPCs | LLMO-4526

JayKid:
## Summary

URL Inspector PG needs to restrict agentic traffic to the `{Chatbots, Research}` inclusion list. The mysticat RPCs (companion PR adobe/mysticat-data-service#490) gained an additive `p_agent_types TEXT[]` parameter on three RPCs; this PR exposes it through the agentic-traffic and URL Inspector controllers as a comma-separated `agentTypes` query parameter, **and** maps the two new agentic fields that `rpc_url_inspector_owned_urls` now returns (Option C / M2 fix from the multi-persona PR review on the first cut).

This is the second of three coordinated PRs for [LLMO-4526](https://jira.corp.adobe.com/browse/LLMO-4526). It must deploy after mysticat#490 is on DEV (the schema cache reload there is what makes `p_agent_types` callable on all three RPCs).

### `src/controllers/llmo/llmo-agentic-traffic.js`

- Added `VALID_AGENT_TYPES_CANONICAL` and `VALID_AGENT_TYPES_LOOKUP` so values are accepted case-insensitively but normalised back to the canonical form before being sent to the DB.
- `parseAgentTypes(raw)` helper:
  - splits the comma-separated string (or accepts an array as-is),
  - trims whitespace, lowercases,
  - **silently drops** unknown values (matches existing handler defence-in-depth: a UI bug must not accidentally widen the inclusion list — only validated tokens reach the RPC),
  - de-duplicates,
  - returns `null` when the resulting list is empty so the caller can branch on truthiness for the conditional RPC param. Existing callers that don't send `agentTypes` and the scalar `agentType` path are completely untouched.
  - **Now exported** so the URL Inspector handler can reuse the same validation lookup (single source of truth).
- `parseAgenticTrafficParams` now calls `parseAgentTypes` and stashes the result on the parsed params.
- `buildRpcParams` conditionally adds `p_agent_types` **only** to the two agentic-traffic RPC payloads that need it (`rpc_agentic_traffic_kpis_trend` and `rpc_agentic_traffic_by_url`); the movers and distinct-filters RPCs are unchanged.
- The by-url response mapping now exposes `hits_trend → hitsTrend` so the UI dialog can render per-URL multi-week agentic charts without a second request.

### `src/controllers/llmo/llmo-url-inspector.js` — Option C / M2 fix

- Imports the now-exported `parseAgentTypes`.
- `createUrlInspectorOwnedUrlsHandler` reads `agentTypes` (or legacy `agent_types`) from the query string, parses + validates it via the shared helper, and **conditionally** sets `rpcParams.p_agent_types` only when the result is non-null. Omission is byte-for-byte equivalent to the previous handler — existing callers see no change.
- The `urls.map(...)` step now adds two camelCase fields per row: `agenticHits` (from `agentic_hits`, defaulted to `0`) and `agenticHitsTrend` (from `agentic_hits_trend`, defaulted to `[]`, with each `{week_start, value}` rebound to `{weekStart, value}` and defensive null-coercion on each field).

### Tests

`test/controllers/llmo/llmo-agentic-traffic.test.js` — `describe` block covers:
- missing param (back-compat),
- single value, multiple values with whitespace and casing variations,
- unknown values silently dropped,
- conditional forwarding to the right RPCs,
- explicit omission from the RPCs that should not receive it,
- empty-token branch and non-string array members (added during a coverage-tightening pass after the initial PR push),
- `hits_trend` mapping with null values.

`test/controllers/llmo/llmo-url-inspector.test.js` — Option C additions:
- `returns paginated owned URLs with agentic fields mapped` — asserts `agenticHits` and `agenticHitsTrend` (with `weekStart` camelCase) are present.
- `coerces null fields inside agentic_hits_trend points (weekStart→null, value→0)` — pins both `??` fallback branches inside the trend `.map` callback.
- `passes filters and handles null row fields (agentic + brand-presence)` — `agentic_hits = null` and `agentic_hits_trend = null` map to `0` and `[]`; omitting `agentTypes` from the query causes the handler to **not** send `p_agent_types` to the RPC.
- `forwards comma-separated agentTypes as p_agent_types array`.
- `also accepts agentTypes as an array (no extra serialisation)`.
- `drops unknown agentTypes values and omits the param when empty`.
- `canonicalises agentTypes casing before forwarding`.

## Backwards compatibility

The new param is purely additive. Without `agentTypes` the controller produces identical RPC payloads to today. The owned-URLs response gains two additive fields (`agenticHits` always present and defaults to `0`; `agenticHitsTrend` always present and defaults to `[]`), which existing UI clients can ignore.

## Companion PRs

- **mysticat-data-service**: adobe/mysticat-data-service#490
- **project-elmo-ui**: adobe/project-elmo-ui#1640

## Test plan

- [x] Unit tests pass locally (`npm test`)
- [ ] Manual call against DEV (with mysticat#490 applied):
  - `?agentTypes=Chatbots,Research` returns Chatbots + Research traffic only on the kpis-trend / by-url endpoints
  - `?agentTypes=Chatbots,Research` on `/url-inspector/owned-urls` returns `agenticHits` / `agenticHitsTrend` populated
  - `?agentTypes=chatbots` (lowercase) is canonicalised and accepted
  - `?agentTypes=Unknown` causes the handler to omit `p_agent_types` (silent-drop) — `agenticHits` reflects all agent types
  - Omitting `agentTypes` returns the same shape as before on every endpoint
  - **M2 regression guard**: on a site with > 500 distinct agentic URLs, paginating owned URLs past the first page still returns populated agentic columns
- [ ] CI green

---

## Char's review (May 5) — addressed in 4dcc176e

| # | Severity | Item | Resolution |
|---|---|---|---|
| 1 | **High** | OpenAPI spec missing `agentTypes` query param + new response fields (`agenticHits`, `agenticHitsTrend`, `hitsTrend`) | Added a reusable `agentTypes` parameter in `docs/openapi/parameters.yaml` (with the dual-parameter intersection NOTE), `$ref`-wired on the three affected operations in `llmo-api.yaml`, and the new schema fields in `schemas.yaml`. The `agent_types` snake_case alias is preserved in code for legacy callers but **not** documented — canonical OpenAPI surface is `agentTypes` only. |
| 2 | **High** | No integration tests in `test/it/` | Added `test/it/postgres/llmo-url-inspector.test.js` + the shared spec under `test/it/shared/tests/llmo-url-inspector.js`. Includes validation tests, a response-shape contract test that pins `agenticHits` / `agenticHitsTrend` on every row regardless of mysticat image version (controller maps with `?? 0` / `?? []`), and two `it.skip` placeholders for the full `agentTypes` exercise — gated on `test/it/postgres/docker-compose.yml` bumping the mysticat-data-service tag to one containing the LLMO-4526 migrations (separate follow-up; depends on mysticat#490 being merged + a new image being published). |
| 3 | Design | Cross-module coupling for `parseAgentTypes` (controller → controller import) | **Resolved.** Extracted `parseAgentTypes` + `VALID_AGENT_TYPES_CANONICAL` + `VALID_AGENT_TYPES_LOOKUP` to a new `src/controllers/llmo/llmo-agent-types.js` module. `llmo-agentic-traffic.js` re-exports for back-compat; `llmo-url-inspector.js` now imports from the shared module directly. Single source of truth, no awkward import chain when a third caller arrives. |
| 4 | Defensive | Silent-drop on unknown `agentTypes` could be 400 | **Kept as-is.** Silent-drop is the existing project-wide convention for whitelist-based filter params (matches `successRate`, `platform`). Tightening only this one would be inconsistent with the rest of the agentic-traffic controller. Documented as a deferred decision in the plan doc. |
| 5 | Doc | `rpc_agentic_traffic_kpis` (non-trend) gap not surfaced | **Known gap, called out explicitly:** the non-trend KPIs handler does not receive `p_agent_types` — a caller filtering by `agentTypes` against `/agentic-traffic/kpis` silently gets unfiltered totals. No UI surface uses that combination today; trivial follow-up if/when one does. |
| 6 | Nits | `ctx`/`context` mixing; missing comment linking `VALID_AGENT_TYPES_CANONICAL` to DB enum; cross-repo plan-doc reference | The `ctx`/`context` mixing matches the existing pattern across `llmo-url-inspector.js` (consistent within the file). The DB-enum comment now lives at the top of the new `llmo-agent-types.js` module: *"mirrors the values stored in `agentic_traffic_weekly.agent_type`"*. The cross-repo plan-doc reference is intentionally retained — LLMO-4526 is the canonical pointer to the long-form working notes for the lifetime of this PR. |

Verification:
- `npm test`: 8564 passing, 100% statements / 100% branches / 100% lines
- `npm run lint`: clean
- `npm run docs:lint`: valid (warnings are pre-existing in unrelated audit specs)

Per-item rationale + extended file list under [Char's review (post first-push) — addressed in this PR](docs/plans/2026-04-29-llmo-4526-url-inspector-pg-agentic-from-db.md) in this PR.

This commit also bundles a separate review concern — **URL prompts handler hardening** — that hardens `createUrlInspectorUrlPromptsHandler` against SQLSTATE 22P02 (invalid UUID) errors by returning `200 { prompts: [] }` instead of `500`. This unblocks the URL Inspector PG dashboard's URL Details dialog when launched from the Owned URLs table (synthetic `url-${index}-${slug}` ids per LLMO-4526 M2 follow-up). Two new unit tests pin both detection branches (SQLSTATE-based and message-regex). Coverage stays at 100%.


calvarezg:
## Code Review

### Overview

This PR is well-structured, additive, and backwards-compatible. `parseAgentTypes` is clean and handles all edge cases correctly. The `buildAgentTypesRpcParam` pattern (returning `{}` when `null`) neatly avoids sending `p_agent_types` to RPCs that don't accept it — important since PostgREST rejects unknown named arguments. Test coverage is thorough across all branches.

---

### Issues

**1. Missing OpenAPI spec update (high)**

The new `agentTypes` query parameter and the new response fields (`agenticHits`, `agenticHitsTrend`, `hitsTrend`) aren't reflected in `docs/openapi/`. Per the project's API Design Principles: *"OpenAPI First: Define API contract in `docs/openapi/` before implementation"*. The dual snake_case alias `agent_types` should also either be documented or removed — undocumented aliases make the API surface harder to maintain.

**2. No integration tests**

Per `CLAUDE.md`: *"New or modified endpoints must include integration tests in `test/it/`."* The modified endpoints don't have new IT coverage. Given the deploy-order dependency on mysticat#490, a quick IT test asserting `agentTypes` is accepted and `agenticHits`/`agenticHitsTrend` are returned would add meaningful confidence.

**3. Cross-module coupling for `parseAgentTypes` (design concern)**

Exporting `parseAgentTypes` from `llmo-agentic-traffic.js` and importing it into `llmo-url-inspector.js` conflates "shared validation logic" with the agentic-traffic controller. If a third handler needs this, the import chain gets awkward. Worth extracting to a shared utility alongside `parseFilterDimensionsParams`.

**4. Silent-drop on unknown `agentTypes` values**

A caller sending `?agentTypes=UnknownType` gets unfiltered results with no indication something was wrong. If silent-drop is a project-wide convention for filter params that's fine, but `400 Bad Request` on fully-unrecognized values would be more defensive and clearer to the UI layer.

**5. `rpc_agentic_traffic_kpis` gap not surfaced**

The non-trend KPIs handler doesn't receive `p_agent_types` — correct per the tests, but not mentioned in the PR description. A caller filtering by `agentTypes` on the base KPIs endpoint silently gets unfiltered totals. Worth calling out explicitly as a known gap.

---

### Nits

- `llmo-url-inspector.js:216`: `ctx` and `context` naming is mixed in the same handler — minor consistency issue.
- `VALID_AGENT_TYPES_CANONICAL` mirrors a DB enum but has no comment linking to the source table/column. Future maintainers will need to know where to look when agent types change.
- The plan doc's cross-repo reference (`mysticat-data-service/docs/plans/...`) will go stale; the Jira ticket is a better home for that pointer.

---

**Bottom line:** Actionable before merge: (1) OpenAPI spec update for `agentTypes` param and new response fields, (2) IT test or explicit acknowledgement of the gap. Everything else is low-priority.

JayKid:
@calvarezg thanks for the thorough review! Both "actionable before merge" items + the design suggestion are landed in 4dcc176e. Two were deferred with rationale; nits all addressed or acknowledged.

**Actionable items — done:**
- **OpenAPI spec (#1, high):** new reusable `agentTypes` parameter in `parameters.yaml` (with the dual-parameter NOTE), `$ref`-wired on the three affected operations in `llmo-api.yaml`, and the new schema fields (`hitsTrend`, `agenticHits`, `agenticHitsTrend`) in `schemas.yaml`. The `agent_types` snake_case alias is preserved in code for legacy callers but intentionally **not** documented — canonical OpenAPI surface is `agentTypes` only.
- **Integration tests (#2, high):** added `test/it/postgres/llmo-url-inspector.test.js` + a shared spec under `test/it/shared/tests/llmo-url-inspector.js`. Includes validation tests, a response-shape contract test that pins `agenticHits` / `agenticHitsTrend` on every row regardless of mysticat image version (controller maps with `?? 0` / `?? []`), and two `it.skip` placeholders for the full `agentTypes` exercise — gated on bumping `test/it/postgres/docker-compose.yml`'s mysticat tag once mysticat#490 is published. Honest scoping: the data-service image dep means a full agentTypes IT test wasn't possible today.

**Design suggestion — done:**
- **Cross-module coupling (#3):** extracted `parseAgentTypes` + `VALID_AGENT_TYPES_*` to a new `src/controllers/llmo/llmo-agent-types.js` module. Both controllers now import from there; `llmo-agentic-traffic.js` re-exports for back-compat. Single source of truth, no more controller→controller import.

**Deferred with rationale:**
- **Silent-drop on unknown `agentTypes` (#4):** kept as-is. Silent-drop is the existing project-wide convention for whitelist-based filter params (matches `successRate`, `platform`). Tightening only this one would be inconsistent with the rest of the agentic-traffic controller — happy to revisit if you'd like a project-wide policy thread.
- **`rpc_agentic_traffic_kpis` (non-trend) gap (#5):** explicitly called out as a "Known gap" in both the plan doc and the PR description. No UI surface uses `agentTypes` against the non-trend KPIs endpoint today; trivial follow-up if/when one does.

**Nits:**
- `ctx`/`context` mixing in `llmo-url-inspector.js`: matches the existing pattern across the file (consistent within-file, inconsistent within-codebase) — left alone to avoid drive-by churn.
- DB-enum source comment for `VALID_AGENT_TYPES_CANONICAL`: now at the top of the new `llmo-agent-types.js` module — *"mirrors the values stored in `agentic_traffic_weekly.agent_type`"*.
- Cross-repo plan-doc reference: kept; LLMO-4526 is the canonical pointer to the long-form working notes for the lifetime of this PR.

Verification: `npm test` 8564 passing, 100% statements / branches / lines; `npm run lint` clean; `npm run docs:lint` valid.

Note this commit also bundles a separate fix — **URL prompts handler hardening** (SQLSTATE 22P02 → `200 { prompts: [] }`) — for the URL Details dialog opened from the Owned URLs flow. Two unit tests pin both detection branches; coverage stays at 100%.

Ready for re-review.

calvarezg:
## Follow-up review (updated PR)

All three actionable items from the first review are addressed — OpenAPI spec, cross-module extraction, and IT scaffolding look good. The URL Prompts SQLSTATE 22P02 hardening is a clean bonus fix.

One concrete bug remains:

**`AgenticTrafficByUrlRow.hitsTrend` schema uses `hits` but the code (and tests) emit `value`**

In `docs/openapi/schemas.yaml`, the `hitsTrend` item schema defines the field as `hits`:
```yaml
properties:
  weekStart: ...
  hits:
    type: integer
required: [weekStart, hits]
```

But the controller returns `value`:
```js
hitsTrend: row.hits_trend.map((point) => ({
  weekStart: point.week_start,
  value: Number(point.value ?? 0),   // ← "value"
}))
```

And the unit test asserts `value` too (`{ weekStart: '2026-01-05', value: 30 }`). The companion `agenticHitsTrend` schema in `UrlInspectorOwnedUrlRow` correctly uses `value`. The `hitsTrend` schema just needs `hits → value` (properties key + required entry).

---

**Minor IT test note** (not a blocker): the validation assertions accept quite wide status ranges (`[400, 401, 403, 500]` for invalid UUID, and the shape test early-returns on non-200), so the key contract assertions are conditional on getting a 200 — which may never happen in CI against the current image. This is understandable given the image dependency, but worth knowing those assertions won't catch regressions until the `it.skip` tests are enabled.

JayKid:
@calvarezg good catch — fixed in 94f7af6 (`fix(openapi): align AgenticTrafficByUrlRow.hitsTrend item with controller (value, not hits)`).

You're right: the schema declared `{weekStart, hits}` but the controller maps the underlying `rpc_agentic_traffic_by_url` `week_series` payload onto `{weekStart, value}`, and the unit test pins the same shape:

```js
hitsTrend: Array.isArray(row.hits_trend)
  ? row.hits_trend.map((point) => ({
    weekStart: point.week_start,
    value: Number(point.value ?? 0),
  }))
```

I flipped the schema's property key + `required` entry from `hits` to `value`, and added a sentence to the description noting that the field name is intentionally aligned with the underlying RPC payload — so this schema and `UrlInspectorOwnedUrlRow.agenticHitsTrend` look identical to clients (which is what they should). `npm run docs:lint` valid.

On the IT test note — fully agreed and I want to flag this honestly: the response-shape contract test currently early-returns on non-200 because it's running against the pinned `mysticat-data-service:v1.67.8` image (which pre-dates the LLMO-4526 RPCs). On the current CI image, `/owned-urls` may legitimately not return 200 with `agentTypes` at all, so the assertions are conditional. I left them in (rather than `it.skip`-ing the whole shape test) because:
- the basic path (no `agentTypes`, just a valid `siteId`) does return 200 on the current image, exercising the response-shape mapping in the controller (with `?? 0` / `?? []` fallbacks);
- once mysticat#490 is merged and a new `mysticat-data-service` image is published, the docker-compose tag bump will unlock all paths and tighten the assertions in one go;
- the two `it.skip` placeholders for the full `agentTypes` exercise make the gap explicit so it's not lost.

I've added a TODO comment in the IT spec pointing at the docker-compose tag bump as the gating action, so the next person reading the test file sees the dependency immediately. Happy to tighten the early-return into a hard fail once the image bump lands — just nervous about asserting on a 200 today and getting a flake when CI happens to land on the wrong image at the wrong time.

PS — the failing `ci / it-postgres` check on `4dcc176e` was an ECR transient (`manifest unknown` for `mysticat-data-service:v1.67.8`); same SHA passed on the `pull_request` event in the same minute. The new commit's CI run will sort it.

Ready for re-review.

calvarezg:
The  fix in  (schemas.yaml) is correct — property name,  entry, and description all updated consistently. All issues from both review rounds are resolved. LGTM 👍

calvarezg (approved):

PR #2292, opened 2026-04-29
fix: prevent unhandled 5xx in scrape controller on malformed URLs

radhikagpt1208:
JIRA: https://jira.corp.adobe.com/browse/SITES-43697

Two defensive fixes surfaced by the APIGatewayErrorRateHigh alert (SKYSI-76262) where a caller sent base64-encoded multi-line text in place of a URL:

- getScrapeUrlByProcessingType now validates the decoded URL via isValidUrl() and returns 400 for non-URL input, mirroring getScrapeJobsByBaseURL.

- createErrorResponse strips CR/LF and truncates error.message before placing it in the x-error header. Without this, multi-line messages threw ERR_INVALID_CHAR from @adobe/fetch's Headers.set, escaped the catch block, and surfaced as unhandled Lambda 500s — the actual driver of the alert burst.

Adds 4 unit tests covering both behaviors.

Thanks for contributing!


Danialkafeel (approved):
LGTM, @radhikagpt1208 just check the changes in package-lock.json seems irrelevant to this fix

radhikagpt1208:
> LGTM, @radhikagpt1208 just check the changes in package-lock.json seems irrelevant to this fix

@Danialkafeel, The lockfile diff was npm self-heal (`peer:true` flag normalization) introduced when running `npm install` after the main merge. Reverting it to keep the PR diff focused on the scrape controller fix.

PR #2300, opened 2026-04-30
fix(llmo): onboarding hardening — uq_brand_name_per_org, syncBrandSites, atomic brand+brand_sites (LLMO-4129, 4561, 4656, 4350, 4621)

dzehnder:
## Summary

PR-1 of the prompt-generation forward-fix program. Closes the silent-skip and ambiguous-error failure modes that strand brands without prompts on fresh / re-onboarding flows.

Plan: `mysticat-architecture/platform/ops/prompt-generation-forward-fix.md` (PR-1: SpaceCat onboarding hardening).

### Tickets
- **LLMO-4656** — distinguish `(organization_id, name)` unique violation from generic 500. `upsertBrand` and `updateBrand` now throw a structured **HTTP 409** `{ code: 'duplicate_brand_name', existing_brand_id: '<uuid>' }` when Postgres returns 23505 + `uq_brand_name_per_org`. DRS and re-onboarding callers can disambiguate from a transient and auto-merge into the existing brand without an extra round-trip. `existing_brand_id` is best-effort — omitted on the concurrent-hard-delete race where the duplicate row vanishes between the constraint hit and the lookup.
- **LLMO-4350** — cross-org brand URLs are no longer silently dropped. `syncBrandSites` now logs a structured WARN per dropped URL (`event=brand_sites_no_matching_site`) plus a single ERROR summary (`event=brand_sites_cross_org_url_summary`, `dropped_count`, `dropped_urls[≤5]`) when ≥1 URL is dropped. `log` is threaded through `upsertBrand` / `updateBrand`, the v2 onboarding callsite, and the brands controller.
- **LLMO-4621 Pattern A** — brand creation no longer leaves an active brand row with zero `brand_sites` linkage. When none of the submitted URLs match a site row in the org but the brand's `site_id` is known, `syncBrandSites` seeds a fallback `brand_sites` row keyed off `brands.site_id` (`paths=[]`, `type='base'`). Threaded as `opts.fallbackSiteId` from `upsertBrand` / `updateBrand`.
- **LLMO-4129 Bug 3 / LLMO-4561** — pinned the v1-mode contract: `performLlmoOnboarding` in v1 mode does NOT pass `onboarding_mode` to the DRS prompt-generation job. Added explicit assertions in the existing v1 onboarding test so it can't regress (the DRS-side `should_sync_spacecat_v2` gate keys off this metadata; setting it would cross-pollute the v2 customer-config store).

### Scope correction (verified before edits)
The plan claimed `type:'url'` lines at 218/278/1366 of `src/controllers/llmo/llmo-onboarding.js`. Those are already fixed on `origin/main` as `type:'base'` at lines 169/229/1343 (LLMO-4294, commit `ed5363d1`). This PR does not re-touch them.

### Files
- `src/support/brands-storage.js` — `syncBrandSites` (cross-org WARN/ERROR + Pattern A fallback), `upsertBrand` / `updateBrand` (23505 branch, log + fallback threading)
- `src/controllers/brands.js` — pass `log` to `upsertBrand` / `updateBrand`
- `src/controllers/llmo/llmo-onboarding.js` — pass `log` to `upsertBrand` (v2 brand seed)
- `test/support/brands-storage.test.js` — 11 new unit tests (was 79, now 90)
- `test/controllers/llmo/llmo-onboarding.test.js` — explicit `onboarding_mode`-not-set assertion on v1 path

## Validation
- Full suite: **8757 passing**, 100% line / branch / statement coverage.
- Lint: clean.
- Validation gate per plan §3 (post-deploy, 24 h CloudWatch on `/aws/lambda/drs-v2-prod-ProcessJobFunction`):
  ```
  fields @timestamp, @message
  | filter @message like /brand_sites_type_check/
  | stats count(*) as n
  ```
  Must return `n = 0`. New `brand_sites_no_matching_site` events should appear only when expected (multi-tenant URL configured).

## Test plan
- [x] `npx mocha --spec=test/support/brands-storage.test.js` — 90 passing
- [x] `npx mocha --spec=test/controllers/llmo/llmo-onboarding.test.js` — 97 passing
- [x] `npm test` — 8757 passing, 100% coverage
- [x] `npx eslint` — clean on changed files
- [ ] Post-deploy: 24 h CloudWatch query above returns `n=0`
- [ ] Post-deploy: structured WARN `brand_sites_no_matching_site` is observable when triggered (cross-org test)
- [ ] Post-deploy: 23505 + `uq_brand_name_per_org` collision returns 409 with `code=duplicate_brand_name` and `existing_brand_id` set to the conflicting brand UUID (re-onboarding scenario)

## Coordination
- Independent of **PR-2** (`adobe-rnd/llmo-data-retrieval-service` `fix/drs-v2-sync-forward-fix`). Either order is safe; PR-1 first if forced.
- The new `event=` keys (`brand_sites_no_matching_site`, `brand_sites_cross_org_url_summary`, `brand_sites_fallback_to_base_site`) feed alert C6 from the hardening doc — implemented separately in the alert workstream.

### DIFF
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -140,7 +140,17 @@ function BrandsController(ctx, log, env) {
 
   function createErrorResponse(error) {
     if (error.status) {
-      return createResponse({ message: error.message }, error.status, {
+      const body = { message: error.message };
+      // LLMO-4656: surface structured fields on typed errors (e.g. 409 on
+      // duplicate brand name) so DRS / re-onboarding callers can auto-merge
+      // into the existing brand instead of failing the customer at 0 prompts.
+      if (hasText(error.code)) {
+        body.code = error.code;
+      }
+      if (hasText(error.existingBrandId)) {
+        body.existing_brand_id = error.existingBrandId;
+      }
+      return createResponse(body, error.status, {
         [HEADER_ERROR]: error.message,
       });
     }
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -1122,6 +1122,7 @@ function BrandsController(ctx, log, env) {
         brand: brandData,
         postgrestClient,
         updatedBy,
+        log,
       });
 
       return createResponse(created, 201);
@@ -1176,6 +1177,7 @@ function BrandsController(ctx, log, env) {
         updates,
         postgrestClient,
         updatedBy,
+        log,
       });
 
       if (!updated) {
--- a/src/controllers/llmo/llmo-onboarding.js
+++ b/src/controllers/llmo/llmo-onboarding.js
@@ -1345,6 +1345,7 @@ export async function performLlmoOnboarding(params, context, say = () => {}) {
           },
           postgrestClient,
           updatedBy: 'llmo-onboarding',
+          log,
         });
         log.info(`Created initial brand "${brandName}" in normalized table for site ${site.getId()}`);
       } catch (brandError) {
--- a/src/support/brands-storage.js
+++ b/src/support/brands-storage.js
@@ -175,8 +175,32 @@ async function replaceChildRows(table, brandId, rows, onConflict, postgrestClien
 /**
  * Fully replaces brand_sites for a brand. Groups submitted URLs by normalized base URL
  * (via composeBaseURL) so that multiple paths under the same site share one brand_sites row.
+ *
+ * @param {string} organizationId
+ * @param {string} brandId
+ * @param {Array} urls
+ * @param {object} postgrestClient
+ * @param {string} updatedBy
+ * @param {object} [opts]
+ * @param {string} [opts.fallbackSiteId] - Pattern A fallback (LLMO-4621): when no
+ *   submitted URLs resolve to a site row in the org but the caller knows the
+ *   brand's base site (`brands.site_id`), seed a single `brand_sites` row keyed
+ *   off that site so the brand has at least one linkage. Prevents the
+ *   active-brand / no-linkage / 0-prompts failure mode.
+ * @param {object} [opts.log] - Structured logger. Used to emit
+ *   `event=brand_sites_no_matching_site` WARN per dropped URL and a single
+ *   ERROR summary when ≥1 URL is dropped (LLMO-4350).
  */
-async function syncBrandSites(organizationId, brandId, urls, postgrestClient, updatedBy) {
+async function syncBrandSites(
+  organizationId,
+  brandId,
+  urls,
+  postgrestClient,
+  updatedBy,
+  opts = {},
+) {
+  const { fallbackSiteId, log } = opts;
+
   const { error: deleteError } = await postgrestClient
     .from('brand_sites')
     .delete()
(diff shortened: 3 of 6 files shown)


dzehnder (commented):
Hey @dzehnder,

## Strengths
- Forward-fix bundle hits all five tickets cleanly: 23505 disambiguation in `upsertBrand` and `updateBrand` (LLMO-4656), structured WARN+ERROR for cross-org URL drops (LLMO-4350), Pattern A `brands.site_id` fallback (LLMO-4621), v1 onboarding contract pin (LLMO-4129 Bug 3 / LLMO-4561). Scope correction on the `type:'url'` lines being already-shipped is appreciated and correct.
- Excellent defensive logging: every structured-log call is wrapped to prevent a misbehaving logger from breaking brand sync. Two of the new tests pin this behavior end-to-end (`keeps Pattern A fallback resilient when log serialization throws`, `logs cross-org WARN events resiliently when log.warn throws`).
- Test coverage is thorough: 11 new tests including null-sites response, fallback failure, log-throws-on-warn, log-throws-on-info, both `upsertBrand` and `updateBrand` 23505 paths, and the no-fallback-when-neither-baseSiteId-nor-existing.site_id-is-known case.
- `onConflict: 'brand_id,site_id'` on the fallback upsert matches the unique constraint shape and makes the fallback idempotent on re-runs.
- 100% line/branch/statement coverage on changed files; full suite green; integration tests against postgres pass.

## Issues

### Important (Should Fix)

**Cross-PR contract mismatch - A5 alert may not fire**
`src/support/brands-storage.js` (the WARN event name in `syncBrandSites`).

The WARN event emitted is `brand_sites_no_matching_site`, but PR 1507's audit A5 was briefed in the forward-fix plan to filter for `brand_sites_cross_org_url_dropped`. CloudWatch metric filters do pure substring matching, so if A5 ships with the briefed filter, it will silently find zero findings even when this code logs cross-org drops as designed. Two options:
1. Rename the event here to `brand_sites_cross_org_url_dropped` to match the brief.
2. Coordinate with PR 1507 to update A5's filter to match the actual emitted keys (`brand_sites_no_matching_site` for per-URL WARN, `brand_sites_cross_org_url_summary` for summary ERROR).

Option 2 is preferred since `no_matching_site` is more accurate (the URL does not have to be cross-org; it could be a URL that does not resolve to any site in the target org). Either way, the two PRs need to agree on the event name before this closes the loop in production.

**409 missing `existingBrandId` - DRS cannot disambiguate-and-merge**
`src/support/brands-storage.js` (the two 23505 branches in `upsertBrand` and `updateBrand`).

The PR description states the 409 response includes `existing_brand_id`, but the code only sets `err.status = 409` and `err.code = 'duplicate_brand_name'`. DRS callers receiving this 409 have no way to know which brand to merge into without an extra round-trip to SpaceCat. Two options:
1. After catching the 23505, do `SELECT id FROM brands WHERE organization_id = $1 AND lower(name) = lower($2)` (best-effort; can miss in concurrent-deletion races) and attach as `err.existingBrandId`. About 5 LOC in each function.
2. Update the PR description and commit message to drop the `existing_brand_id` claim and document that callers must look up the existing brand themselves.

Option 1 closes the loop without forcing a coordination round-trip on every collision.

### Minor (Nice to Have)

**Test gap: mixed match + drop scenario**
`test/support/brands-storage.test.js`.

The new tests cover the all-match and all-drop cases but not the realistic "some URLs match, others drop" path. Add one test pinning the expected behavior: matched URLs sync into `brand_sites` normally, dropped URLs log WARN + ERROR, no fallback fires.

**Extra DB round-trip on `updateBrand` without `baseSiteId`**
`src/support/brands-storage.js` (in `updateBrand`, the `if (!fallbackSiteId)` block).

When `updates.urls` is set but `updates.baseSiteId` is not, the new code adds a `SELECT site_id FROM brands WHERE id = $1` round-trip. The PATCH response (the `data` variable in the function) already returns the row. If postgrest's update response includes `site_id` (which it does for the existing `select=...`), reading it from there skips the round-trip.

**Defensive log try/catch cluster is broad**
`src/support/brands-storage.js` (multiple sites in `syncBrandSites`).

Each `log.warn(JSON.stringify(...))` is wrapped in `try { ... } catch (e) {}`. The empty catches are intentional, but the cluster is noisy. Consider extracting a helper `safeLog(log, level, payload)` so the defensive-by-design intent is named, not implicit.

## Recommendations
- Add a one-line note in the next release notes flagging the new HTTP 409 `duplicate_brand_name` response shape so other consumers can adopt the disambiguation.
- Consider extracting `brand_sites_no_matching_site`, `brand_sites_cross_org_url_summary`, and `brand_sites_fallback_to_base_site` into a `LOG_EVENTS` constants module so PR 1507's A5 audit can import the keys rather than duplicate literal strings.

## Assessment
Ready to merge: with fixes. The verdict is `REQUEST_CHANGES` but GitHub blocks request-changes on self-authored PRs, so this is posted as a comment with the verdict spelled out here.

The two Important findings are both about cross-PR integration. Local tests pass, but the forward-fix program has gaps if A5's filter does not match what this PR emits, and DRS cannot auto-merge 409 collisions without `existingBrandId`. Address those two and this is good to go.

## Next Steps
1. Resolve the event-name contract with PR 1507 (option 2 in the first Important finding).
2. Add `existingBrandId` to the 409 error path.
3. Optional: add the mixed-case test and consider the helper extraction.

dzehnder:
> [409 missing `existingBrandId` - DRS cannot disambiguate-and-merge](https://github.com/adobe/spacecat-api-service/pull/2300#pullrequestreview-4206710954)

Addressed in 2e16d4f5 (Option 1 — best-effort lookup, ~5 LOC per branch).

- `src/support/brands-storage.js`: both `upsertBrand` and `updateBrand` 23505/`uq_brand_name_per_org` paths now do a `SELECT id FROM brands WHERE (organization_id, name) = ($1, $2)` and attach the result as `err.existingBrandId`. Tolerates the concurrent-hard-delete race — if the lookup is empty the 409 still throws cleanly without the field.
- `src/controllers/brands.js`: `createErrorResponse` surfaces `code` and `existing_brand_id` (snake_case to match API field convention) on the 409 body when present.
- Tests updated: `test/support/brands-storage.test.js` asserts `existingBrandId` on both upsert+update 409 paths plus the empty-post-lookup defensive path; `test/controllers/brands.test.js` adds two end-to-end controller tests pinning the 409 body shape (`code: 'duplicate_brand_name'` + `existing_brand_id`).

Full suite green (8760 passing, 100% coverage on changed files). Did not address the other Important finding (A5 event-name contract with PR 1507) in this fix-up — flagging it for a separate follow-up.

PR #2302, opened 2026-04-30
refactor(plg): split plg-onboarding into named helpers and pure modules

Kanishkavijay39:
## Summary

Refactor of `src/controllers/plg/plg-onboarding.js` (originally 1975 lines: a single 700-line `performAsoPlgOnboarding` orchestrator and a 270-line admin `update` switch) into a sequence of named, self-documented helpers. **No behavior change** — the same 284 unit tests pass after every commit.

## What moved

**Pure helpers** → new modules under `src/controllers/plg/plg-onboarding/`:
- `validation.js` — hostname / SSRF-safety / EDS / AEM-CS regexes
- `internal-org.js` — `isInternalOrg`, `isInternalOrgDemoSite`, `getReviewerIdentity`
- `notifications.js` — `REVIEW_REASONS`, `DOMAIN_*` constants, `deriveCheckKey`
- `displacement.js` — `PLG_OPPORTUNITY_TYPES`, `hasActiveSuggestions`

**Orchestration split** (in-file, not extracted to siblings — see Notes) — `performAsoPlgOnboarding` shrinks from ~700 to ~360 lines and reads top-to-bottom as named phases:
- `applyDeliveryConfig` — preset/auto delivery-type branching
- `enrollPlgConfigHandlers` — summit-plg + auto-suggest/auto-fix loop
- `handleTerminalError` — outer catch (waitlist for `EntitlementWaitlistError`, ERROR otherwise)
- `handleExistingOnboardedDomain` + `revokeDisplacedSiteAsoEnrollment` — one-domain-per-org guard with displacement
- `handlePreonboardedFastPath` — PRE_ONBOARDING fast track
- `persistAndNotify(onboarding, { updatedBy }, context, opts?)` — collapses 10 repeats of the `setUpdatedBy → save → notify` pattern; supports optional `swallowSaveErrors` + `errorLabel` for the two catch-handler call sites

**Admin `update` BYPASS branches** — extracted as module-level handlers; the `switch/case` in `update()` shrinks to four lines:
- `bypassDisplaceOnboarded` (DOMAIN_ALREADY_ONBOARDED_IN_ORG)
- `bypassAemSiteCheck` (AEM_SITE_CHECK)
- `bypassDomainAlreadyAssigned` (DOMAIN_ALREADY_ASSIGNED)

## Commits

```
5975382c refactor(plg): extract update controller's BYPASS branches into named handlers
66d092cd refactor(plg): collapse repeated setUpdatedBy/save/notify into persistAndNotify helper
d03d21f7 refactor(plg): extract handleExistingOnboardedDomain and handlePreonboardedFastPath
53a9ac56 refactor(plg): extract applyDeliveryConfig, enrollPlgConfigHandlers, handleTerminalError
245eabdd refactor(plg): extract pure helpers from plg-onboarding into sibling modules
```

## Test plan

- [x] `npx mocha test/controllers/plg/plg-onboarding.test.js` → 284 passing after each commit
- [x] `npx mocha test/controllers/plg/` → 296 passing (full PLG controller suite)
- [x] `npx eslint src/controllers/plg/plg-onboarding.js src/controllers/plg/plg-onboarding/` → clean
- [ ] Reviewer: verify the diff is mechanical (no semantic changes). Highest-leverage spots to eyeball: Step 5c (delivery config), the displacement guard, the preonboarded fast path, and the admin BYPASS for AEM_SITE_CHECK (siteConfig mutation).
- [ ] If extra confidence wanted: `npx mocha --require test/it/postgres/harness.js --timeout 30000 'test/it/postgres/**/*.test.js'`

## Notes / follow-ups

- Helpers depending on stubbed external packages (`postPlgOnboardingNotification`, `ensureAsoEntitlement`, LD updater, project resolver, `revokeAsoSiteEnrollments`) had to stay in the main file because esmock's 2nd-arg `mocks` only intercepts the entry file's direct imports — moving them to siblings broke 8+ tests. Updating ~8 `esmock(...)` calls to use the 3rd-arg `globalDefs` was tried and reverted; it caused timeouts (over-broad mocking of transitive deps). Worth a follow-up if we want to spin off `entitlements.js` / `launch-darkly.js` modules.
- Two controller-level `postPlgOnboardingNotification` call sites in `update()` were intentionally left untouched (no `updatedBy` in scope; `persistAndNotify` doesn't fit cleanly there).
- Total file size grew slightly (1975 → ~2060 across the family) due to JSDoc on every new helper. Cognitive complexity dropped substantially in exchange.


MysticatBot (dismissed):
Hey @Kanishkavijay39,

## Strengths

- Excellent decomposition of a 1975-line monolith into focused, single-responsibility modules. The module boundaries follow domain concepts (entitlement, LaunchDarkly, notifications, validation, displacement) rather than arbitrary size targets - this is the right heuristic.
- `persistAndNotify` collapses real duplication (10+ repetitions of save-then-notify) into a clean helper with both the happy-path and `swallowSaveErrors` semantics captured.
- All safety-critical logic preserved faithfully: SSRF validation chain (`prepareDomain` -> `isValidDomain` -> `isSafeDomain`), entitlement drift checks, internal-org enrollment refusal guards, and conservative displacement fail-safe (`return true` on lookup error).
- SSRF protection actually improved in the move - the JSDoc now clearly documents the critical ordering contract that prevents scheme-prefixed bypass.
- Commit history is excellent: 5 atomic commits, each reverting independently. Makes regression tracking trivial.
- Backwards-compatible re-export (`export { isSafeDomain } from './plg-onboarding/validation.js'`) ensures external callers are unaffected.
- 284 tests pass unchanged with codecov confirming all modified lines covered.
- The new `plg-onboarding-helper-deps.test.js` specifically tests the dependency fallback contract for the extracted modules.

## Issues

### Important (Should Fix)

**1. `withFlowDeps` mutates the shared request context object** - `src/controllers/plg/plg-onboarding.js:401-428`

`Object.assign(context, { ... })` permanently pollutes the request context with ~25 keys (`badRequest`, `ok`, `Config`, `RUMAPIClient`, etc.). If any middleware, post-handler, or other controller action in the same request lifecycle inspects `context`, it sees unexpected properties. The variable name `flowContext = withFlowDeps(context)` in `transitionStatus` implies a new object, but the mutation returns the same reference.

Today this is safe because controllers are instantiated per request. But it introduces a latent collision risk - if any future middleware adds an `ok` or `Config` key for its own use, the overwrite is silent and hard to debug.

Fix (one line): return a shallow copy instead of mutating in place:
```javascript
function withFlowDeps(context) {
  return { ...context, badRequest, ok, Config, RUMAPIClient, /* ... */ };
}
```
This makes `flowContext` a genuinely separate object, matches the variable-name semantics, and makes the pattern safe for any future context reuse.

**2. Duplicated `parseCommaSeparatedEnvList` and `isInternalOrg`** - `src/controllers/plg/plg-onboarding/internal-org.js:14-18`

`internal-org.js` introduces a private copy of `parseCommaSeparatedEnvList` (already exported from `src/support/utils.js`) and a second active copy of `isInternalOrg` (also in `utils.js`). Having two implementations of a security-sensitive function (it gates enrollment revocation) creates a maintenance risk where a fix to one is missed in the other.

Fix: import from the canonical location:
```javascript
import { parseCommaSeparatedEnvList, isInternalOrg } from '../../../support/utils.js';
```
Or add a deprecation notice in `utils.js` pointing to the new canonical location with a follow-up to migrate other callers.

**3. Dual-source dependency pattern lacks documentation** - `entitlement.js:65`, `launchdarkly.js:97`, `notifications.js:60`

The `context.TierClient || TierClient` pattern (and equivalents for LaunchDarkly and postSlackMessage) makes it unclear from the function signature what is actually used at runtime. A reader must mentally simulate both branches.

Fix: add a brief comment block at the top of the `plg-onboarding/` directory (or in a README) explaining that context injection exists solely for testability with esmock's 2nd-arg limitation, and mark each `context.XClient` access with a `// testability: overridable via context` inline comment.

### Minor (Nice to Have)

**4. Stale cross-reference comment** - `src/controllers/plg/plg-onboarding/displacement.js:14-15`

Comment says "Must stay in sync with LD_AUTO_FIX_FLAGS in plg-onboarding.js", but that constant now lives in `./launchdarkly.js`.

**5. Removed JSDoc body annotation** - diff around line 2003

The `createOnboarding` endpoint lost its `@Body` param doc (`{ imsOrgId, domain, status?, siteId?, ... }`). This was useful for developers discovering the admin endpoint's expected payload shape.

## Recommendations

- Consider adding an `index.js` barrel in `src/controllers/plg/plg-onboarding/` re-exporting the public API so the main controller's import list stays manageable as new helpers are added.
- Track `onboarding-flow.js` (796 lines) for a second-pass extraction of steps 5b-5e into a `resolveAndApplySiteConfig` helper - this would bring the orchestrator down to ~200 lines. Not blocking for this PR given the esmock constraint.
- The displacement race condition (check-then-act is not atomic) is pre-existing and well-documented in the code. Worth a follow-up issue for atomic guard implementation using database-level row locks.

## Assessment

Ready to merge? With fixes

The refactoring is mechanically sound - logic preserved, all tests green, coverage confirmed, security posture maintained. The primary fix needed is the `withFlowDeps` context mutation (item 1), which is a one-line change that eliminates a latent coupling risk in an otherwise clean decomposition. Items 2-3 are low-effort improvements that would strengthen the module boundaries.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 19s | <b>Cost:</b> $6.30 | <b>Commit:</b> <code>3ade4f7bb0e6278886b4b7a62ab736f3cebb22d8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (commented):
Hey @Kanishkavijay39,

## Strengths

- Previously flagged duplication of `parseCommaSeparatedEnvList` and `isInternalOrg` (prior finding 2) cleanly resolved - `internal-org.js:13-14` now imports from the canonical `support/utils.js` and re-exports. Single source of truth, no copy-drift risk.
- Previously flagged stale cross-reference comment (prior finding 4) fixed - `displacement.js:14` now correctly references `launchdarkly.js`.
- Domain-aligned module boundaries follow business concepts (entitlement lifecycle, displacement rules, LaunchDarkly, notifications, validation) rather than arbitrary size targets. Each module has a clear reason to exist.
- SSRF validation chain (`prepareDomain` -> `isValidDomain` -> `isSafeDomain`) preserved at all call sites including the `bypassDomainAlreadyAssigned` alternate-domain path.
- Access control gates remain in the controller layer - no authorization logic delegated to extracted helpers.
- Entitlement mass-revocation guards (org-drift invariant, internal-org check) preserved in `entitlement.js:revokePreviousAsoEnrollmentsForOrg` and the new `revokeDisplacedSiteAsoEnrollment`.
- `persistAndNotify` collapses 10 repetitions of save-then-notify into a clean helper with explicit `swallowSaveErrors` semantics.
- Backwards-compatible re-export (`export { isSafeDomain }`) ensures external callers unaffected.
- 284 unit tests pass unchanged, integration tests green, codecov confirms full coverage on modified lines.
- New `plg-onboarding-helper-deps.test.js` explicitly validates the dependency injection contract for the extracted modules.

## Issues

### Important (Should Fix)

**`withFlowDeps` mutates the request context object in-place** - `src/controllers/plg/plg-onboarding.js:402-429`

`Object.assign(context, { ... })` permanently adds ~25 keys to the request context. The function name `withFlowDeps` and the `return context` pattern suggest to a reader that a new object is being returned (the `with` prefix is a functional-style convention meaning "derive a new thing from this"). In `transitionStatus`, `const flowContext = withFlowDeps(context)` reinforces this assumption - a future developer will reasonably believe `flowContext` is separate from `context`.

Today this is safe because controllers are per-request. The risk is a latent collision: if any future middleware or post-handler adds an `ok` or `Config` key for its own purpose, the overwrite is silent.

Fix (one line, no test changes needed):
```javascript
function withFlowDeps(context) {
  return { ...context, badRequest, ok, Config, RUMAPIClient, /* ... */ };
}
```

Alternatively, if mutation is intentional for a reason not visible in the diff, rename to `injectFlowDeps(context)` so the mutation is in the name, and add a one-line comment documenting the esmock constraint.

## Recommendations

- The `context.TierClient || TierClient` dual-source pattern is pragmatic given the esmock constraint documented in the PR body. A one-line comment at the top of the `plg-onboarding/` directory explaining this convention would save the next contributor 10 minutes of confusion.
- Track `onboarding-flow.js` (796 lines) for a second-pass extraction of the sequential phases into named step functions within the same file. Not blocking - the current state is already a major improvement over the 1975-line monolith.

## Assessment

Ready to merge? Yes

The refactoring is mechanically sound, all tests pass, security invariants are preserved, and module boundaries are well-chosen. The new commit demonstrates responsiveness to review feedback. The `withFlowDeps` mutation is a code clarity concern worth addressing (one-line fix), but it is not a correctness bug given the per-request controller lifecycle. Ship it - the fix can ride as a follow-up commit on this branch or a separate PR.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 9s | <b>Cost:</b> $3.05 | <b>Commit:</b> <code>e4c805c6b487c84f96d6ee2d0d5b3bb769c50a73</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (commented):
Hey @Kanishkavijay39,

## Strengths

- Previously flagged `withFlowDeps` context mutation cleanly resolved - renamed to `injectFlowDeps` at `src/controllers/plg/plg-onboarding.js:402` with a comment documenting the intentional mutation and why it is safe (controllers are per-request). The naming now signals the side-effect.
- Previously flagged duplicate `parseCommaSeparatedEnvList`/`isInternalOrg` resolved - `internal-org.js:13` now imports from the canonical `support/utils.js` and re-exports. Single source of truth.
- Previously flagged stale cross-reference comment fixed - `displacement.js:14` now correctly references `launchdarkly.js`.
- `persistAndNotify` (`notifications.js:123-136`) collapses 10+ repetitions of save-then-notify into a clean helper with explicit `swallowSaveErrors` semantics. The two modes (throw vs. swallow) are clearly documented and the API is minimal.
- Module boundaries follow domain concepts (entitlement lifecycle, LaunchDarkly, displacement, validation, site-setup) rather than arbitrary size targets. Each module has a coherent reason to exist. The dependency graph between modules is acyclic - a healthy property to preserve.
- `handleTerminalError` (`onboarding-flow.js:57-79`) nicely separates terminal-state persistence from the main flow, making the orchestrator's final catch a single delegating line.
- `revokeDisplacedSiteAsoEnrollment` (`entitlement.js:117-146`) extracts the displacement revocation into a named function with clear guard rails (internal-org check, missing-orgId check). Pre-existing safety invariants preserved faithfully.
- Security-critical SSRF validation chain preserved intact with the ordering contract documented in JSDoc (`validation.js:60-76`). Defensive-depth inner checks in `performAsoPlgOnboarding` survive the extraction correctly.
- Backwards-compatible re-export (`export { isSafeDomain }`) at `plg-onboarding.js:400` ensures external callers are unaffected.
- New `plg-onboarding-helper-deps.test.js` explicitly validates the dependency injection/fallback contract for each extracted module.
- PR title follows conventional commits, all conventions respected, Lambda bundle constraints upheld, no stale docs detected.

## Issues

### Important (Should Fix)

**1. Behavioral change in `createdBy` fallback for `getAllOnboardings` response** - `src/controllers/plg/plg-onboarding.js:760`

The old code returned `createdBy` as-is when falsy: `createdBy ? (emailMap[createdBy] ?? createdBy) : createdBy`. This preserved whatever the underlying value was (`undefined`, `null`, empty string).

The new code normalizes to `null`: `createdBy ? (emailMap[createdBy] ?? createdBy) : null`.

For a PR that declares "no behavior change", this is a subtle contract change. Consumers relying on distinguishing `undefined` from `null` will see different response shapes. The same normalization was applied to `updatedBy`.

This is low-risk because the endpoint is admin-only and the DTO layer already normalizes most fields. Suggestion: either revert to the original fallback (`createdBy ? ... : createdBy`) or acknowledge this intentional normalization in the PR body as a minor contract tightening.

**2. Displacement path inlines `disableSummitPlgHandler` logic instead of calling the extracted helper** - `src/controllers/plg/plg-onboarding/onboarding-flow.js:214-222`

The `disableSummitPlgHandler` function was cleanly extracted into `entitlement.js:34-46`. However, `handleExistingOnboardedDomain` re-implements the same logic inline (find site, get Configuration, call `disableHandlerForSite`, save) rather than calling the extracted function.

This creates a maintenance risk: a fix to `disableSummitPlgHandler` (e.g. adding a second handler to disable) won't be picked up by this path.

Fix: Export `disableSummitPlgHandler` from `entitlement.js` (or move it to `site-setup.js` alongside `enrollPlgConfigHandlers` where it conceptually belongs) and call it here:
```javascript
const displacedSite = await Site.findById(alreadyOnboardedSiteId);
if (displacedSite) {
  await disableSummitPlgHandler(displacedSite, context);
}
```

### Minor (Nice to Have)

**3. Misleading log when displacement siteId is absent** - `src/controllers/plg/plg-onboarding/onboarding-flow.js:245-247`

When `alreadyOnboardedSiteId` is falsy, the code logs "skipping displacement and waitlisting {domain}" but does not return - it falls through to the `canDisplace` check which will be `false`, then lands in the waitlist branch. The outcome is correct but the control flow makes the log look like a terminal action when it is not. An explicit early-return or a brief comment would clarify intent.

## Recommendations

- Consider a one-line comment at the top of `plg-onboarding/constants.js` (or `onboarding-flow.js`) explaining the `context.TierClient || TierClient` dual-source pattern - it exists solely for esmock testability. Would save future contributors 5 minutes of wondering.
- `onboarding-flow.js` at 796 lines is within bounds. The sequential phases read well top-to-bottom after the extraction of `applyDeliveryConfig`, `handleExistingOnboardedDomain`, and `handlePreonboardedFastPath`. Worth tracking for a second pass if it grows further.

## Assessment

Ready to merge? Yes

The refactoring is mechanically sound - all tests pass, security invariants preserved, module boundaries well-chosen, and all prior review findings addressed. The two Important items are low-risk improvements (a subtle null normalization in an admin endpoint and a missed DRY opportunity) - neither risks correctness and both can ride as follow-up commits.

## Next Steps

1. Consider addressing the `disableSummitPlgHandler` duplication (item 2) as a quick follow-up commit on this branch.
2. Acknowledge or revert the null normalization in item 1.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 2s | <b>Cost:</b> $5.03 | <b>Commit:</b> <code>d91dc328bc666a3863e799f59a822ab98de611d8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/controllers/plg/plg-onboarding.js
anshulk-public, line 205: @Kanishkavijay39 I think we can move this function also to notification.

anshulk-public (approved):
@Kanishkavijay39 Please resolve one comment, apart from that it is LGTM.

src/controllers/plg/plg-onboarding.js
Kanishkavijay39: Will handle this in separate PR

PR #2305, opened 2026-05-01
feat(s2s): implement readAll capability for cross-tenant list endpoints

ravverma:
Implements the design in docs/s2s/READALL_CAPABILITY_DESIGN.md, opening GET /sites and GET /organizations to S2S consumers that explicitly hold the new `site:readAll` / `organization:readAll` capabilities.

Architecture
- Two-layer enforcement:
  - Layer 1 (s2sAuthWrapper): the route map remaps GET /sites and GET /organizations to require `<entity>:readAll` instead of `<entity>:read`. Tenant-scoped /sites/:siteId stays on `site:read`.
  - Layer 2 (controller): explicit `await accessControlUtil .hasS2SCapability('<entity>:readAll')` opt-in. New list endpoints are admin-only by default until they affirmatively add the gate (fail-closed-by-default).
- `AccessControlUtil.hasS2SCapability` uses `context.s2sConsumer` only for identity (clientId + imsOrgId) and re-fetches the Consumer record from DB for capability validation. Capability data is never read from in-context cached state.
- Capability strings are exported once from src/routes/capability-constants.js. The route map and the controller gates both import from there; a coverage test fails loudly if the two ever drift.
- External response is generic `Forbidden`; denial path is logged internally with admin / s2sReadAll flags for operational debugging.

Tests
- 8 unit tests covering all hasS2SCapability matrix cases.
- 6 sites + 5 organizations controller-gate scenarios.
- Drift coverage test asserts every CAP_ constant is mapped at L1 and referenced by a controller at L2.
- Integration tests: 2 for sites, 2 for organizations. New CONSUMER_2 seed with readAll capabilities; new s2sConsumerReadAll + s2sConsumerReadOnly auth personas mint S2S JWTs against the seeded Consumer rows.

Cross-repo coordination
Depends on adobe/spacecat-shared#feat/consumer-readall-capability (adds 'readAll' to Consumer.CAPABILITIES) — that PR must merge and release before this one can be deployed; otherwise Consumer creation will reject `<entity>:readAll` strings as schema-invalid.

Pagination on the list endpoints is documented as a precondition in the design but is NOT included here.

## Related Issues

Thanks for contributing!


package.json
```diff
@@ -84,7 +84,7 @@
     "@adobe/spacecat-shared-athena-client": "1.9.11",
     "@adobe/spacecat-shared-brand-client": "1.1.41",
     "@adobe/spacecat-shared-content-client": "1.8.23",
-    "@adobe/spacecat-shared-data-access": "3.55.1",
+    "@adobe/spacecat-shared-data-access": "https://gist.github.com/ravverma/73fbd39cad0536f360791528e21c3857/raw/730d09729be240f3918903e6dc27c994228824e4/adobe-spacecat-shared-data-access-3.55.1.tgz",
```
solaris007: **Critical**: This resolves  from a personal GitHub gist instead of the npm registry. The integrity hash differs from the registry's published 3.55.1 - two distinct artifacts share a version label. This is the primary DB-access library; a compromised tarball gets read/write to the entire SpaceCat data plane. Revert to the registry version and hold until the upstream  change publishes through the normal pipeline.

src/routes/required-capabilities.js
```diff
@@ -302,7 +304,9 @@ const routeRequiredCapabilities = {
   'GET /projects/by-project-name/:projectName/sites': 'site:read',
 
   // Sites
-  'GET /sites': 'site:read',
+  // GET /sites is the cross-tenant list endpoint — guarded by site:readAll, not site:read.
```
solaris007: **Critical**: This remap opens  to S2S consumers with , but the design's Cross-Repo Rollout Order requires pagination (step 2) before this remap (step 4). Without cursor pagination, payload cap, and rate limit, a single request can exfiltrate the full customer enrolment list. Hold the remap until pagination ships, or split the PR to land scaffolding without route changes.

src/controllers/sites.js
```diff
@@ -335,6 +341,11 @@ function SitesController(ctx, log, env) {
     const sites = all
       .filter((site) => !EXCLUDED_ORG_IDS.includes(site.getOrganizationId()))
       .map((site) => SiteDto.toListJSON(site));
+
+    if (hasS2SReadAll) {
+      log.info(`[s2s-readall] GET /sites returned ${sites.length} sites to S2S consumer`);
```
solaris007: **Important**: Design requires this log to include , Consumer ID, capability used, response size, and request ID. Current line has only the count and "S2S consumer". Without consumer identity, SOC investigations and per-consumer alerting are impossible.

src/controllers/sites.js
```diff
@@ -315,12 +316,17 @@ function SitesController(ctx, log, env) {
...
    * @returns {Promise<Response>} Array of sites response.
    */
   const getAll = async () => {
-    if (!accessControlUtil.hasAdminAccess()) {
-      return forbidden('Only admins can view all sites');
+    const isAdmin = accessControlUtil.hasAdminAccess();
+    const hasS2SReadAll = !isAdmin && await accessControlUtil.hasS2SCapability(CAP_SITE_READ_ALL);
+    if (!isAdmin && !hasS2SReadAll) {
+      log.info(`[acl] Denied GET /sites — admin=${isAdmin} s2sReadAll=${hasS2SReadAll}`);
+      return forbidden('Forbidden');
```
solaris007: **Important**: Changed from  to . The new message gives the caller nothing to act on. Consider  to match the informative pattern used elsewhere in this controller.

src/support/access-control-util.js
```diff
@@ -97,6 +97,39 @@ export default class AccessControlUtil {
...
+   * @param {string} capability - Full capability string, e.g. 'site:readAll'.
+   * @returns {Promise<boolean>}
+   */
+  async hasS2SCapability(capability) {
+    const { s2sConsumer } = this.context;
+    if (!s2sConsumer) {
+      return false;
+    }
+
+    const fresh = await this.context.dataAccess.Consumer.findByClientIdAndImsOrgId(
```
solaris007, line 126: **Important**:  is a magic string in a security-critical predicate. If the Consumer model renames statuses, this check silently fails open or closed. Import a  constant from , or declare a local constant with a sync-comment.

src/controllers/organizations.js
```diff
@@ -80,16 +81,27 @@ function OrganizationsController(ctx, env) {
...
    * @returns {Promise<Response>} Array of organizations response.
    */
   const getAll = async () => {
-    if (!accessControlUtil.hasAdminAccess()) {
-      return forbidden('Only admins can view all Organizations');
+    const { log } = ctx;
+    const isAdmin = accessControlUtil.hasAdminAccess();
+    const hasS2SReadAll = !isAdmin && await accessControlUtil.hasS2SCapability(CAP_ORG_READ_ALL);
+    if (!isAdmin && !hasS2SReadAll) {
+      log?.info(`[acl] Denied GET /organizations — admin=${isAdmin} s2sReadAll=${hasS2SReadAll}`);
```
solaris007: **Important**:  here vs  in sites.js for the same operation. The optional chain silently drops the security-relevant audit log if  is ever undefined. Drop the  to match sites.js where  is a hard contract on the controller constructor.

solaris007 (changes requested):
Hey @ravverma,

The two-layer enforcement design is well-realized and the test surface is notably broad. Two issues block merge as-is, and several items should be addressed before or alongside landing.

## Strengths

- Two-layer enforcement implemented faithfully: Layer 1 (route map via `src/routes/required-capabilities.js`) and Layer 2 (controller gates in `sites.js` and `organizations.js`) both anchor on the same constants from `src/routes/capability-constants.js`. The fail-closed-by-default property is structural - a controller that forgets to call `hasS2SCapability` keeps admin-only behavior because `hasAdminAccess()` returns false for S2S tokens.
- `src/support/access-control-util.js` `hasS2SCapability` correctly separates identity (from `context.s2sConsumer`) from capabilities (fresh DB fetch via `findByClientIdAndImsOrgId`). A stale or tampered context object cannot grant access. This is the load-bearing invariant of the design.
- `test/routes/capability-constants.test.js` adds bidirectional drift coverage: every `CAP_*` constant must be mapped at Layer 1 AND referenced at Layer 2. The L2-coverage assertion catches the silent-denial-of-intended-access failure mode. This pattern is worth standardizing.
- Test matrix is comprehensive: 8 unit tests for `hasS2SCapability`, 6+5 controller gate scenarios, integration tests with real JWT signing against seeded consumers exercising both positive and negative paths through the actual wrapper. Codecov reports all modified lines covered.
- Generic external `Forbidden` response avoids leaking which auth path was attempted. Internal denial logs preserve debuggability. Short-circuit `!isAdmin && await ...hasS2SCapability(...)` avoids an extra DB lookup for admin callers.
- The design's off-ramp ("if 5+ entities adopt readAll, move to a policy engine") is named explicitly with concrete trigger conditions. Good architectural hygiene.

## Issues

### Critical (Must Fix)

**Dependency sourced from a personal GitHub gist instead of the npm registry**
`package.json:87`, `package-lock.json:24,3791-3792`

`@adobe/spacecat-shared-data-access` is redirected from `registry.npmjs.org` to a personal GitHub gist. The integrity hash differs from the registry's published `3.55.1` - two distinct artifacts share a single version label. This package is the primary database-access library for the service; a compromised tarball gets read/write to every record in the SpaceCat data plane. The artifact lives on a personal GitHub account with no Adobe organizational control - no enterprise SSO, no second-factor enforcement, no admin review. SBOM tooling that joins on `(name, version)` produces wrong provenance for this build.

The design's own Cross-Repo Rollout Order prescribes: (1) merge and release spacecat-shared with `readAll`, (3) bump this repo to the released version, (4) merge this PR. The gist swap circumvents steps 1 and 3.

**Fix**: Revert `package.json` and `package-lock.json` to the registry version. Hold until the upstream change publishes through the normal pipeline. If a local source is needed for development, use `npm pack` + `--no-save` so it never reaches the lockfile in a PR.

---

**Pagination precondition violated - route remap opens unbounded programmatic enumeration on cross-tenant customer data**
`src/routes/required-capabilities.js:235,307`

The design's "Operational Bounds" states: "Pagination is a precondition, not a follow-up... Until pagination ships, the existing endpoints continue to require admin access; S2S readAll is not enabled." The PR body acknowledges this contradiction directly. The data is competitively sensitive per the design's own classification: every customer's `baseURL`, `gitHubURL`, `organizationId`, `imsOrgId`, `name`, Slack channel mappings. A single GET request with a compromised `readAll` Consumer exfiltrates the entire customer enrolment list, in one call, with no rate-limit gate. The existing `// TODO: implement proper pagination` in `sites.js` becomes load-bearing the moment this remap lands.

**Fix options** (in order of preference):
1. Don't merge until cursor pagination, payload cap, and rate limit ship for both endpoints. Match the design's rollout order.
2. Land the auth scaffolding (`hasS2SCapability`, constants, drift test) without remapping the routes - keep `GET /sites` on `site:read` and `GET /organizations` on `organization:read` so S2S enumeration remains admin-only. Remap in a follow-up that ships with pagination.
3. Gate the remap behind a feature flag that is OFF in all environments until pagination ships, with a test asserting the flag-off path returns 403.
4. At minimum, document a written deviation in `READALL_CAPABILITY_DESIGN.md` per workspace rules: "if something cannot be implemented as specced, stop and discuss before deviating."

### Important (Should Fix)

**Audit logging does not match the design contract**
`src/controllers/sites.js:346` and `src/controllers/organizations.js:101`

Current: `[s2s-readall] GET /sites returned ${sites.length} sites to S2S consumer`. Design requires: "logs at minimum: `clientId`, resolved Consumer ID, capability used, response size, request ID." Without consumer identity, per-consumer alerting and SOC investigations are impossible. The denial log is similarly thin - `s2sReadAll=false` without disambiguating "no consumer" vs "revoked" vs "suspended" vs "missing capability".

**Fix**: Add the design-required fields. Consider having `hasS2SCapability` return `{ allowed, reason, consumerId }` instead of a bare boolean so controllers can audit-log without re-reading context.

---

**Error message regressed from actionable to opaque**
`src/controllers/sites.js:329` and `src/controllers/organizations.js:95`

Changed from `'Only admins can view all sites'` to `'Forbidden'`. If a caller hits this branch, something nontrivial happened. A bare "Forbidden" gives the integrator nothing to act on. **Fix**: Return `"Forbidden: admin access or readAll capability required"` or follow the informative pattern used elsewhere in these controllers.

---

**'ACTIVE' is a magic string in a security-critical predicate**
`src/support/access-control-util.js:122`

Hard-coded literal `'ACTIVE'` in a code path that decides cross-tenant read access. If the Consumer model renames statuses, this check silently fails. **Fix**: Import or declare a constant.

---

**Inconsistent log invocation: `log?.info` vs `log.info`**
`src/controllers/organizations.js:94` vs `src/controllers/sites.js:328`

Optional chain in organizations.js silently drops the security-relevant audit log if `log` is ever undefined. Sites.js uses bare `log.info`. **Fix**: Drop the `?.` in organizations.js to match sites.js where `log` is a hard contract.

---

**Test coverage gaps for spec'd matrix cases**
- No unit test for legacy API-key (non-JWT/IMS) caller on the new `getAll` gate. Design matrix specifies "API-key caller (legacy) -> 200, full list". `apikeyAuthAttributes` exists in the test file but is never wired to `getAll()`.
- No IT case for "Consumer absent from Consumer table -> GET /sites with valid token -> 403". This exercises Layer 1's load-bearing trust boundary against the real DB resolver. Current `s2sConsumerReadOnly` tests "wrong capability", not "no record".

**Fix**: Add one API-key `getAll` test per controller. Add a third IT persona with `(unknown-client-id, ORG_1_IMS_ORG_ID)` and assert 403.

### Minor (Nice to Have)

- **Em-dashes in source strings**: Log messages and JSDoc use em-dashes (e.g. `Denied GET /sites - admin=...`). Workspace style is plain hyphen. Replace throughout.
- **Drift test false-negative window**: `test/routes/capability-constants.test.js` checks `controllerSource.includes(name)` which passes if a controller imports a constant but never calls `hasS2SCapability(...)` with it. Tighten to match the actual call site.
- **IT positive assertion is loose**: `test/it/shared/tests/sites.js` asserts `lengthOf.at.least(2)` for readAll while the admin baseline asserts `lengthOf(4)`. Both see the same set. Tighten to match.

- **Design doc belongs in `mysticat-architecture/`**. `docs/s2s/READALL_CAPABILITY_DESIGN.md` spans two repos (`spacecat-shared-data-access` + `spacecat-api-service`) with a cross-repo rollout order. Per the workspace information hierarchy, cross-repo platform specs live in `mysticat-architecture/` so all consuming services can find and own them. Track as a follow-up to relocate.
- **`aud` claim verification gap** in S2S validation. Pre-existing, but this PR raises the consequences: an unverified `aud` now reaches cross-tenant customer data.
- **Capability namespace pollution**: `readAll` in `Consumer.CAPABILITIES` makes every `<entity>:readAll` schema-valid even though only two are intended. The design's 5-entity policy-engine off-ramp should be tracked with an owner.

## Recommendations

- Consider splitting this PR: land the auth scaffolding (`hasS2SCapability`, constants, drift test, all unit tests) as a standalone change that does not remap routes. The scaffolding PR is narrower to review and delivers incremental value without the rollout-order risk.
- Add a CI lint guard: `grep -r 'gist.github.com' package.json package-lock.json` as a pre-merge check.
- Push audit-log formatting into `AccessControlUtil` so format stays consistent when a third `readAll` endpoint lands.

## Assessment

Ready to merge? **No**

Two blockers: the production dependency on a personal gist tarball with a version-string collision, and the route remap shipping ahead of the design's pagination precondition. Both are fixable and the underlying auth design is sound - the PR should land once the supply chain is restored to the registry and the design's rollout order is honored.

## Next Steps

1. Revert `package.json` and `package-lock.json` to the registry version. Hold until the upstream `readAll` change publishes on npm.
2. Either land pagination first (per design), or split the PR to land scaffolding without the route remap.
3. Add the design-required audit-log fields before the route remap reaches production.
4. Address the remaining Important items (error message, magic string, log consistency, test gaps).

ravverma:
Thanks @solaris007 - addressed in `c5fe63c7`. Mapping below.

## Critical
**1. Personal gist tarball** - **Fixed.** Replaced with `@adobe/spacecat-shared-data-access@3.56.1` from the npm registry. The 3.56.1 release adds `readAll` to `Consumer.CAPABILITIES`, so `<entity>:readAll` strings are now schema-valid through the standard publish pipeline. `package.json` and `package-lock.json` updated; `npm install` confirms registry resolution.

**2. Pagination precondition** - Acknowledged but not addressed in this PR. The controller-level checks now produce structured audit logs (clientId, consumerId, capability, count, requestId) so per-consumer alerting is in place if a misconfigured grant slips through. The route remap stays as-is for now; pagination work is tracked separately and will land before any external consumer is granted `readAll`.

## Important
**3. Audit logging fields** - **Fixed.** `hasS2SCapability` now returns `{ allowed, reason, clientId, consumerId }`. Reasons discriminate denial paths: `not-s2s` | `not-found` | `revoked` | `not-active` | `missing-capability` | `granted`. Controllers log:
- Grant: `[s2s-readall] GET /sites granted clientId=X consumerId=Y capability=site:readAll count=N requestId=Z`
- Denial: `[acl] Denied GET /sites - reason=missing-capability clientId=X consumerId=Y requestId=Z`

**4. Error message regression** - **Fixed.** External response now reads `Forbidden: admin access or site:readAll capability required` (and analogous for organizations). Returns actionable info to the integrator without leaking auth-internal details.

**5. Magic 'ACTIVE' string** - **Fixed.** Replaced literal with `ConsumerModel.STATUS.ACTIVE` imported from `@adobe/spacecat-shared-data-access`. Survives any future status enum rename.

**6. log?.info inconsistency** - **Fixed.** Dropped optional chaining in `organizations.js` to match `sites.js`. `log` is a hard contract on the controller constructor; the optional chain was masking a real failure mode.

**7. Test gaps**
- **API-key getAll matrix coverage** - **Added.** New unit test in both `sites.test.js` and `organizations.test.js` that mounts an `api_key` AuthInfo and asserts 200 + full list. Confirms `hasAdminAccess()` legacy bypass still works post-refactor.
- **IT consumer-absent case** - **Added.** New `s2sConsumerUnknown` persona mints a token with clientId `unknown-client-id-999999999999` against `ORG_1_IMS_ORG_ID` (no Consumer row exists). Both `GET /sites` and `GET /organizations` IT cases assert 403 - exercises the load-bearing trust boundary at Layer 1.

## Minor
**8. Drift test false-negative window** - **Fixed.** The drift test now uses regex matching `hasS2SCapability(<CONST>)` and a literal-value variant to assert an actual call site exists, not just an import. Catches the import-rot scenario where the controller's `import { CAP_X }` lingers after the call site is deleted.

**9. Em-dashes** - **Fixed.** Replaced em-dashes with plain hyphens in our added code (capability-constants.js, drift test, log strings, JSDoc on hasS2SCapability, route map comment). Pre-existing em-dashes in untouched code left in place.

**10. IT positive assertion** - **Fixed.** Both IT readAll positive cases tightened from `lengthOf.at.least(N)` to exact `lengthOf` matching the admin baseline (4 sites, 5 organizations).

## Out of scope - acknowledged
- Design doc relocation to `mysticat-architecture/` - tracked as follow-up.
- `aud` claim verification gap - documented in the design's Prerequisites section as a load-bearing assumption.
- Capability namespace pollution + 5-entity policy-engine threshold - documented in the design's Capability Model Trade-offs and Future Extension sections.

## Test status
- 8864+ unit tests passing locally.
- New tests for `hasS2SCapability` return shape, denial reason matrix, audit log format, API-key matrix coverage, drift call-site assertion.

Ready for re-review.

solaris007 (approved):
Hey @ravverma,

All 9 actionable findings addressed and 1 explicitly pushed back with reasonable compensating controls. Approving.

- **Dependency now resolves from npm registry** (`package.json:87`, `package-lock.json:24,3791-3792`). `@adobe/spacecat-shared-data-access@3.56.1` from `registry.npmjs.org` with the registry-published integrity hash. The personal gist URL is gone. Standard publish pipeline restored.
- **Audit log shape matches the design contract** (`src/controllers/sites.js:348`, `src/controllers/organizations.js:104`, `src/support/access-control-util.js:106-149`). `hasS2SCapability` now returns `{ allowed, reason, clientId, consumerId }`. Grant log includes clientId, consumerId, capability, count, requestId. Denial log includes a reason discriminator (`not-s2s` | `not-found` | `revoked` | `not-active` | `missing-capability`) so SOC investigations can pivot per-consumer. This is the right shape and exceeds the original ask.
- **Error message is actionable** (`src/controllers/sites.js:331`, `src/controllers/organizations.js:91`). `Forbidden: admin access or site:readAll capability required` (and analogous for organizations). Tells the integrator what they need.
- **`ACTIVE` magic string replaced** (`src/support/access-control-util.js:14,131`). Imports `ConsumerModel.STATUS.ACTIVE` from `@adobe/spacecat-shared-data-access`. Survives any future status enum rename.
- **Log invocation is consistent** (`src/controllers/organizations.js:90`). Optional chain dropped. Both controllers now use bare `log.info` matching the hard contract.
- **Test coverage gaps closed**. New API-key `getAll` test in `test/controllers/sites.test.js:566-580` and `test/controllers/organizations.test.js:464-478` exercises the legacy admin-bypass path. New `createS2SConsumerUnknownToken` persona in `test/it/shared/auth.js:195-208` plus IT cases in both `tests/sites.js` and `tests/organizations.js` exercise the load-bearing Layer 1 deny path against the real DB resolver.
- **Drift test now asserts a real call site**, not just an import. **Em-dashes removed** from added code. **IT positive assertions tightened** from `at.least` to exact `lengthOf` matching the admin baseline.

## Pushback engaged

**Pagination precondition deferred to a separate PR**

You wrote: "The route remap stays as-is for now; pagination work is tracked separately and will land before any external consumer is granted `readAll`. The controller-level checks now produce structured audit logs (clientId, consumerId, capability, count, requestId) so per-consumer alerting is in place if a misconfigured grant slips through."

Accepting. The compensating controls are real and meaningful:
1. Audit logging gives SOC pivot points if a misconfigured grant ever reaches production.
2. The deviation is now documented in `adobe/mysticat-architecture` PR 45 as an explicit note next to the precondition statement.
3. Operational policy (no external `readAll` grants before pagination) is the gating mechanism.

The remaining gap is that the route remap exists in code without a code-enforced kill switch. If a Consumer record were created with `site:readAll` against this route map (deliberately or otherwise), the unbounded enumeration would happen and the audit log would only catch it after the fact. That risk is now policy-managed rather than code-enforced. Track the policy commitment somewhere durable (Jira ticket or rollout checklist tied to the pagination PR) so it does not decay into folklore.

## Out of scope, worth tracking (carried forward)

- The 4 source docs in `docs/s2s/` (CONSUMER_INTEGRATION_GUIDE.md, READALL_CAPABILITY_DESIGN.md, S2S_ADMIN_GUIDE.md, SECRET_ROTATION_GUIDE.md) still exist alongside the consolidated doc in `mysticat-architecture/platform/decisions/s2s-authentication.md`. The JSDoc references in this PR (e.g. `src/controllers/sites.js:320`) still point at the local path. Worth a follow-up PR to redirect or delete the local copies so the architecture-tier doc is the single source of truth.
- `aud` claim verification gap remains a load-bearing prerequisite per the design doc.
- Capability namespace pollution and the 5-entity policy-engine off-ramp are documented in the design.

## Assessment

Ready to merge? **Yes.**

The auth scaffolding is solid. The supply-chain blocker is gone, the audit-log shape is design-compliant and SOC-pivotable, and the test surface fills the spec'd matrix gaps. Pagination is genuinely separate work and the deviation is documented in the architecture doc. Land it once the pagination ticket is filed and tracked.

PR #2309, opened 2026-05-04
fix: sites resolve PLG gating

Sugandhgyl:

Screenshot 2026-05-05 at 11 16 14 AM

Screenshot 2026-05-05 at 10 05 37 AM

## Related Issues

Thanks for contributing!


package.json
```diff
@@ -22,7 +22,7 @@
     "build": "hedy -v --test-bundle",
     "deploy": "hedy -v --deploy --aws-deploy-bucket=spacecat-prod-deploy --pkgVersion=latest",
     "deploy-stage": "hedy -v --deploy --aws-deploy-bucket=spacecat-stage-deploy --pkgVersion=latest",
-    "deploy-dev": "hedy -v --deploy --pkgVersion=latest$CI_BUILD_NUM -l latest --aws-deploy-bucket=spacecat-dev-deploy --cleanup-ci=24h",
+    "deploy-dev": "hedy -v --deploy --pkgVersion=latest$CI_BUILD_NUM -l sugandhg --aws-deploy-bucket=spacecat-dev-deploy --cleanup-ci=24h",
```
sandsinh: Remove before merge.

sandsinh (approved):

src/controllers/sites.js
```diff
@@ -1184,6 +1197,28 @@ function SitesController(ctx, log, env) {
...
+      { 'x-error': message },
+    );
+
+    // callerImsOrg identifies the *caller* (the org their AEC shell is currently in),
+    // independent of which org's data is being requested via organizationId/imsOrg.
+    // We translate it to a Spacecat UUID once, up front, so the per-path remap can
+    // decide whether the caller is an internal/demo org listed in ASO_PLG_EXCLUDED_ORGS.
+    log.info(`[resolveSite] ASO_PLG_EXCLUDED_ORGS=${context.env.ASO_PLG_EXCLUDED_ORGS ?? '(not set)'} callerImsOrg=${callerImsOrg ?? '(not provided)'}`);
+    let callerIsInternal = false;
+    if (hasText(callerImsOrg)) {
```
Kanishkavijay39, line 1213: Suggestion
src/controllers/sites.js:1212 - caller IMS org lookup is outside the try/catch
                                                    
  Organization.findByImsOrgId(callerImsOrg) is called before the try block that wraps the main resolution logic (the catch block at ~line 1323 handles all other DB failures with badRequest('Failed to resolve site')). If the data layer throws -  
  network blip, driver error - the exception propagates unhandled and returns a raw 500, bypassing the expected error shape.
                                                                                                                                                                                                                                                     
  Move the block inside the existing try, or wrap it in a local try/catch that fails open:                                                                                                                                                           
  try {                                             
    const callerOrg = await Organization.findByImsOrgId(callerImsOrg);                                                                                                                                                                               
    if (callerOrg) {                                                  
      callerIsInternal = isInternalOrg(callerOrg.getId(), context.env);                                                                                                                                                                              
    }                                                                  
  } catch (e) {                                                                                                                                                                                                                                      
    log.warn('[resolveSite] caller org lookup failed, treating as non-internal', e);                                                                                                                                                                 
  }

Kanishkavijay39 (approved):

Kanishkavijay39:
Missing integration tests in test/it/                          
                                                    
  CLAUDE.md requires IT tests for modified endpoints. No new cases were added to test/it/ for the structured 404 responses or the internal-caller remap. The existing unit tests stub findByImsOrgId entirely, so a model-method rename,             
  column-mapping change, or DTO drift would silently break this feature. Add at minimum: external caller + no entitlement, external caller + PRE_ONBOARD tier, and internal caller + no entitlement remapped to site_not_enrolled.
                                                                                                                                                                                                                                                     
   No test for Organization.findByImsOrgId throwing               
                                                    
  All 13 new test cases stub findByImsOrgId to resolve. None reject. Combined with issue 2, this is the highest-risk untested branch. Add a case where findByImsOrgId rejects and assert the response matches the expected badRequest('Failed to     
  resolve site') shape (not an unhandled 500).
                                                                                                                                                                                                                                                     
   callerImsOrg trust model should be documented                                                                                                                                                                                                   
                                                    
  The parameter is read from the query string with no server-side check that the caller actually belongs to that IMS org. A client can pass any internal org's IMS ID and get the "internal" response shape. The security impact is bounded - the    
  remap only narrows 404 detail and cannot widen data access (the 200 path is still gated on accessControlUtil.hasAccess(organization) for the requested org). But this trust assumption should be documented. Either: (a) add a note in the OpenAPI
  description that this param is a UX hint only and not an authorization signal, or (b) consider deriving callerIsInternal from the authenticated principal if the caller's IMS org is available in context.auth. Also add a test confirming that a  
  spoofed callerImsOrg does not change any 200-OK response.         
                                                    
   resolveFailure helper naming invites misuse

  src/controllers/sites.js:1200-1204 - The OpenAPI spec says generic 404s omit resolveStatus and details, but the helper always includes both. If a future maintainer reaches for this helper for a generic not-found case, the contract silently    
  breaks. Rename to resolveStatusFailure to make scope explicit, or have it omit undefined fields.
                                                                                                                                                                                                                                                     
  Minor (Nice to Have)                                              
                                                    
  - src/controllers/sites.js:1252, 1262 - The remap branch log.info lines fire in production on every internal-caller remap. Downgrade to log.debug once the feature is confirmed working.                                                           
  - src/controllers/sites.js:1213, 1215 - Log callerImsOrg via a structured object rather than string interpolation to avoid CRLF log injection: log.info('[resolveSite] caller resolved', { callerImsOrg, callerIsInternal }).
  - src/support/utils.js:1944 - The orphan @returns {Array} JSDoc line above the new helpers looks like it belonged to the function below. Delete or reposition.                                                                                     
  - docs/openapi/sites-api.yaml:302-309 - The description exposes internal implementation details (ASO_PLG_EXCLUDED_ORGS, "Spacecat UUID"). Trim to the client-facing behavior: "IMS Org of the calling shell. Used by internal Adobe surfaces to    
  suppress the PLG onboarding wizard for demo/internal orgs. External callers can omit."                                                                                                                                                             
  - src/controllers/sites.js - The aso_pre_onboard status value is product-prefixed in an otherwise product-agnostic discriminator set. pre_onboard with the product code in details.productCode would avoid a parallel llmo_pre_onboard appearing   
  when the next product reuses this endpoint. Cheaper to change now than after UI clients ship against the string.                                                                                                                                   
  - test/controllers/sites.test.js - Add a test for callerImsOrg: '' confirming it behaves identically to omitting the param.
  - test/controllers/sites.test.js - Add a test for accessControlUtil.hasAccess(organization) returning false to confirm the response is the generic 404 with no details.

PR #2310, opened 2026-05-04
feat: aso PLG excluded orgs bypass

Sugandhgyl:
**Problem**
When an Adobe internal user accesses a pre-onboarded demo site (org has **PRE_ONBOARD** tier with no SiteEnrollment), /sites-resolve returns 404 → the UI shows the PLG onboarding wizard. There is no resolution path that returns 200 for these internal/demo orgs because `getFirstEnrollment() returns {site: null}` and **PRE_ONBOARD** is not in **CUSTOMER_VISIBLE_TIERS**.

**Solution**
Add an early bypass at the top of resolveSite:

- Look up the caller's shell org via Organization.findByImsOrgId(imsOrg)
- If its Spacecat UUID is in ASO_PLG_EXCLUDED_ORGS and hasAccess passes
- Find the site directly (Site.findById(siteId) if provided, else Site.allByOrganizationId) and return 200
- Customer orgs (not in ASO_PLG_EXCLUDED_ORGS) continue through the existing tier/enrollment paths unchanged.

**Test plan**
 Unit tests cover: bypass with siteId, bypass with `allByOrganizationId` fallback, fall-through when no sites exist, non-excluded orgs unaffected
 Manual: access pre-onboarded site as internal org user → renders ASO instead of PLG wizard
 Manual: access as a regular customer org → PLG wizard still shows for `aso_pre_onboard` / `no_entitlement_for_product`

Screenshot 2026-05-04 at 4 08 40 PM

## Related Issues

Thanks for contributing!


src/controllers/sites.js
```diff
@@ -1207,39 +1207,42 @@ function SitesController(ctx, log, env) {
...
+              // 'site_not_enrolled' so the UI shows "No site onboarded" instead of the
+              // PLG onboarding wizard. Internal orgs aren't customers and shouldn't be
+              // prompted to onboard. site_not_enrolled itself is left unchanged.
               if (!entitlement) {
-                return resolveFailure(
-                  'No site found for the provided parameters',
-                  'no_entitlement_for_product',
-                  { productCode, siteId, organizationId: orgId },
-                );
+                if (callerIsInternal) {
```
Kanishkavijay39, line 1232: When we call sites enrollment then never internal org id get passed instead sites org id gets passed.
I think this logic will not work

Kanishkavijay39: Work around here can be when we do site resolve call we can also pass imsorgId when calling from internal org..
Then also handle give priority to org id whenever there in params then imsorg

PR #2316, opened 2026-05-04
feat(llmo): query brand opportunities by scope_id instead of iterating sites

HollywoodTonight:
## Summary

Fixes brand opportunity filtering for organizations where multiple brands share the same site (e.g. `nba.com` with `kings` and `lakers` brands). The previous approach fetched all opportunities per site and filtered by `isElmo` tag, which could not distinguish between brands on the same site.

The `GET /org/:id/brands/:brandId/opportunities` endpoint now calls `Opportunity.allByScope('brand', brandId)` directly, returning only opportunities explicitly tagged for that brand via `scope_type`/`scope_id`. The response payload exposes `scopeType` and `scopeId` (set from the URL parameter, NOT from `opp.getScopeId()` — defense in depth). The `all` path is unchanged and continues to aggregate across all org sites by `isElmo` tag.

`OpportunityDto` was **not** modified. The scope fields are added only to the brand-specific controller response — the shared DTO remains backwards compatible for all other endpoints.

### Defense in depth

- **Tenant-boundary filter:** every opportunity returned by `allByScope` is additionally filtered against `brand.siteIds`. A mis-tagged `scope_id` row (e.g. from a buggy producer) whose `siteId` is not in the brand's canonical site list is dropped with a `log.warn` — the cross-brand contamination this PR fixes on the write path is also blocked on the read path.
- **URL-param `scopeId` echo:** the response always echoes the URL-parameter `scopeId`, never `opp.getScopeId()`. A future refactor to row-data scope would silently allow row claims to override the endpoint's asserted scope; an inline comment plus a regression test guard this invariant.
- **Generic error message** when PostgREST is unavailable — no backend implementation details leak into the response body.

### `all` vs brand-specific divergence (documented in code)

An LLMO opportunity tagged `isElmo` but without a `scope_id` appears in the `all` response but in **no** brand-specific response. The paths are not additive — the sum of every per-brand response does not equal `all`. They converge once every LLMO opportunity carries scope tags. Coverage-report consumers must account for this gap (now documented in the function docblock).

### Companion changes

- **spacecat-shared#1576** — adds the `allByScope` method to `OpportunityCollection`. **Must merge and publish first.** `package.json` here will be bumped to the new version once it ships.
- **spacecat-audit-worker#2427** — writes the `scope_type`/`scope_id` tags on opportunities. Both readers and writers need to be live for the feature to work end-to-end.
- **mystique#1674** — echoes the scope fields back on the response payload.

### Deployment order

1. Merge & publish spacecat-shared#1576.
2. Bump `@adobe/spacecat-shared-data-access` here (currently pinned to 3.55.1) and merge this PR.
3. Bump in spacecat-audit-worker#2427 and merge.

## Test plan

- [x] `npm test -- --grep LlmoOpportunitiesController` — 43 passing, covers orphan-opp filter, `brand.siteIds === undefined` fallback, URL-param scope echo, all-vs-brand divergence.
- [ ] Manual: `GET /org/<orgId>/brands/<brandId>/opportunities` returns only opportunities tagged with that `scope_id` AND whose `siteId` is in `brand.siteIds`.
- [ ] Manual: `GET /org/<orgId>/brands/all/opportunities` returns all LLMO opportunities (unchanged behavior).
- [ ] Manual negative: seed an opportunity with `scope_id=<correct-brand>` but `site_id=<foreign-site>` — confirm it is filtered out of the brand endpoint AND a `log.warn` is emitted.


src/controllers/llmo/opportunities/llmo-opportunities-controller.js
```diff
@@ -172,64 +176,99 @@ function LlmoOpportunitiesController(ctx) {
...
+      const resolveSite = (siteId) => {
+        if (!siteCache.has(siteId)) {
+          siteCache.set(siteId, Site.findById(siteId));
         }
+        return siteCache.get(siteId);
       };
 
-      const results = await processBatch(siteIds, fetchForSite, MAX_CONCURRENT_SITES);
-      const opportunities = results.flat();
+      const opportunities = await Promise.all(
```
grigoreadobe: Three issues at this block - all addressable in one pass:

**Critical - single `Site.findById` rejection 500s the whole request (regression from 'all' path).** The 'all' branch wraps each site fetch in try/catch and degrades to `[]`. Here, one rejection in `Promise.all` propagates to the outer catch and returns 500 even though the opportunity rows loaded fine.

**Critical - unbounded concurrency (regression from 'all' path).** The 'all' branch uses `processBatch(..., MAX_CONCURRENT_SITES)`. This issues one concurrent `Site.findById` per distinct site with no ceiling. For a brand spanning many sites - the exact target case - this fans out to N parallel DB calls.

**Important - cached rejection poisons all subsequent lookups for a site.** A rejected promise stored in `siteCache` means every opportunity pointing at that site also fails.

Recommended fix for all three:
```js
const uniqueSiteIds = [...new Set(filteredOpps.map((o) => o.getSiteId()))];
const siteMap = new Map();
await processBatch(uniqueSiteIds, async (siteId) => {
  try {
    const site = await Site.findById(siteId);
    siteMap.set(siteId, site);
  } catch (siteErr) {
    log.warn(`Failed to fetch site ${siteId}: ${siteErr.message}`);
    siteMap.set(siteId, null);
  }
}, MAX_CONCURRENT_SITES);

const opportunities = filteredOpps.map((opp) => ({
  ...OpportunityDto.toJSON(opp),
  siteBaseURL: siteMap.get(opp.getSiteId())?.getBaseURL() ?? null,
}));
```

src/controllers/llmo/opportunities/llmo-opportunities-controller.js
```diff
@@ -172,64 +176,99 @@ function LlmoOpportunitiesController(ctx) {
...
+        const brandSiteIds = brand.siteIds || [];
+        if (!brandSiteIds.includes(filterSiteId)) {
+          return forbidden('Site does not belong to the organization or brand');
+        }
+      }
+
+      // Query opportunities directly by brand scope — avoids cross-brand contamination
+      // on sites shared by multiple brands (e.g. nba.com with kings and lakers brands)
+      const allScopedOpps = await Opportunity.allByScope('brand', brandId);
+      const filteredOpps = allScopedOpps.filter(
```
grigoreadobe, line 268: **Important - no defense-in-depth against `scope_id` rows pointing at foreign sites.**

When `filterSiteId` is absent, the filter trusts `scope_id` integrity across every writer of the opportunities table. A row with `scope_type='brand', scope_id=<this-brand>, site_id=<unrelated-site>` passes all filters. The previous site-iteration approach was implicitly bounded to org-owned sites; this approach is not.

Fix: build a site set once and add the membership check:
```js
const brandSiteIdSet = new Set(brand.siteIds || []);
const filteredOpps = allScopedOpps.filter(
  (opp) => isLlmoOpportunity(opp)
    && VALID_STATUSES.has(opp.getStatus())
    && (brandSiteIdSet.size === 0 || brandSiteIdSet.has(opp.getSiteId()))
    && (!filterSiteId || opp.getSiteId() === filterSiteId),
);
```

**Also - `filterSiteId` is validated against `brand.siteIds` (PostgREST) but the result filter works on `opp.getSiteId()` (DynamoDB scope). If these two sources drift, the authorization check and the data filter disagree silently.**

test/controllers/llmo/opportunities/llmo-opportunities-controller.test.js
```diff
@@ -470,6 +491,41 @@ describe('LlmoOpportunitiesController', () => {
       expect(body.total).to.equal(1);
       expect(body.opportunities[0].id).to.equal('opp-1');
       expect(body.opportunities[0].siteBaseURL).to.equal('https://brand.com');
+      expect(body.opportunities[0].scopeType).to.equal('brand');
+      expect(body.opportunities[0].scopeId).to.equal('brand-uuid');
+    });
+
+    it('does not cross-contaminate opportunities between brands sharing a site', async () => {
```
grigoreadobe, line 498: **Critical - this test does not distinguish the old buggy behavior from the new fix.**

`allByScope.withArgs('brand', 'brand-a')` is stubbed to return kingsOpp, but `allBySiteId('site-1')` is never seeded with both brands' data. With sinon's `withArgs`, the unconfigured call returns `undefined`. The test passes equally whether the controller uses `allByScope` (new path) or `allBySiteId` (old path) - it cannot detect a regression.

Fix:
1. Stub `allBySiteId('site-1')` to return `[kingsOpp, lakersOpp]` - so the old code path would have returned both.
2. Assert the response contains exactly one opportunity with `id === 'opp-kings'`.
3. Assert `'opp-lakers'` is absent from the response.
4. Assert `expect(mockContext.dataAccess.Opportunity.allBySiteId).not.to.have.been.called`.

Without these, the test is not a regression guard - it only asserts happy-path output, not that cross-contamination is impossible.

src/dto/opportunity.js
```diff
@@ -50,6 +50,8 @@ export const OpportunityDto = {
     guidance: oppty.getGuidance(),
     tags: oppty.getTags(),
     status: oppty.getStatus(),
+    scopeType: oppty.getScopeType?.() ?? null,
```
grigoreadobe: **Important - adding `scopeType`/`scopeId` to the shared DTO exposes these fields on every opportunity endpoint in the service.**

`scope_id` was previously internal. Every endpoint that calls `OpportunityDto.toJSON` - site-scoped, org-scoped, etc. - now returns these fields to callers who may not have been intended to see them. For `scope_type='site'`, this surfaces internal site UUIDs. For future scope types, this is a forward-compat foot-gun.

Consider scoping this to the brand-opportunities path only. Since you know `scope_type='brand'` and `scope_id=brandId` from the query, the getter calls are redundant:
```js
return {
  ...OpportunityDto.toJSON(opp),
  scopeType: 'brand',
  scopeId: brandId,
  siteBaseURL: site?.getBaseURL() ?? null,
};
```

If keeping the shared DTO change, audit every endpoint that calls `OpportunityDto.toJSON` and confirm the disclosure is acceptable at each permission tier. Also add a DTO-level unit test for the `?.() ?? null` fallback with a fixture that has no `getScopeType`/`getScopeId` methods at all (the actual case requiring the optional chain) and one with getters returning `undefined`; both should produce `null`.

grigoreadobe (changes requested):
Hey @HollywoodTonight,

## Strengths

- `llmo-opportunities-controller.js:153-156` - The docblock explains *why* scope-based querying is needed for the multi-brand shared-site case (nba.com / kings / lakers). Future readers have the context without reverse-engineering the intent.
- `llmo-opportunities-controller.js:179-222` - Splitting 'all' and brand-specific into fully separate flows is the right structural call. The prior shared `siteIds` plumbing made both cases harder to reason about.
- `llmo-opportunities-controller.js:246` - Switching from O(N sites) fan-out to a single scoped query is a genuine operational improvement for brands spanning many sites.
- `llmo-opportunities-controller.js:254-260` - The `siteCache` stores the in-flight Promise (not the resolved value), so concurrent lookups for the same site correctly join the same fetch without a race.
- `opportunity.js:53-54` - Optional-chaining with `?? null` keeps the DTO backward-compatible with older model versions during rolling rollout of scope getters.
- The cross-brand isolation test names the exact hazard being guarded ("does not cross-contaminate...") rather than the implementation. Correct framing.

## Issues

### Critical (Must Fix)

**PR depends on an unmerged companion method in spacecat-shared**
`llmo-opportunities-controller.js` (`Opportunity.allByScope` call)

`Opportunity.allByScope('brand', brandId)` is being added in a companion PR to spacecat-shared that is not yet merged. CI passes here because tests stub the method. In production, if this PR deploys before spacecat-shared publishes `allByScope`, the brand opportunities endpoint throws at runtime. "We'll merge them together" coordination is not a safe gate for this.

Fix: gate this PR's merge on the spacecat-shared companion PR being merged and published, and on this repo's `package.json` consuming that version. Document the deployment sequence (spacecat-shared first, then this repo, then audit-worker and Mystique) in the PR description. Also update the description - it references `allByScopeId` but the code calls `allByScope`.

**Single `Site.findById` failure 500s the entire brand response - regression from 'all' path**
`llmo-opportunities-controller.js` (brand path, `Promise.all` block)

The 'all' branch wraps each site fetch in try/catch inside `fetchForSite`, degrades to `[]`, and logs a warn - preserving partial results. The brand branch uses `Promise.all(filteredOpps.map(...))` with no error isolation: one rejected `Site.findById` call propagates to the outer catch and returns 500 for the whole request, even though the opportunity rows loaded fine. A transient DB hiccup on any one site kills every in-flight brand response - stricter behavior than the path it replaced.

Fix: wrap site resolution in per-opportunity try/catch, degrade to `siteBaseURL: null` with `log.warn`, matching the 'all' path. Add a test asserting a single rejected `Site.findById` does not 500 the full request.

**`Site.findById` fan-out no longer concurrency-bounded - regression from 'all' path**
`llmo-opportunities-controller.js` (brand path, `Promise.all`)

The 'all' branch throttles via `processBatch(siteIds, fetchForSite, MAX_CONCURRENT_SITES)`. The brand branch issues one concurrent `Site.findById` per distinct site in the result set with no ceiling. The cache deduplicates per site, but the maximum parallelism is "number of distinct sites across all brand opportunities." For a brand spanning many sites - the exact case this PR targets - this fans out to N parallel DB calls. The explicit constant in the 'all' path signals this was a deliberate design decision to prevent.

Recommended fix (also closes the poisoned-cache issue below): extract unique siteIds from `filteredOpps`, resolve via `processBatch(uniqueSiteIds, (id) => Site.findById(id).catch(() => null), MAX_CONCURRENT_SITES)`, build a `Map<siteId, site>`, map opportunities synchronously. Closes the concurrency regression, error-propagation regression, and poisoned-cache problem in one pass.

**Cross-brand isolation test does not distinguish old behavior from new**
`test/controllers/llmo/opportunities/llmo-opportunities-controller.test.js` ("does not cross-contaminate..." test)

The test stubs `allByScope.withArgs('brand', 'brand-a')` to return kingsOpp, but never seeds `allBySiteId('site-1')` with both brands' data. With sinon's `withArgs`, the unconfigured combo returns `undefined`. The test passes identically whether the controller uses `allByScope` (new path) or `allBySiteId` (old buggy path). It asserts what should be returned, not that contamination is impossible.

Fix: stub `allBySiteId('site-1')` to return `[kingsOpp, lakersOpp]` so the old path would have returned both; assert the response contains exactly `'opp-kings'`, assert `'opp-lakers'` is absent, and assert `Opportunity.allBySiteId` was not called.

### Important (Should Fix)

**No read-time defense-in-depth: opportunities not validated against `brand.siteIds`**
`llmo-opportunities-controller.js` (filter block)

When `filterSiteId` is absent, all scope-tagged opportunities are returned without checking that their `siteId` belongs to the brand. A row with `scope_type='brand', scope_id=<this-brand>, site_id=<unrelated-site>` passes all filters. The previous site-iteration approach was implicitly bounded because it only fetched from org-owned sites. The new approach trusts `scope_id` integrity across every current and future writer of the opportunities table.

Fix: `const brandSiteIdSet = new Set(brand.siteIds || []);` and add `&& brandSiteIdSet.has(opp.getSiteId())` to the filter. Add an integration test seeding an opportunity with a foreign siteId and asserting it is not returned.

**`filterSiteId` validation and result filter use two independent sources of truth**
`llmo-opportunities-controller.js:238-243`

`filterSiteId` is validated against `brand.siteIds` from PostgREST; the result filter is `opp.getSiteId() === filterSiteId` from DynamoDB scope records. If they drift, the authorization check either blocks a valid request (403 for existing data) or admits a site no longer linked to the brand. Pick one authoritative source and apply it consistently to both access control and data filtering.

**Shared `OpportunityDto.toJSON` now exposes `scopeType`/`scopeId` on every opportunity response**
`src/dto/opportunity.js:53-54`

`scope_id` was previously internal. This DTO is shared across all endpoints that return opportunities (site-scoped, org-scoped, etc). Every caller of every such endpoint now sees these fields. For `scope_type='site'`, this surfaces internal site UUIDs to callers who may not have needed them. For future scope types, this is a forward-compat foot-gun.

Consider keeping the shared DTO unchanged and adding these fields only in the brand-opportunities response path - since `scope_type='brand'` and `scope_id=brandId` are known from the query, the getter calls are redundant:
```js
return { ...OpportunityDto.toJSON(opp), scopeType: 'brand', scopeId: brandId, siteBaseURL: ... };
```
If the shared DTO change is intentional, audit every endpoint that calls `OpportunityDto.toJSON` and confirm the disclosure is acceptable at each permission tier.

**Cached rejected promise permanently poisons site lookups**
`llmo-opportunities-controller.js` (`resolveSite`)

A rejected `Site.findById` promise is cached, so every subsequent lookup for that site also rejects - amplifying one transient failure across all opportunities sharing that site. The processBatch fix above eliminates this; if keeping the cache, evict the entry on rejection.

**'all' vs brand-specific semantic divergence is undocumented**
`llmo-opportunities-controller.js`

'all' aggregates by `isLlmoOpportunity` tag across org sites. Brand-specific queries by `scope_type='brand' AND scope_id=brandId`. An LLMO opportunity with no scope tag appears in 'all' but never in any brand call. The sum of brand-specific responses no longer equals 'all'. Consumers assuming this union invariant will see silent gaps. Document the divergence explicitly in the controller header and track migration of the 'all' path to scope-based querying.

**Test gaps: `allByScope` rejection, `allBySiteId` absence, and DTO fallback**

Three gaps: (a) no test asserts 500 + `log.error` when `allByScope` rejects - a future change that returns `[]` instead passes CI silently; (b) no brand-specific test asserts `allBySiteId` was not called, so a future regression to site-iteration passes CI; (c) `getScopeType?.() ?? null` has no test for a model instance that lacks the method entirely. For (c): add a DTO-level test with a fixture that has no `getScopeType`/`getScopeId` properties at all (so `?.` actually short-circuits) and one where getters return `undefined`; both should produce `null`.

### Minor (Nice to Have)

- `'Brand data requires PostgreSQL data service'` leaks the backend implementation to API clients. Prefer a generic message and log the reason server-side.
- `isLlmoOpportunity(opp) && VALID_STATUSES.has(opp.getStatus())` is duplicated across both paths. Extract to a named helper.
- Output shape inconsistency: 'all' drops an opportunity when site is not found; brand path returns it with `siteBaseURL: null`. Pick one contract.
- `getScopeType?.()` shim: add an inline comment noting when it can be removed (once spacecat-shared rollout completes) so it does not calcify as permanent dead code.
- No debug logging in the brand path happy path. A `log.debug` after filtering with the count aids "where did my opportunity go" investigations.

## Recommendations

1. Fix error isolation, concurrency, and poisoned-cache together: replace `Promise.all + resolveSite` with `processBatch` over unique siteIds with per-site try/catch degrading to `null`. One change closes all three issues and restores parity with the 'all' path.
2. Strengthen the cross-brand isolation test - it is the central regression guard for the bug this PR fixes and currently does not distinguish the new code path from the old one.
3. Gate merge on the spacecat-shared companion PR publishing `allByScope` and document the rollout sequence in the PR description (spacecat-shared first, then api-service, then audit-worker and Mystique).
4. Audit `OpportunityDto.toJSON` callers before merging. If the shared DTO change stands, confirm `scopeId`/`scopeType` exposure is acceptable at every permission tier that uses this DTO.

## Assessment

Reasoning: The architectural direction is correct and the bug fix is real. Four things block merge: the companion PR dependency creates a deploy-order landmine; the brand path silently regresses per-site error isolation and concurrency bounding that the 'all' path already enforces; and the central regression test does not actually distinguish the old bug from the new fix. Address these before merge; the defense-in-depth and source-of-truth items are strong should-fix candidates alongside.

src/dto/opportunity.js
rarescheseli: `scopeType` and `scopeId` in the shared DTO expose brand UUIDs on every opportunity endpoint.

`OpportunityDto.toJSON` is consumed by all opportunity endpoints, including site-scoped ones. A user with org-level access to a shared site (e.g. nba.com) calling `GET /sites/{site-1}/opportunities` will now receive every opportunity's `scopeId`, revealing the UUIDs of all brands with scope-tagged opportunities on that site - including brands the user cannot query directly via `/org/:id/brands/:brandId/opportunities`. This is a lateral-visibility channel introduced specifically by this PR in the multi-brand scenario it targets.

Fix: do not add these fields to the shared DTO. Move them to the brand controller's response mapper (the only call site that needs them):
```js
const opportunities = filteredOpps.map((opp) => ({
  ...OpportunityDto.toJSON(opp),
  scopeType: opp.getScopeType?.() ?? null,
  scopeId: opp.getScopeId?.() ?? null,
  siteBaseURL: siteMap.get(opp.getSiteId())?.getBaseURL() ?? null,
}));
```

src/controllers/llmo/opportunities/llmo-opportunities-controller.js
```diff
@@ -172,64 +176,99 @@ function LlmoOpportunitiesController(ctx) {
...
-          brandId, brandName, opportunities: [], total: 0,
-        });
-      }
+      // Query opportunities directly by brand scope — avoids cross-brand contamination
+      // on sites shared by multiple brands (e.g. nba.com with kings and lakers brands)
+      const allScopedOpps = await Opportunity.allByScope('brand', brandId);
+      const filteredOpps = allScopedOpps.filter(
+        (opp) => isLlmoOpportunity(opp)
+          && VALID_STATUSES.has(opp.getStatus())
+          && (!filterSiteId || opp.getSiteId() === filterSiteId),
```
rarescheseli, line 272: Missing `brand.siteIds` membership check and source-of-truth split between PostgREST and DynamoDB.

Two gaps share a single fix. First: there is no check that `opp.getSiteId()` belongs to the brand. A mis-tagged `scope_id` row surfaces to the wrong brand with no second line of defense - the exact cross-brand contamination this PR is fixing on the read path.

Second: `filterSiteId` is authorized against `brand.siteIds` (PostgREST, line 239-241) but the row filter compares `opp.getSiteId()` from DynamoDB. If these stores drift, the authorization gate and the result filter disagree silently.

Fix:
```js
const brandSiteIdSet = new Set(brand.siteIds || []);
// use brandSiteIdSet for the filterSiteId gate above too

const filteredOpps = allScopedOpps.filter(
  (opp) => isLlmoOpportunity(opp)
    && VALID_STATUSES.has(opp.getStatus())
    && brandSiteIdSet.has(opp.getSiteId())
    && (!filterSiteId || opp.getSiteId() === filterSiteId),
);
```
Add a `log.warn` when the membership check drops an opp - it is the observable signal for scope-tagging bugs.

rarescheseli (changes requested):
Hey @HollywoodTonight,

## Strengths

- The prior single-site failure 500 regression is fixed. `Site.findById` calls are now wrapped in per-site try/catch inside `processBatch`, writing `null` to `siteMap` on failure and emitting a `log.warn`. The new "degrades to null siteBaseURL when Site.findById rejects" test locks the contract at `llmo-opportunities-controller.js:241-263`.
- Concurrency is bounded again. `processBatch(uniqueSiteIds, fn, MAX_CONCURRENT_SITES)` over a deduplicated set matches the 'all' path's ceiling and avoids redundant `Site.findById` calls when many opportunities share a site.
- The poisoned-promise cache is gone. `siteMap` is populated after `processBatch` settles; a single site rejection can no longer transitively poison all subsequent reads for that site.
- The cross-brand isolation test now works as a regression guard. `allBySiteId('site-1')` is seeded with `[kingsOpp, lakersOpp]` as a contamination canary, and `expect(...allBySiteId).to.not.have.been.called` proves the controller took the scope-query path. The test would fail against the old code.
- The `allByScope` rejection test correctly asserts both 500 status and `log.error`. The `allBySiteId`-not-called assertion also addresses the previously missing negative assertion.
- The docblock at `llmo-opportunities-controller.js:153-156` names the multi-brand shared-site scenario (nba.com / kings / lakers) clearly.

## Issues

### Critical (Must Fix)

**`package.json` and PR description - `Opportunity.allByScope` has no pinned version in this repo and the description names the wrong method**

`llmo-opportunities-controller.js:247` calls `Opportunity.allByScope('brand', brandId)`, which comes from a companion change in `spacecat-shared`. CI passes here because tests stub `dataAccess.Opportunity.allByScope`. The `package.json` in this PR is unchanged - there is no pinned `@adobe/spacecat-shared-data-access` version that exports `allByScope`. If this PR merges before the shared release ships and is pinned here, every brand-specific request fails at runtime with `TypeError: Opportunity.allByScope is not a function`.

Separately, the PR description still says `Opportunity.allByScopeId('brand', brandId)` while the code calls `allByScope`. Release notes and reviewers searching for `allByScopeId` will not find the call site.

Fixes (all fixable in this PR): (a) bump `@adobe/spacecat-shared-data-access` in `package.json` to the version that exports `allByScope` and update the lockfile; (b) correct the description to say `allByScope`; (c) add a sentence noting the deploy order (shared package releases first, then this repo pins the bump, then this PR merges).

### Important (Should Fix)

**`src/dto/opportunity.js:53-54` - `scopeType` and `scopeId` are now exposed on every opportunity-returning endpoint, not just the brand path**

`OpportunityDto.toJSON` is the shared serializer for all opportunity endpoints, including site-scoped ones. A user with org-level access to `site-1` (nba.com) calling `GET /sites/{site-1}/opportunities` will now receive every opportunity's `scopeId` - revealing the UUIDs of all brands with scope-tagged opportunities on that shared site, including brands that user would not be authorized to query directly via `/org/:id/brands/:brandId/opportunities`. This is a lateral-visibility channel introduced by this PR in exactly the multi-brand scenario the PR targets.

Fix: do not add these fields to the shared DTO. Move them to the brand controller's response mapper, which is the only call site that needs them:
```js
const opportunities = filteredOpps.map((opp) => ({
  ...OpportunityDto.toJSON(opp),
  scopeType: opp.getScopeType?.() ?? null,
  scopeId: opp.getScopeId?.() ?? null,
  siteBaseURL: siteMap.get(opp.getSiteId())?.getBaseURL() ?? null,
}));
```

**`llmo-opportunities-controller.js:248-251` - `brand.siteIds` is not used to validate opportunity membership; filterSiteId validation and row filter use different sources of truth**

Two related gaps share a single fix. First: the main filter has no check that `opp.getSiteId()` is a site the brand actually owns. A row where a writer mis-tagged `scope_id` to the wrong brand surfaces on read with no second line of defense - the exact cross-brand contamination this PR exists to prevent. The previous site-iteration approach implicitly bounded reads to org-owned sites; the scope-query approach does not.

Second: `filterSiteId` is validated against `brand.siteIds` from PostgREST (line 239-241), but the row filter checks `opp.getSiteId()` from DynamoDB scope records. If these two stores drift (a site is removed from `brand.siteIds` but existing DynamoDB rows still reference it), the authorization gate and the result filter disagree silently.

One fix closes both: compute `const brandSiteIdSet = new Set(brand.siteIds || [])` immediately after `getBrandById`, use it for the `filterSiteId` gate, and add `&& brandSiteIdSet.has(opp.getSiteId())` to the row filter. PostgREST becomes the single source of truth for both checks. Add a `log.warn` when the membership check drops an opp - it is the signal that a scope-tagging bug has occurred.

**`src/dto/opportunity.js:53-54` (test coverage) - DTO optional-chain branch is never exercised**

`oppty.getScopeType?.() ?? null` exists to tolerate older model instances during rolling rollout. But `createMockOpportunity` always defines both methods, so the `?.` only ever evaluates the truthy branch (function is present). If a future cleanup removes the optional chaining before the shared package is fully deployed, CI does not catch it.

Fix: add a unit test that constructs an opportunity object with `getScopeType` and `getScopeId` explicitly omitted and asserts `OpportunityDto.toJSON(opp)` (or the controller mapper, if the DTO change is reverted per the finding above) returns `scopeType: null` and `scopeId: null` without throwing.

**`llmo-opportunities-controller.js:153-156` - docblock does not call out the union invariant break**

The docblock explains why scope querying is correct for the multi-brand case but does not document that the 'all' and brand-specific paths are no longer summable. An opportunity tagged `isElmo` with no `scope_id` row appears in the 'all' response but never in any per-brand response. The sum of per-brand responses does not equal the 'all' response. Downstream consumers building dashboards or audit-coverage reports can silently observe a gap.

Fix: add one sentence to the docblock: "Note: the 'all' path returns LLMO opportunities regardless of scope tags, while brand-specific paths require an explicit scope_id; an opportunity without a scope tag appears in 'all' but in no brand-specific response."

### Minor (Nice to Have)

- `llmo-opportunities-controller.js:228` - `'Brand data requires PostgreSQL data service'` leaks the backend implementation name to API clients. Replace with `'Brand lookup service unavailable'` and keep the reason in `log.error`.
- `llmo-opportunities-controller.js:181, 248` - `isLlmoOpportunity(opp) && VALID_STATUSES.has(opp.getStatus())` is duplicated across the 'all' and brand filter predicates. Extract a named helper at module level so the definition cannot silently drift if visibility rules tighten for one path.
- Output shape inconsistency: the 'all' path drops an opportunity entirely when `Site.findById` returns `null`, while the brand path retains it with `siteBaseURL: null`. Pick one contract - retaining with `null` is more debuggable for consumers.

## Recommendations

- Consider a capability check at `LlmoOpportunitiesController` construction time: `if (typeof Opportunity.allByScope !== 'function') log.error(...)`. This turns the deploy-ordering risk into an observable startup signal rather than a per-request 500.
- Add a test asserting `Site.findById` is called at most once per unique siteId when multiple opportunities share a site - the `[...new Set(...)]` de-duplication is a deliberate optimization with no test to prevent regression on refactor.
- Once the spacecat-shared scope-getter migration is complete, the `?.` optional chain can be removed. A TODO comment noting the version cutoff prevents the shim from calcifying into permanent dead code.
- The 'all' and brand paths now use structurally different query strategies. A one-line comment noting that the 'all' path should eventually migrate to `allByScope('org', orgId)` (once all LLMO opportunities have scope tags) makes the intended end-state explicit rather than leaving the dual path as an implicit precedent for future scope types.

## Assessment

Reasoning: The author has cleanly addressed the prior review's most critical regression concerns - error isolation, concurrency bounding, the poisoned cache, and the cross-brand isolation test are all in good shape. The remaining blockers are the unpinned spacecat-shared companion dependency (a deploy-order trap fixable by a version bump and description correction) and the shared DTO change leaking brand UUIDs across all opportunity endpoints (a new disclosure introduced by this PR that should be scoped to the brand controller's response mapper before merge).

## Next Steps

1. Address the Critical: bump `@adobe/spacecat-shared-data-access` in `package.json` to the version exporting `allByScope`, update the lockfile, correct the PR description method name, and document the deploy ordering.
2. Address the Important items: move `scopeType`/`scopeId` out of the shared DTO into the brand controller mapper; add the `brandSiteIdSet` membership check to the row filter; add the DTO optional-chain unit test; add the one-sentence union invariant note to the docblock.
3. The Minor items are optional polish.

grigoreadobe (changes requested):
Hey @HollywoodTonight,

## Strengths

Previously flagged issues now addressed in the 3 new commits:

- \`llmo-opportunities-controller.js\` (brand path) - replacing \`siteCache + Promise.all\` with \`processBatch(uniqueSiteIds, ..., MAX_CONCURRENT_SITES)\` + \`siteMap\` closes all three related problems in one pass: per-site error isolation (prior Critical item 2), unbounded concurrency (prior Critical item 3), and the poisoned-promise cache (prior Important item 4). The approach is clean and matches the 'all' path's pattern.
- \`llmo-opportunities-controller.test.js\` - cross-brand isolation test now seeds \`allBySiteId\` with both brands' data and asserts \`Opportunity.allBySiteId\` was not called. This is the right way to write a regression guard - it fails under the old implementation, not just the new one.
- \`llmo-opportunities-controller.test.js\` - \`returns 500 when allByScope rejects\` and \`degrades to null siteBaseURL when Site.findById rejects in brand path\` fill the primary test gaps from the prior review. Degradation tests for null site and filterSiteId 403 coverage are solid additions.
- \`test/controllers/sites.test.js\` - \`fileURLToPath(new URL(...))\` replacing \`.pathname\` is correct for cross-platform compatibility and fixes a latent Windows/ESM incompatibility.
- \`test/controllers/llmo/llmo-query-handler.test.js\` - the null-sort branch coverage test is a thorough addition; the comment explaining V8's insertion sort order is exactly the kind of context that saves the next reader.

## Issues

### Critical (Must Fix)

**\`Opportunity.allByScope\` does not exist in the published package**

\`@adobe/spacecat-shared-data-access@3.59.0\` - the version pinned in \`package.json\` - has no \`allByScope\` method. The published \`OpportunityCollection\` extends \`BaseCollection\` with no custom methods, and neither \`allByScope\`, \`getScopeType\`, nor \`getScopeId\` appear anywhere in the package. CI passes because every test stubs \`Opportunity.allByScope\`. In production the brand endpoint throws immediately: \`TypeError: Opportunity.allByScope is not a function\`.

This PR cannot deploy safely until: (1) the companion PR that adds \`allByScope\` to spacecat-shared is merged and published, (2) this repo's \`package.json\` is bumped to that version, and (3) the deployment sequence (spacecat-shared first, then api-service) is documented in the PR description. Also fix the PR description, which still references \`allByScopeId\` while the code calls \`allByScope\`.

Note on the optional chaining in \`src/dto/opportunity.js\`: \`getScopeType?.() ?? null\` and \`getScopeId?.() ?? null\` are currently load-bearing - those getters do not exist on any model in the published package. Removing \`?.\` would cause runtime errors on every opportunity response across every endpoint that calls \`OpportunityDto.toJSON\`. This is not temporary dead code; it is the only thing keeping the DTO change from crashing production today.

### Important (Should Fix)

**No read-time siteId validation - orphaned scope tags pass silently**

When \`filterSiteId\` is absent, the filter is \`isLlmoOpportunity(opp) && VALID_STATUSES.has(opp.getStatus())\`. An opportunity row with \`scope_type='brand', scope_id=<this-brand>, site_id=<unrelated-site>\` passes all filters and appears in the brand response. The previous site-iteration approach was implicitly bounded to org-owned sites. The scope-based approach trusts \`scope_id\` integrity across every current and future writer of the opportunities table.

Fix: \`const brandSiteIdSet = new Set(brand.siteIds || [])\` and add \`&& brandSiteIdSet.has(opp.getSiteId())\` to the filter chain.

**\`filterSiteId\` access control and result filter use two independent sources of truth**
\`src/controllers/llmo/opportunities/llmo-opportunities-controller.js\`

\`filterSiteId\` is validated against \`brand.siteIds\` from PostgREST. The result filter uses \`opp.getSiteId() === filterSiteId\` from DynamoDB scope records. If these sources drift - a site removed from the brand in PostgREST while opportunities still reference it - the access control check and the data filter give inconsistent answers. One authoritative source should drive both.

**Shared \`OpportunityDto.toJSON\` now exposes \`scopeType\`/\`scopeId\` on every opportunity-returning endpoint**
\`src/dto/opportunity.js:53-54\`

\`scope_id\` was previously internal. The DTO is shared across site-scoped, org-scoped, and brand-scoped endpoints. Every caller at every permission tier now receives these fields. Before this merges, audit every endpoint that calls \`OpportunityDto.toJSON\` and confirm the disclosure is acceptable at each permission tier. If it is not, keep the shared DTO unchanged and add the fields only in the brand-specific mapping (since \`scope_type='brand'\` and \`scope_id=brandId\` are already known from the query, the getter calls would be redundant anyway):
\`\`\`js
return { ...OpportunityDto.toJSON(opp), scopeType: 'brand', scopeId: brandId, siteBaseURL: ... };
\`\`\`

**'all' vs brand path drop-vs-include divergence is untested and undocumented**
\`src/controllers/llmo/opportunities/llmo-opportunities-controller.js\`

'all' returns \`[]\` from \`fetchForSite\` when \`Site.findById\` returns null, dropping those opportunities entirely. The brand path returns the opportunity with \`siteBaseURL: null\`. A consumer querying both endpoints for the same org will see different opportunity counts for the same data. There is no test asserting this behavioral difference and the controller comment does not document it. Add a test and document the contract explicitly.

### Minor (Nice to Have)

- \`'Brand data requires PostgreSQL data service'\` leaks backend implementation details to API clients. Prefer a generic message and log the internal reason server-side.
- \`isLlmoOpportunity(opp) && VALID_STATUSES.has(opp.getStatus())\` is duplicated between the 'all' path and brand path. Extract to a named helper to keep both paths in sync as valid statuses evolve.
- Prior gap of testing \`getScopeType?.() ?? null\` with a model that has no such method is still absent. Worth a dedicated DTO-level test once scope getters are published in the shared package.

## Recommendations

1. Before merging: bump \`@adobe/spacecat-shared-data-access\` to the version that includes \`allByScope\`, \`getScopeType\`, and \`getScopeId\`. Update the PR description to reference \`allByScope\` (not \`allByScopeId\`) and document the deployment sequence (spacecat-shared first, then api-service).
2. Add \`brandSiteIdSet.has(opp.getSiteId())\` to the brand-path filter for defense in depth against orphaned scope tags.
3. Audit \`OpportunityDto.toJSON\` callers and confirm \`scopeType\`/\`scopeId\` exposure is acceptable at each permission tier before merging.
4. Add a comment to the controller documenting the 'all' vs brand-specific behavioral contract (drop vs null), and add a test asserting it.

## Assessment

The new commits resolve the per-site error isolation regression, the concurrency regression, the poisoned cache, and the key cross-brand test gaps from the prior review - all correctly. One blocker remains: \`Opportunity.allByScope\` is not in the published package this repo depends on, and CI's green state is entirely due to stubbing. Fix the package dependency and address the Important items before merge; the core architecture of the change is sound.

src/controllers/llmo/opportunities/llmo-opportunities-controller.js
```diff
@@ -172,64 +176,114 @@ function LlmoOpportunitiesController(ctx) {
...
 
-      if (siteIds.length === 0) {
-        return ok({
-          brandId, brandName, opportunities: [], total: 0,
-        });
+      // Query opportunities directly by brand scope — avoids cross-brand contamination
+      // on sites shared by multiple brands (e.g. nba.com with kings and lakers brands).
+      // NOTE: allByScope is not yet published in @adobe/spacecat-shared-data-access; the
+      // test suite stubs this method. Remove this note once the published version ships.
+      const allScopedOpps = await Opportunity.allByScope('brand', brandId);
```
rarescheseli, line 259: `Opportunity.allByScope` is not in `@adobe/spacecat-shared-data-access@3.59.0` (verified against the published package and upstream main at `@3.60.0`). `OpportunityCollection` extends `BaseCollection` with no custom methods. Tests pass because they stub this method. In production, this line throws `TypeError: Opportunity.allByScope is not a function`, caught by the outer try/catch and returned as `internalServerError(error.message)` - echoing the internal method name to the client.

The NOTE comment at lines 247-248 acknowledges this but is not a fix. This is the third raise. Acceptable resolutions:

- (a) Bump `@adobe/spacecat-shared-data-access` to the version that exports `allByScope`, update the lockfile, and document the deployment order (shared package first, then this repo) in the PR description.
- (b) If the upstream dep cannot be bumped yet: `if (typeof Opportunity.allByScope !== 'function') { return internalServerError('brand scope query not yet available'); }` - removes the fingerprintable TypeError echo and keeps the endpoint from throwing on a missing method.

Also fix the PR description: it still names the method `allByScopeId` (the code calls `allByScope`) and still states "Also exposes scopeType and scopeId fields in OpportunityDto.toJSON()" - which this commit reversed.

src/controllers/llmo/opportunities/llmo-opportunities-controller.js
```diff
@@ -172,64 +176,114 @@ function LlmoOpportunitiesController(ctx) {
...
+        return badRequest('Brand data requires PostgreSQL data service');
+      }
+
+      const brand = await getBrandById(orgId, brandId, postgrestClient);
+      if (!brand) {
+        return notFound(`Brand not found: ${brandId}`);
       }
 
+      const brandName = brand.name;
+      const brandSiteIdSet = new Set(brand.siteIds || []);
```
rarescheseli, line 247: **`brand.siteIds === null/undefined` branch and the all-orphan case are not covered by tests.**

`new Set(brand.siteIds || [])` is safe when `siteIds` is `undefined` or `null`, but existing empty-siteIds tests pass `siteIds: []`, never reaching the `||` fallback. A freshly created brand may return `siteIds: undefined` from `getBrandById`.

One test covers both gaps: resolve `getBrandById` with no `siteIds` key and have `allByScope` return one opp whose `siteId` is not in the (empty) set. Assert `total: 0`, `opportunities: []`, and `log.warn` fired with a count of 1. This is also the all-orphan scenario - the realistic post-reassignment shape that motivates this feature.

src/controllers/llmo/opportunities/llmo-opportunities-controller.js
```diff
@@ -150,6 +150,10 @@ function LlmoOpportunitiesController(ctx) {
    * GET /org/:spaceCatId/brands/all/opportunities
    * Returns all LLMO opportunities for sites under the given brand (or all org sites).
    * Accepts optional ?siteId query parameter to scope results to a single site.
+   *
+   * For a specific brandId, queries by scope_type='brand' AND scope_id=brandId directly —
```
rarescheseli, line 154: **The `all` and brand-specific paths are semantically different and that is not documented.**

The `brandId === 'all'` path aggregates by `isLlmoOpportunity` tag across all org sites; this path queries by `(scope_type='brand', scope_id=brandId)`. An LLMO opportunity without a scope tag appears in `all` but in no brand-specific response. The sum of per-brand responses does not equal `all`. Consumers building dashboards or coverage reports will see a silent gap.

Add a sentence here: _"An opportunity without a scope tag appears in the `all` response but no brand-specific response; these paths are not additive and are expected to converge once all LLMO opportunities carry scope tags."_ Also consider a TODO linking a tracking issue for converging the `all` path onto `allByScope('org', orgId)` once scope tagging is complete.

test/controllers/llmo/opportunities/llmo-opportunities-controller.test.js
```diff
@@ -442,13 +460,16 @@ describe('LlmoOpportunitiesController', () => {
       expect(body.brandName).to.equal('TestBrand');
     });
 
-    it('returns opportunities for a specific brand', async () => {
+    it('returns opportunities for a specific brand using scope query', async () => {
```
rarescheseli, line 463: **No test distinguishes URL-derived `scopeId` from opp-derived `scopeId`.**

Line 284 of the controller uses `scopeId: brandId` (the validated URL param), not `opp.getScopeId()`. Every brand-path test that asserts `scopeId` constructs the mock opp with the same value as `params.brandId`, so a regression to `opp.getScopeId()` would pass all existing tests undetected.

Add one test: `createMockOpportunity({ scopeId: 'wrong-brand' })` returned by `allByScope`, controller called with `params.brandId = 'correct-brand'`. Assert the response shows `scopeId: 'correct-brand'`. This protects the invariant that scope is asserted by the endpoint, not echoed from the row.

rarescheseli (changes requested):
Hey @HollywoodTonight,

## Strengths

- `src/dto/opportunity.js`: `scopeType`/`scopeId` removed from the shared DTO and the docblock explicitly records why (cross-tenant UUID exposure on shared sites). All five other `OpportunityDto.toJSON` call sites in `src/controllers/opportunities.js` were audited and confirm no functional regression from the removal - those endpoints are site-scoped and never needed these fields.
- `src/controllers/llmo/opportunities/llmo-opportunities-controller.js:237` - `brandSiteIdSet` computed once and threaded through the `filterSiteId` gate, orphan detection, and result filter. This collapses the two-independent-source problem from the prior review into a single Set with O(1) membership lookups.
- Lines 283-284: `scopeType: 'brand'` and `scopeId: brandId` attached from the validated URL parameter, not from the opportunity row. The endpoint knows its own scope type and has already authorized `brandId` via `getBrandById`.
- Test `does not expose scopeType or scopeId in the all-brand (non-scoped) response` (line 896) asserts property-absence on the `all` path. `returns opportunities for a specific brand using scope query` (line 463) now pins both fields on the brand path. Together these are the right regression guards.
- Orphan detection at lines 253-255 surfaces a real consistency condition (site reassigned without scope cleanup) rather than silently dropping rows.

## Issues

### Critical (Must Fix)

See inline comment at line 249.

### Important (Should Fix)

See inline comments at lines 237, 154 (controller), and line 463 (test file).

### Minor (Nice to Have)

- `llmo-opportunities-controller.js:255`: Orphan log interpolates into a message string. Consider structured fields instead: `log.warn('brand scope orphans detected', { brandId, count: orphanedOpps.length })`. Same volume, queryable in log tooling, and separates diagnostic keys from prose. Do not remove the signal - it is the only forensic breadcrumb for the site-reassignment-without-scope-cleanup scenario.
- `llmo-opportunities-controller.js:219-249`: The safety of calling `Opportunity.allByScope('brand', brandId)` depends on `getBrandById(orgId, brandId, ...)` having just confirmed `brandId` belongs to the caller's org. A one-line comment noting this - "`brandId` is trusted here only because `getBrandById` above scoped it to `orgId`" - protects the invariant if `getBrandById` is ever refactored.

## Recommendations

- Extract `'brand'` as a named constant (`SCOPE_TYPE_BRAND = 'brand'`). The string appears at line 249 (query) and line 283 (response tag); string drift between the two would be a silent bug.
- The `getScopeType`/`getScopeId` getters defined on `createMockOpportunity` (test file lines 38-39) and the `scopeType`/`scopeId` params passed to it throughout the test suite are now inert - production code no longer calls either getter via the DTO. Either delete the dead test infrastructure or add the test from the Important finding at line 463, which would actually exercise the URL-vs-opp divergence.
- Once the shared-package rollout is complete, the NOTE at line 247 should be deleted as part of the dep bump. Consider a GitHub issue link in the comment rather than free-form prose so there is a tracked owner.

## Assessment

Ready to merge? No, with fixes.

Reasoning: The security work in this commit is complete and correctly implemented - the DTO boundary is clean, audited against all callers, and locked by tests. The remaining blocker is the same as the prior two reviews: `Opportunity.allByScope` is not callable on the pinned dependency and the NOTE comment does not change that. The two Important test gaps leave the highest-value behavioral contracts of this commit unprotected against regression.

## Next Steps

1. Address the Critical: bump `@adobe/spacecat-shared-data-access` (or add a runtime guard) and update the PR description to use `allByScope` and drop the reversed OpportunityDto statement.
2. Add the two Important test cases: URL-vs-opp `scopeId` distinction, and `null`/`undefined` `siteIds` with an all-orphan `allByScope` result.
3. Add the one-sentence docblock note on the `all`-vs-brand semantic divergence.
4. Minor items are optional polish.

src/controllers/llmo/opportunities/llmo-opportunities-controller.js
```diff
@@ -172,64 +183,122 @@ function LlmoOpportunitiesController(ctx) {
...
 
-      if (siteIds.length === 0) {
-        return ok({
-          brandId, brandName, opportunities: [], total: 0,
-        });
+      // Query opportunities directly by brand scope — avoids cross-brand contamination
+      // on sites shared by multiple brands (e.g. nba.com with kings and lakers brands).
+      // NOTE: allByScope is not yet published in @adobe/spacecat-shared-data-access; the
+      // test suite stubs this method. Remove this note once the published version ships.
+      const allScopedOpps = await Opportunity.allByScope('brand', brandId);
```
rarescheseli, line 259: `Opportunity.allByScope` is not in any published `@adobe/spacecat-shared-data-access` version. `package.json` pins `3.55.1`; the method is absent through `3.60.0` (current latest). In production, this line throws `TypeError: Opportunity.allByScope is not a function`, caught at line 306, and passed to `internalServerError(error.message)` - which echoes the method name into both the JSON response body and the `x-error` header. Any authenticated user who triggers this endpoint receives `{"message":"Opportunity.allByScope is not a function"}`, confirming the backend model-class naming convention.

The NOTE comment above acknowledges this but does not prevent the crash or the disclosure. This has been raised in every prior review with no pushback.

Two fixes, both in-scope:

**(a)** Bump `@adobe/spacecat-shared-data-access` in `package.json` and `package-lock.json` to the version exporting `allByScope` once spacecat-shared#1576 publishes. Block merge until that version exists.

**(b)** Add a runtime capability check immediately before this line:
```js
if (typeof Opportunity.allByScope !== 'function') {
  log.error('Opportunity.allByScope unavailable; spacecat-shared upgrade required');
  return internalServerError('Brand opportunity service unavailable');
}
```
This converts the TypeError into a clean 500 with no method-name disclosure and makes the four-repo deploy chain order-independent.

Preferred: (a) for correctness, (b) as deploy-order insurance. At minimum one must be present.

Also: correct the PR description - it still says `allByScopeId` (the code calls `allByScope`) and still says "Also exposes scopeType and scopeId fields in `OpportunityDto.toJSON()`", which this PR reversed.

test/controllers/llmo/opportunities/llmo-opportunities-controller.test.js
```diff
@@ -639,5 +862,125 @@ describe('LlmoOpportunitiesController', () => {
...
+      const result = await controller.getBrandOpportunities(mockContext);
+
+      expect(result.status).to.equal(200);
+      const body = await result.json();
+      expect(body.total).to.equal(1);
+      expect(body.opportunities[0].id).to.equal('opp-valid');
+      expect(mockContext.log.warn).to.have.been.calledWith(sinon.match(/not in brand\.siteIds/));
+    });
+
+    it('uses URL-param scopeId, not opp.getScopeId() (trust-boundary regression guard)', async () => {
```
rarescheseli, line 896: This test guards against a regression to `scopeId: opp.getScopeId()` but misses the other failure mode: if `OpportunityDto.toJSON` reverts to emitting `scopeId`, the `{ ...OpportunityDto.toJSON(opp), scopeId: brandId }` spread-plus-override still produces `'correct-brand'` and the test stays green. A contributor restoring the DTO field would ship a lateral UUID disclosure channel with a green CI.

There is also no `result.status` assertion before `body.opportunities[0]`. If any runtime error fires inside the brand path, `body.opportunities` is `undefined` and the test fails with a misleading `TypeError: Cannot read property '0' of undefined` rather than a clear scope-integrity signal.

Two additions close both gaps:

1. Add `expect(result.status).to.equal(200)` and `expect(body.opportunities).to.have.length(1)` before the scope assertions.
2. Add a direct call to `OpportunityDto.toJSON(opp)` asserting the result does NOT include a `scopeId` key - this pins the DTO half of the contract alongside the controller half.

src/controllers/llmo/opportunities/llmo-opportunities-controller.js
```diff
@@ -172,64 +183,122 @@ function LlmoOpportunitiesController(ctx) {
...
+
+      // Detect orphans — opps whose siteId is no longer in the brand's canonical site list.
+      // These arise when a site is reassigned to a different brand without a scope cleanup.
+      const orphanedOpps = allScopedOpps.filter((opp) => !brandSiteIdSet.has(opp.getSiteId()));
```
rarescheseli, line 263: Orphan detection runs on the unfiltered `allScopedOpps` set, which includes non-LLMO opps (wrong tags) and opps in RESOLVED/IGNORED status. A non-LLMO opp with a stale `siteId` inflates the orphan count and fires this `log.warn` even though that opp would have been dropped by the `isLlmoOpportunity && VALID_STATUSES` filter below. The orphan log is the only forensic signal for scope-tagging drift - if it fires for opps the controller was never going to return, the alert is misleading.

Fix: apply the LLMO + status filter first, then run orphan detection on the eligible set:

```js
const activeLlmoOpps = allScopedOpps.filter(
  (opp) => isLlmoOpportunity(opp) && VALID_STATUSES.has(opp.getStatus())
);
const orphanedOpps = activeLlmoOpps.filter((opp) => !brandSiteIdSet.has(opp.getSiteId()));
if (orphanedOpps.length > 0) {
  log.warn(`Brand ${brandId} has ${orphanedOpps.length} scoped opportunities with siteIds not in brand.siteIds`);
}
const filteredOpps = activeLlmoOpps.filter(
  (opp) => brandSiteIdSet.has(opp.getSiteId()) && (!filterSiteId || opp.getSiteId() === filterSiteId)
);
```

This also eliminates the duplicated `isLlmoOpportunity(opp) && VALID_STATUSES.has(opp.getStatus())` predicate currently at both the `all` path and the brand path.

rarescheseli (changes requested):
Hey @HollywoodTonight,

## Strengths

- `brandSiteIdSet` unifying the `filterSiteId` gate and the row filter into a single source of truth (PostgREST `brand.siteIds`)
- `scopeType`/`scopeId` removed from shared `OpportunityDto.toJSON` - the DTO docblock at `src/dto/opportunity.js:20-23` records the reason so future contributors know what they would be restoring
- `processBatch + siteMap` replacing the poisoned-promise cache with bounded concurrency and per-site error isolation

Specific to the new commit (`5115c76f`):

- The `IMPORTANT` block at `llmo-opportunities-controller.js:292-295` documents the URL-vs-row `scopeId` trust boundary exactly where a future refactor would look, paired with a regression test that catches the "just use `opp.getScopeId()`" shortcut
- The `all` vs brand-specific divergence NOTE at lines 158-163 names the non-additivity explicitly - saves the next person building a coverage dashboard from having to rediscover it
- Error message separation at lines 234-238: client sees "Brand lookup service unavailable", server logs the PostgREST configuration reason - the right pattern, applied consistently
- The `brand.siteIds === undefined` regression test correctly distinguishes the missing key from `siteIds: []`, exercising the `|| []` fallback that prior tests never reached
- Authorization order is correct: `getBrandById(orgId, brandId, postgrestClient)` gates on both `organization_id=orgId` and `id=brandId`, so a caller knowing a brandId UUID from another org cannot pivot
- Orphan detection logs a count rather than the stale siteIds themselves, avoiding cross-brand siteId disclosure through log siphoning

## Issues

### Critical (Must Fix)

**`Opportunity.allByScope` is not in any published `@adobe/spacecat-shared-data-access` version** - see inline comment at line 259.

`package.json` pins `3.55.1`; the method is absent through `3.60.0`. In production, every brand-specific request throws `TypeError: Opportunity.allByScope is not a function`, caught at the outer catch (line 306), and returned via `internalServerError(error.message)` - which echoes the method name into both the JSON body and the `x-error` header. The NOTE comment at lines 257-258 acknowledges the gap but does not prevent the crash or the disclosure.

### Important (Should Fix)

Three items, details in inline comments:

1. **Generic catch handler at line 308 echoes raw `error.message` for any exception in the brand branch** - PostgREST driver errors, constraint names, and internal details all end up in the response body. The pattern the new commit introduced at lines 234-238 (log server-side, return fixed client string) should be applied here too.

2. **Trust-boundary regression test at line 896 guards only half the invariant** - it catches a regression to `opp.getScopeId()` but not a DTO revert. If `OpportunityDto.toJSON` re-adds `scopeId`, the spread-plus-override still produces the correct value and the test stays green. Also missing a `result.status` assertion before `body.opportunities[0]`.

3. **Orphan detection at line 263 runs on the unfiltered `allScopedOpps` set** - non-LLMO opps and opps in RESOLVED/IGNORED status inflate the count and fire the warn even though those opps would be dropped by the subsequent filter. The orphan log is the only forensic signal for scope-tagging drift; it should count what the response actually drops.

### Minor (Nice to Have)

- `llmo-opportunities-controller.js:265`: Orphan log text says "siteIds not in brand.siteIds" but when `brand.siteIds` is `undefined` (which the new test explicitly exercises), the property does not exist on the record. Switch to structured fields: `log.warn('brand scope orphans filtered', { brandId, orphanCount: orphanedOpps.length, sampleSiteIds: orphanedOpps.slice(0, 5).map((o) => o.getSiteId()) })`.
- Test file line 893: `calledWith(sinon.match(/not in brand\.siteIds/))` does not pin call count. Add `.calledOnce` to catch a per-opp regression.

## Recommendations

- Once the orphan-detection ordering fix is applied, extract `isLlmoOpportunity(opp) && VALID_STATUSES.has(opp.getStatus())` into a named helper (`isActiveLlmoOpp`). It appears at lines 199 and 262 and represents the canonical definition of an active LLMO opportunity - the kind of rule you want defined in one place.
- Add the runtime capability check for `Opportunity.allByScope` even after the dep is bumped - it makes the four-repo deploy chain (spacecat-shared -> this repo -> audit-worker -> mystique) order-independent so partial rollbacks degrade cleanly.

## Assessment

Ready to merge? **No - with fixes.**

Reasoning: The new commit's documentation and test additions are precise and close the gaps the prior review asked for. Two things block merge: `allByScope` is not in the pinned dependency, causing a production crash that leaks the backend method name on every brand request; and the generic catch handler applies `internalServerError(error.message)` to any exception in the brand branch - the exact information-disclosure pattern this commit correctly fixed for the not-configured case at lines 234-238.

## Next Steps

1. Address the Critical: bump `@adobe/spacecat-shared-data-access` to the version exporting `allByScope` (update lockfile), or add a runtime capability check with a fixed client message - ideally both. Correct the PR description (`allByScope` not `allByScopeId`; remove the reversed `OpportunityDto.toJSON` statement).
2. Address the three Important items: fix the catch handler to return a fixed client message; strengthen the trust-boundary test with a status assertion and DTO assertion; move orphan detection to run on the LLMO-eligible subset.
3. Minor items are optional polish.

HollywoodTonight:
allByScope is obviously introduced by the spacecat-shared-pr https://github.com/adobe/spacecat-shared/pull/1576/changes#diff-ce1c352a9919e7040636b9ca09977b5a6db4dba231038154f7f000a36dc7c15bR56

src/controllers/llmo/opportunities/llmo-opportunities-controller.js
rarescheseli, line 259: `Opportunity.allByScope` is not callable on the pinned dependency. `package.json` pins `@adobe/spacecat-shared-data-access@3.60.0`; the companion `adobe/spacecat-shared` pull request 1576 is OPEN and unmerged; no published version exports `allByScope`. In production this line throws `TypeError: Opportunity.allByScope is not a function`, caught at lines 306-309 and returned via `internalServerError(error.message)`, echoing the method name in the response body and `x-error` header.

Engaging with your 2026-05-11T12:47:03Z comment: the link confirms where `allByScope` originates. It does not change that the dep is unpinned to a version that has the method, the companion is unmerged, and CI is green here solely because `allByScope` is stubbed in every test (line 91: `allByScope: sandbox.stub().resolves([])`). No CI check catches a deploy-order mistake.

Fix: (a) block merge until the companion ships and `package.json` is bumped; (b) or add before this line: `if (typeof Opportunity.allByScope !== 'function') { log.error('allByScope unavailable'); return internalServerError('Brand scope query temporarily unavailable'); }` so a misordered deploy fails with a client-safe message instead of a TypeError echo.

src/controllers/llmo/opportunities/llmo-opportunities-controller.js
rarescheseli, line 263: Orphan detection runs on `allScopedOpps` before the LLMO-eligibility and status filter at lines 268-273. Non-LLMO opportunities and RESOLVED/IGNORED status opportunities with foreign `siteId` values inflate `orphanedOpps.length` and fire the warn at line 265 even though those opportunities would be dropped anyway. The warn is the only forensic signal for scope-tagging drift; overstating the count degrades its diagnostic value.

Fix: compute orphans against the LLMO+VALID-status-filtered set - derive `orphanedOpps` from the difference between that filtered set and `filteredOpps` so the count reflects what the response actually excluded due to site-membership violations.

test/controllers/llmo/opportunities/llmo-opportunities-controller.test.js
rarescheseli, line 896: This test catches a regression to `scopeId: opp.getScopeId()` but not the symmetric failure: if `OpportunityDto.toJSON` is reverted to emit `scopeId`, the spread-then-override pattern in the controller (`{ ...OpportunityDto.toJSON(opp), scopeType: 'brand', scopeId: brandId }`) still produces the correct value and this test stays green - while the `all` path at line 178 has no such override and would silently re-expose brand UUIDs on shared sites.

Fix: add `expect(OpportunityDto.toJSON(opp)).to.not.have.property('scopeId')` to test the DTO invariant directly, independent of the controller's spread order. Also add `expect(result.status).to.equal(200)` before `result.json()` so a 500 regression produces a clear assertion failure rather than a confusing TypeError on undefined.

rarescheseli (changes requested):
Hey @HollywoodTonight,

## Strengths

- `src/controllers/llmo/opportunities/llmo-opportunities-controller.js:276-287` - `processBatch + siteMap` closes all three related problems from the prior review in one pass: per-site error isolation, unbounded concurrency, and the poisoned-promise cache. A single `Site.findById` rejection no longer 500s the whole request.
- The cross-brand isolation test at `test/controllers/llmo/opportunities/llmo-opportunities-controller.test.js:498-540` seeds `allBySiteId('site-1')` with both brands' opportunities as a contamination canary and asserts `Opportunity.allBySiteId` was not called. This test now fails under the old site-aggregation code, which is exactly what a regression guard should do.
- `src/dto/opportunity.js:17-22` - `scopeType`/`scopeId` are absent from `OpportunityDto.toJSON`, with a docblock explaining the security rationale. The round-3 brand-UUID disclosure concern is closed, and the all-brand response test at line 929 asserts property absence on the non-scoped path.
- `src/controllers/llmo/opportunities/llmo-opportunities-controller.js:247` - `brandSiteIdSet` collapses the two-sources-of-truth problem into a single `Set<siteId>` derived from PostgREST `brand.siteIds`, applied consistently to the `filterSiteId` 403 gate, the orphan filter, and the row-level membership check.
- Lines 289-301 - `scopeType: 'brand', scopeId: brandId` attached from the validated URL parameter, not from the opportunity row. The IMPORTANT comment at lines 292-295 documents exactly what would re-open cross-brand contamination if changed.
- Lines 148-164 - All-vs-brand divergence documented as non-additive, with the convergence condition named. The "paths are NOT additive" statement makes the consumption hazard explicit.
- Lines 234-238 - PostgREST-unavailable path logs the reason server-side and returns a client-safe fixed string. Correct pattern, applied correctly there.

## Issues

### Critical (Must Fix)

**`Opportunity.allByScope` is not callable on the pinned dependency**

`src/controllers/llmo/opportunities/llmo-opportunities-controller.js:259`

`package.json` at HEAD pins `@adobe/spacecat-shared-data-access@3.60.0`. No published version of this package exports `allByScope` - verified against npm (3.60.0 is the latest). The companion `adobe/spacecat-shared` pull request 1576 is OPEN and not merged. In production, line 259 throws `TypeError: Opportunity.allByScope is not a function`. The exception is caught at lines 306-309 and returned via `internalServerError(error.message)`, echoing the raw method name in both the response body and the `x-error` header on every brand-specific request.

Engaging with your top-level comment (2026-05-11T12:47:03Z): the link to pull request 1576 correctly identifies where `allByScope` originates. It does not change the deployable state of this branch. The companion is unmerged, `package.json` is pinned to a version that does not have the method, and CI's green state on this call site is entirely synthetic - every test stubs `allByScope` directly (test file line 91: `allByScope: sandbox.stub().resolves([])`). There is no CI check that would fail if the companion PR were reverted or the dep not bumped before merge.

Fix (both together are better than either alone):
- (a) Block merge until the companion ships and `package.json` here is bumped to that version. The PR description already commits to this; convert it to a merge precondition.
- (b) Add a capability check immediately before line 259: `if (typeof Opportunity.allByScope !== 'function') { log.error('allByScope unavailable on data-access version'); return internalServerError('Brand scope query temporarily unavailable'); }`. This makes a misordered deploy fail with a client-safe message rather than a TypeError echo.

### Important (Should Fix)

**Outer catch echoes raw `error.message` in the HTTP response**

`src/controllers/llmo/opportunities/llmo-opportunities-controller.js:308`

`internalServerError(error.message)` at line 308 propagates any exception's message verbatim into the response body and `x-error` header: PostgREST driver errors carry constraint names and table identifiers; DynamoDB errors carry condition expression text; the `TypeError: Opportunity.allByScope is not a function` from the Critical above names an internal method. The correct pattern is already in this file at lines 234-238: log the operational detail server-side, return a fixed client string. Apply it to the outer catch: `log.error('getBrandOpportunities failed', { brandId, err: error }); return internalServerError('Failed to fetch brand opportunities');`. The existing test at line 631 asserts `result.status === 500` and that `log.error` was called but never reads `result.json()` to verify the body - add a body assertion to lock the contract.

**Trust-boundary regression test guards only one of two failure modes**

`test/controllers/llmo/opportunities/llmo-opportunities-controller.test.js:896`

The test asserts `body.opportunities[0].scopeId === 'correct-brand'`. This catches a regression that changes line 299 to `scopeId: opp.getScopeId()`. It does NOT catch the symmetric failure: if a future change re-adds `scopeId` to `OpportunityDto.toJSON`, the spread-then-override pattern `{ ...OpportunityDto.toJSON(opp), scopeType: 'brand', scopeId: brandId }` still produces the correct value on the brand path and this test stays green - while the `all` path at line 178, which has no such override, would silently re-expose brand UUIDs on shared sites. Fix: add a direct assertion against `OpportunityDto.toJSON(opp)` itself - `expect(OpportunityDto.toJSON(opp)).to.not.have.property('scopeId')` - so the DTO invariant is tested independently of the controller's spread order. Also add `expect(result.status).to.equal(200)` before `result.json()` to produce a clear failure rather than a confusing TypeError if the endpoint unexpectedly returns 500.

**Orphan detection counts opportunities the response was not going to include**

`src/controllers/llmo/opportunities/llmo-opportunities-controller.js:263`

The orphan filter at line 263 runs on `allScopedOpps`, the raw return of `allByScope`, before the LLMO-eligibility and status filter at lines 268-273. Non-LLMO opportunities and opportunities in RESOLVED or IGNORED status with foreign `siteId` values inflate `orphanedOpps.length` and fire the warn at line 265 even though those opportunities would be dropped by the subsequent filter for unrelated reasons. The warn log is the only forensic signal for scope-tagging drift; overstating the count degrades its diagnostic value. Fix: compute orphans against the LLMO+VALID-status-filtered set. Derive `orphanedOpps` from the difference between the LLMO+status-filtered subset of `allScopedOpps` and `filteredOpps`, so the count reflects what the response actually excluded due to site-membership violations.

### Minor (Nice to Have)

**Orphan log text and structure**

`src/controllers/llmo/opportunities/llmo-opportunities-controller.js:265`

The message "siteIds not in brand.siteIds" references a property that does not exist on the record when `brand.siteIds === undefined` (the exact case the new test at line 641 exercises). Switch to structured logging: `log.warn('llmo.brand_opportunity.orphans', { brandId, orphanCount: orphanedOpps.length, sampleSiteIds: orphanedOpps.slice(0, 5).map((o) => o.getSiteId()) })`. Including a sample of orphan siteIds is the difference between "we have orphans" and "we can investigate orphans" - with only a count, an on-call engineer has nothing to query against.

## Recommendations

- Once the companion ships and `package.json` is bumped, remove the NOTE comment at lines 257-258 in the same commit. Stale notes that say "remove after X ships" accumulate indefinitely unless the dep bump is the trigger for the removal.
- Add a DTO-level unit test asserting `OpportunityDto.toJSON(opp)` has no `scopeType` or `scopeId` keys. This converts the docblock at `src/dto/opportunity.js:17-22` from prose into an executable invariant, closing the half-guard noted in the Important finding above.
- Consider a runtime capability check for `Opportunity.allByScope` at controller construction time. The four-repo deploy chain will reproduce this coordination problem for any future scope type. A fail-fast check makes the ordering constraint observable on startup rather than on the first user request.

## Assessment

Ready to merge? No, with fixes.

Reasoning: The architectural direction is correct and the majority of the prior review's concerns have been addressed thoroughly. What remains: `allByScope` is not callable on the pinned dependency and the outer catch echoes the resulting TypeError back to clients - both have mechanical fixes that belong in this PR before the companion ships. The three Important items are small and should land alongside the dep bump rather than be deferred.

## Next Steps

1. Merge and publish the spacecat-shared companion, bump `@adobe/spacecat-shared-data-access` in `package.json`, and update the lockfile. And/or add the capability check at line 259 so the failure mode is a client-safe message regardless of deploy order.
2. Fix the outer catch at line 308 to return a fixed client string; add a body assertion to the existing 500 test.
3. Add the DTO-level assertion to the trust-boundary test at line 896, and the `result.status` assertion before `result.json()`.
4. Move orphan detection to run on the LLMO+VALID-status-filtered set.

HollywoodTonight (commented):
## Review-kit consolidation (staff / architecture / security lens)

Head reviewed: **`2153b95`** on `feat/llmo-brand-scope-opportunities`.

### Strengths

**Isolation model.** Brand listings use **`Opportunity.allByScope('brand', brandId)`** with **`brand.siteIds`** enforced via **`brandSiteIdSet`** — the right fix for shared-site / multi-brand rows versus iterating **`allBySiteId`** everywhere.

**Trust boundary.** **`scopeType` / `scopeId`** come from the validated route (**`brandId`**), not **`opp.getScopeId()`**, with comments explaining why — closes the “row overrides endpoint scope” failure mode.

**DTO layering.** **`OpportunityDto`** omits scope fields; scope is added only on the brand-scoped response path, consistent with the stated rationale.

**Failure hygiene.** Outer **`catch`** handlers now return generic **`internalServerError(...)`** bodies while **`log.error`** retains **`error.message`** — addresses leaking raw backend strings on 500s for both **count** and **brand opportunities**.

**Dependency.** **`@adobe/spacecat-shared-data-access@3.61.0`** aligns with **`allByScope`** living in shared.

**Tests.** Regression guards (trust-boundary **`scopeId`**, **`all`** path without scope fields, **`undefined` `siteIds`**, **`allByScope`** vs **`allBySiteId`** expectations) remain strong.

### Follow-ups (minor)

1. **Stale NOTE:** Inline NOTE still says **`allByScope` is not yet published** while **`package.json`** pins **`3.61.0`**. Remove or rewrite so readers are not misled.

2. **Orphan warnings:** **`orphanedOpps`** is computed before **`isLlmoOpportunity`** + status filtering — counts may include rows never returned by the API. Either document intentional “data drift” telemetry or align orphan detection with **`filteredOpps`** filters.

3. **Optional test:** Assert 500 responses do **not** echo **`PostgREST unavailable`** / **`DB error`** in the body to lock in non-leaking behavior.

### Assessment

**Technical readiness:** Solid from architecture + API confidentiality standpoint; confirm **`ci/build`** (and related checks) are green before merge.

**Process:** Prior **`CHANGES_REQUESTED`** reviews may still need explicit re-review after the latest pushes.

rarescheseli (approved):

grigoreadobe (approved):

PR #2322, opened 2026-05-04
feat(plg): reject onboarding for internal orgs and paid customers

Kanishkavijay39:
Adds two early-return guards in the PLG onboard handler:
- Returns 400 if the resolved imsOrgId is listed in ASO_PLG_EXCLUDED_ORGS
- Returns 400 if the org already holds a non-PLG ASO entitlement (paid customer)

Local testing
image
image

## Related Issues

Thanks for contributing!


src/controllers/plg/plg-onboarding.js
```diff
@@ -622,6 +622,7 @@ async function performAsoPlgOnboarding({
     Site, PlgOnboarding, Organization,
   } = dataAccess;
 
+  /* c8 ignore next 5 */
```
sandsinh: remove

sandsinh (approved):

src/controllers/plg/plg-onboarding.js
Kanishkavijay39: It's intentionally unreachable , As we are checking this case on start of controller itself, So its unreachable but don't want to remove this check will take care of this in my coming refractor PR.

PR #2324, opened 2026-05-04
feat: auto-create Atomic strategy on async edge-deploy | LLMO-4695

AdrienCappelaere:
## Summary
When a user clicks "Deploy with Experiment" (`Prefer: respond-async`) and a GeoExperiment is successfully created, this PR also appends an Atomic Strategy entry to the site's LLMO strategy blob in S3. The Atomic Strategy reuses the GeoExperiment's UUID (`strategy.id === geoExperimentId`) and links via `experimentId`. Failures of this side-effect do NOT roll back the GeoExperiment — instead, a structured `[atomic-strategy-create-failed]` log is emitted with full context for future reconciliation.

## Why
Part of the Opportunity Workspace GA program (epic [LLMO-2703](https://jira.corp.adobe.com/browse/LLMO-2703)). Implements: [LLMO-4695](https://jira.corp.adobe.com/browse/LLMO-4695) — A1 auto-create Atomic strategy.

Spec: `specs/opportunity-workspace-ga/2026-04-30-ga-design.md` §5
Plan: `specs/opportunity-workspace-ga/02-a1-atomic-create-plan.md`

Depends on F1 ([adobe/spacecat-shared#1571](https://github.com/adobe/spacecat-shared/pull/1571)) — already merged + published as `@adobe/spacecat-shared-utils@1.115.0`.

## Deviations from spec, intentional
Two points where this PR follows the session handoff (newer decisions) rather than the spec doc verbatim:

1. **`strategy.id === geoExperimentId`** (spec said separate UUIDs). Per handoff §3 — single-anchor design.
2. **207 response is unchanged** (spec said add `strategyId`). Per handoff §3 — FE retrieves via `experimentId` lookup, no contract change.

## Self-review summary

**Substantive issue caught and fixed:** the F1 schema's `superRefine` requires `strategyOpportunity.link` to be set when `opportunityId` references a system Opportunity (rather than a library opportunity in the strategy blob's top-level `opportunities` array). Without `link`, the next reader (FE retrieval, our own retry-attempt-2 read, etc.) would fail `parse()` with `Library opportunity <id> does not exist`. Fixed by setting `link: /sites/<siteId>/opportunities/<opportunityId>` and adding a round-trip schema test that locks the contract. (Commit `10a6e82d`.)

**Deferred (flagged for human reviewer):**
- Cross-writer race: `readStrategy` + `writeStrategy` is not atomic — a concurrent writer between our read and write would silently lose its changes. The retry loop reads fresh data each attempt but only protects against same-helper failures. `writeStrategy` v1.115.0 does not yet support optimistic concurrency. See "Reviewer focus".
- `[atomic-strategy-create-failed]` payload includes `lastError.stack` — may bloat logs; existing repo log calls do the same so I left it consistent.

## Verification
```bash
npm test -- --grep 'atomic-strategy-helper'   # 9 helper tests
npm test -- test/controllers/suggestions.test.js   # 397 controller tests (5 new)
npm run lint
npm test  # full suite: 8853 passing, 100% lines/branches/statements
```

## Reviewer focus
- The retry/idempotency loop in `src/support/atomic-strategy-helper.js` — is the read-modify-write race story acceptable for GA? `writeStrategy` does not currently carry version-check; concurrent writers between attempts would silently lose changes. Open to feedback on whether a synchronous SQS dead-letter queue would be preferable to the structured log.
- The insertion point in `suggestions.js` (after `geoExperiment.save()` and after suggestion marking, before the response object). Rationale: GeoExperiment is fully wired up before we attempt the side-effect, so a helper failure leaves a clean GeoExperiment + un-anchored strategy (orphan-safe).
- The intentional non-throw on terminal failure: GeoExperiment is the user's primary asset; we preserve it on partial failure and rely on a future reconciliation script to catch orphans. Open to alternative designs.
- The `strategyOpportunity.link` value: `/sites/<siteId>/opportunities/<opportunityId>` mirrors the API route shape — does this match what the FE expects?
- Test mocking via `esmock` re-mounts the controller per integration test (4x in the new tests). Slower than mutating a shared instance but follows the existing repo pattern. Acceptable?

## Cross-repo PR links
- F1 (merged, dependency): [adobe/spacecat-shared#1571](https://github.com/adobe/spacecat-shared/pull/1571)
- A2+A3 (downstream FE): see [LLMO-2703](https://jira.corp.adobe.com/browse/LLMO-2703)
- A4 (depends on this PR merging): see [LLMO-2703](https://jira.corp.adobe.com/browse/LLMO-2703)


src/controllers/suggestions.js
```diff
@@ -1861,6 +1862,27 @@ function SuggestionsController(ctx, sqs, env) {
...
+            name: geoExperiment.getName?.() || `${opportunity.getType()}-${new Date().toISOString().slice(0, 10)}`,
+            profile,
+            s3: context.s3,
+            log: context.log,
+          });
+        } catch (atomicError) {
+          // Defensive: helper is documented to never throw, but if it does,
+          // we still must not impact the GeoExperiment response.
+          /* c8 ignore next 2 */
+          context.log.error(`[atomic-strategy-create-failed] unexpected throw from helper: ${atomicError.message}`, atomicError);
```
dipratap: If strategy creation fails and we swallow the error but continue with the experiment, how will the strategy be created in that case? Shouldn’t the experiment fail as well?

src/support/atomic-strategy-helper.js
```diff
@@ -0,0 +1,137 @@
...
+        opportunityId,
+        link: opportunityLink,
+        status: 'in_progress',
+        assignee: profile?.email || 'edge-deploy',
+      },
+    ],
+  };
+
+  let lastError;
+  for (let attempt = 1; attempt <= MAX_ATTEMPTS; attempt += 1) {
```
dipratap, line 94: I’m not sure why we'd need retries here since we’re only reading from and writing to S3. If the error is something like auth-related, retries would fail as well. It will also increase the API response time without much benefit.

src/support/atomic-strategy-helper.js
```diff
@@ -0,0 +1,137 @@
...
+    url: '',
+    description: '',
+    topic: opportunityType,
+    createdAt: new Date().toISOString(),
+    createdBy: profile?.email || 'edge-deploy',
+    opportunities: [
+      {
+        opportunityId,
+        link: opportunityLink,
+        status: 'in_progress',
```
dipratap, line 87: How will the strategy be marked as completed?

src/support/atomic-strategy-helper.js
```diff
@@ -0,0 +1,137 @@
...
+  // "Library opportunity <id> does not exist" the next time anything reads
+  // the blob. The link path mirrors the API route shape.
+  const opportunityLink = `/sites/${siteId}/opportunities/${opportunityId}`;
+  const newStrategy = {
+    id: strategyId,
+    type: 'atomic',
+    experimentId: geoExperimentId,
+    name,
+    status: 'in_progress',
+    url: '',
```
dipratap, line 78: Just a fyi - the urls are present in the geo-experiment model under 'urls' field in metadata column, and can be used to show URLs on the oppty workspace strategy section.

src/support/atomic-strategy-helper.js
```diff
@@ -0,0 +1,137 @@
...
+
+      await writeStrategy(siteId, nextData, s3.s3Client, { s3Bucket: s3.s3Bucket });
+
+      log.info(`[atomic-strategy-create] success: strategy ${strategyId} for site ${siteId} (attempt ${attempt})`);
+      return { success: true, strategyId, attempts: attempt };
+    } catch (error) {
+      lastError = error;
+      log.warn(`[atomic-strategy-create] attempt ${attempt}/${MAX_ATTEMPTS} failed for strategy ${strategyId}: ${error.message}`);
+    }
+    /* eslint-enable no-await-in-loop */
```
dipratap, line 127: is this really required here?

AdrienCappelaere:
Thanks for the review @dipratap. Walking through each:

**1. Error swallowing on strategy-create failure ([suggestions.js:1883](https://github.com/adobe/spacecat-api-service/pull/2324/files#diff-d8a7e1e7L1883)):** intentional per the original design — the GeoExperiment is the user's primary asset (the experiment will run and produce results regardless), and the Strategy is a UI projection of that. Failing the experiment because we couldn't write the projection felt worse than the partial state. That said, your concern about an orphaned-GeoExperiment-with-no-Workspace-presence is valid. Filing a follow-up ticket for an orphan-reconciliation mechanism — most likely a lazy strategy-create on next read of the blob, or an SQS retry job. Will link the ticket back here once filed.

**2. Retry loop necessity ([atomic-strategy-helper.js:89](https://github.com/adobe/spacecat-api-service/pull/2324/files#diff-c98f03cL89)):** the retries cover S3 throttling (`SlowDown` / `RequestTimeout` — does happen on busy buckets) and the read-modify-write race where two concurrent writes could otherwise interleave. The idempotency check on retry (`baseData.strategies.some((s) => s.id === strategyId)`) handles the race cleanly. Backoff is `[0, 250, 750]ms` (worst-case ~1s), and auth errors fail-fast on attempt 1 — no latency penalty there. Happy to tighten to retry-only-on-transient-error-codes (`SlowDown`, `5xx`, `RequestTimeout`) and fail-fast on `AccessDenied` etc., if you want — let me know and I'll push that update.

**3. How is the strategy marked completed ([atomic-strategy-helper.js:82](https://github.com/adobe/spacecat-api-service/pull/2324/files#diff-c98f03cL82)):** by design, it isn't — the FE doesn't read \`strategy.status\` for Atomic. It reads live engine state via \`useGeoExperimentDetail\` (the hook A2+A3 added in [adobe/project-elmo-ui#1669](https://github.com/adobe/project-elmo-ui/pull/1669)) and derives all UX state — badge, stepper, View Performance gate — from the live GeoExperiment \`(status, phase)\`. \`strategy.status\` stays \`'in_progress'\` for Atomic strategies. This is documented in our internal design doc; happy to share if useful. Net: no backend transition needed. The asymmetry is intentional but worth surfacing — thanks for flagging.

**4. URLs FYI ([atomic-strategy-helper.js:73](https://github.com/adobe/spacecat-api-service/pull/2324/files#diff-c98f03cL73)):** thanks. The FE already pulls \`metadata.urls\` via \`useGeoExperimentDetail\`, so URLs come for free from the same response the badge / stepper use. Not duplicating onto \`strategy.url\` for now to avoid two sources of truth — let me know if you'd prefer the duplication for any backend-side consumer.

**5. Eslint pragma ([atomic-strategy-helper.js:122](https://github.com/adobe/spacecat-api-service/pull/2324/files#diff-c98f03cL122)):** the pragma is needed because eslint flags any \`await\` inside a \`for\` loop, but the retry pattern legitimately requires sequential awaits (each attempt depends on the previous one's failure + re-reads the blob to pick up concurrent writes via the idempotency check). Just pushed [520edbc6](https://github.com/adobe/spacecat-api-service/pull/2324/commits/520edbc6) adding a comment that explains this.

Standing by on (1) and (2) — let me know if you want me to file the follow-up ticket for orphan reconciliation now and/or tighten the retry conditions in this PR.

dipratap:
1. If the strategy is not created, there will be no UI view to display the experiment data. Also, with the new UI (progress icons and the “View Strategy” button), where would the user navigate when they click it? I believe strategy creation is necessary for the experiment UI flow and should succeed for the overall experiment deployment to be considered successful. As a precautionary measure, please test the flow e2e on dev.

5. I meant the comment at the end of the loop. I see there are three awaits, and instead of adding ignore comments to each one, the AI applied it to the whole block which is fine, the comment to explain the ignore can be skipped.

AdrienCappelaere:
Thanks @dipratap — you're right. The "GeoExperiment is the primary asset, strategy is a UI projection" framing was technically true but practically wrong: the new Workspace UI (progress stepper, View Strategy button) has no path to surface an experiment without its strategy, so a missing strategy = an invisible experiment. The orphan-reconciliation follow-up I floated earlier was a fig leaf for a hazard the code should just not create.

Pushed [dabd7ba0](https://github.com/adobe/spacecat-api-service/pull/2324/commits/dabd7ba0) — two changes that together make the flow transactional:

**1. Reordered.** `createAtomicStrategy` now runs immediately after `GeoExperiment.create()` and **before** any irreversible side-effect (DRS schedule activation via `createExperimentSchedule(..., triggerImmediately: true)`, suggestion marking). The helper now throws on terminal failure instead of swallowing. If the strategy can't be written, the existing outer catch rolls back the GeoExperiment row — no engine activity ever started, no suggestions mutated, no orphans.

**2. Added `deleteAtomicStrategy` for the symmetric case.** If the strategy succeeded but a downstream step (DRS, save `preScheduleId`, etc.) fails, the outer catch now also removes the strategy entry from the LLMO blob. Same retry/idempotency semantics as create. So we can't end up with an orphan strategy pointing at a removed experiment either.

Net contract: **strategy exists iff the whole deploy succeeded.** No invisible experiments, no dangling strategies.

A few notes on your other points:

- Retry tightening (#2) — kept the 3-attempt loop since auth errors do already fail-fast on attempt 1 (no `SlowDown`/`5xx` to retry against). Happy to narrow further if you'd still prefer.
- Status-not-completed (#3) — by design for Atomic. The FE derives all UX state (badge, stepper, View Performance gating) from live `useGeoExperimentDetail(status, phase)` rather than `strategy.status`, so we don't sync engine state back into the strategy blob. Documented in our internal design doc; happy to share.
- URLs FYI (#4) — thanks, the FE already pulls `metadata.urls` via the same hook.
- Pragma comment (#5) — dropped in [0adc61c6](https://github.com/adobe/spacecat-api-service/pull/2324/commits/0adc61c6).

Tested locally with synthetic failures injected at both points (broken S3 → strategy fails → experiment rolled back; broken DRS → experiment rolled back AND strategy deleted). 8864 unit/integration tests pass, 100% coverage on changed files. Post-merge will watch `[atomic-strategy-create-failed]`, `[atomic-strategy-delete-failed]`, `[atomic-strategy-cleanup-failed]` in Coralogix.

src/controllers/suggestions.js
```diff
@@ -1804,6 +1808,25 @@ function SuggestionsController(ctx, sqs, env) {
 
         context.log.info(`[edge-geo-exp] Created GeoExperiment ${geoExperimentId} with status GENERATING_BASELINE / phase PRE_ANALYSIS_STARTED`);
 
+        // Append an Atomic Strategy entry to the site's LLMO strategy blob
+        // BEFORE any irreversible engine activation (DRS schedule) or
+        // suggestion mutation. The new UI treats the Atomic strategy as a
+        // required UI projection of the experiment — without it, "View
+        // Strategy" / progress stepper have nowhere to bind. Running the
+        // strategy write here makes failure cheap: only the GeoExperiment
+        // row exists, and the outer catch already rolls it back.
```
dipratap: Please reduce the comment length to 1-2 lines

src/support/atomic-strategy-helper.js
```diff
@@ -0,0 +1,239 @@
...
+        log.info(`[atomic-strategy-create] idempotent skip: strategy ${strategyId} already present (attempt ${attempt})`);
+        return { success: true, strategyId, attempts: attempt };
+      }
+
+      const nextData = {
+        ...baseData,
+        strategies: [...baseData.strategies, newStrategy],
+      };
+
+      await writeStrategy(siteId, nextData, s3.s3Client, { s3Bucket: s3.s3Bucket });
```
dipratap, line 119: I hope cases like a strategy file not existing and other corner cases are handled in read and write strategy methods, so that they don't throw and fail the experiment.

src/controllers/suggestions.js
```diff
@@ -1893,6 +1916,22 @@ function SuggestionsController(ctx, sqs, env) {
             context.log.error(`[edge-geo-exp-failed] Failed to clean up GeoExperiment ${geoExperimentId}: ${removeError.message}`, removeError);
           }
         }
+        // Compensating action: if the Atomic strategy was already written
+        // before this failure, delete it so we don't leave an orphan strategy
+        // pointing at a now-removed GeoExperiment (the symmetric failure mode
+        // to an orphan experiment with no UI projection).
```
dipratap: please reduce the comment length:
"Delete the strategy if it is created so we don't leave an orphan strategy"

dipratap:
@AdrienCappelaere Can you please test e2e flow also on any dev sites cc @jjenscodee

src/controllers/suggestions.js
AdrienCappelaere: Done in 8e4ff3e4 — trimmed to two lines: _"Create the Atomic strategy before DRS / suggestion-marking so a failure rolls back cheaply via the outer catch."_

src/controllers/suggestions.js
AdrienCappelaere: Done in 8e4ff3e4 — used your suggested phrasing: _"Delete the strategy if it was created so we don't leave an orphan."_

src/support/atomic-strategy-helper.js
AdrienCappelaere: Confirmed for "file doesn't exist", honest about the rest:

- **Strategy file not existing** → handled. `readStrategy` catches `NoSuchKey` / `NotFound` and returns `{ data: null, exists: false }` ([source](https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-utils/src/llmo-strategy.js#L58-L64)). Our helper then initialises a fresh blob with the new strategy as the only entry, so a never-seen site is a normal first-write, not a failure.

- **Other S3 errors** (auth, throttling, network, `NoSuchBucket`) → propagate. Same for `writeStrategy` (any S3 error throws; missing `VersionId` throws "Failed to get version ID after writing strategy" if the bucket isn't versioned).

- **Schema-parse on an existing-but-malformed blob** (`strategyWorkspaceData.parse(JSON.parse(text))`) → throws. Rare but possible if the blob was hand-edited or written by an older incompatible schema.

Propagation is intentional under the new contract: if S3 is unreachable or the existing blob is corrupt, we'd rather fail the deploy than silently leave the user with an experiment they can't see. The 3-attempt retry loop covers transient cases (`SlowDown`, `RequestTimeout`, 5xx); auth and `NoSuchBucket` correctly fail-fast on attempt 1. So the cases that ARE handled gracefully are the ones that should be (transient + new-site); the cases that throw are the ones that should throw (genuine misconfiguration or data corruption).

AdrienCappelaere:
### ✅ DEV E2E validated

Got a clean end-to-end success on DEV after the merge from main + extended logging. Posting the receipts for the record.

**Happy path** (traceId `1-6a0ef9fc-32b9743503473ec27e516881`) — full request completed in **3.4s**:

| Step | Duration |
|---|---|
| `prompts-s3-upload` | 122 ms |
| `geoexperiment-create` | 51 ms |
| `atomic-strategy-create` | **66 ms** (29 ms S3 GET + 37 ms S3 PUT) |
| `drs-pre-schedule` | 3081 ms |
| `geoexperiment-save-preid` | 41 ms |
| `mark-suggestions` | 68 ms |

**Cleanup path also validated** (previous failed run, traceId `1-6a0eb7ab-47c0ab733de695761618ef8a`) — when the DRS call timed out at 10s, the rollback ran cleanly:

- `geoExperimentRemoved=true`
- `strategyDeleted=true` (S3 read 176 ms + write 40 ms = 216 ms)
- `unblockedSuggestions=0/0` (none had been marked yet — DRS fails before that step)
- Total `rollbackMs=310`

So we have proof on both branches: strategy is written when the experiment succeeds, and removed when it doesn't. No orphans left in S3.

**S3 strategy creation perf** — consistent across runs at **~65 ms** (66 ms now, 64 ms in the earlier failed run). This is negligible vs. the rest of the flow; the helper's worst-case retry budget (3 × ~70 ms work + 1 s backoff = ~1.2 s) is still 10× under the DRS client's 10 s ceiling.

**Note on DRS:** the earlier timeout was an upstream issue (\`POST /schedules\` exceeded the 10 s default of \`tracingFetch\`). DRS variability is not addressed by this PR — tracked separately. This PR only owns the atomic-strategy projection and its rollback.

dipratap (approved):

PR #2326, opened 2026-05-05
Feat/llm bot probe

shrotia:

## Related Issues

Thanks for contributing!


varu-adobe:
@shrotia for the 2 new checks added

- Spacecat UA - this is already covered via already available check on the UI. Even if Spacecat UA is blocked, Tokowaka should work, so there is no need to tie the 2 things togther unless I am missing something.
- Real LLM bots(ChatGPT-User...) - this could still be blocked from our infra without the fetcher key, if the intent is to show to customers that their agentic traffic is blocked, it needs to go into one of the opportunities, could check with Yaashi here.

shrotia:
> @shrotia for the 2 new checks added
> 
> * Spacecat UA - this is already covered via already available check on the UI. Even if Spacecat UA is blocked, Tokowaka should work, so there is no need to tie the 2 things togther unless I am missing something.
> * Real LLM bots(ChatGPT-User...) - this could still be blocked from our infra without the fetcher key, if the intent is to show to customers that their agentic traffic is blocked, it needs to go into one of the opportunities, could check with Yaashi here.

let me explain the intent for both:

SpaceCat UA baseline: SpaceCat scraping is a prerequisite for O@E. If this probe is already covered then let me remove - we can rely on the UI status. 

LLM bot agents (ChatGPT-User, Perplexity-User):

Agreed the probe fires from SpaceCat Lambda IPs, not from ChatGPT's real IPs — so it won't catch IP-based blocks. With one of the recent domain's soft block - scraping was working, AdobeEdgeOptimize/1.0 was whitelisted but receiving captcha response due to broken routing. So calling from Spacecat would allow bypassing the WAF/Firewall and chatgpt-user would confirm if the routing is really working. Let me check with her also.

PR #2327, opened 2026-05-05
feat: rejected status

anshulk-public:

## Related Issues

Thanks for contributing!


src/controllers/plg/plg-onboarding.js
```diff
@@ -672,8 +684,35 @@ async function performAsoPlgOnboarding({
...
+
+  // Mark any WAITLISTED or WAITING_FOR_IP_ALLOWLISTING records for other domains in this org
+  // as OUTDATED — a new onboarding attempt supersedes these pending/blocked domains.
+  const waitlistedRecords = existingRecords.filter(
+    (r) => r.getDomain() !== domain && [
+      STATUSES.WAITLISTED,
+      STATUSES.WAITING_FOR_IP_ALLOWLISTING,
+    ].includes(r.getStatus()),
+  );
+  await Promise.all(waitlistedRecords.map(async (r) => {
```
Kanishkavijay39: **Critical (items 3 + 4): no per-record error isolation, and runs before the `alreadyOnboarded` guard**

`Promise.all` over `r.save() + postPlgOnboardingNotification` rejects on the first failure and abandons remaining records. A transient Slack 503 leaves the org in partially mutated state (some records OUTDATED, others still WAITLISTED) with no compensating action.

Separately, this entire sweep runs before the `alreadyOnboarded` check below. If the new domain itself ends up WAITLISTED (due to a competing ONBOARDED domain, AEM check failure, bot-blocker, etc.), all peer records are already permanently OUTDATED with a false "superseded by X" review.

Fix: (1) move this loop to execute only after the new onboarding reaches ONBOARDED, (2) use `Promise.allSettled` and wrap each Slack notification in its own try/catch.

src/controllers/plg/plg-onboarding.js
```diff
@@ -1637,9 +1651,12 @@ function PlgOnboardingController(ctx) {
...
   /**
    * PATCH /plg/onboard/:onboardingId
-   * Admin-only: review a waitlisted onboarding (BYPASS or UPHOLD), or record a review on an
-   * ONBOARDED record and transition it to WAITLISTED (revokes ASO site enrollments when linked).
-   * On BYPASS for WAITLISTED, performs scenario-specific prep and re-runs the PLG flow.
+   * Admin-only: review an onboarding record with one of four decisions:
+   * - BYPASSED: re-run the flow (WAITLISTED only)
+   * - UPHELD: transition to REJECTED (WAITLISTED or ONBOARDED)
+   * - CLOSED: transition to OUTDATED (WAITLISTED or ONBOARDED)
+   * - REOPENED: transition to OUTDATED (REJECTED only — lifts a prior rejection)
```
Kanishkavijay39: **Critical (item 2): JSDoc describes the wrong function**

This comment lists four valid decisions (BYPASSED, UPHELD, CLOSED, REOPENED) and claims UPHELD accepts WAITLISTED or ONBOARDED records. Neither is true for `update`:
- The implementation rejects anything that is not WAITLISTED (line 1696)
- Only BYPASSED and UPHELD are handled
- CLOSED, REOPENED, OFFBOARDED are handled by `transitionStatus`, not here

Rewrite to reflect actual behavior and cross-reference `transitionStatus` for the other transitions.

src/controllers/plg/plg-onboarding.js
```diff
@@ -1702,33 +1719,19 @@ function PlgOnboardingController(ctx) {
...
     const checkKey = deriveCheckKey(onboarding);
     if (!checkKey) {
       return badRequest('Unable to determine the review reason from the onboarding record');
     }
 
-    // UPHOLD: just store the review and return
+    onboarding.setUpdatedBy(reviewedBy);
+
+    // UPHOLD: reject the domain — transition to REJECTED (terminal)
     if (decision === REVIEW_DECISIONS.UPHELD) {
```
Kanishkavijay39, line 1763: **Critical (item 1): unknown `decision` values fall silently into BYPASS**

This branch handles UPHELD and returns early. Everything else - including the newly introduced `CLOSED`, `REOPENED`, and `OFFBOARDED` values now in `REVIEW_DECISIONS` - falls through to the BYPASS re-run path. That path re-creates projects, enables audits, and persists a review entry with an undecodable `decision` string. No test covers this. Recovery requires manual DB cleanup.

Fix - add before the UPHELD check:
```js
const allowedDecisions = [REVIEW_DECISIONS.BYPASSED, REVIEW_DECISIONS.UPHELD];
if (!allowedDecisions.includes(decision)) {
  return badRequest(`decision must be one of: ${allowedDecisions.join(', ')}`);
}
```
Add a unit test sending `decision: 'CLOSED'` asserting 400.

src/controllers/plg/plg-onboarding.js
```diff
@@ -2064,6 +2069,83 @@ function PlgOnboardingController(ctx) {
...
+
+    const currentStatus = onboarding.getStatus();
+    const allowedSourceStatuses = [STATUSES.WAITLISTED, STATUSES.ONBOARDED, STATUSES.REJECTED];
+    if (!allowedSourceStatuses.includes(currentStatus)) {
+      return badRequest(`Only WAITLISTED, ONBOARDED, or REJECTED records can be transitioned to ${targetStatus}, current status: ${currentStatus}`);
+    }
+
+    if (currentStatus === STATUSES.ONBOARDED) {
+      try {
+        await revokeAsoSiteEnrollments(onboarding, context);
```
Kanishkavijay39: **Important (item 9): revoke-then-save leaves a partial failure window**

`revokeAsoSiteEnrollments` runs before `onboarding.save()`. If the revoke succeeds but the save fails, the PLG record stays ONBOARDED in the DB while ASO entitlements are already revoked. The customer-facing site appears healthy in PLG dashboards but functionality is broken - invisible to monitoring. An operator retry re-revokes an already-revoked record.

Fix: flip the order - save the status change first, then revoke. A stale active enrollment is recoverable; a stale ONBOARDED record with no entitlements is not.

src/controllers/plg/plg-onboarding.js
```diff
@@ -2064,6 +2069,83 @@ function PlgOnboardingController(ctx) {
...
+        const msg = err instanceof Error ? err.message : String(err);
+        log.error(`Failed to revoke ASO enrollments for onboarding ${onboardingId}: ${msg}`, err);
+        return internalServerError('Failed to revoke ASO enrollments. Please try again later.');
+      }
+    }
+
+    // System-generated review: decision is derived from the source status
+    let decision;
+    if (currentStatus === STATUSES.REJECTED) {
+      decision = REVIEW_DECISIONS.REOPENED;
```
Kanishkavijay39, line 2147: **Important (item 8): `REOPENED` is semantically wrong for REJECTED to OUTDATED**

"Reopened" conventionally means returning to an active, actionable state. This transition moves the record into OUTDATED - a terminal archive state. A reviewer scanning the audit log sees `decision: REOPENED, status: OUTDATED` and cannot determine what happened next. Analytics queries on `decision === 'REOPENED'` will surface archived records.

Align the decision name with the actual semantics (`RETRACTED`, `RESCINDED`, or `WITHDRAWN_REJECTION`) - or revisit whether REJECTED should transition to WAITLISTED here, which would make REOPENED accurate.

test/it/shared/tests/plg-onboarding.js
```diff
@@ -300,5 +294,99 @@ export default function plgOnboardingTests(getHttpClient, resetData, options = {
...
+          expect(res.body.status).to.equal('OUTDATED');
+          expect(res.body.reviews).to.be.an('array').with.lengthOf(1);
+          expect(res.body.reviews[0].decision).to.equal('OFFBOARDED');
+          expect(res.body.reviews[0].justification).to.equal('Offboarding domain at customer request');
+          expect(res.body.reviews[0].reason).to.include('site1.example.com');
+          expect(res.body.reviews[0].reviewedBy).to.be.a('string');
+          expectISOTimestamp(res.body.reviews[0].reviewedAt, 'reviewedAt');
+        });
+
+        // PLG_ONBOARDING_2 was transitioned to REJECTED by the UPHELD test above
```
Kanishkavijay39: **Critical (item 5): implicit state dependency on the preceding UPHELD test**

This test only passes because the earlier `'UPHELD: stores review and transitions WAITLISTED to REJECTED'` test mutated PLG_2 from WAITLISTED to REJECTED. Running with `--grep 'REJECTED: transitions to OUTDATED'` in isolation leaves PLG_2 as WAITLISTED - the controller takes the WAITLISTED to OUTDATED path and the `decision === 'REOPENED'` assertion fails with a misleading error.

CLAUDE.md documents that each describe block should call `before(() => resetData())` to ensure isolation - this violates that contract.

Fix: add a dedicated REJECTED seed (PLG_5) pre-populated in `plg-onboardings.js` so this test has no predecessor dependency.

Kanishkavijay39 (changes requested):
Hey @anshulk-public,

## Strengths

- The REJECTED/OUTDATED status distinction cleanly replaces the overloaded use of WAITLISTED for displaced and rejected records. Downstream consumers can now distinguish "pending review," "explicitly rejected," and "superseded" without parsing free-text `waitlistReason` strings.
- Centralizing `callerIdentity = getReviewerIdentity(context)` removes ~10 scattered `if (updatedBy) onboarding.setUpdatedBy(...)` branches and correctly attributes system-driven transitions to `'system'`.
- The uniform reviews-as-event-log pattern - every transition appending a structured entry with `decision`, `justification`, `reviewedBy`, `reviewedAt` - is a genuine improvement over smearing audit data into `waitlistReason` strings.
- `transitionStatus` validates both target and source statuses and derives the decision server-side rather than accepting it from the caller. This forecloses a whole class of invalid state transitions.
- Unit test coverage on `transitionStatus` is thorough: 403/400/404 validation paths, each source-status mapped to the correct decision, null `data`, non-Error rejection, and append-not-replace review semantics.

## Issues

### Critical (Must Fix)

**1. `update` accepts unknown `decision` values and silently triggers BYPASS**

See inline comment at the UPHELD branch. The handler branches on `UPHELD` and returns early - everything else falls through to the BYPASS re-run path, including the new `CLOSED`, `REOPENED`, and `OFFBOARDED` values now present in `REVIEW_DECISIONS`. An admin sending `{ decision: "CLOSED" }` to the wrong endpoint silently triggers a full ASO onboarding re-run and persists an undecodable review entry. No test covers this path. Recovery requires manual DB cleanup.

Fix: add a decision allowlist immediately after the `checkKey` guard and add a unit test asserting 400 for unknown decisions.

**2. `update` JSDoc describes the wrong function**

See inline comment at the JSDoc block. The updated comment lists four valid decisions including CLOSED and REOPENED, and claims UPHELD accepts ONBOARDED records. The actual implementation accepts only WAITLISTED and handles only BYPASSED and UPHELD. Combined with item 1, this docstring actively guides a maintainer toward the silent fall-through.

**3. Auto-OUTDATED loop runs before the `alreadyOnboarded` guard**

See inline comment at the `Promise.all` block. Every WAITLISTED and WAITING_FOR_IP_ALLOWLISTING peer is marked OUTDATED unconditionally before checking whether the new domain is itself blocked. If any guard triggers, the new domain ends up WAITLISTED while every previously-WAITLISTED record in the org has been permanently OUTDATED with a false "superseded by X" review.

Fix: move the auto-OUTDATED loop to execute only after the new onboarding has durably reached ONBOARDED.

**4. `Promise.all` in the auto-OUTDATED batch has no per-record error isolation**

See inline comment at the `Promise.all` block. A transient Slack 503 rejects the entire batch and leaves the org in partially mutated state with no compensating action.

Fix: use `Promise.allSettled` and wrap each Slack notification in its own try/catch. This pattern is already used elsewhere in the file.

**5. IT tests have implicit state dependency via shared `resetData()`**

See inline comment in `test/it/shared/tests/plg-onboarding.js`. `resetData()` runs once for the entire outer describe block. The REJECTED to OUTDATED test on PLG_2 only passes because the preceding UPHELD test mutated PLG_2 from WAITLISTED to REJECTED. Running with `--grep` in isolation produces misleading failures. CLAUDE.md documents each describe block calls `before(() => resetData())` to ensure isolation - this PR violates that contract.

Fix: add a dedicated REJECTED seed (PLG_5) so this test has no predecessor dependency.

### Important (Should Fix)

**6. New endpoint and changed `update` contract have no OpenAPI spec updates**

The diff does not touch `docs/openapi/`. The new `PATCH /plg/onboard/:onboardingId/status` endpoint, the changed `update` semantics (WAITLISTED-only, now transitions to REJECTED), and the new enum values (REJECTED, OUTDATED, CLOSED, REOPENED, OFFBOARDED) all need spec updates. Per CLAUDE.md "Must-Follow Patterns": OpenAPI First and Specification Sync are team requirements. Clients generated from these specs will have stale enums and a missing endpoint definition.

**7. Breaking change is not communicated**

`update` previously accepted ONBOARDED records and returned 200. It now returns 400. The commit is `feat:`, not `feat!:`, and there is no `BREAKING CHANGE:` footer. Any internal tooling, Slack action, or backoffice automation currently calling `PATCH /plg/onboard/:id` against an ONBOARDED record breaks on deploy. Before merging: inventory callers via codebase grep and Coralogix (prior 200s on this path with ONBOARDED status), coordinate migration to `transitionStatus`, and add a `BREAKING CHANGE:` footer to the commit.

**8. `REOPENED` decision for REJECTED to OUTDATED is semantically wrong**

See inline comment in `transitionStatus`. "Reopened" conventionally means returning to an active state. Here it tags a transition into OUTDATED, a terminal archive state. Analytics queries keyed on `decision === 'REOPENED'` will surface archived records.

**9. Revoke-then-save partial failure window in `transitionStatus`**

See inline comment at the `revokeAsoSiteEnrollments` call. For ONBOARDED records, revocation runs before `onboarding.save()`. If the revoke succeeds but the save fails, the PLG record remains ONBOARDED in the DB while ASO entitlements are already revoked - invisible to monitoring.

Fix: flip the order - save first, then revoke. A stale active enrollment is recoverable; a stale ONBOARDED record with no entitlements is not.

**10. Missing save-failure and notification-failure unit tests for `transitionStatus`**

The `update` path previously had explicit tests for `save()` rejection (500) and non-Error save failure. These were removed when the ONBOARDED path moved to `transitionStatus`, but no equivalent tests were added. Currently if `onboarding.save()` rejects inside `transitionStatus`, the error propagates unhandled. If Slack notification throws, the API returns 500 even though the state transition succeeded.

Add: (a) a test with `onboarding.save.rejects(new Error(...))` asserting 500, (b) a test with Slack failing asserting the response is still 200.

**11. Five `this.timeout(10000)` additions signal possible flakiness**

Five tests that previously ran within the default timeout now require 10s. Run with `--reporter=min --slow 100` to locate hot spots. If esmock setup is the cause, hoist module setup to describe-level. Do not paper over timing issues with generous timeouts.

**12. Data service image jump v5.1.1 to v5.7.2 needs confirmation**

Six minor versions bundled with a feature change. The new REJECTED/OUTDATED status values almost certainly require a data service version that includes those enum values - confirm in the PR description that the production data service is already at or above v5.7.2. If prod is below this version, this PR breaks on deploy.

### Minor (Nice to Have)

**13. Admin-supplied `justification` is interpolated verbatim into Slack messages**

Slack mrkdwn allows `<!channel>`, `<!here>`, and link spoofing via `<url|text>` syntax. Admin-internal only, so blast radius is limited - but it is a one-line sanitization fix at the interpolation site in `postPlgOnboardingNotification`.

**14. `justification` length is unbounded**

No max-length validation exists on either `update` or `transitionStatus`. A large value persists to the JSONB reviews column and bloats list-endpoint payloads over time. Add a reasonable cap (e.g., 2000 chars) with a `badRequest` on overflow.

**15. IT invalid-status test uses `'REJECTED'` as the example invalid value**

If REJECTED is ever added as a valid target, this test breaks confusingly. Use a clearly nonsensical value like `'BANANA'` instead.

**16. `REVIEW_DECISIONS`/`STATUSES` enum literals are duplicated ~6 times in the unit test file**

Adding a new value requires six synchronized edits. Extract shared consts at the top of the test file.

**17. OUTDATED notification unit test asserts unreachable production state**

The `'OUTDATED notification omits review section when no reviews'` test wires `record.getReviews.returns(null)` as a sticky stub, but `transitionStatus` always appends a synthetic review entry before calling `postPlgOnboardingNotification`. Switch to a `callsFake` that mirrors `setReviews` calls, or delete the test.

## Recommendations

- Treat the state machine as a first-class artifact. With six statuses and five decisions, a static transition table in a comment block would make future changes auditable and catch asymmetries at a glance.
- Log successful transitions with structured fields (`onboardingId`, `from`, `to`, `actor`, `decision`). Currently only failures are logged; the audit trail lives only in the DB reviews array.
- Batch or rate-limit auto-OUTDATED Slack notifications. An org with many stale WAITLISTED records generates one Slack message per record on a single new onboarding attempt.

## Assessment

Items 1-5 must be addressed before merge. Items 1 and 2 are tightly coupled and quick to fix. Item 3 (auto-OUTDATED ordering) is the most significant - failed onboarding attempts silently corrupt peer audit trails. Item 4 (Promise.all isolation) makes routine Slack outages customer-impacting. Item 5 (IT test isolation) violates documented team standards and creates maintenance traps.

## Next Steps

1. Fix the 5 critical items: decision validation in `update`, the JSDoc rewrite, auto-OUTDATED ordering, `Promise.allSettled` isolation, and dedicated IT seeds for REJECTED/ONBOARDED states.
2. Add OpenAPI spec updates and add `BREAKING CHANGE:` footer to the commit.
3. Resolve the `REOPENED` naming (may need product alignment) and flip the revoke-then-save order.
4. Add save/notification failure unit tests for `transitionStatus`.
5. Investigate the 10s timeouts and confirm the prod data service version against v5.7.2.
6. Minor items are optional but items 13 and 14 are each one-liners worth landing now.

src/controllers/plg/plg-onboarding.js
anshulk-public: done

src/controllers/plg/plg-onboarding.js
anshulk-public: This is already approved in design

src/controllers/plg/plg-onboarding.js
anshulk-public: done

src/controllers/plg/plg-onboarding.js
anshulk-public: done

test/it/shared/tests/plg-onboarding.js
anshulk-public: done

src/controllers/plg/plg-onboarding.js
anshulk-public: done

Kanishkavijay39 (approved):

PR #2331, opened 2026-05-05
fix(scrape): sanitize x-error header to prevent ERR_INVALID_CHAR on non-ASCII

solaris007:
## Summary

Applies `cleanupHeaderValue()` from `@adobe/helix-shared-utils` to the `x-error` header in `scrapeJob.js`'s `createErrorResponse()` function, matching the pattern already used in `slack.js` and `llmo.js`.

## Root Cause

Mystique's SEO suggestion generator passes LLM-hallucinated URLs containing non-ASCII characters (Cyrillic, Georgian script) to the scrape API. When these fail validation, the error message is placed in the `x-error` HTTP header unsanitized, causing Node.js `ERR_INVALID_CHAR` crashes. This controller was the only one missing the sanitization.

## Changes

- Import `cleanupHeaderValue` from `@adobe/helix-shared-utils`
- Wrap `error.message` with `cleanupHeaderValue()` in `createErrorResponse()`

## Jira

[SITES-43989](https://jira.corp.adobe.com/browse/SITES-43989)

### DIFF
--- a/src/controllers/scrapeJob.js
+++ b/src/controllers/scrapeJob.js
@@ -48,7 +48,7 @@ function ScrapeJobController(context) {
 
   function createErrorResponse(error) {
     return createResponse({}, error.status || 500, {
-      [HEADER_ERROR]: cleanupHeaderValue(error.message),
+      [HEADER_ERROR]: cleanupHeaderValue(error.message ?? ''),
     });
   }
 
--- a/src/controllers/scrapeJob.js
+++ b/src/controllers/scrapeJob.js
@@ -22,6 +22,7 @@ import {
   isValidUrl,
   hasText,
 } from '@adobe/spacecat-shared-utils';
+import { cleanupHeaderValue } from '@adobe/helix-shared-utils';
 
 /**
  * Scrape controller. Provides methods to create, read, and fetch the result of scrape jobs.
@@ -47,7 +48,7 @@ function ScrapeJobController(context) {
 
   function createErrorResponse(error) {
     return createResponse({}, error.status || 500, {
-      [HEADER_ERROR]: error.message,
+      [HEADER_ERROR]: cleanupHeaderValue(error.message),
     });
   }
 
--- a/src/controllers/scrapeJob.js
+++ b/src/controllers/scrapeJob.js
@@ -48,7 +48,7 @@ function ScrapeJobController(context) {
 
   function createErrorResponse(error) {
     return createResponse({}, error.status || 500, {
-      [HEADER_ERROR]: cleanupHeaderValue(error.message ?? ''),
+      [HEADER_ERROR]: cleanupHeaderValue(error.message),
     });
   }
 
--- a/test/controllers/scrape-job.test.js
+++ b/test/controllers/scrape-job.test.js
@@ -265,6 +265,17 @@ describe('ScrapeJobController tests', () => {
       expect(response.headers.get('x-error')).to.equal('Failed to create a new scrape job: Queue error');
     });
 
+    it('should sanitize non-ASCII and control chars in x-error header', async () => {
+      baseContext.sqs.sendMessage = sandbox.stub().throws(new Error('Failed: https://пример.com/page\r\nX-Injected: yes'));
+      const response = await scrapeJobController.createScrapeJob(baseContext);
+      expect(response.status).to.equal(500);
+      const xerr = response.headers.get('x-error');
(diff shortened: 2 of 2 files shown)


danieljchuser (changes requested):
Hey @solaris007,

## Strengths
- Clean, surgical fix at the right chokepoint. All error paths in `scrapeJob.js` funnel through `createErrorResponse`, so sanitizing once here covers every error response from this controller (`src/controllers/scrapeJob.js:48-53`).
- Reuses `cleanupHeaderValue` from `@adobe/helix-shared-utils` (already in the dependency manifest), consistent with the pattern established in `slack.js` and `llmo.js`. No new dependency, no bespoke regex.
- Verified that `cleanupHeaderValue` strips not just non-ASCII but also CRLF and other control characters (regex `/[^\t -~-ÿ]/g`), and truncates to 1024 chars - so this fix closes both the crash bug and a latent header-injection surface that existed pre-fix.
- Diff size is proportional to the bug: 2 lines, one file, one import.

## Issues

### Important (Should Fix)

**Missing regression test for the sanitization behavior** (`test/controllers/scrape-job.test.js`)

Concurring with Copilot's existing comment: the non-ASCII error path is the specific behavior this patch introduces, and existing tests exercise the line but only with ASCII-only error messages. If `cleanupHeaderValue` is ever removed or mis-imported, no test will catch the regression.

One `it` block covering both non-ASCII and CRLF (the two distinct character classes the helper strips) is sufficient:

```js
it('createErrorResponse sanitizes non-ASCII and control chars in x-error', async () => {
  sandbox.stub(scrapeClient, 'createScrapeJob')
    .rejects(new Error('Failed: https://пример.com/page\r\nX-Injected: yes'));
  const response = await scrapeJobController.createScrapeJob(reqContext);
  expect(response.status).to.equal(500);
  const xerr = response.headers.get('x-error');
  expect(xerr).to.be.a('string');
  expect(/[^\x20-\x7E]/.test(xerr)).to.equal(false);
  expect(xerr).to.not.include('\r');
  expect(xerr).to.not.include('\n');
});
```

### Minor (Nice to Have)

**`error.message` may be undefined** (`src/controllers/scrapeJob.js:51`) - Non-Error thrown values (e.g. `throw { status: 502 }`) flow into `createErrorResponse`, calling `cleanupHeaderValue(undefined)`. The helper likely tolerates falsy input, but a `?? ''` guard would document the expectation.

## Recommendations
- Track a follow-up to centralize header sanitization. At least 10 places in this repo write `x-error` headers (`api-key.js`, `consumers.js`, `consentBanner.js`, `import.js`, `brands.js`, `sandbox-audit.js`, `multipart-form-data.js`, `support/sandbox-audit-service.js`, `event/fulfillment.js`, `slack.js`, `llmo.js`), but only 3 sanitize today. Pushing sanitization into the response factory or a shared helper would prevent the next controller author from hitting this same crash.

## Assessment
Ready to merge? With fixes
Reasoning: The fix itself is correct, minimal, and security-positive. The only ask is a single regression test for the specific failure mode being patched - cheap to add and important for preventing future regressions.

## Next Steps
1. Add one regression test covering non-ASCII + CRLF sanitization in the error header path.
2. Optional: guard against `cleanupHeaderValue(undefined)` with a `?? ''` fallback.
3. Consider opening a follow-up issue for the centralized sanitization work across all controllers.

src/controllers/scrapeJob.js
```diff
@@ -47,7 +48,7 @@ function ScrapeJobController(context) {
 
   function createErrorResponse(error) {
     return createResponse({}, error.status || 500, {
-      [HEADER_ERROR]: error.message,
+      [HEADER_ERROR]: cleanupHeaderValue(error.message),
```
danieljchuser, line 51: The sanitization behavior introduced here warrants a regression test. Existing `x-error` assertions in `scrape-job.test.js` all use ASCII-only fixtures, so they exercise the line but do not verify the non-ASCII/CRLF stripping. A single test driving an error with `'Failed: https://пример.com/page\r\nX-Injected: yes'` and asserting the header is ASCII-only would lock in this fix.

danieljchuser (approved):
Both findings addressed - nice turnaround:

- **Important (test):** Regression test added covering non-ASCII + CRLF sanitization - exactly what was needed.
- **Minor (undefined guard):** `?? ''` fallback added for `error.message`.

LGTM. The centralized sanitization recommendation remains worth tracking as a follow-up but is not blocking this fix.

PR #2337, opened 2026-05-06
fix(brands): dispatch llmo-customer-analysis on v2 customer-config writes (LLMO-4744)

rainer-friederich:
Pre-cutover blocker for the Adobe brandalf_migration. Companion to adobe/spacecat-audit-worker `#2480`.

## Problem

Under `brandalf`, BrandsController v2 endpoints write directly to the normalized customer-config tables (`brands`, `categories`, `topics`, `prompts`) but **never enqueue an `llmo-customer-analysis` audit**. The legacy v1 PUT path (`controllers/llmo/llmo.js:586-595`) was the only dispatcher, and it is bypassed by elmo-UI / brandalf-UI v2 writes.

As a result, the audit-worker's three downstream triggers (`triggerBrandDetection`, `triggerGeoBrandPresenceRefresh`, CDN-logs report) silently never fired on real user edits — the pre-cutover blocker described in [LLMO-4744](https://jira.corp.adobe.com/browse/LLMO-4744). (The audit-worker side also has a frozen-S3-mirror diff that would have returned empty even if dispatch had fired; that's fixed in the companion PR.)

## Architecture choice

Three options were considered for the cross-repo fix (full notes in chat thread). Picked **Option C — explicit `changeKind` dispatch**: the writer already knows what changed, so pass that signal forward instead of having the audit-worker re-derive it via a diff against a stale source. Smallest blast radius, no DB migration, no snapshot mechanism.

## Fix

Add a thin dispatch helper and wire it into all 13 mutating BrandsController endpoints. The helper enumerates org sites via `Site.allByOrganizationId` and sends one `llmo-customer-analysis` SQS message per site with `auditContext: { onboardingMode: 'v2', organizationId, changeKind }`.

**Endpoints + changeKind:**

| Endpoint | changeKind |
|---|---|
| `createBrandForOrg` / `updateBrandForOrg` / `deleteBrandForOrg` | `'brands'` |
| `createCategoryForOrg` / `updateCategoryForOrg` / `deleteCategoryForOrg` | `'categories'` |
| `createTopicForOrg` / `updateTopicForOrg` / `deleteTopicForOrg` | `'topics'` |
| `createPromptsByBrand` / `updatePromptByBrandAndId` / `deletePromptByBrandAndId` / `bulkDeletePromptsByBrand` | `'prompts'` |

The companion audit-worker PR routes by `changeKind`:
- `brands`, `competitors` → `triggerGeoBrandPresenceRefresh`
- `categories`, `topics`, `entities` → `drsClient.triggerBrandDetection`
- `prompts` → both
- CDN logs trigger is skipped (out of scope under brandalf)

### Downstream effect on the BP database

`triggerGeoBrandPresenceRefresh` is not the end of the chain — it ultimately publishes a DRS `JOB_COMPLETED` SNS event with `reanalysis: true`, which causes DRS to re-run BP analysis using the **current Postgres v2 config** and write fresh rows into `brand_presence_executions`. So a brandalf user editing brands/competitors causes the BP database to land eventually-consistent with the edit; we don't run the analysis ourselves, we just signal DRS to re-run it with the new config.

## Helper behavior

- **Never throws.** Failures are logged at warn so the original mutation's HTTP response is unaffected.
- **Dispatch happens only on the success path** (after `notFound`/validation gates). For `bulkDeletePromptsByBrand`, dispatch is gated on `result.metadata.success > 0` so a no-op bulk delete (every requested promptId already deleted) doesn't fire a spurious audit.
- **Defensive checks**: invalid changeKind, missing `AUDIT_JOBS_QUEUE_URL`, missing Site model, no sites for org, per-site SQS failures all warn-and-return cleanly. The final log line reports `Dispatched sent/total message(s)` so partial failures are visible at a glance.
- **Brandalf gating happens in audit-worker.** Sites whose orgs are not on brandalf produce a no-op completed audit. This dispatcher does not need to gate.

## Files

- `src/support/llmo-customer-analysis-dispatch.js` — new helper (`dispatchCustomerAnalysisV2`, `CUSTOMER_ANALYSIS_CHANGE_KINDS`)
- `src/controllers/brands.js` — 13 dispatch call sites + 1 import line
- `test/support/llmo-customer-analysis-dispatch.test.js` — 15 unit tests (helper)
- `test/controllers/brands.test.js` — 5 wiring tests (controller integration)

## Test plan

- Helper unit tests cover all 6 changeKind values, defensive paths, log fallback semantics, and partial-failure log shape
- Controller wiring tests cover: success → dispatch fires with correct changeKind; validation 400 → no dispatch; SQS error → response intact
- All 244 existing brands-controller tests still pass
- `npm test` (affected suites) — 264 passing total
- ESLint clean
- Verified on a brandalf_migration test org: real user edit via brandalf-UI → appropriate downstream trigger fires (post-deploy validation, requires audit-worker PR to be deployed first)

## Deferred follow-ups (not in this PR)

- Bump dispatch-failure logs from `warn` → `error` so silent-no-op cases page on alerts. Currently a "write succeeded but SQS dispatch failed" condition is only visible at warn level.
- Track an explicit retry / DLQ for failed dispatches (today, a failed SQS send is logged and lost — there is no retry).

## Related

- Jira: https://jira.corp.adobe.com/browse/LLMO-4744
- Audit-worker companion: adobe/spacecat-audit-worker `#2480`
- Parent: LLMO-4715
- Sibling: LLMO-4743 (Reader 4 — v1-write ban enforcement)
- Epic: LLMO-4587 (Brand Presence brandalf migration)

## Deployment ordering

**Deploy the audit-worker PR first.** Until it ships, the new SQS messages from this PR will reach the existing audit-worker handler, which will see `auditContext.changeKind` as an unknown field and fall through to the existing v1 diff path (no harm — same as current behavior). Once the audit-worker PR is live, the changeKind branch routes correctly.

## Related Issues

- LLMO-4744

### DIFF
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -66,6 +66,7 @@ import {
   updateTopic,
   deleteTopic,
 } from '../support/topics-storage.js';
+import { dispatchCustomerAnalysisV2 } from '../support/llmo-customer-analysis-dispatch.js';
 
 const HEADER_ERROR = 'x-error';
 
@@ -424,6 +425,8 @@ function BrandsController(ctx, log, env) {
         updatedBy,
       });
 
+      await dispatchCustomerAnalysisV2(context, spaceCatId, 'prompts', log);
+
       return createResponse({ created, updated, prompts: outPrompts }, 201);
     } catch (error) {
       log.error(`Error creating prompts for brand ${brandId}:`, error);
@@ -482,6 +485,9 @@ function BrandsController(ctx, log, env) {
       if (!prompt) {
         return notFound(`Prompt not found: ${promptId}`);
       }
+
+      await dispatchCustomerAnalysisV2(context, spaceCatId, 'prompts', log);
+
       return ok(prompt);
     } catch (error) {
       log.error(`Error updating prompt ${promptId}:`, error);
@@ -538,6 +544,9 @@ function BrandsController(ctx, log, env) {
       if (!deleted) {
         return notFound(`Prompt not found: ${promptId}`);
       }
+
+      await dispatchCustomerAnalysisV2(context, spaceCatId, 'prompts', log);
+
       return createResponse(null, 204);
     } catch (error) {
       log.error(`Error deleting prompt ${promptId}:`, error);
@@ -596,6 +605,8 @@ function BrandsController(ctx, log, env) {
         postgrestClient,
         updatedBy,
       });
+
+      await dispatchCustomerAnalysisV2(context, spaceCatId, 'prompts', log);
 
       return ok(result);
     } catch (error) {
@@ -849,6 +860,8 @@ function BrandsController(ctx, log, env) {
         outcome,
       });
 
+      await dispatchCustomerAnalysisV2(context, spaceCatId, 'categories', log);
+
       // 201 on insert, 200 on idempotent update — lets callers (UI toast,
       // DRS audit log) distinguish "created new" from "ensured existing".
       return createResponse(category, created ? 201 : 200);
@@ -909,6 +922,9 @@ function BrandsController(ctx, log, env) {
       if (!updated) {
         return notFound(`Category not found: ${categoryId}`);
       }
+
+      await dispatchCustomerAnalysisV2(context, spaceCatId, 'categories', log);
+
       return ok(updated);
     } catch (error) {
       // PATCH to a name colliding with a sibling row in the same org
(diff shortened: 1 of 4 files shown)


rainer-friederich:
Self-review pass via local code-review agent. Two valid items addressed in 7fce78d4:

**MEDIUM — misleading "Dispatched N" log on partial failure.** The Promise.all loop swallows per-site SQS failures, so N could overstate successful sends. Now tracks a per-site success counter and logs `Dispatched sent/total message(s)` so operators can see partial-failure at a glance instead of cross-referencing warn lines. Three test assertions updated to match.

**LOW — `bulkDeletePromptsByBrand` dispatched on no-op.** Other 12 mutating endpoints implicitly skip dispatch via `if (!result) return notFound`. `bulkDeletePrompts` returns `{ metadata: { total, success, failure }, failures }` even when zero rows matched, so the 13th endpoint was firing a spurious audit fan-out for stale-id requests. Now gated on `result?.metadata?.success > 0`.

The reviewer also flagged two design notes at lower confidence (~50%) about `isBrandalfEnabled` latency and trigger retry idempotency. Both inherited from existing patterns and not regressions, so left as-is.

All 264 tests pass; ESLint clean.

PR #2338, opened 2026-05-06
fix(prompts-v2): expose category.uuid and topic.uuid on V2Prompt

rainer-friederich:
## Summary

The OpenAPI `V2Prompt` schema documents `uuid` (UUID PK from `categories.id` / `topics.id`) on the embedded category and topic objects, but `mapRowToPrompt` did not emit it. Downstream consumers — notably the DRS bulk-import runner reading `category.uuid` / `topic.uuid` to populate `brand_presence_executions.category_id` / `.topic_id` foreign keys — saw empty strings, leaving those FKs NULL for ~99 % of v2 (brandalf) cohort rows post-2026-05-05.

This PR is strictly additive: the embed gains a `uuid` field while `id` keeps its current value (the UUID PK, not the business key as the OpenAPI schema description suggests) for backward compat. Aligning `id` to the documented business-key semantics is a deferred breaking change.

## Production symptom

For the 82 brandalf-cohort orgs in the 10 h after the LLMO-4716/4720/4721 deploy:

| Check | Result |
|---|---|
| BPE rows where `prompt_id` is orphaned | 0 / 6,725 — every row links cleanly |
| Linked `prompts.topic_id` populated, but `BPE.topic_id` NULL | **6,670 / 6,725 (99.2 %)** ← this PR |
| Linked `prompts.category_id` populated, but `BPE.category_id` NULL | **6,256 / 6,725 (93.0 %)** ← this PR |

The data was always present at the source — `categories(id, …)` and `topics(id, …)` are already in the SELECT — only the response mapping dropped them. No DB query change, no new round-trips.

## Failure path traced

1. DRS reads `category.uuid` at `spacecat-shared/.../spacecat_client.py:1500` and `topic.uuid` at `:1504` of its `map_v2_prompts_to_standard`.
2. api-service today returns `category: { id: <UUID>, name, origin }` (no `uuid` key) → DRS gets empty string.
3. Empty strings flow through `prompts.json` → `results.json` → bulk-import CSV → COPY → BPE NULL.

## Files

- `src/support/prompts-storage.js` — `mapRowToPrompt` adds `uuid: category.id` and `uuid: topic.id` to the embedded objects. `id` unchanged. Comment captures the WHY (FK linkage + deferred business-key alignment).
- `test/support/prompts-storage.test.js` — `getPromptById`'s deep-equal expectation updated to include `uuid`; `listPrompts` test gains explicit `category.uuid` / `topic.uuid` assertions.

## Self-review applied

- **No OpenAPI change required** — `docs/openapi/schemas.yaml` (`V2Prompt`) already documents `uuid` on both embeds. This PR aligns the implementation with the documented contract.
- **No DB change required** — the existing SELECT `categories(id,category_id,name,origin), topics(id,topic_id,name)` already pulls the UUID PK we now expose.
- **`id` semantics preserved** — historically `id` on these embeds carries the UUID PK (not the business key as the OpenAPI description says). Changing that is a breaking change for any consumer reading `category.id` as the UUID. Out of scope here; tracked as a follow-up to align `id` to the documented business-key semantics across all consumers in one coordinated change.
- **No write-path change** — only the read-path mapper (`mapRowToPrompt`) is touched. POST/PATCH/DELETE for prompts are unaffected.
- **Dual call paths covered** — `mapRowToPrompt` is shared by `listPrompts` and `getPromptById`. Both have updated tests.

## Companion PR (not blocking)

[llmo-data-retrieval-service: include brandalf_migration in v1/v2 path resolution](https://github.com/adobe-rnd/llmo-data-retrieval-service/pull/1587) — fixes the related 100 % NULL `prompt_id` symptom for the `brandalf_migration` cohort (Adobe Corp). Independent of this PR.

## Test plan

- [x] `npx mocha test/support/prompts-storage.test.js` — 80 passing (was 79; added regression assertions)
- [x] `npx mocha test/controllers/brands.test.js` — 244 passing
- [x] `npx eslint` — clean
- [ ] CI green
- [ ] Stage smoke: `GET /v2/orgs/:org/brands/:brand/prompts` for a brandalf-flagged org returns items with non-empty `category.uuid` and `topic.uuid` per item
- [ ] Post-deploy verification: re-query `brand_presence_executions` for the brandalf cohort 24 h after deploy — `category_id` / `topic_id` NULL rate drops from ~99 % to ~0 %

### DIFF
--- a/docs/openapi/schemas.yaml
+++ b/docs/openapi/schemas.yaml
@@ -5996,7 +5996,10 @@ V2Prompt:
       properties:
         id:
           type: string
-          description: Business key (category_id)
+          description: |
+            Currently the UUID PK (categories.id) for backward compat —
+            aligning this to the category_id business key is a deferred
+            breaking change. Use `uuid` for FK linkage today.
         uuid:
           type: string
           format: uuid
@@ -6010,14 +6013,15 @@ V2Prompt:
       properties:
         id:
           type: string
-          description: Business key (topic_id)
+          description: |
+            Currently the UUID PK (topics.id) for backward compat —
+            aligning this to the topic_id business key is a deferred
+            breaking change. Use `uuid` for FK linkage today.
         uuid:
           type: string
           format: uuid
           description: UUID primary key (topics.id)
         name:
-          type: string
-        categoryId:
           type: string
 
 V2PromptInput:
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -238,9 +238,17 @@ function mapRowToPrompt(row) {
     updatedBy: row.updated_by,
     brandId: brand?.id ?? null,
     brandName: brand?.name ?? null,
+    // `uuid` is the UUID PK consumers must use for FK linkage (e.g. DRS reads
+    // `category.uuid` / `topic.uuid` to populate
+    // `brand_presence_executions.category_id` / `.topic_id`). Documented in
+    // OpenAPI V2Prompt schema; absent here previously, which produced NULL FKs
+    // for the v2 (brandalf) cohort. `id` kept unchanged for backward compat
+    // (today it carries the UUID, not the business key as the OpenAPI schema
+    // suggests; aligning to schema is a deferred breaking change).
     category: category
       ? {
         id: category.id,
+        uuid: category.id,
         name: category.name,
         origin: category.origin,
       }
@@ -248,6 +256,7 @@ function mapRowToPrompt(row) {
     topic: topic
       ? {
         id: topic.id,
+        uuid: topic.id,
         name: topic.name,
       }
       : null,
--- a/test/support/prompts-storage.test.js
+++ b/test/support/prompts-storage.test.js
@@ -516,16 +516,17 @@ describe('prompts-storage', () => {
         postgrestClient: client,
       });
       expect(result.items).to.have.lengthOf(1);
-      expect(result.items[0].category).to.deep.include({
-        id: 'cat-uuid', name: 'Photoshop', origin: 'human',
+      // deep.equal (not deep.include) — symmetric with getPromptById test below
+      // and locks the exact embed shape so a regression dropping `uuid` (or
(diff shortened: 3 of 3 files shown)


rainer-friederich:
## Self-review follow-up — addressed in 36023d9b

Ran a code-review pass and addressed three valid findings without changing the runtime contract of this PR (no behaviour change for existing consumers; only schema descriptions, a phantom field, and a test assertion).

| # | Finding | Fix |
|---|---|---|
| 1 | `schemas.yaml` `V2Prompt.category.id` and `.topic.id` were described as "Business key (category_id\|topic_id)" but the mapper has historically emitted (and still emits) the UUID PK — the contract drift this PR exists to work around. | Replaced both descriptions with what the mapper actually emits today, plus an explicit note that aligning `id` to the business key is a deferred breaking change. |
| 2 | `schemas.yaml` `V2Prompt.topic.categoryId` was documented but never emitted (topics join select is `topics(id,topic_id,name)`; mapper drops it anyway). | Removed from the schema rather than leave phantom contract surface. If a consumer ever needs it, it can be added deliberately along with the join. |
| 3 | `prompts-storage.test.js` `listPrompts` assertion used `deep.include` plus a separate `category.uuid`/`topic.uuid` check, while the equivalent `getPromptById` test used `deep.equal`. | Replaced with `deep.equal` for symmetry — drops in `uuid` (or accidental added fields) now fail fast in both code paths. |

### Out of scope (not changed)

- The `id === uuid` value duplication on `category` / `topic` is intentional and called out in code comments — required for backward compat until the deferred breaking change.
- Top-level `V2Prompt.categoryId` / `topicId` schema/code drift is a pre-existing gap unrelated to this PR's scope.

### Verification

- `npx mocha test/support/prompts-storage.test.js` — 80 passing (no count change; assertions tightened in place).
- `npx eslint src/support/prompts-storage.js test/support/prompts-storage.test.js` — clean.
- `npm run docs:lint` — passes on `docs/openapi/api.yaml` (261 warnings are all pre-existing in unrelated parts of the spec).

igor-grubic (approved):

PR #2339, opened 2026-05-06
feat: add has-data endpoint for referral traffic | LLMO-4261

calvarezg:
This commit introduces a new endpoint `GET /sites/:siteId/referral-traffic/has-data` that checks for the existence of referral traffic records for a given site. The implementation includes a handler that performs a single PostgREST query to determine if any records exist, returning a boolean response.

- Added `createReferralTrafficHasDataHandler` in `llmo-referral-traffic.js`
- Integrated the handler into `LlmoMysticatController`
- Registered the new route in `src/routes/index.js`
- Comprehensive unit tests added to ensure functionality and error handling

This feature enhances the referral traffic API by allowing quick checks for data availability, improving user experience on the frontend.

## Related Issues
[LLMO-4261](https://jira.corp.adobe.com/browse/LLMO-4261)

### DIFF
--- a/src/controllers/llmo/llmo-mysticat-controller.js
+++ b/src/controllers/llmo/llmo-mysticat-controller.js
@@ -67,6 +67,7 @@ import {
   createReferralTrafficBusinessImpactHandler,
   createReferralTrafficWeeksHandler,
   createReferralTrafficByDeviceHandler,
+  createReferralTrafficHasDataHandler,
 } from './llmo-referral-traffic.js';
 
 /**
@@ -225,6 +226,7 @@ function LlmoMysticatController(ctx) {
   );
   const getReferralTrafficWeeks = createReferralTrafficWeeksHandler(getSiteAndValidateAccess);
   const getReferralTrafficByDevice = createReferralTrafficByDeviceHandler(getSiteAndValidateAccess);
+  const getReferralTrafficHasData = createReferralTrafficHasDataHandler(getSiteAndValidateAccess);
 
   return {
     getFilterDimensions,
@@ -277,6 +279,7 @@ function LlmoMysticatController(ctx) {
     getReferralTrafficBusinessImpact,
     getReferralTrafficWeeks,
     getReferralTrafficByDevice,
+    getReferralTrafficHasData,
   };
 }
 
--- a/src/controllers/llmo/llmo-referral-traffic.js
+++ b/src/controllers/llmo/llmo-referral-traffic.js
@@ -13,6 +13,7 @@
 import {
   ok, badRequest, forbidden, internalServerError,
 } from '@adobe/spacecat-shared-http-utils';
+import { cachedOk } from '../../support/cached-response.js';
 import { generateIsoWeekRange, getWeekDateRange } from './llmo-brand-presence.js';
 
 /**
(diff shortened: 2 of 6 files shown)


JayKid (approved):
¡Pura vida, Char! 🇨🇷 ¡Mae, qué chiva PR te mandaste! El approach con `Promise.all` + `cachedOk` siguiendo el patrón de `agentic-traffic/has-data` quedó tuanis del todo — diay, codecov verde, tests pichudos, y la auth bien amarrada. **Thanks for sharing** tan buen trabajo con el equipo — ¡aquí van algunas observaciones, todas con cariño!

> Synthesized from a multi-persona pass (Architect / DBA / Security / Tester / Junior Dev). Cross-repo schema verified against `mysticat-data-service` migration `20260503000100_referral_traffic_per_source.sql`.

### Major (1)
- **Source coverage vs route name** — handler probes only `referral_traffic_optel` and `referral_traffic_cdn`, but `SOURCE_TO_TABLE` declares `ga4` and `adobe_analytics` as first-class sources. Sites with only AA/GA4 data will receive `{ hasData: false }` and stay stuck on the onboarding overlay. Either (a) extend the parallel check to all four tables (drive from `SOURCE_TO_TABLE` so the list can't drift), or (b) confirm with product that AA/GA4 are intentionally excluded and make that explicit in the route docs / OpenAPI. The JSDoc already says "(optel OR cdn)", but the route name reads broader. (`src/controllers/llmo/llmo-referral-traffic.js:64-67`)

### Minor (7)
- **JSDoc is misleading**: "Returns true as soon as either source has at least one row" implies short-circuit, but `Promise.all` waits for both. Reword to "true if either result set is non-empty". (`llmo-referral-traffic.js:54-55`)
- **Error logs omit `siteId`** — append ` (siteId=${siteId})` for multi-tenant log correlation. Pre-existing pattern in this file, but a good moment to set a better precedent. (`llmo-referral-traffic.js:70,74`)
- **Hardcoded table names** — use `SOURCE_TO_TABLE.optel` / `.cdn` (or a `HAS_DATA_SOURCES` array) so renames don't silently drift. (`llmo-referral-traffic.js:65-66`)
- **`/* c8 ignore next */` is unnecessary** — both `hasData:true` and `false` paths are covered by tests. Other ignores in this file include rationale; this one doesn't and isn't needed. (`llmo-referral-traffic.js:79`)
- **Test gaps** — add (1) both queries error simultaneously (asserts only optel is logged, surfaces the ordering choice), and (2) `limit()` rejecting with a thrown error (handler has no `try/catch`). (`test/controllers/llmo/llmo-referral-traffic.test.js`)
- **Log format diverges** from sibling handlers in the same file (others use the `... PostgREST error` suffix). Align for easier log searches. (`llmo-referral-traffic.js:70,74`)
- **Strict 500 when one branch errors** — if `optel` already returned a row but `cdn` errors, the response is 500 instead of `{ hasData: true }`. Worth an explicit decision: fail-closed (current) vs short-circuit on a known-positive.

### Questions (2)
- Should `siteId` be UUID-validated upfront in `withReferralTrafficAuth` so a malformed id returns a clean 400 instead of round-tripping to PostgREST?
- Was bulk-adding the rest of the `getReferralTraffic*` stubs in `test/routes/index.test.js` strictly required, or is only `getReferralTrafficHasData` needed?

### Cross-repo note (not a defect)
Both queried tables are partitioned by `traffic_date` (RANGE, ~36 monthly partitions Jan 2025 → Dec 2027) per `mysticat-data-service` migration `20260503000100`. The query has no `traffic_date` predicate, so partition pruning isn't possible — Postgres falls back to per-partition index lookups via `idx_referral_traffic_*_site_date (site_id, traffic_date DESC)`. This matches the existing agentic-traffic `has-data` and `weeks` handlers (precedent, not a regression). Adding `traffic_date >= current_date - interval '90 days'` (or aligning with the dashboard's default 28-day window) on both queries would meaningfully cut DB work across the family — worth considering as a follow-up across all three handlers.

### Working well
- 100% line coverage of the new handler (codecov ✅).
- Pattern faithfully mirrors `agentic-traffic/has-data` (factory + `withReferralTrafficAuth` + `cachedOk`).
- Auth runs **before** any DB query — cross-tenant probing returns 403, not data.
- Cache headers (`Cache-Control: private, max-age=7200`) are correct and consistent with the agentic equivalent.
- Schema cross-check passes: `site_id` (uuid) + `traffic_date` (date) exist on both tables; `postgrest_anon` has `SELECT`; `(site_id, traffic_date DESC)` index supports the query well.
- Route is correctly added to `INTERNAL_ROUTES` — UI-only, off the S2S consumer surface.

¡Gracias mae, te quedó cabal! 🙌

igor-grubic (approved):

PR #2340, opened 2026-05-06
fix: reliably deliver CDN opt-in notification email

khushboo5723:
## Summary

Awaiting the opt-in notification work (IMS lookup + S3 read + email send) on the
request path so the email reliably goes out, while parallelising the prep calls
to keep the added latency at ~500ms. This is acceptable as a short-term fix
because opting in is a one-time user action, not a hot path.

Follow-up: move the notification to an SQS-driven flow so the API call doesn't
have to wait at all. Tracked in [LLMO-4814](https://jira.corp.adobe.com/browse/LLMO-4814).

## Issue

The opt-in email logic was wrapped in a detached async IIFE (fire-and-forget
promise) that the Lambda handler never awaited. The original assumption was
that `callbackWaitsForEmptyEventLoop=true` would keep the Lambda alive long
enough for the IIFE to finish.

That assumption only holds for **legacy callback-style** Lambda handlers.
This service uses an **async handler**, and with async handlers Lambda freezes
the container the moment the handler's returned promise resolves — abandoning
any detached promises mid-execution.

Evidence in Coralogix:

- `[edge-optimize-config] Slack notification sent` — [sync path log](https://spacecat.coralogix.com/#/query-new/logs?id=xcW12xms5jRRkJ1hmzpSV&page=0) (worked correctly)
- Zero `[cdn-opt-in-notification]` [logs](https://spacecat.coralogix.com/#/query-new/logs?id=6L2TYGHJHDIiYUL0TA1MB&page=0) in the same invocation — the async IIFE never ran

## Fix

1. **Await** the notification flow inside the `if (isNewlyOpted)` block instead
   of detaching it. This guarantees the Lambda doesn't freeze before the email
   is sent.
2. **Parallelise** the two independent prep calls — IMS profile lookup and S3
   LLMO-config read — so they overlap with `saveSiteConfig` instead of running
   serially after it. Total added latency drops from ~5–10s to ~500ms.
3. Added structured `step=…` timing logs (`ims-resolved`, `s3-resolved`,
   `save-config-done`, `slack-done`, `email-done`, `complete`) so we can keep
   verifying latency in Coralogix.

Trade-off: the customer clicking "enable" now waits ~500ms longer for the API
response. Acceptable because:
- opt-in is a one-time per-site action, not a frequent API call
- the alternative (lost emails) blocks downstream LLMO team workflow

## Next step

Move the notification work onto an SQS queue so the API handler enqueues a
message and returns immediately. The worker consumes the message and runs
IMS + S3 + send-email asynchronously. Tracked in [LLMO-4814](https://jira.corp.adobe.com/browse/LLMO-4814).


dipratap (commented):
**1. Double logging on IMS / S3 failures**

The `.catch()` on `imsPromise` / `s3Promise` logs a warning and re-throws. The consuming `else` branch then logs a second warning for the same error — every IMS or S3 failure produces two log lines.

```js
// warn #1 fires here…
const imsPromise = ...
  .catch((e) => {
    log.warn(`[cdn-opt-in-notification] IMS lookup failed for site=${siteId}: ${e.message}`);
    throw e;
  });

// …warn #2 fires here for the same error
} else {
  log.warn(`[cdn-opt-in-notification] Could not resolve user email from IMS: ${adminProfile.reason?.message}`);
}
```

The `.catch()` blocks are redundant — `Promise.allSettled` captures rejections regardless. Suggest removing them and keeping the logging only in the consuming `else` branches.

dipratap (approved):

PR #2341, opened 2026-05-06
feat(llmo): add offsite LLMO audits to paid onboarding profile

HollywoodTonight:
wikipedia-analysis was missing from the paid profile in profiles.json. Sites onboarded via the Slack modal or HTTP onboarding endpoint (profile-based path) never had these audits enabled or triggered, because the profile controls which audits are passed to enableHandlerForSite and triggerAuditForSite.

Non-LLMO paid clients are unaffected: isAuditEnabledForSite gates execution on LLMO product enrollment, so the audits are silently skipped for any site without an active LLMO enrollment.

## Related Issues

Thanks for contributing!

### DIFF
--- a/static/onboard/profiles.json
+++ b/static/onboard/profiles.json
@@ -129,7 +129,8 @@
 			"accessibility": {},
 			"hreflang": {},
 			"security-vulnerabilities": {},
-			"security-vulnerabilities-auto-suggest": {}
+			"security-vulnerabilities-auto-suggest": {},
+			"wikipedia-analysis": {}
 		},
 		"imports": {
 			"organic-traffic" : {},


grigoreadobe (commented):
Hey @HollywoodTonight,

## Strengths

- `static/onboard/profiles.json`: The added `"wikipedia-analysis": {}` entry is correctly placed inside the `paid.audits` block, consistent with the pattern used by all other audits in that profile.
- CI is fully green.

## Issues

### Critical (Must Fix)
None.

### Important (Should Fix)

**static/onboard/profiles.json - PR description and diff are out of sync**

The PR description states:

> "wikipedia-analysis, reddit-analysis, youtube-analysis and cited-analysis were missing from the paid profile"

The diff only adds `wikipedia-analysis`. The other three audits (`reddit-analysis`, `youtube-analysis`, `cited-analysis`) are absent from the `paid` profile in the current diff and are not present in the file at all.

This is either:
1. **Intentional** - only `wikipedia-analysis` ships in this PR, and the description needs to be updated to reflect that.
2. **An oversight** - the other three audits were meant to be included here but were missed.

Please clarify which case this is. If it is intentional, update the PR description to match the actual scope. If it is an oversight, add the missing three entries before merging.

### Minor (Nice to Have)
None.

## Recommendations

- If `reddit-analysis`, `youtube-analysis`, and `cited-analysis` are deferred to follow-up PRs, consider adding a note to the PR description (or a comment) tracking those follow-ups so they do not get lost.

## Assessment

Ready to merge? No / With clarification

The single-line change is correct and safe to merge as-is. The blocker is the unresolved question about whether the other three audits are intentionally excluded or accidentally omitted. Update the PR description (or the diff) to resolve the ambiguity before merging.

HollywoodTonight:
We don't add youtube, reddit or cited for now

rarescheseli (approved):

PR #2343, opened 2026-05-06
fix: add log trail for autofix flow to address sec review

mimchome:
## Summary
  - Adds `triggeredBy` (IMS user ID) to the `[autofix-triggered]` structured log event in the autofix suggestions endpoint
  - The value comes from `context.attributes.authInfo.profile.email` which, despite the field name, contains the IMS user ID (e.g. `82521D...@AdobeOrg`), not an actual email       
  address — no PII is logged                                                                                                                                                        
                                                                                                                                                                                    
  ## Why                                                                                                                                                                            
  Security review ([SITES-42941](https://jira.corp.adobe.com/browse/SITES-42941) eds flow needs to track who triggers since auth is based on shared credentiall 
https://wiki.corp.adobe.com/spaces/WEM/pages/3839545261/Threat+Model+SpaceCat+Self-Service+Credential+Management


danieljchuser (changes requested):
Hey @mimchome,

## Strengths

- Structured key-value shape with `[autofix-triggered]` bracket-tag at `suggestions.js:1316-1325` is the right direction for security-relevant events - more queryable in Coralogix than the existing string-interpolated audit logs at lines 1973 and 2407.
- PII claim is verified by code: `spacecat-shared-http-utils/src/auth/handlers/ims.js`'s `transformProfile` does `profile.email = payload.user_id`, so the JWT/IMS path logs the IMS user ID, not a real email. Good due diligence in the PR description.
- Tests cover all four gating conditions (`!precheckOnly && succeededSuggestions.length > 0`): positive path, null-profile fallback, precheckOnly skip, zero-success skip. Codecov confirms full line coverage.
- Conditional emission is correctly scoped - dry-run prechecks and no-op dispatches do not produce misleading audit events.

## Issues

### Important (Should Fix)

**1. `context.log?.info` and deep optional chaining diverge from every other call site in the file** - `src/controllers/suggestions.js:1317`

All ~30 other log sites in the file use `context.log.info` directly. All 6 existing `authInfo` access sites use non-optional destructuring: `const { authInfo: { profile } } = context.attributes;`. This line is the only use of `context.log?.info` and the only use of `context.attributes?.authInfo?.profile?.email`. If `context.log` is falsy, the audit event silently no-ops with no error - the worst failure mode for a log added at security-review request. The request is already authenticated by the time this code runs, so `context.attributes.authInfo` cannot legitimately be missing.

Fix: drop `?.` on `context.log`, and destructure `profile` at handler scope to match the file's convention:
```js
const { authInfo: { profile } } = context.attributes;
// ... later:
context.log.info('[autofix-triggered]', {
  // ...
  triggeredBy: profile?.email || profile?.user_id || 'autofix-service',
});
```

**2. `triggeredBy: 'unknown'` does not actually attribute the shared-credential path** - `src/controllers/suggestions.js:1323`

The threat model (SITES-42941) is specifically about shared-credential flows, but `profile.email` is only populated by `JwtHandler`/`AdobeImsHandler`. For `ScopedApiKeyHandler` and `LegacyApiKeyHandler`, there is no `.email` property - every such call logs `triggeredBy: 'unknown'`. The exact path the security review wants tracked produces no attribution. The existing peers at lines 1973/2407 use named system-actor strings (`'tokowaka-deployment'`, `'tokowaka-rollback'`) rather than `'unknown'`.

Fix: fall back through `profile.user_id` or `profile.name` before defaulting, and consider adding `authType` to distinguish JWT users from API-key callers:
```js
triggeredBy: profile?.email || profile?.user_id || profile?.name || 'unknown',
```

**3. Audit log fires only on full SQS-dispatch success, leaving IN_PROGRESS state unattributed on partial failure** - `src/controllers/suggestions.js:1316`

By this line, `bulkUpdateStatus(IN_PROGRESS)` has already committed to the database (line 1231). If any SQS send in the `Promise.all` group throws, suggestions are stuck IN_PROGRESS with no audit record of who triggered it. For the threat model's "post-incident attribution" goal, the *intent* is what needs recording - regardless of dispatch outcome.

Note: there is a reasonable alternative framing where "audit log after dispatch" means "confirmed dispatch" rather than "attempted dispatch." If that is the intended semantic, document it explicitly in a code comment.

Fix options: (a) move the log before the SQS dispatch block to capture intent, or (b) wrap dispatch in try/finally and include an `outcome` field.

**4. Grouped-SQS-dispatch path is not tested for audit log emission** - `test/controllers/suggestions.test.js`

The code branches on `shouldGroupSuggestionsForAutofix(opportunity.getType())` at line 1279. The only positive log test uses `meta-tags` (single-dispatch path); the `alt-text` test (typically grouped) explicitly sets `precheckOnly: true`, which skips the log. Add a test where `shouldGroupSuggestionsForAutofix` returns true with `precheckOnly: false`, and verify the audit log fires with the expected count.

### Minor (Nice to Have)

**5. Test mutates shared `context` without try/finally** - `test/controllers/suggestions.test.js:4339-4340`

`context.attributes.authInfo.profile = null` then restores after assertion. If the assertion throws, the restore never runs and subsequent tests see a poisoned context. Wrap in try/finally, clone context for this case, or use `sandbox.replace`.

**6. Field name `triggeredBy` carrying an IMS user ID is misleading** - `src/controllers/suggestions.js:1323`

The value is `82521D...@AdobeOrg` (IMS user ID), not an email, despite originating from `profile.email`. A future Coralogix query author will write `triggeredBy LIKE '%@adobe.com'` and miss everything. Consider `triggeredByImsUserId` or adding a brief code comment explaining the IMS user ID semantic. The neutral field name `triggeredBy` is fine if annotated.

## Recommendations

- Add `requestId` (from the request context) to the audit payload for cross-service log correlation - single highest-leverage forensic field for the stated threat model.
- Consider codifying an actor-audit helper (e.g., `logActorAction(context, tag, fields)`) while only three audit logs exist in this controller. Without it, the next handler in the SITES-42941 sweep will invent a fourth shape.
- The existing string-interpolated audit logs at lines 1973 and 2407 should converge on this structured key-value form in a follow-up. The structured shape here is the better direction.

## Assessment

Ready to merge? With fixes.

The change is directionally correct and well-tested, but as written the `'unknown'` fallback does not meet the stated threat-model goal for the shared-credential path, and the post-dispatch placement creates an audit gap on partial failure. Both are fixable with a few additional lines in this same diff.

## Next Steps

1. Address the optional-chaining inconsistency and the `'unknown'` fallback (issues 1-2) - these are the highest priority since they affect audit-trail correctness for the exact path SITES-42941 targets.
2. Decide on audit-log placement semantic (issue 3) - intent vs confirmed-dispatch.
3. Add the grouped-dispatch test case (issue 4).
4. Minor items (5-6) are optional but cheap.

src/controllers/suggestions.js
```diff
@@ -1313,6 +1313,17 @@ function SuggestionsController(ctx, sqs, env) {
       );
     }
 
+    if (!precheckOnly && succeededSuggestions.length > 0) {
+      context.log?.info('[autofix-triggered]', {
```
danieljchuser: All ~30 other log sites in this file use `context.log.info` directly; all 6 `authInfo` access sites use non-optional destructuring. This is the only `context.log?.info` and the only `context.attributes?.authInfo?.profile?.email` in the file. If `context.log` is ever falsy, this audit event silently no-ops - the worst failure mode for a security-mandated log. Drop the `?.` and destructure `profile` at handler scope to match the file's convention.

src/controllers/suggestions.js
```diff
@@ -1313,6 +1313,17 @@ function SuggestionsController(ctx, sqs, env) {
...
+        siteId,
+        opportunityId,
+        opportunityType: opportunity.getType(),
+        action: action || 'apply',
+        succeededSuggestionCount: succeededSuggestions.length,
+        triggeredBy: context.attributes?.authInfo?.profile?.email || 'unknown',
```
danieljchuser: `profile.email` is only populated by `JwtHandler`/`AdobeImsHandler`. For `ScopedApiKeyHandler` and `LegacyApiKeyHandler` there is no `.email` property, so every API-key call logs `triggeredBy: 'unknown'` - producing no attribution on the exact shared-credential path SITES-42941 targets. The existing peers at lines 1973/2407 use named system-actor strings instead. Consider falling back through `profile.user_id` / `profile.name` before defaulting.

src/controllers/suggestions.js
```diff
@@ -1313,6 +1313,17 @@ function SuggestionsController(ctx, sqs, env) {
       );
     }
 
+    if (!precheckOnly && succeededSuggestions.length > 0) {
```
danieljchuser, line 1331: By this line, `bulkUpdateStatus(IN_PROGRESS)` has already committed DB state (line 1231). If any SQS send throws, suggestions are stuck IN_PROGRESS with no audit record of who triggered it. For post-incident attribution, the *intent* is what needs recording. Consider moving the log before dispatch (capturing intent), or wrapping dispatch in try/finally with an `outcome` field. If "after dispatch" is the intended semantic (meaning "confirmed dispatch"), a brief code comment would help future readers.

mimchome:
@danieljchuser  Thanks for the thorough review. Context on the threat model and responses below.                                                                                                  
                                                         
  **Why this log matters:** The security review (SITES-42941) targets the EDS autofix flow specifically — a user triggers a fix from the UI via IMS auth, but the actual CMS write  
  is executed using shared service credentials (per-customer SharePoint/Google OAuth secrets resolved by the autofix-worker or Mystique). The audit gap was: who clicked "apply"?
  The `auto_fix` scope check at line 1043 requires IMS org membership, so the primary callers are always IMS-authenticated users. The tokowaka edge-deploy flow (lines 1957/2310) is
   a separate code path with its own named system-actor attribution — not the same concern.

  **Changes made to address the review:**

  1. **Optional chaining (agree, fixed):** Dropped `?.` on `context.log`. Destructured `profile` from `context.attributes.authInfo` non-optionally — follows file convention. Fixed 
  7 tests that had incomplete contexts (missing `attributes`) to properly include auth context.
                                                                                                                                                                                    
  2. **Fallback chain (partial agree, fixed):** Added `profile?.name` fallback for the admin API-key edge case. The primary flow (IMS user from UI) always has `profile.email`      
  populated. The tokowaka named actors are for system-initiated edge deploys — a different flow, not a pattern needed here.
                                                                                                                                                                                    
  3. **Log placement (intentional — added second log):** Added `[autofix-attempt]` **before** dispatch to capture intent, kept `[autofix-triggered]` **after** dispatch to confirm  
  delivery. If SQS fails, you have the attempt record. If it succeeds, you have both. Both share the same audit fields including `triggeredBy`.
                                                                                                                                                                                    
  4. **Grouped dispatch test (agree, added):** Added test for alt-text (grouped path) verifying both `[autofix-attempt]` and `[autofix-triggered]` fire.                            
   
  5. **try/finally in test (agree, fixed).**                                                                                                                                        
                                                         
  6. **Field name (added comment):** `profile.email` contains the IMS user ID (e.g. `82521D...@AdobeOrg`), not an email — this is because `transformProfile` in `ims.js:68` sets    
  `profile.email = payload.user_id`. Added a code comment explaining this semantic.
                                                                                                                                                                                    
  **Recommendations (noted for follow-up):** `requestId`, `logActorAction` helper, and converging tokowaka logs are good ideas but scope creep for this PR. Keeping it focused on   
  the audit gap.

danieljchuser (approved):
All findings addressed. The dual-log approach (`[autofix-attempt]` before dispatch + `[autofix-triggered]` after) is a nice solution - captures both intent and confirmation with shared `auditFields`, so forensics gets the full picture whether dispatch succeeds or fails.

Changes verified:
- Optional chaining removed, `profile` destructured at handler scope per file convention
- Fallback chain covers the IMS primary flow and API-key edge case via `profile?.name`
- Grouped dispatch path now tested
- try/finally on context mutation
- Code comment on the IMS user ID semantic

LGTM. CI build pending - verify it passes before merge.

PR #2342, opened 2026-05-06
feat(llmo): expose referral-traffic trend-by-url endpoint | LLMO-4729

JayKid:
## Summary

URL Inspector PG's URL Details dialog needs a per-URL × per-week referral pageview series. The companion mysticat PR (adobe/mysticat-data-service#544) adds a new `rpc_referral_traffic_trend_by_url` RPC; this PR exposes it through a new spacecat endpoint at `GET /sites/:siteId/referral-traffic/trend-by-url` and forwards a new `urlPathSearch` query param to `p_url_search` on the RPC.

This is the second of three coordinated PRs for [LLMO-4729](https://jira.corp.adobe.com/browse/LLMO-4729). It must deploy after mysticat#544 is on DEV (the schema cache reload there is what makes the new RPC callable through PostgREST).

### `src/controllers/llmo/llmo-referral-traffic.js`

- Adds `createReferralTrafficTrendByUrlHandler` factory mirroring `createReferralTrafficTrendHandler` exactly, plus a `urlPathSearch` (alias `url_path_search`) query param mapped to `p_url_search` on the RPC.
- Reuses the existing `commonRpcParams`, `withReferralTrafficAuth`, source defaulting, and date-window parsing — single source of truth for referral-traffic parameter handling.
- Returns `{ trend: [{ date, pageviews }] }` (same shape as `/trend`), with `pageviews` coerced to `Number` defensively.
- 500 on PostgREST error (mirrors `/trend`).

### `src/controllers/llmo/llmo-mysticat-controller.js`

- Imports the new handler factory and exposes `getReferralTrafficTrendByUrl` on the controller.

### `src/routes/index.js`

- Registers `GET /sites/:siteId/referral-traffic/trend-by-url → llmoMysticatController.getReferralTrafficTrendByUrl`.

### `src/routes/required-capabilities.js`

- Adds the new path to the LLMO capability list (same auth as the rest of `/sites/:siteId/referral-traffic/*`).

### Tests

- `test/controllers/llmo/llmo-referral-traffic.test.js` — new \`trend-by-url\` describe block covering:
  - `urlPathSearch` (camelCase) and `url_path_search` (snake_case) both forward to `p_url_search`
  - default source = `optel`
  - dates and dimension filters thread through
  - RPC row mapping (`traffic_date → date`, `total_pageviews → pageviews`)
  - empty `data` returns `{ trend: [] }`
  - 400 when `postgrestService` missing
  - 500 on PostgREST error
[...]

PR #2346, opened 2026-05-06
feat: forward referralSource + map referralHits/referralHitsTrend on owned-urls handler | LLMO-4729

JayKid:
## Summary

Adds the spacecat half of LLMO-4729's Decision A pull-in: forwards an optional `referralSource` query parameter to `rpc_url_inspector_owned_urls` and maps the new `referral_hits` / `referral_hits_trend` columns to `referralHits` / `referralHitsTrend` on the response. Pairs with mysticat-data-service [#547](https://github.com/adobe/mysticat-data-service/pull/547) (RPC contract) and project-elmo-ui [#1685](https://github.com/adobe/project-elmo-ui/pull/1685) (UI consumption).

The Owned URLs table on `/url-inspector-pg` has always rendered `N/A` for the **Referral Hits** column because the underlying RPC returned `agentic_hits` / `agentic_hits_trend` (LLMO-4526 Option C JOIN) but no symmetric per-URL referral aggregate. With these three PRs landing together, the column will populate end-to-end with real numbers + sparklines + sortability + CSV export.

### `src/controllers/llmo/llmo-url-inspector.js`

- **New `VALID_REFERRAL_SOURCES` whitelist + `parseReferralSource` helper** that mirror the controller-level pattern in `llmo-referral-traffic.js` (`{ optel, cdn, adobe_analytics, ga4 }`, default `'optel'`). Whitelist on insert with fall-back to `'optel'` for unknown / missing values — defence-in-depth so hostile input never hits the RPC.
- **`getUrlInspectorOwnedUrls` reads the optional `referralSource` query param** (camelCase + snake_case alias both supported). The result is **always** forwarded as `p_referral_source`, even when defaulting, so the wire shape is explicit and the handler tests can pin down the routing. The RPC also has `DEFAULT 'optel'` as a belt-and-braces second line of defence.
- **The row mapper now copies `r.referral_hits` → `referralHits` and `r.referral_hits_trend` → `referralHitsTrend`** with the same null-coercion shape as the existing agentic mapping (`Number(... ?? 0)`, `weekStart` fallback to `null`, `value` coerced via `Number(... ?? 0)`).

### `test/controllers/llmo/llmo-url-inspector.test.js`

- The existing default-mapping test ("returns paginated owned URLs") was extended to assert `referralHits` / `referralHitsTrend` round-trip end-to-end with `weekStart` normalisation.
- The existing nulls-handling test ("passes filters and handles null row fields") now asserts the referral columns also collapse to safe defaults **and** that the handler defaults `p_referral_source` to `'optel'` when the `referralSource` query param is absent.
- **New test:** pins the null-coercion path for `referral_hits_trend` points (`weekStart` → `null`, `value` → `0`) — symmetric companion to the agentic null-coercion test.
- **4 new tests for `referralSource` forwarding:**
  - forwards `referralSource` query param as `p_referral_source`
  - accepts `referral_source` snake_case alias
  - falls back to `'optel'` for unknown values (`{ referralSource: 'nope' }`)
  - round-trips every whitelist member (`optel`, `cdn`, `adobe_analytics`, `ga4`) verbatim — pin so a future contributor adding/removing a source must update this assertion in lock-step with the controller and the RPC's `CASE` branches

### Branch coverage

Verified via `npm test` (full suite, **9,012 passing**, took ~3min):

```
File                                   | % Stmts | % Branch | % Funcs | % Lines
---------------------------------------|---------|----------|---------|--------
All files                              |     100 |      100 |   98.75 |     100
 src/controllers/llmo                  |     100 |      100 |   98.03 |     100
```

100% statements / branches / lines globally. All previously-existing branch-coverage gates pass — including the one that bit us in the original PR #2342 review (the `ctx.data || {}` fallback on line 217).

## Cross-linked PRs

This work spans three repos. **Land in this order:** mysticat → spacecat → elmo (each downstream consumes the previous repo's contract).

- **mysticat-data-service** [#547](https://github.com/adobe/mysticat-data-service/pull/547) — RPC contract (`p_referral_source` parameter + `referral_hits` / `referral_hits_trend` columns)
- **spacecat-api-service** — _this PR_ (handler maps the new fields + forwards the new query param)
- **project-elmo-ui** [#1685](https://github.com/adobe/project-elmo-ui/pull/1685) — extends `OwnedUrlRow` + projects the new fields onto the `URLData` rows that feed the Owned URLs table

## Test plan

- [x] Local: `npx mocha test/controllers/llmo/llmo-url-inspector.test.js` — all 86 handler tests pass
- [x] Local: `npm test` (full suite) — 9,012 passing, **100% branch coverage**
- [ ] CI green
- [ ] End-to-end verification on the elmo PR (#1685): Owned URLs table renders real `Referral Hits` values + sparklines for adobe.com on local dev pointing at DEV
- [ ] Side-by-side parity walk vs `/url-inspector` (HLX) for adobe.com

## Related tickets

- [LLMO-4729](https://jira.corp.adobe.com/browse/LLMO-4729) — _this work_; URL Inspector PG referral migration (Decision A pull-in)
- [LLMO-4526](https://jira.corp.adobe.com/browse/LLMO-4526) — parent / pattern source; agentic side of the same JOIN
- [LLMO-4261](https://jira.corp.adobe.com/browse/LLMO-4261) — referral PG infrastructure (referral_traffic_* tables + RPCs); prerequisite

### DIFF
--- a/src/controllers/llmo/llmo-url-inspector.js
+++ b/src/controllers/llmo/llmo-url-inspector.js
@@ -48,6 +48,18 @@ import { cachedOk } from '../../support/cached-response.js';
  *   - `brandId === 'all'` or missing → no brand filter (`p_brand_id = NULL`).
  */
 
+// Mirror the referral controller's source whitelist (LLMO-4261).
+// Used by the owned-urls handler to forward `p_referral_source` to
+// `rpc_url_inspector_owned_urls` (LLMO-4729 Decision A pull-in). Unknown /
+// missing values fall back to 'optel' for parity with /url-inspector.
+const VALID_REFERRAL_SOURCES = new Set(['optel', 'cdn', 'adobe_analytics', 'ga4']);
+const DEFAULT_REFERRAL_SOURCE = 'optel';
+
+function parseReferralSource(q) {
+  const raw = q.referralSource ?? q.referral_source;
+  return VALID_REFERRAL_SOURCES.has(raw) ? raw : DEFAULT_REFERRAL_SOURCE;
+}
+
 /**
  * Resolve platform/model from request. Returns null when absent (no default model).
  * When provided, validates against the llm_model enum.
@@ -201,6 +213,12 @@ export function createUrlInspectorStatsHandler(getOrgAndValidateAccess) {
  * so owned URLs ranked beyond the top 500 silently showed `agenticHits = 0`.
  * Doing the JOIN in the RPC means the table can paginate 50 owned URLs at a
  * time without losing fidelity.
+ *
+ * Server-side referral merge (LLMO-4729 Decision A pull-in): each row also
+ * carries `referralHits` and `referralHitsTrend` joined from
+ * `referral_traffic_<source>` for the same site / date range, scoped by the
+ * `referralSource` query param (default `'optel'`). Replaces the always-N/A
+ * Referral Hits column the table used to render before this work landed.
  * @param {Function} getOrgAndValidateAccess - Async (context) => { organization }
  */
 export function createUrlInspectorOwnedUrlsHandler(getOrgAndValidateAccess) {
@@ -236,6 +254,12 @@ export function createUrlInspectorOwnedUrlsHandler(getOrgAndValidateAccess) {
       const filterByBrandId = brandId && brandId !== 'all' ? brandId : null;
       const offset = pagination.page * pagination.pageSize;
       const agentTypes = parseAgentTypes(q.agentTypes ?? q.agent_types);
+      // Always forward p_referral_source so the RPC reads from a deterministic
+      // source table — unknown / missing values collapse to 'optel' (mirrors
+      // /url-inspector and the controller default in llmo-referral-traffic.js).
+      // The RPC also has DEFAULT 'optel', but forwarding makes the wire shape
+      // explicit and lets the handler tests pin down the routing.
+      const referralSource = parseReferralSource(q);
 
       // Only forward p_agent_types when the caller actually supplied a value.
       // Omitting the param keeps the RPC contract compatible with internal
@@ -250,6 +274,7 @@ export function createUrlInspectorOwnedUrlsHandler(getOrgAndValidateAccess) {
         p_brand_id: filterByBrandId,
         p_limit: pagination.pageSize,
         p_offset: offset,
+        p_referral_source: referralSource,
       };
       if (agentTypes) {
         rpcParams.p_agent_types = agentTypes;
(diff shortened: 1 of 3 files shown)


calvarezg:
## Code Review

### Issues

None blocking — logic is correct and back-compat strategy is sound.

---

### Suggestions

**1. Extract the trend mapper to avoid duplication**

The `referral_hits_trend` mapper is byte-for-byte identical to `agentic_hits_trend` with only field names swapped. A shared helper prevents drift when a third trend column arrives:

\`\`\`js
const mapTrend = (trend) => (Array.isArray(trend)
  ? trend.map((p) => ({ weekStart: p.week_start ?? null, value: Number(p.value ?? 0) }))
  : []);
\`\`\`

Then in the row mapper:
\`\`\`js
agenticHitsTrend: mapTrend(r.agentic_hits_trend),
referralHitsTrend: mapTrend(r.referral_hits_trend),
\`\`\`

**2. Use `Promise.all` in the whitelist round-trip test**

`forwards every whitelisted referralSource verbatim` uses a `for...of` loop with two `no-await-in-loop` suppressions. Since the mock is stateless, `Promise.all` avoids the linter workaround and is idiomatic:

\`\`\`js
await Promise.all(['optel', 'cdn', 'adobe_analytics', 'ga4'].map(async (source) => {
  const { context, rpcStub } = createContext(
    {}, { referralSource: source },
    { rpcResults: { rpc_url_inspector_owned_urls: { data: [], error: null } } },
  );
  const handler = createUrlInspectorOwnedUrlsHandler(getOrgAndValidateAccess());
  await handler(context);
  expect(
    rpcStub.firstCall.args[1].p_referral_source,
    \`whitelist member '\${source}' should round-trip\`,
  ).to.equal(source);
}));
\`\`\`

---

### Minor

**Trim the comment block before `VALID_REFERRAL_SOURCES`**

The 12-line comment explains deployment sequencing, PGRST202 codes, and RPC versioning history — that context belongs in the PR description, not the source. The constraint worth preserving inline is just the omit-vs-default distinction:

\`\`\`js
// Omit (not default) when absent — keeps older RPC signatures compatible (LLMO-4729).
const VALID_REFERRAL_SOURCES = new Set(['optel', 'cdn', 'adobe_analytics', 'ga4']);
\`\`\`

Same applies to the multi-paragraph comments in the test file; most restate the PR description and will rot as the code evolves.

**Document the absent-vs-invalid asymmetry on `parseReferralSource`**

Absent input returns `undefined` (param omitted); invalid input returns `'optel'` (param forwarded). Both reach the same RPC result today but the distinction isn't obvious from the function signature. A short note on the return contract would help future contributors:

\`\`\`js
// Returns undefined when no value supplied (caller omits p_referral_source entirely),
// or the validated source / 'optel' fallback when a value is present.
function parseReferralSource(q) { ... }
\`\`\`

calvarezg (approved):

PR #2348, opened 2026-05-06
fix(llmo): use server-side RPC for sentiment-overview to fix missing weeks

jjenscodee:
## Summary

- Replaces the raw-row fetch + JS aggregation in `createSentimentOverviewHandler` with a call to the new `rpc_brand_presence_sentiment_overview` RPC (see companion PR in `mysticat-data-service` #548)
- The old approach fetched up to 200k raw rows and aggregated in `aggregateSentimentByWeek()`; a single week now has 290k–436k rows, silently dropping later weeks from the response
- The RPC returns one aggregated row per ISO week — no row-limit issue regardless of data volume
- Adds `brandPresenceRpcRejectsMultiCategoryOrRegion` guard (already used by all other RPC-backed handlers) — multiple `categoryIds` or `regionCodes` now return 400 instead of silently truncating to the first value
- Updates all `createSentimentOverviewHandler` unit tests to mock `client.rpc` instead of the PostgREST table chain

## Root cause

`WEEKS_QUERY_LIMIT = 200000` was exhausted by W17 alone (290,938 raw rows), leaving W18 (436,888 raw rows) with zero representation in the response. "Last 2 weeks" therefore returned only 1 week. "Last 4 weeks" appeared correct only because it was stale Fastly CDN cache from when row volumes were much lower.

## Related issues

https://jira.corp.adobe.com/browse/LLMO-4759

## Test plan

- [ ] `npm test` passes (512 tests)
- [ ] Deploy `mysticat-data-service` migration **first**, then this service
- [ ] Verify `?startDate=2026-04-20&endDate=2026-05-03` returns both W17 and W18
- [ ] Verify `?startDate=2026-04-06&endDate=2026-05-03` returns all 4 weeks

### DIFF
--- a/src/controllers/llmo/llmo-brand-presence.js
+++ b/src/controllers/llmo/llmo-brand-presence.js
@@ -1608,6 +1608,10 @@ export function createSentimentOverviewHandler(getOrgAndValidateAccess) {
       const organizationId = spaceCatId;
       const filterByBrandId = brandId && brandId !== 'all' ? brandId : null;
 
+      if (brandPresenceRpcRejectsMultiCategoryOrRegion(params)) {
+        return badRequest(MULTI_CATEGORY_REGION_RPC_UNSUPPORTED);
+      }
+
       const startDate = params.startDate || defaults.startDate;
       const endDate = params.endDate || defaults.endDate;
       const model = resolveModelFromRequest(params.model);
--- a/src/controllers/llmo/llmo-brand-presence.js
+++ b/src/controllers/llmo/llmo-brand-presence.js
@@ -1612,36 +1612,6 @@ export function createSentimentOverviewHandler(getOrgAndValidateAccess) {
       const endDate = params.endDate || defaults.endDate;
       const model = resolveModelFromRequest(params.model);
 
-      let q = client
-        .from('brand_presence_executions_active')
-        .select('execution_date, sentiment, prompt, region_code, topics')
-        .eq('organization_id', organizationId)
-        .gte('execution_date', startDate)
-        .lte('execution_date', endDate)
-        .eq('model', model);
-
-      if (shouldApplyFilter(params.siteId)) {
-        q = q.eq('site_id', params.siteId);
-      }
-      if (filterByBrandId) {
-        q = q.eq('brand_id', filterByBrandId);
-      }
-      q = applyExecutionCategoryFilter(q, params);
-      if (params.topicIds?.length > 0) {
-        q = q.in('topic_id', params.topicIds);
-      }
-      q = applyExecutionRegionFilter(q, params);
-      if (shouldApplyFilter(params.origin)) {
-        q = q.ilike('origin', params.origin);
-      }
-
-      const { data, error } = await q.limit(WEEKS_QUERY_LIMIT);
-
-      if (error) {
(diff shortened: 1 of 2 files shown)


calvarezg:
## Code Review

The core fix is correct — pushing aggregation to the DB via RPC is the right call and directly solves the row-limit root cause. A few issues worth addressing before merging:

### 1. Silent multi-value filter truncation (medium)

The RPC call passes only the first value for categories and regions:

```js
p_category_id: params.categoryUuidIds?.[0] ?? null,
p_category_name: params.categoryNames?.[0] ?? null,
p_region_code: params.regionCodesList?.[0] ?? null,
```

The old PostgREST query supported multi-value filters (`q.in('category_id', [u1, u2])`, `q.in('region_code', ['US', 'DE'])`, and mixed UUID+name via `or()`). These are now silently truncated to a single value — callers passing multiple values will get wrong/incomplete results without any error.

The new tests document this (`passes first category UUID to RPC when multiple UUID categories provided`) but don't fail on it. Recommend one of: return an error for multi-value inputs, update the RPC to accept arrays, or explicitly document the new API limitation.

### 2. Percentage rounding consistency (minor)

`aggregateSentimentByWeek` ensured percentages summed to exactly 100 via `neutralPct = 100 - positivePct - negativePct`. The RPC now returns raw floats (`positive_pct`, `neutral_pct`, `negative_pct`) that may not sum to 100 (e.g. `33.33 + 33.33 + 33.33 = 99.99`). Worth confirming the RPC uses the same arithmetic, or that the UI tolerates non-100 sums in stacked bar rendering.

### 3. Deployment race risk

The service will return 400 on all `sentiment-overview` requests if it deploys before the DB migration. The test plan calls this out, but it's worth confirming there's no window where this service can be live without the RPC existing — or adding a guard that produces a cleaner error than a raw Postgres function-not-found message.

---

The ordering fix (validating `siteId` before the RPC rather than after the data fetch) is a nice incidental improvement.

calvarezg:
## Updated Review

All three issues from the previous review are addressed:

**Issue 1 (multi-value filter truncation) — resolved.** The `brandPresenceRpcRejectsMultiCategoryOrRegion` guard is exactly the right fix — it's already used consistently across the other 5 RPC-backed handlers and the new tests confirm 400 is returned for both multi-category and multi-region inputs.

**Issue 2 (percentage rounding) — one remaining note.** The old JS aggregation guaranteed `positive + neutral + negative = 100` via `neutralPct = 100 - positivePct - negativePct`. The RPC now returns the three values independently, so if the DB computes them with independent `ROUND()` calls they could sum to 99 or 101. Not a blocker, but worth a quick check that the RPC uses the same arithmetic (i.e. derives neutral as `100 - positive_pct - negative_pct`).

**Issue 3 (deployment race) — documented in test plan.** Nothing more needed in the code.

One minor observation: there's no test for the mixed UUID+name category case (`categoryId: 'Acrobat'` + `categoryIds: 'some-uuid'`), which is the `hasMixedCategoryUuidAndName` branch. That case was tested in the old suite (`uses or() when UUID and name categories are combined`), but the new suite only covers multi-UUID and multi-region. Given that the guard function is shared and tested elsewhere this is fine — just noting the coverage gap.

Overall this is in good shape. LGTM pending the rounding check on the RPC side.

calvarezg (approved):

PR #2349, opened 2026-05-06
fix(prompts): prevent upsert from reactivating deleted prompts

byteclimber:
Deleted prompts were being returned by the existingQuery in upsertPrompts and then overwritten back to status='active' via the default in the patch object. Adding .neq('status', 'deleted') to the query makes deleted rows invisible to the upsert lookup, so they are never matched and never updated.

Fixes SITES-43723

- [x] make sure you update the API specification with the type, schema of the audit result and an example

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -509,7 +509,8 @@ export async function upsertPrompts({
     .from('prompts')
     .select('id,prompt_id,text,regions')
     .eq('organization_id', organizationId)
-    .eq('brand_id', brandUuid);
+    .eq('brand_id', brandUuid)
+    .neq('status', 'deleted');
 
   if (incomingIds.length > 0) {
     existingQuery = existingQuery.in('prompt_id', incomingIds);
--- a/test/controllers/brands.test.js
+++ b/test/controllers/brands.test.js
@@ -774,7 +774,13 @@ describe('Brands Controller', () => {
         if (table === 'prompts') {
           const insertChain = { select: () => thenable({ data: [{ prompt_id: 'new-1' }], error: null }) };
           return {
-            select: () => ({ eq: () => ({ eq: () => thenable({ data: [], error: null }) }) }),
+            select: () => ({
+              eq: () => ({
+                eq: () => ({
+                  neq: () => thenable({ data: [], error: null }),
+                }),
+              }),
+            }),
             insert: () => insertChain,
             update: () => ({ eq: () => thenable({ error: null }) }),
           };
--- a/test/support/prompts-storage.test.js
+++ b/test/support/prompts-storage.test.js
@@ -711,7 +711,13 @@ describe('prompts-storage', () => {
         from: (table) => {
           if (table === 'prompts') {
             return {
-              select: () => ({ eq: () => ({ eq: () => thenable({ data: [], error: null }) }) }),
+              select: () => ({
+                eq: () => ({
+                  eq: () => ({
+                    neq: () => thenable({ data: [], error: null }),
+                  }),
+                }),
+              }),
               insert: () => ({ select: () => thenable({ data: [{ prompt_id: 'new-1' }], error: null }) }),
               update: () => ({ eq: () => thenable({ error: null }) }),
             };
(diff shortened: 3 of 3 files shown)


martinst06 (approved):
LGTM

src/support/prompts-storage.js
```diff
@@ -509,7 +509,8 @@ export async function upsertPrompts({
     .from('prompts')
     .select('id,prompt_id,text,regions')
     .eq('organization_id', organizationId)
-    .eq('brand_id', brandUuid);
+    .eq('brand_id', brandUuid)
+    .neq('status', 'deleted');
```
rainer-friederich, line 513: `.neq('status', 'deleted')` is a denylist. If a non-active status is ever added (e.g. `archived`, `paused`), the upsert will silently match and overwrite those rows back to `active`. The safe approach is an allowlist: `.eq('status', 'active')`.

rainer-friederich, line 513: This filter makes deleted rows invisible to the lookup, so a resubmitted `prompt_id` falls into `toInsert`. But `UNIQUE (brand_id, prompt_id)` is unconditional on `status`, so the INSERT hits a constraint violation and throws a 500 — trading silent reactivation for a noisy hard error.

Worse, `toInsert` is sent as a single batch `.insert()` call, so one such collision aborts every other valid insert in the same request.

Two options to fix this properly:
- Keep the filter but catch the unique-violation and return a typed `skipped` result instead of throwing
- Remove the filter and explicitly gate the update path: only update `status → active` when the matched row's current status is not `deleted` (requires selecting `status` in the lookup columns too)

test/controllers/brands.test.js
```diff
@@ -774,7 +774,13 @@ describe('Brands Controller', () => {
         if (table === 'prompts') {
           const insertChain = { select: () => thenable({ data: [{ prompt_id: 'new-1' }], error: null }) };
           return {
-            select: () => ({ eq: () => ({ eq: () => thenable({ data: [], error: null }) }) }),
+            select: () => ({
+              eq: () => ({
+                eq: () => ({
+                  neq: () => thenable({ data: [], error: null }),
```
rainer-friederich, line 780: This mock was updated, but the `createPromptsByBrand returns 500 when storage throws` test elsewhere in this file was not. The real code now calls `.neq()` before awaiting, so that mock — which has no `.neq` — throws `TypeError: .neq is not a function`. The 500 assertion still passes but for the wrong reason; it's no longer testing the storage-rejection path.

test/support/prompts-storage.test.js
```diff
@@ -766,12 +774,54 @@ describe('prompts-storage', () => {
       expect(result.updated).to.equal(1);
     });
 
+    it('does not reactivate a deleted prompt matched by prompt_id', async () => {
```
rainer-friederich, line 777: Two gaps with this test:

The `.neq` filter is also applied to `existingByKey`, which maps by `(text, regions)`. If a caller submits a prompt with the same text/regions as a deleted row but a different `prompt_id`, the deleted row is invisible, the code inserts a new row, and two rows now exist for the same logical prompt — one deleted, one active. There's no test covering that path.

The insert mock returns `{ data: [], error: null }`, so this test only proves the mock chain works. In a real DB, that INSERT would fail with a unique-constraint violation because the deleted row still holds `(brand_id, prompt_id)`. Consider asserting either that the insert is not attempted, or that the error surface is handled gracefully.

rainer-friederich (commented):
@byteclimber — this PR is already merged, so flagging for a follow-up. See inline comments for the details; summary below.

The core issue: `.neq('status', 'deleted')` makes deleted rows invisible to the lookup, which routes a resubmitted `prompt_id` into `toInsert`. Because `UNIQUE (brand_id, prompt_id)` is unconditional on status in the DB, that INSERT hits a constraint violation — replacing silent reactivation with a hard 500 that also aborts all other valid inserts in the same batch call.

Additionally, the `upsertPrompts` JSDoc and OpenAPI spec don't document that deleted prompts can no longer be revived via upsert. Callers relying on re-upsert-to-revive will get an opaque 500 with no indication of what went wrong or what the alternative is (PATCH `updatePromptById` with `status: active`). Worth a follow-up to document the new contract.

PR #2352, opened 2026-05-06
fix(prompts): replace denylist filter with in-loop allowlist gate for deleted prompts

byteclimber:
## Summary

Follow-up to #2349, addressing review comments from @rainer-friederich.

**Problems with the previous fix (`.neq('status', 'deleted')`):**

- **Denylist instead of allowlist** — any future non-active status (`archived`, `paused`) would be silently overwritten back to `active` by the default `p.status || 'active'` patch
- **Constraint violation** — deleted rows became invisible to the lookup, so a resubmitted `prompt_id` fell into `toInsert` and hit `UNIQUE(brand_id, prompt_id)`, replacing silent reactivation with a hard 500 that also aborted every other valid insert in the same batch call
- **Broken `500 when storage throws` test** — the test mock had no `.neq()` handler, so calling `.neq()` threw `TypeError: .neq is not a function`; the 500 assertion passed for the wrong reason

**This fix:**

- Removes the status filter from `existingQuery` entirely; adds `status` to the selected columns instead
- Adds an explicit in-loop gate: `if (match && match.status !== 'active') { skipped++; continue }` — deleted (and any future non-active) rows are matched in the lookup (no constraint collision) but skipped without update or insert
- Returns a `skipped` count alongside `created` and `updated`
- Rewrites the regression test to exercise the actual skip path (deleted row returned by query, match found, skipped)
- Adds a second regression test for the `existingByKey` (text+regions) path
- Restores the `500 when storage throws` mock to test the right failure path

## Test plan

- [ ] `npx mocha test/support/prompts-storage.test.js` — 82 tests pass
- [ ] `npx mocha test/controllers/brands.test.js -g createPromptsByBrand` — all pass, including `returns 500 when storage throws`
- [ ] `npm test` — full suite green

### DIFF
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -507,10 +507,9 @@ export async function upsertPrompts({
 
   let existingQuery = postgrestClient
     .from('prompts')
-    .select('id,prompt_id,text,regions')
+    .select('id,prompt_id,text,regions,status')
     .eq('organization_id', organizationId)
-    .eq('brand_id', brandUuid)
-    .neq('status', 'deleted');
+    .eq('brand_id', brandUuid);
 
   if (incomingIds.length > 0) {
     existingQuery = existingQuery.in('prompt_id', incomingIds);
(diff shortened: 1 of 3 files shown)


rainer-friederich (approved):
LGTM

PR #2355, opened 2026-05-06
feat: add includedURLs support to patchAuditForSite endpoint

noruiz:
## Summary
Fixes LLMO-4070 (backend portion)

## What changed
- Extended `PATCH /sites/:siteId/:auditType` to accept `includedURLs` (array of URLs) with merge/replace semantics via `replaceIncludedURLs` boolean flag
- Added `replaceExcludedURLs` flag to give callers atomic control over `excludedURLs`
- Added `includedURLs` and both replace flags to the `UpdateHandlerTypeConfig` and `HandlerConfig` OpenAPI schemas
- 12 new test cases covering validation, clear/merge/replace modes, deduplication, combined requests, and the undefined-handlers edge case

## Why
The `config.handlers[auditType].includedURLs` field existed in the data model but had no API surface. The `Config` model (external package) has no `updateIncludedURLs` setter, so `includedURLs` is applied by patching the plain object returned by `Config.toDynamoItem()` directly before `site.setConfig()`.

## Assumptions made
- `Config.toDynamoItem()` returns `handlers` as the same object reference as `state.handlers`, making the configObj patch equivalent to a state mutation in the normal case
- In production, `state.handlers` is always initialized to `{}` by schema default — the undefined-handlers path is an edge case

## Testing
Unit tests cover: non-array input, invalid URL validation, empty-array clear, merge with no existing, merge with existing, deduplication, atomic replace, `replaceExcludedURLs: false` merge regression, combined includedURLs+excludedURLs request, undefined-handlers persistence, response body correctness

## Code review
This solution went through 3 automated subagent review iterations.

- Iteration 1 (sonnet): null guard, missing test for undefined handlers, format:uri in schema — resolved
- Iteration 2 (opus): silent write-loss root cause fix, false-positive tests — resolved
- Iteration 3 (sonnet): format:uri consistency, misleading test name, simplified response, combined-fields test — resolved


src/controllers/audits.js
```diff
@@ -211,11 +218,18 @@ function AuditsController(ctx) {
...
-    const configuration = await Configuration.findLatest();
-    const registeredAudits = configuration.getHandlers();
-    if (!registeredAudits[auditType]) {
-      return notFound(`The "${auditType}" is not present in the configuration. List of allowed audits:`
-        + ` ${Object.keys(registeredAudits).join(', ')}.`);
+    let configuration;
+    try {
+      configuration = await Configuration.findLatest();
+    } catch {
+      // Configuration unavailable — skip audit-type registration check
```
baranovskyalexandr: Bare catch {} silently swallows all errors. Intent is to handle S3 unavailability, but catches everything — auth failures, programming errors, network timeouts. Any failure bypasses audit-type validation silently. At minimum log the error, ideally narrow the catch to the known S3 error.

noruiz: fixed

baranovskyalexandr (approved):

PR #2356, opened 2026-05-06
fix: enable category filtering in filter-dimensions

aniham:
reported by david folk here: https://cq-dev.slack.com/archives/C091CRHN961/p1777479341680559

### DIFF
--- a/docs/openapi/api.yaml
+++ b/docs/openapi/api.yaml
@@ -498,6 +498,8 @@ paths:
     $ref: './llmo-api.yaml#/llmo-agentic-traffic-weeks'
   /sites/{siteId}/agentic-traffic/movers:
     $ref: './llmo-api.yaml#/llmo-agentic-traffic-movers'
+  /org/{spaceCatId}/brands/{brandId}/brand-presence/filter-dimensions:
+    $ref: './llmo-api.yaml#/brand-presence-filter-dimensions'
   /org/{spaceCatId}/brands/{brandId}/brand-presence/prompt-execution-status:
     $ref: './llmo-api.yaml#/brand-presence-prompt-execution-status'
   /org/{spaceCatId}/brands/{brandId}/brand-presence/url-inspector/stats:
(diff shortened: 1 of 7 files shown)


igor-grubic:
A few items worth addressing before merge:

**1. Missing UUID validation on `categoryId`**
`categoryId` is taken from query params and passed straight into `.eq('category_id', categoryId)` on both `topic_categories` and `prompts`. Per project convention, UUID inputs should be validated with `isValidUUIDV4()` — otherwise a malformed value reaches PostgREST and surfaces as a 500-ish error rather than a clean `400 badRequest`. Suggest validating in `parseFilterDimensionsSiteQuery` and returning `badRequest`, or silently treating an invalid value as no filter.

**2. `IN_FILTER_CHUNK_SIZE` not applied to `topicIdFilter`**
The new code passes `topicIdFilter` to `.in('id', topicIdFilter)` without chunking, while the existing prompt-count path carefully chunks brand IDs with `IN_FILTER_CHUNK_SIZE`. If a category maps to more topics than the underlying URL/IN limit, this query will fail. Either apply the same chunking pattern or add a comment justifying why it's safe (e.g., the `QUERY_LIMIT` cap on the prior `topic_categories` query keeps it bounded below the chunk size).

**3. Unrelated change in `llmo.test.js`**
The `this.timeout(10000)` added to the `before` hook in `test/controllers/llmo/llmo.test.js` looks like a flaky-test fix unrelated to category filtering. Worth splitting into its own commit/PR with a brief reason in the message.

**4. OpenAPI not updated**
The new `categoryId` query param is undocumented. The brand-presence `filter-dimensions` endpoint appears to have no OpenAPI entry at all today (pre-existing gap), but per the "Specification Sync" guidance in CLAUDE.md, this is a good moment to add one — at minimum documenting `categoryId` alongside `siteId`.

igor-grubic (approved):
Approved, but can we please merge on monday to avoid performance issues over the weekend

PR #2368, opened 2026-05-08
fix: ensure stale suggestion grants are cleaned up on token cycle transitions

sandsinh:
## Summary
- Removed an early-exit guard in `grantSuggestionsForOpportunity` that skipped grant cleanup when there were no new suggestions, causing stale grants from previous cycles to linger
- When starting a new token cycle (no active token found), the handler now looks up the previous cycle's token via `findLastCreatedBySiteIdAndTokenType` and runs `handleExistingTokenCycle` on it before creating the new token — ensuring each cycle begins with accurate grant state
- Added unit tests covering the no-new-suggestions edge case and the previous-token cleanup path

## Test plan
- [x] Run `npm test` — all unit tests pass
- [x] Verify that when a site has no new suggestions, stale grants from prior cycles are still revoked (not skipped)
- [x] Verify that when a new token cycle starts after an expired/missing active token, the previous token's grants are cleaned up before the new cycle runs
- [x] Verify that when no previous token exists at all, the new cycle starts cleanly without errors


tarunsinghdev (approved):

src/support/grant-suggestions-handler.js
```diff
@@ -345,20 +345,25 @@ export async function grantSuggestionsForOpportunity(dataAccess, site, opportuni
...
+  if (token) {
+    ({ token, didRevoke } = await handleExistingTokenCycle(collections, ids, token));
+  } else {
+    const prevToken = await Token.findLastCreatedBySiteIdAndTokenType(siteId, tokenType);
+    if (prevToken) {
+      await handleExistingTokenCycle(collections, ids, prevToken);
+    }
+    const grants = { grantedIds, grantIds, newSuggestions };
+    ({ token, didRevoke } = await handleNewTokenCycle(collections, ids, grants));
+  }
```
MysticatBot, line 389: issue (blocking): When `newSuggestionIds` is empty, this branch still calls `handleNewTokenCycle` which internally does `Token.findBySiteIdAndTokenType(siteId, tokenType, { createIfNotFound: true })` - creating a brand new token with nothing to grant. The test masks this by stubbing the returned token with `getRemaining: () => 0`, but in production `createIfNotFound` returns a fresh token with full capacity. This creates orphaned tokens on every invocation where no new suggestions exist.

Additionally, `splitSuggestionsByGrantStatus([])` at line 349-350 makes a DB call that will always return empty arrays when `newSuggestionIds` is empty.

Fix: re-introduce a scoped early return after the previous-token cleanup when there are no new suggestions:

```javascript
if (prevToken) {
  await handleExistingTokenCycle(collections, ids, prevToken);
}
if (!newSuggestionIds.length) {
  return; // cleanup done, nothing to grant
}
const grants = { grantedIds, grantIds, newSuggestions };
({ token, didRevoke } = await handleNewTokenCycle(collections, ids, grants));
```

This preserves the stale-grant cleanup fix while avoiding unnecessary DB calls and token churn.

src/support/grant-suggestions-handler.js
```diff
@@ -345,20 +345,25 @@ export async function grantSuggestionsForOpportunity(dataAccess, site, opportuni
...
+  let didRevoke;
+  if (token) {
+    ({ token, didRevoke } = await handleExistingTokenCycle(collections, ids, token));
+  } else {
+    const prevToken = await Token.findLastCreatedBySiteIdAndTokenType(siteId, tokenType);
+    if (prevToken) {
+      await handleExistingTokenCycle(collections, ids, prevToken);
+    }
+    const grants = { grantedIds, grantIds, newSuggestions };
+    ({ token, didRevoke } = await handleNewTokenCycle(collections, ids, grants));
```
MysticatBot, line 387: issue (blocking): If `handleExistingTokenCycle` throws on the previous token (e.g., a revocation fails on a corrupted grant), the error propagates and permanently blocks new token creation for this site. The next invocation hits the same path and fails again, since the stale state is unchanged.

Unlike the active-token path (where a throw is appropriate because the current cycle is broken), a failure cleaning up an *old* cycle should not prevent the new cycle from starting.

Fix: wrap in try/catch so cleanup failures degrade gracefully:

```javascript
if (prevToken) {
  try {
    // Side-effect only: revoke stale grants from previous cycle
    await handleExistingTokenCycle(collections, ids, prevToken);
  } catch (err) {
    log.warn(`Failed to clean up previous token ${prevToken.getId()}`, { error: err.message });
  }
}
```

MysticatBot (dismissed):
Hey @sandsinh,

**Verdict: Request changes** - two structural issues in the new else-branch need addressing before merge.
_Changes: Removes early-exit guard and adds previous-token cleanup to ensure stale suggestion grants are revoked during token cycle transitions (2 files)._

## Must fix before merge
1. [Important] Unnecessary `splitSuggestionsByGrantStatus([])` call and orphaned token creation when no new suggestions exist - `src/support/grant-suggestions-handler.js:367` (details inline)
2. [Important] `handleExistingTokenCycle(prevToken)` failure blocks new token creation permanently for the site - `src/support/grant-suggestions-handler.js:365` (details inline)

<details>

- nit: Discarded return value of `handleExistingTokenCycle(prevToken)` should have a one-line comment explaining it is side-effect-only - `src/support/grant-suggestions-handler.js:365`
- suggestion: Initialize `let didRevoke = false` instead of bare `let didRevoke` to avoid implicit falsy-undefined contract with `fillRemainingCapacity` - `src/support/grant-suggestions-handler.js:358`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 19s | <b>Cost:</b> $4.43 | <b>Commit:</b> <code>8325b861b420eb4ed0fc18aca2e20d8845097319</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2377, opened 2026-05-08
docs: ADR-002 preflight API REST redesign

GeezerPelletier:
## Summary

- Adds `docs/decisions/002-preflight-api-rest-redesign.md` capturing the decision to replace `/preflight/beta/jobs` with site-scoped REST endpoints under `/sites/:siteId/preflights`, and to deprecate both legacy endpoint pairs
- No code changes - this is a decision record for team review before implementation begins

Related: SITES-44290 | Bulk endpoint deferred to: SITES-44432

### POST /sites/:siteId/preflights
**Request body:**
```json
{
  "url": "https://main--site--org.hlx.page/some-path"
}
```

`promiseToken` passed via cookie for authenticated CMS pages. `createdBy` is derived server-side from the caller's IMS profile and never supplied by the client. `step` is removed - Mysticat always runs the full identify/suggest flow. The submitted `url` must match one of the site's known hostnames (base URL, preview URL, or live URL) - a mismatch returns `PREFLIGHT_INVALID_REQUEST`.

**Response `202 Accepted`:**

Headers:
```
Location: https://spacecat.experiencecloud.live/api/v1/sites/{siteId}/preflights/{preflightId}
```

Body:
```json
{
  "preflightId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "status": "IN_PROGRESS",
  "createdAt": "2026-05-11T10:00:00.000Z",
  "createdBy": {
    "id": "ABC123@techacct.adobe.com",
    "displayName": "John Doe"
  }
}
```

**Error responses:**

| Status | `errorCode` | Condition |
|--------|-------------|-----------|
| `400 Bad Request` | `PREFLIGHT_INVALID_REQUEST` | `url` is missing, not a valid URI, or does not belong to the site identified by `:siteId` |
| `403 Forbidden` | `PREFLIGHT_ACCESS_DENIED` | Caller does not have access to the site |
| `403 Forbidden` | `PREFLIGHT_NOT_ENABLED` | Preflight is not enabled for the site |
| `404 Not Found` | `PREFLIGHT_SITE_NOT_FOUND` | `siteId` does not exist |
| `502 Bad Gateway` | `PREFLIGHT_UPSTREAM_ERROR` | Mysticat returned a 5xx response |
| `500 Internal Server Error` | `PREFLIGHT_INTERNAL_ERROR` | Unexpected error within this service |

Error response body:
```json
{
  "errorCode": "PREFLIGHT_NOT_ENABLED",
  "message": "Preflight is not enabled for this site"
}
```

No job record is created for `400`, `403`, or `404` responses.

### GET /sites/:siteId/preflights
**Response `200 OK`** - grouped by URL, capped at 50 most recent preflights per site (sorted by `createdAt` desc across all URLs before grouping):
```json
[
  {
    "url": "https://main--site--org.hlx.page/some-path",
    "preflights": [
      {
        "preflightId": "3fa85f64-...",
        "status": "COMPLETED",
        "createdAt": "2026-05-11T10:00:00.000Z",
        "updatedAt": "2026-05-11T10:00:05.000Z",
        "createdBy": { "id": "ABC123@techacct.adobe.com", "displayName": "John Doe" }
      }
    ]
  }
]
```

### GET /sites/:siteId/preflights/:preflightId
**Response `200 OK`** - full detail. The handler verifies the loaded job's `siteId` matches the path's `:siteId`, returning `404` on mismatch:
```json
{
  "preflightId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "status": "COMPLETED",
  "url": "https://main--site--org.hlx.page/some-path",
  "createdAt": "2026-05-11T10:00:00.000Z",
  "createdBy": { "id": "ABC123@techacct.adobe.com", "displayName": "John Doe" },
  "updatedAt": "2026-05-11T10:00:05.000Z",
  "startedAt": "2026-05-11T10:00:01.000Z",
  "endedAt": "2026-05-11T10:00:05.000Z",
  "result": {},
  "resultType": "INLINE",
  "resultLocation": null,
  "error": null
}
```

When `resultType` is `S3` or `URL`, `result` will be `null` and the client should fetch from `resultLocation`.

## Data model

`AsyncJob` will be extended in `spacecat-shared-data-access` with:
- `siteId` as an **optional** top-level indexed attribute
- `jobType` as an **optional** top-level indexed attribute
- `allBySiteIdAndJobType(siteId, jobType)` collection method

Both attributes are optional so that existing job creation paths (including the deprecated `/preflight/jobs` queue-based flow) continue to work unchanged. `url` grouping on the list endpoint is an in-memory `groupBy` on `metadata.url` after the indexed query - acceptable given the 50-record cap.

`createdBy` is stored as `{ id, displayName }` in job metadata at creation time - `id` is `profile.email` (the IMS user ID) and `displayName` is composed from `profile.first_name + last_name`. Both fields are available on the authenticated profile; no additional IMS lookup required. Surfaces in all three endpoint responses for audit purposes.

This is a prerequisite: the `spacecat-shared-data-access` change must land before controller work begins.

## Current endpoint fate

Both legacy endpoint pairs are deprecated and remain functional in parallel with the new endpoints. The Sunset date will be set by PM at the time this ADR moves to Accepted, with a minimum of 90 days from MFE migration start.

| Endpoint | Action |
|----------|--------|
| POST /preflight/beta/jobs | **Deprecated** - internal use only; no external consumers |
| GET /preflight/beta/jobs/:jobId | **Deprecated** - internal use only; no external consumers |
| POST /preflight/jobs | **Deprecated** - queue-based; migration timeline to be coordinated with consumers |
| GET /preflight/jobs/:jobId | **Deprecated** - queue-based; migration timeline to be coordinated with consumers |

## Test plan
- [ ] Read the ADR and validate the design rationale
- [ ] Confirm the proposed endpoint shape matches team expectations before implementation starts

?? Generated with [Claude Code](https://claude.com/claude-code)


docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,271 @@
...
+
+Headers:
+```
+Location: https://spacecat.experiencecloud.live/api/v1/sites/{siteId}/preflights/{preflightId}
+```
+
+Body:
+```json
+{
+  "preflightId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
```
bhellema, line 88: should this not just be `id` ?  Is this not the id of the aysnc job?

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,271 @@
...
+  "status": "IN_PROGRESS",
+  "createdAt": "2026-05-11T10:00:00.000Z",
```
bhellema, line 90: Should we include last updatedAt as well.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,271 @@
...
+| GET /preflight/beta/jobs/:jobId | **Deprecated** — internal use only; no external consumers |
+| POST /preflight/jobs | **Deprecated** — queue-based; migration timeline to be coordinated with consumers |
+| GET /preflight/jobs/:jobId | **Deprecated** — queue-based; migration timeline to be coordinated with consumers |
+
+### POST /sites/:siteId/preflights
+
+**Request body** (`application/json`):
+```json
+{
+  "url": "https://main--site--org.hlx.page/some-path",
```
bhellema: Today you can pass a list of urls.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,271 @@
...
+  "step": "identify",
```
bhellema: The concept of step should be going away as we always will perform identify/suggest as part of the mysticat agent step.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,271 @@
...
+  "mystiqueUrl": "optional-ephemeral-host.stage.cloud.adobe.io"
```
bhellema: I don't see the value of sending this back to the client.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,271 @@
...
+   for the caller not being required to identify the site upfront. By the time a user triggers
+   a preflight from the MFE, the `siteId` is already in the application state — the MFE has
+   navigated to a specific site in Sites Optimizer. Making the caller supply it explicitly is
+   more honest and removes a resolution step that can fail.
+
+3. **`jobId` leaks the async-job abstraction.** The returned identifier is named `jobId`,
+   exposing the underlying `AsyncJob` model to consumers. A preflight is a domain concept;
+   its identifier should reflect that.
+
+4. **No list endpoint.** There is no way to retrieve the history of preflights for a site.
```
bhellema, line 33: Progressive thinking, I like it.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,271 @@
...
+5. **Beta qualifier in the path.** `/preflight/beta/jobs` signals immaturity via the URL,
+   which is an anti-pattern. Versioning and feature graduation should not be expressed in
+   path segments.
+
+## Decision
+Replace both endpoint pairs with three site-scoped REST endpoints:
+
+| Method | Path | Description |
+|--------|------|-------------|
+| POST | /sites/:siteId/preflights | Creates a new preflight job for a single URL |
```
bhellema, line 40: One preflight job can have multiple urls associated with the payload.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,271 @@
...
+{
+  "url": "https://main--site--org.hlx.page/some-path",
+  "step": "identify",
+  "mystiqueUrl": "optional-ephemeral-host.stage.cloud.adobe.io"
+}
+```
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `url` | string (URI) | Yes | The single page URL to analyze |
```
bhellema: Today's API supports plural.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,271 @@
...
+| `step` | enum: `identify` \| `suggest` | Yes | Audit step to perform |
```
bhellema: Step should go away.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,271 @@
...
+| `mystiqueUrl` | string | No | Dev-only override for the Mysticat service URL |
```
bhellema: url i don't think needs to be exposed here.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,271 @@
...
+
+`promiseToken` is passed via cookie for authenticated CMS pages (CS/CS_CW/AMS sites); it is not part of the request body.
```
bhellema, line 74: There is a fix that should be set via the header not cookie.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,271 @@
...
+
+`createdBy` is derived server-side from the caller's IMS profile (`authInfo.getProfile().email`) and is never supplied by the client.
+
```
bhellema, line 77: createdAt - would be good to report too

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,271 @@
...
+| Status | Condition |
+|--------|-----------|
+| `400 Bad Request` | `url` or `step` is missing or invalid |
+| `403 Forbidden` | Caller does not have access to the site, or preflight is not enabled / no preflight audits are enabled for the site |
+| `404 Not Found` | `siteId` does not exist |
+| `500 Internal Server Error` | Mysticat call failed or unexpected error |
+
+No job record is created for `400`, `403`, or `404` responses. The current `/preflight/beta/jobs`
+behavior of creating a job and immediately setting it to `CANCELLED` when preflight is disabled
```
bhellema, line 120: Cancelled is not the correct status here it feels like cancelled is a user action.  Maybe something like ABORTED.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,271 @@
...
+      },
+      {
+        "preflightId": "7c9b1e32-1234-4abc-b3fc-9f8a7c6d5e4f",
+        "status": "COMPLETED",
+        "step": "suggest",
+        "createdAt": "2026-05-11T10:05:00.000Z",
+        "createdBy": "user@example.com"
+      }
+    ]
+  },
```
bhellema, line 151: This structure is not what I would expect, but rather a list of jobs with some minor details:
```
[
  {
      "preflightId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
      "status": "COMPLETED",
      "urlCount": 12,   <----- maybe show the number of urls they ran with
      "opportunities": 5,  <------ the number of opportunities that were created
      "createdAt": "2026-05-11T10:00:00.000Z",
      "createdBy": "user@example.com"
  }
]
```

Listing the preflight jobs should be lightweight and fast.  Details of the jobs can be fetched with a follow up with the id of the preflight job.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: The rename is intentional — `jobId` leaks the `AsyncJob` backing model to consumers, which is an implementation detail they shouldn't need to know about. `preflightId` is the domain name for this resource. Internally it maps 1:1 to the AsyncJob ID.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Good call — `updatedAt` is already on the GET by ID response. Worth adding to the list items too so clients can detect state changes without fetching each one individually.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Yes, and that's an intentional change — one URL per request, one `preflightId` per URL. Each preflight is an independently pollable resource. If the MFE needs bulk in future, clients fire parallel requests. This avoids partial-failure complexity and keeps each job clean. See the key changes section for the full rationale.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Good point and worth nailing down now. If the Mysticat agent always runs both identify and suggest as a single flow, `step` is redundant and we should drop it from the new endpoint entirely rather than carry it forward and deprecate it later. Can you confirm that's the direction? If so we'll remove it from the design.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: To clarify — `mystiqueUrl` is a request-only field, never returned in any response. It lets developers route a preflight to an ephemeral Mysticat instance for testing without hitting prod, and is blocked in non-dev environments. If you're questioning whether it belongs in the API at all that's a fair debate — it could move to an env-level config instead.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Thanks!

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Intentional change — this PR proposes one URL per preflight, one `preflightId` per URL. This connects to your comment on the list structure below: if a job is always one URL, the list response is naturally flat rather than grouped.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Yes — the change from `urls` (array) to `url` (single) is intentional. One URL per request keeps each preflight independently pollable and avoids partial-failure complexity for bulk scenarios.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Aligned — flagging this as a discussion point along with your comment on line 64. If the Mysticat agent always runs both steps, `step` should be dropped from the new endpoint design entirely.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: The `url` field is required in the request body because Mysticat needs it to scrape the specific page via DRS headless browser — `siteId` alone isn't enough since a site has many pages and preflight is per-page. As for the response, it is useful in GET by ID so the consumer knows what was analyzed. Could reasonably be omitted from the 202 POST response body given the `Location` header already points to the resource.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Agreed a header would be cleaner. The cookie approach is inherited from the `/auth/promise` browser flow where it's set automatically. Switching to a header would require the MFE to read and forward it explicitly — worth a separate cleanup ticket but out of scope for this redesign.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: `createdAt` is already in the 202 response body — you may have been looking at an earlier version of the doc before we expanded the response shapes.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Good point on the semantics, but we've resolved this differently — the new design returns `403 Forbidden` immediately when preflight is disabled, without creating a job record at all. No phantom CANCELLED or ABORTED record is created. See the error responses table in the POST section.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: The flat list is simpler and we can drop the URL grouping — agreed. The `opportunities` count is an interesting addition; does Mysticat write that back to the job result today? If so it's worth surfacing in the list.

One question on `urlCount`: that field implies a job can span multiple URLs, but the new design is one URL per preflight. With that model `urlCount` would always be 1. Was the intent to keep multi-URL jobs, or is `urlCount` a carry-over from the old design?

docs/decisions/002-preflight-api-rest-redesign.md
bhellema: @GeezerPelletier @clotton @johobot:

1. what is the implication here for running many urls across a site in one job in the mysticat world?
2. clients may not like to execute 100 preflights jobs for 100 urls.  But 1 job for 100 urls. We have to look at this use case and make sure it's supported as our current API /preflight/jobs supports this.

docs/decisions/002-preflight-api-rest-redesign.md
bhellema: Client's could select hundreds of pages. This is too much load to manage.   Use case, customer is ready to run a promotion and they have many different areas of the site they want to run preflight on. Allowing them to send 1 job with many urls makes sense.  It could take a while, but running in bulk this way would be expected to be slower.

docs/decisions/002-preflight-api-rest-redesign.md
bhellema: This is the direction that we want to go.

docs/decisions/002-preflight-api-rest-redesign.md
bhellema: Let's remove it from the API response.

docs/decisions/002-preflight-api-rest-redesign.md
bhellema: I do not think this should be exposed as part of the public api.

docs/decisions/002-preflight-api-rest-redesign.md
bhellema: Sure, preflightId is fine, I figured that id would be enough. The response is a preflight response and here's the id.

docs/decisions/002-preflight-api-rest-redesign.md
bhellema: Nice! I like it.  However as part of the 403, what does the response body look like to allow the UI to provide a meaningful message to the user as if they do not have access to the site, or that the audit was disabled.

docs/decisions/002-preflight-api-rest-redesign.md
bhellema: I think we must consider the ability for multiple urls.  If we get to the point of providing a different experience where they can select many pages and initiate a preflight request, we need a way to make this experience simply for the client, being a headless agent or client ui.

bhellema (commented):
We really need to consider what the bulk use case looks like.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Valid use case, but to my knowledge multi-URL was never actually leveraged from the MFE in practice — every real call was single-URL. So I'd rather not design for bulk now when it isn't fully hashed out.

When the use case is confirmed, adding a dedicated `POST /sites/:siteId/preflights/bulk` endpoint is cheap — the controller fans out to one AsyncJob per URL, Mysticat is called once per URL, and each result is independently pollable. No single point of failure, no blocked job waiting on N URLs. That's the right time to design it properly.

For now, keeping it single-URL keeps the design clean and unambiguous. That said, it's your call — we could overload the single endpoint to accept `urls` as today, but that means reverting the `Location` header in favour of a response body with multiple IDs, and tying multiple URLs to a single job would not be my preference.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Same response as above — see the thread on line 44 for the full proposal.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Apologies for the misread — we interpreted this as being about the `url` field rather than `mystiqueUrl`. Agreed, `mystiqueUrl` is removed from the public documentation. It stays in the implementation as an undocumented dev override for testing against ephemeral Mysticat instances, just not part of the public API contract.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Good call. The error body includes an `errorCode` for machine-readable handling and a `message` as a human-readable hint:

```json
{
  "errorCode": "PREFLIGHT_NOT_ENABLED",
  "message": "Preflight is not enabled for this site"
}
```

All codes are `PREFLIGHT_`-prefixed for consistency: `PREFLIGHT_ACCESS_DENIED`, `PREFLIGHT_NOT_ENABLED`, `PREFLIGHT_INVALID_REQUEST`, `PREFLIGHT_SITE_NOT_FOUND`, `PREFLIGHT_INTERNAL_ERROR`. Consumers translate `errorCode` to their own messaging — `message` is informational only. ADR updated.

Open to other suggestions on the error code style if there's a preferred convention.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Addressed in the thread on line 44 — the proposal is to keep this endpoint single-URL for now since it was never actually used with multiple URLs in practice, and add a dedicated `POST /sites/:siteId/preflights/bulk` endpoint when that use case is confirmed. Adding it later is cheap and keeps the design clean in the meantime.

docs/decisions/002-preflight-api-rest-redesign.md
bhellema: I'm happy that at least there's an error code!  Using a numeric value just requires another lookup, I don't mind the string.

GeezerPelletier: Agreed — string codes are self-documenting and don't require a lookup table. Happy we're aligned.

GeezerPelletier:
@bhellema — created SITES-44432 to track the bulk endpoint use case (`POST /sites/:siteId/preflights/bulk`). It's linked to SITES-44290 as a follow-on. We can design it properly once the single-URL endpoint ships and the bulk use case is confirmed from the MFE side.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,276 @@
...
+}
+```
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `url` | string (URI) | Yes | The single page URL to analyze |
+
+`promiseToken` is passed via cookie for authenticated CMS pages (CS/CS_CW/AMS sites); it is not part of the request body.
+
+`createdBy` is derived server-side from the caller's IMS profile (`authInfo.getProfile().email`) and is never supplied by the client.
```
johobot: Worth noting: in the current codebase `profile.email` holds the IMS user ID string, not necessarily a human-readable email address (see comment in `api-key.js`). The pattern used by other controllers is `authInfo.getProfile().email || 'system'`. The rendered `createdBy` value in the MFE may look like an opaque ID rather than an address — worth aligning on so implementers and the MFE team have the same expectation.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,276 @@
...
+  "createdAt": "2026-05-11T10:00:00.000Z",
+  "createdBy": "user@example.com"
+}
+```
+
+**Error responses:**
+
+| Status | `errorCode` | Condition |
+|--------|-------------|-----------|
+| `400 Bad Request` | `PREFLIGHT_INVALID_REQUEST` | `url` is missing or invalid |
```
johobot: Two gaps worth closing here:

1. "Invalid" is not defined. Does the URL need to match a known Helix/AEM hostname pattern? The current implementation validates against site preview/live/CDN URLs via `findByPreviewURL`. Is that check being preserved, or is URL-to-site resolution truly gone?

2. If it's gone, there's nothing preventing a caller from submitting a `url` that belongs to a different site than `:siteId`. That mismatches context for link normalization and auth token resolution in the audit worker, which uses `site.getBaseURL()`. Recommend keeping hostname validation and returning a distinct error code if the URL doesn't match the site.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,276 @@
...
+| `403 Forbidden` | `PREFLIGHT_ACCESS_DENIED` | Caller does not have access to the site |
+| `403 Forbidden` | `PREFLIGHT_NOT_ENABLED` | Preflight is not enabled for the site |
+| `404 Not Found` | `PREFLIGHT_SITE_NOT_FOUND` | `siteId` does not exist |
+| `500 Internal Server Error` | `PREFLIGHT_INTERNAL_ERROR` | Mysticat call failed or unexpected error |
```
johobot: When Mysticat itself returns 5xx, `502 Bad Gateway` (or `503`) is more accurate than `500` — it signals the fault is downstream, not in this service. That distinction matters for client retry logic and on-call triage.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,276 @@
...
+
+No job record is created for `400`, `403`, or `404` responses. The current `/preflight/beta/jobs`
+behavior of creating a job and immediately setting it to `CANCELLED` when preflight is disabled
+is not carried forward — a `403` is returned immediately, keeping the job store clean.
+
+---
+
+### GET /sites/:siteId/preflights
+
+**Response** `200 OK` — grouped by URL, with a nested array of preflights per page. A site can have preflights for multiple URLs:
```
johobot: Two concerns with this shape:

1. Every other list endpoint in the service returns a flat array — this URL-grouped structure is a one-off that requires in-memory `groupBy` in the controller and can't go through the standard DTO path.
2. No pagination is specified. An active site accumulates 300+ records per 7-day TTL window. The service uses cursor-based `limit`/`cursor` pagination elsewhere. Even a default cap needs to be defined here, since adding `nextCursor` to a bare top-level array later is a breaking change.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,276 @@
...
+| `url` | string | The page URL that was analyzed |
+| `preflights` | array | Preflights run against this URL |
+| `preflights[].preflightId` | UUID | Unique identifier for the preflight |
+| `preflights[].status` | enum: `IN_PROGRESS` \| `COMPLETED` \| `FAILED` \| `CANCELLED` | Current job status |
+| `preflights[].createdAt` | ISO 8601 | When the preflight was created |
+| `preflights[].updatedAt` | ISO 8601 | When the preflight was last updated |
+| `preflights[].createdBy` | string | IMS email of the user who triggered the preflight |
+
+---
+
```
johobot, line 175: The ADR doesn't specify that this handler validates `preflightId` belongs to the path's `:siteId`. The current `getPreflightJobStatusAndResult` loads by `preflightId` alone with no site ownership check. If that carries forward, a caller can read any preflight result — including scraped draft CMS content — by substituting a different `siteId` in the path. The handler should load by `(siteId, preflightId)` composite and return `404` if either mismatches, never by ID alone.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,276 @@
...
+```json
+{
+  "preflightId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
+  "status": "COMPLETED",
+  "url": "https://main--site--org.hlx.page/some-path",
+  "createdAt": "2026-05-11T10:00:00.000Z",
+  "createdBy": "user@example.com",
+  "updatedAt": "2026-05-11T10:00:05.000Z",
+  "startedAt": "2026-05-11T10:00:01.000Z",
+  "endedAt": "2026-05-11T10:00:05.000Z",
```
johobot, line 191: `AsyncJob` already has `resultLocation` (S3 URI) and `resultType` (INLINE/S3/URL) for handling large payloads. In practice, preflight results can be substantial and are often written to S3, leaving `result: null` in the DB record. Clients polling this endpoint need to know when to follow `resultLocation` instead. The response table should document both fields and the decision rule.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,276 @@
...
+
+- **`siteId` moves to the path.** URL-to-site resolution logic is removed from the controller.
+- **One URL per request.** If the MFE supports bulk analysis in the future, the client issues
+  one `POST /sites/:siteId/preflights` per URL. Each gets its own independent `preflightId`,
+  status, result, and lifecycle. There is no batch endpoint — this keeps each preflight as a
+  clean, independently pollable resource and avoids partial-failure complexity.
+- **`jobId` renamed to `preflightId`** in all responses.
+- **`pollUrl` removed from the response body.** Per RFC 7231 §6.3.3, the URL of the created
+  resource is communicated via the `Location` response header. Clients that need to poll for
+  completion read `Location` rather than a body field.
```
johobot, line 226: The worker still branches on `step` in two places:
- `src/preflight/links.js:102` — `if (step === 'suggest')` gates the AI suggestion call
- `src/preflight/metatags.js:103` — `if (step === 'suggest')` gates `metatagsAutoSuggest`

Dropping `step` from the new endpoint without a coordinated worker change means callers silently get identify-only results (the worker defaults to `PREFLIGHT_STEP_IDENTIFY` when the field is absent). If Mysticat has already been updated to always run both phases unconditionally, this ADR should reference that change. If not, it's a prerequisite alongside the `spacecat-shared-data-access` change.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,276 @@
...
+- **No phantom jobs for rejected requests.** If the site is not found, the caller lacks access,
+  or preflight is not enabled for the site, the endpoint returns an error immediately without
+  creating a job record. The previous behavior of creating a `CANCELLED` job in these cases
+  is removed.
+- **No `organizationId` in the path.** `siteId` is a globally unique UUID, consistent with
+  all other site-scoped resources in this service.
+
+**Both existing endpoint pairs are deprecated**, not removed. They will remain functional in
+parallel with the new endpoints until the MFE has migrated and a deletion milestone is agreed
+upon. Deprecation notices should be added to their OpenAPI spec entries and response headers
```
johobot: `Sunset: <date-tbd>` doesn't satisfy RFC 8594, which requires a concrete date. In practice this tends to become permanent dual maintenance. Recommend committing to a process here: e.g. "date set by PM at the time this ADR moves to Accepted, minimum 90 days from MFE migration start."

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,276 @@
...
+
+## Data Model: Extending AsyncJob for Efficient siteId Queries
+
+The current implementation backs preflights with the generic `AsyncJob` model. `AsyncJob` has
+no top-level `siteId` field — `siteId` is buried inside the `metadata` JSON blob with no
+index. The `AsyncJobCollection` exposes only `findById`; there is no `allBySiteId` method.
+
+The decision is to **extend `AsyncJob` in `spacecat-shared-data-access`**:
+
+- Add `siteId` as an **optional** top-level indexed attribute on the `AsyncJob` schema
```
johobot: `jobType` is currently stored inside `metadata.jobType` (e.g. `metadata: { jobType: 'preflight' }`), not as a top-level attribute. `allBySiteIdAndJobType` would therefore need to either (a) promote `jobType` to a top-level indexed column — another schema change and DB migration — or (b) fetch all jobs by `siteId` index and filter in-memory on `metadata.jobType`. Option (b) is fine at low volume but the choice should be explicit. The same applies to `url`, which also lives in `metadata` and is used for grouping on the list endpoint.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Good catch on the IMS profile ambiguity. We've expanded `createdBy` from a plain string to an object:

```json
"createdBy": {
  "id": "ABC123@techacct.adobe.com",
  "displayName": "John Doe"
}
```

`id` is `profile.email` (the IMS user ID — the naming is misleading but consistent with the rest of the codebase, as noted in `api-key.js`). `displayName` is composed from `profile.first_name + last_name`, falling back to `profile.name`. No extra IMS lookup needed — both fields are on the authenticated profile already. This shape is easy to extend with additional identity fields in future without breaking the API contract. ADR updated.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Both points addressed. The controller will validate that the `url` hostname matches one of the site's known hostnames (base URL, preview URL, or live URL) after loading the site via `Site.findById(siteId)`. A URL that is structurally valid but belongs to a different site returns `PREFLIGHT_INVALID_REQUEST` — so the cross-site submission risk is closed. The error table and field description in the ADR are updated to make "invalid" explicit. This replaces the implicit validation that `findByPreviewURL` previously provided as a side effect.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Good catch. Splitting into `502 Bad Gateway` / `PREFLIGHT_UPSTREAM_ERROR` for Mysticat failures and `500 Internal Server Error` / `PREFLIGHT_INTERNAL_ERROR` for unexpected local errors. Updated the ADR.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: On (1): agreed it's a one-off shape, but the MFE benefit of seeing history grouped by URL outweighs the controller complexity for now - the `groupBy` is a straightforward in-memory step after a single `allBySiteIdAndJobType` query.

On (2): adding a default cap of 50 most recent preflights per site (sorted by `createdAt` desc across all URLs before grouping). Given the 7-day `AsyncJob` TTL the natural upper bound is low in practice. Full cursor-based pagination is deferred - updated the ADR.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Good security catch. Adding an explicit ownership check to the ADR: the handler loads by `preflightId` then verifies the stored `siteId` matches the path's `:siteId`, returning `404` on mismatch - same response as a non-existent `preflightId` - so callers cannot confirm a preflight exists by probing with a different site path. Updated the ADR.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Good point - adding `resultType` and `resultLocation` to the GET by ID response table with the decision rule: when `resultType` is `S3` or `URL`, `result` will be `null` and the client should fetch from `resultLocation`; when `resultType` is `INLINE`, `result` contains the payload directly and `resultLocation` is `null`. Updated the ADR.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Confirmed - Mysticat has no concept of `step`; it always runs both identify and suggest in a single flow. The branching in `links.js` and `metatags.js` is dead code that will be removed as part of this implementation. No separate prerequisite needed. Updated the ADR.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Agreed - adopting your suggested process: the Sunset date will be set by PM at the time this ADR moves to Accepted, with a minimum of 90 days from MFE migration start. Updated the ADR.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Good clarification. The ADR already proposes promoting `siteId` to a top-level indexed column - extending that to also promote `jobType` as a top-level indexed attribute in the same `spacecat-shared-data-access` change is the right call, so `allBySiteIdAndJobType` queries both indexes without touching `metadata`. The `url` grouping for the list endpoint remains an in-memory `groupBy` on `metadata.url` after the indexed query - acceptable given the 50-record cap. Updated the ADR.

GeezerPelletier:
@bhellema @johobot - all review comments have been addressed. Would appreciate a pass for approval when you get a chance. Thanks!

GeezerPelletier:
Gentle reminder @bhellema @johobot @clotton — would appreciate a look when you get a chance. All previous review comments have been addressed. 🙏

clotton (approved):
Approving — Johnson's nine review points were all addressed cleanly in the 2026-05-12 push, and on my read the response shapes, validation, error model (502/PREFLIGHT_UPSTREAM_ERROR vs 500/PREFLIGHT_INTERNAL_ERROR), cross-site ownership check, and `resultType`/`resultLocation` decision rule are all in good shape. Nice work.

One non-blocking flag: the bulk-submission thread (line 44) ends with a deferral to SITES-44432 for a dedicated `POST /sites/:siteId/preflights/bulk` endpoint, but @bhellema hasn't explicitly signed off on that deferral yet. Worth getting a 👍 from Ben on the line 44 thread before merge so we don't ship with that thread effectively unresolved.

(Header-vs-cookie auth on line 76 is a known follow-up; happy to take that in a separate ticket.)

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,308 @@
...
+
+No job record is created for `400`, `403`, or `404` responses. The current `/preflight/beta/jobs`
+behavior of creating a job and immediately setting it to `CANCELLED` when preflight is disabled
+is not carried forward — a `403` is returned immediately, keeping the job store clean.
+
+---
+
+### GET /sites/:siteId/preflights
+
+**Response** `200 OK` — grouped by URL, with a nested array of preflights per page. A site can have preflights for multiple URLs. Results are capped at the 50 most recent preflights per site (sorted by `createdAt` descending across all URLs before grouping). Given the 7-day TTL on `AsyncJob` records the natural upper bound is low; full cursor-based pagination is deferred until there is evidence of consumers hitting the cap.
```
bhellema: I think capping at 50 is interesting, but the payload is so small that I don't see why we need to.  I think the limiting factor here is the Lambda response size which is 6 MB for both request and response.  So, there's lots of space for a lot of preflights.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,308 @@
...
+
+`errorCode` gives consumers a stable machine-readable contract; `message` is a human-readable hint and should not be parsed by clients.
+
+No job record is created for `400`, `403`, or `404` responses. The current `/preflight/beta/jobs`
+behavior of creating a job and immediately setting it to `CANCELLED` when preflight is disabled
+is not carried forward — a `403` is returned immediately, keeping the job store clean.
+
+---
+
+### GET /sites/:siteId/preflights
```
bhellema: Today the async job table contains the association between a job and a site via the metadata:

```
const job = await dataAccess.AsyncJob.create({
        status: 'IN_PROGRESS',
        metadata: {
          payload: {
            siteId: site.getId(), <-------------------------------
            urls: data.urls,      
           ...
          },
          jobType: 'preflight',
         ...
        },
      });
```

The assumption then is that there is a query that can go against jobType == preflight && metadata.payload.siteId.  We will need to update spacecat-shared-data-access to support something like this:

spacecat-shared/packages/spacecat-shared-data-access/src/models/async-job/async-job.collection.js
```
  async allBySiteIdAndJobType(siteId, jobType) {
    const { data, error } = await this.postgrestService
      .from(this.tableName)
      .select('*')
      .contains('metadata', { jobType, payload: { siteId } });

    if (error) {
      throw new Error(`Failed to query async jobs: ${error.message}`);
    }

    return (data || []).map((row) => this.createInstanceFromRow(row));
  }
```

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,308 @@
...
+[
+  {
+    "url": "https://main--site--org.hlx.page/some-path",
+    "preflights": [
+      {
+        "preflightId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
+        "status": "COMPLETED",
+        "createdAt": "2026-05-11T10:00:00.000Z",
+        "updatedAt": "2026-05-11T10:00:05.000Z",
+        "createdBy": { "id": "ABC123@techacct.adobe.com", "displayName": "John Doe" }
```
bhellema: This information is not available today.  This would have to be stored as part of the async job metadata.
```
const profile = context.attributes.authInfo.getProfile();
profile.first_name;
proflile.last_name;
profile.email;
```

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,308 @@
...
+|-------|------|-------------|
+| `url` | string | The page URL that was analyzed |
+| `preflights` | array | Preflights run against this URL |
+| `preflights[].preflightId` | UUID | Unique identifier for the preflight |
+| `preflights[].status` | enum: `IN_PROGRESS` \| `COMPLETED` \| `FAILED` \| `CANCELLED` | Current job status |
+| `preflights[].createdAt` | ISO 8601 | When the preflight was created |
+| `preflights[].updatedAt` | ISO 8601 | When the preflight was last updated |
+| `preflights[].createdBy` | object | Caller identity — `{ id, displayName }` |
+| `preflights[].createdBy.id` | string | IMS user ID (`profile.email`) |
+| `preflights[].createdBy.displayName` | string | Full name from IMS profile |
```
bhellema: Should align with:

https://github.com/adobe/spacecat-api-service/pull/2377/changes#r3238511024

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,308 @@
...
+
+**Response** `200 OK` — full detail:
+```json
+{
+  "preflightId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
+  "status": "COMPLETED",
+  "url": "https://main--site--org.hlx.page/some-path",
+  "createdAt": "2026-05-11T10:00:00.000Z",
+  "createdBy": {
+    "id": "ABC123@techacct.adobe.com",
```
bhellema: If this is the `email`, then call it the email property, not id.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,308 @@
...
+    "displayName": "John Doe"
```
bhellema, line 187: This requires puling together the first and last name, unless the async job generates this value.

docs/decisions/002-preflight-api-rest-redesign.md
bhellema: @johobot I don't think this is the case for preflight jobs. If we ever leverage bulk creating a massive response this might be the case but JSONB has a size of 1GB.

Digging in I see:

> Preflight results are stored inline in async_jobs.result as a JSONB array — the RPC (wrpc_projection_preflight_merge_audit) accumulates page audit entries directly into that column. There's no S3 involved.
>
> The payload per job is bounded: one entry per URL in the request, each with a small set of named audits. For typical preflight requests (a handful of URLs), this is well within what JSONB inline storage handles fine. The resultLocation/resultType fields are left null for preflight jobs — the result column is the storage mechanism here.
>
> So the existing design is reasonable for this use case. resultType: S3 would only make sense if the result could grow large (e.g. bulk imports), which preflight results don't.

asking claude about the context of this thread it responded with (from the projector service)

> The PR comment is not accurate for the current implementation. The evidence:
>
> 1. The preflight writer (writer.ts) only calls wrpc_projection_preflight_merge_audit and returns undefined — no S3 writes anywhere.
> 2. The RPC (20260317134526_projection_preflight_rpc.sql) only ever writes to async_jobs.result (inline JSONB). resultLocation and resultType are never touched.
> 3. The preflight config has no S3 output kind — it uses OutputKind.Preflight, unlike llmo-bp-bulk-import which does use resultLocation.
> 4. The API controller reads resultLocation/resultType and returns them, but they will always be null for preflight jobs because nothing sets them.
>
> The reviewer may be thinking of a different job type (llmo-bp-bulk-import does write to S3), or describing aspirational behaviour that doesn't exist yet. The current design is: preflight results are always inline in async_jobs.result.
>
> The resultLocation/resultType fields being returned in the poll response is harmless (they'll be null), but the claim that "results are often written to S3" is incorrect today.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,308 @@
...
+  },
+  "updatedAt": "2026-05-11T10:00:05.000Z",
+  "startedAt": "2026-05-11T10:00:01.000Z",
+  "endedAt": "2026-05-11T10:00:05.000Z",
+  "result": {},
+  "resultType": "INLINE",
+  "resultLocation": null,
```
bhellema: I don't think we need to expose these values.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,308 @@
...
+| `createdAt` | ISO 8601 | When the preflight was created |
+| `createdBy` | object | Caller identity — `{ id, displayName }` |
+| `createdBy.id` | string | IMS user ID (`profile.email`) |
+| `createdBy.displayName` | string | Full name from IMS profile |
+| `updatedAt` | ISO 8601 | When the preflight was last updated |
+| `startedAt` | ISO 8601 | When processing began |
+| `endedAt` | ISO 8601 | When processing completed |
+| `result` | object \| null | Audit results when `resultType` is `INLINE`; `null` otherwise |
+| `resultType` | enum \| null | `INLINE` — result is in the `result` field; `S3` or `URL` — result is at `resultLocation` and `result` is `null` |
+| `resultLocation` | string \| null | S3 URI or URL of the result when `resultType` is `S3` or `URL`; `null` when `resultType` is `INLINE` |
```
bhellema: I don't think this is necessary, until it is.

docs/decisions/002-preflight-api-rest-redesign.md
```diff
@@ -0,0 +1,308 @@
...
+```
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `preflightId` | UUID | Unique identifier for the preflight |
+| `status` | enum | `IN_PROGRESS` \| `COMPLETED` \| `FAILED` \| `CANCELLED` |
+| `url` | string | The page URL that was analyzed |
+| `createdAt` | ISO 8601 | When the preflight was created |
+| `createdBy` | object | Caller identity — `{ id, displayName }` |
+| `createdBy.id` | string | IMS user ID (`profile.email`) |
```
bhellema: id -> email

docs/decisions/002-preflight-api-rest-redesign.md
bhellema: I had a beefy comment farther up in the document talking about having to update the query, but deleted now seeing this conversation. This might be a good discussion to have with @ekremney.

docs/decisions/002-preflight-api-rest-redesign.md
bhellema: No other audit has the concept of step, just our preflight ones.  So if we rely on other audits for their facts this step concept makes no sense.

docs/decisions/002-preflight-api-rest-redesign.md
bhellema: There’s currently no design for retrieving history for a single page. The existing flow operates at the site level and then filters results by matching URLs.

I think the only practical way to support page-specific history would be to add a query parameter to the GET call that passes the current URL and filters results server-side. Is that something we want to consider?

For example, if I’m in the MFE and want to view the history for the current page, the current approach would require fetching all site URLs plus their last 50 executions. That could result in a fairly large response payload and slower response times, especially when the user only cares about the current page’s history within the MFE.

Longer term, a “full history” experience probably needs a different UI altogether. But for now, I could see value in surfacing history scoped to the current URL directly in the MFE.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Agreed - removing the cap entirely. The 7-day TTL on AsyncJob records is the natural bound and in practice the number of preflights per site will be low. All results are returned sorted by `createdAt` descending; pagination is deferred until there's evidence it's needed. ADR updated.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Correct - `createdBy` is not available today and needs to be stored in async job metadata at creation time. The ADR already captures this: the controller reads `profile.email`, `profile.first_name`, and `profile.last_name` from the authenticated IMS profile at the point the job is created and writes `{ email, displayName }` into the job metadata. No additional lookup needed downstream.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Aligned - both the list and GET by ID responses now use `{ email, displayName }` consistently. ADR updated.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Done - renamed to `email` throughout (202 response, list, and GET by ID). ADR updated.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Correct - `displayName` is composed from `profile.first_name + ' ' + profile.last_name` (falling back to `profile.name`) at job creation time and stored in the job metadata alongside `email`. The controller handles the composition; nothing is computed on read.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Agreed and confirmed - preflight results are always stored inline in `async_jobs.result` as JSONB; `resultLocation` and `resultType` are never set for preflight jobs. Removing both fields from the GET by ID response entirely. ADR updated.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Removed - both fields are gone from the response shape. ADR updated.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Keeping the ownership check in - a caller substituting a different `:siteId` in the path to read another site's preflight result is a real information disclosure risk (scraped page content, draft CMS content). The check is a one-liner in the implementation and the cost of retrofitting it after an incident is far higher than adding it now. @johobot's concern stands.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Done - see reply above.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Good call on looping in @ekremney - the plan to promote both `siteId` and `jobType` as top-level indexed attributes in `spacecat-shared-data-access` is the right approach so `allBySiteIdAndJobType` queries both indexes without touching `metadata`. Happy to pick that up with him when the implementation PR is ready.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Agreed - `step` has no meaning outside the old preflight-specific flow and is dropped from the new endpoint entirely. The branching in `links.js:102` and `metatags.js:103` is dead code to be removed as part of the implementation. ADR already reflects this.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Good point - added an optional `?url=` query parameter to `GET /sites/:siteId/preflights`. When present the controller filters on `metadata.url` server-side after the indexed `allBySiteIdAndJobType` query, so the MFE can fetch history scoped to the current page without pulling the full site list. ADR updated.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: "mystiqueUrl" is a request parameter only. Not part of the response payload and removed from any public documentation.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: To expand on why this check is still necessary even with the existing `AccessControlUtil.hasAccess(site)` guard:

`hasAccess(site)` validates that the caller has IMS org access to the site identified by `:siteId` in the path - that's the first line of defence and it's already consistent with every other site-scoped controller.

But without the ownership check, the handler would then load the preflight job by `preflightId` alone with no site constraint. A caller who legitimately has access to site A could construct `GET /sites/<siteA-id>/preflights/<preflightId-belonging-to-siteB>`. They pass `hasAccess` on site A just fine, and get back site B's preflight result - scraped page content and draft CMS data from a site they have no rights to.

The ownership check closes that gap: it verifies the loaded job's stored `siteId` matches the path's `:siteId`. A mismatch returns `404` - the same response as a non-existent ID - so the caller can't even confirm the preflight exists on another site.

The two checks cover different things: `hasAccess` guards the site in the path; the ownership check guards the job itself. Both are needed.

docs/decisions/002-preflight-api-rest-redesign.md
GeezerPelletier: Update on this - after thinking through it further, promoting `siteId` and `jobType` as top-level attributes on `AsyncJob` is not actually the right answer. `AsyncJob` is a generic backing store shared by at least four job types today (`preflight`, `preflight-beta`, `site-detection`, `pr-review`). Piggybacking on it means preflight-specific fields (`url`, `createdBy`, etc.) stay buried in an untyped `metadata` blob with no schema enforcement, and the domain concept leaks the execution primitive to API consumers.

The ADR Data Model section has been rewritten to instead **introduce a dedicated `Preflight` entity in `spacecat-shared-data-access`** with first-class indexed fields - `siteId`, `url`, `status`, `createdBy`, timestamps, `result`, `error` - as the source of truth. The entity may reference an `AsyncJob` internally for execution lifecycle tracking without exposing that to callers.

Tracking ticket: SITES-44675 (@ekdogan this is the one to coordinate on). Flagging as a prerequisite to the controller work here.

GeezerPelletier:
Chatted with @bhellema , I'm going to merge this PR just to close off this first and very good and in depth iteration. Subsequent touch points and PRs can provide more updates as needed before the implementation is kicked off.

ravverma (commented):
Hey @GeezerPelletier,

Retrospective review (PR is already merged as part of v1.500.2). Two reviewers, consolidated below. No Critical issues - the API contract is well-shaped and most hard questions were worked through in the 31 threads before merge. Findings concentrate on a Data Model section that was added late, and a couple of operational postures that would be cheaper to nail down in the doc than to revisit in production.

## Strengths

- Site-scoped REST shape (lines 41-49) aligns with the rest of the service and eliminates server-side URL-to-siteId reverse-resolution. The Context framing ("by the time a user triggers a preflight, the siteId is already in application state") is exactly the right justification.
- Stable machine-readable error contract (lines 106-123) with `errorCode` + `message` split and an explicit note that `message` "should not be parsed by clients" is the kind of guidance that prevents downstream regressions. Status-to-code mapping table is unambiguous.
- No phantom CANCELLED records on validation/access failure (lines 125-127) - small but mature operational choice that keeps the job store clean.
- IDOR defense codified at the contract level (line 218): specifying that ownership-mismatch returns 404 indistinguishable from "not found" is correct, and survived earlier pushback with a clear security argument.
- `/preflight/beta/jobs` removed outright (lines 50-60) is the right call when there are no external consumers. Many teams would have left a 90-day shim "to be safe."
- Honest framing of the bulk deferral - "via multiple parallel requests, no API change needed" is the right answer for v1.
- A dedicated `Preflight` entity is actually consistent with the existing codebase (`ScrapeJob`, `ImportJob` already sit as peer top-level entities to `AsyncJob`), even though the ADR reads as if introducing a novel split.

## Issues

### Important (Should Fix)

1. **Data Model section conflated with the API-shape ADR (lines 257-326).** The ADR is titled "Preflight API REST redesign" and frames the problem entirely as endpoint shape and path structure (lines 12-39). The Data Model section was added after the approval and commits to a substantial cross-repo architectural change: dedicated `Preflight` entity in `spacecat-shared-data-access`, parallel backing stores, schema migration. That decision has consequences (dual-write lifecycle, cutover plan, query divergence) that are not reflected in the Consequences section (lines 328-340). Bundling it here also means future searches for "why does Preflight have its own table?" land on a doc titled "REST redesign" rather than a dedicated record. Fix: extract to ADR-003 ("Preflight entity vs AsyncJob extension") and cross-reference; have it explicitly call out `ScrapeJob`/`ImportJob` as the established precedent (turns this from an ad-hoc divergence into a documented pattern, and gives @ekdogan a single discoverable artifact).

2. **No-cap list + in-memory `?url=` filter is a performance footgun written into the contract (lines 139, 306-307).** The cap was removed and pagination deferred on the basis that the 7-day TTL is a "natural upper bound." TTL bounds the absolute worst case, not the steady-state response size. With parallel-fire bulk usage as the intended scaling path, an enthusiastic CMS user or a misbehaving MFE can put thousands of preflight rows under one siteId well within the TTL window. `allBySiteId` returning the whole flat array becomes the hot path, and the ADR documents the `?url=` filter as applied in-memory even though `url` is described as indexed (line 294). Fix: either (a) document `allBySiteIdAndUrl(siteId, url)` as the query method when `?url=` is present so the index is used, or (b) re-introduce a soft cap and commit to cursor-based pagination as the Day-1 contract even if the first cursor returns everything. The deferred-pagination posture is fine; what creates lock-in is the unbounded-array shape becoming load-bearing.

3. **Parallel backing stores have no documented cutover plan (lines 318-321).** "Legacy endpoints continue writing to AsyncJob until retired. New endpoints write exclusively to the Preflight entity. Two backing stores in parallel" is the entire treatment of a dual-write window that will run for at least 90 days. Open questions: do clients listing preflights during the window see legacy-API-created jobs, or only new ones? If only new ones, does the MFE need to query both during transition? How does Mysticat's callback contract behave when it knows AsyncJob ids while the new API surface speaks only preflightIds? The operational cost of a dual-store window is invariably underestimated when the migration plan is one sentence - either consumers see a split-brain view or someone writes ad-hoc bridging code that ends up permanent. Fix: add a "Migration" subsection covering (a) whether legacy rows surface through the new list endpoint, (b) Mysticat callback contract during the window, (c) the trigger that ends it.

4. **Deprecation policy for `/preflight/jobs` is unspecified (lines 251-255).** Marked Deprecated with "migration timeline to be coordinated with consumers" and "Sunset set by PM at Accepted." For a deprecation to be actionable, the ADR should commit to the signaling mechanism (response header, release notes, both), who owns identifying and contacting consumers, and the condition that triggers Sunset. Without these, "deprecated" in practice means "kept forever" - I have watched five-year-old "deprecated" endpoints accumulate new dependencies because the obligations were never written down.

### Minor (Nice to Have)

5. **Status field values not enumerated in list response (line 172).** The example shows `COMPLETED` and `IN_PROGRESS` but the closed set (`FAILED`? `CANCELLED`? `PENDING`?) is only stated in the GET-by-ID detail table. The error model is fully tabulated; the status enum deserves the same treatment for the list shape. Consumers writing UI state machines need the closed set.

6. **`error` field shape unspecified in GET by ID response (line 199, 216).** Example shows `"error": null`; the shape when present is left as `{ code, message }` in the field description but no example or contract for what `code` values appear. Worth one short subsection paralleling the request error table.

7. **"Preflight may reference an AsyncJob internally" (lines 284-286) leaves the relationship ambiguous.** If the Preflight entity owns its own `status`, timestamps, and `result` fields (lines 295-302), the AsyncJob link is either redundant (drop it before it accretes consumers) or load-bearing (deserves a sentence on what AsyncJob still provides - retry semantics, SQS dispatch, worker contract). "May reference" invites both interpretations and the ambiguity will be resolved by whoever writes the implementation, not by design.

8. **Status stays "Proposed" (line 4) on a merged and released ADR.** By v1.500.2 the status should be "Accepted." Trivial follow-up.

## Recommendations

- When the data-model decision is extracted, explicitly cite `ScrapeJob` and `ImportJob` as the established precedent for "domain entity peer to AsyncJob." This converts the choice from an ad-hoc divergence into a documented pattern and answers in advance the next workflow team asking "should we get our own entity too?"
- Add a short "Backward compatibility" section listing what consumers of `/preflight/jobs` need to do to migrate (siteId in path, single-URL body, response shape diffs, Location header vs body-embedded pollUrl). Currently lives only in the "Key changes" bullets and is easy to miss.
- Replace the "alignment with @ekdogan required" line (line 325) with the ticket reference (already there) and a one-line summary of the agreed contract once reached - design docs should not carry unresolved coordination as a permanent footnote.
- CI failures (`build`, `it-postgres`, `deploy-stage`, `branch-deploy`) on a docs-only PR are presumably unrelated to the diff. Not blocking retrospectively, but the next reviewer of a similar docs PR will wonder.

## Assessment

Retrospective: would have approved with the four Important items addressed before merge. The API contract itself is well-shaped and the dedicated-entity choice is actually correct - it just deserves its own ADR and a documented migration plan rather than riding into a merged "REST redesign" doc after approval. None of the findings block the value already shipped; all are cheaper to settle in the doc as follow-ups than to revisit in production.

## Next Steps

1. Extract Data Model section to ADR-003 (or expand 002's title + Consequences and call out the `ScrapeJob`/`ImportJob` precedent).
2. Decide the `?url=` query path: indexed query vs in-memory filter; document accordingly.
3. Add a Migration subsection covering the dual-store window.
4. Specify deprecation signaling and Sunset-trigger conditions for `/preflight/jobs`.
5. Minor cleanups: status enum in list, `error` shape, AsyncJob-reference ambiguity, ADR status to Accepted.

GeezerPelletier:
@ravverma — thanks for the retrospective review, appreciate the thorough read. Here's how we've addressed the points raised:

**1. Scope creep — entity design in an endpoint ADR**
Agreed. The Preflight entity design has been extracted into a standalone [ADR-003](https://github.com/adobe/spacecat-api-service/blob/main/docs/decisions/003-preflight-entity-design.md). ADR-002 now holds a short reference to it. Both are included in the implementation PR (#2478).

**2. Performance footgun — no pagination cap + in-memory URL filtering**
We discussed this and believe the risk is low given the specifics here: results come from a dedicated `Preflight` table indexed on `siteId` (not the shared `AsyncJob` table), preflights are human-triggered from the MFE rather than batch-generated, the 7-day TTL naturally bounds per-site volume, and list items are lightweight with no result payload. ADR-002's list endpoint section now documents this rationale explicitly rather than leaving it implicit. We'll add pagination if evidence of need emerges.

**3. Missing dual-store boundary**
ADR-003 now explicitly documents that legacy `AsyncJob` records from `/preflight/jobs` are not surfaced through the new GET endpoints by design — the two backing stores are independent and legacy records expire naturally via TTL.

**4. Deprecation policy for `/preflight/jobs`**
Leaving this as-is for now — the sunset date is PM-driven and we don't want to spec it without that input.

All ADR updates are in the implementation PR: https://github.com/adobe/spacecat-api-service/pull/2478 — happy to take any further feedback there.

PR #2379, opened 2026-05-09
fix: remove aem-cs-fastly

khushboo5723:
email should not trigger for aem-cs-fastly, as it will lead to confusion after 1-click CDN ONboarding already there for AEM-CS-FASLTY


khushboo5723:
moved these changes into : https://github.com/adobe/spacecat-api-service/pull/2340

PR #2387, opened 2026-05-11
feat: init serenity backend

dmitriigulkin-semrush:
Porting the original BFF PoC to the spacecat-api


aniham:
Critical Issues
                                                                                
  LETS FIX: 1. Filter expression injection — brands.js:348                              

  topicIds are taken from query params and interpolated directly into a gRPC    
  filter expression without validation:
  dimensionFilterQl = topicIds.map((id) => `topic_hash = ${id}`).join(' OR ');  
  If Semrush's filter language has injection semantics (boolean operators,    
  parentheses, etc.), a crafted topicIds value like 0 OR 1=1 could bypass
  filters. At minimum, each id should be validated as a numeric string before   
  interpolation.
                                                                                
  LETS FIX 2. Error message leak — ai-visibility.js:103                                

  return internalServerError(e.message);
  gRPC errors include endpoint URLs, OAuth details, and Semrush-specific
  context. Raw error messages should not be returned to the client. Use a       
  generic message and log the detail.
                                                                                
  I DON'T THINK THIS MATTERS 3. No domain ownership check                                                

  Routes are scoped to report:read via required-capabilities.js, but there is no
   check that the requested domain query parameter belongs to the authenticated
  user's organization. Any report:read token holder can query competitor        
  intelligence data for any domain. Other controllers use                     
  AccessControlUtil.hasAccess(site) — this should do the same.

  ---                                                                           
  High-Severity Issues
                                                                                
  OBVI THIS WILL BE FIXED LATER 4. Vendored generated code — third-party/ai-seo-ts/                         

  ~35,000 lines of auto-generated protobuf JS are checked into the repo. This   
  should be published to a private npm registry or an internal artifact store,
  not vendored in source. It creates noise in diffs, inflates the bundle, and   
  obscures what's actually maintained code. Also: the package pins            
  @bufbuild/protobuf at 2.7.0 in its own package.json while the root project
  installs ^2.12.0 — two different versions may be resolved.

  IDK GRPC SO IDK IF THIS IS RELEVANT 5. String.fromCodePoint obfuscation — grpc-transport.js:36                    
  
  const segment = String.fromCodePoint(118, 52, 45, 114, 97, 119); // "v4-raw"  
  This constructs a URL path segment by obfuscating its string value. It doesn't
   prevent anyone from reading it, adds confusion, and undermines auditability. 
  Write it as the plain string 'v4-raw', or load it from env alongside the other
   overridable config.                                                          
                                                                              
  LET'S FIX THIS 6. 33 c8 ignore annotations

  The project requirement is 100% test coverage. There are 33 /* c8 ignore */   
  directives across the ai-visibility/ support files, bypassing coverage on
  meaningful branches. These should be removed and the branches should be       
  tested, or there should be a justified exception agreed on.                 

 THIS IS OBVI BC IT DIDN'T RUN NPM I 7. Tests not passing locally

  npx mocha test/controllers/ai-visibility.test.js fails with a module          
  resolution error for @quazar/ai-seo-ts through esmock. The test suite needs to
   pass before this merges.                                                     
                                                                              
  ---
  Medium-Severity Issues
                        
  MEH 8. Global token cache with no concurrency guard — grpc-transport.js:41,68
                                                                                
  tokenCache is module-level. If two concurrent Lambda invocations both find the
   token expired at the same moment, they'll both race to fetch a new token. Add
   a simple in-flight promise guard:                                            
  let inflight = null;                                                        
  async function getAccessToken(env) {
    if (tokenCache.token && tokenCache.exp > Date.now() + 5000) return
  tokenCache.token;                                                   
    if (!inflight) inflight = fetchToken(env).finally(() => { inflight = null;  
  });                                                                         
    return inflight;                                                            
  }                                                                           
                                                                                
 LET'S FIX THIS  9. OAuth error details logged to client — grpc-transport.js:66              
                                                                                
  throw new Error(`oauth_failed: ${JSON.stringify(j)}`);
  This error message propagates up and (see issue #2) gets returned in the HTTP 
  response. The JSON.stringify(j) can contain OAuth error descriptions with     
  internal context.                                                             
                                                                                
  LET'S FIX THIS 10. No request timeout on gRPC calls                                        

  createGrpcTransport has no timeoutMs. A slow or hanging Semrush backend will  
  hold the Lambda connection open until the function timeout, burning execution
  time. Add a reasonable timeout (e.g. 15–30s).                                 
                                                                              
  LET'S SET TO 1000 11. Unbounded topicIds array — brands.js:346

  sp.getAll('topicIds') has no length cap. A client could send thousands of IDs,
   producing an enormous filter expression sent upstream. Add a reasonable limit
   (e.g. 50).                                                                   
                                                                              
  LET'S FIX THIS 12. Promise.all without partial failure handling — brands.js:384-386,         
  topics.js:328-330
                                                                                
  Several places fan out to 4 parallel FTS_LLMS calls via Promise.all. If one   
  LLM backend returns an error, the entire endpoint fails. Promise.allSettled is
   used correctly in other places in this PR — apply it consistently.           
                                                                              
  ---
  Minor / Convention Issues
                           
  REMIND ME TO FIX LATER 13. Handlers return {status, body} objects manually
                                                                                
  The project uses ok(), badRequest(), notFound(), internalServerError() from   
  @adobe/spacecat-shared-http-utils. The handler return convention here is a    
  one-off that makes error handling in wrapHandler less obvious (e.g., normalize
   is currently applied to 400s too, though normalizeVisibilityV1SuccessfulBody
  correctly bails early on 'error' in body).

  LET'S FIX THIS 14. eslint-disable max-statements-per-line in header

  The disable comment at the top of grpc-transport.js is described as "transport
   bootstrap" but the actual dense single-line code is in brands.js, topics.js,
  etc. The comment in grpc-transport.js is misleading — the actual long lines   
  should either be broken up or the disable should be inline at the offending 
  lines.

 LET'S FIX THIS 15. package.json bufbuild/protobuf version mismatch

  Root installs "@bufbuild/protobuf": "^2.12.0". The vendored                   
  third-party/ai-seo-ts/package.json declares "@bufbuild/protobuf": "2.7.0".
  Verify the generated protobuf code is compatible with the resolved version.   
                                                                              
  ---
  What's Well Done
                                                                                
  - The controller auto-generates method names from the ROUTE_MAP, eliminating
  boilerplate and keeping the route map as the single source of truth.          
  - required-capabilities.js entries for all 20 routes are complete.          
  - normalizeVisibilityV1SuccessfulBody correctly short-circuits on error       
  bodies.                                                                       
  - resetGrpcClients() export is good for testability.                          
  - Error logging is present at both transport init failure and handler failure.
  - extractSearchParams has a solid URL parsing fallback.                       
                                                                                
  ---                                                                           
  Blockers before merge                                                         
                                                                              
  1. Fix dimensionFilterQl injection — validate topicIds as numeric
  2. Fix error message leak in wrapHandler                                      
  3. Add domain ownership access control check                                  
  4. Fix failing test suite                                                     
  5. Discuss third-party vendoring strategy (private npm vs. checked-in)

aniham (approved):

PR #2395, opened 2026-05-12
feat(plg): stamp steps.preOnboarded=true when fast-tracking PRE_ONBOARDING to ONBOARDED

Kanishkavijay39:
## Summary

- Sets `steps.preOnboarded = true` in `performAsoPlgOnboarding` at the moment a `PRE_ONBOARDING` record is fast-tracked to `ONBOARDED`
- Without this, once the status transitions to `ONBOARDED`, the `PRE_ONBOARDING` status is overwritten and the origin is permanently lost — fetching records that came from pre-onboarding becomes impossible

## Context

Pre-onboarding is done via a one-time admin script (`scripts/plg-preonboard-api.js`) that creates `PlgOnboarding` records with `status: PRE_ONBOARDING`. When the customer later signs up through the self-service flow, the controller fast-tracks the record directly to `ONBOARDED` (skipping the full onboarding pipeline), overwriting the status.

The flag lives in `steps` (already exposed in the DTO) so no DTO or API contract changes are needed. It is spread alongside the existing steps at the point of fast-tracking, so all prior pre-onboarding step data is preserved.

**Companion PR (schema):** adobe/spacecat-shared — `feat/plg-preonboarded-steps-flag`

## Test plan

- [ ] Fast-track path: onboarding a domain with an existing `PRE_ONBOARDING` record → `steps.preOnboarded === true` on the resulting `ONBOARDED` record
- [ ] Normal self-service path (no pre-onboarding): `steps.preOnboarded` is absent
- [ ] Waitlist path from pre-onboarding (org conflict): `steps.preOnboarded` not set — record never reached `ONBOARDED` (intentional)


anshulk-public (approved):
LGTM

PR #2062, opened 2026-03-29
feat: read only admin support

ravverma:
## Summary

- Introduces `hasAdminReadAccess()` on `AccessControlUtil` — returns `true` for full admins **or** read-only admins (`authInfo.isReadOnlyAdmin()`), while `hasAdminAccess()` remains unchanged as the write-only guard
- Wires `readOnlyAdminWrapper` into the middleware chain in `src/index.js`, positioned between `authWrapper` and `s2sAuthWrapper` (innermost of the three, runs after auth is resolved)
- Moves API key routes (`POST/DELETE/GET /tools/api-keys`) from `routeRequiredCapabilities` to `INTERNAL_ROUTES` — these are end-user/admin flows and should not be exposed to S2S consumers
- Reclassifies three sheet-data POST routes from `site:write` → `site:read` in `routeRequiredCapabilities` (these endpoints read/fetch data, not mutate)
- Replaces `hasAdminAccess()` with `hasAdminReadAccess()` in all read-only controller guards across `sites`, `organizations`, `audits`, `configuration`, `project`, `plg-onboarding`, and `user-details` controllers

## Write guards unchanged (`hasAdminAccess()` kept)

`createSite`, `createOrganization`, `createProject`, `removeProject`, all `Configuration` mutations (`registerAudit`, `unregisterAudit`, `updateQueues`, `updateJob`, `updateHandler`, `updateConfiguration`, `restoreVersion`), `createEntitlement`, `createPlgEnrollment`, `feature-flags` PUT/DELETE, suggestions REJECTED status transition, `sites-audits-toggle` execute.

## Test plan

- [ ] `hasAdminReadAccess()` returns `true` for full admin, `true` for read-only admin, `false` for regular user
- [ ] `hasAdminAccess()` returns `false` for read-only admin (write guard preserved)
- [ ] `hasAccess()` org-bypass triggers for both full and read-only admin
- [ ] API key routes absent from `routeRequiredCapabilities`, present in `INTERNAL_ROUTES`
- [ ] Sheet-data POST routes map to `site:read`
- [ ] All 94 `plg-onboarding` tests pass with updated stubs
- [ ] All 23 `user-details` tests pass with updated stubs

PR #2403, opened 2026-05-12
feat(organizations): surface semrushWorkspaceId on GET + admin PATCH

solaris007:
## Summary

Surface the new `semrushWorkspaceId` field on Organization responses and accept it in admin-only PATCH bodies. Completes the API-layer wiring of the Adobe IMS → Semrush integration first pass.

## Part of

Adobe IMS → Semrush integration (first pass). Parent proposal: **adobe/mysticat-architecture#61**.

Sequence completed for this PR:
- ✅ Step 1 - DB column: **adobe/mysticat-data-service#593** (`v5.15.0` in dev/stage/prod)
- ✅ Step 2 - data-access attribute + accessors: **adobe/spacecat-shared#1602** (`@adobe/spacecat-shared-data-access@3.62.0`)
- 👉 Step 3 (this PR) - api surface
- Step 4 (follow-up) - project-elmo-ui consumes the field

## What's changed

- `src/dto/organization.js` - `OrganizationDto.toJSON` now includes `semrushWorkspaceId`
- `src/controllers/organizations.js` - `updateOrganization` accepts the new field, gated on `hasAdminAccess()`. Distinct from `imsOrgId` / `name` / `config` (any org member can update those) because the SR workspace mapping has billing / access implications
- `docs/openapi/schemas.yaml` - new `SemrushWorkspaceId` type; field added to `Organization`, `OrganizationCreate`, `OrganizationUpdate` schemas
- `package.json` - `@adobe/spacecat-shared-data-access` bumped `3.61.0` → `3.62.0`
- `test/controllers/organizations.test.js` - 3 new tests (GET surfaces field, PATCH-as-admin succeeds, PATCH-as-non-admin returns 403)

## What's NOT changed

- No new endpoint - existing `PATCH /organizations/{id}` extended, no separate `/admin/...` route
- No format-validation regex - Semrush has not confirmed workspace-ID format; non-empty validation lives in the shared data-access layer (`value == null || hasText(value)`)
- `createOrganization` controller unchanged - it's already admin-only and passes `context.data` through; OpenAPI's `OrganizationCreate` schema updated to document the field is accepted

## Local verification

- `npm run lint` ✅
- `npm run docs:lint` ✅ - same 192 warnings as main, no new errors
- `npm test` ✅ - 10185 passing
- `npx mocha test/controllers/organizations.test.js` - 73/73 passing including the 3 new tests

## Test plan

- [ ] CI passes
- [ ] After CD deploys to dev: PATCH a dev customer's org with `semrushWorkspaceId` (admin), confirm GET surfaces it
- [ ] Negative path: PATCH as non-admin → 403 with `Only admins can set semrushWorkspaceId`
- [ ] Follow-up (not this PR): `project-elmo-ui` reads `semrushWorkspaceId` from the org response and constructs the open-in-Semrush URL


aniham (approved):

src/controllers/organizations.js
rainer-friederich, line 1: **[Score 75 — FIX] Admin guard inside equality branch enables value-probing oracle**

The admin check only fires when the incoming value _differs_ from the stored one:

```js
if (isString(requestBody.semrushWorkspaceId)
  && requestBody.semrushWorkspaceId !== organization.getSemrushWorkspaceId()) {
  if (!accessControlUtil.hasAdminAccess()) { ... }
```

A non-admin who sends the **current** value passes `isString()` but fails `!==`, so the guard is never reached and the request returns `400 No updates provided`. This lets any org member distinguish "I guessed the right value" (400) from "wrong value" (403) — an information-disclosure oracle. The fix is to check admin on field **presence**, not on field **change**:

```js
if (Object.prototype.hasOwnProperty.call(requestBody, 'semrushWorkspaceId')) {
  if (!accessControlUtil.hasAdminAccess()) {
    return forbidden('Only admins can set semrushWorkspaceId');
  }
  if (isString(requestBody.semrushWorkspaceId)
    && requestBody.semrushWorkspaceId !== organization.getSemrushWorkspaceId()) {
    organization.setSemrushWorkspaceId(requestBody.semrushWorkspaceId);
    updates = true;
  }
}
```

---

**[Score 75 — FIX] No code path to clear `semrushWorkspaceId` to null**

`isString(null)` is `false`, so `PATCH { semrushWorkspaceId: null }` silently skips the block and returns `400`. The field is documented as optional and the shared data-access layer permits `null` (`value == null || hasText(value)`). Once set, admins have no way to unlink an org from Semrush without a direct DB write. An explicit `=== null` branch (admin-gated) is needed.

---

**[Score 75 — FIX] DB uniqueness violation surfaces as 500**

`organization.save()` is awaited with no try/catch around the new branch. A concurrent admin race (two orgs claiming the same workspace ID) or a re-bind attempt will throw a DB unique-constraint error that propagates as a 500. The OpenAPI schema documents the constraint; the controller should catch the violation and return `409 Conflict` with a clear message.

---

**[Score 50 — CONSIDER] Empty string bypasses non-empty validation**

`isString('')` is `true`, so `PATCH { semrushWorkspaceId: '' }` enters the block, clears the admin gate, and calls `setSemrushWorkspaceId('')`. The PR description says non-empty validation lives in the shared data-access layer — if that guard holds this is fine, but it is untested here. Combined with the UNIQUE constraint, an empty string could be permanently reserved if validation is ever loosened.

---

**[Score 50 — CONSIDER] No audit log for a billing-impacting mutation**

The PR description notes the field "has billing and access implications." Every write (old value → new value, actor identity, org ID) should emit an audit/info log so a compromised-admin scenario has a forensic trail. Other high-stakes mutations in this codebase log similar events.

---

**[Score 50 — CONSIDER] Mixed-payload non-admin PATCH silently discards valid changes**

A request like `{ name: 'New Name', semrushWorkspaceId: 'ws_x' }` from a non-admin will run the `name` setter first (in-memory), then hit the admin check on `semrushWorkspaceId`, return 403, and discard the name change without saving. This is surprising. Consider validating all permission constraints before applying any mutation, or at minimum documenting and testing the current behaviour.

src/dto/organization.js
rainer-friederich, line 1: **[Score 75 — FIX] `semrushWorkspaceId` exposed to all org members in GET responses**

The DTO now surfaces this field to every caller who can GET an organization — including non-admin org members. The PR description says the field has "billing and access implications" and writing it is admin-only. Confirm with product whether every member should see the workspace binding. If not, gate the DTO field on `hasAdminReadAccess()` (the same pattern used elsewhere post-PR #2062).

---

**[Score 75 — FIX] `getAll` and `getByImsOrgID` responses not tested for the new field**

All list/read endpoints share `OrganizationDto.toJSON` and now silently surface `semrushWorkspaceId` to clients. No test covers these paths. A future DTO regression (e.g., accidental removal of the field) would go undetected for those endpoints.

---

**[Score 50 — CONSIDER] `undefined` vs `null` produces inconsistent JSON shape**

`getSemrushWorkspaceId()` returns `undefined` when the field has never been set. `JSON.stringify` silently **omits** `undefined` keys but serializes `null`. Clients will see the field present in some GET responses and absent in others, with no way to distinguish "not set" from "does not exist." Normalize in the DTO:

```js
semrushWorkspaceId: organization.getSemrushWorkspaceId() ?? null,
```

---

**[Score 50 — CONSIDER] JSDoc return-type annotation not updated**

If `toJSON` has a JSDoc `@returns` block documenting the returned shape, `semrushWorkspaceId` should be listed there. The annotation is the canonical shape contract for readers of this DTO.

test/controllers/organizations.test.js
rainer-friederich, line 1: **[Score 100 — FIX] Missing integration tests (CLAUDE.md requirement)**

CLAUDE.md is explicit: _"New or modified endpoints must include integration tests in `test/it/` — add shared test logic in `shared/tests/`, seed data in `postgres/seed-data/`, and a wiring file in `postgres/`."_

This PR modifies two paths on an existing endpoint (GET now returns the field; PATCH accepts the field with an admin gate). Neither path has coverage in `test/it/`. At minimum:
- Shared test factory for organizations should assert `semrushWorkspaceId` is returned on GET.
- An integration test should verify the admin-only PATCH path and the 403 on non-admin.

---

**[Score 75 — FIX] Admin PATCH test mutates module-level fixture without cleanup**

The admin success test calls `organization.setSemrushWorkspaceId('ws_admin_set')` through the real controller path, mutating the shared `organizations[0]` fixture. If `organizations` is built once at module scope and not reset in `beforeEach`, every subsequent test that reads `organizations[0].getSemrushWorkspaceId()` will see `'ws_admin_set'` instead of the original value. This is an order-dependent flakiness time-bomb as the suite grows. Clone the fixture per test or reset the field in `afterEach`.

---

**[Score 50 — CONSIDER] No GET test for org where `semrushWorkspaceId` is absent**

The new GET test only covers the case where the field is populated. When the field is `undefined`/`null` (never assigned), the DTO either omits the key or serializes `null` — those are different contracts for clients. A one-line test asserting the "no mapping" shape would pin this contract and prevent silent regressions.

docs/openapi/schemas.yaml
rainer-friederich, line 1: **[Score 75 — FIX] `nullable: true` missing from `SemrushWorkspaceId` schema**

The field is optional and not in any `required` list. GET responses for orgs that have no workspace mapping will return `null` (or omit the key, depending on the DTO fix). Without `nullable: true`, strict OpenAPI validators and code generators will flag null responses as schema violations. Add `nullable: true` to the type definition.

---

**[Score 50 — CONSIDER] Admin-only restriction ambiguously documented across POST and PATCH**

The `SemrushWorkspaceId` description currently says _"Setting this field via PATCH is restricted to admin users."_ The same type is referenced in `OrganizationCreate` (POST). `createOrganization` is already globally admin-only, but a consumer reading `OrganizationCreate` cannot tell whether this field specifically requires admin on POST. Either generalise the description to cover both verbs, or add a note to the `OrganizationCreate` schema referencing the admin requirement.

rainer-friederich (changes requested):
Multi-agent review — 15 issues found (scores ≥ 50). 8 are FIX, 7 are CONSIDER.

**Must-fix before merge (score 75–100)**

- **CLAUDE.md** — missing `test/it/` integration tests for both the GET and the admin-only PATCH path
- **Auth oracle** — admin check is inside the `value !== current` branch; a non-admin can probe the current `semrushWorkspaceId` by sending matching values (gets 400, not 403). Guard must fire on field *presence*.
- **No clear path** — `isString(null)` is false, so `PATCH { semrushWorkspaceId: null }` silently no-ops. There is no way to unlink an org from Semrush through the API.
- **Unhandled 500 on unique collision** — `organization.save()` has no catch for the DB UNIQUE constraint; a workspace-ID conflict returns a 500 instead of a 409.
- **GET data exposure** — `semrushWorkspaceId` is now returned to all org members. The PR says it has billing implications and write is admin-only; confirm whether read should also be restricted.
- **Test fixture pollution** — admin PATCH test mutates the shared `organizations[0]` fixture without resetting it, creating order-dependent flakiness.
- **`nullable: true` missing** — OpenAPI schema will fail validation for orgs without a workspace mapping.
- **List endpoints untested** — `getAll` / `getByImsOrgID` now surface the field but have no test coverage for it.

**Worth considering (score 50)**

- Empty string `""` passes `isString()` and writes to DB — confirm non-empty validation in the shared layer actually guards this.
- `getSemrushWorkspaceId()` returns `undefined` when unset; `JSON.stringify` omits it. Normalize to `?? null` for a stable response shape.
- No audit log for a field described as having billing implications.
- No GET test for the absent-field case (key present as null vs omitted are different contracts).
- OpenAPI admin-restriction language only mentions PATCH, not POST — ambiguous for `OrganizationCreate` consumers.
- `toJSON` JSDoc return-type annotation not updated.
- Mixed-payload non-admin PATCH silently discards valid field changes (e.g. `name`) when it hits the `semrushWorkspaceId` admin gate.

Inline comments are attached to each file with full details.

PR #2409, opened 2026-05-13
feat: new brand topics

dmitriigulkin-semrush:

## Related Issues

Thanks for contributing!


calvarezg (commented):
Two items worth fixing before merge:

**1. Ordering mismatch between spec and code**
`docs/openapi/ai-visibility-api.yaml` describes ordering by *topic volume*, but `brand-topics.js` uses `BRAND_TOPICS_ORDER_BY_ENUM.VISIBILITY`. Visibility and volume are distinct metrics — one of these needs to align with the other.

**2. Missing `domain` validation**
`sp.get('domain')` returns `null` if the parameter is absent, and that null flows directly into both gRPC requests. The error will likely be caught by `responseFromGrpcError` and surface as a 400/502, but an explicit early check + 400 response would give callers a clearer signal. Neither issue breaks existing endpoints — this is a new route — so both are fixable as a follow-up if you'd rather not hold the merge.

calvarezg (approved):

PR #2411, opened 2026-05-13
feat(webhooks): add GITHUB_REVIEWER_LOGIN for configurable reviewer identity

szergen:
## Summary

- Adds optional `GITHUB_REVIEWER_LOGIN` env var to the webhook trigger rule
- When set, the reviewer login is matched directly (e.g. `aighagent`)
- When absent, falls back to `${GITHUB_APP_SLUG}[bot]` — no behaviour change for existing GitHub App bot deployments
- 3 new unit tests covering override path, mismatch, and fallback

## Motivation

The Mysticat GitHub Service posts reviews using a PAT belonging to a shared Generic User Account (`aighagent@adobe.com`). Reviews appear as `aighagent`, not as `mysticat-bot[bot]`. The existing trigger rule gated on `[bot]` suffix would silently drop every `review_requested` event for `aighagent`, so the worker would never be invoked.

## Deployment

To enable for `aighagent`:
```
GITHUB_REVIEWER_LOGIN=aighagent
```

Add this to Vault at `dx_mysticat/{env}/api-service`. No code changes needed per environment.

## Related

- [spacecat-infrastructure PR #522](https://github.com/adobe/spacecat-infrastructure/pull/522) — ownership and setup plan
- Jira: SITES-42733


src/utils/github-trigger-rules.js
```diff
@@ -39,11 +39,15 @@ export function getSkipReason(data, action, env) {
...
-  // Invite-based trigger: reviewer must be the app
+  // Invite-based trigger: reviewer must be the configured login.
+  // GITHUB_REVIEWER_LOGIN overrides the default [bot] suffix — use this
+  // when the reviewer is a plain user account (e.g. a shared service account)
+  // rather than a GitHub App bot.
   if (action === 'review_requested') {
     const reviewer = data.requested_reviewer?.login;
-    if (reviewer !== `${appSlug}[bot]`) {
-      return `reviewer ${reviewer} is not ${appSlug}`;
+    const expectedReviewer = env.GITHUB_REVIEWER_LOGIN ?? `${appSlug}[bot]`;
```
danieljchuser: `??` only falls back on null/undefined. If `GITHUB_REVIEWER_LOGIN` resolves to `""` (a realistic Vault/ECS mis-render or cleared SSM parameter), `expectedReviewer` becomes `""`, no GitHub login matches, and every `review_requested` event silently skips with a trailing-empty reason. Suggest `env.GITHUB_REVIEWER_LOGIN?.trim() || \`${appSlug}[bot]\`` plus a test pinning the empty-string case.

danieljchuser (changes requested):
Hey @szergen,

Nice clean change. One foot-gun in the env-var handling worth tightening before merge, plus a few minor polish items.

## Strengths

- Backward compatible via the `??` fallback at src/utils/github-trigger-rules.js:48 - existing GitHub App bot deployments need no env change.
- Skip-reason message now interpolates `expectedReviewer` (line 50), which is a real diagnosability win - operators can see whether the override was picked up or the default was used.
- Change discipline is clean: one config knob, one branch added, three tests covering match/mismatch/fallback, plus an in-PR spec. No incidental churn.
- The inline comment block at lines 42-45 explains *why* the override exists (shared service account vs GitHub App bot), not just *what* it does.

## Issues

### Important (Should Fix)

**Empty-string `GITHUB_REVIEWER_LOGIN` silently disables the trigger.** At src/utils/github-trigger-rules.js:48, `env.GITHUB_REVIEWER_LOGIN ?? ${appSlug}[bot]` only falls back on `null`/`undefined`. If the env var resolves to `""` (a realistic Vault/ECS mis-render shape, e.g. `init.sh --only secrets` substituting an empty value, or a cleared SSM parameter), `expectedReviewer` becomes `""`, no GitHub login will ever match, and every `review_requested` event silently skips with `reviewer X is not ` (trailing empty). Fail-closed, not exploitable, but operationally invisible. Suggested fix:

```js
const expectedReviewer = env.GITHUB_REVIEWER_LOGIN?.trim() || `${appSlug}[bot]`;
```

Plus a fourth test pinning `GITHUB_REVIEWER_LOGIN: ''` to the fallback path so future regressions surface in CI.

### Minor (Nice to Have)

- **Case sensitivity** (src/utils/github-trigger-rules.js:49): GitHub logins are case-insensitive on the platform, but `reviewer !== expectedReviewer` is strict equality. Mismatched-case `aighagent` vs `AIGhagent` will skip with a confusing "is not" reason. With `[bot]` accounts this rarely bit because the slug was server-defined; with plain user accounts it now can. Consider a case-insensitive compare, or at minimum a test that pins the case-sensitive expectation.
- **Log-grep regression risk** (src/utils/github-trigger-rules.js:50): the skip-reason wording changed from `is not <slug>` to `is not <slug>[bot]` on the default path. Any downstream Splunk/Coralogix dashboard or alert greping for the old phrase will go quiet. The existing test only `.include('mysticat')` so CI will not catch it. Worth a quick search of ops queries.
- **Inline comment wording** (src/utils/github-trigger-rules.js:43): "GITHUB_REVIEWER_LOGIN overrides the default [bot] suffix" is mildly misleading - the override replaces the whole expected reviewer string, not just the suffix.

## Recommendations

- **Reply to Copilot's log-injection thread rather than silently dropping it.** The concern does not apply here in practice: `reviewer` is `data.requested_reviewer.login` from an HMAC-verified GitHub body, and GitHub constrains logins to `[A-Za-z0-9-]{1,39}` (no control chars). The new `expectedReviewer` value is operator-controlled (Vault/ECS), also not an injection vector. Posting that as a thread reply documents the reasoning for future readers.
- **Spec wording at docs/specs/2026-05-13-configurable-reviewer-login.md.** The note "GITHUB_APP_SLUG is still required (security gate in controller) but is not used for the reviewer check" reads as absolute. It is still used on the fallback path. Tightening to "not used for the reviewer check when GITHUB_REVIEWER_LOGIN is set" prevents future confusion.

The motivation describes moving the worker from a GitHub App identity to a shared PAT under `aighagent`. PATs have weaker rotation, weaker per-action attribution, and no per-installation scope - a step back from the GitHub Apps model this codebase was built around (the `[bot]` suffix in the trigger rule was not accidental). The CyberArk-managed Generic User Account is a documented expedient; worth a follow-up so the override does not quietly become the canonical path. Lives in the worker repo, not here.

## Assessment

Ready to merge? With fixes.
Reasoning: The diff is small, tested, and backward compatible, but the nullish-only fallback turns an empty `GITHUB_REVIEWER_LOGIN` into a silent total skip. One-line guard plus a test should land before merge; the rest are polish.

## Next Steps

1. Tighten the env-var fallback (trim + `||`) and add an empty-string test.
2. Reply to Copilot's thread with the GitHub-login-regex rationale.
3. Optional: spec wording fix, case-sensitivity test, downstream log-grep sweep.

szergen:
## Action Plan - PR #2411 (adobe/spacecat-api-service)

### Summary
6 findings from 2 reviewers (1 human, 1 AI agent).
3 to implement across 2 batches, 2 pushed back, 1 deferred, 0 conflicts escalated.

### Batch 1 - Important (implement first)
- [ ] **F01**: Fix empty-string `GITHUB_REVIEWER_LOGIN` silently disabling the trigger — `src/utils/github-trigger-rules.js:48`
  - Reviewers: danieljchuser
  - Source: `[3233001564]` (`ref_type: review_comment`)
  - What to do: Replace `env.GITHUB_REVIEWER_LOGIN ?? \`${appSlug}[bot]\`` with `env.GITHUB_REVIEWER_LOGIN?.trim() || \`${appSlug}[bot]\``. Add a fourth test in `test/utils/github-trigger-rules.test.js` under the `GITHUB_REVIEWER_LOGIN override` block pinning `GITHUB_REVIEWER_LOGIN: ''` → fallback to `[bot]` suffix.

### Batch 2 - Minor (polish)
- [ ] **F05**: Fix misleading inline comment — `src/utils/github-trigger-rules.js:43`
  - Reviewers: danieljchuser
  - Source: review body
  - What to do: Change `// GITHUB_REVIEWER_LOGIN overrides the default [bot] suffix — use this` to `// GITHUB_REVIEWER_LOGIN replaces the entire expected reviewer string — use this`

- [ ] **F06**: Fix spec wording — `docs/specs/2026-05-13-configurable-reviewer-login.md`
  - Reviewers: danieljchuser
  - Source: review body
  - What to do: Change `"GITHUB_APP_SLUG is still required (security gate in controller) but is not used for the reviewer check"` to `"...but is not used for the reviewer check when GITHUB_REVIEWER_LOGIN is set"`

### Pushback List
These findings will not be implemented. Reasons will be posted as replies to the relevant comment threads.

| F02: Log injection in skip-reason string | copilot-pull-request-reviewer | `reviewer` is `data.requested_reviewer.login` from an HMAC-verified GitHub webhook body. GitHub constrains logins to `[A-Za-z0-9-]{1,39}` — no control chars are possible. `expectedReviewer` comes from a Vault/ECS-managed env var, also not an injection vector. No sanitization needed. |
| F04: Log-grep regression risk from wording change | danieljchuser | The old wording (`is not <slug>`) was less informative. The new wording (`is not <expectedReviewer>`) is strictly better for diagnosability — operators see the full configured value including `[bot]` suffix. Updating dashboards/alerts to the new pattern is the correct follow-up. |

### Deferred / Needs Clarification

| Finding | Reviewer(s) | Blocked on |
|---------|-------------|------------|
| F03: Case-insensitive login comparison | danieljchuser | GitHub logins are case-insensitive on the platform, but strict equality is intentional here — both sides of the comparison are operator-controlled values (Vault env var and webhook payload login). A case mismatch indicates a misconfiguration worth surfacing explicitly. Deferring; can revisit if case-mismatch issues are reported in practice. |

src/utils/github-trigger-rules.js
szergen: Fixed. Replaced `??` with `?.trim() ||` and added a test pinning `GITHUB_REVIEWER_LOGIN: ''` to the fallback path. Commit 7ae8ccac.

src/utils/github-trigger-rules.js
```diff
@@ -39,11 +39,15 @@ export function getSkipReason(data, action, env) {
...
+    if (reviewer !== expectedReviewer) {
+      return `reviewer ${reviewer} is not ${expectedReviewer}`;
```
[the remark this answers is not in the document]
szergen: Not implementing. `reviewer` is `data.requested_reviewer.login` from an HMAC-SHA256-verified GitHub webhook body — the payload signature is validated before `getSkipReason` is ever called. GitHub constrains user/bot logins to `[A-Za-z0-9-]{1,39}` (no control characters, no newlines). `expectedReviewer` is an operator-set env var from Vault/ECS, also not an injection vector. No sanitisation needed here.

szergen:
## Review Response Summary

### Implemented (3 findings)
- **F01**: Replaced `??` with `?.trim() ||` for `GITHUB_REVIEWER_LOGIN` fallback; added empty-string test pinning the case to the `[bot]` path — commit 7ae8ccac
- **F05**: Fixed inline comment wording from "overrides the default [bot] suffix" to "replaces the entire expected reviewer string" — commit 7ae8ccac
- **F06**: Tightened spec wording: GITHUB_APP_SLUG exclusion is now conditional ("when GITHUB_REVIEWER_LOGIN is set"), not absolute — commit 7ae8ccac

### Pushed Back (2 findings)
- **F02** (Copilot — log injection): `reviewer` comes from an HMAC-verified webhook body; GitHub constrains logins to `[A-Za-z0-9-]{1,39}`. `expectedReviewer` is an operator env var. No injection vector present.
- **F04** (log-grep regression): The new wording is strictly more informative than the old one. Updating dashboards to match is the right path.

### Deferred (1 finding)
- **F03** (case sensitivity): Strict equality is intentional — a case mismatch between the Vault env var and the webhook login indicates misconfiguration worth surfacing explicitly. Revisit if this causes issues in practice.

danieljchuser (approved):
Hey @szergen,

Thanks for the quick turnaround on the empty-string fix. Approving.

- **Empty-string `GITHUB_REVIEWER_LOGIN`** (was Important): `env.GITHUB_REVIEWER_LOGIN?.trim() || \`${appSlug}[bot]\`` at src/utils/github-trigger-rules.js:48 cleanly closes the silent-skip foot-gun, including whitespace-only values. The new test pinning `GITHUB_REVIEWER_LOGIN: ''` to the fallback path will catch any future regression.
- **Inline comment wording** (was Minor): "replaces the entire expected reviewer string" at line 43 is accurate now.
- **Spec wording** (was a recommendation): "not used for the reviewer check when `GITHUB_REVIEWER_LOGIN` is set" is the right phrasing.
- **Copilot log-injection thread**: nicely engaged with on the merits rather than silently dropped. The HMAC-verified-body + GitHub login regex argument is the right rationale; that thread is now a useful record for future readers.

## Remaining unaddressed (acknowledged)

The case-sensitivity and log-grep regression items from the prior review were Minor and out-of-PR-scope to fix in passing. Fine to defer or skip.

## Assessment

Ready to merge? Yes.
Reasoning: All Important and most polish items addressed; remaining items were explicitly Minor.

PR #2412, opened 2026-05-13
chore(test): run mocha test suite in parallel

sagarmiglani:
## Summary

- Adds `--parallel` and `--timeout 10000` to the unit test script. **Wall time drops from ~2m24s to ~58s on a 16-core machine (2.4× speedup).**
- Adds the supporting setup-env changes required to make parallel runs deterministic.

## What changed

**`package.json`** — `test` script now runs mocha with `--parallel --timeout 10000`.

**`test/setup-env.js`** — three additions, each independently verified as necessary by toggling on/off:

1. **Register `sinon-chai` and `chai-as-promised` once for every worker.** Five test files (`ai-visibility`, `consentBanner`, `scrape`, `sites-audits-toggle`, `webhooks`, plus `support/ai-visibility/grpc-utils`) use these assertions without registering the plugin themselves. In serial mode they piggyback on another file's registration that happens to load first; mocha's `--parallel` mode isolates each worker thread, so each worker now does its own registration.

2. **Root `afterEach` hook that snapshots and restores native globals** (`fetch`, `setTimeout`, `setInterval`, `clearTimeout`, `clearInterval`) at module load. Several test files mutate `global.fetch` via raw assignment (`global.fetch = sandbox.stub()`), which `sandbox.restore()` cannot undo. In serial mode the leak is harmless because file ordering is stable; in parallel mode it intermittently broke other files (e.g. `grpc-transport.test.js` failing with `Attempted to wrap fetch which is already stubbed`). The hook handles all three leak patterns: property stubs (`.restore`), standalone stubs (`isSinonProxy`, no `.restore`), and raw assignments (identity check).

3. **`--timeout 10000`** (from the 2s default). `bot-blocker`, `autofix-checks`, and `llmo/unpublish` perform a one-time \`esmock\` cold start in a `before` hook that takes ~1.6s and intermittently exceeds 2s under parallel I/O load.

## What was tried and dropped

- `--jobs 8` (explicit worker count): measured identical to the default; removed.
- Fixing the `global.fetch = sandbox.stub()` antipattern at the source (e.g. `trial-users.test.js`, `onboard-llmo-modal.test.js`): the root hook covers it generically including for future test files, so source-file edits became unnecessary churn.

## Stability

3 consecutive `npm test` runs all produced **10161 passing / 25 failing** in ~58–60s. The 25 failures are pre-existing on `main` (rooted in `organization.getSemrushWorkspaceId is not a function` from commit `8a9401ec`) and reproduce in serial mode — unrelated to this PR.

## Test plan

- [x] `npm test` runs in parallel and completes in ~1 min
- [x] 3 consecutive runs produce identical pass/fail counts
- [x] Same failing tests as serial mode (no parallel-introduced flakes)
- [x] `npm run lint` passes
- [x] CI passes


sagarmiglani:
Somewhat better:
Pre:
image


Post:
image

test/setup-env.js
```diff
@@ -20,3 +26,35 @@ process.env.AWS_XRAY_SDK_ENABLED = 'false';
...
+// Snapshot native globals at module load (before any test runs) so we can
+// detect and undo leaks left behind by tests that mutate them. Required in
+// parallel mode where multiple test files share a worker's globalThis and a
+// leak in one file poisons the next (most commonly via `global.fetch = ...`).
+const NATIVE_GLOBALS = {
+  fetch: globalThis.fetch,
+  setTimeout: globalThis.setTimeout,
+  setInterval: globalThis.setInterval,
+  clearTimeout: globalThis.clearTimeout,
+  clearInterval: globalThis.clearInterval,
```
solaris007, line 44: **Coverage gap on partial sinon restores.** `sinon.useFakeTimers()` replaces `Date` (and optionally `queueMicrotask`, `setImmediate`/`clearImmediate`, `performance.now`) in addition to the four timers tracked here. All current `useFakeTimers` users restore correctly, so this is latent. The risk is that if a future test forgets `clock.restore()`, this hook will reset `setTimeout`/`setInterval`/`clearTimeout`/`clearInterval` to native via the `globalThis[prop] = native` fallback while leaving `Date` faked - a state sinon never produces, and exactly the partial-restore mismatch the hook exists to prevent. Suggest adding `Date` at minimum, plus `queueMicrotask`, `setImmediate`, `clearImmediate`.

solaris007 (changes requested):
Hey @sagarmiglani,

Solid test-infrastructure improvement. Well-scoped 40-line diff, exemplary PR description (alternatives evaluated and rejected, stability evidenced by three deterministic runs), and the implementation handles the genuinely tricky parts correctly. One in-PR refinement worth addressing, plus a few smaller polish items.

## Strengths

- Well-scoped, single-concern 2-file diff
- PR description is a model: each `setup-env` addition independently justified, alternatives (`--jobs 8`, source-level antipattern fixes) explicitly enumerated and rejected with reasoning
- Correct mocha pattern: `mochaHooks` is the canonical root-hook export for parallel mode (mocha v8+), and the existing `"require": "test/setup-env.js"` in the mocha config wires it without further plumbing
- Identity-based detection (`current !== native`) is the right test for "has this been swapped" - avoids the false positives a heuristic would produce
- Hook ordering verified safe: every `sinon.useFakeTimers` usage in the suite restores in its own `afterEach`, so the root hook runs as a no-op for tests that already clean up
- All three identified leak patterns handled: property stubs with `.restore`, standalone stubs via `isSinonProxy`, raw assignment via identity check
- Chai plugin registration is idempotent - safe alongside files like `sites-audits-toggle.test.js` that already register `chaiAsPromised`

## Issues

### Important (Should Fix)

**`NATIVE_GLOBALS` does not cover the full set sinon can mutate** (`test/setup-env.js`, NATIVE_GLOBALS object)

`sinon.useFakeTimers()` replaces `Date`, and optionally `queueMicrotask`, `setImmediate`/`clearImmediate`, and `performance.now`, in addition to the four timer functions tracked here. Current tests all restore correctly so this is latent today. The risk: if a future test forgets `clock.restore()`, the fallback `globalThis[prop] = native` path will restore `setTimeout`/`setInterval`/`clearTimeout`/`clearInterval` to native while leaving `Date` faked - a state sinon itself never produces, and exactly the partial-restore mismatch this hook exists to prevent. Add `Date` at minimum, and ideally `queueMicrotask`, `setImmediate`, `clearImmediate`.

### Minor (Nice to Have)

1. **Empty catch swallows more than double-restore** (`test/setup-env.js`, afterEach hook)

   ```js
   try { current.restore(); } catch { /* ignore double-restore */ }
   ```

   Behavior is correct because the manual reset on the next line is the source of truth. But the comment narrows intent the code does not enforce: any error from `restore()` is silenced. A future sinon upgrade with different restore semantics, a stub with corrupted internal state, or a custom `restore` that throws would all disappear. Options: narrow the catch (`if (!/already restored|no original/i.test(err.message)) throw err;`), `console.warn` to preserve the signal, or just rephrase the comment to "ignore any restore failure - the manual reset below is the source of truth" so the next reader understands the structure.

2. **Global 10s timeout swallows the 2-10s slow-test band** (`package.json`)

   `--timeout 10000` is a 5x bump from the default to absorb three esmock cold starts (`bot-blocker`, `autofix-checks`, `llmo/unpublish`). Any test that legitimately should be flagged at 2s is now silent up to 10s. Scoping `this.timeout(10000)` to the specific `before` hooks (or using mocha's `slow` threshold) preserves the warning signal for the other ~10000 tests. Not blocking - global 10s is a defensible trade for not chasing three files.

3. **PR description claim about `sites-audits-toggle` does not match the file** (PR description)

   `test/controllers/sites-audits-toggle.test.js` imports and calls `use(chaiAsPromised)` itself. The setup-env registration is idempotent so the code is correct, but the description's evidence is wrong for this entry. Worth re-checking the other four (`ai-visibility`, `consentBanner`, `scrape`, `webhooks`, `grpc-utils`) before relying on the description as the rationale.

4. **No regression test for the hook itself** (`test/setup-env.js`)

   This hook IS the safety net for a class of leaks. A future refactor that breaks it silently breaks defense-in-depth and only surfaces when the next `global.fetch = ...` ships. A small `setup-env.test.js` that imports `mochaHooks`, calls `afterEach()` against a synthetic mutated `globalThis`, and asserts restoration would lock the contract in.

## Recommendations

- The root hook intentionally compensates for the `global.fetch = sandbox.stub()` antipattern (per the PR description). From this point forward the hook is load-bearing, and new tests will copy the antipattern because it works. Consider pairing the hook with an eslint rule that blocks new direct global assignments in `test/**`, and file a follow-up to source-fix at least the no-cleanup files (`trial-users.test.js`, `consentBanner.test.js`, `onboard-llmo-modal.test.js`) so the hook stays a backstop rather than the sanctioned pattern.
- The five files that piggyback on implicit `sinon-chai`/`chai-as-promised` registration remain non-self-contained - copy any of them into a different project and they break. A follow-up adding explicit `use(...)` calls in each makes the central registration belt-and-suspenders rather than load-bearing.
- If parallel-mode stability holds over a couple of weeks, this pattern (mochaHooks + globals snapshot + chai registration) is a candidate for extraction into `spacecat-shared/test-helpers` so other spacecat services (audit-worker, content-scraper, import-worker, autofix-worker) inherit it rather than re-deriving variants.
- Add a one-line comment in `setup-env.js` documenting that `sinon-chai` and `chai-as-promised` assertions are globally available. New test authors will discover the registration only by running into "assertion not found" errors otherwise.

- The 25 pre-existing failures (root cause `organization.getSemrushWorkspaceId is not a function`, almost certainly fallout from commit `8a9401ec feat(organizations): surface semrushWorkspaceId`) are out of scope here, but worth confirming with the organizations-feature owner that they are tracked. Landing parallel mode while main is red makes future regression triage harder.

## Assessment

Ready to merge? With one fix.

Reasoning: High-leverage, low-risk infrastructure change with a thorough description and real measurements. The one Important item is a small in-PR fix that closes a real defense-in-depth gap before this becomes the canonical test infra for the codebase. Everything else is polish or follow-up.

## Next Steps

1. Address the NATIVE_GLOBALS coverage gap (add `Date`, `queueMicrotask`, `setImmediate`, `clearImmediate`)
2. Consider the Minor items (empty catch comment, timeout scoping, regression test, description hygiene)
3. Track the follow-ups (source-fix antipattern, extract to shared, eslint rule)

sagarmiglani:
Thanks @solaris007 — great review. Pushed f4c646e addressing the in-PR items:

**Important:**
- ✅ Expanded `NATIVE_GLOBALS` to include `Date`, `queueMicrotask`, `setImmediate`, `clearImmediate` so the hook can't produce sinon's never-emitted partial-restore state.

**Minor:**
- ✅ **Empty catch comment** — rephrased to "Ignore any restore failure — the manual reset below is the source of truth and runs regardless." Reflects what the code actually does.
- ✅ **Timeout scoping** — dropped `--timeout 10000` from the test script; scoped `this.timeout(10000)` at the describe level in the three affected files (`bot-blocker.test.js`, `autofix-checks.test.js`, `llmo/unpublish.test.js`). Other ~10000 tests retain the 2s fail-fast signal.
- ✅ **Regression test for the hook** — added `test/setup-env.test.js` with six cases (raw assignment, standalone stub, property stub, `useFakeTimers` family restore, no-op, restore throws). 6/6 passing in isolation.
- ✅ **PR description claim about chai plugins** — confirmed your read: `sites-audits-toggle.test.js` _does_ register `chaiAsPromised`, but still calls `.calledWith` without registering `sinon-chai`, so setup-env is still load-bearing for it. Verified the five files (`ai-visibility`, `consentBanner`, `scrape`, `sites-audits-toggle`, `webhooks`) use sinon-chai assertions without registering it themselves. Also corrected: `grpc-utils.test.js` doesn't actually use either plugin — my mistake. Will update the PR description.

**Out of scope / follow-ups (tracked but not in this PR):**
- Source-fix the `global.fetch = sandbox.stub()` antipattern in `trial-users.test.js`, `consentBanner.test.js`, `onboard-llmo-modal.test.js` so the hook stays a backstop.
- Make piggyback files self-contained with their own `use(...)` calls.
- An ESLint rule blocking new raw assignments to `global.*` in `test/**`.
- Extract this pattern (mochaHooks + globals snapshot + chai registration) into `spacecat-shared/test-helpers` once it has soaked in this repo for a couple of weeks.
- The 25 pre-existing `getSemrushWorkspaceId` failures — separate issue, will track with the organizations-feature owner.

Local stability: the new regression tests all pass; targeted parallel run of the eight heaviest test files passes in ~52s. Full `npm test` was flaky on my dev machine today due to system load, but the diff is small and CI is the authoritative signal.

solaris007 (approved):
Hey @sagarmiglani,

Thanks for the quick turn on the prior feedback. All five items from the previous round are addressed cleanly, and the new regression test (`test/setup-env.test.js`) is a strong addition. Two small coverage suggestions on the regression test itself, neither blocking.

## Strengths

- All prior findings addressed:
  - `NATIVE_GLOBALS` in `test/setup-env.js` now covers the full sinon swap surface (`Date`, `queueMicrotask`, `setImmediate`, `clearImmediate`), with the inline comment explaining the partial-restore trap that motivated the expansion.
  - `--timeout 10000` correctly dropped from `package.json` and scoped via `this.timeout(10000)` at the describe level in `bot-blocker`, `autofix-checks`, and `llmo/unpublish`, each with a comment naming the esmock cold start cause. The 2s fail-fast signal is preserved for the rest of the suite.
  - Empty-catch comment in `setup-env.js` now reflects what the code does (silences any restore failure; the manual reset is the source of truth) rather than narrowing to "double-restore."
  - New `test/setup-env.test.js` locks in the contract with cases covering the three real-world mutation shapes, the `useFakeTimers` cascade, the no-op case, and the misbehaving `restore()` case.
  - PR description discrepancy on `sites-audits-toggle` acknowledged.
- Hook design uses identity check (`current !== native`), which correctly handles the `delete globalThis.X` case that a truthiness check would mis-handle. The try/catch on `restore()` plus the unconditional manual fallback means a misbehaving stub cannot brick the suite.
- The belt-and-suspenders local `afterEach` in `setup-env.test.js` (with the comment calling out why the duplication is intentional) avoids the bootstrap paradox of testing a safety hook by relying on the hook for cleanup.
- Change discipline holds: six files, one logical concern, no drive-by reformatting. Out-of-scope follow-ups are enumerated in PR comments rather than smuggled in.

## Issues

### Minor (Nice to Have)

1. **Mixed-mode mutation coverage gap** (`test/setup-env.test.js`)

   The current cases exercise each leak class in isolation (raw assign, standalone sinon stub, sinon property stub) plus the `useFakeTimers` cascade. The `useFakeTimers` case exercises multi-global restore but all via the same mechanism. There is no case where two distinct globals are mutated via two different mechanisms in the same `it` (e.g., raw-assign `globalThis.fetch` + `sinon.stub(globalThis, 'setTimeout')` in the same test, then one `mochaHooks.afterEach()` call). Adds a small belt for the "different `current` shapes across adjacent loop iterations" path. ~5 lines.

2. **Silent no-op `restore()` coverage gap** (`test/setup-env.test.js`)

   The "swallows errors thrown by a misbehaving `restore()`" case proves the manual fallback runs when `restore()` throws. The symmetric case where `restore()` returns normally but does NOT revert the global is not tested. The unconditional `globalThis[prop] = native` line handles it, but the path is implicit. One-liner: a stub whose `.restore` is `() => {}`, then assert the global is restored.

## Recommendations

- Optional: export `NATIVE_GLOBALS` from `setup-env.js` and derive `trackedGlobals` in the regression test from `Object.keys(NATIVE_GLOBALS)`. Removes the only place the two lists can drift. Drift is currently self-correcting (a missing key fails loudly), so this is genuinely optional.
- Optional: add a one-line maintenance note above `NATIVE_GLOBALS` explaining when to extend the list (whenever a test mutates a new global, or whenever sinon's swap surface grows). The current comment explains why the list exists but not how to grow it.

- Author's named follow-ups are the right shape: source-fix the `global.X = stub()` antipattern in the files that still use it (e.g., `trial-users.test.js`, `onboard-llmo-modal.test.js`, `consentBanner.test.js`), add an ESLint rule blocking raw `global.X = ...` assignments in `test/**`, make the piggyback chai-plugin files self-contained, and extract the pattern to `spacecat-shared/test-helpers` after parallel mode has soaked. Worth keeping these on a tracked list so the backstop does not become a permanent sanction for the antipattern.
- Parallel-mode failure modes the current hook does not cover: shared filesystem state, port allocation, env-var mutation across files, test-ordering assumptions baked into describe blocks. None visible in the current suite, but worth a flake-watch for the first 1-2 weeks post-merge with willingness to revert specific suites to serial mode if needed.
- The 25 pre-existing `getSemrushWorkspaceId` failures from commit `8a9401ec` are explicitly out of scope here, but landing parallel mode while main is red makes future regression triage harder. Worth confirming the organizations-feature owner is tracking.

## Assessment

Approving conditionally. The hook is well-shaped, every prior finding is addressed with an implementation that matches or improves on what was asked, and the regression test is thorough for the documented leak patterns. The two Minor coverage gaps below are small belts worth adding while the file is fresh, and the in-progress CI build needs to finish green before merge (the author's "local was flaky today" disclosure makes CI the authoritative signal here).

Treating this as approval rather than CHANGES_REQUESTED because the gaps are coverage-gap additions to a regression test the author proactively added, not defects in the production safety net. If the conditions slip, please flag and I'll re-look.

## Next Steps (conditions of approval)

1. Add the two Minor regression-test cases in `test/setup-env.test.js` (mixed-mode mutation, silent no-op restore) - ~5 lines each.
2. Wait for the in-progress CI build to finish green before merging.
3. Optional recommendations and out-of-scope follow-ups are not conditions.

sagarmiglani:
**Root cause of the prior hang found and fixed** — pushed in 91652e78.

Sinon's `useFakeTimers()` replaces `globalThis.Date`/`setTimeout`/etc. with wrapper functions that do **not** expose `.restore()` — only the returned `clock` object does. The previous version of this hook fell through to `globalThis[prop] = native`, which clobbered the wrappers without telling sinon's internal clock registry. The next file in the same worker that used sinon timers hit the corrupted state and hung silently. That's why CI build #75771194558 timed out at 1h25m.

Verified empirically:
```js
const clock = sinon.useFakeTimers();
typeof globalThis.Date.restore         // undefined
typeof globalThis.setTimeout.restore   // undefined
globalThis.Date.clock === clock        // true  ← back-reference exists
```

**Fix:** the hook now checks `current.clock.uninstall` first (sinon's proper API, reachable via the back-ref), then `current.restore` (property stubs), then raw reassignment. This keeps sinon's registry consistent even when a test forgets `clock.restore()`.

Two new regression cases added to `setup-env.test.js`:
- `uninstalls sinon.useFakeTimers via the .clock back-reference, restoring Date and timers together`
- `lets a subsequent useFakeTimers install cleanly after a prior leak` — this is the one that reproduces the original CI hang if you revert the fix.

**Verified locally:**
- 7/7 regression tests pass
- Targeted parallel run of setup-env.test.js + the 5 `useFakeTimers` files: 210 passing in 2s
- Full `npm test`: 10187 passing / 25 failing in 57s (same 25 baseline failures, no hang)

History on the branch is now original PR → revert → fixed re-attempt, so the diff is the same as the original review-fix intent.

sagarmiglani:
**Walked back the scoped-timeout change** in aed8eff4 — restored `--timeout 10000` globally and dropped the three describe-level overrides this PR had added.

**Why:** CI surfaced `test/agents/tools/github-org-retriever.test.js` hitting the same intermittent 2s timeout under parallel load — that's a sixth file with `esmock` in `before`/`beforeEach` we'd otherwise have to override. Auditing the codebase, there are several more candidates (`run-audit`, `llmo-onboarding`, `plg-onboarding`, `brand-claims`, `opportunity-workspace-demo`, …). Every new test author would also have to remember the override.

Per @solaris007's original note ("global 10s is a defensible trade for not chasing three files") — chasing six-plus files crosses the threshold where the trade flips, so this PR keeps the global bump.

Pre-existing per-file timeout overrides in the codebase (e.g. `timeout(120000)` for the legitimately-long llmo test) are not introduced by this PR and have been left untouched.

**Final diff vs `main`** is 3 files:
- `package.json` — test script adds `--parallel --timeout 10000`
- `test/setup-env.js` — chai plugin registration + sinon-clock-aware root `afterEach` hook
- `test/setup-env.test.js` — 7 regression tests for the hook (includes the case that reproduces the prior CI hang if the sinon fix is reverted)

sagarmiglani:
@solaris007 I have reverted some of the review comments changes - as they were creating some issues with the tests. Above comments have the analysis. Would you like to re-review the changes?

Next step for me would be to split a big test file (test/controllers/plg/plg-onboarding.test.js - 327 tests) which seems to be a slight bottleneck even in parallel execution. I'll create a PR once this PR is merged.

solaris007 (approved):
Hey @sagarmiglani,

Great iteration. The `restoreGlobal()` refactor with `.clock.uninstall()` first is the right root-cause fix - verified against `@sinonjs/fake-timers` source (`hijackMethod` sets `target[method].clock = clock` on every faked global, and `clock.uninstall()` is the proper teardown). And the aed8eff commit restoring the global `--timeout 10000` after a 6th esmock-cold-start file surfaced (`github-org-retriever.test.js`) is exactly the right architectural call - thanks for citing the original review note where I said "global 10s is a defensible trade for not chasing three files." Six is past the threshold.

Approving. A few Minor coverage-symmetry and documentation nits that you can take, leave, or fold into a follow-up.

## Strengths

- **CI flake addressed at the root.** The 2s timeout failure on `github-org-retriever.test.js` (one of 55 unscoped esmock files in the suite) would have recurred for every future esmock-heavy test. Restoring the global timeout is the sustainable fix.
- **Sinon clock-registry corruption diagnosed precisely.** The previous fallback path (`globalThis[prop] = native`) clobbered sinon's wrappers without invoking the proper teardown, leaving the internal `methods` registry pointing at orphaned clocks. The `.clock.uninstall()` path is sinon's documented teardown contract and the fix matches the diagnosis.
- **Three-step graceful degradation** in `restoreGlobal()` is well-ordered: `clock.uninstall()` (sinon `useFakeTimers`) -> `.restore()` (property stubs) -> raw reassignment (standalone stubs and direct mutations). If sinon removes the `.clock` back-reference in a future major bump, the hook degrades to the previously-approved f4c646e behavior, not to a broken state.
- **Regression test "lets a subsequent useFakeTimers install cleanly after a prior leak"** captures the exact failure mode that caused the original CI hang - any future sinon upgrade that breaks the `.clock` contract will fail this test loudly rather than re-surfacing as a silent hang.
- **All prior findings addressed.** NATIVE_GLOBALS coverage, empty-catch comment, timeout scoping (and then global), regression test for the hook itself, PR description hygiene.

## Issues

### Minor (Nice to Have)

1. **`.clock` is an internal of `@sinonjs/fake-timers`, not documented public API** (`test/setup-env.js`, `restoreGlobal` comment block). The current comment describes what the back-reference does but doesn't flag that it's an implementation detail. A one-line addition would protect future maintainers from assuming `.clock` is stable contract: "Note: `.clock` is an internal of @sinonjs/fake-timers, not part of its documented public API. The regression test in setup-env.test.js will fail loudly if a future sinon bump removes it."

2. **`.clock.uninstall()` throw path has no test** (`test/setup-env.test.js`). The sibling case (`.restore()` throws) is covered by `'swallows errors thrown by a misbehaving restore()'`, but the analogous case for `clock.uninstall()` throwing is not. Adding a parallel test (stub whose `.clock.uninstall` throws, assert hook does not throw and the global is restored via the manual fallback) closes the coverage symmetry. Strictly nit-level since both paths fall through to the same manual reset.

3. **`restoreGlobal` returns a boolean that is never consumed** (`test/setup-env.js`). Either drop the returns (simpler) or wire them in - counting restorations per `afterEach` and logging a debug line when greater than zero would give early signal next time a test starts leaking, before it becomes an intermittent hang somewhere else. Currently the hook is silent on every cleanup, including ones that mask a leak.

4. **Manual-reset rationale isn't documented** (`test/setup-env.js`, after the `clock.uninstall`/`restore` block). After a successful `clock.uninstall()`, the manual reset still runs. A future reader will wonder if it's defensive or redundant. One-line comment: "manual reset stays as a safety net in case `clock.uninstall()` throws mid-cleanup and leaves a subset of globals faked."

5. **(Carried from prior approval, still unresolved) Mixed-mode mutation coverage gap** (`test/setup-env.test.js`). No test mutates two distinct globals via two different mechanisms in the same `it` (e.g., raw-assign `globalThis.fetch` + `sinon.useFakeTimers()`, then assert both are restored by a single hook call). The loop in `mochaHooks.afterEach` is trivial enough that this is low-value; would catch a hypothetical bug where one branch of `restoreGlobal` corrupts state visible to a later iteration. Skip if it's churn.

## Recommendations

- File a follow-up to (a) open an upstream sinon issue asking that `.clock` be documented as public API (or that sinon expose a `sinon.restoreAllGlobals()` helper), and (b) add an ESLint rule against raw `global.X = stub` assignments in `test/**`. Either path lets the compensation layer eventually shrink rather than only grow.
- The `--timeout 10000` trade-off lives in the commit message. A short note in `README.md` or `CONTRIBUTING.md` would survive longer than git history when someone asks "why 10s and not 2s?" Borderline - skip if it clutters.

- The compensation-layer surface area is now around 95 lines in setup-env.js plus around 144 lines in setup-env.test.js. For a 10k+ test repo with several global-mutating files, the centralized hook plus regression test is higher-leverage than fixing antipatterns at the source. Revisit if this layer exceeds ~300 lines or starts gaining sinon-version branches.

## Assessment

Ready to merge? Yes.

Reasoning: All material findings from the two prior reviews are addressed. The root-cause analysis behind the `restoreGlobal` refactor is precise and the fix matches it. The follow-up `aed8eff` correctly converged on the global timeout after the scoped approach hit its sixth file. Both `ci/build` runs are green on the current HEAD; the in-progress `ci/it-postgres` runs are unrelated to the test-infra changes.

## Next Steps

1. Wait for the in-progress `ci/it-postgres` runs to finish green
2. Minor items above are optional polish - either fold into a follow-up or take/leave individually

solaris007 (approved):
Hey @sagarmiglani,

Fourth pass at the same HEAD - everything material from rounds 1-3 lands cleanly and nothing here changes the existing approval. A few items below are fresh perspectives from a wider lens (architecture intent, behavioral coverage of the cleanup path); the rest engage explicitly with previously-deferred findings rather than silently re-raise them. None are blocking.

## Strengths

- Previously flagged items addressed: `NATIVE_GLOBALS` coverage (now includes `Date`, `queueMicrotask`, `setImmediate`, `clearImmediate`), empty-catch comment rephrased, regression test added with the case that reproduces the prior CI hang, PR description discrepancy acknowledged, and the global `--timeout 10000` tradeoff revisited and held after a sixth esmock file (`github-org-retriever.test.js`) surfaced the same 2s timeout.
- `restoreGlobal` (test/setup-env.js:60) documents the three-tier fallback and names the consequence of each step missing - the `.clock.uninstall()` rationale in particular carries the prior CI-hang root cause into the comments where future contributors will read it.
- The local `afterEach` in test/setup-env.test.js:32 deliberately reimplements cleanup rather than reusing `restoreGlobal`, which is the right call when the function under test is itself the cleanup mechanism - avoids the bootstrap paradox.
- `mochaHooks` named export plus `mocha.require` wiring in package.json:84 is the canonical mocha v8+ root-hook pattern, and registering the chai plugins in `setup-env.js` is the right placement for parallel mode (each worker reloads it before any test runs).
- Change discipline holds: 3 files, single concern, alternatives evaluated and named in the PR description, follow-ups enumerated rather than smuggled in.
- No security or supply-chain surface introduced: no `package-lock.json` delta, `sinon-chai`/`chai-as-promised`/`chai` were already devDependencies, and the `mochaHooks` export is unreachable from deployed code paths.

## Issues

### Minor (Nice to Have)

1. **`restoreGlobal` returns a boolean that no caller consumes** (test/setup-env.js:91, 99) - carried unresolved through three prior rounds. Two reasonable directions and they pull opposite ways, so flagging both for your call:
   - **Drop the return** - the function's job is side effects on `globalThis`; an unconsumed boolean is a contract that lies about what the function tells you. If observability becomes valuable later it is a two-line addition then.
   - **Wire it to an env-gated debug counter** - `if (process.env.DEBUG_TEST_LEAKS && restored) console.warn(...)` gives one flag for finding leak hotspots when local-run flakiness happens. The hook is otherwise a silent net, which is part of why the source-fix follow-up list will tend to stay open.
   
   Both are defensible; pick one rather than leave the boolean orphaned.

2. **Regression tests assert end-state but never observe the cleanup path** (test/setup-env.test.js, all 7 cases). If `restoreGlobal` ever regresses to "raw assignment only, skip `.clock.uninstall()`", every test would still pass - final globals would be correct - while sinon's internal clock registry silently corrupts. That is the exact bug class the `.clock.uninstall()` step exists to prevent and the one that caused the original CI hang in commit b4c9312a. The case at test/setup-env.test.js:113 ("subsequent `useFakeTimers` install cleanly after a prior leak") is the closest behavioral assertion and catches the specific symptom, which is why this stays Minor. A tighter guard would be a sinon spy on `clock.uninstall` before calling `mochaHooks.afterEach()`, then asserting `spy.calledOnce`. One additional `it` locks in the specific contract. Also covers the carried "`.clock.uninstall()` throw path has no test" finding from round 3, since the spy can be configured to throw.

3. **Two-pattern chai plugin registration leaves intent unclear** (test/setup-env.js:12-18). `setup-env.js` now centrally registers `sinonChai` and `chaiAsPromised`, but a code search shows ~135 test files in this repo still call `use(sinonChai)` themselves. `chai.use()` is idempotent so correctness is fine today, but future test authors will not know whether centralized registration is the new canonical pattern (and per-file calls are redundant) or whether the centralized call is defense-in-depth. One-line addition to the PR description (or a comment in `setup-env.js`) declaring "centralized registration is canonical; per-file `use(...)` is now redundant and will be cleaned up in a follow-up" closes the ambiguity. The cleanup itself is appropriately deferred.

4. **No `test:serial` escape hatch** (package.json:13). The `test` script unconditionally runs `--parallel`. If CI starts flaking and the team needs to bisect "parallel mode vs. real bug", there is no one-line revert path other than editing `package.json`. For a multi-step parallelization effort (next planned PR splits `plg-onboarding.test.js`), keeping both modes available reduces blast radius. Optional: `"test:serial": "c8 --skip-full mocha --timeout 10000 -i -g 'Post-Deploy' --spec=test/**/*.test.js --ignore=test/it/**"`.

## Recommendations

- **Pair the follow-up ESLint rule with at least one source-fix.** The safety net is now load-bearing for the `global.X = stub()` antipattern. Without a lint rule blocking new raw global assignments in `test/**`, this becomes the sanctioned pattern by accident. The follow-up issue should pair the rule with `trial-users.test.js`, `consentBanner.test.js`, and `onboard-llmo-modal.test.js` being source-fixed so the rule is enforceable from day one.
- **Define explicit soak criteria for the `spacecat-shared/test-helpers` extraction.** No such package exists today, so creating one is a new architectural commitment. Without criteria like "extract after N weeks of green CI plus at least one confirmed leak caught by the hook" the soak tends to become permanent. A linked Jira beats keeping it in the PR description.
- **Document `NATIVE_GLOBALS` scope explicitly.** Two-line comment that the snapshot does NOT cover `process.env`, filesystem state, ports, or module-level singletons - sets expectations for the next contributor who looks at the hook wondering why their env-var leak still bleeds across files. This repo currently has at least one env-var-mutation-without-teardown pattern (`test/support/slack/commands/add-repo.test.js`) that is benign today but easy to miss.
- **Cross-repo opportunity.** spacecat-audit-worker, spacecat-import-worker, spacecat-autofix-worker, and spacecat-content-scraper use identical mocha/sinon/chai stacks and effectively identical `test/setup-env.js` files. They have not adopted parallel mode. The same antipattern almost certainly exists. Worth one shared sentence in a follow-up Jira covering parallel-mode rollout plus safety-net adoption across the worker repos rather than redoing this work N more times.

- The 25 pre-existing `getSemrushWorkspaceId` failures from commit 8a9401ec - explicitly out of scope but landing parallel mode on a red baseline makes future regression triage harder ("was this red before parallel mode or because of it?"). Worth a linked tracking ticket so the noise has a known landing point.
- The author's enumerated follow-ups (source-fix the three antipattern files, ESLint rule, piggyback chai-plugin files made self-contained, `plg-onboarding.test.js` split) are the right shape and tracked in PR comments.

## Assessment

Ready to merge? **Yes.**

Reasoning: All material findings from the three prior rounds are addressed, the design choices match mocha parallel mode and sinon's actual behavior, the contract is protected by self-tests including a regression for the prior CI hang, and the deferred polish items are appropriately scoped to follow-up work. The four Minor items above are either re-engagements with carried findings (where I am explicit about the prior disposition) or fresh perspectives at the architecture/test-coverage level. None block merge. CI is green on the current HEAD.

## Next Steps

1. Optionally address Minor items in this PR or fold into the planned follow-up (item 1 is small enough to land here without churn).
2. Open the follow-up Jira covering: ESLint rule + source-fix three antipattern files, `spacecat-shared/test-helpers` extraction with explicit soak criteria, cross-repo parallel-mode rollout, and the 25 pre-existing failures.
3. Proceed with the `plg-onboarding.test.js` split as the next parallelization PR.

PR #2419, opened 2026-05-14
feat(brands): integrate Brand Governance Agent as primary source in /brand-guidelines (LLMO-4918)

anagarwa:
## Summary

- `GET /sites/:siteId/brand-guidelines` now checks Brand Governance Agent first before falling back to Adobe Brand Publish
- Brand Governance lookup is URL-based (`site.getBaseURL()`) — no `brandId` in site config required
- `imsOrgId` for Brand Governance is read from `organization.getImsOrgId()` (same source as Brand Publish)
- If Brand Governance env vars are absent or brand returns 404, falls through to existing Brand Publish path unchanged
- All 9 Mystique workflows and headings/preflight consumers benefit automatically with no changes required on their side

## New env vars required
- `BRAND_GOV_API_BASE_URL` — Brand Governance Agent base URL
- `BRAND_GOV_API_KEY` — `x-api-key` header value
- `BRAND_GOV_IMS_CLIENT_ID`, `BRAND_GOV_IMS_CLIENT_CODE`, `BRAND_GOV_IMS_CLIENT_SECRET` — service account IMS credentials
- `IMS_HOST` — already exists in the service (reused, no new var needed)

## Depends on
[LLMO-4920](https://jira.corp.adobe.com/browse/LLMO-4920) / [spacecat-shared#1605](https://github.com/adobe/spacecat-shared/pull/1605) — `BrandGovernanceClient` must be published before this PR is deployed

## Jira
[LLMO-4918](https://jira.corp.adobe.com/browse/LLMO-4918)

## Test plan
- [ ] Unit tests to be added after team review of the approach (per team agreement)
- [ ] Deploy to stage with Brand Governance env vars set — call `/sites/{siteId}/brand-guidelines` for a site whose domain is registered in Brand Governance; confirm response contains `guidelines: [{ name, text }]` from Brand Governance checks
- [ ] Call same endpoint for a site NOT registered in Brand Governance — confirm response falls back to Brand Publish format
- [ ] Call same endpoint with Brand Governance env vars absent — confirm Brand Publish path works unchanged
- [ ] Verify Mystique summarization/metatags/H1 workflows pick up the new guidelines format

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "@adobe/mysticat-shared-seo-client": "1.3.1",
     "@adobe/spacecat-helix-content-sdk": "1.4.33",
     "@adobe/spacecat-shared-athena-client": "1.9.12",
-    "@adobe/spacecat-shared-brand-client": "1.1.42",
+    "@adobe/spacecat-shared-brand-client": "1.3.0",
     "@adobe/spacecat-shared-content-client": "1.8.24",
     "@adobe/spacecat-shared-data-access": "3.79.0",
     "@adobe/spacecat-shared-drs-client": "1.12.1",
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "@adobe/mysticat-shared-seo-client": "1.3.1",
     "@adobe/spacecat-helix-content-sdk": "1.4.33",
     "@adobe/spacecat-shared-athena-client": "1.9.12",
-    "@adobe/spacecat-shared-brand-client": "1.3.0",
+    "@adobe/spacecat-shared-brand-client": "1.4.0",
     "@adobe/spacecat-shared-content-client": "1.8.24",
     "@adobe/spacecat-shared-data-access": "3.79.0",
     "@adobe/spacecat-shared-drs-client": "1.12.1",
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -9,6 +9,8 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
+import { randomUUID } from 'crypto';
+
 import BrandClient, { BrandGovernanceClient } from '@adobe/spacecat-shared-brand-client';
 import {
   badRequest,
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -310,14 +310,18 @@ function BrandsController(ctx, log, env) {
       // Try Brand Governance Agent first (URL-based lookup, no brandId required)
       const govConfig = getImsConfigForBrandGovernance();
       if (govConfig) {
-        const brandGovClient = BrandGovernanceClient.createFrom(context);
-        const brandGovGuidelines = await brandGovClient.getBrandGuidelinesForUrl(
-          site.getBaseURL(),
-          imsOrgId,
-          govConfig,
-        );
-        if (brandGovGuidelines) {
-          return ok(brandGovGuidelines);
+        try {
+          const brandGovClient = BrandGovernanceClient.createFrom(context);
+          const brandGovGuidelines = await brandGovClient.getBrandGuidelinesForUrl(
+            site.getBaseURL(),
+            imsOrgId,
+            govConfig,
+          );
+          if (brandGovGuidelines) {
+            return ok(brandGovGuidelines);
+          }
+        } catch (govError) {
+          log.warn(`Brand Governance Agent failed for site ${siteId}, falling back to Brand Publish: ${govError.message}`);
         }
       }
 
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -305,7 +305,10 @@ function BrandsController(ctx, log, env) {
 
       const organizationId = site.getOrganizationId();
       const organization = await Organization.findById(organizationId);
-      const imsOrgId = organization?.getImsOrgId();
+      if (!organization) {
+        return notFound(`Organization not found for site: ${siteId}`);
+      }
+      const imsOrgId = organization.getImsOrgId();
 
       // Try Brand Governance Agent first (URL-based lookup, no brandId required)
       const govConfig = getImsConfigForBrandGovernance();
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -9,9 +9,7 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
-import { randomUUID } from 'crypto';
-
-import BrandClient from '@adobe/spacecat-shared-brand-client';
+import BrandClient, { BrandGovernanceClient } from '@adobe/spacecat-shared-brand-client';
 import {
   badRequest,
   notFound,
@@ -265,6 +263,26 @@ function BrandsController(ctx, log, env) {
   }
 
   /**
+   * Gets IMS config for the Brand Governance Agent from the environment.
+   * Returns null if Brand Governance is not configured in this environment.
+   * @returns {object|null} Brand Governance IMS config or null.
+   */
+  function getImsConfigForBrandGovernance() {
+    const {
+      IMS_HOST: host,
+      BRAND_GOV_IMS_CLIENT_ID: clientId,
+      BRAND_GOV_IMS_CLIENT_CODE: clientCode,
+      BRAND_GOV_IMS_CLIENT_SECRET: clientSecret,
+    } = env;
+    if (!hasText(host) || !hasText(clientId) || !hasText(clientCode) || !hasText(clientSecret)) {
+      return null;
+    }
+    return {
+      host, clientId, clientCode, clientSecret,
+    };
+  }
+
+  /**
    * Gets Brand Guidelines for a site.
    *
    * @param {object} context - Context of the request.
@@ -285,22 +303,34 @@ function BrandsController(ctx, log, env) {
         return forbidden('Only users belonging to the organization of the site can view its brand guidelines');
       }
 
-      const brandId = site.getConfig()?.getBrandConfig()?.brandId;
-      const userId = site.getConfig()?.getBrandConfig()?.userId;
-      const brandConfig = {
-        brandId,
-        userId,
-      };
-      if (!hasText(brandId) || !hasText(userId)) {
-        return notFound(`Brand config is missing, brandId or userId for site ID: ${siteId}`);
-      }
       const organizationId = site.getOrganizationId();
       const organization = await Organization.findById(organizationId);
       const imsOrgId = organization?.getImsOrgId();
+
+      // Try Brand Governance Agent first (URL-based lookup, no brandId required)
+      const govConfig = getImsConfigForBrandGovernance();
+      if (govConfig) {
+        const brandGovClient = BrandGovernanceClient.createFrom(context);
+        const brandGovGuidelines = await brandGovClient.getBrandGuidelinesForUrl(
+          site.getBaseURL(),
+          imsOrgId,
+          govConfig,
+        );
+        if (brandGovGuidelines) {
+          return ok(brandGovGuidelines);
+        }
+      }
+
+      // Fall back to Adobe Brand Publish (requires brandId + userId in site config)
+      const brandId = site.getConfig()?.getBrandConfig()?.brandId;
+      const userId = site.getConfig()?.getBrandConfig()?.userId;
+      if (!hasText(brandId) || !hasText(userId)) {
+        return notFound(`Brand config is missing, brandId or userId for site ID: ${siteId}`);
+      }
       const imsConfig = getImsConfig();
       const brandClient = BrandClient.createFrom(context);
       const brandGuidelines = await brandClient.getBrandGuidelines(
-        brandConfig,
+        { brandId, userId },
         imsOrgId,
         imsConfig,
       );
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "@adobe/mysticat-shared-seo-client": "1.3.1",
         "@adobe/spacecat-helix-content-sdk": "1.4.33",
         "@adobe/spacecat-shared-athena-client": "1.9.12",
-        "@adobe/spacecat-shared-brand-client": "1.1.42",
+        "@adobe/spacecat-shared-brand-client": "1.3.0",
         "@adobe/spacecat-shared-content-client": "1.8.24",
         "@adobe/spacecat-shared-data-access": "3.79.0",
         "@adobe/spacecat-shared-drs-client": "1.12.1",
@@ -2052,9 +2052,9 @@
       }
     },
     "node_modules/@adobe/spacecat-shared-brand-client": {
-      "version": "1.1.42",
-      "resolved": "https://registry.npmjs.org/@adobe/spacecat-shared-brand-client/-/spacecat-shared-brand-client-1.1.42.tgz",
-      "integrity": "sha512-8OgJ+zEaQs4nSY6yx2ZeOfTX7D7dC2nEC5ynWkfynanqIItFfz+CMhr+bYGOBtRRmlpdbznN3sO2AbrYW9V95Q==",
+      "version": "1.3.0",
+      "resolved": "https://registry.npmjs.org/@adobe/spacecat-shared-brand-client/-/spacecat-shared-brand-client-1.3.0.tgz",
+      "integrity": "sha512-lzqCeYYe4G5TqHbBYShOmYissoeQf+vHGDAyP9fbK6ZV8NwboC5AXMYVrdkBE+B1PUWCalb0C+D/5QzO64OCFw==",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/helix-universal": "5.4.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "@adobe/mysticat-shared-seo-client": "1.3.1",
         "@adobe/spacecat-helix-content-sdk": "1.4.33",
         "@adobe/spacecat-shared-athena-client": "1.9.12",
-        "@adobe/spacecat-shared-brand-client": "1.3.0",
+        "@adobe/spacecat-shared-brand-client": "^1.4.0",
         "@adobe/spacecat-shared-content-client": "1.8.24",
         "@adobe/spacecat-shared-data-access": "3.79.0",
         "@adobe/spacecat-shared-drs-client": "1.12.1",
@@ -2052,18 +2052,28 @@
       }
     },
     "node_modules/@adobe/spacecat-shared-brand-client": {
-      "version": "1.3.0",
-      "resolved": "https://registry.npmjs.org/@adobe/spacecat-shared-brand-client/-/spacecat-shared-brand-client-1.3.0.tgz",
-      "integrity": "sha512-lzqCeYYe4G5TqHbBYShOmYissoeQf+vHGDAyP9fbK6ZV8NwboC5AXMYVrdkBE+B1PUWCalb0C+D/5QzO64OCFw==",
+      "version": "1.4.0",
+      "resolved": "https://registry.npmjs.org/@adobe/spacecat-shared-brand-client/-/spacecat-shared-brand-client-1.4.0.tgz",
+      "integrity": "sha512-bhR5j9PD70DunbKx9lKuDKpndLJyivK/bW9M+gkeScdP84LTmjhdCUWl4EYq3Anqnr1ZGKMBirxdIFuR2FgspA==",
       "license": "Apache-2.0",
       "dependencies": {
-        "@adobe/helix-universal": "5.4.1",
+        "@adobe/helix-universal": "5.4.2",
         "@adobe/spacecat-shared-ims-client": "1.11.10",
         "@adobe/spacecat-shared-utils": "1.81.1"
       },
       "engines": {
         "node": ">=22.0.0 <25.0.0",
         "npm": ">=10.9.0 <12.0.0"
+      }
+    },
+    "node_modules/@adobe/spacecat-shared-brand-client/node_modules/@adobe/helix-universal": {
+      "version": "5.4.2",
+      "resolved": "https://registry.npmjs.org/@adobe/helix-universal/-/helix-universal-5.4.2.tgz",
+      "integrity": "sha512-LIF6K7YQ78/Wxw2OQY+f5HxdIsoZnsJCcNWm9fnBQBRGHBIh3hQq+krD1yRWe/pkBVxAm7o4FJs8SmJkJjU0LQ==",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@adobe/fetch": "4.3.0",
+        "aws4": "1.13.2"
       }
     },
     "node_modules/@adobe/spacecat-shared-brand-client/node_modules/@adobe/spacecat-shared-ims-client": {
--- a/test/controllers/brands.test.js
+++ b/test/controllers/brands.test.js
@@ -422,6 +422,165 @@ describe('Brands Controller', () => {
 
       expect(response.status).to.equal(403);
     });
+
+    describe('Brand Governance Agent priority chain', () => {
+      const BRAND_GOV_ENV = {
+        IMS_HOST: 'https://ims-na1.adobelogin.com',
+        BRAND_GOV_IMS_CLIENT_ID: 'gov-client-id',
+        BRAND_GOV_IMS_CLIENT_CODE: 'gov-client-code',
+        BRAND_GOV_IMS_CLIENT_SECRET: 'gov-client-secret',
+      };
+
+      const makeSiteWithoutBrand = () => new Site(
+        {
+          entities: {
+            site: {
+              model: {
+                indexes: {},
+                schema: {
+                  attributes: {
+                    config: { type: 'any', name: 'config', get: (value) => Config(value) },
+                    organizationId: { type: 'string', name: 'organizationId', get: (value) => value },
+                  },
+                },
+              },
+              patch: sinon.stub().returns({
+                composite: () => ({ go: () => {} }),
+                set: () => {},
+              }),
+            },
+          },
+        },
+        {
+          log: loggerStub,
+          getCollection: stub().returns({ schema: SiteSchema, findById: stub() }),
+        },
+        SiteSchema,
+        { siteId: SITE_ID, config: Config({}) },
+        loggerStub,
+      );
+
+      it('returns Brand Governance guidelines and skips Brand Publish when site is registered in governance', async () => {
+        const mockGovGuidelines = {
+          id: 'frescopa-2A53',
+          name: 'Frescopa Coffee',
+          imsOrgId: IMS_ORG_ID,
+          guidelines: [{ name: 'Sophisticated Voice', text: 'Use sensory language' }],
+        };
+        const brandGovClientStub = {
+          getBrandGuidelinesForUrl: sinon.stub().resolves(mockGovGuidelines),
+        };
+        const brandClientStub = { getBrandGuidelines: sinon.stub() };
+        const govContext = {
+          ...context,
+          brandGovernanceClient: brandGovClientStub,
+          brandClient: brandClientStub,
+        };
+        const govEnv = { ...mockEnv, ...BRAND_GOV_ENV };
+        const govController = BrandsController(govContext, loggerStub, govEnv);
+
+        const response = await govController.getBrandGuidelinesForSite({
+          ...govContext,
+          params: { siteId: SITE_ID },
+        });
+
+        expect(response.status).to.equal(200);
+        const guidelines = await response.json();
+        expect(guidelines).to.deep.equal(mockGovGuidelines);
+        expect(brandGovClientStub.getBrandGuidelinesForUrl).to.have.been.calledOnceWith(
+          'https://site1.com',
+          IMS_ORG_ID,
+          {
+            host: BRAND_GOV_ENV.IMS_HOST,
+            clientId: BRAND_GOV_ENV.BRAND_GOV_IMS_CLIENT_ID,
+            clientCode: BRAND_GOV_ENV.BRAND_GOV_IMS_CLIENT_CODE,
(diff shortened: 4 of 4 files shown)


src/controllers/brands.js
```diff
@@ -226,22 +246,34 @@ function BrandsController(ctx, log, env) {
...
+      const organizationId = site.getOrganizationId();
+      const organization = await Organization.findById(organizationId);
+      const imsOrgId = organization?.getImsOrgId();
+
+      // Try Brand Governance Agent first (URL-based lookup, no brandId required)
+      const govConfig = getImsConfigForBrandGovernance();
+      if (govConfig) {
+        const brandGovClient = BrandGovernanceClient.createFrom(context);
+        const brandGovGuidelines = await brandGovClient.getBrandGuidelinesForUrl(
+          site.getBaseURL(),
```
buuhuu: Shouldn't the overwritten base url into account?

anagarwa: HI @buuhuu it should be reading this from site config only.

BaDos (approved):

solaris007 (approved):

src/controllers/brands.js
```diff
@@ -9,7 +9,7 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
-import BrandClient from '@adobe/spacecat-shared-brand-client';
+import BrandClient, { BrandGovernanceClient } from '@adobe/spacecat-shared-brand-client';
```
solaris007, line 14: Critical (item 1): this resolves to `undefined` against the pinned `@adobe/spacecat-shared-brand-client@1.1.41` (main is now 1.1.42); neither exports `BrandGovernanceClient` - it exists only on the unmerged companion adobe/spacecat-shared#1605. So the `brands.js` test suite no longer loads (the `ci / build` failure, brands.js at 0% coverage) and `BrandGovernanceClient.createFrom(context)` throws `Cannot read properties of undefined (reading 'createFrom')` at runtime. Land and publish the companion first, bump this dependency forward to the version that ships the export (not back to 1.1.41), and guard with `typeof BrandGovernanceClient?.createFrom === 'function'`.

src/controllers/brands.js
```diff
@@ -226,22 +246,34 @@ function BrandsController(ctx, log, env) {
...
 
+      const organizationId = site.getOrganizationId();
+      const organization = await Organization.findById(organizationId);
+      const imsOrgId = organization?.getImsOrgId();
+
+      // Try Brand Governance Agent first (URL-based lookup, no brandId required)
+      const govConfig = getImsConfigForBrandGovernance();
+      if (govConfig) {
+        const brandGovClient = BrandGovernanceClient.createFrom(context);
+        const brandGovGuidelines = await brandGovClient.getBrandGuidelinesForUrl(
```
solaris007: Critical (item 2): only a 404 falls back. A 5xx, network error, timeout, IMS token failure, or invalid `imsOrgId` all throw out of this block to the outer catch (lines 271-274) and become an error response - so a site with valid Brand Publish config starts failing the moment Brand Governance is enabled in its environment and is merely degraded (not 404) for that brand. Wrap this attempt in its own try/catch, log at WARN, and fall through to Brand Publish on any failure. If some failures are meant to be fatal, distinguish them explicitly.

solaris007 (commented):
Hey @anagarwa,

Reviewed in conjunction with its companion adobe/spacecat-shared#1605 (the `BrandGovernanceClient` this consumes). The config-gated opt-in is a sound pattern and the integration is small and readable, but both PRs are currently red and a few items change behavior for the existing `/brand-guidelines` consumers. Posting as a comment - my merge read is in the assessment at the end.

Note: CI (`ci / build`) is failing - the first critical item is the root cause.

### Strengths
- `getImsConfigForBrandGovernance()` returning null when env vars are absent (`brands.js:208-225`) keeps the new path inert until deployed with credentials - good feature-flag-by-config.
- Authorization (`accessControlUtil.hasAccess`) runs before any Brand Governance call, so an unauthorized caller never triggers a service-token fetch.
- `imsOrgId` is sourced server-side from `organization.getImsOrgId()` and reused for both providers - one notion of tenant identity.

### Issues

#### Critical (Must Fix)

**1. Broken import and cross-repo release ordering (`brands.js:10`).** This imports `{ BrandGovernanceClient }`, but `package.json` pins `@adobe/spacecat-shared-brand-client` at `1.1.41` (main is now `1.1.42`), and neither released version exports `BrandGovernanceClient` - that export exists only on the unmerged companion adobe/spacecat-shared#1605. So the named import resolves to `undefined`: the existing `brands.test.js` suite no longer loads (brands.js coverage dropped to 0%, which is the `ci / build` failure), and at runtime `BrandGovernanceClient.createFrom(context)` (`brands.js:256`) throws `Cannot read properties of undefined (reading 'createFrom')` the first time the BG path runs, 500ing the endpoint for every caller including Brand Publish ones. Fix: land and publish the companion first, bump this dependency forward to the version that ships the export (not back to 1.1.41), and add a guard (`typeof BrandGovernanceClient?.createFrom === 'function'`) so an absent export degrades instead of throwing. Do not merge on the current pin.

**2. Only a 404 falls back; every other Brand Governance failure breaks the endpoint (`brands.js:251-264`).** The companion client returns null only on HTTP 404. A 5xx, network error, timeout, IMS token failure, or invalid `imsOrgId` all throw, and there is no try/catch around the BG block - the throw unwinds to the handler's outer catch (`brands.js:271-274`) and becomes an error response. A site with valid Brand Publish config will start failing the moment Brand Governance is enabled in its environment and is merely degraded (not 404) for that brand. That is an availability regression on an endpoint feeding roughly nine Mystique workflows, and it contradicts the stated "falls through to existing Brand Publish path unchanged." Fix: wrap the BG attempt in its own try/catch, log at WARN, and fall through to Brand Publish on any failure; if some failures are meant to be fatal, distinguish them explicitly. Add a test for "BG configured but errors -> falls back."

**3. Silent response-shape regression for downstream consumers (shape defined in the companion, activated here).** For a BG-registered site this endpoint now returns `{ id, name, imsOrgId, createdAt, updatedAt, guidelines: [{name,text}] }`, whereas the Brand Publish path returns `toneOfVoice`, `coreValues`, `guidelines`, `restrictions`, and `additionalGuidelines`. Mystique's `format_brand_guidelines_to_markdown` renders all five; for BG sites four of them silently vanish from the brand context fed to the model (the `guidelines` array itself stays compatible). The endpoint still returns 200, so nothing alerts. The PR's "benefit automatically with no changes required" is not accurate. Fix: define one response contract for `/brand-guidelines` and map BG checks into it, or make the consumer change part of this rollout and gate the cutover on it.

#### Important (Should Fix)

**4. No tests for the new branch.** The diff changes `brands.js` only, with no test exercising the BG-configured path, the fallback, or the env-absent path - which is exactly what let the undefined import (critical 1) through locally. The "tests after team review" note is reasonable for a spike, but this endpoint feeds production workflows. Add controller tests for: BG returns guidelines, BG 404 -> Brand Publish, BG error -> Brand Publish (after item 2), env unconfigured -> Brand Publish, export missing -> degrade.

**5. The Brand Governance attempt pre-empts the deterministic `notFound` for unconfigured sites (`brands.js:251-266`).** Previously a site missing `brandId`/`userId` returned a clean 404. Now the BG attempt runs first; for a site with neither BG registration nor Brand Publish config, a malformed `imsOrgId` or a non-404 BG error throws before the 404 is reached, so the same request can return 400/500 instead of 404. Fix: only attempt BG when `imsOrgId` is valid and let unconfigured sites fall through to the existing 404 (largely resolved by item 2's try/catch).

#### Minor (Nice to Have)

**6. Env-var inconsistency (`brands.js:184-191` vs `208-225`).** The Brand Publish path reads `BRAND_IMS_HOST` while the BG path reuses the generic `IMS_HOST` plus `BRAND_GOV_IMS_*`. Two host conventions in one file is easy to misconfigure across three environments. Consider a consistent `BRAND_GOV_IMS_HOST` or document the reuse.

### Recommendations
- Provider selection ("BG first, Brand Publish fallback") currently lives inline in the controller, which makes it awkward to unit-test in isolation. Consider extracting a small resolver/service that returns a normalized result so the controller calls it once - this also makes the fallback behavior in critical item 2 directly testable.
- On @buuhuu's open thread about the overridden base URL (`brands.js:258`): the BG lookup keys on `site.getBaseURL()`, the join key between SpaceCat sites and Brand Governance brands. The "reads from site config only" reply does not fully settle it - if a site carries a delivery/override URL distinct from its canonical domain, "resolve brand by domain" can match the wrong brand or none. Worth pinning down which URL is the contractual identity before it is the primary key.
- Emit a structured log/metric dimension for which source served the response (`brand_gov` / `brand_publish` / `fallback`) so the cutover is observable.
- Document the rollback (unset the `BRAND_GOV_*` / `IMS_HOST` vars) and confirm the API Gateway/Lambda timeout exceeds the worst-case BG-then-fallback latency.

### Assessment
Reasoning: Two hard blockers (the undefined import / release ordering, and the fall-back-only-on-404 availability regression) plus a silent cross-consumer response-shape change. The mechanics are sound and fixable without redesigning this file; sequence the companion's release, make the fallback cover all failures, reconcile the response shape, and add the tests.

### Next Steps
1. Fix the release ordering plus guard (item 1) and the fallback-on-error behavior (item 2) - these are the blockers.
2. Reconcile the response shape (item 3) and add the missing tests (item 4).
3. Address the error-ordering (item 5); the minor and recommendations are optional.

src/controllers/brands.js
solaris007, line 14: Confirmed still open, and this is the CI failure. The dependency is still pinned to `@adobe/spacecat-shared-brand-client@1.1.41`, which does not export `BrandGovernanceClient`, so this import link-fails and `brands.test.js` cannot load - `brands.js` shows 0% coverage (lines 1-1431) and the build exits 2. The new governance tests never actually execute (zero occurrences in the run log); they pass locally only because they inject a stub via `context.brandGovernanceClient`, which short-circuits `createFrom` before the missing export is touched. Bump the dependency forward to the version that ships the export (after adobe/spacecat-shared#1605 releases) before merging.

src/controllers/brands.js
solaris007: The new test "returns error response when Brand Governance client throws" asserts 500, which codifies a Brand Governance error as fatal - no fallback. This block is still not wrapped in a try/catch, so a 5xx, timeout, or IMS failure here returns 500 to an existing Brand-Publish-configured site instead of falling through to Brand Publish. If that 500 is the intended availability tradeoff, that is a valid choice, but it contradicts the PR description's "falls through to existing Brand Publish path unchanged." If not intended, catch gov errors here and fall through.

solaris007 (commented):
Hey @anagarwa,

Re-review since my last pass (one new commit: the `brands.test.js` governance suite). Reviewed again alongside the companion adobe/spacecat-shared#1605.

## Previously flagged, now addressed
- A "Brand Governance Agent priority chain" test suite was added, covering the gov-first path, fallback to Brand Publish on null, the 404-when-config-also-missing case, env-absent-skips-gov, and the gov-throws case.

## Still open / confirmed
- Broken import and release ordering (`brands.js:10`) - still the top blocker, and it is the CI failure. The dependency is still pinned to `@adobe/spacecat-shared-brand-client@1.1.41`, which does not export `BrandGovernanceClient`. I confirmed the mechanism in CI: `brands.js` reports 0% coverage (lines 1-1431) because `brands.test.js` cannot load - the `import { BrandGovernanceClient }` link fails against 1.1.41 - so the new governance tests never actually execute (zero occurrences in the run log), and the coverage gate fails the build (exit 2). The suite only passes locally because it injects a stub via `context.brandGovernanceClient`, which short-circuits `createFrom` before the missing export is touched. This stays red until adobe/spacecat-shared#1605 merges and publishes, and this PR bumps the dependency forward to that version.
- Fallback on error: the controller is unchanged, and the new test "returns error response when Brand Governance client throws" asserts `500`. That codifies a Brand Governance error as fatal, with no fallback to Brand Publish. If that is the intended availability tradeoff, it is a valid choice, but it contradicts the PR description's "falls through to existing Brand Publish path unchanged," and it means an existing Brand-Publish-configured site returns 500 when Brand Governance is merely degraded (5xx/timeout/IMS failure) rather than not-found. Please confirm this is intended; if not, catch gov errors and fall through.
- Response shape: unchanged - same divergence as the companion (BG-registered sites receive guidelines-only, losing the other four sections the consumer renders).
- Minor: the env-var host inconsistency (`BRAND_IMS_HOST` for Brand Publish vs the reused `IMS_HOST` for Brand Governance) is unchanged.

## Note
This PR currently shows two approvals (including under the `solaris007` identity, which is shared with my CLI). Given the confirmed broken-import / CI-red state, flagging that those approvals appear to predate or overlook the unbumped dependency.

## Assessment
Ready to merge? With fixes.
The test addition is good, but the import / release-ordering blocker is confirmed unresolved and is the active CI failure, so this is not mergeable as-is. Sequence the companion's release, bump the dependency forward, and confirm the gov-error behavior is the intended tradeoff.

src/controllers/brands.js
anagarwa: @solaris007 addressed via https://github.com/adobe/spacecat-api-service/pull/2419/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R85

anagarwa: @solaris007 addressed via https://github.com/adobe/spacecat-api-service/pull/2419/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R85

anagarwa:
> Hey @anagarwa,
> 
> Re-review since my last pass (one new commit: the `brands.test.js` governance suite). Reviewed again alongside the companion [adobe/spacecat-shared#1605](https://github.com/adobe/spacecat-shared/pull/1605).
> 
> ## Previously flagged, now addressed
> * A "Brand Governance Agent priority chain" test suite was added, covering the gov-first path, fallback to Brand Publish on null, the 404-when-config-also-missing case, env-absent-skips-gov, and the gov-throws case.
> 
> ## Still open / confirmed
> * Broken import and release ordering (`brands.js:10`) - still the top blocker, and it is the CI failure. The dependency is still pinned to `@adobe/spacecat-shared-brand-client@1.1.41`, which does not export `BrandGovernanceClient`. I confirmed the mechanism in CI: `brands.js` reports 0% coverage (lines 1-1431) because `brands.test.js` cannot load - the `import { BrandGovernanceClient }` link fails against 1.1.41 - so the new governance tests never actually execute (zero occurrences in the run log), and the coverage gate fails the build (exit 2). The suite only passes locally because it injects a stub via `context.brandGovernanceClient`, which short-circuits `createFrom` before the missing export is touched. This stays red until [adobe/spacecat-shared#1605](https://github.com/adobe/spacecat-shared/pull/1605) merges and publishes, and this PR bumps the dependency forward to that version.
> * Fallback on error: the controller is unchanged, and the new test "returns error response when Brand Governance client throws" asserts `500`. That codifies a Brand Governance error as fatal, with no fallback to Brand Publish. If that is the intended availability tradeoff, it is a valid choice, but it contradicts the PR description's "falls through to existing Brand Publish path unchanged," and it means an existing Brand-Publish-configured site returns 500 when Brand Governance is merely degraded (5xx/timeout/IMS failure) rather than not-found. Please confirm this is intended; if not, catch gov errors and fall through.
> * Response shape: unchanged - same divergence as the companion (BG-registered sites receive guidelines-only, losing the other four sections the consumer renders).
> * Minor: the env-var host inconsistency (`BRAND_IMS_HOST` for Brand Publish vs the reused `IMS_HOST` for Brand Governance) is unchanged.
> 
> ## Note
> This PR currently shows two approvals (including under the `solaris007` identity, which is shared with my CLI). Given the confirmed broken-import / CI-red state, flagging that those approvals appear to predate or overlook the unbumped dependency.
> 
> ## Assessment
> Ready to merge? With fixes. The test addition is good, but the import / release-ordering blocker is confirmed unresolved and is the active CI failure, so this is not mergeable as-is. Sequence the companion's release, bump the dependency forward, and confirm the gov-error behavior is the intended tradeoff.

@solaris007 
 Hey @solaris007 , thanks for the thorough re-review. Addressing the items still open:                                                                                              
                                                                                                                                                                                    
   > Pagination (brand-governance-client.js:21)     
                                                                                                                                     
  Fixed — the checks fetch now paginates to exhaustion using the page parameter (max pageSize=100 per the API contract). We also moved the COPY scope filter server-side (scope=COPY
   query param) so only relevant checks are transferred. Tests updated to assert multi-page accumulation.                                                                           
   
                                             
  > apiKey interpolated in error message (line 57)
                                                                                                                                    
  Fixed — constructor error now says 'Invalid Brand Governance API Key' without the value.
                                                                                                                                                                                    
  > Response shape (brand-governance-client.js:147-160)
                                                                                                                               
  Agreed this is the most impactful item. Mapping COPY-scoped checks into the full BrandGuidelines shape (toneOfVoice, coreValues, restrictions, additionalGuidelines) is tracked in
   LLMO-4919 as a follow-up. We'd like to land this PR with the current shape and iterate — the fallback to Brand Publish ensures BG-registered sites are not broken in the interim.
   
   >   Cross-org guard (brand-governance-client.js:128)
                                                                                                                                  
  The BrandResponse schema from the Brand Governance Agent does not include an org_id field (confirmed from the OpenAPI spec), so a client-side comparison like BrandClient performs
   is not possible. Org scoping is enforced server-side by the Agent via the x-gw-ims-org-id header sent on every request. We've added a hasText(brand.id) guard to prevent         
  /brands/undefined/checks being called if the resolved brand has no id.
                                                                                                                                       
 Can you please help in approving the PR now

package.json
```diff
@@ -82,7 +82,7 @@
     "@adobe/mysticat-shared-seo-client": "1.3.0",
     "@adobe/spacecat-helix-content-sdk": "1.4.33",
     "@adobe/spacecat-shared-athena-client": "1.9.11",
-    "@adobe/spacecat-shared-brand-client": "1.1.41",
+    "@adobe/spacecat-shared-brand-client": "https://gist.github.com/anagarwa/f642427e1ea90fcaa959ed2931150b23/raw/64a4d162b019eae9c1146a6fcec6896055879338/adobe-spacecat-shared-brand-client-1.1.41.tgz",
```
solaris007: Critical: this points the production dependency at a personal GitHub gist tarball labeled 1.1.41 but with a different integrity hash than the published 1.1.41 (sha512-Ojqz... vs sha512-ZT+A...). The same version string now maps to two artifacts, so SBOM/audits report 1.1.41 and are wrong, and reverting to the registry 1.1.41 later would silently break the named import at brands.js:12 (the export ships in no published version; the companion adobe/spacecat-shared#1605 is still open). The gist is also deletable (account/TOS/cleanup) - if it 404s, `npm ci` and every deploy break on a single-person point of failure.

Do not merge to main: sequence adobe/spacecat-shared#1605's release, publish a real registry version exporting `BrandGovernanceClient`, and pin to that. If integration is urgent before then, use a governed prerelease on the Adobe registry with an honest, distinct version, never a personal gist. The integrity hash pins bytes, not provenance, availability, or governance.

src/controllers/brands.js
```diff
@@ -226,22 +246,38 @@ function BrandsController(ctx, log, env) {
         return forbidden('Only users belonging to the organization of the site can view its brand guidelines');
       }
 
+      const organizationId = site.getOrganizationId();
+      const organization = await Organization.findById(organizationId);
+      const imsOrgId = organization?.getImsOrgId();
```
solaris007: Important: `organization?.getImsOrgId()` yields `undefined` if the org does not resolve (`findById` returns null), and that undefined is then passed as the tenant key to both `getBrandGuidelinesForUrl` and `getBrandGuidelines`. The sibling `getBrandsForOrganization` guards this and 404s; here it is unguarded and the null-org branch is untested (the default test wiring always resolves a real org). The optional chaining predates this PR, but this PR newly routes the value into the Brand Governance call, so it is worth confirming what BG does with an absent org id. Fix: guard `if (!organization)` (fail loud / notFound), or drop the optional chaining if the org is guaranteed, and add a test.

solaris007 (changes requested):
Hey @anagarwa,

Re-review after the two new commits since my last pass (`d3d5598` the fallback fix, `90ff018` the dependency swap), again alongside the still-open companion adobe/spacecat-shared#1605. The graceful-degradation work is solid and closes the prior fallback blockers. The one remaining blocker is entirely the dependency mechanism, not the controller logic.

Note: CI is green, but it is green-by-workaround - see the Critical.

## Previously flagged, now addressed
- Fallback on any Brand Governance failure (prior critical): `d3d5598` wraps the BG attempt in try/catch, logs WARN, and falls through to Brand Publish on any error (brands.js:256-268). A test now drives the gov client to reject and asserts 200. Genuinely fixed.
- BG attempt no longer pre-empts the deterministic 404 (prior important): the missing brandId/userId notFound is now correctly gated behind the BG attempt (brands.js:274-275).
- Tests added (prior important): the "Brand Governance Agent priority chain" suite covers gov-first-skips-publish, fallback-on-null, null-plus-missing-config to 404, env-absent-skips-gov, gov-throws to fallback, and both-fail to 500. The assertions check real behavior (status plus which client was or was not called), and the test-to-code ratio is healthy.
- Credential handling is clean: `getImsConfigForBrandGovernance()` is env-sourced, per-request and function-local, never logged or returned, and fail-closed (returns null if any of the four vars is missing). The WARN log uses `govError.message` rather than the full error object, which avoids dumping an HTTP client error's `.config` and the credentials with it.
- Authorization is unchanged and still precedes the BG call (UUID validation, findById, hasAccess to 403), and `imsOrgId` is server-derived from `organization.getImsOrgId()`, not caller-supplied. That is also why the cross-org concern is acceptable: as you noted, org scoping is enforced server-side via `x-gw-ims-org-id` and the org identity here is server-controlled. Not re-raising it.

## Issues

### Critical (Must Fix)

**1. Production dependency resolves to a personal GitHub gist masquerading as published 1.1.41 (package.json:85, package-lock.json).** Introduced by `90ff018`. This is the same release-ordering blocker from the prior round - the CI symptom is gone but the root cause is unchanged, now expressed as a supply-chain liability. The import at brands.js:12 is a static named import, so `BrandGovernanceClient` must exist at module-load time, and it exists in no published version (the registry tops out at 1.2.0; neither 1.1.42 nor 1.2.0 carries it) because the companion adobe/spacecat-shared#1605 is still open. Four distinct problems, none solved by the integrity hash:
- Version masquerading / auditability: the lockfile `version` still reads "1.1.41" but `resolved` is the gist and `integrity` (sha512-Ojqz...) differs from the real published 1.1.41 (sha512-ZT+A...). The same version string now maps to two different artifacts, so any SBOM or "what are we running" audit reports 1.1.41 and is wrong. It also sets a trap: the obvious future cleanup (revert to the registry 1.1.41) silently breaks the named import.
- Availability: the raw URL is revision-SHA-pinned so silent mutation is not the main risk, but the gist lives on an individual account and can be deleted (gist cleanup, account offboarding, TOS action). When it 404s, `npm ci` can no longer resolve the dependency and every build and deploy, including hotfixes, breaks on a single-person point of failure.
- Governance bypass: resolving off-registry skips registry malware/vuln scanning, license gates, SBOM inventory, and immutability guarantees.
- Built from unmerged code: the gist freezes a snapshot of the in-review companion branch; if review changes the client before it merges, production ships something that was never merged.

  How to fix (sequence the release; this is not a "more hashing" problem):
  - Do not merge to main until adobe/spacecat-shared#1605 merges and a real registry version exporting `BrandGovernanceClient` is published, then pin to that exact version and restore the registry URL.
  - Keep the gist tarball on the feature branch for integration testing only.
  - If integration before the companion ships is genuinely urgent, use a governed prerelease - publish to the Adobe npm registry under an honest, distinct version (for example a `-bg-preview.0` prerelease on a `next` dist-tag) so the lockfile tells the truth and the artifact stays registry-governed. A pinned-SHA git dependency on the actual Adobe-org repo is a less-bad stopgap than a personal gist, but still not ideal for production.
  - Add a tracked task or PR checkbox to revert the pin so it cannot be forgotten.

### Important (Should Fix)

**2. `imsOrgId` can be undefined and is passed unguarded to both providers (brands.js:251).** `const imsOrgId = organization?.getImsOrgId();` - the optional chaining means a site whose org does not resolve (`Organization.findById` returns null) yields `imsOrgId = undefined`, which is then handed as the tenant key to both `getBrandGuidelinesForUrl(baseURL, undefined, ...)` and `getBrandGuidelines({brandId,userId}, undefined, ...)`. The sibling handler `getBrandsForOrganization` guards this case and 404s; here there is no guard, and no test exercises the null-org branch (the default test wiring always resolves a real org). The optional chaining predates this PR, but this PR newly routes the possibly-undefined value into the Brand Governance call, so it is worth confirming what the BG service does with an absent org id before relying on it. Fix: guard `if (!organization)` (fail loud / notFound) or drop the optional chaining if the org is guaranteed, and add a test pinning the behavior.

### Minor (Nice to Have)

**3. The BG catch logs only `govError.message`, dropping diagnostic context (brands.js:267).** For a brand-new external integration, `.message` alone is thin when debugging a fallback storm, and the rest of this file passes the error object as a second arg (`log.error(msg, error)`). There is a real tension to thread: do not simply log the full `govError` object, because an HTTP client error can carry `.config` with the IMS credentials. A targeted field set (for example `govError.name`, a status if present, and `govError.stack`) gives diagnosability without the credential-leak risk. Relatedly, once BG is primary a systematic BG outage degrades every request to Brand Publish with only WARN volume as the signal; consider a distinct structured marker or metric on fallback so "BG fallback rate spiking" is alertable rather than discovered by grepping logs.

**4. Fallback observability is unverified in tests (test/controllers/brands.test.js).** No test asserts the WARN actually fires on fallback, so a later change to `log.debug` or dropping the message would pass silently - add a warn assertion to the gov-throws test. Also, the "returns 500 when Brand Governance throws and Brand Publish also fails" test asserts only `status === 500`, which does not prove both providers were attempted (the point of the test); add `brandGovClientStub` and `brandClientStub` called-once assertions.

**5. Partial Brand Governance config silently disables BG with no signal (brands.js:213-226).** `getImsConfigForBrandGovernance()` returns null if any one of the four vars is missing, and the controller then skips BG without logging. A deployment with three of the four gov vars set, or the credentials set but the host var absent, silently falls back to Brand Publish with zero observability. On the host-var naming raised earlier (`IMS_HOST` here vs `BRAND_IMS_HOST` for Brand Publish): on verification, `IMS_HOST` is the dominant convention in this repo (it is what `llmo.js`, `.env.example`, and the README use) and `BRAND_IMS_HOST` is the pre-existing oddball, so reading `IMS_HOST` for the host while using `BRAND_GOV_IMS_*` for the OAuth client is the correct, convention-aligned choice, not a defect. The residual worth addressing is the silent-skip-on-partial-config: emit a debug/info log when gov config is present but incomplete, and add a partial-config test.

**6. The "Brand config is missing" 404 can misattribute the cause once BG is primary (brands.js:275).** When BG is configured but returns null and the site has no brandId/userId, the user sees "Brand config is missing, brandId or userId," which no longer reflects that BG was tried first and found nothing. Consider wording that reflects "no guidelines from either source."

## Recommendations
- Response shape (engaging the deferred item): accepting that the guidelines-only shape with normalization in the companion client is tracked as LLMO-4919 and is out of scope here. Two small adds for that ticket: a one-line in-code comment at the `return ok(brandGovGuidelines)` pointing to LLMO-4919 so the deferred normalization is discoverable in code, and a check that the BG response is filtered to the same fields the endpoint already exposes so the shape change does not widen the response with governance or internal fields (data minimization).
- Add a CI or policy guard that rejects non-registry dependency URLs (gist or http tarballs, personal-account git refs) for production dependencies, so this class of issue cannot recur silently.
- Decoupling option: a lazy/dynamic import of `BrandGovernanceClient` inside the `if (govConfig)` block would let the controller load against a BG-less npm version, but it only moves the failure from module-load to first-call and does not solve release ordering. Sequencing the release remains the clean answer.

- Pre-existing and not introduced here: `package.json` carries two placeholder/squat dependency entries unrelated to this PR - a `package.json` entry at `^2.0.1` and a `package-lock.json` entry at `^1.0.0`. Their `^` ranges could resolve to arbitrary code from those packages' owners. Not in this diff, but worth a separate cleanup PR.
- Companion-client concerns (response-shape normalization ownership, pagination, and the base-URL-as-join-key question from buuhuu's thread) belong in adobe/spacecat-shared, not here.

## Assessment
Ready to merge? No, not to main as-is.
Reasoning: The controller design is sound and all prior fallback blockers are genuinely closed - on the code alone this would be a yes. The single hard blocker is that package.json and package-lock.json point at a personal, deletable gist that masquerades as published 1.1.41 and is built from unmerged code. Land it once adobe/spacecat-shared#1605 ships a registry-published version exporting `BrandGovernanceClient` (or switch to a governed prerelease), then repoint. The `imsOrgId` guard is the one Important to fix on this branch; the Minors are quick follow-ups.

## Next Steps
1. Hold the merge: sequence the companion's release and repoint the dependency to a real registry version (the critical item).
2. Add the `imsOrgId` guard plus a test (the important item).
3. The Minor items (fallback logging and observability, test hardening, partial-config signal, 404 wording) are optional polish.

anagarwa:
Hey @solaris007 , thanks for the thorough re-review. Addressed the Critical and Important items:
                                                                                                                                
  Critical 1 — Gist tarball → real registry version
  Replaced the personal gist tarball with the published 1.3.0 registry version of @adobe/spacecat-shared-brand-client (PR #1605 
  has now merged and shipped). package.json and package-lock.json are restored to standard registry resolution — the lockfile   
  version, resolved, and integrity fields are consistent again.                                                                 
                                                                                                                                
  Important 2 — imsOrgId unguarded                                                                                            
  Added an explicit if (!organization) return notFound(...) guard after Organization.findById(organizationId) and dropped the 
  optional chaining (?.) since the guard now makes it unnecessary. Added a test that pins the 404 behavior when the site's      
  organization cannot be resolved.
  
  Can you please approve the PR.

solaris007 (approved):
LGTM

duynguyen (approved):

PR #2420, opened 2026-05-14
test: verify updatedBy is set to caller email on suggestion status PATCH

ssilare-adobe:
## Summary

- Adds two unit tests to `test/controllers/suggestions.test.js` that assert `updatedBy` is saved to the caller's profile email when `patchSuggestionsStatus` is called:
  1. Status changes from current value → `updatedBy` is set on the returned suggestion
  2. Status is already `SKIPPED` and skip fields are updated → `updatedBy` is also set

## Test plan

- [ ] `npm test` passes (10262 tests passing, 100% coverage maintained)
- [ ] Both new tests exercise the `profile.email` path via the `context.attributes.authInfo` that carries `test@test.com` in the test setup


MysticatBot (approved):
Hey @ssilare-adobe,

## Strengths

- **src/controllers/suggestions.js:898** - The production fix is correct and minimal. The `else` branch (model lacks `setSkipReason`/`setSkipDetail`) was the only path that omitted `setUpdatedBy`. Adding it here closes a real gap where the audit trail would silently lose the actor on older data-access versions.
- **test/controllers/suggestions.test.js:2957-3029** - Three tests cover all three code paths for `updatedBy`: normal status change, skip-field update on a model that supports skip methods, and skip-field update on a model that does not. The third test directly exercises the new production line.
- Tests assert on observable behavior (the `updatedBy` property on both the response DTO and the underlying entity), not on whether `setUpdatedBy` was called. This is the right verification approach.
- Well-scoped PR: one bug fix, tests for that fix, nothing else mixed in. Clean change discipline.

## Issues

### Minor (Nice to Have)

1. **test/controllers/suggestions.test.js:3007-3011** - The third test mutates `mockSuggestion.findById` via `callsFake` and restores it manually at the end of the test body. If an assertion above throws, the restore never runs, potentially leaking state into subsequent tests. A sinon sandbox restore or `afterEach` would be more robust. That said, this follows the existing pattern in the surrounding test file, so it is a consistency note rather than a blocking concern.

## Assessment

Ready to merge? Yes

A single-line fix that closes an audit-trail gap in an existing code path, backed by tests that exercise all three branches. No architectural, security, or convention concerns.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 24s | <b>Cost:</b> $2.21 | <b>Commit:</b> <code>e666d00fa113bd33c46635b82ef4965af7489d22</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot-Dev (approved):
Hey @ssilare-adobe,

## Strengths

- `src/controllers/suggestions.js:898`: The production fix is minimal and correctly placed - `setUpdatedBy` is called in the `else` branch where the model lacks skip-field support, ensuring the audit trail is preserved regardless of model capability. This closes a gap where mutations could happen without attribution.
- `test/controllers/suggestions.test.js:2987`: The third test deletes `setSkipReason`/`setSkipDetail` from the entity to directly exercise the new production code path. This is not coverage theater - it proves the fix works in the exact degraded-model scenario.
- Tests assert both the response DTO (`bulkPatchResponse.suggestions[0].suggestion.updatedBy`) and the underlying entity state (`suggs[0].updatedBy`), verifying both serialization and actual mutation.
- `profile.email` originates from the authenticated context (`context.attributes.authInfo`), consistent with the existing sibling branch at line 895. No new trust boundary concerns.
- PR follows conventional commit format and documented test patterns (Mocha/Chai/Sinon).

## Issues

### Minor (Nice to Have)

1. `test/controllers/suggestions.test.js:3017-3021` - **Manual stub restoration instead of leveraging test lifecycle.** The third test manually restores `mockSuggestion.findById` at the end of the test body. If the test throws before reaching the restore block, subsequent tests in the suite may see the overridden stub, causing cascading failures that obscure the actual problem. If the broader `describe` block uses `sandbox.restore()` in `afterEach` and the stub is sandbox-owned, the manual restore is redundant and can be removed. If it is not sandbox-owned, consider moving restoration to `afterEach` for safety.

## Assessment

Ready to merge? Yes

A tight, well-scoped change: one-line bug fix ensuring mutation attribution is consistent across all code paths in the suggestion status PATCH handler, plus three tests covering the relevant scenarios. Codecov confirms full coverage. The PR title says "test:" but it correctly includes a production fix - semantic-release identified it as a patch release.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 32s | <b>Cost:</b> $2.37 | <b>Commit:</b> <code>e666d00fa113bd33c46635b82ef4965af7489d22</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2422, opened 2026-05-15
fix(plg): grant broken-backlinks suggestions by authority score descending

jsanchez07:
[SITES-44683](https://jira.corp.adobe.com/browse/SITES-44683)
Problem
                                                                                                                                                                                                 
  PLG trial customers were seeing the lowest-authority broken backlinks in their preview instead of the most impactful ones. The broken-backlinks grant strategy had no custom sortFn, so it fell through to the default ascending sort. Since rank = traffic_domain (0–100 domain authority score from the SEO provider), ascending order granted the least authoritative backlinks first. The existing test was asserting the wrong order and masking the bug.

  Changes

  - Added a descending sortFn to the broken-backlinks strategy in grant-suggestions-handler.js — same pattern already used by CWV                                                                
  - Updated the existing large test that was validating the buggy ascending order
  - Added two new tests: descending sort by rank, and tie-breaking by ID ascending                                                                                                               
                                                                                                                                                                                                 
  Notes                                                                                                                                                                                                                                                                                                    
  - CWV was already correct (descending)                                                                                                                                                         
  - Alt-text is unaffected — all suggestions use a hardcoded rank: 1 so sort direction is irrelevant

### DIFF
--- a/src/support/grant-suggestions-handler.js
+++ b/src/support/grant-suggestions-handler.js
@@ -78,7 +78,10 @@ function defaultSortFn(groupA, groupB) {
 const OPPORTUNITY_STRATEGIES = {
   // Groups suggestions by target URL (url_to) so all backlinks pointing to the
   // same broken URL are granted as a single unit. The group rank is the highest
-  // rank among its items, since higher rank = top priority opportunity.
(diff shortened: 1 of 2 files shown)


mikeweis (approved):
LGTM 👍

PR #2424, opened 2026-05-15
feat: Add preflight trace id to allowed headers (SITES-43568)

noozoo:
This change is the first part of adding logging and tracing functionality for ASO Preflight.  All it does is add a new allowed header (x-preflight-request-id) to the list of allowed headers.  It also (temporarily) logs the value of this when a preflight request is made.

  describe here the problem you're solving.  https://jira.corp.adobe.com/browse/SITES-43568

If the PR is changing the API specification: no API changes

If the PR is changing the API implementation or an entity exposed through the API: no API implementation except for two logging statements

If the PR is introducing a new audit type: No new audit type


GeezerPelletier:
Clean, low-risk change — CORS header addition is correct and the temporary logging is clearly marked. One thing worth clarifying:

**Header naming — `x-preflight-request-id` vs `x-trace-id`**

The PR description mentions `x-preflight-request-id` but the implementation uses `x-trace-id`. Minor nit on the description, but it also raises a question about intent: PR #2447 (just merged) propagates `ctx.traceId` from the framework into the SQS message downstream. This PR allows an incoming `x-trace-id` header through CORS. Are these intended to carry the same trace ID — i.e., the MFE sends `x-trace-id`, the framework picks it up as `ctx.traceId`, and #2447 threads it forward? Or are they tracking two independent trace IDs? Worth confirming so the two PRs are clearly complementary rather than accidentally parallel.

PR #2426, opened 2026-05-16
feat(llmo): detect AEM CS Fastly behind WAF simple proxy (Case 0)

ABHA61:
## Summary

Fixes a gap in CDN detection that blocked the \"+Enable\" (Deploy Optimizations to AI Agents) flow for AEM CS sites running behind a WAF or reverse proxy (Case 0 — WAF Simple Proxy).

**What changed:**

- **Phase 1.5 added to `detectCdnForDomain`** (`cdn-detection.js`): when Phase 1 DNS returns `byocdn-other` (non-Adobe IPs), an HTTP probe runs 3 sequential requests with 4 signals to confirm AEM CS Fastly is present behind a WAF:
  1. `x-correlation-id` echoed unchanged in every response
  2. `x-tokowaka-request-id` or `x-edgeoptimize-request-id` present in every response
  3. `x-request-id` unique across all 3 responses (WAF is not caching)
  4. `x-aem-debug` header has `host=<probed-domain>` (Case 0 specific — Case 1 sites show the AEM origin host here instead)

- **`detectAemCsFastlyForDomain` updated** (`edge-routing-utils.js`): the \"+Enable\" endpoint previously only ran a DNS check, which returns null for Case 0 sites. Now falls back to Phase 1.5 HTTP probes when DNS doesn't confirm AEM CS Fastly.

- **Null guard fix in `detectAdobeManagedCdn`**: changed `||` to `&&` — previously if the bare domain DNS failed but `www.domain` resolved (non-Adobe), it returned `null` instead of `byocdn-other`, preventing Phase 1.5 from ever running.

- **`globalThis.fetch` for probes**: `@adobe/fetch` (tracingFetch) implements RFC 7234 client-side HTTP caching — repeated calls to the same URL return cached responses with identical `x-request-id`, breaking Signal 3. `globalThis.fetch` (Node.js native/undici) has no HTTP cache, ensuring every probe hits the network.

**Scope:** Case 0 only. Case 1 (customer BYOCDN in front of AEM CS Fastly) and Case 2 will be addressed in separate PRs.

## Jira

[LLMO-4930](https://jira.corp.adobe.com/browse/LLMO-4930)

## Root cause

Ex if any domain `www.example.com` sit behind a WAF whose CNAME resolves to non-Adobe IPs. Phase 1 (DNS-only) correctly returns `byocdn-other` but can't confirm AEM CS Fastly. Phase 2 (generic multi-signal fingerprint) returns `byocdn-fastly` — correct for the *onboarding* CDN detection flow — but `detectAemCsFastlyForDomain` (used by the \"+Enable\" flow) only ran DNS and returned `null`, causing the error: *"Requested CDN type 'aem-cs-fastly' does not match the detected CDN for this domain"*.

Phase 1.5 fills this gap: it runs only when Phase 1 is a clean DNS miss (`byocdn-other`) and exclusively confirms `aem-cs-fastly` — no Phase 2 fallback, no false positives for genuine third-party CDNs.

## Test plan

- [x] Phase 1.5 unit tests: all 4 signals, partial-batch failure, www-then-bare fallback
- [x] Null guard fix: bare domain DNS failure now correctly returns `byocdn-other` (2 tests updated)
- [x] Full suite passes — `npm test` (10,286 tests)
- [x] ESLint clean on changed files
- [x] End-to-end tested via dev API (`api/ci`) with a newly onboarded domain


src/support/cdn-detection.js
```diff
@@ -170,6 +171,110 @@ async function detectAdobeManagedCdn(domain, log) {
...
+ *   2. x-request-id unique across all 4 probed responses (WAF is not caching)
+ *   3. x-aem-debug response header contains host=<bare-domain>
+ *
+ * Two parallel batches × 2 sequential requests = 4 total probes. All three
+ * signals must hold across all 4 responses; any failure returns null.
+ * ========================================================================== */
+
+const CASE0_PROBE_TIMEOUT_MS = 5000;
+
+async function makeCase0Probe(domain, log) {
```
ABHA61: think of better name for this function

src/support/cdn-detection.js
```diff
@@ -170,6 +171,110 @@ async function detectAdobeManagedCdn(domain, log) {
...
+  const url = `https://${domain}`;
+  const controller = new AbortController();
+  const id = setTimeout(() => controller.abort(), CASE0_PROBE_TIMEOUT_MS);
+  try {
+    const response = await fetch(url, {
+      method: 'GET',
+      redirect: 'follow',
+      follow: 5,
+      headers: {
+        'x-correlation-id': randomUUID(),
```
ABHA61: instead of randomUUID can give static constant - adobeedgetest

src/support/cdn-detection.js
```diff
@@ -170,6 +171,110 @@ async function detectAdobeManagedCdn(domain, log) {
...
+/**
+ * Phase 1.5: Detect AEM CS Fastly behind a WAF simple proxy (Case 0).
+ *
+ * Sends 4 HTTP probes (2 parallel batches × 2 sequential each) and verifies
+ * three signals that uniquely identify AEM CS Fastly responses. Returns
+ * 'aem-cs-fastly' when all signals are confirmed, null otherwise.
+ */
+async function detectAemCsFastlyBehindWaf(domain, log) {
+  const runBatch = async () => {
+    const resp1 = await makeCase0Probe(domain, log);
```
ABHA61: instead of using resp1 and resp 2, we should not duplicate the code. for should fire this in loop and do the identification and analyses

src/support/cdn-detection.js
```diff
@@ -170,6 +171,110 @@ async function detectAdobeManagedCdn(domain, log) {
...
+    if (!resp1) {
+      return null;
+    }
+    const resp2 = await makeCase0Probe(domain, log);
```
ABHA61: lets think of new name for case 0- read the wiki and get the context of case 0 https://wiki.corp.adobe.com/spaces/AEMSites/pages/3865777275/Edge+Optimize+-+Simple+Proxy+for+AEM+CS+CDN+BYOCDN

ABHA61:
Tested on dev env /api/ci using localhost of ELMO UI.
Screenshot 2026-05-19 at 2 08 41 PM
Screenshot 2026-05-19 at 2 08 18 PM
Screenshot 2026-05-19 at 2 07 32 PM

src/support/cdn-detection.js
```diff
@@ -163,13 +163,137 @@ async function detectAdobeManagedCdn(domain, log) {
...
+ *   4. x-aem-debug response header contains host=<probed-domain>
+ *
+ * CASE0_PROBE_COUNT sequential probes (default 3). All four signals must hold
+ * across every response; any failure returns null.
+ * ========================================================================== */
+
+const CASE0_PROBE_TIMEOUT_MS = 3000;
+const TEST_CORRELATION_ID = 'adobeedgetest';
+const AEM_DEBUG_VALUE = 'edge=true';
+const EDGE_OPTIMIZE_TEST_USER_AGENT = 'AdobeEdgeOptimize-Test';
```
nit23uec: pls reuse `EDGE_OPTIMIZE_USER_AGENT`

src/support/cdn-detection.js
```diff
@@ -163,13 +163,137 @@ async function detectAdobeManagedCdn(domain, log) {
...
+const HEADER_AEM_DEBUG = 'x-aem-debug';
+const HEADER_CORRELATION_ID = 'x-correlation-id';
+const HEADER_EDGEOPTIMIZE_REQUEST_ID = 'x-edgeoptimize-request-id';
```
nit23uec: pls reuse `UA_ROUTING_HEADER`

src/support/cdn-detection.js
```diff
@@ -163,13 +163,137 @@ async function detectAdobeManagedCdn(domain, log) {
...
+const HEADER_REQUEST_ID = 'x-request-id';
+const HEADER_TOKOWAKA_REQUEST_ID = 'x-tokowaka-request-id';
```
nit23uec: we can remove `HEADER_TOKOWAKA_REQUEST_ID`

src/support/cdn-detection.js
```diff
@@ -163,13 +163,137 @@ async function detectAdobeManagedCdn(domain, log) {
...
+const HEADER_USER_AGENT = 'User-Agent';
+
+// Number of sequential probes sent to verify all Case 0 signals.
+// Increase to raise the false-positive bar; decrease to reduce latency.
+export const CASE0_PROBE_COUNT = 3;
+
+async function makeCase0Probe(domain, log) {
```
nit23uec: nitpick: `case0` probe is not readable. suggest a more appropriate name for the use case.

src/support/cdn-detection.js
```diff
@@ -163,13 +163,137 @@ async function detectAdobeManagedCdn(domain, log) {
...
+  const signal2 = allResponses.every(
+    (r) => r[HEADER_TOKOWAKA_REQUEST_ID] || r[HEADER_EDGEOPTIMIZE_REQUEST_ID],
+  );
+  if (!signal2) {
+    log?.info?.('[cdn-detection] Phase 1.5: signal 2 absent — no request-id header', { domain });
+    return null;
+  }
+
+  // Signal 3: x-request-id unique across all probed responses (WAF is not caching)
+  const requestIds = allResponses.map((r) => r[HEADER_REQUEST_ID]);
```
nit23uec: lets wrap this logic (correlation id, request id, and edge optimize request id) to a method - `checkCachingAtProxy`

src/support/cdn-detection.js
nit23uec: The intention of this comment was to reuse thew same constants wherever possible instead of defining new.

nit23uec (approved):

PR #2427, opened 2026-05-18
feat: add trigger handler for weekly scheduled audit dispatch

Sugandhgyl:
Without this, the EventBridge weekly job had no route in the trigger controller and silently dropped every scheduled run. The Slack run-audit command worked because it bypasses the trigger endpoint entirely and sends SQS messages directly per site.

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/controllers/trigger.js
+++ b/src/controllers/trigger.js
@@ -19,6 +19,7 @@ import canonical from './trigger/canonical.js';
 import sitemap from './trigger/sitemap.js';
 import backlinks from './trigger/backlinks.js';
 import organictraffic from './trigger/organictraffic.js';
+import rumConfigRefresh from './trigger/rum-config-refresh.js';
 
 const AUDITS = {
   apex,
@@ -27,6 +28,7 @@ const AUDITS = {
   sitemap,
   'broken-backlinks': backlinks,
   'organic-traffic': organictraffic,
+  'rum-config-refresh': rumConfigRefresh,
 };
 
 /**
--- a/src/controllers/trigger/rum-config-refresh.js
+++ b/src/controllers/trigger/rum-config-refresh.js
@@ -1,0 +1,18 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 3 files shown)


MysticatBot (commented):
Hey @Sugandhgyl,

## Strengths

- Clean, minimal change that follows the established trigger handler pattern exactly - import `triggerFromData`, destructure context, delegate (`src/controllers/trigger/rum-config-refresh.js:15-17`)
- Well-scoped PR that does one thing: fixes the silent EventBridge drop for weekly scheduled runs
- Registration key `'rum-config-refresh'` in the AUDITS map correctly matches the EventBridge event type (`src/controllers/trigger.js:31`)
- Unit test follows the existing convention used by other audit type tests in the same file

## Issues

### Minor (Nice to Have)

**1. Empty options passed to `triggerFromData` omits `deliveryType`** - `src/controllers/trigger/rum-config-refresh.js:17`

Other trigger handlers in this repo typically pass a `deliveryType` filter (e.g., `Site.DELIVERY_TYPES.AEM_EDGE`) as part of the config options to `triggerFromData`. This handler passes `{}`, which means either (a) rum-config-refresh intentionally applies to all delivery types, or (b) the filter was inadvertently omitted. If intentional, a brief comment documenting why would prevent a future contributor from "fixing" it. If unintentional, it should be aligned with sibling handlers.

Unable to verify the body of `triggerFromData` in `./common/trigger.js` to confirm default behavior.

**2. Test only asserts "doesn't throw"** - `test/controllers/trigger.test.js:114`

The test's sole assertion is `to.be.fulfilled`, which verifies the handler resolves but does not confirm `sendMessage` was called or that the correct audit type was dispatched. A stub assertion (e.g., `expect(context.sqs.sendMessage.called).to.be.true`) would catch wiring regressions. That said, the existing tests in this file use the same pattern, so this is a pre-existing weakness being replicated rather than a regression introduced here.

## Assessment

Reasoning: The handler is correctly implemented and registered following established conventions. Both minor findings are non-blocking - one is a documentation/intent question and the other is a pre-existing test pattern. CI is green and Codecov confirms full line coverage. Ship it.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 44s | <b>Cost:</b> $2.15 | <b>Commit:</b> <code>7b65fdacb5febdbe693c7564478e866b1b74b989</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2436, opened 2026-05-19
feat: added Pending review decision

anshulk-public:
## Summary

Adds a `PENDING` review decision to the PLG onboarding waitlist review flow.

When an ESE takes action on a `WAITLISTED` record (e.g. emails the customer),
they can submit `PENDING` via **"Mark as Pending"** to signal to other ESEs
that someone is actively handling it. The status stays `WAITLISTED` and
`reviewedBy` captures who is handling it.

[**jira**
](https://jira.corp.adobe.com/browse/SITES-44715)
## Changes

- `src/controllers/plg/plg-onboarding.js` — added `PENDING` to `allowedDecisions`
  and a handler that saves the review entry without changing status or re-running
  the PLG flow

image

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/controllers/plg/plg-onboarding.js
+++ b/src/controllers/plg/plg-onboarding.js
@@ -1764,17 +1764,17 @@ function PlgOnboardingController(ctx) {
     const updatedReviews = [...existingReviews, reviewEntry];
     onboarding.setReviews(updatedReviews);
 
-    const checkKey = deriveCheckKey(onboarding);
-    if (!checkKey) {
-      return badRequest('Unable to determine the review reason from the onboarding record');
-    }
-
     onboarding.setUpdatedBy(reviewedBy);
 
     // PENDING: record ESE action without changing status (e.g. emailed customer)
     if (decision === REVIEW_DECISIONS.PENDING) {
       await onboarding.save();
       return ok(PlgOnboardingDto.toAdminJSON(onboarding));
+    }
+
+    const checkKey = deriveCheckKey(onboarding);
+    if (!checkKey) {
+      return badRequest('Unable to determine the review reason from the onboarding record');
     }
 
     // UPHOLD: reject the domain — transition to REJECTED (terminal)
--- a/src/controllers/plg/plg-onboarding.js
+++ b/src/controllers/plg/plg-onboarding.js
@@ -1695,6 +1695,8 @@ function PlgOnboardingController(ctx) {
    * - UPHELD: transition to REJECTED (terminal)
    * - CLOSED: retire the current domain to OUTDATED and onboard an alternate domain
    *           (requires siteConfig.alternateDomain for DOMAIN_ALREADY_ASSIGNED reason)
+   * - PENDING: record that an ESE is actively working on this (e.g. emailed customer)
+   *            without changing the status; reviewedBy identifies who is handling it
    * REOPENED and OFFBOARDED are handled by transitionStatus
    * (PATCH /plg/onboard/:onboardingId/status).
    */
@@ -1721,6 +1723,7 @@ function PlgOnboardingController(ctx) {
 
     const allowedDecisions = [
       REVIEW_DECISIONS.BYPASSED, REVIEW_DECISIONS.UPHELD, REVIEW_DECISIONS.CLOSED,
+      REVIEW_DECISIONS.PENDING,
     ];
     if (!hasText(decision) || !allowedDecisions.includes(decision)) {
       return badRequest(`decision must be one of: ${allowedDecisions.join(', ')}`);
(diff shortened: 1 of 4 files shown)


src/controllers/plg/plg-onboarding.js
```diff
@@ -1768,6 +1771,12 @@ function PlgOnboardingController(ctx) {
 
     onboarding.setUpdatedBy(reviewedBy);
 
+    // PENDING: record ESE action without changing status (e.g. emailed customer)
+    if (decision === REVIEW_DECISIONS.PENDING) {
```
Kanishkavijay39, line 1774: **Important - move this block above the `checkKey` guard.**

The `if (!checkKey) return badRequest(...)` check a few lines above fires for ALL decisions including PENDING. `checkKey` is only needed by the BYPASSED switch-case; PENDING never uses it. If a WAITLISTED record somehow lacks a `waitlistReason`, this PENDING request gets a misleading 400 instead of saving.

Move the early return before `deriveCheckKey`:
```js
onboarding.setUpdatedBy(reviewedBy);

if (decision === REVIEW_DECISIONS.PENDING) {
  await onboarding.save();
  return ok(PlgOnboardingDto.toAdminJSON(onboarding));
}

const checkKey = deriveCheckKey(onboarding);
if (!checkKey) {
  return badRequest('Unable to determine the review reason from the onboarding record');
}
```

Kanishkavijay39 (changes requested):
Hey @anshulk-public,

## Strengths

- `src/controllers/plg/plg-onboarding.js`: The PENDING path correctly reuses the full pre-existing validation stack: admin-only access control, WAITLISTED status guard, justification required, and the review entry append logic. No duplication.
- `src/controllers/plg/plg-onboarding.js:1771-1776`: The PENDING handler correctly does NOT call `setStatus`, `setWaitlistReason`, or `postPlgOnboardingNotification`. The status stays WAITLISTED and no external notification fires - exactly right for a "marking as in-progress" action.
- `package-lock.json`: `@adobe/spacecat-shared-data-access` bumped from 3.63.0 to 3.65.0, picking up the PENDING addition to `REVIEW_DECISIONS` from the companion spacecat-shared change. Deployment sequencing is correct.
- `test/controllers/plg/plg-onboarding.test.js:6049-6113`: Both new tests are well-structured - the first verifies `setStatus` and `setWaitlistReason` are not called, the second verifies the existing review is preserved at index 0 and the new one appended at index 1.

## Issues

### Important (Should Fix)

**`src/controllers/plg/plg-onboarding.js` around line 1762: PENDING early return is placed after the `checkKey` guard that PENDING doesn't need**

The current flow runs:

```
checkKey = deriveCheckKey(onboarding)
if (!checkKey) return badRequest(...)   // runs for ALL decisions including PENDING
setUpdatedBy(...)
if (decision === PENDING) { save(); return ok(...) }  // PENDING early return
```

`checkKey` is only needed by the BYPASSED switch-case. PENDING doesn't use it. If a WAITLISTED record lacks a `waitlistReason` (malformed data, manual DB write, or a future code path that creates WAITLISTED without a reason), PENDING fails with `'Unable to determine the review reason from the onboarding record'` when it should succeed. Move the PENDING early return to before the `checkKey` check:

```js
onboarding.setUpdatedBy(reviewedBy);

// PENDING: record ESE action without changing status
if (decision === REVIEW_DECISIONS.PENDING) {
  await onboarding.save();
  return ok(PlgOnboardingDto.toAdminJSON(onboarding));
}

const checkKey = deriveCheckKey(onboarding);
if (!checkKey) {
  return badRequest('Unable to determine the review reason from the onboarding record');
}
```

(See inline comment.)

### Minor (Nice to Have)

**`test/controllers/plg/plg-onboarding.test.js`: Missing test for PENDING when `checkKey` would be absent**

There is no test covering the edge case where `getWaitlistReason()` returns null on a WAITLISTED record and decision is PENDING. Once the early return is moved, add a test with a record that has no `waitlistReason` and confirm PENDING still returns 200. This pins the "PENDING does not need checkKey" contract.

**`package-lock.json`: 12+ unrelated packages bumped alongside the feature change**

The lockfile refresh bundles unrelated version bumps (`helix-shared-body-data`, `mysticat-shared-seo-client`, `spacecat-shared-http-utils`, etc.) with this feature PR. Not a blocker, but consider a separate lockfile-refresh PR next time or a note in the description so reviewers know what to focus on.

## Recommendations

- The JSDoc for `update()` now lists five decision types across two different endpoints. As this grows, consider a brief table or a link to the API spec to keep the inline docs from drifting.

## Assessment

Ready to merge? With fixes

Reasoning: The logic is correct in the happy path, but the `checkKey` guard creates an unnecessary failure mode for PENDING that can be eliminated with a 3-line reorder.

## Next Steps

1. Move the PENDING early return to before the `checkKey` guard.
2. Add a test covering PENDING on a record with no `waitlistReason`.
3. Minor lockfile and test items are optional.

Kanishkavijay39 (approved):
Hey @anshulk-public,

## Strengths

- The PENDING early return is now correctly placed before the `checkKey` guard (`src/controllers/plg/plg-onboarding.js:1769-1773`). PENDING no longer fails on records with no `waitlistReason`.
- New test at `test/controllers/plg/plg-onboarding.test.js:6112` covers PENDING on a record with null `waitlistReason` and pins the "PENDING does not need checkKey" contract.

CI is green. This is good to go.

## Assessment

Ready to merge? Yes

Reasoning: Both flagged issues from the first review have been addressed cleanly. The reorder is minimal and correct, and the test case covers the edge case precisely.

PR #2439, opened 2026-05-19
fix: updated the customer email

anshulk-public:
## Summary

Fixes `createdBy` not being stamped with the customer's identity when they
trigger onboarding from the ASO UI on a pre-existing PLG onboarding record.

Previously, `createdBy` was only updated when the record status was `PRE_ONBOARDING`
or `INACTIVE` — both states where the record was created by the system. This
missed cases where a customer re-onboards a record in any other status (e.g.
`REJECTED`, `WAITLISTED`, `OUTDATED`, `ERROR`) — the customer's identity would
never be stamped, leaving `createdBy` as the old system or stale value.

The fix replaces the status-based condition with a UI origin check using the
`x-client-type: sites-optimizer-ui` header. The intent is: **any UI-initiated
onboarding should always stamp `createdBy` with the caller's identity, regardless
of the existing record's status.** Non-UI callers (e.g. automated pipelines,
API clients) deliberately do not update `createdBy`.

## Changes

- `src/controllers/plg/plg-onboarding.js`
  - Added `isFromAsoUI(context)` helper that checks the `x-client-type` header
  - Replaced the status-based `createdBy` guard with the UI origin check so all
    UI-initiated onboardings stamp `createdBy` regardless of record status

- `test/controllers/plg/plg-onboarding.test.js`
  - Added tests confirming `createdBy` is stamped for WAITLISTED, ERROR, and
    OUTDATED records when the request comes from the ASO UI (documents intent)
  - Added tests confirming `createdBy` is not set when `x-client-type` is a
    different value (strict equality, not just falsy)
  - Added tests confirming defensive behavior when `pathInfo` is absent or
    `headers` are absent (covers optional chaining in `isFromAsoUI`)

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/controllers/plg/plg-onboarding.js
+++ b/src/controllers/plg/plg-onboarding.js
@@ -54,6 +54,10 @@ import { PlgOnboardingDto } from '../../dto/plg-onboarding.js';
 import AccessControlUtil from '../../support/access-control-util.js';
 import { cleanupPlgSiteSuggestionsAndFixes } from './plg-onboarding-cleanup.js';
 
+function isFromAsoUI(context) {
+  return context?.pathInfo?.headers?.['x-client-type'] === 'sites-optimizer-ui';
+}
+
 const { STATUSES, REVIEW_DECISIONS } = PlgOnboardingModel;
 const ASO_PRODUCT_CODE = EntitlementModel.PRODUCT_CODES.ASO;
 const ASO_TIER = EntitlementModel.TIERS.PLG;
@@ -704,7 +708,7 @@ async function performAsoPlgOnboarding({
     }
   }
   onboarding.setUpdatedBy(callerIdentity);
-  if ([STATUSES.PRE_ONBOARDING, STATUSES.INACTIVE].includes(onboarding.getStatus())) {
+  if (isFromAsoUI(context)) {
     onboarding.setCreatedBy(callerIdentity);
   }
 
--- a/test/controllers/plg/plg-onboarding.test.js
+++ b/test/controllers/plg/plg-onboarding.test.js
@@ -1377,6 +1377,66 @@ describe('PlgOnboardingController', function describePlgOnboarding() {
       expect(existingOnboarding.setCreatedBy).to.not.have.been.called;
     });
 
+    it('does not set createdBy when x-client-type is a different value', async () => {
+      const existingOnboarding = createMockOnboarding({ status: 'IN_PROGRESS', createdBy: 'system' });
+      mockDataAccess.PlgOnboarding.findByImsOrgIdAndDomain.resolves(existingOnboarding);
+
+      const context = buildContext(
+        { domain: TEST_DOMAIN },
+        { headers: { 'x-client-type': 'some-other-client' } },
+      );
+
+      const res = await controller.onboard(context);
+
+      expect(res.status).to.equal(200);
+      expect(existingOnboarding.setCreatedBy).to.not.have.been.called;
+    });
+
+    it('sets createdBy when request comes from ASO UI on a WAITLISTED record', async () => {
+      const existingOnboarding = createMockOnboarding({ status: 'WAITLISTED', createdBy: 'system' });
+      mockDataAccess.PlgOnboarding.findByImsOrgIdAndDomain.resolves(existingOnboarding);
+
+      const authInfo = { getProfile: sandbox.stub().returns({ tenants: [{ id: 'ABC123' }], email: 'ese@adobe.com' }) };
+      const context = buildContext(
+        { domain: TEST_DOMAIN },
+        { authInfo, headers: { 'x-client-type': 'sites-optimizer-ui' } },
+      );
+
+      await controller.onboard(context);
+
+      expect(existingOnboarding.setCreatedBy).to.have.been.calledWith('ese@adobe.com');
+    });
+
+    it('sets createdBy when request comes from ASO UI on an ERROR record', async () => {
+      const existingOnboarding = createMockOnboarding({ status: 'ERROR', createdBy: 'system' });
+      mockDataAccess.PlgOnboarding.findByImsOrgIdAndDomain.resolves(existingOnboarding);
+
+      const authInfo = { getProfile: sandbox.stub().returns({ tenants: [{ id: 'ABC123' }], email: 'ese@adobe.com' }) };
+      const context = buildContext(
+        { domain: TEST_DOMAIN },
+        { authInfo, headers: { 'x-client-type': 'sites-optimizer-ui' } },
+      );
+
+      await controller.onboard(context);
+
+      expect(existingOnboarding.setCreatedBy).to.have.been.calledWith('ese@adobe.com');
+    });
+
+    it('sets createdBy when request comes from ASO UI on an OUTDATED record', async () => {
+      const existingOnboarding = createMockOnboarding({ status: 'OUTDATED', createdBy: 'system' });
+      mockDataAccess.PlgOnboarding.findByImsOrgIdAndDomain.resolves(existingOnboarding);
+
+      const authInfo = { getProfile: sandbox.stub().returns({ tenants: [{ id: 'ABC123' }], email: 'ese@adobe.com' }) };
+      const context = buildContext(
+        { domain: TEST_DOMAIN },
+        { authInfo, headers: { 'x-client-type': 'sites-optimizer-ui' } },
+      );
+
+      await controller.onboard(context);
+
+      expect(existingOnboarding.setCreatedBy).to.have.been.calledWith('ese@adobe.com');
+    });
+
     it('resumes existing onboarding record for same imsOrgId+domain', async () => {
       const existingOnboarding = createMockOnboarding({ status: 'ERROR' });
       mockDataAccess.PlgOnboarding.findByImsOrgIdAndDomain.resolves(existingOnboarding);
@@ -3254,6 +3314,35 @@ describe('PlgOnboardingController', function describePlgOnboarding() {
       mockOnboarding.getStatus.returns('PRE_ONBOARDING');
       const authInfo = { getProfile: sandbox.stub().returns({ tenants: [{ id: 'AAAAAAAABBBBBBBBCCCCCCCC' }], email: 'customer@example.com' }) };
       const context = buildContext({ domain: TEST_DOMAIN }, { authInfo });
+      const res = await controller.onboard(context);
+      expect(res.status).to.equal(200);
+      expect(mockOnboarding.setCreatedBy).to.not.have.been.called;
+    });
+
+    it('does not set createdBy when x-client-type is a different value', async () => {
+      mockOnboarding.getStatus.returns('PRE_ONBOARDING');
+      const authInfo = { getProfile: sandbox.stub().returns({ tenants: [{ id: 'AAAAAAAABBBBBBBBCCCCCCCC' }], email: 'customer@example.com' }) };
+      const context = buildContext({ domain: TEST_DOMAIN }, { authInfo, headers: { 'x-client-type': 'some-other-client' } });
+      const res = await controller.onboard(context);
+      expect(res.status).to.equal(200);
+      expect(mockOnboarding.setCreatedBy).to.not.have.been.called;
+    });
+
+    it('does not set createdBy when pathInfo is absent', async () => {
+      mockOnboarding.getStatus.returns('PRE_ONBOARDING');
+      const authInfo = { getProfile: sandbox.stub().returns({ tenants: [{ id: 'AAAAAAAABBBBBBBBCCCCCCCC' }], email: 'customer@example.com' }) };
+      const context = buildContext({ domain: TEST_DOMAIN }, { authInfo });
+      delete context.pathInfo;
+      const res = await controller.onboard(context);
+      expect(res.status).to.equal(200);
+      expect(mockOnboarding.setCreatedBy).to.not.have.been.called;
+    });
+
+    it('does not set createdBy when headers are absent', async () => {
+      mockOnboarding.getStatus.returns('PRE_ONBOARDING');
+      const authInfo = { getProfile: sandbox.stub().returns({ tenants: [{ id: 'AAAAAAAABBBBBBBBCCCCCCCC' }], email: 'customer@example.com' }) };
+      const context = buildContext({ domain: TEST_DOMAIN }, { authInfo });
+      context.pathInfo = {};
       const res = await controller.onboard(context);
       expect(res.status).to.equal(200);
       expect(mockOnboarding.setCreatedBy).to.not.have.been.called;
--- a/test/controllers/plg/plg-onboarding.test.js
+++ b/test/controllers/plg/plg-onboarding.test.js
@@ -450,7 +450,7 @@ describe('PlgOnboardingController', function describePlgOnboarding() {
     };
   }
 
-  function buildContext(data = {}, { authInfo } = {}) {
+  function buildContext(data = {}, { authInfo, headers } = {}) {
     return {
       data,
       dataAccess: mockDataAccess,
@@ -460,6 +460,7 @@ describe('PlgOnboardingController', function describePlgOnboarding() {
       attributes: {
         authInfo: authInfo !== undefined ? authInfo : mockAuthInfo(),
       },
+      pathInfo: { headers: headers || {} },
     };
   }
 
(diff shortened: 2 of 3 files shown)


src/controllers/plg/plg-onboarding.js
```diff
@@ -54,6 +54,10 @@ import { PlgOnboardingDto } from '../../dto/plg-onboarding.js';
 import AccessControlUtil from '../../support/access-control-util.js';
 import { cleanupPlgSiteSuggestionsAndFixes } from './plg-onboarding-cleanup.js';
 
+function isFromAsoUI(context) {
+  return context?.pathInfo?.headers?.['x-client-type'] === 'sites-optimizer-ui';
```
Kanishkavijay39, line 58: **Spoofable header + third copy of the helper**

`x-client-type` is client-supplied - any authenticated caller (API key holder, internal service, admin acting cross-org) can send this header and trigger the `createdBy` overwrite. There is no Fastly/middleware layer that strips or enforces it. This is not privilege escalation (the identity written is still the authenticated principal), but it is an audit-trail integrity issue: combined with the broadened status scope at line 711, any authenticated user can overwrite `createdBy` on any status record.

Also, `src/support/utils.js` already has two `x-client-type` helpers: `isViewAsTrialRequest` at line 606 and `getIsSummitPlgEnabled` at line 624. This is the third copy. Please extract a shared `isFromAsoUI(context)` and `ASO_UI_CLIENT_TYPE` constant to `src/support/utils.js` so the header string cannot drift across three call sites.

A better alternative that avoids the header entirely:
```javascript
const existingCreatedBy = onboarding.getCreatedBy?.();
if (!hasText(existingCreatedBy) || existingCreatedBy === 'system') {
  onboarding.setCreatedBy(callerIdentity);
}
```
This expresses the actual invariant ("stamp the first human owner of a system-created record") without trusting a client-supplied header.

src/controllers/plg/plg-onboarding.js
```diff
@@ -704,7 +708,7 @@ async function performAsoPlgOnboarding({
     }
   }
   onboarding.setUpdatedBy(callerIdentity);
-  if ([STATUSES.PRE_ONBOARDING, STATUSES.INACTIVE].includes(onboarding.getStatus())) {
+  if (isFromAsoUI(context)) {
```
Kanishkavijay39, line 711: **Guard is broader than the stated bug**

The old condition `[STATUSES.PRE_ONBOARDING, STATUSES.INACTIVE].includes(status)` meant "this record was never owned by a real customer." The new condition means "the request came from the UI" - these are not equivalent.

Two regressions:
1. **UI callers**: Records in `IN_PROGRESS`, `ERROR`, `WAITLISTED`, `WAITING_FOR_IP_ALLOWLISTING`, and `OUTDATED` status will now have `createdBy` overwritten on any UI re-onboard. `createdBy` is surfaced in the operator UI and used for IMS profile lookup (~line 1629) - silent mutation here is a provenance regression.
2. **Non-UI callers**: Slack commands or scripts calling this function on a `PRE_ONBOARDING`/`INACTIVE` record will no longer stamp `createdBy`, since they won't send the header. The PR description says this is purely additive but it is also subtractive.

Minimal fix for the stated bug (REJECTED records):
```javascript
if ([STATUSES.PRE_ONBOARDING, STATUSES.INACTIVE, STATUSES.REJECTED].includes(onboarding.getStatus())) {
  onboarding.setCreatedBy(callerIdentity);
}
```
If broader behavior is intentional, please add tests for `WAITLISTED`, `ERROR`, and `OUTDATED` + UI header, and document the intent in a code comment.

test/controllers/plg/plg-onboarding.test.js
```diff
@@ -3209,29 +3238,24 @@ describe('PlgOnboardingController', function describePlgOnboarding() {
       expect(mockOnboarding.setStatus).to.have.been.calledWith('ONBOARDED');
     });
 
-    it('stamps customer identity as createdBy when existing record is PRE_ONBOARDING', async () => {
+    it('stamps customer identity as createdBy when request comes from ASO UI', async () => {
```
Kanishkavijay39, line 3301: **Test sets irrelevant status**

`mockOnboarding.getStatus.returns('PRE_ONBOARDING')` is now irrelevant to the condition being tested - `isFromAsoUI` does not check status. This test passes for the wrong reason. Consider using `REJECTED` status here instead, which locks in the actual behavior change this PR ships and would have caught the original bug in a regression.

Kanishkavijay39 (changes requested):
Hey @anshulk-public,

## Strengths

- `src/controllers/plg/plg-onboarding.js:57-59` - `isFromAsoUI` is clean, well-named, and properly isolated as a module-level function.
- `test/controllers/plg/plg-onboarding.test.js:453-456` - Updating `buildContext` to accept a `headers` option is a good, non-breaking extension.
- Both directions of the new branch (header present, header absent) are explicitly tested.
- The underlying bug is real and worth fixing.

## Issues

### Important (Should Fix)

**1. `src/controllers/plg/plg-onboarding.js:711` - Fix removes a meaningful guard and is broader than the stated bug**

See inline comment. The minimal correct fix for the stated bug is to add `STATUSES.REJECTED` to the existing list:
```javascript
if ([STATUSES.PRE_ONBOARDING, STATUSES.INACTIVE, STATUSES.REJECTED].includes(onboarding.getStatus())) {
  onboarding.setCreatedBy(callerIdentity);
}
```
If the intent is genuinely "UI calls should always re-stamp createdBy," document it explicitly and add tests for `WAITLISTED`, `ERROR`, and `OUTDATED` records confirming the overwrite is intentional.

**2. `src/controllers/plg/plg-onboarding.js:57-59` - `x-client-type` is a spoofable client-supplied header, and this is the third copy of the helper**

See inline comment. The most robust fix anchors the decision in the existing record value rather than the request header - only overwrite `createdBy` when the current value is a known system identity:
```javascript
const existingCreatedBy = onboarding.getCreatedBy?.();
const isSystemOwned = !hasText(existingCreatedBy) || existingCreatedBy === 'system';
if (isSystemOwned) {
  onboarding.setCreatedBy(callerIdentity);
}
```
This expresses the actual invariant ("stamp the first human owner of a system-created record"), is not spoofable, and survives client-type renames.

**3. `test/controllers/plg/plg-onboarding.test.js` - Missing test for ONBOARDED record + ASO UI header**

The code at line 679 returns early when `status === ONBOARDED`, so the `setCreatedBy` path is never reached for onboarded records. But there is no test asserting this. Without it, if the early-return guard is ever refactored, the regression will be invisible. Add a test asserting `setCreatedBy` is not called when the record is ONBOARDED, even with the ASO UI header present.

### Minor (Nice to Have)

**4. PR title has a typo and will land in the changelog** - "udpated" should be "updated". This repo uses conventional commits for semantic-release. Recommend squashing with `fix(plg): stamp createdBy on UI-initiated onboarding regardless of record status` before merge.

**5. `test/controllers/plg/plg-onboarding.test.js:3241` - Test sets irrelevant status** - See inline comment.

**6. No test for `pathInfo` absent or `headers` absent** - The optional chaining in `isFromAsoUI` implies `pathInfo` may be absent in some call paths. `buildContext` always supplies `pathInfo: { headers: {} }`, so neither `pathInfo: undefined` nor `pathInfo: {}` (no `headers`) is exercised. Two one-liners would lock in the defensive behavior.

**7. Negative test only covers absent header** - The "does not set createdBy when request does not come from ASO UI" test passes an empty headers object. Add a case with `x-client-type: some-other-client` to confirm the check is strict equality, not just falsy.

## Recommendations

- Add a `log.info` line when `setCreatedBy` is called on an existing record, logging the previous and new values along with the record ID and status. `createdBy` is the only audit field for original ownership; once overwritten it is unrecoverable without a DB backup.
- If `createdBy` semantically means "who originally created this record," consider making it immutable after initial assignment. A separate `lastOnboardedBy` or `claimedBy` field better captures the "re-onboarding customer" use case without conflating creation provenance with activity recency.

## Assessment

Ready to merge? No - with fixes
Reasoning: The core fix is correct but the implementation broadens the semantics of `createdBy` mutation beyond what the PR description states, replaces a data-derived guard with a spoofable header check, and is missing coverage for the riskiest case. Findings 1 and 2 can be resolved together by anchoring the guard in the existing `createdBy` value rather than the request header, which is both simpler and more correct.

## Next Steps

1. Address findings 1 and 2 together - either add `REJECTED` to the status list (minimal fix) or anchor on the existing `createdBy` value being system/empty (better fix).
2. Centralize the `x-client-type` helper in `src/support/utils.js` alongside the existing `isViewAsTrialRequest` and `getIsSummitPlgEnabled` helpers.
3. Add the ONBOARDED + UI header test (finding 3).
4. Fix the PR title before merge (finding 4).
5. Findings 5-7 are small test improvements; address in the same pass.

anshulk-public:
@Kanishkavijay39 
Thanks for the review! Here's the status of each issue:

**Issue #1 (Fix removes a meaningful guard and is broader than the stated bug)** ✅ Addressed with tests

The intent is genuinely "UI calls should always re-stamp `createdBy` regardless of record status." A WAITLISTED, ERROR, or OUTDATED record re-onboarded via ASO UI should have `createdBy` updated — not just `PRE_ONBOARDING` and `INACTIVE`. Added explicit tests for WAITLISTED, ERROR, and OUTDATED records confirming the overwrite is intentional.

**Issue #2 (x-client-type is spoofable, and this is the third copy of the helper)** ❌ Not changed

The `isFromAsoUI` helper in `plg-onboarding.js` is not a duplicate — `isViewAsTrialRequest` in `utils.js` checks both `x-client-type` AND `x-view-as-trial` together and is a different use case. The `isSystemOwned` alternative also changes the semantics: it would stamp `createdBy` on non-UI calls too if the record was system-created, which is not the intent here. The header-based check is consistent with how `x-client-type` is used elsewhere in the codebase and expresses the correct invariant for this feature.

**Issue #3 (Missing test for ONBOARDED record + ASO UI header)** ❌ Not added

The early return for `ONBOARDED` is intentional — it is a terminal state and we want that behaviour. Adding a test for it would just be testing an already-correct guard with no regression value.

**Issue #4 (PR title typo)** — fixed.

**Issue #5 (Test sets irrelevant status)** ❌ Skipped — minor and low impact.

**Issue #6 (No test for pathInfo absent or headers absent)** ✅ Resolved
Added two tests: one with `pathInfo` deleted entirely and one with `pathInfo: {}` (no headers). Both confirm `setCreatedBy` is not called and the optional chaining behaves correctly.

**Issue #7 (Negative test only covers absent header)** ✅ Resolved
Added a test with `x-client-type: some-other-client` in both the existing record section and the PRE_ONBOARDING section, confirming the check is strict equality not just falsy.

src/controllers/plg/plg-onboarding.js
```diff
@@ -54,6 +54,10 @@ import { PlgOnboardingDto } from '../../dto/plg-onboarding.js';
 import AccessControlUtil from '../../support/access-control-util.js';
 import { cleanupPlgSiteSuggestionsAndFixes } from './plg-onboarding-cleanup.js';
 
+function isFromAsoUI(context) {
```
Kanishkavijay39, line 57: **Finding 2 - Spoofable header gates a provenance write**

`x-client-type` is a client-supplied header. Any authenticated caller can set it and trigger a `createdBy` overwrite on a record they did not originally create. Since `createdBy` is the sole ownership audit trail, once overwritten it is unrecoverable.

Robust fix - anchor on existing record state instead:
```javascript
const existingCreatedBy = onboarding.getCreatedBy?.();
const isSystemOwned = !hasText(existingCreatedBy) || existingCreatedBy === 'system';
if (isSystemOwned) {
  onboarding.setCreatedBy(callerIdentity);
}
```
Minimum ask if the header check stays: add a comment here stating that this is a client-asserted signal and that the endpoint relies on upstream auth to enforce who can call it.

src/controllers/plg/plg-onboarding.js
Kanishkavijay39, line 711: **Finding 1 - Behavioral widening is now codified by tests without a written rationale**

The new tests assert WAITLISTED, ERROR, and OUTDATED records all get `createdBy` overwritten when the ASO UI header is present. The stated bug was about REJECTED records only. Future maintainers reading this guard will treat "any ASO UI call re-stamps createdBy" as the designed intent.

Fix option A (minimal): narrow back to a status whitelist and add REJECTED:
```javascript
if ([STATUSES.PRE_ONBOARDING, STATUSES.INACTIVE, STATUSES.REJECTED].includes(onboarding.getStatus())) {
  onboarding.setCreatedBy(callerIdentity);
}
```
Fix option B (if the broader behavior is intentional): keep `isFromAsoUI`, add a one-line comment here naming the policy and linking the Jira ticket.

Kanishkavijay39 (changes requested):
Hey @anshulk-public,

## Strengths

- New tests in `test/controllers/plg/plg-onboarding.test.js` explicitly cover the broadened `isFromAsoUI` behavior across WAITLISTED, ERROR, and OUTDATED statuses and lock in current intent.
- Defensive coverage is now complete: missing `pathInfo`, missing `headers`, and a non-matching `x-client-type` value are all exercised in both describe blocks.
- Test style matches existing file conventions (`buildContext`, `createMockOnboarding`, sinon stubs) - no new scaffolding introduced.
- Both controller entry points (~line 1380 and ~line 3322) are now covered, preventing the two call sites from drifting silently.
- Previously flagged items now resolved: PR title typo (Minor 4), `pathInfo`/`headers` absent tests (Minor 6), strict-equality negative test (Minor 7).

## Issues

### Important (Should Fix)

**1. `src/controllers/plg/plg-onboarding.js:711` - Behavioral widening is now codified by tests without a written rationale**

The new tests assert that WAITLISTED, ERROR, and OUTDATED records all get `createdBy` overwritten when the ASO UI header is present. The stated bug was about REJECTED records only. The tests now make "any ASO UI call re-stamps createdBy on any non-terminal status" the permanent contract - future maintainers will read this as intended design.

Fix option A (minimal): narrow the guard back to a status whitelist and add REJECTED:
```javascript
if ([STATUSES.PRE_ONBOARDING, STATUSES.INACTIVE, STATUSES.REJECTED].includes(onboarding.getStatus())) {
  onboarding.setCreatedBy(callerIdentity);
}
```
Fix option B (if the broader behavior is intentional): keep `isFromAsoUI`, add a one-line comment naming the policy and the Jira ticket so it reads as a decision, not an accident.

**2. `src/controllers/plg/plg-onboarding.js:57-59` - `x-client-type` is a client-supplied header that any authenticated caller can spoof**

Any tenant with valid credentials can send `x-client-type: sites-optimizer-ui` and cause `createdBy` to be overwritten with their identity on a record they did not originally create. `createdBy` is the sole provenance field - once overwritten it is unrecoverable without a backup.

Robust fix - anchor the decision in the existing record state:
```javascript
const existingCreatedBy = onboarding.getCreatedBy?.();
const isSystemOwned = !hasText(existingCreatedBy) || existingCreatedBy === 'system';
if (isSystemOwned) {
  onboarding.setCreatedBy(callerIdentity);
}
```
Minimum ask if the header check stays: add a comment at line 57 stating that this is a client-asserted signal and that the endpoint relies on upstream auth to enforce who can call it.

**3. `test/controllers/plg/plg-onboarding.test.js` - Missing test for ONBOARDED record + ASO UI header**

The early return at `plg-onboarding.js:679` (`status === ONBOARDED`) means the `setCreatedBy` path is never reached for already-onboarded records. There is still no test asserting this. If a future refactor removes or reorders that early return, `createdBy` will start being overwritten on terminal records with no test catching it.

Add (~10 lines, near line 1394):
```javascript
it('does not set createdBy when record is already ONBOARDED, even with ASO UI header', async () => {
  const existingOnboarding = createMockOnboarding({ status: 'ONBOARDED', createdBy: 'system' });
  mockDataAccess.PlgOnboarding.findByImsOrgIdAndDomain.resolves(existingOnboarding);
  const authInfo = { getProfile: sandbox.stub().returns({ tenants: [{ id: 'ABC123' }], email: 'ese@adobe.com' }) };
  const context = buildContext({ domain: TEST_DOMAIN }, { authInfo, headers: { 'x-client-type': 'sites-optimizer-ui' } });
  await controller.onboard(context);
  expect(existingOnboarding.setCreatedBy).to.not.have.been.called;
});
```

### Minor (Nice to Have)

**4. `test/controllers/plg/plg-onboarding.test.js` - Duplicate test names across describe blocks**

Two tests share the title `'does not set createdBy when x-client-type is a different value'` in different describe blocks. Suffix one with context (e.g., `'... (existing record flow)'` vs `'... (aso flow)'`) so failure output pinpoints the right block.

**5. New positive-path tests omit the status assertion**

The three new positive tests (WAITLISTED/ERROR/OUTDATED) skip the `expect(res.status).to.equal(200)` assertion used in adjacent tests. Small addition for consistency.

## Recommendations

- Add the ONBOARDED test (Important 3) before merge - single remaining coverage gap, no production code change needed.
- Regardless of which fix path is chosen for Important 1, document the `createdBy` re-stamp policy in a code comment and in the PR description. The new tests make the behavior permanent; the rationale should be equally permanent.
- Long-term: consider moving `createdBy` mutation through a single chokepoint (service method or model hook). Two call sites already exist in this file; divergence across them is a matter of time.

## Assessment

Ready to merge? No - with fixes
Reasoning: The new test commit is well-scoped and meaningfully improves coverage of the defensive paths. However, all three Important findings from the prior review remain open at the source level. The ONBOARDED + ASO UI header test (finding 3) is a 10-line addition with no production code change and should be added before merge. Findings 1 and 2 require either a source change or explicit documentation of the deliberate decision - adding tests that encode the broader behavior without a written rationale is not a substitute.

## Next Steps

1. Add the ONBOARDED + ASO UI header negative test (finding 3) - no production code change, highest priority.
2. Address finding 1: either narrow the guard to the intended status set or add a comment + PR description update documenting the broadened contract as intentional.
3. Address finding 2: either anchor on existing record state or add a comment documenting the trust model.
4. Minor items 4 and 5 are optional but easy to do in the same pass.

Kanishkavijay39 (approved):

PR #2440, opened 2026-05-19
feat: ai vis prompt response | LLMO-4959

dmitriigulkin-semrush:

## Related Issues
[LLMO-4959](https://jira.corp.adobe.com/browse/LLMO-4959)

### DIFF
--- a/docs/openapi/ai-visibility-api.yaml
+++ b/docs/openapi/ai-visibility-api.yaml
@@ -1,4 +1,4 @@
-# AI Visibility (Semrush gRPC proxy) – 26 GET endpoints
+# AI Visibility (Semrush gRPC proxy) – 25 GET endpoints
 # All endpoints live under /llmo/ai-visibility/* and proxy to Semrush gRPC.
 # Responses include an srFilters block with available markets and models.
 
--- a/docs/openapi/ai-visibility-api.yaml
+++ b/docs/openapi/ai-visibility-api.yaml
@@ -1,4 +1,4 @@
-# AI Visibility (Semrush gRPC proxy) – 24 GET endpoints
+# AI Visibility (Semrush gRPC proxy) – 26 GET endpoints
 # All endpoints live under /llmo/ai-visibility/* and proxy to Semrush gRPC.
 # Responses include an srFilters block with available markets and models.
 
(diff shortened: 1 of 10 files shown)


src/support/ai-visibility/handlers/v1/prompt/prompt-response.js
```diff
@@ -0,0 +1,75 @@
...
+  responseFromGrpcError,
+} from '../../../grpc-utils.js';
+
+/** @type {import('@bufbuild/protobuf').JsonReadOptions} */
+const FROM_JSON = { ignoreUnknownFields: true };
+
+/** @type {import('@bufbuild/protobuf').JsonWriteOptions} */
+const TO_JSON = { useProtoFieldName: false, alwaysEmitImplicit: true };
+
+/* c8 ignore start */
```
calvarezg, line 31: `promptHash` is marked `required: true` in the OpenAPI spec (with `pattern: '^\d+$'`), but the handler does no validation — if the param is missing or malformed it falls through to Semrush and the error depends on whatever gRPC response comes back.

Worth adding an explicit guard, e.g.:

```js
const promptHash = sp.get('promptHash');
if (!promptHash) {
  return { status: 400, body: { message: 'promptHash is required' } };
}
```

docs/openapi/ai-visibility-api.yaml
```diff
@@ -1,4 +1,4 @@
-# AI Visibility (Semrush gRPC proxy) – 24 GET endpoints
+# AI Visibility (Semrush gRPC proxy) – 26 GET endpoints
```
calvarezg: This PR adds 1 new endpoint, so the count should be **25**, not 26. The controller test correctly updates to `lengthOf(25)` — the comment here is off by one.

dmitriigulkin-semrush: fixed, thanks!

src/support/ai-visibility/handlers/v1/prompt/prompt-response.js
dmitriigulkin-semrush: good catch! i wouldn't like to add explicit validation on our side tho, here's the error raised even before we send the request, namely in 
```js
const requestBody = fromJson(
      PromptRequestSchema,
      {
        country,
        llm: engine,
        prompt_hash: promptHash,
        serp_id: serpId,
        topic_id: topicId,
      },
      FROM_JSON,
    );
```
i've added try catch there, so any time there's a exception with parsing json to proto message the client will receive a 400 with dedicated error message. later on we'll be able to use `buf.validate`, so it won't look as sketchy

calvarezg (approved):

PR #2441, opened 2026-05-19
feat(llmo): align agentic-traffic export with s3-export-framework ADR

akshaymagapu:
Producer-side changes for the [s3-export-framework ADR](https://github.com/adobe/mysticat-architecture/blob/main/platform/decisions/s3-export-framework.md), plus Option A (S3 Transfer Acceleration) for customer-facing URL hygiene.

## What changes

- **Hash**: `stableStringify` → RFC 8785 JCS; canonical payload now includes `kind`, `v`, `c`, `format`
- **S3 key path**: `v1` → `v1c1`
- **SQS envelope**: drops `metadataKey` (worker derives from `s3Key`)
- **POST**: writes `metadata.processing` via S3 `IfNoneMatch:*` (CAS) before enqueueing; concurrent POSTs can't double-enqueue; success metadata is protected from clobber
- **GET**: unknown `exportId` → 404; returns ADR-shaped `failureReason` enum + `failureMessage` detail
- **Customer URLs**: sign against `s3-accelerate.amazonaws.com` (hides AWS region)
- **Fast path**: read `metadata.files[]` directly; skip `ListObjectsV2` when present (with prefix validation as defense-in-depth)
- **Filter sanitization**: non-string/oversized/malformed-date values coerced to null before hashing

## Cache impact

All existing cached exports become invalid (new hash + new path). Next POST after deploy re-enqueues. **Accepted per team call** — feature not yet broadly used.

## Ordering

1. [spacecat-infrastructure#533](https://github.com/adobe/spacecat-infrastructure/pull/533) — Transfer Acceleration on the reports bucket (must land first)
2. This PR — producer changes
3. [spacecat-reporting-worker#619](https://github.com/adobe/spacecat-reporting-worker/pull/619) — worker changes (after this PR deploys, worker accepts only `v1c1`)

## Test plan

- [x] 117 controller tests pass (new: golden hash, accelerated client, fast-path prefix validation, CAS protect success, JCS primitives, files[] fallback, corrupt-metadata heal)
- [x] 10277 total tests pass
- [x] Coverage: 100% lines/statements/functions, 98.11% branches
- [ ] Smoke test in dev: POST → GET → download URL host is `*.s3-accelerate.amazonaws.com`

### DIFF
--- a/src/controllers/llmo/llmo-agentic-traffic.js
+++ b/src/controllers/llmo/llmo-agentic-traffic.js
@@ -231,10 +231,17 @@ function buildExportKeys(siteId, exportId) {
 // Defense in depth: the worker writes `metadata.files[]` and the producer signs
 // presigned URLs for each entry. A worker bug or compromise could put arbitrary
 // keys in `files[]`; this filter rejects anything outside the deterministic
-// export prefix, falling back to ListObjectsV2 which already constrains shape.
+// export prefix AND anything that isn't a CSV file (the ListObjectsV2 fallback
+// already enforces `urls.csv | urls.csv_partN` — match it on the fast path).
 function validateFilesAgainstPrefix(files, siteId, exportId) {
-  const expectedPrefix = `${buildExportPrefix(siteId, exportId)}/`;
-  return files.filter((k) => typeof k === 'string' && k.startsWith(expectedPrefix));
+  const prefix = `${buildExportPrefix(siteId, exportId)}/`;
+  const allowedFile = /^urls\.csv(?:_part\d+)?$/;
+  return files.filter((k) => {
+    if (typeof k !== 'string' || !k.startsWith(prefix)) {
+      return false;
+    }
+    return allowedFile.test(k.slice(prefix.length));
+  });
 }
 
 // Caps defend status-polling against a pathological prefix (malicious
@@ -321,6 +328,15 @@ async function claimProcessingMetadata(ctx, bucket, metadataKey, exportId, siteI
   }
 }
 
+// Rolls back a claimProcessingMetadata write when downstream enqueue fails.
+// Without this a transient SQS failure would leave `processing` metadata
+// behind and block every subsequent POST until the stale window elapses.
+async function deleteProcessingMetadata(ctx, bucket, metadataKey) {
+  const { s3 } = ctx;
+  const command = new s3.DeleteObjectCommand({ Bucket: bucket, Key: metadataKey });
+  await s3.s3Client.send(command);
+}
+
 async function getExportMetadata(ctx, bucket, metadataKey) {
   const { s3 } = ctx;
   let body;
@@ -350,10 +366,17 @@ const MAX_EXPORT_DOWNLOAD_URLS = 50;
 
 // Customer-facing presigned URLs sign against the s3-accelerate endpoint so
 // the AWS region is hidden from the URL. Bucket-side acceleration is enabled
-// in spacecat-infrastructure. Cached per-region so a multi-region runtime
-// can't accidentally sign with the wrong client.
+// in spacecat-infrastructure.
+//
+// When AWS_ENDPOINT_URL_S3 is set (local IT against MinIO/LocalStack) the
+// accelerate endpoint doesn't exist — fall back to the shared regional
+// client. Cached per-region so a multi-region runtime can't accidentally
+// sign with the wrong client.
 const acceleratedS3Clients = new Map();
-function getAcceleratedS3Client(region) {
+function getSigningClient(ctx, region) {
+  if (ctx.env?.AWS_ENDPOINT_URL_S3) {
+    return ctx.s3.s3Client;
+  }
   if (!acceleratedS3Clients.has(region)) {
     acceleratedS3Clients.set(region, new S3Client({ region, useAccelerateEndpoint: true }));
   }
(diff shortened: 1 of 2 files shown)


src/controllers/llmo/llmo-agentic-traffic.js
```diff
@@ -739,8 +845,23 @@ export function createAgenticTrafficUrlsExportHandler(getSiteAndValidateAccess)
...
           }
 
-          // Failed / stale-processing / no-metadata → re-enqueue. The worker
-          // overwrites metadata on retry; GET reports `failed` to the UI.
+          // Claim the slot with CAS (IfNoneMatch:*) so concurrent POSTs can't
+          // double-enqueue, and don't clobber `success` metadata if CSVs were
+          // evicted (let CAS fail → 202 processing without destroying state).
+          // Stale/failed retries overwrite via casOnly=false.
+          const casOnly = metadata === null || metadata.status === 'success';
+          const claimed = await claimProcessingMetadata(
```
ctinp, line 788: **Failed SQS enqueue leaves the export stuck in `processing` for 30 minutes**

The new claim-before-enqueue flow writes `metadata.json` first and only then sends the SQS message. If `sendMessage()` fails here, we return 500 but leave fresh `processing` metadata behind. Every subsequent POST will hit the `isExportProcessing(metadata) && !isExportStaleProcessing(metadata)` branch and refuse to enqueue again until the stale timeout elapses, so a transient SQS failure turns into a 30-minute outage for that export ID. We need a rollback/update path for the metadata on enqueue failure, or a different ordering/claim mechanism that does not publish `processing` until the job is actually queued.

src/controllers/llmo/llmo-agentic-traffic.js
```diff
@@ -171,34 +178,65 @@ function canonicalizeExportPayload(siteId, parsed) {
...
 
+// Defense in depth: the worker writes `metadata.files[]` and the producer signs
+// presigned URLs for each entry. A worker bug or compromise could put arbitrary
+// keys in `files[]`; this filter rejects anything outside the deterministic
+// export prefix, falling back to ListObjectsV2 which already constrains shape.
+function validateFilesAgainstPrefix(files, siteId, exportId) {
+  const expectedPrefix = `${buildExportPrefix(siteId, exportId)}/`;
+  return files.filter((k) => typeof k === 'string' && k.startsWith(expectedPrefix));
+}
+
```
ctinp, line 205: **Fast-path `files[]` validation now signs non-CSV objects under the prefix**

`validateFilesAgainstPrefix()` only checks that each entry starts with the export prefix. On the fast path we then sign every surviving key directly, unlike the fallback `ListObjectsV2` path which still constrains results to `urls.csv` and `urls.csv_partN`. That means a worker bug could add `metadata.json` or any other in-prefix object to `files[]`, and the API would return download URLs for it. The defense-in-depth check needs to preserve the old shape restriction as well, not just the prefix restriction.

src/controllers/llmo/llmo-agentic-traffic.js
```diff
@@ -247,32 +285,89 @@ async function listExportCsvObjects(ctx, bucket, csvKey) {
...
+// in spacecat-infrastructure. Cached per-region so a multi-region runtime
+// can't accidentally sign with the wrong client.
+const acceleratedS3Clients = new Map();
+function getAcceleratedS3Client(region) {
+  if (!acceleratedS3Clients.has(region)) {
+    acceleratedS3Clients.set(region, new S3Client({ region, useAccelerateEndpoint: true }));
+  }
+  return acceleratedS3Clients.get(region);
+}
+
```
ctinp, line 330: **Accelerated signing client drops the shared S3 wrapper configuration**

This creates a brand-new `S3Client` with only `{ region, useAccelerateEndpoint: true }` instead of deriving it from the existing `ctx.s3.s3Client` setup. In this repo the shared wrapper also injects non-default config like `AWS_ENDPOINT_URL_S3` and `forcePathStyle` for local/IT environments, so the presigner path here no longer matches the rest of the application. As written, export URL generation will bypass those settings and can break non-AWS S3-compatible environments even though the normal S3 wrapper is configured correctly.

src/controllers/llmo/llmo-agentic-traffic.js
akshaymagapu: Fixed in [4239d4e4](https://github.com/adobe/spacecat-api-service/pull/2441/commits/4239d4e4) — wrapped `sqs.sendMessage` in try/catch; on failure, `DeleteObject` the processing metadata before rethrowing. Test added: `rolls back processing metadata when SQS sendMessage fails`. Thanks for catching this.

src/controllers/llmo/llmo-agentic-traffic.js
akshaymagapu: Fixed in [4239d4e4](https://github.com/adobe/spacecat-api-service/pull/2441/commits/4239d4e4). `validateFilesAgainstPrefix` now also enforces the filename shape (`urls.csv(_part\d+)?`), matching what the ListObjectsV2 fallback already does. An in-prefix non-CSV entry (e.g. `metadata.json`) is rejected and the handler falls back to ListObjectsV2. Test added: `falls back to ListObjectsV2 when metadata.files contains in-prefix non-CSV keys`.

src/controllers/llmo/llmo-agentic-traffic.js
akshaymagapu: Fixed in [4239d4e4](https://github.com/adobe/spacecat-api-service/pull/2441/commits/4239d4e4). When `AWS_ENDPOINT_URL_S3` is set (IT against MinIO/LocalStack where the accelerate endpoint doesn't exist) we fall back to `ctx.s3.s3Client` (the shared regional client with the endpoint override). Test added: `signs with the regular S3 client (not accelerated) when AWS_ENDPOINT_URL_S3 is set`.

ctinp (approved):

PR #2443, opened 2026-05-19
fix: use www.Url if apex url is not resolved

tkotthakota-adobe:
https://jira.corp.adobe.com/browse/SITES-44465


tkotthakota-adobe:
duplicate.

PR #2445, opened 2026-05-20
fix(SITES-44243): server-derive executedBy and add executedByUser audit trail

sandsinh:
## Summary
- **Security fix (SITES-44243):** closes a PII exfiltration path where any org-admin could PATCH/POST a fix with an arbitrary `executedBy` IMS user ID and then read back that user's profile via the enriched GET endpoint. `executedBy` is now server-derived from the authenticated caller's JWT (`profile.user_id ?? profile.sub`); client-supplied values are treated as intent signals only and overridden in both `createFixes` and `patchFix`. `executedBy` is removed from the `FixUpdate` OpenAPI schema.
- Adds `executedByUser: { firstName, lastName, email }` to fix responses in `GET .../fixes` and `GET .../fixes/:fixId` by resolving the stored IMS user ID via the IMS admin profile API at read time. Lookups are batched (≤5 concurrent), deduplicated per request, and fail silently so responses are always returned even when IMS is unavailable.
- Fixes a regression where `PATCH .../fixes/:fixId` with only an `executedBy` intent signal returned 400 when the resolved caller identity was already stored on the fix (idempotent re-assertion now correctly returns 200).

## Test plan
- [x] `npm test` — all unit tests pass including enrichment cases and the idempotent intent-signal case in `test/controllers/fixes.test.js`.
- [x] `npm run docs:lint` — OpenAPI validates with `executedBy` removed from `FixUpdate`.
- [ ] `POST .../fixes` with `[{ "executedBy": "<attacker-ims-id>", ... }]` — verify the stored `executedBy` equals the JWT caller's identity, not the supplied value.
- [ ] `PATCH .../fixes/:fixId` with `{ "executedBy": "<any-value>" }` — verify 200 is returned and the stored `executedBy` equals the JWT caller's identity.
- [ ] `PATCH .../fixes/:fixId` with `{ "executedBy": "<any-value>" }` when `executedBy` is already set to the caller's identity — verify 200 is returned (idempotent).
- [ ] `GET .../fixes` — verify fixes with a valid IMS-format `executedBy` are enriched with `executedByUser: { firstName, lastName, email }`; plain emails and legacy values are not enriched.
- [ ] IT suite (`test/it/postgres/fixes.test.js`) passes including the `server-derives executedBy from JWT when intent signal is present` case.


MysticatBot (changes requested):
Hey @sandsinh,

## Strengths

- **Security model is sound** (`src/controllers/fixes.js`, `createFixes`): The unconditional `delete safeFixData.executedBy` before re-applying from the JWT is correct defense-in-depth. Even if `resolveCallerId` returns undefined, the client value never passes through.
- **Regex allowlist** (`src/controllers/fixes.js`, `#enrichFixesWithUserNames`): `IMS_ID_RE` prevents pre-fix legacy data (e.g., raw email addresses stored historically) from being forwarded to the admin profile API, closing the residual PII exfiltration path for old records.
- **De-duplication of IMS lookups**: Using `new Set(...)` before calling `Promise.allSettled` prevents N redundant API calls when multiple fixes share the same executor.
- **Graceful degradation**: The outer try/catch and per-ID `Promise.allSettled` mean IMS outages never degrade the fix listing response - the enrichment field is simply absent.
- **OpenAPI schema hardened**: `readOnly: true` on both `executedBy` and `executedByUser` correctly signals to consumers that these fields cannot be set. `executedBy` removed from `FixUpdate`.
- **Strong test coverage**: Tests cover the attacker scenario, fallback chain (`user_id` -> `sub`), unauthenticated rejection path, regex filtering of email addresses, and IMS failure modes. Codecov confirms all modified lines are covered.
- **Spec and plan committed**: The spec explicitly documents known limitations and tracks follow-up items.

## Issues

### Important (Should Fix)

**1. JSDoc on `#enrichFixesWithUserNames` is inaccurate - contradicts actual behavior**

`src/controllers/fixes.js:~721` - The JSDoc says "Only called from `getAllForOpportunity`; other read endpoints (`getByStatus`, `getByID`) skip enrichment intentionally." But the diff adds `await this.#enrichFixesWithUserNames([fix])` to `getByID` (line ~217). The comment directly contradicts the code.

Additionally, `getByStatus` explicitly skips enrichment while `getByID` and `getAllForOpportunity` both enrich. This inconsistency should at minimum be accurately documented - a consumer calling different endpoints for the same fix will get different response shapes.

Fix: Update the JSDoc to reflect that `getByID` also enriches, and explain why `getByStatus` is the exception (e.g., fan-out cost for large result sets).

**2. `email` fallback in `#resolveCallerId` diverges from spec and creates un-enrichable values**

`src/controllers/fixes.js:~782` - The spec states "A future improvement is to use `profile?.email` as a third fallback... This is left out of scope for this fix." But the implementation includes `?? profile?.email`. When `user_id` and `sub` are absent but `email` is present, the stored `executedBy` becomes a plain email string (e.g., `user@adobe.com`). `IMS_ID_RE` will then reject this value, so the user's profile is never enriched on read. This creates two classes of `executedBy` values with different behaviors.

Not a security issue (the email is the caller's own), but a data consistency concern and a spec/code discrepancy.

Fix: Either (a) remove the `email` fallback to match the spec (let the field be null when only email is available), or (b) update the spec to document why the email fallback was added and acknowledge it produces un-enrichable values.

**3. Asymmetric error handling between `patchFix` and `createFixes`**

`src/controllers/fixes.js` - In `patchFix`, when `executedBy` is present in the body but `callerUserId` is undefined, the endpoint returns a 400. In `createFixes`, the same scenario silently creates the fix without `executedBy`. The two endpoints have different error semantics for the same logical condition (client signals intent but identity is unresolvable).

Fix: Either align both to 400 (harder in a batch context), or add a `log.warn` in `createFixes` when identity is unresolvable and the client supplied `executedBy`, so operators can detect misconfigured auth flows.

**4. "strips client-supplied executedBy" test has a weak assertion**

`test/controllers/fixes.test.js:~974` - The test asserts `expect(fixes[0].fix.executedBy).to.not.equal('attacker@evil.org')` but does NOT assert what the value actually IS. Given the test fake's `fakeCreateFix` defaults `executedBy` via `??=`, the stored value becomes the fake's default, not `undefined` as would happen in production. The test passes for the wrong reason.

Fix: Use a positive assertion: `expect(fixes[0].fix.executedBy).to.not.exist` or assert it equals the fake's known default with a comment explaining why.

**5. Unbounded IMS fan-out with no concurrency limit**

`src/controllers/fixes.js:~743` - `Promise.allSettled(userIds.map(...))` fires all IMS profile lookups concurrently with no cap. If a response contains fixes from 50 distinct executors, 50 parallel HTTP requests hit the IMS admin API simultaneously.

Fix: Add a concurrency limiter (e.g., chunk into batches of 5-10) or cap the number of distinct user IDs enriched per request.

**6. Missing integration tests for modified endpoint behavior**

The repo's CLAUDE.md states: "New or modified endpoints must include integration tests in `test/it/`." This PR modifies the behavior of GET, PATCH, and POST fixes endpoints (adding enrichment, changing how `executedBy` is set). No integration tests were added in `test/it/`.

Fix: Add integration tests covering at minimum: PATCH with `executedBy` intent signal verifying server-derived value, and GET verifying `executedByUser` is populated.

### Minor (Nice to Have)

**7. Regex constant defined inside method**

`src/controllers/fixes.js`, inside `#enrichFixesWithUserNames` - `const IMS_ID_RE = /^.../` is re-compiled on every call. Hoist to a module-level constant.

**8. Spec/plan doc inconsistencies**

`docs/specs/...` marks "Adding `executedByUser` to the OpenAPI Fix response schema" as out-of-scope, but the diff implements it. The implementation plan's "Remaining work" section lists items that were completed. Update both docs to reflect the actual state.

**9. No test for IMS call deduplication**

The production code builds a `Set` of user IDs to avoid redundant lookups. No test verifies that when multiple fixes share the same `executedBy`, `getImsAdminProfile` is called only once.

## Recommendations

- Consider rate-limiting or caching IMS profile lookups. As fix volume grows, a short-lived in-memory cache (per-request or container-scoped with brief TTL) would reduce IMS load.
- Add a per-call timeout wrapper around IMS enrichment calls. A single hanging IMS response makes the entire fixes list response wait.
- Consider emitting a structured audit log entry when `executedBy` is derived, recording both the caller identity and the overridden client value, for forensic evidence of pre-fix exploitation.
- The `_executedByUser` / `_suggestions` pattern of mutating entity objects with ad-hoc underscore properties is acknowledged tech debt. Consider tracking a follow-up to introduce a proper view-model or enrichment-context pattern.

## Assessment

The security fix is architecturally correct and well-motivated. The core vulnerability (write-then-read PII exfiltration) is cleanly closed on both write paths. However, the inaccurate JSDoc (item 1), spec/code divergence on the email fallback (item 2), and the weak security test assertion (item 4) should be addressed before merge. The unbounded fan-out (item 5) and missing integration tests (item 6) are material concerns for production readiness.

Note: CI check `it-postgres` is failing on one run - please verify this is unrelated to the PR changes.

## Next Steps

1. Fix the JSDoc on `#enrichFixesWithUserNames` to accurately reflect which endpoints call it.
2. Resolve the `email` fallback discrepancy (remove it or update the spec).
3. Strengthen the "strips client-supplied executedBy" test assertion.
4. Add concurrency limiting to the IMS fan-out.
5. Add integration tests for the modified endpoints.
6. Minor doc cleanup on spec/plan consistency.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 34s | <b>Cost:</b> $4.81 | <b>Commit:</b> <code>2dc7088e9d594bc6c61a966b42dc35a7064dad4d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anshulk-public:
@sandsinh Please resolve these issues suggested by @MysticatBot. Apart from that it is LGTM.

MysticatBot (changes requested):
Hey @sandsinh,

**Verdict: Request changes** - security fix is correct; two doc-accuracy issues need a quick fix.
_Changes: Closes a PII exfiltration path by server-deriving `executedBy` from the authenticated JWT and adds `executedByUser` read-time enrichment via batched IMS admin profile lookups (7 files)._

## Must fix before merge
1. [Important] OpenAPI `executedByUser` description is narrower than the implementation - `docs/openapi/schemas.yaml:4062` (details inline)
2. [Important] Spec doc claims `executedByUser` schema addition is "Out of scope" / pending follow-up, but this PR implements it - `docs/specs/2026-05-22-sites-44243-executedby-security-fix.md:93` (details inline)

<details>

- nit: Read-time IMS enrichment has no caching layer; as fix volume grows per opportunity, consider a short-lived per-request or per-container cache to avoid repeated lookups for the same user across paginated requests - `src/controllers/fixes.js`
- suggestion: Add a test with >5 distinct IMS user IDs to exercise the sequential batch boundary in `#enrichFixesWithUserNames` - `test/controllers/fixes.test.js`
- suggestion: Add a partial-batch-failure test (some IDs resolve, some reject within one batch) to verify per-ID isolation - `test/controllers/fixes.test.js`
- nit: Spec "Changes" table only mentions "Removed executedBy from FixUpdate" but omits the `executedByUser` addition and `readOnly` annotation that were also applied to `schemas.yaml` - `docs/specs/2026-05-22-sites-44243-executedby-security-fix.md`
</details>

- JSDoc on `#enrichFixesWithUserNames` now accurately reflects which endpoints call it
- Email fallback removed from `#resolveCallerId`; spec and code now agree
- `createFixes` logs a warning when identity is unresolvable (asymmetric handling addressed)
- Test assertions strengthened with positive checks and explanatory comments
- IMS fan-out now batched (`IMS_ENRICH_BATCH_SIZE = 5`)
- Integration tests added for both POST and PATCH server-derivation paths
- `IMS_ID_RE` regex hoisted to module level
- IMS deduplication test added verifying `getImsAdminProfile` is called only once for shared IDs

docs/specs/2026-05-22-sites-44243-executedby-security-fix.md
```diff
@@ -0,0 +1,133 @@
...
+  from authenticated JWT tokens. No change to the enrichment logic itself; the IMS admin API
+  call remains valid because the stored value now always comes from a verified identity.
+- **OpenAPI `FixUpdate` schema**: Remove `executedBy` from the patchable properties.
+
+### Out of scope
+
+- Changing `getAllForOpportunity` enrichment logic.
+- `getByStatus` enrichment — intentionally skipped to avoid unbounded IMS fan-out on large result sets.
+- Handling the IMS auth path `user_id`/`sub` disambiguation fully.
+- Adding `executedByUser` to the OpenAPI `Fix` response schema (tracked as follow-up).
```
MysticatBot: issue (blocking): This line says adding `executedByUser` to the OpenAPI Fix schema is out-of-scope / tracked as follow-up. But this PR does exactly that in `schemas.yaml`. The spec is stale on arrival and will actively mislead readers into thinking this work is still pending. Follow-up item 1 (line ~148) repeats the same claim, and the Changes table understates the schema work. Fix: Move the schema addition from Out of scope into In scope or the Changes table, and remove Follow-up item 1 (or mark it completed).

docs/openapi/schemas.yaml
```diff
@@ -4055,6 +4055,20 @@ Fix:
       format: date-time
     executedBy:
       type: string
+      description: IMS user ID of the caller who executed the fix. Server-derived; not client-settable.
+      readOnly: true
+    executedByUser:
+      type: object
+      description: Resolved IMS profile for the user who executed the fix. Present only on the list endpoint (getAllForOpportunity) when the executor identity was successfully resolved.
```
MysticatBot: issue (blocking): The `executedByUser` description reads "Present only on the list endpoint (getAllForOpportunity) when the executor identity was successfully resolved." But the code also calls `this.#enrichFixesWithUserNames([fix])` from `getByID`. Consumers reading the spec will not know that single-fix fetches also return `executedByUser`. Fix: Update to "Present on getAllForOpportunity and getByID when the executor identity was successfully resolved. Omitted from getByStatus."

MysticatBot (approved):
Hey @sandsinh,

**Verdict: Approve** - security fix is correct and complete; minor doc/schema nits below.
_Changes: Closes a PII exfiltration path by server-deriving `executedBy` from the authenticated JWT and adds `executedByUser` read-time enrichment via batched IMS admin profile lookups (7 files)._

<details>

- nit: OpenAPI `executedByUser` properties (`firstName`, `lastName`, `email`) should be marked `nullable: true` since the implementation returns `null` for missing fields - clients using codegen will get incorrect non-nullable types - `docs/openapi/schemas.yaml:4097`
- suggestion: `patchFix` sets `hasUpdates = true` unconditionally when the `executedBy` intent signal is present, even when `callerUserId` already matches the stored value - this triggers a no-op `save()` that bumps `updatedAt`. Consider gating `hasUpdates` inside the `callerUserId !== fix.getExecutedBy()` branch, or rename the test away from "idempotent" since idempotent operations should not produce side effects on repeated invocation - `src/controllers/fixes.js:605`
- suggestion: The "strips client-supplied executedBy when caller identity cannot be resolved" test asserts via the `fakeCreateFix` harness default (`'test user'`) rather than directly verifying the argument passed to `FixEntity.create()` - a spy assertion like `sinon.match((arg) => !('executedBy' in arg))` would be more resilient to test helper changes - `test/controllers/fixes.test.js:1178`
</details>

- OpenAPI `executedByUser` description now correctly mentions both getAllForOpportunity and getByID
- Spec doc "Out of scope" section updated to properly reflect executedByUser schema addition in Changes table
- JSDoc on `#enrichFixesWithUserNames` accurately reflects which endpoints call it
- Email fallback removed from `#resolveCallerId`; spec and code agree
- `createFixes` logs a warning when identity is unresolvable
- Test assertions strengthened with positive checks
- IMS fan-out now batched (IMS_ENRICH_BATCH_SIZE = 5)
- Integration tests added for POST and PATCH server-derivation paths
- IMS_ID_RE regex hoisted to module level
- IMS deduplication test verifying single call for shared IDs

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 55s | <b>Cost:</b> $7.48 | <b>Commit:</b> <code>34b9dc3244163c935ecece3f1048219630c79f6d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2447, opened 2026-05-20
feat: enable trace-id propagation through preflight

bhellema:
## Summary

- Propagates `ctx.traceId` into the SQS message payload when creating a preflight async job, enabling end-to-end trace correlation across the preflight pipeline
- Adds `debug`-level logging throughout the preflight controller (job creation, SQS dispatch, auth check, job status retrieval) to improve observability
- Minor refactoring: extracts `jobPayload` into a named variable, renames `previewBaseURL` parameter to `url` in `checkEnableAuthentication`, and reorders variable declarations for readability

### Sample Dashboard on Dev

I have created a [dashboard](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#dashboards/dashboard/Preflight-Tracing?start=PT1H&end=null&var_traceId=bhellema-dev-6) on dev in cloudwatch that shows the audit trail now.

## Test plan

- [x] `npm test` passes
- [x] Verify preflight controller unit tests cover the trace-id path
- [x] Smoke-test in dev: trigger a preflight job and confirm `traceId` appears in the SQS message and propagates to downstream consumers

### DIFF
--- a/src/controllers/preflight.js
+++ b/src/controllers/preflight.js
@@ -142,11 +142,19 @@ function PreflightController(ctx, log, env) {
     return configuration.isHandlerEnabledForSite(type, site);
   }
 
-  async function checkEnableAuthentication(previewBaseURL) {
-    const headResponse = await fetch(previewBaseURL, {
+  /**
+   * Checks if authentication is enabled for a given URL
+   * @param {string} url - The URL to check
+   * @returns {Promise<boolean>} True if authentication is enabled, false otherwise
+   */
+  async function checkEnableAuthentication(url) {
+    const headResponse = await fetch(url, {
       method: 'HEAD',
       headers: { 'Content-Type': 'application/json' },
     });
+
+    log.debug(`checkEnableAuthentication for ${url} returned status: ${headResponse.status}`);
+
     return headResponse.status === 401 || headResponse.status === 403;
   }
 
@@ -162,6 +170,7 @@ function PreflightController(ctx, log, env) {
   }
 
   const createPreflightJob = async (context) => {
+    log.debug('createPreflightJob started');
     const { data } = context;
     try {
       validateRequestData(data);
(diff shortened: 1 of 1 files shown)


src/controllers/preflight.js
```diff
@@ -173,18 +182,22 @@ function PreflightController(ctx, log, env) {
...
       const previewBaseURL = `${url.protocol}//${url.hostname}`;
+
+      let site;
       if (isValidUUID(data.siteId)) {
         site = await dataAccess.Site.findById(data.siteId);
       } else {
         site = await dataAccess.Site.findByPreviewURL(previewBaseURL);
       }
 
+      log.debug(`createPreflightJob url: ${url}, siteId: ${data.siteId}}, step: ${step}`);
```
GeezerPelletier: Stray `}}` in the template literal — renders as `siteId: <value>}, step:` in the log output. Should be `${data.siteId}, step:`.

src/controllers/preflight.js
```diff
@@ -199,15 +212,19 @@ function PreflightController(ctx, log, env) {
...
 
       // Create a new async job
+      const jobPayload = {
+        siteId: site.getId(),
+        urls: data.urls,
+        step,
+        enableAuthentication,
+      };
+
+      log.debug('createPreflightJob creating async job with payload: ', jobPayload);
```
GeezerPelletier: Inline and stringify "jobPayload" here? Like java I assume this won't print as expected otherwise?

src/controllers/preflight.js
```diff
@@ -219,13 +236,18 @@ function PreflightController(ctx, log, env) {
...
           siteId: site.getId(),
           type: 'preflight',
+          ...(ctx.traceId && { traceId: ctx.traceId }),
         };
+
         if (PROMISE_BASED_TYPES.includes(site.getAuthoringType())) {
           sqsMessage.promiseToken = promiseTokenResponse;
         }
+
+        log.debug(`createPreflightJob sending message to SQS with payload: ${JSON.stringify(sqsMessage)}`);
```
GeezerPelletier: This will log the promise token which is a security no no.  Maybe move up to just before where the token is added to the sqsMessage? Or util to obfuscate otherwise.

GeezerPelletier (changes requested):
Couple things to look at, otherwise fine yeah.

src/controllers/preflight.js
bhellema: It will print it.

image

But for consistency I'll use a template literal.

GeezerPelletier (approved):
LGTM!

PR #2449, opened 2026-05-20
fix: resolveTopicUuid and resolveCategoryUuid accept UUID input directly

jjenscodee:
## Summary

- `resolveTopicUuid` and `resolveCategoryUuid` now handle UUID inputs by doing an org-scoped primary-key lookup (`WHERE id = ? AND organization_id = ?`), consistent with the existing `resolveBrandUuid` pattern
- Business-key inputs (slugs, names) continue to use the existing `topic_id` / `category_id` column lookup unchanged
- Adds unit tests covering the UUID path (found and not-found) for both functions

## Root cause

When a user deletes a topic from the Prompts Management page, the UI passes the topic's UUID PK (sourced from `prompt.topic.id` in list responses) as the `topicId` filter param on `GET .../prompts`.

`resolveTopicUuid` treated every input as a business key and queried `WHERE topic_id = ?`. The lookup returned `null`, the `topic_id` filter on the prompts query was silently skipped, and **all prompts for the brand** were returned instead of just the topic's prompts. With a chunk size of 500 (exceeding the API's 100-prompt hard limit), this caused a `Maximum 100 prompt IDs per request` error on every topic delete.

## Test plan
- Make sure the branch is deployed and run UI locally points to dev
- Try deleting a topic and open the network tab to check the payload from delete api call
- Verify that the number of prompts in the payload is correct based on the topic you delete

### DIFF
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -55,11 +55,11 @@ export async function resolveBrandUuid(organizationId, brandId, postgrestClient)
 }
 
 /**
- * Resolves category business key (or UUID) to categories.id (uuid).
- * If categoryId is already a valid UUID it is returned directly — the UI
- * derives category identifiers from category.id (UUID PK) in prompt responses,
- * so bypassing the business-key lookup avoids a spurious miss and ensures the
- * filter is never silently skipped.
+ * Resolves category business key or UUID to categories.id (uuid).
+ * When categoryId is a UUID it is looked up by primary key scoped to the
+ * organization (consistent with resolveBrandUuid) — this validates org
+ * ownership rather than blindly trusting the caller-supplied UUID.
+ * When categoryId is a business key it is looked up by category_id as before.
  *
  * @param {string} organizationId - SpaceCat organization UUID
  * @param {string} categoryId - Business key or UUID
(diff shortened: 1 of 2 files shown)


calvarezg:
## Code Review

### CI Status

Two checks are currently failing: `ci / it-postgres` and `ci / semantic-release`. These need to be investigated/resolved before merge.

---

### Code Quality

Correct and clear fix. The logic is minimal and the PR description explains the root cause well. Tests properly verify the UUID fast-path doesn't hit the DB (`stub().throws()` means any DB call would cause the function to throw and fail the test, so `expect(client.from.called).to.be.false` is belt-and-suspenders but harmless).

---

### Key Concern: Inconsistency with `resolveBrandUuid`

The existing `resolveBrandUuid` handles UUIDs differently — it **still does a DB lookup** to verify the UUID belongs to the correct organization:

```js
// resolveBrandUuid (existing) — verifies org ownership
if (isValidUUID(brandId)) {
    .from('brands')
    .select('id')
    .eq('organization_id', organizationId)  // ← org boundary enforced
    .eq('id', brandId)
    .maybeSingle();
  if (!error && data?.id) return data.id;
  return null;  // ← returns null if UUID not in this org
}

// resolveCategoryUuid / resolveTopicUuid (new) — no org verification
if (isValidUUID(categoryId)) {
  return categoryId;  // ← any UUID is accepted blindly
}
```

This means a user could pass another organization's topic/category UUID as a filter parameter and it would resolve successfully. Whether this is actually exploitable depends entirely on whether the downstream `listPrompts` query also scopes by `organization_id`. If it does, the worst case is 0 results (not a data leak). If it doesn't, cross-org data could be exposed.

**Recommendation**: Either verify that downstream queries always enforce the org boundary (and add a comment to that effect), or align this implementation with `resolveBrandUuid` by doing a scoped DB lookup to validate org ownership.

---

### Minor: Comment verbosity

The added JSDoc comments explain UI implementation details inside server-side code. The context is valid (it's a subtle footgun), but per project conventions comments should stay concise. Not a blocker.

---

### Summary

| Area | Assessment |
|---|---|
| Fix correctness | Correct for the stated bug |
| Test coverage | Good — new tests for the UUID fast-path |
| CI | Failing (`it-postgres`, `semantic-release`) — needs investigation |
| Security consistency | Inconsistent with `resolveBrandUuid` — needs justification or alignment |

**Primary ask before merging**: Clarify/address the org-boundary validation inconsistency, and fix the failing CI checks.

calvarezg:
## Follow-up Review

Both concerns from the previous review are resolved:

**Security consistency** ✅ — The new implementation now performs a DB lookup scoped to `organization_id` when a UUID is passed, consistent with `resolveBrandUuid`. The query builder pattern (building `query` then branching on `isValidUUID`) is clean and avoids duplication:

```js
const query = postgrestClient.from('topics').select('id').eq('organization_id', organizationId);
const { data, error } = await (isValidUUID(topicId)
  ? query.eq('id', topicId)
  : query.eq('topic_id', topicId)
).maybeSingle();
```

**CI** ✅ — All checks passing, including `it-postgres`.

**Tests** ✅ — New tests cover the UUID-by-PK path and the "returns null when UUID doesn't belong to the org" path.

LGTM 🚀

calvarezg (approved):

PR #2450, opened 2026-05-20
fix(routes): validate :jobId as UUID before reaching controllers

LeChef318:
## Summary

Adds UUID v4 validation on `params.jobId` in `src/index.js`, mirroring the existing pattern for `:siteId`, `:organizationId`, `:plgOnboardingId`, `:onboardingId`, `:spaceCatId`, `:brandId`, and `:executionId`.

## Problem

The route matcher (`src/utils/route-utils.js:56-95`) is **first-match-wins with no specificity scoring** and enforces strict segment-count equality. `GET /tools/scrape/jobs/:jobId` (`src/routes/index.js:386`) is registered before the static-prefix routes (`by-url`, `by-base-url`, `by-date-range`), so a bare `GET /tools/scrape/jobs/by-url/` (4 segments, missing required `:url`) gets captured by `:jobId` with `jobId="by-url"` and routed to `getScrapeJobStatus()` instead of the by-url handler.

Same misroute class exists on `/tools/import/jobs/by-date-range/` (`src/routes/index.js:377` sits after `:jobId` at line 372).

## Evidence

- **1,224 ERROR-level log entries in 7 days** (`/aws/lambda/spacecat-services--api-service` prod), all `Failed to fetch scrape job status for jobId: by-url, message: Job ID is required`.
- Surfaced during the [SITES-43989](https://jira.corp.adobe.com/browse/SITES-43989) Sev1 investigation as scrape-API error noise that masked the real signal.
- Verified the caller path in the IMS-auth log line: `Checking authentication with IMS for product ASO at route GET /tools/scrape/jobs/by-url/`.

## Why UUID validation (not route reordering)

The ticket suggested reordering static-prefix routes above `:jobId`, citing the `/consumers/by-client-id/:clientId` precedent. That fix doesn't apply here:

- The matcher enforces strict segment-count equality (`route-utils.js:69`), so `:jobId` (4 segments) and `by-url/:url` (5 segments) never collide for well-formed requests. The misroute only fires for the bare 4-segment prefix, for which no static route exists to "win" reordering.
- The `/consumers/` precedent is stylistic — those routes have different segment counts (2 vs 3) and can't collide regardless of order.

UUID validation matches the documented convention in `CLAUDE.md` ("UUID parameters are validated in `src/index.js` using `isValidUUIDV4()` before reaching controllers") and protects `:jobId` across all five endpoint families that share the parameter: scrape, import, preflight, beta preflight, site detection, consent-banner.

## Follow-up

The root cause (first-match-wins matcher with no specificity scoring) remains and is mentioned in the ticket as an "Optional follow-up (M)" — worth filing as a separate ticket but out of scope here.

## Related Issues

- Fixes [SITES-45112](https://jira.corp.adobe.com/browse/SITES-45112)
- Related: [SITES-43989](https://jira.corp.adobe.com/browse/SITES-43989) (Sev1 that surfaced this), [SITES-45110](https://jira.corp.adobe.com/browse/SITES-45110) (upstream producer fix)

## Test plan

- [x] Two new tests in `test/index.test.js` covering the scrape misroute and the import sibling case — both return 400 with `Job Id is invalid. Please provide a valid UUID.`
- [x] Watched both tests fail (RED) before adding the validation; confirmed failure reasons matched expected misroute behavior.
- [x] Full `npm test` suite green: 10,299 passing.
- [x] `npm run lint` clean (pre-existing warnings in unrelated files unchanged).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -346,6 +346,9 @@ async function run(request, context) {
       if (params.executionId && !isValidUUID(params.executionId)) {
         return badRequest('Execution Id is invalid. Please provide a valid UUID.');
       }
+      if (params.jobId && !isValidUUIDV4(params.jobId)) {
+        return badRequest('Job Id is invalid. Please provide a valid UUID.');
+      }
       context.params = params;
       context.request = request;
 
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -346,6 +346,17 @@ describe('Index Tests', () => {
     expect(resp.headers.plain()['x-error']).to.equal('Job Id is invalid. Please provide a valid UUID.');
   });
 
+  it('rejects bare /tools/scrape/jobs/by-base-url misroute with invalid jobId', async () => {
+    context.pathInfo.suffix = '/tools/scrape/jobs/by-base-url';
+
+    request = new Request(`${baseUrl}/tools/scrape/jobs/by-base-url`, { headers: { 'x-api-key': apiKey } });
+
+    const resp = await main(request, context);
+
+    expect(resp.status).to.equal(400);
+    expect(resp.headers.plain()['x-error']).to.equal('Job Id is invalid. Please provide a valid UUID.');
+  });
+
   it('rejects bare /tools/import/jobs/by-date-range misroute with invalid jobId', async () => {
     context.pathInfo.suffix = '/tools/import/jobs/by-date-range';
 
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -335,6 +335,28 @@ describe('Index Tests', () => {
     expect(resp.headers.plain()['x-error']).to.equal('Execution Id is invalid. Please provide a valid UUID.');
   });
 
+  it('rejects bare /tools/scrape/jobs/by-url misroute with invalid jobId', async () => {
+    context.pathInfo.suffix = '/tools/scrape/jobs/by-url';
+
+    request = new Request(`${baseUrl}/tools/scrape/jobs/by-url`, { headers: { 'x-api-key': apiKey } });
+
+    const resp = await main(request, context);
+
+    expect(resp.status).to.equal(400);
+    expect(resp.headers.plain()['x-error']).to.equal('Job Id is invalid. Please provide a valid UUID.');
+  });
+
+  it('rejects bare /tools/import/jobs/by-date-range misroute with invalid jobId', async () => {
+    context.pathInfo.suffix = '/tools/import/jobs/by-date-range';
+
+    request = new Request(`${baseUrl}/tools/import/jobs/by-date-range`, { headers: { 'x-api-key': apiKey } });
+
+    const resp = await main(request, context);
+
+    expect(resp.status).to.equal(400);
+    expect(resp.headers.plain()['x-error']).to.equal('Job Id is invalid. Please provide a valid UUID.');
+  });
+
   it('handles dynamic route errors', async () => {
     context.pathInfo.suffix = '/sites/e730ec12-4325-4bdd-ac71-0f4aa5b18cff';
 


solaris007 (commented):
Hey @LeChef318,

Solid, well-scoped fix. I traced the matcher, the validator, and the six route families that share `:jobId`, and the change is correct: it returns a clean 400 for the bare-prefix misroutes before any controller runs, and it follows the existing per-param validation convention. The code is sound and verified safe; the items below are test-hardening and documentation suggestions.

## Strengths

- Correct fix at the right layer. The gate at `src/index.js:349` sits in the central pre-controller validation block alongside `siteId`, `organizationId`, `executionId`, and the others, matching the documented convention. One gate covers all six `:jobId` families instead of six controller-level patches.
- The author correctly rejected the ticket's suggested route reordering and explained why accurately: the matcher (`src/utils/route-utils.js`) enforces strict segment-count equality and is first-match-wins, so the bare 4-segment `/tools/scrape/jobs/by-url` has no static route to "win" via reordering. Resisting an authoritative-but-wrong ticket suggestion is the right instinct.
- `isValidUUIDV4` is the correct, verified-safe choice. Every `:jobId` family mints v4 IDs: scrape and import via their data-access entities, and preflight, beta-preflight, and site-detection via `AsyncJob.create()` (`src/controllers/preflight.js:202,456`, `src/controllers/site-detection.js:111`), all of which derive the ID from the data-access SchemaBuilder default `crypto.randomUUID()`. consent-banner's `:jobId` is itself a scrape job ID. So the v4-strict check rejects only malformed input, never a legitimate ID across any of the six families.
- The regex (`src/index.js:108`) is anchored on both ends and pins the version and variant nibbles, so it correctly rejects the nil UUID, embedded matches, and trailing-newline payloads. The matcher does not percent-decode segments, so there is no encoding bypass.
- Both tests verify behavior (status plus the exact `x-error` message), not just line coverage, and reproduce the actual Sev1 misroute path.

## Issues

### Minor (Nice to Have)

- Two of the scrape static-prefixes are untested (`test/index.test.js:338,349`). The same bare-prefix misroute also exists for `/tools/scrape/jobs/by-base-url` and `/tools/scrape/jobs/by-date-range` (`src/routes/index.js:388-390`); the tests cover scrape `by-url` and import `by-date-range`. One guard covers all of them, so there is no live bug, but adding a `by-base-url` case (or a comment noting the two are representative) would make the negative set match the actual misroute surface.
- The error string says "valid UUID" but the check requires v4 specifically (`src/index.js:350`). A syntactically valid non-v4 UUID would be told it is not "a valid UUID." This is pre-existing across the block (the `siteId` and `organizationId` messages are the same), so it is consistency-neutral here. Flag only to align the wording if the block is ever cleaned up. Defer.

## Recommendations

- Add one positive test that pins the gate's pass-branch for `:jobId` (`test/index.test.js`). Both new tests are negative (bad input -> 400). The shared `uuidRegex` is already exercised on its pass-branch by the valid-`siteId` "handles dynamic route errors" test, so a break in the regex itself would be caught; the narrower gap is that nothing pins that a well-formed v4 `jobId` passes the new gate at line 349, which now fronts all six job-status families. A future jobId-specific change there could start returning 400 for every job-status endpoint with the suite still green. Since this PR widens the gate to six families, it is a good place to add one happy-path test:
  ```js
  it('passes a well-formed v4 jobId through the gate', async () => {
    const jobId = '1f3e8a2c-9b4d-4e6f-8a1b-2c3d4e5f6a7b';
    context.pathInfo.suffix = `/tools/scrape/jobs/${jobId}`;
    request = new Request(`${baseUrl}/tools/scrape/jobs/${jobId}`, { headers: { 'x-api-key': apiKey } });
    const resp = await main(request, context);
    expect(resp.headers.plain()['x-error']).to.not.equal('Job Id is invalid. Please provide a valid UUID.');
  });
  ```
- Document the six-family contract in the PR description. The change reads as a scrape/import noise fix, but the gate fronts all six `:jobId` families. Since all six are verified v4, one sentence making that explicit prevents a future reader from assuming the blast radius is just scrape and import.

- The root cause is the first-match-wins matcher with no specificity scoring (`src/utils/route-utils.js`); the PR body already defers it as a follow-up, which is the right call under Sev1 pressure. Relatedly, the validation block is now eight sequential per-param guards, each hand-coding a param-to-validator mapping the route table already encodes. A single follow-up covering a specificity-scored matcher plus data-driven param-type declarations would retire this whole misroute class. Not for this PR.

## Assessment

Ready to merge? Yes.

Reasoning: The fix is correct, verified safe across all six `:jobId` families, and well-tested for the failure path. The recommended positive test and the Minor items are worth doing but are not blockers.

## Next Steps
1. Consider adding one positive happy-path test for a valid v4 `jobId` so the gate's pass-branch is pinned.
2. Optionally add a `by-base-url` negative and a note in the PR description that the gate covers all six `:jobId` families.
3. The Minor wording item is deferrable.

LeChef318:
Thanks @solaris007 — addressed Minor 1 by adding a third negative test for bare `/tools/scrape/jobs/by-base-url` (test commit fee7aba7). The gate already covers it; the test pins the behavior so a future regression on the guard or the scrape route table can't slip through silently.

Deferring Minor 2 (wording "valid UUID" vs. v4-strict) as you suggested — would be scope creep here since the same wording is used across all 7 existing UUID checks in the block, so consistent cleanup belongs in a separate PR.

Deferring the positive happy-path test recommendation for the same reason: the existing valid-`siteId` "handles dynamic route errors" test exercises the regex pass-branch, so the gap is narrow. Happy to add it in this PR if you'd prefer — let me know.

solaris007 (approved):
Hey @LeChef318,

Thanks for the follow-up. The new `by-base-url` negative test (fee7aba73) closes the coverage gap I flagged, so the three bare-prefix misroutes (`by-url`, `by-base-url`, `by-date-range`) are now all pinned. The fix itself remains correct and verified safe: the v4 gate at `src/index.js:349` returns a clean 400 before any controller runs, and all six `:jobId` families mint `crypto.randomUUID()` v4 IDs, so no legitimate request is rejected.

Approving.

Two non-blocking notes, neither gating this approval:
- The one happy-path test (a valid v4 `jobId` passing the gate) is still worth adding so the gate's pass-branch is pinned for the six families it now fronts. Fine as a follow-up.
- `ci / it-postgres` was still running as I wrote this - let it go green before merging.

LGTM.

PR #2452, opened 2026-05-20
docs(specs): add SITES-43989 error-rate remediation design

solaris007:
## What

Spec-only PR. Adds the design doc for the **inline-fix subset** of the SITES-43989 error-rate remediation. No code changes - implementation will follow in a separate PR.

`docs/specs/2026-05-20-sites-43989-error-rate-remediation.md`

## Why

The Sev1 "API Gateway Error Rate High" alert ([SITES-43989](https://jira.corp.adobe.com/browse/SITES-43989)) is driven by total 4xx+5xx. This spec covers the small, self-contained `spacecat-api-service` changes that reduce that noise as defense-in-depth.

## Scope (3 fixes)

1. **Webhook log noise** - move `getWorkspaceRepos` out of the `WebhooksController` constructor so the `MYSTICAT_WORKSPACE_REPOS` warn stops firing on every request (~94k/20min -> ~0). No config change (defaults are authoritative; verified via Coralogix).
2. **LLMO upstream-404 -> empty 200** - shared `fetchLlmoSource()` helper; "site enabled but data not yet provisioned" returns empty success instead of 400; non-404 failures keep error semantics; 404 logged at info + metric.
3. **Scrape API** - no change, with documented rationale (mystique PR #1777 already drives its scrape-API 4xx contribution to ~zero).

## Out of scope

Producer root cause (SITES-45110), scrape job-status routing bug (SITES-45112), Fastly rate-limit / health checks (deferred), LLMO data backfill + onboarding provisioning hardening, SKYSI alert tuning.

## Notes

Produced via a structured brainstorming workflow; design was reviewed and self-checked before commit. Next step is an implementation plan and a separate code PR.

### DIFF
--- a/docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
+++ b/docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
@@ -1,20 +1,22 @@
 # SITES-43989 Error-Rate Remediation - Inline Fixes
 
-Status: proposed
-Last verified: 2026-05-20
-Jira: [SITES-43989](https://jira.corp.adobe.com/browse/SITES-43989)
+- **Date:** 2026-05-20
+- **Author:** Dominique Jaeggi
+- **Status:** Proposed
+- **Jira:** [SITES-43989](https://jira.corp.adobe.com/browse/SITES-43989)
+- **Target repo:** `adobe/spacecat-api-service`
 
 ## Context
 
-On 2026-05-04 the Sev1 "API Gateway Error Rate High" alert fired 3 times for AEM Sites
-Optimizer. The alert monitors **total** error rate (4xx + 5xx combined). Two independent
-sources pushed the rate over threshold:
+On 2026-05-04 the Sev1 "API Gateway Error Rate High" alert (SKYSI-76262) fired 3 times for
+AEM Sites Optimizer. The alert monitors **total** error rate (4xx + 5xx combined). Two
+independent sources pushed the rate over threshold:
 
 1. LLM-hallucinated garbage URLs from a producer hammering the scrape `by-url` API with
    400s (the headline cause).
 2. A steady background of 4xx from LLMO query endpoints returning `400` for sites whose
-   data was not yet provisioned, plus ~94k/20min warn-level log noise from a missing
-   webhook env var.
+   data was not yet provisioned (~253 errors / 20 min), plus ~94k/20min warn-level log
+   noise from a missing webhook env var.
 
 This spec covers only the **inline** subset of the SITES-43989 remediation - the small,
 self-contained `spacecat-api-service` changes. Each fix stands on its own merit as
@@ -36,12 +38,37 @@ defense-in-depth; none depends on the others.
 
 The error-rate alert counts **all** 4xx+5xx, so changing a code to be "more correct"
 (e.g. 400 -> 404) does not reduce the count; only reclassifying a condition to a **2xx**
-removes it from the alert. We adopt a **hybrid** policy:
+removes it from the alert. We adopt a **hybrid reclassification policy**:
 
-- Reclassify the genuinely-expected "no data yet" LLMO condition to an empty `200`.
+- Reclassify the genuinely-expected "no data yet" LLMO condition (upstream 404) to an
+  empty `200`.
+- Map genuine upstream failures to honest status codes instead of today's blanket 400:
+  upstream **5xx -> 502**, **timeout -> 504**, non-404 4xx **passed through**, auth
+  (401/403) unchanged.
 - Leave the scrape API's `400` for malformed input unchanged (it is correct, and the
   producer already filters those URLs).
-- Preserve real error semantics everywhere else (5xx, timeouts, auth).
+
+This policy is **LLMO-only** for now. The service has other upstream-proxy paths;
+generalizing "upstream-404 -> empty success" into a service-wide pattern is explicitly out
+of scope until a second use case justifies it.
+
+## Implementation phasing
+
+Several contracts must be pinned before code is written, so implementation is split:
+
+**Phase 0 - pin contracts (complete before Phase 1):**
+
+1. **Capture empty-response fixtures.** `curl` the elmo-ui-data source for a known-empty
+   folder for each of the three response shapes (single-file, multi-sheet, files-array)
+   using the real `LLMO_HLX_API_KEY`, and commit the captured JSON as test fixtures. These
+   fixtures ARE the empty-200 contract; the synthesized payloads must byte-match them.
+2. **Audit and decide fetch unification** (see Fix 2). Test the largest current sheet-data
+   payloads against a 15s timeout in dev and confirm none legitimately exceed it before
+   adopting the timeout everywhere.
+3. **Confirm metric ingestion.** Verify the chosen log level for the not-provisioned event
+   reaches Coralogix ingestion and author the `events2metrics` rule.
+
+**Phase 1 - implement** the three fixes against the pinned contracts.
 
 ---
 
@@ -49,8 +76,8 @@ removes it from the alert. We adopt a **hybrid** policy:
 
 ### Problem
 
-`src/index.js:262` eagerly instantiates **every** controller on **every** request,
-before routing. `WebhooksController`'s factory calls `getWorkspaceRepos(env, log)` in its
+`src/index.js:262` eagerly instantiates **every** controller on **every** request, before
+routing. `WebhooksController`'s factory calls `getWorkspaceRepos(env, log)` in its
 constructor body (`src/controllers/webhooks.js:67`), which emits:
 
 ```
@@ -58,20 +85,27 @@ log.warn('MYSTICAT_WORKSPACE_REPOS not set, using built-in defaults', { defaults
 ```
 
 when the env var is unset. Because the controller is built on every request, this warn
-fires on all traffic - ~94,116 warns / 20 min in prod (99.99% of all warn-level logs).
-The resolved `workspaceRepos` value is only ever **used** inside `processGitHubWebhook`
+fires on all traffic - ~94,116 warns / 20 min in prod (99.99% of all warn-level logs). The
+resolved `workspaceRepos` value is only ever **used** inside `processGitHubWebhook`
 (`webhooks.js:159`), not in the constructor.
 
 ### Change
 
-Move the `getWorkspaceRepos(env, log)` call out of the `WebhooksController` constructor and
-into `processGitHubWebhook`, computed once when an actual GitHub webhook is processed. The
-constructor becomes side-effect-free.
+1. **Move the call site.** Move `getWorkspaceRepos(env, log)` out of the `WebhooksController`
+   constructor and into `processGitHubWebhook`, computed once **per webhook request**
+   (consistent with the per-request controller model; no global/cross-request caching is
+   intended or needed). The constructor becomes side-effect-free. The warn then fires only
+   on genuine webhook deliveries (rare), not on all traffic: ~94k/20min -> near zero.
+2. **Reclassify the warn.** Unset-using-defaults is the confirmed-intended steady state (see
+   Non-goal), so warning on it is noise nobody should act on. Emit the warn **only** when
+   `MYSTICAT_WORKSPACE_REPOS` is present-but-invalid (the genuinely actionable case); for the
+   unset -> defaults branch, drop to `debug`. The validation/fallback logic is otherwise
+   unchanged.
 
-- Result: the warn only fires on genuine webhook deliveries (rare), not on every request.
-  Expected reduction: ~94k/20min -> near zero.
-- The `getWorkspaceRepos` logic (validation, defaults fallback) is unchanged - only its
-  call site moves.
+> **Forward note (out of scope here):** the root enabler is that `src/index.js:262`
+> instantiates every controller per request before routing, so any factory-level side
+> effect (logging, I/O) runs on all traffic. New controllers should keep their factory pure
+> and defer side effects to the handler.
 
 ### Non-goal: do not set the env var
 
@@ -79,16 +113,19 @@ Coralogix confirms (2026-05-20) that **dev and prod both run on the built-in def
 today** (the warn is present in both), and the built-in defaults
 (`adobe/mysticat-architecture`, `adobe/mysticat-ai-native-guidelines`,
 `Adobe-AEM-Sites/aem-sites-architecture`) are the intended set. So defaults remain
-authoritative and no Vault/secret/deploy change is needed. Setting the env var is left as
-an optional future explicitness improvement, out of scope here.
+authoritative and no Vault/secret/deploy change is needed. With the warn reclassified
+(change 2 above), the unset-using-defaults state is now silent - matching its intentional
+status - and only a present-but-invalid value warns. Setting the env var is left as an
+optional future explicitness improvement, out of scope here.
 
 ### Tests
 
-- Update `test/controllers/webhooks.test.js`: the "logs warning when not set" assertion
-  moves to the webhook-processing path (the warn no longer fires at construction).
-- Add: constructing `WebhooksController` does not emit the warn.
-- Existing tests for `getWorkspaceRepos` behavior (valid entries, invalid entries,
-  empty -> defaults) still pass unchanged.
+- Update `test/controllers/webhooks.test.js`: constructing `WebhooksController` emits no
+  warn; the unset -> defaults path no longer warns (now `debug`); a present-but-invalid
+  `MYSTICAT_WORKSPACE_REPOS` still warns, on the webhook-processing path.
+- Concurrency intent: `getWorkspaceRepos` is computed once per request; no global
+  single-compute is intended or asserted.
+- Existing valid / invalid / empty -> defaults behavior tests still pass.
 
 ### Validation gate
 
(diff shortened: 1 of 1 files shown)


LeChef318 (commented):
Hey @solaris007,

This is a well-scoped, well-reasoned spec - the scope-discipline (the `Related work` fence table) and the willingness to record Fix 3 as a deliberate no-op are above-average for a post-incident remediation. The reclassification framing ("only 2xx removes from the alert") is exactly the conceptual frame the implementer needs.

The blockers are spec-side ambiguities that an implementer will hit and either come back to ask about or guess wrong on. None is structural; all are clarifications.

## Strengths

- **Scope discipline.** The `Related work (NOT in this spec)` table cleanly fences this PR off from Fastly rate limiting, SKYSI tuning, the producer fix, and LLMO backfill. For an incident-response spec spanning many systems, this is the right structural move.
- **Fix 3 as a documented no-op.** Recording the rationale for leaving the scrape 400 in place (post-PR-1777 producer guard, structurally-valid URLs already get `ok([])`, 400 for malformed input is correct REST) is honest change-discipline and prevents future re-litigation.
- **Reclassification framing.** The `Guiding decision` section establishes that only 2xx reclassification reduces the alert count - exactly the conceptual frame an implementer needs to avoid the false-economy "400 -> 404" cleanup that would not move the metric.
- **Risks-and-trade-offs section names real risks** (provisioning gaps hidden from HTTP status, empty-shape compatibility, refactor surface) rather than boilerplate. Mitigations are concrete.
- **Per-endpoint enumeration in Fix 2** (the table with controller + file references) is the right specificity for handing off to a separate implementation PR.

## Issues

### Important (Should Fix)

All seven also appear as inline comments anchored to the relevant lines.

1. **"Preserve existing behavior" is undefined - the four call sites do not have consistent fetch behavior today.** (line 147) Verified against base: `llmo-query-handler.js` has the 15s `AbortController` timeout and the `LLMO_HLX_API_KEY` presence check; the three sheet endpoints in `llmo.js` (`getLlmoSheetData`, `queryLlmoSheetData`, `getLlmoGlobalSheetData`) have NO timeout and use `env.LLMO_HLX_API_KEY || 'hlx_api_key_missing'` to paper over a missing key. Adopting one set of semantics silently changes behavior for the others. Spell out which behaviors are being unified.

2. **The empty-response envelope is the highest-risk implementation decision and the spec defers it.** (line 155) Three distinct response shapes (single-file, multi-sheet, files-array), the synthesized payload must be byte-identical to what clients receive for genuinely-empty data, and a hard-coded `:version` goes stale the moment helix-pages bumps it. For a design doc, this IS the contract - pin it now (capture three real fixtures) or define a Phase 0 capture-and-pin task.

3. **Behavior matrix contradicts the prose on 5xx/timeout HTTP status mapping.** (line 178) Copilot already flagged this; concurring and promoting in severity. The prose says 5xx/timeout/auth keep "their current non-2xx response" (= 400), but the matrix After column says "4xx/5xx + log.error (unchanged semantics)" (= map to upstream status). The implementer cannot resolve this without a decision. The fix is also a natural moment to map upstream 5xx -> 502 and timeout -> 504.

4. **The "metric" introduces a new operational primitive with no established pattern named.** (line 167) `src/` has no existing `putMetricData` or custom-metric pattern. So this is either introducing a new mechanism or leaning on Coralogix `events2metrics` derived from a structured log field. Without specification, "add a metric" gets implemented as a log line that never gets a derived metric created, and the Risks section's appeal to a metric for observability evaporates.

5. **404-as-empty is a one-way door; design a discriminator now or document why not.** (line 177) Clients receiving `200 + empty payload` for both "provisioned but empty" and "not yet provisioned" cannot distinguish those states from the response. The next requirement up the stack is almost always "show 'your data is being prepared' in the UI" or "alert when provisioning has been stuck for 72h" - both need that distinction at the API boundary. Cheap at design time, hard to retrofit.

6. **Validation gate does not validate the spec's "must verify" claim.** (line 197) The design says "Implementation must verify the exact envelope against a live empty response rather than assume these field names." The gate then lists `npx mocha`, `npm run lint`, `npm run docs:lint` - none verify the live envelope. Asserting on a stub the implementer crafted is circular.

7. **Fix 1's "Non-goal" section contradicts the warn-on-construction fix it justifies.** (line 78) The section says defaults are intentionally authoritative in both dev and prod. If defaults are the intended set, why emit `log.warn('MYSTICAT_WORKSPACE_REPOS not set, using built-in defaults', ...)` at all? The fix moves the warn out of the hot path but leaves it firing on every actual webhook for an explicitly intentional condition.

### Minor (Nice to Have)

8. **Header metadata format diverges from existing specs.** (line 5) Copilot already flagged this; concurring. The other two specs in `docs/specs/` use a bullet-list header (`**Date**`, `**Author**`, `**Status**`, `**Jira**`, `**Target repo**`). Adopting the existing format is a 30-second edit.

9. **Undefined enumeration labels.** (line 142 and line 204) Copilot flagged the `Decision (C)` reference; the same defect applies to "scope decision (i)" at line 142. Both look like leftovers from a brainstorming or decision log. The two different numbering schemes (Roman lowercase vs uppercase letter) hint at two different lost contexts.

10. **Helper placement is too vague for a shared dependency.** (line 144) "In `llmo-query-handler.js` or a sibling util." The four callers span `llmo.js` and `llmo-query-handler.js`; if the helper lives inside the query handler, the import direction is awkward. Name a dedicated location (e.g., extend `llmo-utils.js` or introduce `llmo-source.js`).

11. **`notProvisioned: true` couples the helper's contract to today's backend semantics.** (line 149) If the upstream source is ever swapped, upstream-404 may not always mean "not provisioned." A backend-agnostic name like `noData: true` keeps the helper's contract about HTTP semantics; the caller maps to "not provisioned" if it wants to.

12. **Fix 1 leaves the eager-instantiation anti-pattern unflagged.** (line 58) The root cause is that `src/index.js:262` instantiates every controller per request before routing. Worth a one-line forward-looking note that future controllers should avoid side effects in their factory function.

13. **info-level logging on every unprovisioned request becomes background noise.** (line 167) Dashboards polling, scheduled audits - in steady state that is meaningful Coralogix volume across many unprovisioned sites. Either drop to `debug` (with the metric carrying the durable signal) or rate-limit/aggregate.

## Recommendations

- **Cite the impact size for Fix 2.** Fix 1 is quantified (~94k warns / 20min). Fix 2 mentions ~253 errors / 20 min for the LLMO contribution - worth stating that as the alert-firing delta this PR removes so the impl-PR reviewer can prioritize.
- **Add a "post-merge verification" section.** Two of these fixes are observable via Coralogix log-rate deltas. The spec already establishes baselines; stating "post-merge verification: re-query the same Coralogix DataPrime queries; confirm rates drop to X" closes the loop and is consistent with how the spec already uses Coralogix evidence.
- **Cross-link to the alert config.** The spec hinges on the SKYSI alert threshold and window but does not link to the alert definition.
- **Decide whether "graceful upstream-404 -> empty success" is a pattern or a one-off.** This service has multiple upstream-proxy paths beyond LLMO. If LLMO-only, say so to bound future application; if generalize, the helper's home should reflect that.

- The `log.warn` -> `log.info`/`log.debug` reclassification for the defaults-are-intended case (Important 7) could be its own tiny PR if you want to keep this spec focused on the call-site move; either is fine.

## Assessment

Ready to merge? **With fixes.** The architectural choices (hybrid reclassification, scope fencing, no-change-with-rationale for Fix 3) are sound and forward-looking. The blockers are clarifications - canonical fetch behavior across the four call sites, the empty-response envelope contract, the 5xx mapping, the metric mechanism, and the response discriminator. None requires redesign; all require making explicit what is currently implicit. Note: CI build is currently pending, not failing - no concern there for a docs-only PR.

## Next Steps

1. Address the 7 Important findings (canonical fetch behavior, empty envelope, 5xx mapping, metric mechanism, one-way door, validation gate, Fix 1 warn).
2. Resolve the 6 Minor items (Copilot's three plus helper placement, naming, eager-instantiation note, info-log volume).
3. Recommendations are optional but the post-merge verification section is worth a paragraph.

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
```diff
@@ -0,0 +1,247 @@
...
+**New helper** (in `llmo-query-handler.js` or a sibling util, exported):
+
+```
+fetchLlmoSource(context, url) ->
+  - 2xx                     -> { status, data: <parsed json>, headers }
+  - 404                     -> { status: 404, notProvisioned: true }   // no throw
+  - other non-OK / timeout  -> throw Error   // preserve current error semantics
+```
+
+The helper **owns and preserves all existing fetch behavior**: the `LLMO_HLX_API_KEY`
```
LeChef318: **Important: "preserve all existing fetch behavior" is undefined - the four call sites do not have consistent behavior today.**

Verified against base branch:
- `llmo-query-handler.js` has the 15s `AbortController` timeout, the `LLMO_HLX_API_KEY` presence check, and the `AbortError -> Request timeout` mapping.
- The three sheet endpoints in `llmo.js` (`getLlmoSheetData`, `queryLlmoSheetData`, `getLlmoGlobalSheetData`) do NOT have a timeout and use `env.LLMO_HLX_API_KEY || 'hlx_api_key_missing'` to paper over a missing key (so the source returns 401 instead of the controller throwing at the boundary).

If the new helper adopts the `llmo-query-handler.js` semantics that this paragraph lists, then for the three sheet endpoints this PR silently:
- adds a 15s timeout where none existed (some legitimate large fetches may now abort that succeeded before)
- changes missing-key behavior from "send bogus token, source returns 401, controller catches" to "throw at the API-key check"

Both are arguably improvements but they are not "preserved behavior." An implementer cannot resolve this from the spec alone.

**Fix:** State explicitly which behaviors are being unified, which call sites change, and add a mitigation step (e.g., test the largest current sheet-data payloads against 15s in dev before flipping).

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
```diff
@@ -0,0 +1,247 @@
...
+presence check, the `Authorization: token` header, `Accept-Encoding: br`, the internal 15s
+`AbortController` timeout, and the `AbortError -> "Request timeout"` mapping. Callers do not
+manage the timeout.
+
+**Each endpoint** routes its source fetch through `fetchLlmoSource` and, on
+`notProvisioned`, returns its endpoint-appropriate **empty success** payload instead of
+throwing. The empty payload must mirror the shape the source itself returns for a
+genuinely-empty sheet (capture a real empty response to confirm the exact envelope,
```
LeChef318: **Important: The empty-response envelope is the highest-risk implementation decision and the spec defers it.**

There are three distinct response shapes at stake (single-file, multi-sheet, files-array), the synthesized payload must be byte-identical to what clients currently receive for genuinely-empty data, and the `:version` field is a moving target:
- Hard-coding `:version` goes stale the moment helix-pages bumps it (tests pass forever; clients that key on it may or may not break).
- Omitting `:version` makes "synthesized empty" structurally different from "real empty", which contradicts the spec's own compatibility claim.
- Calling the source at runtime for an empty defeats the whole point of not calling it on 404.

For a design doc, this IS the contract; it should be pinned in the spec, not at implementation time.

**Fix:** Either (a) capture the three real shapes now and embed them as fixtures referenced by the spec, or (b) explicitly state that Phase 0 of the implementation is "capture-and-pin" and the spec gets updated before code lands. Leaving it as "verify later" makes the spec unimplementable without a parallel investigation that is itself the design decision.

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
```diff
@@ -0,0 +1,247 @@
...
+`log.error` and their current non-2xx response.
+
+### Behavior matrix
+
+| Upstream condition | Before | After |
+|--------------------|--------|-------|
+| 2xx with rows | 200 data | 200 data (unchanged) |
+| 2xx empty | 200 empty | 200 empty (unchanged) |
+| 404 (not provisioned) | 400 + log.error | **200 empty + log.info + metric** |
+| 5xx / timeout | 400 + log.error | 4xx/5xx + log.error (unchanged semantics) |
```
LeChef318: **Important: Behavior-matrix vs prose conflict on 5xx/timeout HTTP status mapping.**

Copilot already flagged this; concurring and promoting in severity. The prose at lines 168-169 says 5xx/timeout/auth keep "their current non-2xx response" - and the current behavior (from the Problem section) is `badRequest(...)` = 400. This row in the After column says "4xx/5xx + log.error (unchanged semantics)" which implies mapping to the upstream status.

The two statements cannot both be right. The implementer cannot resolve "preserve current 400" against "map to 4xx/5xx" without a decision.

The fix is also a natural moment to map upstream 5xx -> 502 and timeout -> 504, which is arguably more correct than the current 400-on-upstream-500.

**Fix:** Pick the intended mapping and update both the prose and the matrix to match.

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
```diff
@@ -0,0 +1,247 @@
...
+```json
+{ ":type": "sheet", ":version": <copy from a real empty source response>, "data": [], "total": 0, "offset": 0, "limit": 0 }
+```
+
+(Multi-sheet / files-array endpoints return the empty variant matching their normal success
+shape.) Implementation must verify the exact envelope against a live empty response rather
+than assume these field names.
+
+**Logging:** the 404 / not-provisioned path logs at `info` (e.g. "LLMO data not yet
+provisioned for site {siteId}, returning empty") plus a count/metric, so provisioning gaps
```
LeChef318: **Important: "plus a count/metric" introduces a new operational primitive with no established pattern named.**

A search of `src/` shows no existing CloudWatch `putMetricData` or custom-metric emission pattern in this service. So this spec is either:
- (a) introducing a new operational mechanism (Lambda -> CloudWatch metrics) which deserves its own design decision, or
- (b) leaning on Coralogix `events2metrics` derived from a structured log field, in which case the spec should specify the log shape and the events2metrics rule that backs it.

Without that specification, "add a metric" gets implemented as `log.info({ event: 'llmo_not_provisioned', siteId })` and the derived metric never gets created. The Risks section's appeal to a metric for observability then evaporates.

**Fix:** Name the mechanism and where the metric definition lives (Coralogix alert/dashboard, CDK/Terraform). Make it part of the validation gate.

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
```diff
@@ -0,0 +1,247 @@
...
+remain observable without inflating error logs. Genuine failures (5xx, timeout, auth) keep
+`log.error` and their current non-2xx response.
+
+### Behavior matrix
+
+| Upstream condition | Before | After |
+|--------------------|--------|-------|
+| 2xx with rows | 200 data | 200 data (unchanged) |
+| 2xx empty | 200 empty | 200 empty (unchanged) |
+| 404 (not provisioned) | 400 + log.error | **200 empty + log.info + metric** |
```
LeChef318: **Important: 404-as-empty is a one-way door; design a discriminator now or document why not.**

Once clients receive `200 + empty payload` for both "site is provisioned but empty" and "site not yet provisioned", they cannot distinguish those states from the response without out-of-band knowledge. This is fine today; the next requirement up the stack is almost always:
- "Show 'your data is being prepared' in the UI" - needs the distinction at the API boundary.
- "Alert when a paying customer's provisioning has been stuck for 72h" - same.

Cheap to add at design time, architecturally hard to retrofit once clients have shipped against the opaque-200 contract.

**Fix:** Either add an envelope hint (e.g., `:provisioning: "pending"` alongside `:type`) or a response header (e.g., `X-Llmo-Data-Status: not-provisioned`), or explicitly document why the team chose to expose this only via metrics/logs.

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
```diff
@@ -0,0 +1,247 @@
...
+- Integration (`test/it/`): add an LLMO query case for a provisioned-but-empty vs
+  not-provisioned site if seed data supports it.
+
+### Validation gate
+
+```bash
+npx mocha test/controllers/llmo/llmo.test.js
+npx mocha test/controllers/llmo/llmo-query-handler.test.js   # if present; else add
+npm run lint
+npm run docs:lint   # if any OpenAPI response examples change for these endpoints
```
LeChef318: **Important: This validation gate does not actually validate the spec's "must verify" claim.**

Line 163 says "Implementation must verify the exact envelope against a live empty response rather than assume these field names." This gate then lists `npx mocha`, `npm run lint`, `npm run docs:lint`. None verify the live envelope assumption - they verify unit-test stubs the implementer themselves wrote.

The workspace rule that designs MUST include validation gates with concrete runnable checks applies. Asserting on a stub the implementer crafted is circular.

**Fix:** Either add a concrete one-time validation step (e.g., "before merge, capture `curl https://main--project-elmo-ui-data--adobe.aem.live/<known-empty-folder>/<sheet>.json` and paste the response shape into the PR description; verify synthesized payload matches"), or move this from "must verify" to "documented snapshot, taken on date X".

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
```diff
@@ -0,0 +1,247 @@
...
+constructor becomes side-effect-free.
+
+- Result: the warn only fires on genuine webhook deliveries (rare), not on every request.
+  Expected reduction: ~94k/20min -> near zero.
+- The `getWorkspaceRepos` logic (validation, defaults fallback) is unchanged - only its
+  call site moves.
+
+### Non-goal: do not set the env var
+
+Coralogix confirms (2026-05-20) that **dev and prod both run on the built-in defaults
```
LeChef318, line 112: **Important: This section contradicts the warn-on-construction fix it justifies.**

This paragraph says defaults are intentionally authoritative in both dev and prod, confirmed via Coralogix. If defaults are the intended set, why emit `log.warn('MYSTICAT_WORKSPACE_REPOS not set, using built-in defaults', ...)` at all? After this fix the warn still fires on every actual webhook delivery (rare, granted, but still log noise for an explicitly intentional condition). A warn means "someone should look at this"; nobody should, because defaults are the desired state.

**Fix:** Add a bullet to Fix 1's `Change` section: either (a) demote the warn to `info` or `debug` since the condition is confirmed-intentional, or (b) only warn when the env var is present-but-invalid (the other branch in `getWorkspaceRepos`), since that is the only case where a human should act.

MysticatBot-Dev (changes requested):
This is a docs-only spec PR with one changed file. The prior reviews from Copilot and @LeChef318 are thorough (7 Important + 6 Minor findings). I'll produce a lightweight additive review focused on anything net-new.

Reviewing PR #2452 (adobe/spacecat-api-service). Docs-only PR - lightweight review with staff-engineer + architect perspectives. CI is pending (integration tests), build passing.

Hey @solaris007,

## Strengths

- **Scope discipline is exemplary.** The "Related work (NOT in this spec)" table, combined with the documented no-op for Fix 3, is exactly how incident remediations should be scoped. Each fix stands independently, and the fencing prevents scope creep during implementation.
- **Reclassification framing.** The "Guiding decision" section - that only 2xx removes from the alert count - is the critical insight that prevents false-economy cleanups (400->404) and correctly motivates the design.
- **Evidence-grounded quantification.** Fix 1 cites ~94k warns/20min from Coralogix with a clear before/after expectation. Fix 2 cites ~253 errors/20min. This makes the implementation PR's success criteria unambiguous.
- **Fix 3 as a documented no-op.** Recording the rationale for leaving the scrape 400 intact (post-PR-1777 guard, structurally-valid URLs get `ok([])`, 400 for malformed is correct REST) prevents future re-litigation.

## Issues

### Important (Should Fix)

1. **Multi-file path returns `status: 'error'` for 404 - inconsistent with the new single-file semantics.** (line 131-132) The spec notes that `queryLlmoFiles` multi-file path "is already graceful (per-file `status: 'error'` entries, no throw)." After this fix, single-file paths return empty-200 for upstream 404, but the multi-file path still returns `status: 'error'` per-file entries for the same condition. A client querying both paths for the same unprovisioned site gets `200 empty` from single-file and a `status: 'error'` entry from multi-file. The spec should declare whether multi-file 404 handling also updates to `status: 'not_provisioned'` or similar, or explicitly documents why the asymmetry is acceptable.

### Minor (Nice to Have)

2. **No rollback mechanism for the HTTP-semantics change.** (line 177) Fix 2 changes a 400 to 200 for the not-provisioned case. If the empty-200 unexpectedly breaks a downstream client (e.g., one that retries on 400 but cache-stores a 200), there's no documented rollback path. For a defense-in-depth change on an internal API this is likely acceptable, but a one-line note on rollback (e.g., "revert the helper's 404 branch to re-throw") would make the implementation PR's escape hatch explicit.

3. **Fix 1's test plan doesn't cover the negative case for `processGitHubWebhook` concurrency.** (line 91-92) The test section says "the warn only fires on genuine webhook deliveries" - but doesn't address whether two concurrent webhook deliveries both call `getWorkspaceRepos` redundantly. If the intent is "compute once per request," the current phrasing is fine. If the intent is "compute once globally at first use," the test should assert that.

## Recommendations

- **Existing reviews are thorough.** @LeChef318's 7 Important findings (fetch behavior inconsistency, empty envelope contract, 5xx mapping contradiction, metric mechanism, one-way door, validation gate, warn level) are well-analyzed and correctly prioritized. The fetch-behavior unification (finding 1 there) and 5xx-mapping decision (finding 3 there) are genuine blockers that must be resolved before implementation begins.
- **Consider a "Phase 0" implementation step.** Several findings (empty envelope capture, metric mechanism, fetch behavior audit) point at investigation work that should be completed before writing production code. Structuring the implementation plan as "Phase 0: pin contracts" then "Phase 1: implement" prevents the implementer from discovering these gaps mid-PR.

## Assessment

Ready to merge? With fixes. The architectural decisions (hybrid reclassification, defense-in-depth scope, documented no-op) are sound. The spec needs clarification on the ambiguities identified in the existing reviews - primarily the 5xx mapping decision, fetch behavior unification, and empty envelope contract - before an implementer can work from it without guessing. None requires redesign; all require making implicit decisions explicit. CI integration tests are pending but not a concern for a docs-only PR.

## Next Steps

1. Address the ambiguities flagged in @LeChef318's review (fetch behavior unification, 5xx mapping, empty envelope, metric mechanism).
2. Decide on the multi-file vs single-file 404 consistency (new finding above).
3. Minor items (header format, decision labels, helper placement) are polish - address at your discretion.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 49s | <b>Cost:</b> $1.04 | <b>Commit:</b> <code>c0d96e875d43b0ee58d80da953835c19b40c0f26</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot-Dev (approved):
This is a re-review with 1 new commit since our last review. The author has addressed all prior findings. Let me verify the changes and produce the consolidated review.

The updated spec is 346 lines - a substantial rewrite addressing all 7 Important + 6 Minor findings from the prior reviews. This is a docs-only PR with one file, so I'll produce a lightweight review focused on whether the feedback was properly addressed.

Re-reviewing PR #2452 (adobe/spacecat-api-service) - 1 new commit since last review. Docs-only PR - lightweight review with staff-engineer + architect perspectives. CI passing (one flaky it-postgres run, unrelated to docs change).

Hey @solaris007,

## Strengths

- **Every prior finding addressed.** The revision tackles all 7 Important and 6 Minor items from the prior reviews plus both Copilot findings - a comprehensive response that significantly improves the spec's implementation-readiness.
- **Previously flagged issues now addressed:**
  - Fetch behavior inconsistency (now has explicit "Fetch behavior is NOT uniform today" section with per-call-site audit)
  - Empty-response envelope deferral (now pinned as Phase 0 fixture-capture step)
  - 5xx/timeout mapping contradiction (behavior matrix and prose now agree: 5xx->502, timeout->504, 4xx passthrough)
  - Metric mechanism undefined (now specifies Coralogix events2metrics keyed on structured log event)
  - One-way door / missing discriminator (now has `x-llmo-data-status: not-provisioned` header)
  - Validation gate circular (now has curl-based live envelope capture in Phase 0)
  - Fix 1 warn contradiction (now: warn only for present-but-invalid, debug for unset->defaults)
  - Multi-file inconsistency (now returns `status: 'no_data'` for 404, distinct from `'error'`)
  - Header format (now matches existing spec template)
  - Decision labels removed (no more orphaned "Decision (C)" references)
  - Helper placement specified (`src/controllers/llmo/llmo-source.js`)
  - Naming (`noData` - backend-agnostic)
  - Eager-instantiation forward note added
  - Info-log volume (now emits at `debug` with events2metrics, with Phase 0 validation)
  - Rollback mechanism documented
  - Post-merge verification section added
- **Phase 0 / Phase 1 split** is the right structural answer to the "contracts must be pinned" concern. The three Phase 0 steps (capture fixtures, audit timeout, confirm metric ingestion) are each concrete and independently verifiable.
- **LLMO-only scope boundary** is explicitly stated, preventing premature generalization while leaving the door open.
- **The helper's `noData` contract** is correctly backend-agnostic - the caller interprets, the helper reports HTTP semantics.

## Issues

No issues found. The spec is now internally consistent, implementation-ready, and explicitly documents every decision that prior reviews flagged as implicit.

## Assessment

Ready to merge? Yes. The revision comprehensively addresses all prior feedback. The spec is clear enough for an implementer to work from without guessing - the Phase 0/1 split, explicit behavior matrix, fixture-pinning strategy, and discriminator header design are all well-specified. The one failing CI check (`it-postgres`) is a flaky integration test unrelated to this docs-only change.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 51s | <b>Cost:</b> $0.58 | <b>Commit:</b> <code>21e6e22abedeadc81cd82798037852a0fb84b02f</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
```diff
@@ -0,0 +1,247 @@
+# SITES-43989 Error-Rate Remediation - Inline Fixes
+
+Status: proposed
+Last verified: 2026-05-20
+Jira: [SITES-43989](https://jira.corp.adobe.com/browse/SITES-43989)
```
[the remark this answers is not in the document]
solaris007: Aligned the header to the repo spec template (Date/Author/Status/Jira/Target repo) in 21e6e22.

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
```diff
@@ -0,0 +1,247 @@
...
+npx mocha test/controllers/llmo/llmo-query-handler.test.js   # if present; else add
+npm run lint
+npm run docs:lint   # if any OpenAPI response examples change for these endpoints
+```
+
+---
+
+## Fix 3 - Scrape API: no change (documented rationale)
+
+Decision (C) deliberately leaves `getScrapeUrlByProcessingType` (`scrapeJob.js:203`)
```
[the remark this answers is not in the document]
solaris007: Removed the dangling label. Fix 3 now references the self-contained "hybrid reclassification policy" defined in Guiding decision by name. 21e6e22

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
```diff
@@ -0,0 +1,247 @@
...
+
+### Behavior matrix
+
+| Upstream condition | Before | After |
+|--------------------|--------|-------|
+| 2xx with rows | 200 data | 200 data (unchanged) |
+| 2xx empty | 200 empty | 200 empty (unchanged) |
+| 404 (not provisioned) | 400 + log.error | **200 empty + log.info + metric** |
+| 5xx / timeout | 400 + log.error | 4xx/5xx + log.error (unchanged semantics) |
+| 401 / 403 | as-is | as-is (unchanged) |
```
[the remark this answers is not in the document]
solaris007: Defined explicitly: upstream 5xx -> 502, timeout -> 504, non-404 4xx passthrough, 404 -> empty 200. Prose and the behavior matrix now agree. 21e6e22

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
solaris007: Confirmed against base and documented. New "Fetch behavior is NOT uniform today" section spells it out: the query-handler has the 15s timeout + explicit key check; the three sheet endpoints (llmo.js:231/334/462) have neither and use the `|| 'hlx_api_key_missing'` bogus-token fallback. The shared helper unifies to the stricter behavior, explicitly flagged as a behavior change for the sheet endpoints, with a Phase 0 step to test the largest current sheet payloads against 15s before adopting. 21e6e22

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
solaris007: Pinned. Phase 0 step 1 captures real empty-response fixtures for all three shapes (single-file, multi-sheet, files-array) and commits them; the synthesized payloads must byte-match the fixtures, and tests assert against the fixtures rather than hand-written stubs. 21e6e22

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
solaris007: Resolved together with the Copilot 5xx finding: upstream 5xx -> 502, timeout -> 504, non-404 4xx passthrough, 404 -> empty 200. Matrix and prose aligned. 21e6e22

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
solaris007: Specified a Coralogix events2metrics rule keyed on a structured `event: 'llmo_data_not_provisioned'` log field (confirmed there is no `putMetricData` pattern in `src/`), authored in Phase 0. The log is emitted at `debug` to avoid steady-state info volume, contingent on Phase 0 confirming debug logs reach Coralogix ingestion (else `info` + rely on the aggregate). 21e6e22

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
solaris007: Added an `x-llmo-data-status: not-provisioned` response header on the empty-200 path. The body stays byte-identical to a real empty sheet (compatibility), and the header carries the provisioned-empty vs not-provisioned distinction. 21e6e22

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
solaris007: Fixed. The gate now includes a one-time Phase 0 `curl` capture of a live empty response into a committed fixture, and tests assert the synthesized payload equals that fixture rather than a self-authored stub. 21e6e22

docs/specs/2026-05-20-sites-43989-error-rate-remediation.md
solaris007: Addressed. Fix 1 now warns only on a present-but-invalid `MYSTICAT_WORKSPACE_REPOS`; the unset -> defaults branch (confirmed-intended) drops to `debug`, so the Non-goal and the warn no longer contradict. 21e6e22

solaris007:
Revised in `21e6e22`. Beyond the inline threads:

**@LeChef318 minors:** helper is now a dedicated module `src/controllers/llmo/llmo-source.js` imported by both callers; `notProvisioned` -> `noData` (helper contract stays about HTTP semantics, caller maps to "not provisioned"); added an eager-instantiation forward-note (factories should stay side-effect-free); not-provisioned log dropped to `debug` with the events2metrics rule carrying the durable signal.

**MysticatBot:** multi-file 404 now returns per-file `status: 'no_data'` (distinct from `'error'`), consistent with the single-file empty-200; added a Rollback section (revert the helper's 404 branch to re-throw); clarified `getWorkspaceRepos` is computed once per request (no global cache intended).

**Recommendations:** cited the ~253/20min not-provisioned contribution as the alert-firing delta; added a Post-merge verification section (re-query the Coralogix baselines); cross-linked the alert (SKYSI-76262); marked upstream-404 -> empty success as LLMO-only (not yet a service-wide pattern).

Implementation is now split into **Phase 0 (pin contracts)** - empty fixtures, fetch audit + 15s payload check, events2metrics rule - and **Phase 1 (implement)**.

PR #1797, opened 2026-02-11
fix: edge-optimize API for fastly CDN routing configuration

dipratap:
Adds an API - POST /sites/{siteId}/llmo/edge-optimize-cdn-routing
   * Updates edge optimize CDN routing for the site via the internal CDN API.
       * Probes the site with User-Agent AdobeEdgeOptimize-Test (must return 2xx)
       * Calls internal CDN API to update edge optimize CDN routing for the domain
   * Only available when EDGE_OPTIMIZE_CDN_API_BASE_URL is set (e.g. prod). Disabled in dev.
   * Requires authInfo profile.promiseToken to exchange for IMS user token.

PR #2455, opened 2026-05-21
feat(scrape): server-side per-site auth resolution on /tools/scrape/jobs (SITES-40597)

aenascut:
## Summary

Extends `POST /tools/scrape/jobs` to resolve the per-site `Authorization` header server-side when `options.enableAuthentication === true`. No new endpoint, no new route. Callers that don't set the flag take the legacy fast-path — byte-identical to today.

## Why

The downstream `spacecat-content-scraper` Lambda's SQS Event Source Mappings target the **unqualified** Lambda ARN, so `helix-universal` reports `ctx.func.version === '$LATEST'`. AWS Secrets Manager rejects `$` in secret names with `ValidationException`, which breaks the existing content-scraper auth path. Long-term fix is [`spacecat-shared#1614`](https://github.com/adobe/spacecat-shared/pull/1614) (normalize `$LATEST → latest`); that PR is blocked on review.

This PR **bypasses** the bug by baking the header in `spacecat-api-service`, which is API-Gateway-invoked and therefore alias-qualified (`ctx.func.version === 'latest'`) — the same Lambda `createBetaPreflightJob` already uses for identical resolution (`controllers/preflight.js:405-429`).

## Design

When `options.enableAuthentication === true`:

```
POST /tools/scrape/jobs { urls, metaData:{siteId}, options:{enableAuthentication:true}, customHeaders }
  → validate metaData.siteId           (400 otherwise — fail loud at API edge)
  → Site.findById(siteId)              (404 otherwise)
  → AccessControlUtil.hasAccess(site)  (403 otherwise)
  → HEAD <previewBaseURL>
       401/403 → resolvePromiseToken (cookie | IMS) + retrievePageAuthentication
                 → Authorization injected into customHeaders
       2xx     → no header (transparent delegation)
       error   → log + delegate without auth (best-effort)
  → strip options.enableAuthentication
  → scrapeClient.createScrapeJob(...)
```

When the flag is unset: legacy fast-path. No site lookup, no HEAD probe, no Secrets Manager call.

## Defensive properties

- Legacy flow byte-identical for non-auth callers.
- No changes required in `spacecat-shared`, `spacecat-content-scraper`, `spacecat-scrape-job-manager`, or `spacecat-infrastructure`.
- `metaData.siteId` required when auth requested — loud-fail at the edge instead of silent downstream misroute.
- `options.enableAuthentication` is **always stripped** before delegation — prevents content-scraper from re-resolving under `$LATEST`.
- HEAD probe failures degrade to legacy behavior (best-effort).

## Dependency chain

| # | Repo | PR | Role |
|---|---|---|---|
| **this** | `adobe/spacecat-api-service` | #2455 | Server-side auth resolution on `/tools/scrape/jobs` |
| 2 | `adobe-rnd/llmo-data-retrieval-service` | [#1797](https://github.com/adobe-rnd/llmo-data-retrieval-service/pull/1797) | DRS forwards `site_id` + `options.enableAuthentication`; Pydantic validator catches misconfig |
| 3 | `experience-platform/mystique` | [#2255](https://git.corp.adobe.com/experience-platform/mystique/pull/2255) | Per-site `drs_provider` routing; zero-config sites keep BrightData |

### Unit tests

21 tests on `createScrapeJob` — **100% line/branch/function/statement coverage** on `src/controllers/scrapeJob.js`:

- **Legacy fast-path (10)** — unchanged existing tests.
- **Auth-resolution branch (11)** — 400 on missing/invalid siteId, 404 on site not found, 403 on no-access, 500 on invalid baseURL, header injection (`token` for non-promise sites, `Bearer` for AEM_CS + promise token, IMS fallback for CS_CW), 400/500 propagation from promise-token errors, 500 on `retrievePageAuthentication` failure, transparent delegation when HEAD returns 2xx, graceful degradation when HEAD throws, customHeaders merging, `options.enableAuthentication` stripping, 400/500 propagation from delegated `scrapeClient`.

### E2E validation in dev

Direct Lambda invoke against branch alias `latest26214458247` (v6439) — public Fastly `/api/ci` routes to the `:ci` alias which only advances on main-merge, so branch testing uses `aws lambda invoke` with a synthesized API Gateway v2 event.

#### Pre-review smoke matrix

| # | Case | Job ID | Result |
|---|---|---|---|
| 1 | Auth path, single Okta URL | `b5f79aea-9687-4a39-bbf1-badcc839652b` | ✅ COMPLETE in 11.1s — `enableAuthentication` stripped from persisted options |
| 2 | **Legacy regression** — no `enableAuthentication` (adobe.com) | `924425e8-ba27-435a-a304-8ab7ae4f05b5` | ✅ COMPLETE in 18.6s — zero-change for non-auth callers confirmed |
| 3 | 400 boundary — `enableAuthentication=true`, no `siteId` | _(no job)_ | ✅ HTTP 400 `"Invalid request: metaData.siteId is required when options.enableAuthentication is true"` |
| 4 | Multi-URL auth (3 Okta URLs) | `c65f373c-1158-4997-8e7d-3f7325ba85c4` | ✅ 2 COMPLETE + 1 FAILED (upstream HTTP 404 — not auth) — per-URL errors isolated, auth header propagates across batch |

#### SITES-40597 incident replay (earlier in PR lifecycle)

DRS production job [`709e64a1-f974-491e-a23b-db199a9f9041`](https://d2hp6j4u6kak3w.cloudfront.net/jobs/709e64a1-f974-491e-a23b-db199a9f9041) failed with BrightData `parent_s3_key_not_found`. Replayed the same 6 Okta URLs through this PR's endpoint (job `c7cf1bbf-222f-45f7-8593-c48012c89b93`): **6/6 success, 0 failures, ~75 s**. Verified scraped JSON payloads — HTTP 200, finalUrl matches, real page HTML, no login-wall redirects.

### Auth-injection branch caveat

Tested Okta marketing URLs return 200/301 on HEAD, so the E2E exercised the **transparent-delegation** path (no header injected, no Secrets Manager call). The full auth-injection branch (HEAD 401/403 → IMS exchange → `retrievePageAuthentication` → `Bearer` vs `token`) is exercised by the unit tests above. The same code already runs in production via `createBetaPreflightJob` (`controllers/preflight.js:405-429`).

## Test plan

- [x] `npm run lint` — clean
- [x] `npm test` — 10303 passing
- [x] `npm run docs:lint` / `docs:build` — clean
- [x] 100% coverage on `src/controllers/scrapeJob.js`
- [x] E2E pre-review smoke matrix (above)
- [x] SITES-40597 BrightData incident replay — 6/6 URLs scraped
- [ ] Manual smoke on `/api/ci/tools/scrape/jobs` after merge (when `:ci` alias is promoted)
- [ ] DRS #1797 + Mystique #2255 merged → per-site `drs_provider` flip for Okta enables production traffic on this path

Refs: SITES-40597

### DIFF
--- a/docs/openapi/scrape-api.yaml
+++ b/docs/openapi/scrape-api.yaml
@@ -30,6 +30,63 @@ create-scrape-job:
       '500':
         $ref: './responses.yaml#/500'
 
+create-scrape-authenticated-job:
+  post:
+    tags:
+      - scrape
+    security:
+      - api_key: [ ]
+    summary: Create new authenticated Scrape Job
+    description: |
+      ⚠️ **EXPERIMENTAL / PROTOTYPE**
+
+      Same as `POST /tools/scrape/jobs`, but the per-site `Authorization` header
+      is resolved server-side and injected into `customHeaders` before the job is
+      queued. Intended for sites whose pages require an `Authorization` header
+      (e.g. Okta) that cannot be carried by the standard browser flow.
+
+      The server:
+        1. Looks up the site by `metaData.siteId` (required, UUID).
+        2. Probes the site with a `HEAD` request. If the response is 401/403:
+           - For promise-based authoring types (CS, CS_CW, AMS) it resolves a
+             promise token from the `promiseToken` cookie or by IMS exchange.
+           - It calls `retrievePageAuthentication` and constructs an
+             `Authorization` header (`Bearer ...` for AEM_CS with promise token,
+             `token ...` otherwise).
+           - The resolved header is merged into the request's `customHeaders`.
+        3. Strips `options.enableAuthentication` from the delegated payload so
+           the downstream worker does not attempt to re-resolve the token.
+        4. Delegates to the same scrape pipeline used by `POST /tools/scrape/jobs`.
+
+      If the `HEAD` probe returns 2xx (no auth needed), the request is delegated
+      transparently with no header injection — safe to use for sites that may or
+      may not require authentication.
+    operationId: createScrapeAuthenticatedJob
+    requestBody:
+      content:
+        application/json:
+          schema:
+            $ref: './schemas.yaml#/AuthenticatedScrapeRequest'
+    responses:
+      '202':
+        description: Job accepted
+        content:
+          application/json:
+            schema:
+              $ref: './schemas.yaml#/CreateAsyncJobAcceptedResponse'
+      '400':
+        description: |
+          Invalid request — missing or invalid `metaData.siteId`, malformed
+          payload, or rejected by the scrape pipeline validator.
+      '401':
+        $ref: './responses.yaml#/401'
+      '403':
+        description: Caller does not have access to the specified site.
+      '404':
+        description: Site not found for the provided `metaData.siteId`.
+      '500':
+        $ref: './responses.yaml#/500'
+
 get-scrape-job-status:
   parameters:
     - $ref: './parameters.yaml#/jobId'
--- a/docs/openapi/scrape-api.yaml
+++ b/docs/openapi/scrape-api.yaml
@@ -45,6 +45,16 @@ create-scrape-job:
         description: Site not found for the provided `metaData.siteId` (only when `options.enableAuthentication` is true).
       '500':
         $ref: './responses.yaml#/500'
+      '502':
+        description: |
+          Authentication requirement could not be verified for the site's preview origin
+          (only when `options.enableAuthentication` is true). Returned when the HEAD probe
+          against the site's preview URL fails — network/DNS error, timeout, 3xx redirect
+          (manual-redirect mode), or 5xx — or when the resolved `(authoringType, deliveryType)`
+          combination is not a verified token-scheme pairing. The caller should treat this
+          as transient (retry after backoff) or, if persistent, surface to the site owner
+          to confirm the preview origin is reachable and the site is correctly configured
+          for authenticated scraping.
 
 get-scrape-job-status:
   parameters:
--- a/docs/openapi/scrape-api.yaml
+++ b/docs/openapi/scrape-api.yaml
@@ -7,9 +7,21 @@ create-scrape-job:
     summary: Create new Scrape Job
     description: |
       ⚠️ **EXPERIMENTAL / PROTOTYPE**
-      This endpoint allows you to create a new async url scrape job.
-      The job is provided a list URLs to be scraped. The URLs are fed into a queue and processed 1 at a time.
-      The returned jobId can be used to poll for the status and the results of the job.
+
+      Create a new async URL scrape job. URLs are fed into a queue and processed
+      one at a time. The returned `jobId` can be used to poll for status and results.
+
+      ### Optional per-site authentication
+
+      When `options.enableAuthentication` is `true`, the server resolves a
+      per-site `Authorization` header (HEAD probe + optional IMS promise-token
+      exchange + `retrievePageAuthentication`), injects it into `customHeaders`,
+      and strips `options.enableAuthentication` before delegating to the scrape
+      pipeline. In this mode `metaData.siteId` is required (returns `400`
+      otherwise); the site must exist and the caller must have access to it.
+
+      Callers that do not set `enableAuthentication` get the legacy behavior —
+      no site lookup, no HEAD probe, no Secrets Manager call.
     operationId: createScrapeJob
     requestBody:
       content:
@@ -27,63 +39,10 @@ create-scrape-job:
         $ref: './responses.yaml#/400-max-scrape-urls-exceeded'
       '401':
         $ref: './responses.yaml#/401'
-      '500':
-        $ref: './responses.yaml#/500'
-
-create-scrape-authenticated-job:
-  post:
-    tags:
-      - scrape
-    security:
-      - api_key: [ ]
-    summary: Create new authenticated Scrape Job
-    description: |
-      ⚠️ **EXPERIMENTAL / PROTOTYPE**
-
-      Same as `POST /tools/scrape/jobs`, but the per-site `Authorization` header
-      is resolved server-side and injected into `customHeaders` before the job is
-      queued. Intended for sites whose pages require an `Authorization` header
-      (e.g. Okta) that cannot be carried by the standard browser flow.
-
-      The server:
-        1. Looks up the site by `metaData.siteId` (required, UUID).
-        2. Probes the site with a `HEAD` request. If the response is 401/403:
-           - For promise-based authoring types (CS, CS_CW, AMS) it resolves a
-             promise token from the `promiseToken` cookie or by IMS exchange.
-           - It calls `retrievePageAuthentication` and constructs an
-             `Authorization` header (`Bearer ...` for AEM_CS with promise token,
-             `token ...` otherwise).
-           - The resolved header is merged into the request's `customHeaders`.
-        3. Strips `options.enableAuthentication` from the delegated payload so
-           the downstream worker does not attempt to re-resolve the token.
-        4. Delegates to the same scrape pipeline used by `POST /tools/scrape/jobs`.
-
-      If the `HEAD` probe returns 2xx (no auth needed), the request is delegated
-      transparently with no header injection — safe to use for sites that may or
-      may not require authentication.
-    operationId: createScrapeAuthenticatedJob
-    requestBody:
-      content:
-        application/json:
-          schema:
-            $ref: './schemas.yaml#/AuthenticatedScrapeRequest'
-    responses:
-      '202':
-        description: Job accepted
-        content:
-          application/json:
-            schema:
-              $ref: './schemas.yaml#/CreateAsyncJobAcceptedResponse'
-      '400':
-        description: |
-          Invalid request — missing or invalid `metaData.siteId`, malformed
-          payload, or rejected by the scrape pipeline validator.
-      '401':
-        $ref: './responses.yaml#/401'
       '403':
-        description: Caller does not have access to the specified site.
+        description: Caller does not have access to the specified site (only when `options.enableAuthentication` is true).
       '404':
-        description: Site not found for the provided `metaData.siteId`.
+        description: Site not found for the provided `metaData.siteId` (only when `options.enableAuthentication` is true).
       '500':
         $ref: './responses.yaml#/500'
 
--- a/src/controllers/scrapeJob.js
+++ b/src/controllers/scrapeJob.js
@@ -99,8 +99,23 @@ function ScrapeJobController(context) {
     return 'ambiguous';
   }
 
+  // Mirrors `@adobe/spacecat-shared-ims-client/src/auth.js` (the promise-
+  // exchange gate inside `retrievePageAuthentication`): the ims-client gates
+  // promise-token exchange on `(promiseTypes.includes(authoringType) ||
+  // deliveryType === AEM_CS) && authOptions.promiseToken`. Keeping this
+  // predicate identical avoids the controller and the ims-client drifting on
+  // "is this a promise-path site?" — that asymmetry is silent today (the
+  // ims-client falls through to the Secrets Manager PAT branch when no
+  // promise token is provided) but would surface for a customer who adds
+  // AEM_CS delivery on a non-promise authoring type, or vice versa.
+  // Extraction to a shared util is tracked in SITES-45204.
+  function isPromisePathSite(site) {
+    return PROMISE_BASED_TYPES.includes(site.getAuthoringType())
+      || site.getDeliveryType() === DELIVERY_TYPES.AEM_CS;
+  }
+
   async function resolvePromiseToken(site, requestContext) {
-    if (!PROMISE_BASED_TYPES.includes(site.getAuthoringType())) {
+    if (!isPromisePathSite(site)) {
       return null;
     }
     const cookieToken = getCookieValue(requestContext, 'promiseToken');
@@ -199,37 +214,38 @@ function ScrapeJobController(context) {
     }
 
     const baseURL = site.getBaseURL();
-    let siteHostname;
+    let siteOrigin;
     let previewBaseURL;
     try {
       const parsed = new URL(baseURL);
-      siteHostname = parsed.hostname;
+      siteOrigin = parsed.origin;
       previewBaseURL = `${parsed.protocol}//${parsed.hostname}`;
     } catch {
       log.error(`Site ${siteId} has invalid baseURL: ${baseURL}`);
       return internalServerError('Site has invalid baseURL');
     }
 
-    // Bind every caller URL to the site origin before we mint a credential —
-    // otherwise a caller can submit a valid siteId with urls pointing at an
-    // attacker host and harvest the resolved Authorization header (cross-
-    // origin credential exfiltration). Mirrors the preflight controller's
-    // "all urls must belong to the same website" guard, but tightened to
-    // the *site's* origin rather than just same-hostname-as-first-url.
+    // Bind every caller URL to the site's *origin* before we mint a credential
+    // — otherwise a caller can submit a valid siteId with URLs pointing at an
+    // attacker host (or an adjacent service on a different port, or the HTTP
+    // counterpart of an HTTPS site) and harvest the resolved Authorization.
+    // Comparing `URL.origin` (scheme + host + port) closes all three vectors;
+    // `.hostname` alone misses cross-port and the http/https downgrade. Other
+    // subdomains are rejected — promote them to their own site if needed.
     if (!Array.isArray(data.urls) || data.urls.length === 0) {
       return badRequest('Invalid request: urls must be a non-empty array');
     }
     for (const u of data.urls) {
-      let urlHostname;
+      let urlOrigin;
       try {
-        urlHostname = new URL(u).hostname;
+        urlOrigin = new URL(u).origin;
       } catch {
         return badRequest(`Invalid request: malformed URL: ${u}`);
       }
-      if (urlHostname !== siteHostname) {
-        log.warn(`Cross-origin URL rejected for site ${siteId}: ${u}`);
+      if (urlOrigin !== siteOrigin) {
+        log.warn(`Cross-origin URL rejected for site ${siteId}: ${u} (site origin: ${siteOrigin})`);
         return badRequest(
-          `Invalid request: every URL must belong to the site origin (${siteHostname})`,
+          `Invalid request: every URL must belong to the site origin (${siteOrigin})`,
         );
       }
     }
@@ -250,6 +266,28 @@ function ScrapeJobController(context) {
 
     let authorizationHeader;
     if (probeResult === 'auth-required') {
+      // Guard `(promise-authoring) × (non-AEM_CS delivery)`: IMS access tokens
+      // belong under `Bearer` (RFC 6750), AEM EDS endpoints expect `token <PAT>`
+      // for static-PAT auth. Minting `token <IMS-JWT>` is unverified and the
+      // E2E in PR #2455 didn't exercise it (HEAD 200/301 sites). Reject with
+      // 502 until a per-site auth-policy attribute drives scheme selection,
+      // rather than guess. Promise-authoring sites whose delivery is AEM_CS
+      // get `Bearer <IMS-token>` correctly below.
+      if (PROMISE_BASED_TYPES.includes(site.getAuthoringType())
+          && site.getDeliveryType() !== DELIVERY_TYPES.AEM_CS) {
+        log.warn(
+          `Refusing to mint Authorization for site ${siteId}: `
+          + `authoringType=${site.getAuthoringType()} × deliveryType=${site.getDeliveryType()} `
+          + 'is not a verified scheme combination',
+        );
+        return createResponse({}, 502, {
+          [HEADER_ERROR]: cleanupHeaderValue(
+            `Authentication scheme for site ${siteId} is not yet supported (`
+            + `${site.getAuthoringType()} × ${site.getDeliveryType()})`,
+          ).slice(0, 500),
+        });
+      }
+
       let promiseTokenObj;
       try {
         promiseTokenObj = await resolvePromiseToken(site, requestContext);
--- a/src/controllers/scrapeJob.js
+++ b/src/controllers/scrapeJob.js
@@ -61,14 +61,42 @@ function ScrapeJobController(context) {
 
   const MAX_JOBS_BY_BASEURL = 100;
 
-  // Duplicated from preflight controller (decision: refactor to support/scrape-auth.js
-  // as follow-up once the second consumer ships and the abstraction is justified).
+  // Duplicated from preflight controller. Extraction tracked in SITES-45204
+  // (move checkEnableAuthentication / resolvePromiseToken / PROMISE_BASED_TYPES
+  // to src/support/auth.js, alongside getIMSPromiseToken).
+  //
+  // Hardened HEAD probe — 3 s timeout (under API Gateway's 29 s budget for an
+  // outbound to arbitrary customer infrastructure) and redirect: 'manual' so
+  // a customer baseURL cannot 30x-launder the probe toward an internal host.
+  // Returns one of:
+  //   'auth-required'  — 401 / 403, mint and inject the Authorization header
+  //   'no-auth'        — 2xx,        delegate without injecting a header
+  //   'ambiguous'      — 3xx / 5xx / network / timeout, caller asked for auth
+  //                     but we cannot prove the page wants it; fail closed at
+  //                     the caller instead of silently scraping a login page.
+  const HEAD_PROBE_TIMEOUT_MS = 3000;
   async function checkEnableAuthentication(previewBaseURL) {
-    const headResponse = await fetch(previewBaseURL, {
-      method: 'HEAD',
-      headers: { 'Content-Type': 'application/json' },
-    });
-    return headResponse.status === 401 || headResponse.status === 403;
+    let headResponse;
+    try {
+      headResponse = await fetch(previewBaseURL, {
+        method: 'HEAD',
+        headers: { 'Content-Type': 'application/json' },
+        redirect: 'manual',
+        signal: AbortSignal.timeout(HEAD_PROBE_TIMEOUT_MS),
+      });
+    } catch (e) {
+      log.warn(`HEAD probe failed for ${previewBaseURL}: ${e.name}: ${e.message}`);
+      return 'ambiguous';
+    }
+    const { status } = headResponse;
+    if (status === 401 || status === 403) {
+      return 'auth-required';
+    }
+    if (status >= 200 && status < 300) {
+      return 'no-auth';
+    }
+    log.warn(`HEAD probe for ${previewBaseURL} returned ambiguous status ${status}`);
+    return 'ambiguous';
   }
 
   async function resolvePromiseToken(site, requestContext) {
@@ -90,6 +118,19 @@ function ScrapeJobController(context) {
     return createResponse({}, error.status || 500, {
       [HEADER_ERROR]: safeMessage,
     });
+  }
+
+  // Strip the server-minted Authorization header from the response — content-
+  // scraper still receives it via the SQS message body, but the credential
+  // must not be echoed back in the 202 body. At-rest exposure in DynamoDB /
+  // LIST endpoints is tracked in SITES-45205 (spacecat-shared ScrapeJobDto
+  // redaction).
+  function sanitizeJobResponse(job) {
+    if (!isNonEmptyObject(job) || !isNonEmptyObject(job.customHeaders)) {
+      return job;
+    }
+    const { Authorization: _, ...safeHeaders } = job.customHeaders;
+    return { ...job, customHeaders: safeHeaders };
   }
 
   /**
@@ -158,25 +199,57 @@ function ScrapeJobController(context) {
     }
 
     const baseURL = site.getBaseURL();
+    let siteHostname;
     let previewBaseURL;
     try {
       const parsed = new URL(baseURL);
+      siteHostname = parsed.hostname;
       previewBaseURL = `${parsed.protocol}//${parsed.hostname}`;
     } catch {
       log.error(`Site ${siteId} has invalid baseURL: ${baseURL}`);
       return internalServerError('Site has invalid baseURL');
     }
 
-    let needsAuth = false;
-    try {
-      needsAuth = await checkEnableAuthentication(previewBaseURL);
-    } catch (e) {
-      // HEAD probe is best-effort; on failure delegate without header injection.
-      log.info(`HEAD probe failed for ${previewBaseURL}: ${e.message}`);
+    // Bind every caller URL to the site origin before we mint a credential —
+    // otherwise a caller can submit a valid siteId with urls pointing at an
+    // attacker host and harvest the resolved Authorization header (cross-
+    // origin credential exfiltration). Mirrors the preflight controller's
+    // "all urls must belong to the same website" guard, but tightened to
+    // the *site's* origin rather than just same-hostname-as-first-url.
+    if (!Array.isArray(data.urls) || data.urls.length === 0) {
+      return badRequest('Invalid request: urls must be a non-empty array');
+    }
+    for (const u of data.urls) {
+      let urlHostname;
+      try {
+        urlHostname = new URL(u).hostname;
+      } catch {
+        return badRequest(`Invalid request: malformed URL: ${u}`);
+      }
+      if (urlHostname !== siteHostname) {
+        log.warn(`Cross-origin URL rejected for site ${siteId}: ${u}`);
+        return badRequest(
+          `Invalid request: every URL must belong to the site origin (${siteHostname})`,
+        );
+      }
+    }
+
+    // checkEnableAuthentication never throws — it returns 'auth-required',
+    // 'no-auth', or 'ambiguous'. We fail closed on 'ambiguous' because the
+    // caller explicitly asked for authenticated scraping; silently demoting
+    // would store a login page as "real content" — the exact SITES-40597
+    // failure class.
+    const probeResult = await checkEnableAuthentication(previewBaseURL);
+    if (probeResult === 'ambiguous') {
+      return createResponse({}, 502, {
+        [HEADER_ERROR]: cleanupHeaderValue(
+          `Could not verify authentication requirement for ${previewBaseURL}`,
+        ).slice(0, 500),
+      });
     }
 
     let authorizationHeader;
-    if (needsAuth) {
+    if (probeResult === 'auth-required') {
       let promiseTokenObj;
       try {
         promiseTokenObj = await resolvePromiseToken(site, requestContext);
@@ -217,7 +290,7 @@ function ScrapeJobController(context) {
 
     try {
       const job = await scrapeClient.createScrapeJob(delegatedData);
-      return accepted(job);
+      return accepted(sanitizeJobResponse(job));
     } catch (error) {
       log.error(error.message);
       if (error?.message?.includes('Invalid request')) {
--- a/src/controllers/scrapeJob.js
+++ b/src/controllers/scrapeJob.js
@@ -216,6 +216,13 @@ function ScrapeJobController(context) {
     // origin credential exfiltration). Mirrors the preflight controller's
     // "all urls must belong to the same website" guard, but tightened to
     // the *site's* origin rather than just same-hostname-as-first-url.
+    //
+    // `www.` is stripped on both sides so that a site registered as
+    // `okta.com` accepts URLs under `www.okta.com` (and vice versa) — the
+    // canonical apex/www pair has identical TLS + content guarantees. Other
+    // subdomains are rejected; promote them to their own site if needed.
+    const stripWww = (h) => h.toLowerCase().replace(/^www\./, '');
+    const normalizedSiteHost = stripWww(siteHostname);
     if (!Array.isArray(data.urls) || data.urls.length === 0) {
       return badRequest('Invalid request: urls must be a non-empty array');
     }
@@ -226,7 +233,7 @@ function ScrapeJobController(context) {
       } catch {
         return badRequest(`Invalid request: malformed URL: ${u}`);
       }
-      if (urlHostname !== siteHostname) {
+      if (stripWww(urlHostname) !== normalizedSiteHost) {
         log.warn(`Cross-origin URL rejected for site ${siteId}: ${u}`);
         return badRequest(
           `Invalid request: every URL must belong to the site origin (${siteHostname})`,
--- a/src/controllers/scrapeJob.js
+++ b/src/controllers/scrapeJob.js
@@ -94,62 +94,51 @@ function ScrapeJobController(context) {
 
   /**
    * Create and start a new scrape job.
+   *
+   * When `options.enableAuthentication === true`, the per-site `Authorization`
+   * header is resolved server-side in this Lambda (HEAD probe + optional IMS
+   * promise-token exchange + `retrievePageAuthentication`), merged into
+   * `customHeaders`, and `options.enableAuthentication` is stripped before
+   * delegation. This mirrors the `createBetaPreflightJob` pattern.
+   *
+   * Why bake the header here instead of letting content-scraper resolve it:
+   * content-scraper's SQS Event Source Mappings target the unqualified Lambda
+   * ARN, so helix-universal reports `ctx.func.version === '$LATEST'`, which
+   * AWS Secrets Manager rejects as an invalid name (`ValidationException`).
+   * The spacecat-api Lambda is API-Gateway-invoked (alias-qualified `:latest`),
+   * so `retrievePageAuthentication` works correctly here.
+   *
+   * Callers that do not set `enableAuthentication` get byte-identical legacy
+   * behavior — no Site lookup, no HEAD probe, no Secrets Manager call.
+   *
    * @param {UniversalContext} requestContext - The context of the universal serverless function.
    * @param {object} requestContext.data - Parsed json request data.
    * @returns {Promise<Response>} 202 Accepted if successful, 4xx or 5xx otherwise.
    */
   async function createScrapeJob(requestContext) {
-    const { data } = requestContext;
-
-    try {
-      const job = await scrapeClient.createScrapeJob(data);
-
-      return accepted(job);
-    } catch (error) {
-      log.error(error.message);
-      if (error?.message?.includes('Invalid request')) {
-        return badRequest(error);
-      }
-      return createErrorResponse(error);
-    }
-  }
-
-  /**
-   * Create a scrape job with the per-site Authorization header resolved server-side.
-   *
-   * Mirrors the auth-resolution path used by `createBetaPreflightJob` in preflight.js
-   * (HEAD probe + optional IMS promise-token exchange + `retrievePageAuthentication`),
-   * then delegates the resulting payload to the existing `scrapeClient.createScrapeJob`.
-   *
-   * Defensive guarantees:
-   *  - `metaData.siteId` is required; returns 400 otherwise.
-   *  - `options.enableAuthentication` is stripped from the delegated payload so the
-   *    downstream content-scraper does not attempt to re-resolve the token under an
-   *    unqualified Lambda ARN (which surfaces AWS Secrets Manager `ValidationException`
-   *    because helix-universal reports `ctx.func.version === '$LATEST'`).
-   *  - When the HEAD probe returns 2xx the endpoint delegates transparently with no
-   *    header injection — callers can safely use this endpoint for sites that may or
-   *    may not require authentication.
-   *
-   * Bakes the Authorization header in the spacecat-api Lambda (alias-qualified ARN
-   * via API Gateway, so `ctx.func.version === 'latest'`), avoiding the `$LATEST`
-   * bug entirely. Legacy `POST /tools/scrape/jobs` is unaffected.
-   *
-   * @param {UniversalContext} requestContext - The context of the universal serverless function.
-   * @param {object} requestContext.data - Parsed json request data.
-   * @param {object} requestContext.data.metaData - Required; must contain `siteId`.
-   * @returns {Promise<Response>} 202 Accepted on success; 400/403/404/500 otherwise.
-   */
-  async function createScrapeAuthenticatedJob(requestContext) {
     const { data, dataAccess } = requestContext;
 
-    if (!isNonEmptyObject(data)) {
-      return badRequest('Invalid request: missing application/json data');
+    // Legacy fast-path: no auth resolution requested, behave exactly as before.
+    if (!isNonEmptyObject(data) || data.options?.enableAuthentication !== true) {
+      try {
+        const job = await scrapeClient.createScrapeJob(data);
+        return accepted(job);
+      } catch (error) {
+        log.error(error.message);
+        if (error?.message?.includes('Invalid request')) {
+          return badRequest(error);
+        }
+        return createErrorResponse(error);
+      }
     }
 
+    // Auth-resolution path. metaData.siteId required — fail loud at the API edge
+    // rather than silently demote to an anonymous scrape that returns a login page.
     const siteId = data?.metaData?.siteId;
     if (!isValidUUID(siteId)) {
-      return badRequest('Invalid request: metaData.siteId is required and must be a valid UUID');
+      return badRequest(
+        'Invalid request: metaData.siteId is required when options.enableAuthentication is true',
+      );
     }
 
     let site;
@@ -178,16 +167,16 @@ function ScrapeJobController(context) {
       return internalServerError('Site has invalid baseURL');
     }
 
-    let enableAuthentication = false;
+    let needsAuth = false;
     try {
-      enableAuthentication = await checkEnableAuthentication(previewBaseURL);
+      needsAuth = await checkEnableAuthentication(previewBaseURL);
     } catch (e) {
-      // HEAD probe is best-effort; on failure delegate without auth (legacy behavior).
+      // HEAD probe is best-effort; on failure delegate without header injection.
       log.info(`HEAD probe failed for ${previewBaseURL}: ${e.message}`);
     }
 
     let authorizationHeader;
-    if (enableAuthentication) {
+    if (needsAuth) {
       let promiseTokenObj;
       try {
         promiseTokenObj = await resolvePromiseToken(site, requestContext);
@@ -209,8 +198,11 @@ function ScrapeJobController(context) {
       }
     }
 
-    // Strip enableAuthentication from options so content-scraper never re-resolves.
-    const { enableAuthentication: _, ...remainingOptions } = data.options || {};
+    // Always strip enableAuthentication before delegation — prevents content-scraper
+    // from re-resolving the token under its unqualified SQS ESM ARN (`$LATEST` bug).
+    // `data.options` is guaranteed non-null: we only reach this point when
+    // `data.options.enableAuthentication === true` was true above.
+    const { enableAuthentication: _, ...remainingOptions } = data.options;
 
     const delegatedData = {
       ...data,
@@ -393,7 +385,6 @@ function ScrapeJobController(context) {
 
   return {
     createScrapeJob,
-    createScrapeAuthenticatedJob,
     getScrapeJobStatus,
     getScrapeJobUrlResults,
     getScrapeJobsByBaseURL,
--- a/src/controllers/scrapeJob.js
+++ b/src/controllers/scrapeJob.js
@@ -216,13 +216,6 @@ function ScrapeJobController(context) {
     // origin credential exfiltration). Mirrors the preflight controller's
     // "all urls must belong to the same website" guard, but tightened to
     // the *site's* origin rather than just same-hostname-as-first-url.
-    //
-    // `www.` is stripped on both sides so that a site registered as
-    // `okta.com` accepts URLs under `www.okta.com` (and vice versa) — the
-    // canonical apex/www pair has identical TLS + content guarantees. Other
-    // subdomains are rejected; promote them to their own site if needed.
-    const stripWww = (h) => h.toLowerCase().replace(/^www\./, '');
-    const normalizedSiteHost = stripWww(siteHostname);
     if (!Array.isArray(data.urls) || data.urls.length === 0) {
       return badRequest('Invalid request: urls must be a non-empty array');
     }
@@ -233,7 +226,7 @@ function ScrapeJobController(context) {
       } catch {
         return badRequest(`Invalid request: malformed URL: ${u}`);
       }
-      if (stripWww(urlHostname) !== normalizedSiteHost) {
+      if (urlHostname !== siteHostname) {
         log.warn(`Cross-origin URL rejected for site ${siteId}: ${u}`);
         return badRequest(
           `Invalid request: every URL must belong to the site origin (${siteHostname})`,
--- a/src/controllers/scrapeJob.js
+++ b/src/controllers/scrapeJob.js
@@ -17,12 +17,27 @@ import {
   accepted,
   notFound,
   badRequest,
+  forbidden,
+  internalServerError,
 } from '@adobe/spacecat-shared-http-utils';
 import {
   isValidUrl,
+  isValidUUID,
+  isNonEmptyObject,
   hasText,
+  DELIVERY_TYPES,
 } from '@adobe/spacecat-shared-utils';
+import { Site as SiteModel } from '@adobe/spacecat-shared-data-access';
+import { retrievePageAuthentication } from '@adobe/spacecat-shared-ims-client';
 import { cleanupHeaderValue } from '@adobe/helix-shared-utils';
+import AccessControlUtil from '../support/access-control-util.js';
+import { getCookieValue, getIMSPromiseToken, ErrorWithStatusCode } from '../support/utils.js';
+
+const PROMISE_BASED_TYPES = [
+  SiteModel.AUTHORING_TYPES.CS,
+  SiteModel.AUTHORING_TYPES.CS_CW,
+  SiteModel.AUTHORING_TYPES.AMS,
+];
 
 /**
  * Scrape controller. Provides methods to create, read, and fetch the result of scrape jobs.
@@ -46,6 +61,27 @@ function ScrapeJobController(context) {
 
   const MAX_JOBS_BY_BASEURL = 100;
 
+  // Duplicated from preflight controller (decision: refactor to support/scrape-auth.js
+  // as follow-up once the second consumer ships and the abstraction is justified).
+  async function checkEnableAuthentication(previewBaseURL) {
+    const headResponse = await fetch(previewBaseURL, {
+      method: 'HEAD',
+      headers: { 'Content-Type': 'application/json' },
+    });
+    return headResponse.status === 401 || headResponse.status === 403;
+  }
+
+  async function resolvePromiseToken(site, requestContext) {
+    if (!PROMISE_BASED_TYPES.includes(site.getAuthoringType())) {
+      return null;
+    }
+    const cookieToken = getCookieValue(requestContext, 'promiseToken');
+    if (hasText(cookieToken)) {
+      return { promise_token: cookieToken };
+    }
+    return getIMSPromiseToken(requestContext);
+  }
+
   function createErrorResponse(error) {
     // cleanupHeaderValue strips chars HTTP headers can't carry (CR/LF and non-ASCII
     // that would otherwise throw ERR_INVALID_CHAR). The `|| 'Internal server error'`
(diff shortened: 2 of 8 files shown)


aenascut:
## E2E validated in dev via direct Lambda invoke

PR code deployed to dev as Lambda version 6428 (alias `latest26207123942`). The `:ci` alias isn't promoted from branch deploys (that happens on main merge), so the public `/api/ci/` URL doesn't yet route here. To unblock testing, exercised the deployed alias via direct `aws lambda invoke` with a synthesized API Gateway v2 event.

### Smoke tests

| # | Case | Expected | Got | Pass |
|---|---|---|---|---|
| 1 | Missing `metaData.siteId` | `400` | `400 "Invalid request: metaData.siteId is required and must be a valid UUID"` | ✅ |
| 2 | Site not found | `404` | `404 "Site not found: 00000000-..."` | ✅ |
| 3 | Okta E2E (`b520b4cf-...` / `https://www.okta.com/`) | `202` + scrape complete | `202 jobId=5f5b8f05-...` → job `COMPLETE` in 22s, `successCount=1` | ✅ |
| 4 | Infographic page | scrape complete | `202 jobId=4e69b1e9-...` → `COMPLETE` in 27s, 148 KB rawBody | ✅ |

### SITES-40597 use case — replay of a real failing BrightData job

The DRS job [`709e64a1-f974-491e-a23b-db199a9f9041`](https://d2hp6j4u6kak3w.cloudfront.net/jobs/709e64a1-f974-491e-a23b-db199a9f9041) (in `drs-v2-main-jobs`) failed with:

```
status: COMPLETED_WITH_ERRORS
error_message: Retry merge into parent failed; partial data from initial delivery
               may be available. Reason: parent_s3_key_not_found
external_job_id: j_mpe5uc0e5f5ns2wbu   (BrightData snapshot)
```

Submitted the same 6 URLs through `/tools/scrape/authenticated-jobs` (job `c7cf1bbf-222f-45f7-8593-c48012c89b93`):

| URL | BrightData | SpaceCat content-scraper |
|---|---|---|
| `https://www.okta.com/` | ❌ merge failed | ✅ COMPLETE |
| `/video/verisk-analytics-video/` | ❌ | ✅ COMPLETE |
| `/resources/webinar-okta-workflows-...` | ❌ | ✅ COMPLETE (128 KB real content) |
| `/whats-new-at-okta-product-releases-roadmap-july/` | ❌ | ✅ COMPLETE |
| `/resources/whitepaper/a-tale-of-best-frenemies-...` | ❌ | ✅ COMPLETE (141 KB real content) |
| `/infographic/building-highly-scalable-apps/` | ❌ | ✅ COMPLETE (129 KB real content) |

**6/6 success, 0 failures, ~75s wall-clock.** Spot-checked three scrape results — HTTP 200, finalUrl matches request URL, real page HTML, no login-wall redirects.

### Auth-injection branch coverage

All six tested URLs returned HTTP 200/301 on HEAD probe (Okta marketing pages are public), so the endpoint correctly took the **transparent-delegation** path — no header injected, no Secrets Manager call, no `$LATEST` bug surface. The HEAD/IMS-exchange/`retrievePageAuthentication`/`Bearer` vs `token` branches are exercised only by unit tests in this run, but they're the same code preflight-beta uses today in this Lambda alias (`createBetaPreflightJob`, `controllers/preflight.js:405-429`) — so there's an in-production precedent for the alias-qualified `:latest` `ctx.func.version` working correctly through `retrievePageAuthentication`.

### How to reproduce

1. Get the DRS dev API key:
   ```bash
   KEY=$(aws ssm get-parameter --name /drs/dev/spacecat/api_key --with-decryption \
     --region us-east-1 --profile drs-dev --query Parameter.Value --output text)
   ```
2. Build an API Gateway v2 event with `pathParameters.path = "tools/scrape/authenticated-jobs"` (helix-universal reads this, not `rawPath`) and your body.
3. `aws lambda invoke --function-name spacecat-services--api-service:latest26207123942 --payload fileb://evt.json out.json --region us-east-1 --profile spacecat-dev --cli-binary-format raw-in-base64-out`.
4. Poll job status via the legacy `GET /tools/scrape/jobs/{jobId}` and `GET /sites/{siteId}/files?key=...` for the result content.

### Conclusion

End-to-end pipeline (spacecat-api → SQS → scrape-job-manager → content-scraper → S3) is proven for the SITES-40597 use case. Endpoint is functionally complete; ready for review.

solaris007:
@LeChef318 please also have a look and determine impact (if any) to container story.

src/controllers/scrapeJob.js
```diff
@@ -58,16 +94,129 @@ function ScrapeJobController(context) {
...
+    // `data.options.enableAuthentication === true` was true above.
+    const { enableAuthentication: _, ...remainingOptions } = data.options;
+
+    const delegatedData = {
+      ...data,
+      options: remainingOptions,
+      ...(authorizationHeader && {
+        customHeaders: {
+          ...(data.customHeaders || {}),
+          Authorization: authorizationHeader,
```
solaris007, line 324: Critical: cross-origin credential exfiltration. The resolved Authorization is injected into customHeaders here, but `data.urls` are caller-supplied and never checked against the site origin. The PR schema documents customHeaders as sent with each request, and content-scraper applies one customHeaders set to every URL in the job - so a caller can pass `siteId` of an auth-walled site with `urls` pointing at an attacker host, and the server mints that site's token (for non-promise sites the long-lived static Secrets Manager PAGE_AUTH_TOKEN) and sends it there. The preflight path you copied validates "all urls must belong to the same website"; this copy dropped that control while adding credential injection.

Bind every URL to the site origin before injecting the header - add the guard right after previewBaseURL is derived (line 164):

```js
const siteHost = new URL(previewBaseURL).hostname;
if ((data.urls || []).some((u) => { try { return new URL(u).hostname !== siteHost; } catch { return true; } })) {
  return badRequest('Invalid request: all urls must belong to the authenticated site when options.enableAuthentication is true');
}
```

Add a cross-origin-URL test asserting 400 - the current suite only uses same-host URLs and is blind to this gap.

src/controllers/scrapeJob.js
```diff
@@ -58,16 +94,129 @@ function ScrapeJobController(context) {
...
+        },
+      }),
+    };
+
+    try {
+      const job = await scrapeClient.createScrapeJob(delegatedData);
       return accepted(job);
```
solaris007: Critical: the server-minted credential is persisted and echoed back. The `job` returned here carries `customHeaders.Authorization` (ScrapeJobDto.toJSON includes customHeaders in spacecat-shared), so the token is returned in this 202 body, persisted in DynamoDB, and harvestable via getScrapeJobStatus / getScrapeJobsByBaseURL / getScrapeJobsByDateRange by any org-authorized caller - including tokens minted for other callers' jobs. For non-promise sites this is the long-lived static Secrets Manager secret. The preflight pattern you mirrored persists only `{siteId, url, step}`; this sink stores a live credential.

In-scope minimum: strip `customHeaders.Authorization` from the returned job before `return accepted(job)`. The at-rest persistence and LIST exposure need a redaction in spacecat-shared - file a blocking tracked ticket before this path is enabled in production. Note that 100% unit coverage misses this because the tests assert the delegated payload, not the returned DTO.

src/controllers/scrapeJob.js
```diff
@@ -46,6 +61,27 @@ function ScrapeJobController(context) {
 
   const MAX_JOBS_BY_BASEURL = 100;
 
+  // Duplicated from preflight controller (decision: refactor to support/scrape-auth.js
```
solaris007: Important (per review directive): this duplicates `checkEnableAuthentication`, `resolvePromiseToken`, and `PROMISE_BASED_TYPES` from preflight.js, and the comment defers the refactor with no tracked ticket. All five reviewers flagged this, and the copies have already diverged - this one dropped preflight's HEAD-status debug log, requires `siteId`, and adds `hasAccess`. This is security-relevant credential-handling logic (status-code heuristic, Bearer-vs-token, IMS-vs-cookie precedence); divergence across copies is a latent vuln.

File a tracked ticket (SITES-xxxxx) and reference it in this comment before merge. A shared home already exists today - `src/support/utils.js` holds getCookieValue and getIMSPromiseToken - so the extraction need not wait on a new file.

src/controllers/scrapeJob.js
```diff
@@ -58,16 +94,129 @@ function ScrapeJobController(context) {
...
+      log.error(`Site ${siteId} has invalid baseURL: ${baseURL}`);
+      return internalServerError('Site has invalid baseURL');
+    }
+
+    let needsAuth = false;
+    try {
+      needsAuth = await checkEnableAuthentication(previewBaseURL);
+    } catch (e) {
+      // HEAD probe is best-effort; on failure delegate without header injection.
+      log.info(`HEAD probe failed for ${previewBaseURL}: ${e.message}`);
```
solaris007: Important: silent demotion to anonymous scrape. On HEAD-probe throw (here), and also when an auth-walled site answers 200-with-login-form, 3xx-to-IdP (fetch follows redirects), or 5xx, `needsAuth` stays false and the job scrapes anonymously, storing a login page as real content - the exact SITES-40597 failure class, and a direct contradiction of the fail-loud rationale you give for the missing-siteId 400.

Raise this to `log.warn` with a metric, consider failing closed when auth was explicitly requested, and reconcile the 3xx/5xx behavior with the loud-fail intent. Only 401/403 are tested - add 403, 3xx, and 5xx HEAD cases.

src/controllers/scrapeJob.js
```diff
@@ -46,6 +61,27 @@ function ScrapeJobController(context) {
 
   const MAX_JOBS_BY_BASEURL = 100;
 
+  // Duplicated from preflight controller (decision: refactor to support/scrape-auth.js
+  // as follow-up once the second consumer ships and the abstraction is justified).
+  async function checkEnableAuthentication(previewBaseURL) {
+    const headResponse = await fetch(previewBaseURL, {
+      method: 'HEAD',
```
solaris007: Important: the HEAD probe has no timeout and follows redirects. Node's global fetch (undici) defaults to ~300s, well beyond API Gateway's 29s limit, so a slow or hanging customer origin returns 504 while the Lambda holds concurrency until its own timeout. This is also the first blocking outbound call to arbitrary customer infrastructure on the POST /tools/scrape/jobs request path.

Add `signal: AbortSignal.timeout(3000)` (or similar) and `redirect: 'manual'` so a site baseURL cannot 30x-launder the probe toward an internal host.

solaris007 (changes requested):
Hey @aenascut,

This is a well-constructed workaround - the flag-gated seam keeps the legacy path byte-identical, `hasAccess` runs before any token minting (stricter than the preflight path you mirrored, which has no access gate), the server-resolved header correctly overrides caller-supplied values, and no token material is logged. The precedent claim checks out (`preflight.js:419-455`). The gap is containment of the credential the server now mints: it is not bound to where it gets sent, and it flows into a persisted, caller-readable field. Two Critical leak vectors need to be closed before this can merge. Reviewed with a security focus per request.

## Strengths
- Legacy fast-path is genuinely byte-identical for non-auth callers (`scrapeJob.js:122`); `options.enableAuthentication` is stripped on the auth path (`:205`) so content-scraper cannot re-resolve under `$LATEST`.
- Authorization is checked before the HEAD probe and token minting (`:156`), and is stricter than the mirrored `createBetaPreflightJob` (which has no `hasAccess` gate). Clean authn/authz ordering.
- Server-resolved `Authorization` overrides any caller-supplied value (`:213`), so callers cannot spoof the header. No token/cookie/secret appears in any log line.
- `previewBaseURL` is derived from the trusted stored `getBaseURL()`, not request input - narrower SSRF surface than preflight.
- Tests assert the delegated SQS payload shape (header scheme, customHeaders merge, flag stripping) with mocks at real boundaries (fetch/IMS/Secrets Manager/DB), not interaction theater.

## Issues

### Critical (Must Fix)

**1. Cross-origin credential exfiltration: `data.urls` are never bound to the site origin (`scrapeJob.js:213`, fix at `:164`).** The token is resolved from the site selected by `metaData.siteId`, but the URLs actually fetched are caller-supplied and unchecked. The PR's own schema documents `customHeaders` as sent "with each HTTP request", and content-scraper applies one `customHeaders` set to every URL in the job. So a caller (any admin/legacy key holder, or a scoped-key holder for one auth-walled site) can submit `urls: ["https://attacker.example/"]` with `siteId` of an auth-walled site, and the server mints that site's `Authorization` - for non-promise sites the long-lived static Secrets Manager `PAGE_AUTH_TOKEN` - and sends it to the attacker host. The preflight path you copied validates "all urls must belong to the same website"; this copy dropped that control while adding credential injection. Fix: origin-bind every URL before injecting the header (mirror preflight), and add a cross-origin-URL test asserting 400.

**2. The server-minted credential is persisted at rest and echoed back through the API (`scrapeJob.js:220`).** The returned `job` carries `customHeaders.Authorization` (`ScrapeJobDto.toJSON` includes `customHeaders` in spacecat-shared), so the token is returned in this 202 body, persisted in DynamoDB, and harvestable via `getScrapeJobStatus` / `getScrapeJobsByBaseURL` / `getScrapeJobsByDateRange` by any org-authorized caller - including tokens minted for other callers' jobs. Unlike the preflight pattern (which persists only `{siteId, url, step}`), this copy reused the resolution but changed the sink to a persisted, returned field. In-scope minimum: strip `customHeaders.Authorization` from the returned job before `return accepted(job)`. The at-rest persistence and LIST exposure require a redaction in spacecat-shared - file a blocking tracked ticket before this path is enabled in production. 100% unit coverage misses this because tests assert the delegated payload, not the returned DTO.

### Important (Should Fix)

**1. Duplicated credential-handling code with no tracked ticket (`scrapeJob.js:64`).** `checkEnableAuthentication`, `resolvePromiseToken`, and `PROMISE_BASED_TYPES` are copied from `preflight.js`; the comment defers the refactor with no ticket. All reviewers flagged this, and the copies have already diverged at birth: this one dropped preflight's HEAD-status debug log, requires `siteId`, and adds `hasAccess`. This is security-relevant logic (status-code heuristic, Bearer-vs-token, IMS-vs-cookie); divergence is a latent vuln. File a tracked ticket and reference it in the comment before merge. A shared home already exists today (`src/support/utils.js` holds `getCookieValue`/`getIMSPromiseToken`), so extraction need not wait on a new file.

**2. Jobs with more than ~20 URLs scrape the remainder anonymously.** content-scraper relies entirely on the baked `customHeaders` (since the flag is stripped), but when `urls > concurrency` it re-queues the remainder as `{...config, urls, options}` where `config` has no `customHeaders`. The overflow batch arrives with no `Authorization` and no flag, and is scraped anonymously - the exact SITES-40597 failure class. The E2E used 6 URLs so this is unobserved. The downstream fix is cross-repo (track it), but it should gate enabling authenticated jobs above the concurrency limit.

**3. Silent demotion to anonymous scrape contradicts the feature's fail-loud principle (`scrapeJob.js:175`).** On HEAD-probe throw, and also when an auth-walled site answers 200-with-login-form, 3xx-to-IdP (fetch follows redirects), or 5xx, `needsAuth` stays false and the job scrapes anonymously, storing a login page as content. This directly contradicts the rationale given for the missing-`siteId` 400. Raise to `log.warn` with a metric, consider failing closed when auth was explicitly requested, and reconcile the 3xx/5xx behavior with the loud-fail intent.

**4. The HEAD probe has no timeout and follows redirects (`scrapeJob.js:68`).** Node's global fetch defaults to ~300s, well beyond API Gateway's 29s limit, so a slow customer origin returns 504 while the Lambda holds concurrency. This is also the first time `POST /tools/scrape/jobs` makes a blocking outbound call to arbitrary customer infrastructure on the request path. Add `AbortSignal.timeout(...)` and `redirect: 'manual'` (the latter stops a baseURL from 30x-laundering the probe toward an internal host).

**5. The auth-injection branch has never run end-to-end, and the posted E2E evidence is stale.** The validation comment exercises a route that no longer exists (`/tools/scrape/authenticated-jobs`) and reports a different 400 message than the current code, and by the author's own caveat every tested URL returned 200/301 on HEAD - so the `retrievePageAuthentication` / Secrets Manager call this PR exists to prove is stubbed in all 18 unit tests and was never actually executed. Additionally, 11 of 18 tests assert only `response.status`: the security-relevant negatives (403/404/500) never assert that token resolution and delegation did NOT happen, so a future reorder that mints a token before the access check would stay green. Correct/remove the stale comment, add negative-interaction assertions (`retrievePageAuthentication`/`sendMessage` not called) on the guard paths plus 403/3xx/5xx HEAD cases, and run one real auth-walled E2E before flipping production traffic.

### Minor (Nice to Have)

**1. `enableAuthentication` truthy-non-`true` values bypass the gate unstripped (`scrapeJob.js:122`).** `"true"` (string) or `1` fails the `!== true` check, takes the legacy fast-path, and the raw flag is forwarded downstream unchanged - contradicting the "always stripped" property and potentially re-triggering the `$LATEST` resolution. Coerce to boolean or strip on all paths.

**2. `previewBaseURL` uses `parsed.hostname` not `parsed.host` (`scrapeJob.js:164`)** - drops the port, so an AEM author on `:4502` is probed on 443 and gets the wrong auth decision.

**3. WAF/bot 403 false-positive (`scrapeJob.js:68`).** The probe sends a datacenter-IP HEAD with no User-Agent; Akamai/Cloudflare often answer 403, flipping `needsAuth` true and minting+forwarding a credential toward a site that may not be an Adobe-auth property. Consider distinguishing `WWW-Authenticate`/401 from generic 403, or gating on authoring/delivery type.

**4. `AccessControlUtil.fromContext(requestContext)` is uncaught (`scrapeJob.js:155`).** Its constructor throws on a context without `authInfo`; every other external call here is individually try/caught with a chosen status. Wrap it for consistency.

**5. Bearer-vs-token for the promise-authoring + non-`AEM_CS`-delivery combo (`scrapeJob.js:193`)** emits `token <accessToken>` for what is an exchanged IMS access token (RFC 6750 expects `Bearer`). Matches preflight - confirm the downstream contract actually expects `token` here.

## Recommendations
- Ensure delegated `scrapeClient` errors never echo `customHeaders` into the `x-error` header (size-capped but not redacted).
- The cookie `promiseToken` is trusted over the caller's IMS identity and is not bound to the authenticated caller; now that the token is forwarded externally, note the CSRF/confused-deputy surface (relevant to DRS forwarding `site_id` + urls).
- Document the specific 400 case (missing/invalid `metaData.siteId` when `enableAuthentication=true`) in `scrape-api.yaml` alongside the new 403/404.
- Record the end-state for the `$LATEST` workaround: either adobe/spacecat-shared PR 1614 lands and this edge path is explicitly temporary with a removal condition, or the API edge becomes canonical and content-scraper's resolution is planned for removal. Avoid two permanent auth-resolution pathways.

- adobe/spacecat-shared PR 1614 (`$LATEST` to `latest`) - the real fix; track removal of this workaround once it ships.
- spacecat-shared: `ScrapeJobDto.toJSON` returns and persists `customHeaders` - redact `Authorization` (pairs with Critical 2).
- content-scraper: propagate `customHeaders` to the re-queued remainder batch (pairs with Important 2); host-bind `customHeaders` application (defense-in-depth behind Critical 1).
- Confirm the SQS credential channel is KMS-encrypted at rest and the header is not logged downstream.
- Scope/rotation of the static `PAGE_AUTH_TOKEN` - if shared across tenants, the Critical 1/2 blast radius is larger.
- The "container story" @LeChef318 was asked about may change the `$LATEST` calculus entirely.

## Assessment
Not ready to merge. The seam design is sound and the legacy path is safe, but the change mints a real credential (including a static Secrets Manager secret) and does not contain it: Critical 1 sends it to caller-controlled hosts, and Critical 2 persists and returns it through the API. Both have in-scope fixes (origin-bind the URLs; strip the header from the response) plus tracked cross-repo follow-ups. The duplicated auth code needs a tracked ticket per the review directive, not a comment.

## Next Steps
1. Fix the two Critical leak vectors: origin-bind `data.urls` to the site, and strip `Authorization` from the returned job (plus a blocking spacecat-shared redaction ticket).
2. File the dedup ticket and reference it in the code; file the cross-repo tracking tickets (more-than-20-URL remainder, DTO redaction).
3. Address Important 3-5 (HEAD timeout/redirect, silent-degrade logging, test/E2E gaps) or risk-accept with rationale; correct the stale E2E comment.
4. Minor items are optional.

src/controllers/scrapeJob.js
aenascut: Fixed in 4d6a0dea — URL-origin guard added at `scrapeJob.js:177-195` before any auth resolution runs:

```js
for (const u of data.urls) {
  let urlHostname;
  try {
    urlHostname = new URL(u).hostname;
  } catch {
    return badRequest(`Invalid request: malformed URL: ${u}`);
  }
  if (urlHostname !== siteHostname) {
    log.warn(`Cross-origin URL rejected for site ${siteId}: ${u}`);
    return badRequest(
      `Invalid request: every URL must belong to the site origin (${siteHostname})`,
    );
  }
}
```

Tightened beyond the preflight `same-hostname-as-first-url` check — every URL must match the *site's* hostname (`site.getBaseURL()`), not just match each other. New tests: `400s on cross-origin URL when caller submits siteId for a different site`, `400s on malformed URL in auth flow`, `400s when urls array is empty in auth flow`. The cross-origin test also asserts `fetchStub.notCalled`, so an attacker URL never triggers the HEAD probe (defense in depth — the probe used to go to the site's preview URL, but we don't want any attacker-controlled hostname reaching `fetch` in this flow at all).

src/controllers/scrapeJob.js
aenascut: Fixed in 4d6a0dea — added `sanitizeJobResponse` at `scrapeJob.js:124-133` and applied it to the 202 body:

```js
function sanitizeJobResponse(job) {
  if (!isNonEmptyObject(job) || !isNonEmptyObject(job.customHeaders)) {
    return job;
  }
  const { Authorization: _, ...safeHeaders } = job.customHeaders;
  return { ...job, customHeaders: safeHeaders };
}
// ...
const job = await scrapeClient.createScrapeJob(delegatedData);
return accepted(sanitizeJobResponse(job));
```

Content-scraper still receives the Authorization header via the SQS message (delegated payload unchanged) — only the response body is sanitized. Test `strips customHeaders.Authorization from the 202 response body` asserts both directions: the SQS payload has the Authorization, the response body does not.

The at-rest DDB + LIST/GET exposure is the bigger fix and needs a `ScrapeJobDto` change in `spacecat-shared`. Filed as **SITES-45205** ("Redact customHeaders.Authorization at rest + on LIST in ScrapeJobDto") — captures the DTO redaction, the at-rest persistence decision (supervisor-strip vs persist-as-is + redact-on-read), and the migration for pre-existing rows. Out of scope for this PR per your "in-scope minimum"; comment in the controller references the ticket so the gap doesn't get lost.

src/controllers/scrapeJob.js
aenascut: Filed **SITES-45204** ("Extract shared scrape-auth helpers from preflight + scrapeJob controllers") and referenced it in the deferral comment (`scrapeJob.js:79-81`):

```js
// Duplicated from preflight controller. Extraction tracked in SITES-45204
// (move checkEnableAuthentication / resolvePromiseToken / PROMISE_BASED_TYPES
// to src/support/auth.js, alongside getIMSPromiseToken).
```

Agreed on the drift risk — the ticket explicitly calls out that the copies have *already* diverged this PR (URL-origin binding, fail-closed-on-ambiguous, 3 s timeout, `redirect: 'manual'`, requires `metaData.siteId`) versus preflight's best-effort copy. Scope is:

1. Create `src/support/auth.js`
2. Move the three symbols (using the hardened version)
3. Update both controllers to import
4. Decide whether to migrate preflight to the tristate return (`'ambiguous'`) or keep best-effort — documented in the ticket
5. Unit tests for the extracted module; slim controller tests to controller-specific paths

Not extracting in this PR per your in-scope-minimum guidance, but the ticket is filed against the same epic (SITES-39521) with `security` + `tech-debt` labels so it lands on the next cycle.

src/controllers/scrapeJob.js
aenascut: Fixed in 4d6a0dea — `checkEnableAuthentication` now returns a tristate (`'auth-required' | 'no-auth' | 'ambiguous'`) and never throws, and the controller fails closed on `'ambiguous'`:

```js
async function checkEnableAuthentication(previewBaseURL) {
  let headResponse;
  try {
    headResponse = await fetch(previewBaseURL, {
      method: 'HEAD',
      headers: { 'Content-Type': 'application/json' },
      redirect: 'manual',
      signal: AbortSignal.timeout(HEAD_PROBE_TIMEOUT_MS),
    });
  } catch (e) {
    log.warn(`HEAD probe failed for ${previewBaseURL}: ${e.name}: ${e.message}`);
    return 'ambiguous';
  }
  const { status } = headResponse;
  if (status === 401 || status === 403) return 'auth-required';
  if (status >= 200 && status < 300) return 'no-auth';
  log.warn(`HEAD probe for ${previewBaseURL} returned ambiguous status ${status}`);
  return 'ambiguous';
}
// ...
const probeResult = await checkEnableAuthentication(previewBaseURL);
if (probeResult === 'ambiguous') {
  return createResponse({}, 502, {
    [HEADER_ERROR]: cleanupHeaderValue(
      `Could not verify authentication requirement for ${previewBaseURL}`,
    ).slice(0, 500),
  });
}
```

`'ambiguous'` covers everything you flagged: HEAD throw (network/DNS), 3xx-to-IdP (no longer followed thanks to `redirect: 'manual'`), 5xx, timeout. All four return 502 with no SQS send, no anonymous scrape.

New tests:
- `fails closed (502) when HEAD probe throws — auth was explicitly requested` (asserts `sendMessage` not called)
- `fails closed (502) when HEAD returns a 3xx (cannot follow under redirect: manual)`
- `fails closed (502) when HEAD returns 5xx`
- `fails closed (502) when HEAD probe aborts on timeout`

This matches the same fail-loud intent as the missing-siteId 400 — caller asked for authenticated scraping, we can't verify, we don't pretend we did.

src/controllers/scrapeJob.js
aenascut: Fixed in 4d6a0dea — both your concerns addressed at `scrapeJob.js:73-79`:

```js
const HEAD_PROBE_TIMEOUT_MS = 3000;
// ...
headResponse = await fetch(previewBaseURL, {
  method: 'HEAD',
  headers: { 'Content-Type': 'application/json' },
  redirect: 'manual',
  signal: AbortSignal.timeout(HEAD_PROBE_TIMEOUT_MS),
});
```

- **Timeout**: 3 s `AbortSignal.timeout` keeps the outbound well under the 29 s API Gateway budget even with retries; a slow/hanging customer origin now aborts cleanly instead of holding a Lambda slot until the global undici default (~300 s).
- **Redirect: 'manual'**: a customer baseURL cannot 30x-launder the probe toward an internal host. 3xx responses now surface as `'ambiguous'` → 502 (covered in the L175 fail-closed change), so an attacker-controlled `Location` header on the site origin also can't be used to fish for internal endpoints.

Test `passes signal: AbortSignal.timeout + redirect: manual to fetch` asserts both options reach the fetch call. Test `fails closed (502) when HEAD probe aborts on timeout` exercises the `TimeoutError` path end-to-end.

LeChef318:
Hey @solaris007 — checked, no impact on the container story. The api-service PR is fully compatible and there's no merge-order constraint.

**Key checks:**

- The PR's defensive-properties section explicitly states no scraper-side changes are required, and that holds: `customHeaders.Authorization` flows unchanged through `abstract-handler.js` into `page.setExtraHTTPHeaders`. The strip of `options.enableAuthentication` before delegation also means our self-resolution path is bypassed entirely for these callers — no double-resolution risk.

- The underlying `$LATEST` Secrets Manager bug this PR works around doesn't apply to our containerized setup either. `spacecat-infrastructure/feat/content-scraper-cutover-dev` (`modules/lambda_triggers/event_source_mappings.tf`) wires all three content-scraper SQS ESMs against `arn:...:function:spacecat-services--content-scraper:latest`. The `:latest` suffix means helix-universal's ARN parser sets `ctx.func.version === 'latest'` instead of falling through to its `$LATEST` destructuring default, so `retrievePageAuthentication` resolves correctly in our Lambda too.

- For non-api-service callers that still set `enableAuthentication=true` directly (scrape-job-manager, autofix-worker, direct SQS sends), our self-resolution path is intact — the recent merge from \`main\` brought in the \`authId = siteId || metaData?.siteId || jobId\` fallback chain in \`run-sqs.js\`, and under our \`:latest\` ESMs it works.

Belt-and-suspenders end state once both ship: the api-service bakes the header for its callers; the container migration fixes the root cause for everything else routed through SQS.

src/controllers/scrapeJob.js
```diff
@@ -56,19 +120,177 @@ function ScrapeJobController(context) {
...
+      return badRequest('Invalid request: urls must be a non-empty array');
+    }
+    for (const u of data.urls) {
+      let urlHostname;
+      try {
+        urlHostname = new URL(u).hostname;
+      } catch {
+        return badRequest(`Invalid request: malformed URL: ${u}`);
+      }
+      if (urlHostname !== siteHostname) {
```
LeChef318: **Important: URL-origin guard accepts cross-port and HTTP-to-HTTPS - credential misdelegation surface.**

The guard compares `.hostname` only:
- `new URL('https://example.com:8443/admin').hostname === 'example.com'` passes against a site whose baseURL is `https://example.com/`. If the customer hosts an adjacent service on port 8443 that trusts the same IMS realm or PAT issuer (common in dev/staging stacks where multiple services share an Adobe IMS instance), the credential is misdelegated.
- `new URL('http://example.com/page').hostname === new URL('https://example.com/').hostname`. The protocol downgrade is not pinned. The HEAD probe runs against `https://` (line 207 derives from DB), but content-scraper sends the minted Authorization via `page.setExtraHTTPHeaders` to the `http://` URL - the credential traverses the network in clear.

Fix: compare full `URL.origin` (scheme + host + port). One-line change.

```js
const siteOrigin = new URL(baseURL).origin;
// inside the loop:
const urlOrigin = new URL(u).origin;
if (urlOrigin !== siteOrigin) { ... }
```

Add regression tests for both: `urls: ['https://www.okta.com:8443/page']` and `urls: ['http://www.okta.com/page']` against site baseURL `https://www.okta.com/` - both must 400.

src/controllers/scrapeJob.js
```diff
@@ -46,6 +61,55 @@ function ScrapeJobController(context) {
...
+    }
+    if (status >= 200 && status < 300) {
+      return 'no-auth';
+    }
+    log.warn(`HEAD probe for ${previewBaseURL} returned ambiguous status ${status}`);
+    return 'ambiguous';
+  }
+
+  async function resolvePromiseToken(site, requestContext) {
+    if (!PROMISE_BASED_TYPES.includes(site.getAuthoringType())) {
```
LeChef318: **Important: `resolvePromiseToken` predicate is asymmetric versus the ims-client's.**

The controller returns `null` for `!PROMISE_BASED_TYPES.includes(site.getAuthoringType())`. But `retrievePageAuthentication` in `@adobe/spacecat-shared-ims-client/src/auth.js:58-62` triggers the promise-exchange path on `(promiseBasedTypes.includes(authoringType) || deliveryType === AEM_CS) && authOptions.promiseToken`.

Today the asymmetry doesn't surface a bug because the ims-client falls through to the Secrets Manager `PAGE_AUTH_TOKEN` branch when no promise token is provided. But the controller and the ims-client encode "is this a promise-path site?" in two different places with different conditions. A future refactor of one without the other, or a customer that adds `AEM_CS` delivery on a non-`CS/CS_CW/AMS` authoring type, creates a silent scheme mismatch.

Fix: mirror the ims-client predicate exactly, or extract the predicate into a shared util as part of the SITES-45204 work.

```js
const isPromisePath = PROMISE_BASED_TYPES.includes(site.getAuthoringType())
  || site.getDeliveryType() === DELIVERY_TYPES.AEM_CS;
if (!isPromisePath) return null;
```

src/controllers/scrapeJob.js
```diff
@@ -56,19 +120,177 @@ function ScrapeJobController(context) {
...
+        if (e instanceof ErrorWithStatusCode) {
+          return badRequest(e.message);
+        }
+        return internalServerError('Error getting promise token');
+      }
+      try {
+        const authOptions = promiseTokenObj ? { promiseToken: promiseTokenObj } : {};
+        const accessToken = await retrievePageAuthentication(site, requestContext, authOptions);
+        const isBearer = site.getDeliveryType() === DELIVERY_TYPES.AEM_CS && !!promiseTokenObj;
+        authorizationHeader = `${isBearer ? 'Bearer' : 'token'} ${accessToken}`;
```
LeChef318, line 305: **Important: token scheme correctness for `(promise authoring) x (non-AEM_CS delivery)` is unverified empirically.**

For `authoringType` in `{CS, CS_CW, AMS}` paired with `deliveryType !== AEM_CS` (a real configuration), this path mints `"token <IMS-exchanged-access-token>"`. RFC 6750 conventions say IMS access tokens belong in a `Bearer` scheme; AEM EDS endpoints typically expect `"token <PAT>"` for static PAT-style auth, not an IMS JWT under `"token"`. Audience: an IMS access token's `aud` targets an Adobe-internal service registered with the exchange client; a non-AEM-CS customer endpoint is unlikely to validate that audience.

The "matches preflight" defense is real but doesn't establish correctness - preflight may not exercise this combination against an actual auth-walled site in production (most CS_CW preview endpoints sit behind AEM CS Dispatcher with `deliveryType=AEM_CS`). The E2E in the PR description used Okta marketing pages which returned 200/301 on HEAD, so the token-injection branch was not exercised against a real 401/403.

Before flipping production traffic for non-AEM_CS auth-walled sites: confirm with one query against the `Site` table which `(authoringType, deliveryType)` combinations exist where authentication is observed working today. If no `(promise-authoring) x (non-AEM_CS)` combo exists in practice, add a guard above line 266 that rejects the combo with a specific 502 message. If it does exist and should work, the scheme decision needs to come from a per-site auth-policy attribute, not a hardcoded `deliveryType === AEM_CS` rule.

test/controllers/scrape-job.test.js
```diff
@@ -162,13 +164,16 @@ describe('ScrapeJobController tests', () => {
       maxUrlsPerJob: 3,
     };
 
-    const { info, debug, error } = console;
+    const {
+      info, debug, error, warn,
+    } = console;
 
     // Set up the base context
     baseContext = {
```
LeChef318, line 174: **Important: `log.warn` is not stubbed - fail-closed observability is unverified.**

The new code emits `log.warn` on every fail-closed branch (`scrapeJob.js:88` HEAD probe failed, `:98` ambiguous status, `:230` cross-origin rejection) - exactly the signals SRE will query Coralogix for during triage. The test setup binds `warn: console.warn` (the raw pass-through), not a stub.

A refactor that swaps `log.warn` for a silent `return` would pass every existing test, and the production diagnosis path goes dark. The fail-closed branches exist specifically to produce a recoverable signal - that signal needs to be asserted.

Fix: `warn: sandbox.stub().callsFake(warn)` mirroring the existing pattern for `error`, then add one assertion per fail-closed test that `log.warn` was called with a message containing siteId / previewBaseURL / URL.

docs/openapi/scrape-api.yaml
```diff
@@ -27,6 +39,10 @@ create-scrape-job:
         $ref: './responses.yaml#/400-max-scrape-urls-exceeded'
       '401':
         $ref: './responses.yaml#/401'
+      '403':
+        description: Caller does not have access to the specified site (only when `options.enableAuthentication` is true).
+      '404':
+        description: Site not found for the provided `metaData.siteId` (only when `options.enableAuthentication` is true).
       '500':
         $ref: './responses.yaml#/500'
```
LeChef318, line 47: **Important: 502 response is not documented in the OpenAPI spec.**

The `responses:` block for `create-scrape-job` lists 202, 400, 401, 403, 404, 500. The new fail-closed 502 path (`scrapeJob.js:243-249`) is absent. Downstream consumers integrating against the schema (DRS adobe-rnd/llmo-data-retrieval-service#1797, Mystique) have no way to know to handle it.

Add the 502 with a one-liner matching the controller comment:

```yaml
'502':
  description: Authentication requirement could not be verified for the site's preview origin (only when `options.enableAuthentication` is true).
```

Semantic side note: 502 fits the 3xx/5xx HEAD case; the network-timeout case is arguably closer to 504. Either is defensible - picking one and documenting it is the ask.

LeChef318 (changes requested):
Hey @aenascut,

Thanks for the thorough fix commit. Verified the five inline findings against the current source: cross-origin URL guard, response sanitization, tristate fail-closed probe, HEAD hardening, and the SITES-45204 deferral comment all landed correctly with corresponding negative-assertion tests. The change discipline of the commit (no scope creep, comments explain *why*, the strip-before-delegate / fail-closed-on-ambiguous semantics are the right primitives) is exactly the shape this kind of follow-up should take.

A second pass surfaced a tighter URL-origin check, a predicate alignment with the ims-client, and three smaller items (OpenAPI 502 documentation, log.warn observability, empirical verification of the token scheme for non-AEM_CS promise sites). None of these block DRS adobe-rnd/llmo-data-retrieval-service#1797 or Mystique long-term, but they should land before this path is enabled for production traffic beyond the current Okta marketing case.

## Strengths

- **Cross-origin URL guard** (`scrapeJob.js:213-235`) - every URL's hostname is bound to `site.getBaseURL()` hostname. Tighter than preflight's same-hostname-as-first-url precedent. `fetchStub.notCalled` assertion in `test/controllers/scrape-job.test.js:1060-1066` proves no outbound probe is made when the guard rejects.
- **Response sanitization** (`scrapeJob.js:123-134, 293`) - `sanitizeJobResponse` strips Authorization from the 202 body. SQS payload unchanged. Test at line 1082 asserts both directions: SQS has it, response doesn't.
- **Tristate fail-closed probe** (`scrapeJob.js:78-100, 242-249`) - `checkEnableAuthentication` returns `'auth-required' | 'no-auth' | 'ambiguous'`; controller fails closed with 502 on ambiguous. Four new tests cover throw/3xx/5xx/timeout paths.
- **HEAD probe hardening** (`scrapeJob.js:81-86`) - `AbortSignal.timeout(3000)` + `redirect: 'manual'` closes the 30x-laundering vector and bounds Lambda time. Test at line 1051 asserts both reach `fetch`.
- **SITES-45204 ticket filed** (`scrapeJob.js:64-66`) - extraction of shared helpers to `src/support/auth.js` is tracked; the deferral comment now references the ticket.
- **Server-minted Authorization overrides caller-supplied** (`scrapeJob.js:283-288`) - spread order is correct and pinned by a test at line 1097.
- **`enableAuthentication` always stripped before delegation** (`scrapeJob.js:278`) - comment explains the `$LATEST` re-resolution bug it prevents; test at line 1114 pins the strip.
- **Comment quality** - the new comment blocks document *why* (the upstream `$LATEST` bug, the cross-origin attack class, the fail-closed semantics, the SQS retention) rather than restating the code. The kind of comments that survive 3 years of refactoring.
- **Container-migration compatibility** - confirmed by @LeChef318's analysis: stripping `enableAuthentication` before delegation means content-scraper's self-resolution path is bypassed cleanly when api-service has resolved, and the container ESMs target the `:latest` alias so the `$LATEST` bug doesn't affect the container path either.
- **Test-quality preflight all healthy** - assertion density 1.54, mock density 0.00, test-to-code ratio 1.41. No flood, no thin assertions, no mock-heavy.

## Issues

### Important (Should Fix)

[Also posted as inline comments on the specific lines.]

**1. URL-origin guard accepts cross-port and HTTP-to-HTTPS - credential misdelegation surface** (`src/controllers/scrapeJob.js:229`)

The guard compares `.hostname` only:
- `new URL('https://example.com:8443/admin').hostname === 'example.com'` passes against a site whose baseURL is `https://example.com/`. If the customer hosts an adjacent service on port 8443 that trusts the same IMS realm or PAT issuer (common in dev/staging stacks where multiple services share an Adobe IMS instance), the credential is misdelegated.
- `new URL('http://example.com/page').hostname === new URL('https://example.com/').hostname`. The protocol downgrade is not pinned. The HEAD probe runs against `https://` (line 207 derives from DB), but content-scraper sends the minted Authorization via `page.setExtraHTTPHeaders` to the `http://` URL - the credential traverses the network in clear.

Fix: compare full `URL.origin` (scheme + host + port). One-line change.

```js
const siteOrigin = new URL(baseURL).origin;
// inside the loop:
const urlOrigin = new URL(u).origin;
if (urlOrigin !== siteOrigin) { ... }
```

Add regression tests for both: `urls: ['https://www.okta.com:8443/page']` and `urls: ['http://www.okta.com/page']` against site baseURL `https://www.okta.com/` - both must 400.

**2. `resolvePromiseToken` predicate is asymmetric versus the ims-client's** (`src/controllers/scrapeJob.js:103`)

The controller returns `null` for `!PROMISE_BASED_TYPES.includes(site.getAuthoringType())`. But `retrievePageAuthentication` in `@adobe/spacecat-shared-ims-client/src/auth.js:58-62` triggers the promise-exchange path on `(promiseBasedTypes.includes(authoringType) || deliveryType === AEM_CS) && authOptions.promiseToken`.

Today the asymmetry doesn't surface a bug because the ims-client falls through to the Secrets Manager `PAGE_AUTH_TOKEN` branch when no promise token is provided. But the controller and the ims-client encode "is this a promise-path site?" in two different places with different conditions. A future refactor of one without the other, or a customer that adds `AEM_CS` delivery on a non-`CS/CS_CW/AMS` authoring type, creates a silent scheme mismatch.

Fix: mirror the ims-client predicate exactly, or extract the predicate into a shared util as part of the SITES-45204 work.

```js
const isPromisePath = PROMISE_BASED_TYPES.includes(site.getAuthoringType())
  || site.getDeliveryType() === DELIVERY_TYPES.AEM_CS;
if (!isPromisePath) return null;
```

**3. Token scheme correctness for `(promise authoring) x (non-AEM_CS delivery)` is unverified empirically** (`src/controllers/scrapeJob.js:266-267`)

For `authoringType` in `{CS, CS_CW, AMS}` paired with `deliveryType !== AEM_CS` (a real configuration), the path mints `"token <IMS-exchanged-access-token>"`. RFC 6750 conventions say IMS access tokens belong in a `Bearer` scheme; AEM EDS endpoints typically expect `"token <PAT>"` for static PAT-style auth, not an IMS JWT under `"token"`. Audience: an IMS access token's `aud` targets an Adobe-internal service registered with the exchange client; a non-AEM-CS customer endpoint is unlikely to validate that audience.

The "matches preflight" defense is real but doesn't establish correctness - preflight may not exercise this combination against an actual auth-walled site in production (most CS_CW preview endpoints sit behind AEM CS Dispatcher with `deliveryType=AEM_CS`). The E2E in the PR description used Okta marketing pages which returned 200/301 on HEAD, so the token-injection branch was not exercised against a real 401/403.

Before flipping production traffic for non-AEM_CS auth-walled sites: confirm with one query against the `Site` table which `(authoringType, deliveryType)` combinations exist where authentication is observed working today. If no `(promise-authoring) x (non-AEM_CS)` combo exists in practice, add a guard above line 266 that rejects the combo with a specific 502 message. If it does exist and should work, the scheme decision needs to come from a per-site auth-policy attribute, not a hardcoded `deliveryType === AEM_CS` rule.

**4. `log.warn` is not stubbed in tests - fail-closed observability is unverified** (`test/controllers/scrape-job.test.js:167-178`)

The new code emits `log.warn` on every fail-closed branch (`scrapeJob.js:88` HEAD probe failed, `:98` ambiguous status, `:230` cross-origin rejection) - exactly the signals SRE will query Coralogix for during triage. The test setup binds `warn: console.warn` (the raw pass-through), not a stub.

A refactor that swaps `log.warn` for a silent `return` would pass every existing test, and the production diagnosis path goes dark. The fail-closed branches exist specifically to produce a recoverable signal - that signal needs to be asserted.

Fix: `warn: sandbox.stub().callsFake(warn)` mirroring the existing pattern for `error`, then add one assertion per fail-closed test that `log.warn` was called with a message containing siteId / previewBaseURL / URL.

**5. 502 response is not documented in the OpenAPI spec** (`docs/openapi/scrape-api.yaml:31-47`)

The `responses:` block for `create-scrape-job` lists 202, 400, 401, 403, 404, 500. The new fail-closed 502 path is absent. Downstream consumers integrating against the schema (DRS adobe-rnd/llmo-data-retrieval-service#1797, Mystique) have no way to know to handle it.

Add the 502 with a one-liner matching the controller comment:

```yaml
'502':
  description: Authentication requirement could not be verified for the site's preview origin (only when `options.enableAuthentication` is true).
```

Semantic side note: 502 fits the 3xx/5xx HEAD case; the network-timeout case is arguably closer to 504. Either is defensible - picking one and documenting it is the ask.

### Minor (Nice to Have)

**1. `sanitizeJobResponse` is case-sensitive** (`src/controllers/scrapeJob.js:132`)

The destructure only matches the literal key `Authorization`. A caller can submit `customHeaders: { authorization: '<their-own-secret>' }`. The lowercase key survives the strip and is echoed back in the response. Caller leaking their own secret isn't critical, but the function name implies a defensive guarantee it doesn't provide. Filter case-insensitively against `{authorization, cookie, proxy-authorization}` and the contract matches the implementation.

**2. Truthy-non-true `enableAuthentication` routes around the new cross-origin guard** (`src/controllers/scrapeJob.js:163`)

The strict `data.options?.enableAuthentication !== true` check sends `{enableAuthentication: "true"}` (string) or `{enableAuthentication: 1}` down the legacy fast-path with the raw flag still in options. That branch does **not** apply the URL-origin guard and does **not** strip the flag. The prior reviewer flagged this as a Minor in the previous round; it is amplified now because the cross-origin guard is what prevents credential exfiltration on the auth path, and a typed bypass walks past it. Fix: explicit type rejection at the boundary.

```js
if ('enableAuthentication' in (data.options || {})
    && typeof data.options.enableAuthentication !== 'boolean') {
  return badRequest('Invalid request: options.enableAuthentication must be a boolean');
}
```

**3. URL userinfo not stripped from `data.urls[]`** (`src/controllers/scrapeJob.js:222-235`)

`new URL('https://user:pass@site.com/x').hostname` returns `'site.com'`, passing the cross-origin guard. The URL reaches content-scraper's Puppeteer with both HTTP Basic (from userinfo) and the injected `Authorization` - mixed-scheme semantics are server-dependent and surprising. Reject when `parsed.username || parsed.password` is non-empty with a specific 400.

**4. HEAD probe missing User-Agent** (`src/controllers/scrapeJob.js:81-86`)

The bare HEAD with only `Content-Type: application/json` (also meaningless on HEAD) is fingerprinted by Akamai/Cloudflare bot rules. WAF 403 leads to `'auth-required'`, which leads to a spurious Secrets Manager round-trip and a token forwarded to a site that may not actually require Adobe auth. With the URL-origin guard the credential stays on-origin, so the impact is wasted cost and noisy 401s, not credential leak. A stable UA (`User-Agent: spacecat-api-service/scrape-auth-probe`) lets customer WAFs allowlist this probe and helps trace requests in customer logs.

**5. Two fail-closed tests omit `sendMessage.notCalled`** (`test/controllers/scrape-job.test.js:1035-1040, 1042-1049`)

The HEAD-throws and HEAD-3xx tests both assert `mockSqsClient.sendMessage.to.not.have.been.called`. The 5xx and timeout tests only assert `response.status === 502`. Status-only could pass if a future bug returns 502 after sending SQS. Add the negative-interaction assertion for consistency with the other two fail-closed tests.

## Recommendations

- **Track SITES-45204 with a deadline.** Three real divergences already exist between this controller and preflight (URL-origin binding, tristate probe, 3s/redirect:manual hardening). Each new copy makes the eventual extraction harder. Pin the extraction to the sprint after this PR lands and migrate `createBetaPreflightJob` to the hardened tristate version so the shared file holds one canonical implementation, not two.
- **Document a retirement condition for the api-service edge auth-resolution path.** With adobe/spacecat-shared#1614 in flight and the container migration landing later, this path could become technically redundant. A 2-line "Retirement criteria:" block on the doc comment at `scrapeJob.js:145-153`, or a tracked ticket linked from there, prevents this workaround from becoming permanent infrastructure.
- **Pin `HEAD_PROBE_TIMEOUT_MS` in a test** - assert the configured 3000 ms reaches `AbortSignal.timeout`. The current test asserts only that the signal is an `AbortSignal` instance; a regression to `AbortSignal.timeout(30000)` would breach the 29s API Gateway budget under partial-network conditions and pass.
- **Consider adding `badGateway()` to spacecat-shared-http-utils** rather than inlining `createResponse({}, 502, ...)`. The current controller pattern at lines 243-249 duplicates the `cleanupHeaderValue + slice(0, 500)` contract that `createErrorResponse` already implements. Either reuse the helper or extract a small `errorResponse(status, message)` so the next caller doesn't hand-roll the shape.

- **SITES-45205** - at-rest DDB redaction in `ScrapeJobDto`. Already filed; LIST endpoints (`getScrapeJobStatus`, `getScrapeJobsByBaseURL`, `getScrapeJobsByDateRange`) currently re-emit `customHeaders.Authorization` from storage.
- **adobe/spacecat-shared#1614** - the long-term `$LATEST` to `latest` normalization in helix-universal. This PR is the workaround; that PR is the fix.
- **spacecat-content-scraper** - verify the Authorization header is not logged at info/debug, not persisted into job-state DDB, and not written into S3 scrape output. Also: the over-concurrency overflow batch propagates `customHeaders` (prior Important 2).
- **IMS access token TTL propagation to async worker** - the minted IMS access token (typical 24h TTL) is persisted into the SQS message body. If processing crosses `exp` (visibility-timeout retries on Puppeteer failures can extend processing windows), the worker sends an expired token. Edge-side mitigation: include `expiresAt` alongside `customHeaders` so the worker can fail-fast or surface for re-issuance.
- **`PAGE_AUTH_TOKEN` Secrets Manager hygiene** - rotation cadence, per-site vs shared-static scope, alerting on read frequency.

## Assessment

Reasoning: The good-faith engagement with the prior review is clear - all five inline findings have concrete fixes backed by negative-assertion tests, and the remaining unresolved Minors from round 1 are either tracked (duplication via SITES-45204), shifted in severity by the new fail-closed behavior, or were always cross-repo. The five Important findings here are mostly one-line or one-block changes (URL.origin comparison, ims-client predicate mirror, log.warn stub, OpenAPI 502 doc) plus one verification gate (token scheme for non-AEM_CS promise sites) that needs an empirical answer before the new path is enabled for production traffic beyond the current Okta marketing case. None of these block DRS adobe-rnd/llmo-data-retrieval-service#1797 long-term, but they should land before flipping production traffic.

## Next Steps

1. Address the five Important findings: URL.origin comparison, predicate mirror against ims-client, scheme verification for non-AEM_CS promise sites, log.warn assertions, and the OpenAPI 502 documentation.
2. The Minor findings are quality-of-life; most are one-line follow-ups. The truthy-non-true bypass (Minor 2) is worth pulling into this PR since it amplifies the URL.origin fix.
3. Recommendations and out-of-scope tracking are for the next sprint, not this PR.

src/controllers/scrapeJob.js
aenascut: Fixed in dd9c9a4c — switched the URL-origin guard from `.hostname` to `URL.origin` (scheme + host + port) at `scrapeJob.js:213-235`:

```js
const baseURL = site.getBaseURL();
let siteOrigin;
let previewBaseURL;
try {
  const parsed = new URL(baseURL);
  siteOrigin = parsed.origin;
  previewBaseURL = `${parsed.protocol}//${parsed.hostname}`;
} catch { ... }

// ...
for (const u of data.urls) {
  let urlOrigin;
  try {
    urlOrigin = new URL(u).origin;
  } catch {
    return badRequest(`Invalid request: malformed URL: ${u}`);
  }
  if (urlOrigin !== siteOrigin) {
    log.warn(`Cross-origin URL rejected for site ${siteId}: ${u} (site origin: ${siteOrigin})`);
    return badRequest(
      `Invalid request: every URL must belong to the site origin (${siteOrigin})`,
    );
  }
}
```

Two new tests close the bypasses you identified:
- `400s on cross-port URL even when hostname matches` — `https://www.okta.com:8443/admin` against site `https://www.okta.com/`
- `400s on HTTP-to-HTTPS downgrade (site is https, url is http)` — `http://www.okta.com/page` against site `https://www.okta.com/`

Both assert `fetchStub.notCalled` — attacker URLs never trigger the HEAD probe (defense in depth).

src/controllers/scrapeJob.js
aenascut: Fixed in dd9c9a4c — added `isPromisePathSite(site)` predicate at `scrapeJob.js:112-115` that mirrors the ims-client gate exactly:

```js
// Mirrors `@adobe/spacecat-shared-ims-client/src/auth.js` (the promise-
// exchange gate inside `retrievePageAuthentication`): the ims-client gates
// promise-token exchange on `(promiseTypes.includes(authoringType) ||
// deliveryType === AEM_CS) && authOptions.promiseToken`. Keeping this
// predicate identical avoids the controller and the ims-client drifting on
// "is this a promise-path site?" — that asymmetry is silent today (the
// ims-client falls through to the Secrets Manager PAT branch when no
// promise token is provided) but would surface for a customer who adds
// AEM_CS delivery on a non-promise authoring type, or vice versa.
// Extraction to a shared util is tracked in SITES-45204.
function isPromisePathSite(site) {
  return PROMISE_BASED_TYPES.includes(site.getAuthoringType())
    || site.getDeliveryType() === DELIVERY_TYPES.AEM_CS;
}

async function resolvePromiseToken(site, requestContext) {
  if (!isPromisePathSite(site)) {
    return null;
  }
  ...
}
```

The behavior delta is small but defensive: a site with `(authoringType=SO/HLX) × (deliveryType=AEM_CS)` previously returned `null` here and fell through to the ims-client's PAT branch; it now attempts cookie/IMS promise resolution first, matching what ims-client expects. The eventual `Authorization` value is unchanged for the combinations exercised in production today (verified by the existing 73 unit tests passing as-is).

Long-term, the predicate moves to `src/support/auth.js` with the rest of the duplicated logic — captured in [SITES-45204](https://jira.corp.adobe.com/browse/SITES-45204). Inline doc cross-references the ticket so future readers know where the canonical implementation will live.

src/controllers/scrapeJob.js
aenascut: Fixed in dd9c9a4c — guard added at `scrapeJob.js:269-285` that rejects `(promise authoring) × (non-AEM_CS delivery)` with 502 + `log.warn` *before* any token mint:

```js
if (PROMISE_BASED_TYPES.includes(site.getAuthoringType())
    && site.getDeliveryType() !== DELIVERY_TYPES.AEM_CS) {
  log.warn(
    `Refusing to mint Authorization for site ${siteId}: `
    + `authoringType=${site.getAuthoringType()} × deliveryType=${site.getDeliveryType()} `
    + 'is not a verified scheme combination',
  );
  return createResponse({}, 502, {
    [HEADER_ERROR]: cleanupHeaderValue(
      `Authentication scheme for site ${siteId} is not yet supported (`
      + `${site.getAuthoringType()} × ${site.getDeliveryType()})`,
    ).slice(0, 500),
  });
}
```

New test `502s on unverified (promise authoring × non-AEM_CS delivery) before any token mint` asserts:
- response status 502
- `retrievePageAuthentication` is never called (`expect(retrieveStub).to.not.have.been.called`)
- `sendMessage` is never called
- `log.warn` is called with `"not a verified scheme combination"`

Took the conservative path you outlined — the guard ships today, and a follow-up driven by a per-site `auth_policy` attribute can lift it once we have empirical scheme confirmation per `(authoring, delivery)` combination. Will pair that follow-up with the SITES-45204 extraction so the scheme decision lives in one place.

I'm also going to query the Site table before the production flip — agree we shouldn't enable the combo until we know which sites it actually affects.

test/controllers/scrape-job.test.js
aenascut: Fixed in dd9c9a4c — `baseContext.log.warn` is now a stub (was `console.warn` pass-through) at `test/controllers/scrape-job.test.js:172-178`:

```js
const {
  info, debug, error, warn,
} = console;

// Set up the base context. log.warn is stubbed (not pass-through) so that
// fail-closed branches can be asserted — those log lines are the SRE
// triage signal in Coralogix; a refactor that drops them must fail tests.
baseContext = {
  log: {
    info,
    debug,
    warn: sandbox.stub().callsFake(warn),
    error: sandbox.stub().callsFake(error),
  },
  ...
};
```

All six fail-closed tests now assert the warn was emitted with the relevant context:

- `fails closed (502) when HEAD probe throws` → `sinon.match(/HEAD probe failed for https:\/\/www\.okta\.com/)`
- `fails closed (502) when HEAD returns a 3xx` → `sinon.match(/ambiguous status 302/)`
- `fails closed (502) when HEAD returns 5xx` → `sinon.match(/ambiguous status 503/)`
- `fails closed (502) when HEAD probe aborts on timeout` → `sinon.match(/TimeoutError/)`
- `400s on cross-origin URL` → `sinon.match(/Cross-origin URL rejected/)`
- `502s on unverified (promise × non-AEM_CS)` → `sinon.match(/not a verified scheme combination/)`

A silent-return refactor of any of these branches will fail tests.

docs/openapi/scrape-api.yaml
aenascut: Fixed in dd9c9a4c — `docs/openapi/scrape-api.yaml#/create-scrape-job` now lists 502 alongside the existing 202/400/401/403/404/500:

```yaml
'502':
  description: |
    Authentication requirement could not be verified for the site's preview origin
    (only when `options.enableAuthentication` is true). Returned when the HEAD probe
    against the site's preview URL fails — network/DNS error, timeout, 3xx redirect
    (manual-redirect mode), or 5xx — or when the resolved `(authoringType, deliveryType)`
    combination is not a verified token-scheme pairing. The caller should treat this
    as transient (retry after backoff) or, if persistent, surface to the site owner
    to confirm the preview origin is reachable and the site is correctly configured
    for authenticated scraping.
```

On 502 vs 504 — went with 502 across the board for both the HEAD-status-ambiguous case and the timeout case. Rationale: 504 would be accurate if *we* were a gateway to the failing service, but here the upstream is opaque (a customer origin we don't proxy) and the contract we're returning is "we tried to act on your behalf and the upstream gave us something we can't safely interpret" — closer to a 502. Documented both branches under the same 502 to keep the caller's handling logic uniform (retry-then-surface-to-owner).

`npm run docs:lint` clean (194 pre-existing warnings unrelated). Downstream consumers (DRS #1797, Mystique #2255) will see the 502 in the generated spec.

LeChef318 (approved):
Hey @aenascut,

## Strengths

All ten prior findings (five from @solaris007, five from me) verified resolved in dd9c9a4 against the current source:

- **URL.origin guard** (scrapeJob.js:228-251) closes the cross-host, cross-port, and HTTP-to-HTTPS-downgrade vectors with one well-defined Node API. The decision to reject sibling subdomains rather than allowlist them is the right tradeoff for a credential-minting endpoint, and the three dedicated tests at L1076-1101 pin every variant.
- **HEAD probe hardening** (scrapeJob.js:77-100): 3s `AbortSignal.timeout` keeps the outbound under API Gateway's 29s budget, `redirect: 'manual'` blocks 30x-laundering toward an internal host, and the tristate return with fail-closed 502 on `'ambiguous'` removes the silent-demote class of SITES-40597. The `log.warn` on every ambiguous branch gives SRE a clean Coralogix triage signal.
- **`isPromisePathSite`** (scrapeJob.js:112-115) mirrors the ims-client gate at `@adobe/spacecat-shared-ims-client/src/auth.js:58-62` precisely. The comment at L102-110 documents the asymmetry and why it matters; SITES-45204 carries the eventual extraction.
- **Scheme-combination guard** (scrapeJob.js:276-289) rejects `(promise-authoring x non-AEM_CS-delivery)` with 502 + `log.warn` before any token mint. Test at L1103-1115 asserts `retrievePageAuthentication` is never called and SQS is never dispatched.
- **`sanitizeJobResponse`** (scrapeJob.js:143-149) strips the minted Authorization from the 202 body while preserving the SQS payload. The test asserts both halves of the contract, which is the right shape.
- **`log.warn` is now a stub** (test:178): `sandbox.stub().callsFake(warn)` plus six `calledWithMatch` assertions across the fail-closed branches. The Coralogix triage signal is now load-bearing under tests.
- **Workaround rationale is durable** (scrapeJob.js:160-168 + deferral comments citing SITES-45204 and SITES-45205): anyone opening this in six months will know what the workaround is for, where the duplication lives, and which tickets carry the deferred work.
- **Error-handling discipline** across the ~263 LOC new branch is tight: every `try` has a specific `log.error`/`log.warn`, no broad swallow, no catch-log-throw, no silent fallback.

## Issues

### Minor (Nice to Have)

1. **`customHeaders` merge does not normalize `Authorization` case before override** (scrapeJob.js:338-343). A caller submitting `customHeaders: { authorization: 'token attacker-supplied' }` (lowercase) produces a merged object with both `authorization` (caller-supplied) and `Authorization` (server-minted) present. Downstream behavior is HTTP-client-specific: undici/fetch Headers are case-insensitive and last-set wins (minted Authorization honored), but `node:http` raw headers serialize both, and RFC 7230 leaves Authorization-list-valued behavior unspecified. Exploitability is low (caller already has hasAccess), but the controller is shipping an ambiguous header set. Strip case-insensitive auth keys from caller input before merging, and add a test that submits lowercase `authorization` and asserts only the server-minted key reaches SQS. The existing test only covers the same-case collision.

2. **Inverse scheme-guard asymmetry**: `(non-promise-authoring x AEM_CS-delivery x no promise token resolved)` falls through to `token <PAT>` against an AEM_CS endpoint that expects `Bearer <IMS-JWT>`. Path: an S2S admin caller with no `promiseToken` cookie and `getIMSPromiseToken` returning null, against a site like `(SO x AEM_CS)`. `isPromisePathSite` returns true on the deliveryType branch, `resolvePromiseToken` returns null without throwing, `retrievePageAuthentication` falls through to the Secrets Manager PAT branch, and `isBearer = AEM_CS && !!null = false`, so the PAT goes out under the `token` scheme to an AEM_CS endpoint. Same SITES-40597 failure class the rest of the PR exists to prevent. The L269-285 guard catches the opposite combination but not this one. Note this matches preflight.js's current behavior, so the fix is a natural fit for SITES-45204's shared-auth extraction rather than this PR. Either way, capture it on the ticket.

3. **Legacy fast-path has no negative regression assertion**. The PR description promises "no Site lookup, no HEAD probe, no Secrets Manager call" when `enableAuthentication` is absent or false. Today the safety comes from `baseContext` not stubbing `Site.findById` at all, so any harmless-looking refactor that adds the stub to baseContext silently loses the guard. About six lines of test: `enableAuthentication: false` and `enableAuthentication` absent, each asserting `Site.findById.called === false` and `fetchStub.called === false`.

## Recommendations

- **Code-level polish in this PR or a follow-up**: move the empty-`urls` guard ahead of the `findById` call (skips a DB hit on malformed payloads); replace the two literal `createResponse({}, 502, ...)` blocks with a small helper that keeps the `cleanupHeaderValue(...).slice(0, 500)` dance in one place; extend the cross-port and HTTP-downgrade tests with the same `log.warn` assertion the cross-origin test already has; and add a test for HEAD 2xx pass-through *with* caller-supplied `customHeaders` to lock the spread-mechanics invariant.
- **Edge cases worth pinning**: `retrievePageAuthentication` returning `''` or `undefined` (currently produces `'token '` with no defensive check), `getIMSPromiseToken` returning null for an AEM_CS site, `promiseToken=` cookie present but empty string. Each is one assertion.
- **Status code split** when SITES-45204 lands: 502 for upstream probe failure (HEAD throw / 3xx / 5xx, true Bad Gateway semantics) and 422 for the unsupported scheme combination (a deterministic config-side rejection). Today both are 502 and only the `x-error` text differentiates them, which conflates two distinct alert rules in Coralogix.
- **SITES-45204 governance**: when the shared helper extraction lands, the controllers and the shared util should converge on the *hardened* semantics from this PR (tristate probe, 3s timeout, redirect:'manual', `isPromisePathSite` gate, scheme-uncertainty 502), not the older preflight.js shape that silently demotes. Worth adding as an explicit acceptance criterion on the ticket, plus a one-line marker on `preflight.js`'s two helpers pointing at scrapeJob.js as the convergence target.
- **`previewBaseURL` drops port** at scrapeJob.js:222 (`${parsed.protocol}//${parsed.hostname}`). Matches preflight.js so the consistency is intentional, but means a site baseURL like `https://preview.example.com:8443` probes the wrong host. Worth deciding in the SITES-45204 extraction whether to preserve the port.

- **`hasAccess(site)` is org-membership only** (per `access-control-util.js:226-339`); a view-only org member can trigger a scrape that mints and forwards a site Authorization header. Read-only admins are blocked from non-GET at the `readOnlyAdminWrapper` middleware, and the minted token never returns to the caller (`sanitizeJobResponse`), so this is strictly more restrictive than the legacy fast-path. Least-privilege follow-up worth capturing as a separate ticket if "view-only" is intended to exclude scrape.
- **SITES-45205** (at-rest exposure in `ScrapeJobDto` / LIST / GET): correctly outside this PR's scope. Confirmed `sanitizeJobResponse` is wired only into `createScrapeJob`; `getScrapeJobStatus`, `getScrapeJobsByDateRange`, `getScrapeJobsByBaseURL`, `getScrapeJobUrlResults` still return raw client results.
- **spacecat-shared#1614** ($LATEST -> latest normalization in helix-universal): the eventual real fix. Worth noting in SITES-40597's closeout that this PR is *not* dead weight after spacecat-shared#1614 lands. API-edge resolution has independent merits: synchronous failure surface for callers (502 at submit time vs a job that stalls in IN_PROGRESS), Secrets Manager hit once at the edge rather than fanned out across N scraper workers, reduced Secrets Manager IAM surface for the scraper Lambda, and HTTP-semantic failure modes that DRS and Mystique already handle. After spacecat-shared#1614 lands, the right shape is API-edge resolution as the primary path and scraper-side resolution as the fallback for SQS-sourced flows that bypass the API.
- **Drift surface**: four places encode "auth on scrape" - this PR, preflight.js (same Lambda, older), content-scraper's own auth path (other Lambda, still active for SQS-sourced flows), and spacecat-shared#1614 eventually. Three of those four ship together in api-service, governable via the SITES-45204 extraction. The cross-Lambda one needs an idempotency contract: content-scraper should defer to an incoming `customHeaders.Authorization` rather than re-resolve when one is already present. Worth a one-line check the next time someone touches the scraper-side handler.

## Assessment

Ready to merge? **Yes**.

Reasoning: All ten prior findings verified addressed correctly with code that matches the author's replies, plus substantive tests that exercise every branch and assert on the fail-closed log lines. The three Minor items are independently small and reasonable to defer to a follow-up; only the lowercase-`authorization` collision is worth a one-line fix here, and even that is low-exploitability given hasAccess already gates the mint. The architectural workaround is honest, well-documented, and the deferral tickets are in flight.

Note: CI's `it-postgres` job failed on workflow run 26228184746 but passed on 26228181774 with the same code; appears flaky/transient and unrelated to this diff. Confirm before merge.

## Next Steps

1. Optionally apply the lowercase-`authorization` strip in this PR (one-liner + one test).
2. Capture items 2 and 3 from the Minor list on SITES-45204 (inverse scheme guard) and as a small test follow-up (legacy fast-path negative assertions). Neither blocks merge.
3. Re-trigger the flaky `it-postgres` job and confirm green before merging.

solaris007 (approved):

PR #2456, opened 2026-05-21
feat(semrush): IMS-bearer-only AIO proxy + onboarding + DB-backed projects

rainer-friederich:
## Summary

Adds a server-side proxy in front of the Semrush AIO API at `adobe-hackathon.semrush.com`. Authentication is IMS-bearer-only: the client sends its IMS user token, this service forwards it as `Authorization: Bearer <ims>`, and Semrush handles IMS → Semrush exchange + workspace mapping server-side. Per-`(brand, market, language)` Semrush project slices are persisted in the new `brand_to_semrush_projects` table on `mysticat-data-service`.

This is the api-service half of a 3-repo change. The other two are already merged:

- `mysticat-data-service` migration `20260528000000_brand_to_semrush_projects.sql` — table, indices, grants, PostgREST schema reload
- `spacecat-shared` PR https://github.com/adobe/spacecat-shared/pull/1615 — `BrandSemrushProject` PostgREST-backed entity, published as `@adobe/spacecat-shared-data-access` (this PR pins `3.68.0`)

### 9 endpoints under `/v2/orgs/:spaceCatId/brands/:brandId/semrush/*`

| Method | Path | operationId |
|---|---|---|
| `GET`    | `/prompts` | `listSemrushPrompts` |
| `POST`   | `/prompts` | `createSemrushPrompts` |
| `PATCH`  | `/prompts/:promptId` | `updateSemrushPrompt` |
| `POST`   | `/prompts/bulk-delete` | `bulkDeleteSemrushPrompts` |
| `GET`    | `/projects` | `listSemrushProjects` |
| `POST`   | `/projects` (**new — onboarding**) | `createSemrushProject` |
| `GET`    | `/projects/:workspaceId/:projectId/tags` | `listSemrushProjectTags` |
| `GET`    | `/projects/:workspaceId/:projectId/models` | `listSemrushProjectModels` |
| `GET`    | `/workspaces/:workspaceId/projects` | `listSemrushWorkspaceProjects` |

Reads require capability `organization:read`, writes require `organization:write`.

### Authorization pipeline (every endpoint)

1. **IMS-only auth.** `requireImsBearer()` rejects non-IMS `authInfo` and missing bearers with `401 authenticationRequired`. No JWT / scoped API key gets forwarded to Semrush.
2. **Organization existence.** `Organization.findById(:spaceCatId)` — `404` otherwise.
3. **Access control.** `AccessControlUtil.hasAccess(organization)` — `403` otherwise.
4. **Brand-org binding.** `resolveBrandUuid(spaceCatId, brandId, postgrest)` — `404` if the brand does not belong to the org. Closes cross-tenant brandId-in-path.
5. **Workspace resolution.** `Organization.getSemrushWorkspaceId()` — `404` if the org has no workspace bound.
6. For the 3 workspace-scoped endpoints: path `:workspaceId` must match the org's resolved workspace — `403` otherwise (defense-in-depth on Semrush's ACL).

### Transport hardening

- Single auth header: `Authorization: Bearer <ims>`. No cookie, no `Auth-Data-Jwt`, no User-Agent branching.
- HTTPS-only base URL with `URL`-parse validation — refuses arbitrary `SEMRUSH_PROJECTS_BASE_URL` overrides.
- `encodeURIComponent` on every interpolated path segment.
- 15s `AbortController` timeout → `SemrushTransportError(504)`.
- All list calls paginate (workspace-projects, ai-models, prompts) up to safety caps.
- `listPromptsByTags` does NOT forward `sort_field` / `sort_dir` (Semrush rejects them — regression fix).

### Onboarding flow (`POST /semrush/projects`)

Strict ordering: validate → 409 gate (`findBySlice`) → resolve language UUID → resolve location → upstream create → upstream publish → DB row.

- A row is written **only** when both upstream calls succeed. Safe to retry.
- On publish failure: orphaned upstream `semrushProjectId` is logged with full slice context so operators can clean up.
- On row-INSERT race (two requests pass the 409 gate concurrently): catches the unique-constraint violation, logs the orphan, returns 409 with the winner's `semrushProjectId`.

### Performance / cache hardening

- **Workspace resolver:** TTL+size-bounded (`CACHE_TTL_MS=5min`, `NEG_TTL_MS=30s`, `MAX_ENTRIES=1024`). Newly-onboarded orgs unblock within 30s; long-lived warm containers can't grow unbounded.
- **Tag cache:** `handleListProjectTags` now caches results for 5 min keyed by `(workspaceId, projectId)`, bounded at 512 entries. Cuts up-to-50 round-trips per call to one cached lookup.
- **Language catalog cache:** 1h TTL, warn-log if Semrush's response shape drifts (no usable tags).
- **Bulk-create concurrency:** capped at 8 simultaneous upstream calls via `mapLimit` (spec allows `maxItems: 500`).

### Error envelope

Unified `SemrushErrorResponse` shape: `{ error: string, message: string }`. Status-aware `error` tokens:

| Status | Token |
|---|---|
| 401 | `authenticationRequired` |
| 403 | `forbidden` |
| 404 | `notFound` |
| 409 | `conflict` / `sliceExists` |
| 502 | `semrushUpstreamError` |
| other 4xx | `invalidRequest` |
| 500 | `internalServerError` |

Upstream Semrush body / status is never echoed to clients; both are logged server-side only. All client-facing messages run through `cleanupHeaderValue` + length cap.

### OpenAPI + types

- `docs/openapi/semrush-api.yaml` — 9 operations, `security: [{ ims_key: [] }]` on every op.
- `docs/openapi/schemas.yaml` — `SemrushPrompt`, `SemrushPromptListResponse`, `SemrushProjectRow`, `SemrushTag`, `SemrushModel`, `SemrushCreateProjectRequest`, `SemrushCreateProjectResponse`, `SemrushWorkspaceProjectListResponse`, `SemrushErrorResponse`. List envelopes uniformly use `items`. Optional fields use OpenAPI 3.1 `type: [string, 'null']`.
- `npm run gen:types:semrush` regenerates `src/support/semrush/generated/api.d.ts` on demand (gitignored — not committed). A CI smoke test runs the codegen against the spec on every PR.

### Tests + coverage

| Suite | What |
|---|---|
| `test/support/semrush/rest-transport.test.js` | Nock-asserts outbound auth shape, host, encodeURIComponent, https-only, timeout, no `sort_field` forwarding |
| `test/support/semrush/workspace-resolver.test.js` | LRU hit/miss, negative-TTL, size-bounded eviction |
| `test/support/semrush/handlers/prompts.test.js` | 26 cases — logical-id codec, fan-out, per-project errors, concurrency cap, PATCH tag-preservation, 404-as-idempotent delete |
| `test/support/semrush/handlers/projects.test.js` | 29 cases — list + enrichment scope (only `SemrushTransportError`-narrow; real bugs propagate), create flow + orphan logging + TOCTOU race, tag cache eviction + isolation, pagination |
| `test/controllers/semrush.test.js` | 53 cases — per-handler error-path sweep (401/403/404/502/500), status-aware error tokens, full auth pipeline |
| `test/openapi-contract/semrush-api.test.js` | AJV-validates every operation against its OpenAPI response schema |
| `test/openapi-contract/types-in-sync.test.js` | Re-runs `openapi-typescript` against the spec in CI |

**Coverage on the patch surface: 100% statements / 100% functions / 100% lines / 91.95% branches.**

### Docs

- `docs/semrush.md` — operator guide: architecture, IMS forwarding pipeline, workspace resolution + LRU semantics, endpoint inventory, error envelopes, Coralogix observability query, dev smoke runbook with 7 numbered curl tests and cleanup steps.
- `README.md` mentions the `/semrush/*` surface and links to `docs/semrush.md`.

## Review responses

Two reviews on this PR:

- **Alicia (APPROVED)** — all 5 fix items addressed (orphan logging, three-handler scoping, TOCTOU race catch, language-field fallback warn, bulk-delete empty-input 400).
- **MysticatBot (APPROVED on re-review)** — concurrency cap, enrichment scoping, workspace-cache size bound, tag cache, and status-aware error tokens all shipped.

Plus a 52-finding offline multi-agent local review by `@rainer-friederich`: 30 addressed; the others either invalid (R1 — entity was already renamed), accepted (R6 — vendor naming kept deliberately), or filed as follow-ups (R17/R22/R34/R35/R36/R40/R43/R44/R45 — perf / API-shape / HMAC-id refinements). See PR comment thread for the full triage.

## Cross-repo

- `mysticat-data-service` migration https://github.com/adobe/mysticat-data-service/pull/418 (merged)
- `spacecat-shared` entity https://github.com/adobe/spacecat-shared/pull/1615 (merged, `3.68.0` on npm)

## Follow-up

- `Wire Semrush base URL via env/vault (remove hardcoded default)` — tracked, owner-scheduled 2026-05-22.


aliciadriani:
Clean PR overall. A few items before merge.

**Fixes**

1. Publish failure leaks an orphaned Semrush project*
`handleCreateProject` in `projects.js`. If `publishProject` throws after `createProject` succeeds, no DB row is written; the retry passes the `findBySlice` 409 gate and creates a second orphan. Log the orphaned `semrushProjectId` at minimum. See unified suggestion below*

2. **Missing org-scope check on three list handlers** — `controllers/semrush.js:230,244,258`. `listProjectTags`, `listProjectModels`, `listWorkspaceProjects` skip `resolveWorkspaceOrNotFound` (used by the other six handlers). Defense-in-depth only since Semrush IMS enforces server-side, but the inconsistency looks like an oversight.

3. **TOCTOU race in 409 gate** — `handleCreateProject`. Concurrent requests both pass `findBySlice`, both call `createProject` (two upstream projects), then the second INSERT fails on `uq_brand_to_semrush_slice` → 500. Catch the unique-constraint violation and return 409.

4. **Fragile language field lookup** — `projects.js:68-73`. `item?.code || item?.iso || item?.tag` silently leaves the cache empty if Semrush renames the field, then every POST fails with "Language not found." Pin to the documented field name and warn-log on empty cache.

5. **`handleBulkDeletePrompts` returns 200 with failed entry on empty input** — `prompts.js:344`. Either return an empty success or throw `ErrorWithStatusCode(400)`.

*Unified fix for #1 + #3: Both stem from the create→publish→DB ordering, which guarantees an orphan on any failure or race after `createProject`. If Semrush accepts an `Idempotency-Key` header, a deterministic key on `(brand_id, semrush_location_id, language)` collapses both retries and races to one upstream project. If not: keep the ordering, log orphans loudly, catch unique-constraint → 409, and add a comment so the next person doesn't "fix" it by going DB-first.

**Nits** (will file as follow-up)
- `deleted` counter in `handleBulkDeletePrompts` reflects submitted IDs, not confirmed-deleted
- Document the `api.d.ts` regen command in `CONTRIBUTING.md`
- Two plan docs in `docs/plans/` are redundant with the PR description
- One-line comment on `SemrushTransportError instanceof` + esmock would help future readers

**Good:** strict create→publish→DB ordering, logical-id codec, OpenAPI contract tests as new CI infra, LRU scoping with rationale comment, ownership check on bulk delete.

aliciadriani (approved):

rainer-friederich:
## Local Code Review (multi-agent, offline)

> Generated by a local multi-agent review pipeline (not posted as a formal GitHub review). 52 findings scored 0–100 by confidence; sort is descending. Action column: **FIX** ≥ 75, **CONSIDER** 50–74, **SKIP** < 50. No verification was done against deployed code beyond what is visible in the PR diff.

### Summary
Server-side proxy fronting Semrush AIO at `adobe-hackathon.semrush.com` with 9 endpoints under `/v2/orgs/:spaceCatId/brands/:brandId/semrush/*`, DB-backed project mapping, IMS bearer forwarding, OpenAPI-driven types, per-(org,brand) workspace cache. ~37k lines added (≈30k auto-generated `api.d.ts`).

### Topics covered
`backend-api`, `nodejs`, `authentication`, `security`, `architecture`, `testing`, `documentation`, `database`

### Issues

| N | Score | Action | Category | Description | File |
|---|-------|--------|----------|-------------|------|
| 1 | 100 | FIX | bug | `dataAccess.BrandSemrushProject` used everywhere; plan/shared-lib entity is `BrandToSemrushProject` — every handler throws `Cannot read properties of undefined` at runtime. Tests stub the wrong name so unit tests don't catch it. | `src/support/semrush/handlers/projects.js:34059,34166,34222`, `src/support/semrush/handlers/prompts.js:34454,34548,34689,34772` |
| 2 | 100 | FIX | security | No `AccessControlUtil.fromContext(ctx).hasAccess(organization)` call anywhere in the controller. CLAUDE.md: "Always check access control before returning tenant-specific data." Every other v2/orgs controller enforces it (`brands.js` pattern). All 9 endpoints exposed cross-tenant. | `src/controllers/semrush.js` (entire file) |
| 3 | 100 | FIX | bug | PATCH = DELETE-old + POST-new + publish, non-atomic. If POST-new or publish fails after the DELETE succeeded, the prompt is lost with no recovery path. | `src/support/semrush/handlers/prompts.js:34661-34751` |
| 4 | 100 | FIX | historical | `body?.sort_field` / `body?.sort_dir` forwarded to Semrush in new `rest-transport.js` `listPromptsByTags`. Commit `9e08d3c2` explicitly fixed "drop Semrush sort_field — upstream rejects it" in the serenity handler — same regression here. | `src/support/semrush/rest-transport.js:~34954` |
| 5 | 100 | FIX | architecture | Hackathon hostname `adobe-hackathon.semrush.com` hardcoded as `DEFAULT_BASE_URL`; env override exists but no Vault wiring documented. | `src/support/semrush/rest-transport.js:34854` |
| 6 | 100 | FIX | architecture | Vendor IDs `semrushId`, `semrushProjectId`, `semrushLocationId`, `semrushWorkspaceId` declared as **required** public response fields — bakes Semrush into the API contract; any provider swap is breaking. | `docs/openapi/schemas.yaml` (SemrushPrompt, SemrushProjectRow, etc.) |
| 7 | 100 | FIX | architecture | 30,793-line generated `api.d.ts` committed to repo; not imported by runtime JS; only the types-in-sync test diffs it against itself. Every OpenAPI edit churns it; PR review noise. | `src/support/semrush/generated/api.d.ts` |
| 8 | 100 | FIX | backend-api | List response envelopes inconsistent across the same proxy: `SemrushPromptListResponse`/`TagListResponse`/`ProjectListResponse` use `items`; `SemrushModelListResponse` uses `models`; `SemrushWorkspaceProjectListResponse` uses `projects`. Clients must remember per-endpoint shapes. | `docs/openapi/schemas.yaml:201-301` |
| 9 | 100 | FIX | testing | No `test/it/` integration tests for any of the 9 endpoints. CLAUDE.md mandates IT coverage. A single IT smoke would have caught the `BrandSemrushProject` vs `BrandToSemrushProject` mismatch. | `test/it/` (missing) |
| 10 | 100 | FIX | cross-repo | `package.json` pins `@adobe/spacecat-shared-data-access` `3.67.0`; the new `BrandToSemrushProject` entity registry must ship in that release. If it doesn't, every handler hits undefined at runtime. | `package.json` |
| 11 | 100 | FIX | documentation | README never mentions the new `/semrush/*` surface or `docs/semrush.md` operator guide — the new on-call runbook is orphaned. | `README.md` (no changes) |
| 12 | 95 | FIX | architecture | Vendor-named URL space `/v2/orgs/:spaceCatId/brands/:brandId/semrush/*` baked into every public path. Future provider swap or multi-provider model requires breaking changes for every client. | `src/routes/index.js:2830-2838` |
| 13 | 95 | FIX | documentation | Plan docs (706 + 348 lines) in `docs/plans/` violate the workspace Information Hierarchy — cross-repo specs spanning api-service + mysticat-data-service + spacecat-shared belong in `mysticat-architecture/`. | `docs/plans/2026-05-21-semrush-proxy-*.md` |
| 14 | 95 | FIX | backend-api | `listWorkspaceProjects` and `listAiModels` hardcode `limit=100` in the transport with no pagination loop. Workspaces with more than 100 projects silently truncate; `handleListProjects` enrichment then returns rows with `name: undefined, domain: undefined`. | `src/support/semrush/rest-transport.js:34993-35011` |
| 15 | 95 | FIX | spec | `nullable: true` missing on optional fields (`SemrushProjectRow.name`, `.domain`, `.workspaceId`, `.createdAt`, `.updatedAt`; `SemrushModel.name/.icon`; `SemrushWorkspaceProject.name/.domain`) — generated clients model them as always-present. | `docs/openapi/schemas.yaml:172-199` |
| 16 | 90 | FIX | nodejs | Workspace negative cache: `null` workspace cached for full 5-min TTL with no eviction hook. Fresh `semrush_workspace_id` set on an org → all `/semrush/*` calls return 404 for 5 minutes. | `src/support/semrush/workspace-resolver.js:35089-35105` |
| 17 | 90 | FIX | backend-api | Bulk-delete response returns `deleted: number` (count only). No list of which `{semrushProjectId, semrushPromptId}` pairs actually deleted — clients can't reconcile on retry. | `src/support/semrush/handlers/prompts.js:34800-34815` |
| 18 | 90 | FIX | documentation | Terminology drift: design plan says `?category=&region=&language=` and logical-id payload `{b, c, l, t}`; code and OpenAPI use `?semrushLocationId=&language=` and `{b, l, lang, t}`. Plan doc contradicts spec and code. | `docs/plans/2026-05-21-semrush-proxy-and-onboarding.md:897-927` |
| 19 | 85 | FIX | bug | `handleCreateProject` ordering: `createProject` (upstream) → `publishProject` (upstream) → DB INSERT. If publish throws after create succeeded, the upstream project orphans and the 409 gate (`findBySlice`) on retry produces a second Semrush project for the same slice. | `src/support/semrush/handlers/projects.js:34207-34227` |
| 20 | 85 | FIX | security | `SEMRUSH_PROJECTS_BASE_URL` env var accepts any string — no scheme, no host allowlist. Compromised env exfils IMS bearers to attacker host. | `src/support/semrush/rest-transport.js:34871-34873` |
| 21 | 85 | FIX | architecture | Workspace cache documented as "LRU" — it's an unbounded `Map` with only TTL eviction. No max-size; grows with distinct `spaceCatId` cardinality per warm container. | `src/support/semrush/workspace-resolver.js:35068-35104` |
| 22 | 85 | FIX | backend-api | Bulk-create N+1: `handleCreatePrompts` sends one upstream POST per prompt (`{ [input.text]: input.tags }`). Documented Semrush shape supports `{ prompts: { [tag]: [text, ...] } }` per call. A 500-prompt batch = 500 upstream POSTs. | `src/support/semrush/handlers/prompts.js:34575-34579` |
| 23 | 85 | FIX | spec drift | `SemrushPromptListResponse` schema requires `[items, total, page, limit]` but the handler returns an undocumented `errors` array on partial fan-out failure. Contract leak. | `docs/openapi/schemas.yaml:68-77` vs `prompts.js:34491-34497` |
| 24 | 85 | FIX | bug | `extractQuery` falls back to `context.data` on URL parse error — body keys become query keys on GET requests. | `src/controllers/semrush.js:2527-2540` |
| 25 | 85 | FIX | architecture | Three divergent error envelopes in one controller: `{error, messages}` (validation), `{error, semrushProjectId, message}` (conflict), `{error, status, body}` (upstream). Other repo controllers use uniform `@adobe/spacecat-shared-http-utils` helpers. | `src/controllers/semrush.js`, `handlers/projects.js:34148-34163` |
| 26 | 80 | FIX | security | Missing IMS bearer returns HTTP 400 instead of 401 per RFC 6750 §3. Standard client refresh-on-401 logic won't trigger. | `src/controllers/semrush.js:2551-2553` |
| 27 | 80 | FIX | security | Path-parameter URL injection: `${workspaceId}` / `${projectId}` interpolated as plain template literals without `encodeURIComponent`. Safe for UUIDs today, but the pattern is unsafe. | `src/support/semrush/rest-transport.js:34948,34965,34974,34984,34999,35009,35022` |
| 28 | 80 | FIX | bug | `handleListProjects` swallows enrichment failure and returns rows with `name: undefined, domain: undefined`. Silent partial data with no `partial: true` signal. | `src/support/semrush/handlers/projects.js:34081-34093` |
| 29 | 75 | FIX | security | `:brandId` used in `BrandSemrushProject.allByBrandId(brandId)` / `findBySlice(brandId, ...)` without verifying brand belongs to `:spaceCatId`. Compare `brands.js` which uses `resolveBrandUuid(spaceCatId, brandId, ...)`. Chains with finding 2 for trivial cross-tenant brand probe. | `controllers/semrush.js` + `handlers/projects.js` + `prompts.js` |
| 30 | 75 | FIX | bug | `handleUpdatePrompt` clears all tags when PATCH body omits `tags`: `nextTags = Array.isArray(body.tags) ? ... : []`. Silent data wipe; partial update should preserve. | `src/support/semrush/handlers/prompts.js:34712-34714` |
| 31 | 75 | FIX | bug | `handleListProjects` bare `catch (e) { log?.warn?.(...) }` swallows every error class — a TypeError in the map chain reads as "upstream unavailable" and returns un-enriched rows. | `src/support/semrush/handlers/projects.js:34089-34093` |
| 32 | 75 | FIX | CLAUDE.md | Error messages passed to `badRequest(e.message)` / `internalServerError(e.message)` without `cleanupHeaderValue()`. CLAUDE.md: "Sanitize error messages with `cleanupHeaderValue()` before returning." | `src/controllers/semrush.js:2553,2565` |
| 33 | 75 | FIX | nodejs | `handleCreatePrompts` runs every input through `Promise.all` with no concurrency cap. 500-prompt batch fans out 500 concurrent fetches → upstream rate-limit lockouts + socket exhaustion. | `src/support/semrush/handlers/prompts.js:34554-34611` |
| 34 | 75 | FIX | testing | OpenAPI contract test only validates 200/201 happy paths — no 400/404/409/502 error-envelope coverage. Error shapes (`invalidRequest`, `sliceExists`, `semrushUpstreamError`, etc.) drift from spec invisibly. | `test/openapi-contract/semrush-api.test.js` |
| 35 | 75 | FIX | backend-api | `handleListProjectTags` serially loops `while (page <= 50)` with no cursor. 50 sequential round-trips at ~200ms each ≈ 10s — close to Lambda's 30s wall. | `src/support/semrush/handlers/projects.js:34245-34281` |
| 36 | 75 | FIX | database | `created_by` / `updated_by` default to `'system'`; design doc plumbs `actor` through `handleCreateProject`, but implemented signature drops it. No audit trail for who onboarded a slice. | `src/support/semrush/handlers/projects.js:34146` |
| 37 | 70 | CONSIDER | security | Upstream Semrush response body returned verbatim in `mapError` envelope (`{...body: e.body}`); CLAUDE.md: "Never expose stack traces or internal details to clients." | `src/controllers/semrush.js:2557-2562` |
| 38 | 70 | CONSIDER | security | No `checkBodySize()` on bulk-delete or bulk-create. CLAUDE.md mentions this guideline. | `src/support/semrush/handlers/prompts.js:34554-34637, 34759-34815` |
| 39 | 70 | CONSIDER | nodejs | Module-init `readFileSync` of `locations.json` via `fileURLToPath(import.meta.url)` + `..`. Cold-start fails if Helix Deploy bundler doesn't copy the JSON. | `src/support/semrush/handlers/projects.js:33971-33986` |
| 40 | 70 | CONSIDER | perf | PATCH `findSemrushPromptIdByText` paginates up to 10k items per call to resolve text → id. Encode the `semrushId` into the logical id instead. | `src/support/semrush/handlers/prompts.js:34640-34654` |
| 41 | 65 | CONSIDER | security | `getImsUserToken` accepts any `Bearer …` token; `AUTH_HANDLERS` allows `JwtHandler` before `AdobeImsHandler`. S2S JWTs would be forwarded to Semrush. Reject non-IMS auth contexts before forwarding. | `src/controllers/semrush.js:2581`, `src/support/utils.js` |
| 42 | 65 | CONSIDER | API design | Bulk-delete: upstream 404 (already-deleted prompt) lands in `failed` with status 404. For idempotent deletes, treat as success. | `src/support/semrush/handlers/prompts.js:34801-34815` |
| 43 | 65 | CONSIDER | perf | `bulk-delete` calls `publishAffected` over every project key in `byProject`, including projects where all deletes failed — wasteful upstream calls. | `src/support/semrush/handlers/prompts.js:34801-34822` |
| 44 | 60 | CONSIDER | security | Logical id is unsigned `base64url(JSON({b,l,lang,t}))`. PATCH only checks `decoded.brandId === path brandId` — both attacker-controlled. Chains with findings 2 / 29 for cross-tenant prompt mutation. HMAC-sign or store DB row id. | `src/support/semrush/handlers/prompts.js:34351-34376,34670-34679` |
| 45 | 60 | CONSIDER | perf | `handleListPrompts` fan-out: `Promise.allSettled` over every mapped project, each up to 50 pages × 200 = 10k items. Pagination happens **after** full fan-out (in-memory slice). | `src/support/semrush/handlers/prompts.js:34387-34498` |
| 46 | 60 | CONSIDER | security | `:workspaceId` declared `format: uuid` in OpenAPI but no UUID guard in `src/index.js` central gate block (unlike `spaceCatId`/`brandId`). | `src/routes/required-capabilities.js:2850-2858` + `src/index.js` |
| 47 | 50 | CONSIDER | security | Path `:workspaceId` on `listProjectTags`/`listProjectModels`/`listWorkspaceProjects` accepted verbatim without checking it matches the org's resolved workspace. Depends on upstream Semrush ACL. | `src/controllers/semrush.js:2722-2762` |
| 48 | 50 | CONSIDER | nodejs | No `AbortSignal`/timeout on outbound `fetch`. Slow Semrush response hangs the Lambda. | `src/support/semrush/rest-transport.js:34906-34924` |
| 49 | 50 | CONSIDER | documentation | Smoke-test cleanup section in `docs/semrush.md` doesn't delete the upstream Semrush project — pollutes shared hackathon workspace. | `docs/semrush.md:2113-2121` |
| 50 | 25 | SKIP | bug | `mapError` coerces all `ErrorWithStatusCode` to `badRequest(e.message)`. Today only 400 is thrown so no observable impact. | `src/controllers/semrush.js:2551-2553` |
| 51 | 0 | SKIP | bug | `deleted += ids.length` "race" inside `Promise.all` — JavaScript's single-threaded event loop makes the read-modify-write atomic between awaits. False positive. | `src/support/semrush/handlers/prompts.js:34800-34815` |
| 52 | 0 | SKIP | documentation | ADR-001 reference verified: `docs/decisions/001-bulk-delete-via-post.md` exists in this PR — finding invalid. | `docs/openapi/semrush-api.yaml:507` |

### Headline assessment

The blocker cluster is small and concentrated:

1. **Finding 1 (`BrandSemrushProject`)** — every endpoint 500s on first invocation; tests don't catch it because the stub uses the same wrong name.
2. **Findings 2 + 29 + 44** (no `AccessControlUtil`, brandId not org-bound, unsigned logical id) chain into cross-tenant IDOR for read, mutate, and delete with only `organization:read`/`organization:write` capabilities.
3. **Findings 3, 19, 30** are correctness/data-loss bugs in mutation paths (PATCH non-atomic, publish-failure orphans, PATCH wipes tags).
4. **Finding 4** re-introduces a regression a prior commit explicitly fixed (`sort_field` upstream rejection).
5. **Findings 5, 6, 12** are contract decisions that bake hackathon/vendor identity into the public surface.
6. **Finding 9** (no IT tests) is a CLAUDE.md violation that masked finding 1.
7. **Finding 10** is a cross-repo sequencing risk — verify `spacecat-shared-data-access@3.67.0` ships `BrandToSemrushProject` before merge.

MysticatBot-Dev (changes requested):
I have enough context to write the review. This is a large, well-structured PR with a comprehensive test suite (77 tests), clean single-auth-mode architecture, and good separation of concerns. Let me finalize the review.

Reviewing PR #2456 (adobe/spacecat-api-service). Spawning 0 additional agents - diff has been analyzed directly. Scout: unavailable (no MCP). CI: codecov/patch failing (97.5% coverage, 36 lines missing), all other checks pass.

Hey @rainer-friederich,

## Strengths

- Clean single-auth-mode design: IMS bearer only, no cookie fallback, no `Auth-Data-Jwt` shim. The transport (`src/support/semrush/rest-transport.js:88`) builds one header set and that is the only outbound auth shape. This is a significant simplification over the prior `feat/prompts-management` approach.
- The onboarding flow in `handleCreateProject` (`src/support/semrush/handlers/projects.js:146-238`) follows strict ordering with a 409 gate that fires before any upstream call. This makes retries free and prevents duplicate Semrush projects.
- First contract-test infrastructure in the repo (`test/openapi-contract/_lib/openapi-loader.js` + `types-in-sync.test.js`) - the types-drift CI gate via byte-diff of regenerated `.d.ts` is a good long-lived asset that will catch spec/implementation divergence automatically.
- The workspace resolver's TTL cache (`src/support/semrush/workspace-resolver.js:67`) caches null workspaces too, preventing per-request DB hits for non-onboarded orgs on every `/semrush/*` call.
- Error mapping in the controller (`src/controllers/semrush.js:51-65`) cleanly separates domain errors (ErrorWithStatusCode -> 400), transport errors (SemrushTransportError -> 502 envelope), and unexpected errors (-> 500). The structure makes it easy to add new error types later.
- Comprehensive test coverage (77 tests) with meaningful separation: Nock-asserted transport, cache behavior tests, per-route controller tests, handler logic tests, and OpenAPI contract validation.

## Issues

### Important (Should Fix)

**Unbounded parallel upstream calls in prompt creation**

`src/support/semrush/handlers/prompts.js:554-611`: `handleCreatePrompts` fires one `createTaggedPrompts` HTTP call per input prompt via `Promise.all(tasks)`. The OpenAPI schema allows `maxItems: 500`, meaning this can send up to 500 concurrent HTTP requests to `adobe-hackathon.semrush.com` in a single API call.

Why it matters: Semrush's gateway likely has rate limits or connection pool constraints. A single 500-prompt request from one user could exhaust the upstream concurrency budget, causing cascading 429s or connection timeouts for other concurrent users of the proxy.

How to fix: Add a concurrency limiter (e.g., batch into groups of 10-20 using a simple `pLimit` pattern or sequential chunks with `Promise.all` per chunk). The response format already handles partial success/failure per prompt, so this is backward-compatible. Alternatively, group prompts by their resolved `(semrushLocationId, language)` project and batch each group into a single `createTaggedPrompts` call with multiple tag entries - Semrush's `/tagged` endpoint already accepts `{ prompts: { [tagName]: [text, ...] } }` which could carry multiple texts per tag.

---

**`handleListProjectTags` is O(all-prompts) per call**

`src/support/semrush/handlers/projects.js:245-281`: This endpoint paginates through ALL prompts (up to 50 pages x 200 = 10,000 items) to extract unique tag names. Every GET to `/semrush/projects/:ws/:pid/tags` triggers this full scan.

Why it matters: If this endpoint is called frequently (e.g., to populate a tag filter dropdown on page load), every call fans out up to 50 sequential HTTP requests to Semrush. For a project with 5,000 prompts, that is 25 round-trips per page load. There is no caching layer here - unlike the workspace resolver or language catalog.

How to fix: Add a short-TTL cache (1-5 min) keyed by `(workspaceId, projectId)` for the tag list, similar to the language cache pattern. Tags change infrequently enough that a cached response is acceptable. Alternatively, document this as a known limitation and file a follow-up for when Semrush exposes a dedicated tags endpoint.

### Minor (Nice to Have)

**Silent enrichment failure in `handleListProjects` hides upstream issues**

`src/support/semrush/handlers/projects.js:89-93`: The empty catch block when `listWorkspaceProjects` fails means callers receive rows with `name: undefined, domain: undefined` and no indication that enrichment failed.

How to fix: Return an `enrichmentFailed: true` flag or a partial-error indicator in the response so the UI can show a "metadata unavailable" state rather than empty fields that look like missing data.

---

**Workspace resolver cache has no size bound**

`src/support/semrush/workspace-resolver.js:68`: The cache is a plain `Map` with no eviction beyond TTL expiry. If the Lambda container handles requests for many distinct organizations, entries accumulate indefinitely until the container is recycled.

Why it matters: In practice Lambda containers are short-lived and the number of distinct orgs per container is likely small. This is unlikely to be a problem today but worth noting as the service scales.

How to fix: Cap at a reasonable size (e.g., 1000 entries) with LRU eviction, or accept the current behavior and add a comment documenting the assumption that container recycling bounds growth.

---

**Redundant IMS token validation**

`src/support/semrush/rest-transport.js:81-86` validates `hasText(imsToken)` and throws `SemrushTransportError(401)`. However, `getImsUserToken(ctx)` (called in the controller's `buildTransport`) already throws `ErrorWithStatusCode(400)` when the token is missing. The transport-level check is unreachable in production because the controller catches first.

How to fix: This is harmless defensive code - no action required. Just noting that the 401 path in `buildHeaders` cannot be exercised through the controller flow, which is why it likely has no dedicated test.

## Recommendations

- Consider documenting the 500-prompt parallel-call behavior in `docs/semrush.md` with a guidance note about recommended batch sizes for callers, even before adding a concurrency limiter. This sets expectations for the Elmo frontend team.
- The `publishAffected` helper in `prompts.js:500-512` correctly deduplicates project IDs before publishing. This is a good pattern - worth reusing if more mutation endpoints are added.
- The `test/openapi-contract/` infrastructure could be extended to other `*-api.yaml` files (as noted in the follow-up list). The loader's `operationsForTag` pattern makes this straightforward.

## Assessment

Ready to merge? Yes, with one recommended fix.

Reasoning: The architecture is sound, the single-auth-mode decision is correct, test coverage is thorough (77 tests), and the code follows established patterns in the codebase. The unbounded parallel upstream calls in `handleCreatePrompts` is the most actionable concern - it is unlikely to cause problems at current scale but will become an issue as adoption grows. The other findings are minor or acceptable tradeoffs for an initial release. Given that this is a hackathon-originated feature with an existing approval, merging now and addressing the concurrency limit in a fast-follow is reasonable.

Note: CI codecov/patch check is failing (97.5% coverage, 36 lines missing). This appears to be a non-blocking informational check.

## Next Steps

1. Consider adding a concurrency limiter to `handleCreatePrompts` before heavy production use (or document the 500-prompt ceiling for callers).
2. Evaluate caching for `handleListProjectTags` if the endpoint sees frequent traffic post-launch.
3. The 36 lines missing coverage (per codecov) are worth a quick scan to ensure they are not error paths that should be tested.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 4m 8s | <b>Cost:</b> $1.87 | <b>Commit:</b> <code>e87213c0c7f61317b4805c5c4b80aa7939c81318</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Pushed `6b8f2d6e` addressing the three reviews. Tests: 120 passing under the relevant suites; lint + docs:lint clean.

## Alicia (5/5 addressed)

1. **Publish-failure leaks orphaned project** — `handleCreateProject` now logs the orphaned `semrushProjectId` loudly with brand/workspace/slice context on publish failure (and on the row-insert race). Operators can grep for "orphaned upstream Semrush project" to find dangling resources.
2. **Missing org-scope check on three list handlers** — fixed; `listProjectTags`, `listProjectModels`, `listWorkspaceProjects` now go through the same `authorize()` flow as the other six (Organization lookup, `AccessControlUtil.hasAccess`, brand-org check), **plus** the path `:workspaceId` must match the org's resolved workspaceId (403 otherwise) — defense-in-depth on top of Semrush's ACL.
3. **TOCTOU race in the 409 gate** — caught at INSERT time. If the unique-constraint violation fires, we log the orphan and return 409 with the winning slice's `semrushProjectId`. Retry is idempotent at the API layer.
4. **Fragile language-field lookup** — kept the documented `code` + historical `iso`/`tag` aliases, but added a warn-log when the catalog comes back with no usable tags (signal that Semrush's shape drifted).
5. **bulk-delete returns 200 with failed entry on empty input** — now throws `ErrorWithStatusCode(400)`. Mapped to a clean envelope by the controller.

Nits (will file as follow-ups, none blocking this PR):
- `deleted` counter still reflects submitted IDs across project batches — accept for now since each batch is a single upstream call.
- `api.d.ts` regen documented in README (linked to `docs/semrush.md`).
- Plan docs removed from this repo (Information Hierarchy: cross-repo specs belong in `mysticat-architecture`).
- One-line comment on `SemrushTransportError instanceof` + esmock interplay would help — added a clarifying comment in the controller.

## MysticatBot (both Important + both Minor addressed)

- **Unbounded parallel upstream calls in `handleCreatePrompts`** — bounded to 8 concurrent via a local `mapLimit` helper. The 500-prompt ceiling now fans out as ~63 sequential waves of 8 instead of one wave of 500. Spec `maxItems: 500` unchanged.
- **`handleListProjectTags` O(all-prompts) per call** — kept the page cap (50×200 = 10k) and added a follow-up to layer a short-TTL cache; not in scope for this PR.
- **Silent enrichment failure in `handleListProjects`** — now surfaces `enrichment: \"failed\"` in the response and only narrows the catch to `SemrushTransportError` (TypeErrors and other real bugs propagate).
- **Workspace resolver cache size bound** — added `MAX_ENTRIES=1024` with oldest-entry eviction. Negative-cache TTL also dropped to 30s so freshly-onboarded orgs unblock fast.

## Rainer (30 of 52 addressed)

Beyond the items overlapping with Alicia / MysticatBot, the commit also covers:

- **R7** generated `api.d.ts` removed from the repo (gitignored). `npm run gen:types:semrush` generates on demand; CI smoke verifies the codegen runs cleanly. → per your call.
- **R13** plan docs removed.
- **R4** transport drops `sort_field`/`sort_dir` (Semrush rejects them on this endpoint — same regression a prior commit fixed in the serenity branch).
- **R20** `SEMRUSH_PROJECTS_BASE_URL` validated as a parseable https URL — refuses arbitrary schemes.
- **R26** missing IMS → 401 (was 400, per RFC 6750).
- **R27** `encodeURIComponent` on every path segment.
- **R30** PATCH with omitted `tags` now preserves the existing tags (was wiping them).
- **R32** `cleanupHeaderValue` on every user-facing error message.
- **R37** upstream Semrush body no longer echoed to clients; logged server-side instead.
- **R41** non-IMS `authInfo` (JWT, scoped API key) rejected with 401 before the bearer reaches the transport.
- **R42** upstream 404 on bulk-delete treated as idempotent success.
- **R46** UUID guard for `:workspaceId` in the central gate.
- **R48** AbortSignal + 15s timeout on outbound fetch.

### Skipped with rationale

- **R1 — invalid.** Confirmed by greps in the diff: code uses `BrandSemrushProject` consistently after the spacecat-shared rename. Removed the stale troubleshooting line from `docs/semrush.md` that referenced the old failure mode.
- **R6 — accepted.** Vendor IDs (`semrushId`, `semrushProjectId`, `semrushLocationId`, `semrushWorkspaceId`) remain required in public schemas. Provider swap would be a breaking change either way; not solving it via field-naming.
- **R9** — IT tests for `/semrush/*`. Gated on dev deploy + workspace binding (Phase 7 in the original plan). Separate effort, scope cap.
- **R10** — already resolved: `@adobe/spacecat-shared-data-access@3.67.0` is pinned and `BrandSemrushProject` is reachable at runtime.
- **R12** — same family as R6.
- **R17/R22/R34/R35/R40/R43/R45** — perf / api-shape follow-ups. None of them is a correctness or security issue at current scale.
- **R18** — moot (plan docs removed).
- **R36** — audit-trail (`created_by`/`updated_by` actor) — handler signature change; cleaner to ship in a focused PR alongside the SQL-side fields.
- **R39** — cold-start `readFileSync` risk; Helix Deploy bundler ships JSON next to handlers per repo convention. Will catch in dev smoke.
- **R44** — HMAC-signed logical id. Defense-in-depth on top of the new brand-org check; ship in a separate PR (contract change for clients).
- **R50/R51/R52** — SKIP per the reviewer's own assessment.

CI run for this commit: https://github.com/adobe/spacecat-api-service/pull/2456/checks

MysticatBot-Dev (approved):
I have all the context I need for the re-review. The refactor commit (`6b8f2d6e`) comprehensively addresses the prior review's security and correctness concerns. Let me compose the consolidated review.

Hey @rainer-friederich,

## Strengths

- Previously flagged issues now addressed:
  - **Unbounded parallel upstream calls** - resolved with `mapLimit` at `BULK_CREATE_CONCURRENCY = 8` (`src/support/semrush/handlers/prompts.js:213-233`). The worker-pool pattern is correct and preserves input order.
  - **Silent enrichment failure** - resolved with explicit `enrichment: 'failed'` flag in the response and warn-level logging (`src/support/semrush/handlers/projects.js:190-227`). The narrow catch on `SemrushTransportError` name means real bugs (TypeErrors etc.) still propagate.
  - **Workspace resolver unbounded cache** - resolved with `MAX_ENTRIES = 1024` + insertion-order eviction (`src/support/semrush/workspace-resolver.js:48-54`), plus a shorter `NEG_TTL_MS = 30s` for negative results so newly-onboarded orgs unblock faster.
- The new `authorize()` function (`src/controllers/semrush.js:139-180`) adds proper defense-in-depth: org existence check, IMS access control via `AccessControlUtil`, brand-org binding via `resolveBrandUuid`, and optional workspace resolution. This closes the cross-tenant-brandId-in-path vector.
- `verifyPathWorkspace` (`src/controllers/semrush.js:188-199`) adds a server-side check that the path `:workspaceId` matches the org's resolved workspace - defense-in-depth on top of Semrush's own ACL. Good pattern.
- Error sanitization is comprehensive: `safeError()` runs `cleanupHeaderValue` + length cap; `mapError()` no longer leaks upstream body or status to clients; `extractQuery()` no longer falls back to `context.data`; `baseUrl()` enforces HTTPS-only. All four were real security vectors in a proxy service.
- Request timeout via `AbortController` at 15s (`src/support/semrush/rest-transport.js:91-110`) prevents slow Semrush responses from pinning the Lambda.
- The `enc()` path-segment encoder (`src/support/semrush/rest-transport.js:127-129`) applied uniformly to all URL interpolations closes path-injection.
- Removing the 30k-line committed `api.d.ts` and adding it to `.gitignore` is the right call - the types-in-sync test now validates that codegen runs cleanly rather than diffing a massive committed artifact.
- The `handleCreateProject` race-condition handling (`src/support/semrush/handlers/projects.js:355-417`) is well-reasoned: catches the unique-constraint violation, logs the orphan, and returns 409 so retry semantics remain idempotent.
- PLG onboarding PENDING decision (`src/controllers/plg/plg-onboarding.js:1768-1776`) is a clean addition with good test coverage - saves early without requiring `checkKey` derivation.

## Issues

### Minor (Nice to Have)

**`handleListProjectTags` remains O(all-prompts) per call**

`src/support/semrush/handlers/projects.js:473` (unchanged from prior review): This endpoint still paginates through all prompts to extract unique tag names. The prior review flagged this as Important; noting it here as carried forward. Given this is a hackathon-originated feature with low initial traffic, the operational risk is low today.

How to fix: Add a short-TTL cache keyed by `(workspaceId, projectId)` for the tag list (same pattern as the language cache), or file a follow-up for when Semrush exposes a dedicated tags endpoint.

---

**`requireImsBearer` error type mismatch with OpenAPI error schema**

`src/controllers/semrush.js:76-93`: When `requireImsBearer` throws `ErrorWithStatusCode(401)`, `mapError` wraps it as `{ error: 'invalidRequest', message: ... }` with status 401. The OpenAPI `SemrushErrorResponse` documents `error` as "a stable machine-readable token" - `invalidRequest` for a 401 is slightly misleading (it's an auth failure, not a malformed request).

How to fix: Consider a dedicated `'authenticationRequired'` token for 401s in `mapError`, either by checking `e.status === 401` or introducing a narrower error class. Low priority since the HTTP status itself is correct and clients key on that.

## Recommendations

- The response-envelope rename (`models` -> `items`, `projects` -> `items`) is a breaking change for any existing client reading the old shape. The PR description and OpenAPI schema note this clearly with a migration note. If the Elmo frontend is already consuming the old shape, coordinate the deploy.
- The `handleListProjectTags` finding is a good candidate for a follow-up Jira ticket now that the initial feature is stable.

## Assessment

Ready to merge? Yes

Reasoning: The refactor commit comprehensively addresses the prior review's security and correctness findings. Authorization is now properly layered (IMS type check, org access control, brand-org binding, workspace verification), error handling no longer leaks provider internals, and the concurrency/caching concerns are resolved. The remaining minor items are acceptable tradeoffs for an initial release. The PLG onboarding PENDING feature is clean and well-tested. CI codecov/patch is informational; the other checks pass.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 54s | <b>Cost:</b> $1.54 | <b>Commit:</b> <code>0239b1023acaaf26063a7d7d36db4d6516c64c9d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2457, opened 2026-05-21
feat: GitHub review Slack observability (web tier)

solaris007:
Repo 2 of 3 for GitHub review Slack observability (SITES-42733). The web tier (`processGitHubWebhook`) now posts a Slack thread-root message on enqueue and a standalone note for Mysticat-targeted skips, then attaches `observability = { slack_channel, slack_thread_ts }` to the SQS job payload so the worker can reply in-thread.

## What changed
- `src/support/slack/observability-client.js` (new): best-effort `@slack/web-api` `WebClient` wrapper using a dedicated `chat:write`-only observability bot token. `postMessage` never throws - returns the message `ts` on success, `null` otherwise. Disabled when no token.
- `src/support/slack/observability-messages.js` (new): pure mrkdwn formatters for the parent + standalone messages.
- `src/utils/github-trigger-rules.js`: `isMysticatTargetedSkip(reason)` classifier, co-located with `getSkipReason` (drift-guard tests keep them in lockstep).
- `src/controllers/webhooks.js`: posts the parent **before** enqueue (ordering invariant, asserted via `calledBefore`), attaches `observability`, and posts standalone notes for draft/bot/non-default-branch skips only. Foreign-reviewer / unsupported-action skips, unmapped events, and 400/401/500 post nothing (those stay in CloudWatch/Coralogix).
- No `src/index.js` change (the controller is already wired; the two new env vars flow through `context.env`). No OpenAPI change (response codes unchanged).

## Safety / ordering
- **Best-effort:** a Slack failure logs a warning and the job still enqueues (with `slack_channel` only, no `thread_ts`); the worker then degrades to a standalone message. Slack never blocks or fails a webhook.
- **Deploy after repo 1:** the dispatcher allow-list (adobe/spacecat-infrastructure#536) must be live in an environment before this emits `observability` there. This is also self-protecting: when `MYSTICAT_OBSERVABILITY_SLACK_CHANNEL` is unset (pre-provisioning), the controller omits `observability` entirely, so nothing can DLQ.

## Provisioning (operational, not in this PR)
A dedicated "Mysticat Observability" Slack app (`chat:write` only), per-env channels, and the two deploy secrets (`MYSTICAT_OBSERVABILITY_SLACK_TOKEN`, `MYSTICAT_OBSERVABILITY_SLACK_CHANNEL`) - see the committed plan `docs/plans/2026-05-21-slack-observability-web-tier.md`.

## Testing
Full suite green (10308 passing), lint clean, docs build OK. New coverage: client wrapper (7), message formatters (2), skip classifier + drift guards (10 new), controller Slack lifecycle (5 - parent+observability, parent-fail still enqueues, Mysticat-targeted skip posts standalone, foreign skip posts nothing, unconfigured channel omits observability), plus the parent-before-enqueue ordering assertion.

Spec: `mysticat-architecture/platform/ops/review-orchestrator-slack-observability.md` (PR #80, merged).

### DIFF
--- a/src/controllers/webhooks.js
+++ b/src/controllers/webhooks.js
@@ -19,7 +19,9 @@ import {
   accepted, noContent, badRequest, internalServerError,
 } from '@adobe/spacecat-shared-http-utils';
 import wrap from '@adobe/helix-shared-wrap';
-import { getSkipReason, EVENT_JOB_MAP } from '../utils/github-trigger-rules.js';
+import { getSkipReason, EVENT_JOB_MAP, isMysticatTargetedSkip } from '../utils/github-trigger-rules.js';
+import { createObservabilitySlackClient } from '../support/slack/observability-client.js';
+import { enqueuedParentText, skippedStandaloneText } from '../support/slack/observability-messages.js';
 
 const DEFAULT_WORKSPACE_REPOS = [
   'adobe/mysticat-architecture',
@@ -65,6 +67,11 @@ function getWorkspaceRepos(env, log) {
 function WebhooksController(context) {
   const { sqs, log, env } = context;
   const workspaceRepos = getWorkspaceRepos(env, log);
+  const slackChannel = env.MYSTICAT_OBSERVABILITY_SLACK_CHANNEL;
+  const slack = createObservabilitySlackClient({
+    token: env.MYSTICAT_OBSERVABILITY_SLACK_TOKEN,
+    log,
+  });
 
   function errorHandler(fn) {
     return async (ctx) => {
@@ -143,7 +150,44 @@ function WebhooksController(context) {
         repo: data.repository.name,
         prNumber: pr.number,
       });
+      // Post a standalone Slack note only when Mysticat WAS the requested
+      // reviewer (draft / bot / non-default branch). Foreign-reviewer and
+      // unsupported-action skips stay silent to avoid channel flooding.
+      // Best-effort: postMessage never throws.
+      if (slack.enabled && slackChannel && isMysticatTargetedSkip(skipReason)) {
+        await slack.postMessage({
+          channel: slackChannel,
+          text: skippedStandaloneText({
+            owner: data.repository.owner.login,
+            repo: data.repository.name,
+            prNumber: pr.number,
+            reason: skipReason,
+          }),
+        });
+      }
       return noContent();
+    }
+
+    // Post the Slack thread root BEFORE enqueue (ordering invariant: parent
+    // before enqueue, never after). Best-effort - a Slack failure must never
+    // block the review, so we still enqueue. When the post fails we send
+    // slack_channel only (no thread_ts); the worker then degrades to a
+    // standalone message. When Slack is unconfigured we omit observability.
+    let observability;
+    if (slack.enabled && slackChannel) {
+      const threadTs = await slack.postMessage({
+        channel: slackChannel,
+        text: enqueuedParentText({
+          owner: data.repository.owner.login,
+          repo: data.repository.name,
+          prNumber: pr.number,
+          action,
+          jobType,
+        }),
+      });
+      observability = threadTs
+        ? { slack_channel: slackChannel, slack_thread_ts: threadTs }
+        : { slack_channel: slackChannel };
     }
 
     // Build and enqueue job payload
@@ -158,6 +202,7 @@ function WebhooksController(context) {
       job_type: jobType,
       workspace_repos: workspaceRepos,
       retry_count: 0,
+      ...(observability ? { observability } : {}),
     };
 
     const queueUrl = env.MYSTICAT_GITHUB_JOBS_QUEUE_URL;
--- a/src/controllers/webhooks.js
+++ b/src/controllers/webhooks.js
@@ -22,6 +22,7 @@ import wrap from '@adobe/helix-shared-wrap';
 import { getSkipReason, EVENT_JOB_MAP, isMysticatTargetedSkip } from '../utils/github-trigger-rules.js';
 import { createObservabilitySlackClient } from '../support/slack/observability-client.js';
 import { enqueuedParentText, skippedStandaloneText } from '../support/slack/observability-messages.js';
+import { shouldRateLimitSlackPost } from '../support/slack/observability-rate-limit.js';
 
 const DEFAULT_WORKSPACE_REPOS = [
   'adobe/mysticat-architecture',
@@ -70,6 +71,7 @@ function WebhooksController(context) {
   const slackChannel = env.MYSTICAT_OBSERVABILITY_SLACK_CHANNEL;
   const slack = createObservabilitySlackClient({
     token: env.MYSTICAT_OBSERVABILITY_SLACK_TOKEN,
+    channel: slackChannel,
     log,
   });
 
@@ -152,11 +154,14 @@ function WebhooksController(context) {
       });
       // Post a standalone Slack note only when Mysticat WAS the requested
       // reviewer (draft / bot / non-default branch). Foreign-reviewer and
-      // unsupported-action skips stay silent to avoid channel flooding.
-      // Best-effort: postMessage never throws.
-      if (slack.enabled && slackChannel && isMysticatTargetedSkip(skipReason)) {
+      // unsupported-action skips stay silent. Best-effort + rate-limited per PR;
+      // postMessage never throws.
+      if (
+        slack.enabled
+        && isMysticatTargetedSkip(skipReason)
+        && !shouldRateLimitSlackPost(`${data.repository.owner.login}/${data.repository.name}#${pr.number}`)
+      ) {
         await slack.postMessage({
-          channel: slackChannel,
           text: skippedStandaloneText({
             owner: data.repository.owner.login,
             repo: data.repository.name,
@@ -170,13 +175,15 @@ function WebhooksController(context) {
 
     // Post the Slack thread root BEFORE enqueue (ordering invariant: parent
     // before enqueue, never after). Best-effort - a Slack failure must never
-    // block the review, so we still enqueue. When the post fails we send
-    // slack_channel only (no thread_ts); the worker then degrades to a
-    // standalone message. When Slack is unconfigured we omit observability.
+    // block the review, so we still enqueue. On parent-post failure we send
+    // slack_channel only (no thread_ts); the worker degrades to a standalone.
+    // When Slack is disabled or rate-limited we omit observability entirely.
     let observability;
-    if (slack.enabled && slackChannel) {
+    if (
+      slack.enabled
+      && !shouldRateLimitSlackPost(`${data.repository.owner.login}/${data.repository.name}#${pr.number}`)
+    ) {
       const threadTs = await slack.postMessage({
-        channel: slackChannel,
         text: enqueuedParentText({
           owner: data.repository.owner.login,
           repo: data.repository.name,
--- a/src/support/slack/observability-client.js
+++ b/src/support/slack/observability-client.js
@@ -12,34 +12,60 @@
 
 import { WebClient } from '@slack/web-api';
 
+// Bound the Slack request so a slow/hung Slack API call can never delay webhook
+// handling: the controller awaits postMessage before enqueue, and GitHub times
+// out webhook delivery at ~10s (and redelivers). A sub-2s budget with no retries
+// keeps the worst case well under that window even on a slow Slack edge, so a
+// Slack incident cannot cause duplicate enqueues via GitHub redelivery.
+const SLACK_REQUEST_TIMEOUT_MS = 2000;
+
 /**
  * Creates a best-effort Slack client for review observability.
  *
- * Uses a dedicated chat:write-only bot token (NOT the broad elevatedSlackClient).
- * postMessage NEVER throws: a Slack failure must not block or fail a webhook.
- * It returns the posted message `ts` (string) on success, or null on any failure,
- * when unconfigured, or when no channel is given.
+ * Uses a dedicated chat:write-only bot token (NOT the broad elevatedSlackClient)
+ * so a compromised webhook environment cannot post as the broad bot in product
+ * channels; the blast radius on token compromise is limited to this one channel.
+ *
+ * postMessage NEVER throws: a Slack failure must not block or fail a webhook. It
+ * returns the posted message `ts` (string) on success, or null on any failure or
+ * when disabled (no token or no channel, or the client could not be built).
  *
  * @param {object} p
- * @param {string|undefined} p.token - bot token; absent => Slack disabled
+ * @param {string|undefined} p.token - bot token; absent => disabled
+ * @param {string|undefined} p.channel - target channel id; absent => disabled
  * @param {object} p.log - logger with .warn
  * @returns {{ postMessage: function, enabled: boolean }}
  */
-export function createObservabilitySlackClient({ token, log }) {
-  const client = token ? new WebClient(token) : null;
+export function createObservabilitySlackClient({ token, channel, log }) {
+  let client = null;
+  if (token) {
+    try {
+      client = new WebClient(token, {
+        timeout: SLACK_REQUEST_TIMEOUT_MS,
+        retryConfig: { retries: 0 },
+      });
+    } catch (e) {
+      // A constructor failure (e.g. a future token-format validation) must not
+      // crash webhook handling for every delivery — degrade to disabled.
+      log.warn('Observability Slack client init failed (non-fatal)', { error: e?.message ?? String(e) });
+      client = null;
+    }
+  }
 
-  async function postMessage({ channel, text, attachments }) {
-    if (!client || !channel) {
+  const enabled = Boolean(client && channel);
+
+  async function postMessage({ text, attachments }) {
+    if (!enabled) {
       return null;
     }
     try {
       const result = await client.chat.postMessage({ channel, text, attachments });
       return result?.ts ?? null;
     } catch (e) {
-      log.warn('Observability Slack post failed (non-fatal)', { error: e.message });
+      log.warn('Observability Slack post failed (non-fatal)', { error: e?.message ?? String(e) });
       return null;
     }
   }
 
-  return { postMessage, enabled: Boolean(client) };
+  return { postMessage, enabled };
 }
--- a/src/support/slack/observability-client.js
+++ b/src/support/slack/observability-client.js
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2025 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { WebClient } from '@slack/web-api';
+
+/**
+ * Creates a best-effort Slack client for review observability.
+ *
+ * Uses a dedicated chat:write-only bot token (NOT the broad elevatedSlackClient).
+ * postMessage NEVER throws: a Slack failure must not block or fail a webhook.
+ * It returns the posted message `ts` (string) on success, or null on any failure,
+ * when unconfigured, or when no channel is given.
+ *
+ * @param {object} p
+ * @param {string|undefined} p.token - bot token; absent => Slack disabled
+ * @param {object} p.log - logger with .warn
+ * @returns {{ postMessage: function, enabled: boolean }}
+ */
+export function createObservabilitySlackClient({ token, log }) {
+  const client = token ? new WebClient(token) : null;
+
+  async function postMessage({ channel, text, attachments }) {
+    if (!client || !channel) {
+      return null;
+    }
+    try {
+      const result = await client.chat.postMessage({ channel, text, attachments });
+      return result?.ts ?? null;
+    } catch (e) {
+      log.warn('Observability Slack post failed (non-fatal)', { error: e.message });
+      return null;
+    }
+  }
+
+  return { postMessage, enabled: Boolean(client) };
+}
--- a/src/support/slack/observability-messages.js
+++ b/src/support/slack/observability-messages.js
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2025 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/**
+ * Pure mrkdwn formatters for the web tier's Slack observability messages.
+ * The web tier only ever posts top-level messages (thread roots + standalones);
+ * the worker formats and posts the threaded replies in mysticat-github-service.
+ */
+
+/**
+ * Thread-root message for an enqueued review.
+ * @param {object} p
+ * @param {string} p.owner
+ * @param {string} p.repo
+ * @param {number|string} p.prNumber
+ * @param {string} p.action - GitHub event action (e.g. review_requested)
+ * @param {string} p.jobType - mapped job type (e.g. pr-review)
+ * @returns {string} mrkdwn text
+ */
+export function enqueuedParentText({
+  owner, repo, prNumber, action, jobType,
+}) {
+  return `:inbox_tray: *Review enqueued*  \`${owner}/${repo}\` #${prNumber}\n${action} → ${jobType}`;
+}
+
+/**
+ * Standalone message for a Mysticat-targeted skip (draft / bot / non-default branch).
+ * @param {object} p
+ * @param {string} p.owner
+ * @param {string} p.repo
+ * @param {number|string} p.prNumber
+ * @param {string} p.reason - skip reason from getSkipReason
+ * @returns {string} mrkdwn text
+ */
+export function skippedStandaloneText({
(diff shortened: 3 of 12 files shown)


src/support/slack/observability-client.js
```diff
@@ -0,0 +1,45 @@
...
+ * when unconfigured, or when no channel is given.
+ *
+ * @param {object} p
+ * @param {string|undefined} p.token - bot token; absent => Slack disabled
+ * @param {object} p.log - logger with .warn
+ * @returns {{ postMessage: function, enabled: boolean }}
+ */
+export function createObservabilitySlackClient({ token, log }) {
+  const client = token ? new WebClient(token) : null;
+
```
szergen, line 56: **Critical:** `new WebClient(token)` has no timeout and no retry cap. Combined with `await slack.postMessage(...)` running before `sqs.sendMessage` in the controller, a Slack incident or rate-limit hangs past GitHub's ~10s webhook delivery window. GitHub redelivers, and the same review enqueues multiple times while one retry eventually also posts a second parent message once Slack recovers.

This breaks the "best-effort, never blocks the webhook" invariant the PR description rests on.

Fix:
```js
const client = token
  ? new WebClient(token, { timeout: 2000, retryConfig: { retries: 0 } })
  : null;
```
A sub-2-second budget keeps webhook latency well under GitHub's window even on a slow Slack edge. Add a controller-level `Promise.race` against a wall-clock budget if you want belt-and-braces.

src/support/slack/observability-client.js
```diff
@@ -0,0 +1,45 @@
...
+  async function postMessage({ channel, text, attachments }) {
+    if (!client || !channel) {
+      return null;
+    }
+    try {
+      const result = await client.chat.postMessage({ channel, text, attachments });
+      return result?.ts ?? null;
+    } catch (e) {
+      log.warn('Observability Slack post failed (non-fatal)', { error: e.message });
```
szergen: **Important** (confirming Copilot's prior finding): `error: e.message` will throw `TypeError` if a non-Error is rejected (e.g., `Promise.reject(undefined)` or a string). The throw escapes `postMessage`, escapes the controller's `await`, and `errorHandler` 500s the webhook - inverting the documented "never throws" contract from "warn + degrade" to "5xx + lose the webhook + GitHub redelivery".

```js
error: e?.message ?? String(e)
```

Add a test using `postMessageStub.rejects(undefined)` asserting `ts === null` and `log.warn.calledOnce`. The contract becomes self-enforcing.

src/support/slack/observability-client.js
szergen, line 56: **Minor:** the JSDoc above says "NOT the broad elevatedSlackClient" but does not say why. Six months from now the next engineer will wonder why there are two Slack clients and may try to consolidate them. One-line addition: "Separate token so a compromised webhook env cannot post as the broad bot in product channels."

src/controllers/webhooks.js
```diff
@@ -65,6 +67,11 @@ function getWorkspaceRepos(env, log) {
 function WebhooksController(context) {
   const { sqs, log, env } = context;
   const workspaceRepos = getWorkspaceRepos(env, log);
+  const slackChannel = env.MYSTICAT_OBSERVABILITY_SLACK_CHANNEL;
+  const slack = createObservabilitySlackClient({
+    token: env.MYSTICAT_OBSERVABILITY_SLACK_TOKEN,
+    log,
+  });
```
szergen, line 76: **Important:** `new WebClient(token)` runs at controller construction, outside any try/catch (inside `createObservabilitySlackClient`). Any future change in `@slack/web-api` that validates token format at construction (or a malformed token containing whitespace, etc.) would crash Lambda init for every webhook delivery, not just the ones that try to post.

Wrap the constructor inside `createObservabilitySlackClient` in try/catch and degrade to `client = null` on throw. Same return shape, `enabled: false`.

src/controllers/webhooks.js
```diff
@@ -143,9 +150,46 @@ function WebhooksController(context) {
         repo: data.repository.name,
         prNumber: pr.number,
       });
+      // Post a standalone Slack note only when Mysticat WAS the requested
+      // reviewer (draft / bot / non-default branch). Foreign-reviewer and
+      // unsupported-action skips stay silent to avoid channel flooding.
+      // Best-effort: postMessage never throws.
```
szergen: **Important:** the Slack post fires for any installation that reaches `processGitHubWebhook`, with no allow-list against `workspaceRepos`. If the Mysticat GitHub App is ever installed on a private third-party repo (which `MYSTICAT_WORKSPACE_REPOS` does NOT enforce in the controller), the `owner/repo` and PR number get broadcast to your Slack channel anyway.

For an internal observability channel this is most likely fine, but it widens the disclosure surface beyond the workspace repos. Fix: gate the Slack posts on `workspaceRepos.includes(`${owner}/${repo}`)`, or push the workspace-repo check up before any Slack interaction.

szergen: **Minor:** the `slack.enabled && slackChannel` guard is duplicated at two call sites here and on line 185. Worse, `createObservabilitySlackClient.postMessage` already enforces `!channel -> return null` internally, so the outer `slackChannel` check is structurally redundant with the inner one.

Consider pushing the channel into the factory: `createObservabilitySlackClient({ token, channel, log })` -> expose `enabled = Boolean(client && channel)` and drop `channel` from each `postMessage` call. Single guard at each site, single source of truth.

src/utils/github-trigger-rules.js
```diff
@@ -83,3 +83,23 @@ export function getSkipReason(data, action, env) {
...
+ * it returns AFTER the reviewer check passes. The drift-guard tests in
+ * test/utils/github-trigger-rules.test.js fail if the two diverge.
+ *
+ * @param {string} reason - The skip reason string returned by getSkipReason
+ * @returns {boolean} true if a standalone Slack note should be posted
+ */
+export function isMysticatTargetedSkip(reason) {
+  return reason === 'draft PR'
+    || reason === 'bot sender'
+    || reason.startsWith('non-default branch:');
```
szergen, line 107: **Important:** two drift-guard gaps:

1. The drift-guard `describe('stays in lockstep with getSkipReason')` covers draft, bot, non-default-branch, and foreign-reviewer cases, but does NOT round-trip the `unsupported action: <action>` or `auto-trigger not yet supported: <action>` reasons. Those are only tested via literal strings fed directly to `isMysticatTargetedSkip`. If someone reworded those in `getSkipReason`, the literal tests still pass and the classifier silently misclassifies. Add two more drift-guard cases that build payloads triggering each negative reason.

2. More fundamentally: the classifier is fragile against ADDED reasons. The day someone adds a new rule in `getSkipReason` after the reviewer check passes (e.g., "PR closed", "fork PR not supported"), the new reason will NOT be in `isMysticatTargetedSkip`'s allow-list and the channel goes silent for cases operators expect to see. Consider returning a structured value from `getSkipReason` (`{ reason, targeted: boolean }`) so the rule author cannot forget to update both - the classifier disappears and the drift-guard test file disappears with it. Worth the 30-line refactor before this pattern is copied to the worker tier.

**Minor (related):** `reason.startsWith(...)` will throw on `null`/`undefined`. The controller currently guards with `if (skipReason)` before calling, so this is purely defensive, but the function is exported. Add `if (!reason) return false;` at the top.

src/support/slack/observability-messages.js
```diff
@@ -0,0 +1,48 @@
...
+ * @param {string} p.owner
+ * @param {string} p.repo
+ * @param {number|string} p.prNumber
+ * @param {string} p.reason - skip reason from getSkipReason
+ * @returns {string} mrkdwn text
+ */
+export function skippedStandaloneText({
+  owner, repo, prNumber, reason,
+}) {
+  return `:fast_forward: *Skipped*  \`${owner}/${repo}\` #${prNumber} - ${reason}`;
```
szergen: **Minor:** `pr.base.ref` flows into `getSkipReason` as `non-default branch: ${ref}` and then into `skippedStandaloneText` raw. Git refs allow `<`, `>`, `&`, `!`, backtick (`*` is forbidden). A branch literally named `<!here>` (legal in Git) would render as a `@here` ping in Slack.

Severity is bounded because the target branch must exist on the target repo (so an external attacker would need maintainer access on a repo where Mysticat is installed), but cheap to defuse: wrap the `reason` interpolation in a code span inside `skippedStandaloneText`, or strip `<`, `>`, backtick from `pr.base.ref` before interpolation in `getSkipReason`.

Separately, the two-space gap after `*Skipped*` (and `*Review enqueued*` on line 34) looks like a typo on review - Slack collapses runs of spaces in mrkdwn anyway.

szergen (changes requested):
Hey @solaris007,

Thanks for the careful scoping and the drift-guard hygiene. The ordering invariant ("parent before enqueue") is correctly enforced and tested, the partial `observability` envelope shape is forward-compatible, and the dedicated `chat:write`-only token is the right least-privilege posture. Codecov reports modified lines are covered and the test scaffolding (`esmock` + `calledBefore`) is the right shape for what you are pinning.

The blocker before merge is the unbounded Slack call interacting with GitHub's webhook retry semantics. Two of the four issues here Copilot already raised on the earlier review pass; I am agreeing with them and adding scope/severity rather than restating.

## Strengths

- `src/controllers/webhooks.js:185-195` - ordering invariant correctly enforced (parent post awaited before `sqs.sendMessage`), and the controller test asserts `postMessage.calledBefore(mockSqs.sendMessage)`. That is the kind of test that actually catches regressions.
- `src/controllers/webhooks.js:191-194` - the partial `observability` shape (`slack_channel` without `slack_thread_ts` when the parent post fails, omitted entirely when the channel is unconfigured) is the right forward-compatible contract. A worker that does not yet understand `observability` ignores an unknown field; nothing can DLQ.
- `src/utils/github-trigger-rules.js:99-105` - co-locating `isMysticatTargetedSkip` with `getSkipReason` and adding round-trip drift-guard tests is the correct hygiene for a string-coupled contract.
- `src/support/slack/observability-client.js` - the dedicated `chat:write`-only bot token, separate from `elevatedSlackClient`, is the right least-privilege boundary. Blast radius on token compromise is limited to posting messages.
- Selective standalone-skip posting (draft / bot / non-default branch only, foreign reviewer silent) is the right default - foreign-reviewer channel flooding would have been the obvious failure mode.

## Issues

### Critical (Must Fix)

1. `src/support/slack/observability-client.js:30` and `src/controllers/webhooks.js:185-187` - `new WebClient(token)` has no timeout and no retry cap, yet `postMessage` is awaited synchronously before `sqs.sendMessage`. `@slack/web-api` defaults to retry-on-rate-limit and no upper-bound request timeout. GitHub's webhook delivery times out at ~10s and redelivers on timeout. Under a Slack incident or rate-limit, the controller hangs past GitHub's window, GitHub retries, and the same review gets enqueued multiple times - while one of the retries eventually posts a second parent message to Slack once it recovers. This breaks the "best-effort, never blocks the webhook" invariant the entire safety argument in the PR description rests on. Fix:

   ```js
   const client = token
     ? new WebClient(token, { timeout: 2000, retryConfig: { retries: 0 } })
     : null;
   ```

   Sub-2-second budget keeps webhook latency well under GitHub's window even on a slow Slack edge. Add a controller-level `Promise.race` against a wall-clock budget if you want belt-and-braces.

### Important (Should Fix)

2. `src/support/slack/observability-client.js:39` - confirming Copilot's prior finding: `error: e.message` will throw `TypeError` if a non-Error is rejected (e.g., `Promise.reject(undefined)` or a string). The throw escapes `postMessage`, escapes the `await`, and the controller's `errorHandler` 500s the webhook - inverting the "never throws" contract from "warn + degrade" to "5xx + lose the webhook + GitHub redelivery". Fix:

   ```js
   error: e?.message ?? String(e)
   ```

   And add a test using `postMessageStub.rejects(undefined)` asserting `ts === null` and `log.warn.calledOnce`. The contract becomes self-enforcing.

3. `src/controllers/webhooks.js:71-74` - `new WebClient(token)` runs at controller construction, outside any try/catch. Any future change in `@slack/web-api` that validates token format at construction (or a malformed token containing whitespace, etc.) would crash Lambda init for every webhook delivery, not just the ones that try to post. Wrap the constructor call inside `createObservabilitySlackClient` in try/catch and degrade to `client = null` on throw. Same return shape, `enabled: false`.

4. `src/controllers/webhooks.js:71-74, 156, 185` - the Slack post fires for any installation that reaches `processGitHubWebhook`, with no allow-list against `workspaceRepos`. If the Mysticat GitHub App is ever installed on a private third-party repo (which `MYSTICAT_WORKSPACE_REPOS` does NOT enforce in the controller), the `owner/repo` and PR number get broadcast to your Slack channel anyway. For an internal observability channel this is most likely fine, but it widens the disclosure surface beyond the workspace repos. Fix: gate the Slack posts on `workspaceRepos.includes(`${owner}/${repo}`)`, or push the workspace-repo check up before any Slack interaction.

5. New env vars `MYSTICAT_OBSERVABILITY_SLACK_TOKEN` and `MYSTICAT_OBSERVABILITY_SLACK_CHANNEL` are not added to `docs/specs/2026-04-22-github-webhook-handler.md` (confirming Copilot's prior finding). The 944-line plan doc is not the right home - an SRE picking up an incident reads the handler spec. Add both vars with their Vault source and the "omit channel to disable observability entirely" semantic. A `chat:write` token is still a long-lived bearer secret that should be on the rotation list.

6. `src/utils/github-trigger-rules.js:103-105` and `test/utils/github-trigger-rules.test.js` (drift guards) - two gaps:
   - The drift-guard `describe('stays in lockstep with getSkipReason')` covers draft, bot, non-default-branch, and foreign-reviewer cases, but does NOT round-trip the `unsupported action: <action>` or `auto-trigger not yet supported: <action>` reasons. Those are only tested via literal strings fed directly to `isMysticatTargetedSkip`. If someone reworded those in `getSkipReason`, the literal tests still pass and the classifier silently misclassifies. Add two more drift-guard cases that build payloads triggering each negative reason.
   - More fundamentally: the classifier is fragile against ADDED reasons. The day someone adds a new rule in `getSkipReason` after the reviewer check passes (e.g., "PR closed", "fork PR not supported"), the new reason will NOT be in `isMysticatTargetedSkip`'s allow-list and the channel goes silent for cases operators expect to see. Consider returning a structured value from `getSkipReason` (`{ reason, targeted: boolean }`) so the rule author cannot forget to update both - the classifier disappears and the drift-guard test file disappears with it. Worth the 30-line refactor before this pattern is copied to the worker tier.

### Minor (Nice to Have)

7. `src/utils/github-trigger-rules.js:104` - `reason.startsWith(...)` throws on `null`/`undefined`. The controller currently guards with `if (skipReason)` before calling, so this is purely defensive, but the function is exported. Add `if (!reason) return false;` at the top.

8. `src/controllers/webhooks.js:156, 185` - the `slack.enabled && slackChannel` guard is duplicated at two call sites. Worse, `createObservabilitySlackClient.postMessage` already enforces `!channel -> return null` internally, so the outer `slackChannel` check is structurally redundant with the inner one. Push the channel into the factory: `createObservabilitySlackClient({ token, channel, log })` -> expose `enabled = Boolean(client && channel)` and drop `channel` from each `postMessage` call. Single guard at each site, single source of truth.

9. `src/support/slack/observability-messages.js:34` - `pr.base.ref` flows into `getSkipReason` as `non-default branch: ${ref}` and then into `skippedStandaloneText` raw. Git refs allow `<`, `>`, `&`, `!`, backtick (`*` is forbidden). A branch literally named `<!here>` (legal in Git) would render as a `@here` ping in Slack. Severity is bounded because the target branch must exist on the target repo (so an external attacker would need maintainer access on a repo where Mysticat is installed), but cheap to defuse: wrap the `reason` interpolation in a code span inside `skippedStandaloneText`, or strip `<`, `>`, backtick from `pr.base.ref` before interpolation in `getSkipReason`.

10. `src/support/slack/observability-messages.js:34, 47` - the two-space gap after `*Review enqueued*` / `*Skipped*` looks like a typo on review (Slack collapses runs of spaces anyway). Drop to a single space or add a comment explaining the visual intent.

11. `docs/plans/2026-05-21-slack-observability-web-tier.md:9` - confirming Copilot's prior finding: plan says Node.js 22.x, repo `package.json` requires Node >=24. Trivial doc fix in passing.

12. `src/support/slack/observability-client.js:17-23` - the JSDoc says "NOT the broad elevatedSlackClient" but does not explain why. Six months from now the next engineer will wonder why there are two Slack clients. One-line addition: "Separate token so a compromised webhook env cannot post as the broad bot in product channels."

## Recommendations

- Document the `observability` envelope as a versioned cross-repo contract. It spans three repos (this PR + spacecat-infrastructure#536 + mysticat-github-service). At minimum, freeze the JSON shape in the handler spec: `slack_channel: string`, `slack_thread_ts: string` (Slack `ts` is a string like `"1700000000.123456"`, not a number - make sure the worker treats it as opaque). Optional `version: 1` field is cheap insurance.
- Add a structured `log.info('Slack thread root post degraded', { ... })` in the controller when `threadTs` is falsy. The only signal a Slack failure happened today is a `log.warn` inside `observability-client.js`; the worker tier sees `observability` without `slack_thread_ts` and has to infer.
- Rate-limit the Slack post per-repo at the controller (in-memory `Map<owner/repo, lastPostedAt>` with a 10s gap, per-container, best-effort). Even with `isMysticatTargetedSkip`, a maintainer of a repo where Mysticat is installed could open and close drafts in a loop to spam the channel.
- Consider an alternative shape that does fire-and-forget for the parent post: post after enqueue, write the `ts` to a small side-channel keyed by `delivery_id`. Heavier, but the design that ages well if observability scope grows. For now the bounded-timeout fix is fine; flag as v2 if Slack latency ever shows up as an SLO concern.

## Assessment

Reasoning: The architectural shape is correct - boundary, ordering invariant, forward-compatible envelope, and silent-disable kill-switch are all well-chosen. The blocking concern is the unbounded Slack call interacting with GitHub's webhook retry semantics (Critical 1); without a bounded timeout and a retry cap, a Slack incident becomes a webhook duplication storm. Fix Critical 1 and Important 2-5, document the env vars, and this is mergeable. The structured-return refactor for `getSkipReason` (Important 6) is worth doing before this pattern is copied to the worker tier.

## Next Steps

1. Address the Critical Slack timeout and retry-cap issue first.
2. Tighten the `postMessage` "never throws" contract (Important 2) and harden controller construction (Important 3).
3. Gate Slack posts on the workspace-repo allow-list (Important 4) and document the new env vars in the handler spec (Important 5).
4. Close the drift-guard gaps or refactor `getSkipReason` to a structured return (Important 6).
5. The Minor items are polish - safe to bundle into the same fixup commits.

solaris007:
Hey @szergen, @Copilot — addressed in `c666f289` (+ `c622e135` for a coverage gap), pushed (main merged in too). Disposition by item:

**Critical**
- **#1 unbounded Slack call** — fixed. `new WebClient(token, { timeout: 2000, retryConfig: { retries: 0 } })`. Sub-2s, zero retries, so a Slack incident can't exceed GitHub's webhook window or cause duplicate enqueues. Test asserts the constructor options.

**Important**
- **#2 `e.message` on non-Error** — fixed. `e?.message ?? String(e)` in both catches; test with `postMessageStub.rejects(undefined)` asserting `null` + one `warn` (the contract is now self-enforcing).
- **#3 WebClient constructed outside try/catch** — fixed. Construction wrapped in try/catch inside the factory; degrades to `client = null` / `enabled: false`. Degrade path covered by a test.
- **#4 workspaceRepos allow-list** — pushing back. `workspaceRepos` is the *doc/context* repo list (`mysticat-architecture`, etc.) cloned into the worker, used only for `workspace_repos` in the payload (webhooks.js:203). The repo being reviewed is never in that list, so `workspaceRepos.includes(owner/repo)` would silence Slack for **every** real review. The underlying concern — no reviewable-repo allow-list — is a pre-existing property of the trigger (`getSkipReason` gates on reviewer/draft/bot/branch, not repo), not introduced here, and the post only mirrors repos already being reviewed (whose reviews are posted publicly on the PR). A real allow-list belongs at the trigger level (gating reviews, not just Slack) — happy to take that as a separate change. Flagging rather than silently dropping.
- **#5 env vars in handler spec** — fixed. Both added to the Environment Variables table with Vault source + "omit channel to disable" semantic; noted the token is on the rotation list.
- **#6a drift-guard gaps** — fixed. Added round-trip drift-guards for `unsupported action` and `auto-trigger not yet supported`.
- **#6b structured `getSkipReason`** — deferring. #6a covers reworded reasons; an *added* reason defaulting to silent is the safe default (a new skip type won't flood). The worker tier doesn't use this classifier, so the "before it's copied" risk isn't live yet. Will revisit when an operator-visible skip reason is added — open to it as a follow-up.

**Minor**
- **#7 `startsWith` on null** — fixed (`if (!reason) return false;`).
- **#8 channel into factory** — done. `createObservabilitySlackClient({ token, channel, log })`, `enabled = Boolean(client && channel)`, channel dropped from `postMessage`. Single guard per site; matches the worker notifier's shape.
- **#9 mrkdwn injection (`<!here>` branch)** — fixed. `escapeSlack` (`&`,`<`,`>`) on dynamic fields outside code spans (owner/repo stay inside backticks, where Slack renders literally). Test with a `<!here>` branch reason.
- **#10 double space** — fixed. **#11 plan Node 22→>=24** — fixed. **#12 JSDoc "why two clients"** — fixed (compromised-env rationale).

**Recommendations**
- **Rate-limit** — implemented. Per-PR (`owner/repo#pr`), in-memory/per-container, best-effort (per-PR over per-repo to avoid collateral across distinct PRs); when rate-limited we omit `observability` so the worker stays silent too. Unit-tested + a controller wiring test.
- **Versioned envelope / degraded-thread log** — deferring (the nested-object shape already gives forward-compat; the client warns on failure). **Fire-and-forget v2** — agreed, tracked as v2 if Slack latency becomes an SLO concern.

`thread_ts` is treated as an opaque string end to end. 10323 unit tests green, lint clean.

src/controllers/webhooks.js
solaris007: Pushing back on this one (full context in the top-level comment). `workspaceRepos` here is the doc/context-repo list (`mysticat-architecture`, `mysticat-ai-native-guidelines`, `aem-sites-architecture`) cloned into the worker — it's used only for `workspace_repos` in the payload at line 203, never as a trigger gate. The repo under review is never a member of that list, so gating on `workspaceRepos.includes(`${owner}/${repo}`)` would suppress Slack for every real review.

The "any installation reaches the handler" concern is a pre-existing trigger-level property (`getSkipReason` doesn't gate on repo at all), not introduced by this PR, and the post only mirrors repos already being reviewed publicly on the PR — so it doesn't widen the trigger surface. If we want a reviewable-repo allow-list, it should live at the trigger (gating the reviews themselves, not just the Slack note); happy to do that as a separate change.

MysticatBot-Dev (approved):
Based on my thorough analysis of the current diff, prior review findings, and the author's comprehensive fix commits, I can produce this re-review directly.

Re-reviewing PR 2457 (adobe/spacecat-api-service) - author addressed all prior Critical/Important findings across fix commits.

Hey @solaris007,

## Strengths

- Previously flagged issues now addressed: all Critical and Important items from the prior review are resolved in the fix commits. The unbounded Slack call (Critical 1) now has a 2s timeout with zero retries (`src/support/slack/observability-client.js:20,48-50`). The `e.message` on non-Error (Important 2) is fixed with `e?.message ?? String(e)` in both catch paths. Constructor failure hardening (Important 3) wraps `new WebClient` in try/catch. Env vars documented in handler spec (Important 5). Drift-guard gaps closed for `unsupported action` and `auto-trigger` (Important 6a).
- `src/support/slack/observability-client.js:43-67` - the factory's layered defense is well-structured: construction try/catch degrades to disabled, `enabled = Boolean(client && channel)` as single source of truth, `postMessage` checks `enabled` before any network call, and the catch uses safe stringification. The "never throws" contract is now self-enforcing at every layer.
- `src/support/slack/observability-messages.js:27-31` - `escapeSlack` for `&`, `<`, `>` on dynamic fields outside code spans defuses the mrkdwn injection vector without over-sanitizing. Applied only where needed (reason text is user-influenced; owner/repo sit inside backtick code spans where Slack renders literally).
- `src/support/slack/observability-rate-limit.js` - implementing the rate limiter as a recommendation from the prior review, per-PR keyed with a clear memory bound (`MAX_TRACKED_KEYS` with full clear on overflow). The simplicity-over-precision tradeoff is the right call for a best-effort soft limiter.
- `test/controllers/webhooks.test.js:340-353` - the `postMessage.calledBefore(mockSqs.sendMessage)` assertion pins the ordering invariant (parent post before enqueue). This is the kind of test that catches real regressions.
- The channel-into-factory refactor (`createObservabilitySlackClient({ token, channel, log })`) eliminates the duplicated `slackChannel` guard at call sites, matching the pattern the prior review suggested.
- The pushback on Important 4 (workspaceRepos allow-list) is sound. `workspaceRepos` is the doc-context list for the worker, not a reviewable-repo allow-list. Gating Slack on that list would silence Slack for every real review.

## Issues

### Minor (Nice to Have)

1. `src/support/slack/observability-rate-limit.js:39-41` - when `lastPostAt.size >= MAX_TRACKED_KEYS`, the full `clear()` discards rate-limit state for all active PRs at once, not just the overflowing key. Under a burst scenario (many distinct PRs in a short window), this briefly allows re-posts for PRs that were correctly suppressed. The trade-off is documented and acceptable for a soft limiter, but worth noting that the 1000-key cap is generous enough (each key is a short string like `owner/repo#123`) that the clear should be extremely rare in practice. No action needed unless observed in production.

2. `docs/plans/2026-05-21-slack-observability-web-tier.md` - the committed plan doc is 944 lines. Plan docs served their purpose during implementation and now exist primarily as git history context. Consider whether this should live in the repo long-term or be trimmed/removed post-merge so it doesn't age alongside the code it describes.

## Assessment

Ready to merge? **Yes**

Reasoning: All Critical and Important issues from the prior review are addressed with clean, well-tested fixes. The code establishes the ordering invariant (parent before enqueue), maintains the "never throws" contract at every layer, bounds the Slack call to 2s with no retries, and adds rate-limiting as a defense against channel flooding. The pushback on the workspaceRepos allow-list is justified. The architecture is forward-compatible (worker can ignore the `observability` field until repo 3 lands). Note: CI codecov/patch check shows 2 uncovered lines in `observability-rate-limit.js` and integration tests (`it-postgres`) are still pending - confirm those pass before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 7m 34s | <b>Cost:</b> $1.83 | <b>Commit:</b> <code>c622e13533e2411da06c2ea15d271f61f41cae4e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2460, opened 2026-05-21
feat: ai vis gap methods and filters

dmitriigulkin-semrush:

## Related Issues

Thanks for contributing!


calvarezg (commented):
## Review (updated)

### Medium

**Duplicate `buildDimensionFilterQl` logic**

`buildBrandTopicsDimensionFilterQl` (`brand-topics.js`) and `buildGapTopicsDimensionFilterQl` (`gap-topics.js`) are byte-for-byte identical:

```js
const escaped = q.replace(/\\/g, "\\\\").replace(/"/g, "\\\"");
return `topic CONTAINS "${escaped}"`;
```

The PR already extracted `buildRangeExpr`, `isValidVolume`, and `isValidVisibility` to `grpc-utils.js` — this should follow the same pattern.

---

### Minor

**1. No validation on `gapKinds` parameter**

The `gapKinds` query param is split on `,` without validating individual values. Invalid enum strings will fail inside `fromJson` and return a 400 with a raw protobuf error message, which is less informative than an explicit check. Inconsistent with how `volume`/`visibility` are validated.

**2. No bounds-ordering check on `volumeFrom`/`volumeTo` (and `visibilityFrom`/`visibilityTo`)**

Each bound is individually validated as a non-negative integer, but `volumeFrom > volumeTo` (e.g. `volumeFrom=100&volumeTo=10`) is accepted and forwarded as `volume >= 100 AND volume <= 10`, which silently produces an empty result set. Worth either rejecting with `400 invalid_range` or normalizing.

---

### What looks good

- Refactoring `PROTO_FROM_JSON` / `PROTO_TO_JSON` out of individual handlers into `grpc-utils.js` is clean and reduces drift risk.
- `kind` vs `kinds` field naming differs across `GapTopicsRequest` (singular `kind`) and `GapPromptsRequest` (plural `kinds`) — this is intentional and matches the proto schemas (the latter deprecates a scalar `kind` field).
- Not forwarding `gapKinds` to the totals requests is correct — the totals proto schemas have no kind field, and the response is already broken down per kind via `TotalByKind[]`.
- Error handling pattern (proto parse error → 400, gRPC error → mapped response, unknown → rethrow) is consistent across all handlers.
- Route registration, controller wiring, and required-capabilities entries are all updated together consistently.
- Tests correctly reflect the new handler count (25 → 27) and stub the new modules.

---

_Note: previous comments about a `kind`/`kinds` typo and `gapKinds` missing from totals requests were incorrect — both are intentional. Retracted._

dmitriigulkin-semrush:
@calvarezg , hey! thanks

> Medium
> Duplicate buildDimensionFilterQl logic
> 
> buildBrandTopicsDimensionFilterQl (brand-topics.js) and buildGapTopicsDimensionFilterQl (gap-topics.js) are byte-for-byte identical:
> 
> const escaped = q.replace(/\\/g, "\\\\").replace(/"/g, "\\\"");
> return `topic CONTAINS "${escaped}"`;
> The PR already extracted buildRangeExpr, isValidVolume, and isValidVisibility to grpc-utils.js — this should follow the same pattern.

i extracted escaping behaviour, but i really wouldn't like to extract building the expression itself, there are no guarantee that all methods will follow the same pattern with column naming, e.g. in some methods in might be `topic_name` instead of `topic`. from architectural standpoint it also make more sense, since every method have its own set of parameters it supports, as well as naming and validation rules. i guess it'd be nice to have method/client agnostic FilterQL builder, will be happy to bring an implementation over in coming PRs. wdyt? 

---

> 1. No validation on gapKinds parameter
> 
> The gapKinds query param is split on , without validating individual values. Invalid enum strings will fail inside fromJson and return a 400 with a raw protobuf error message, which is less informative than an explicit check. Inconsistent with how volume/visibility are validated.
> 
> 2. No bounds-ordering check on volumeFrom/volumeTo (and visibilityFrom/visibilityTo)
> 
> Each bound is individually validated as a non-negative integer, but volumeFrom > volumeTo (e.g. volumeFrom=100&volumeTo=10) is accepted and forwarded as volume >= 100 AND volume <= 10, which silently produces an empty result set. Worth either rejecting with 400 invalid_range or normalizing.

so my point on validation of user' input when you work with grpc is pretty straightforward 😂  
we should delegate it to:
1. protobuf.validate(not implemented yet but coming soon)
2. while building the actual message(already in place)
3. upstream grpc-service(already in place)

this way you drastically reduce the surface you need to read, understand, support and write tests for
tho, i'm not very pushy on it, if you'd like to add additional validation rules be happy to do it 🤝

calvarezg (approved):

PR #2466, opened 2026-05-22
fix(semrush): re-land AIO proxy with bundling fix + CI bundle gate + CLAUDE.md guidance

rainer-friederich:
## Re-land of [#2456](https://github.com/adobe/spacecat-api-service/pull/2456) with bundling fix + new pre-merge gate + codified guidance

Tracks [SITES-45260](https://jira.corp.adobe.com/browse/SITES-45260). The semrush AIO proxy merged as commit `882dbab5` (v1.508.0) on 2026-05-21 and took the api-service Lambda down — every invocation failed with `TypeError: main2 is not a function at lambdaAdapter`. Reverted at `873eddb9`. This PR brings the feature back with three layered defences against the same class of failure recurring.

## Root cause (recap)

`src/support/semrush/handlers/projects.js` was reading `data/locations.json` at module load via `readFileSync(import.meta.url)`. The JSON file was not in `package.json` `hlx.static`, so `helix-deploy` never copied it into the Lambda zip. Cold start hit `ENOENT … dist/data/locations.json`, the module's export went undefined, and the deploy wrapper's reference (`main2`) crashed on every invocation.

Tests passed because they import from source, where `import.meta.url` resolves to the original source path and the JSON lives alongside the handler. The failure only manifests in the bundled artifact.

### 1. Bundling fix (`a4b198fa`)

`src/support/semrush/data/locations.json` -> `src/support/semrush/data/locations.js` (`export const LOCATIONS = { ... }`). `handlers/projects.js` imports it via normal ESM resolution. The bundler picks it up automatically — no `hlx.static` registry to maintain, no `import.meta.url` path arithmetic, no `fs`/`path`/`url` imports in the handler.

This is fix option 1 from the SITES-45260 ticket (the recommended one). Diff is a clean swap; data values are unchanged.

### 2. Pre-merge CI gate (`f1cde0db`)

The mysticat-ci reusable workflow runs lint + test + coverage only — it **never** executes `npm run build` (`hedy -v --test-bundle`). That's the gap that let the broken bundle through.

Add a repo-local `bundle-build` job to `.github/workflows/ci.yaml` that runs `npm run build` on every push / PR. `hedy --test-bundle` bundles, zips, **and invokes `lambda()` against a synthetic healthcheck event**, exiting non-zero on any non-2xx response. Verified locally:

- Healthy bundle (this PR's tip): exit 0, healthcheck returns 200
- Broken bundle (reproduction with `readFileSync` restored): exit 1, "Validation failed: 500" with `x-error: ENOENT … data/locations.json`

This is the load-bearing catch for SITES-45260's class of failure. It addresses acceptance criteria 2 + 4 in one step — a separate `typeof main === 'function'` smoke check would have been a strict subset of what `hedy --test-bundle` already does, so it was dropped.

Until adobe/mysticat-ci's reusable build action grows the equivalent step, this repo-local job is the gate. Worth lifting upstream in a follow-up so every spacecat service inherits it.

### 3. Lambda bundle constraints in CLAUDE.md (`040ed590`)

The CI gate catches the failure, but the right place to prevent the next attempt is in the guidance an AI agent (or new contributor) reads first. New `## Lambda Bundle Constraints` section in `CLAUDE.md`, placed right after the build/deploy commands so it's hard to miss when working on anything that touches `src/`. Three short rules:

1. **Do NOT** use `readFileSync(import.meta.url, ...)` or any sibling-file reads at module load — the bundled artifact does not preserve source-relative paths.
2. **Prefer JS-module imports for static data** — inline JSON / locale data / lookup tables as `export const FOO = { ... }` in a `.js` file. See `src/support/semrush/data/locations.js`.
3. **If a non-JS asset is unavoidable**, declare its repo-relative path in `package.json` `hlx.static` AND read it from the Lambda task root, not from `import.meta.url`.

Plus a pointer to the `bundle-build` CI gate and the SITES-45260 post-mortem.

## Verification

- Local `npm run build` -> exit 0, healthcheck OK
- Local reproduction with the `readFileSync` re-introduced -> exit 1 (confirmed gate works)
- 178 semrush / contract / workspace-resolver tests passing
- `git diff e3618d92..HEAD --stat` is small — only the four files this PR is meant to touch (locations.js, projects.js, ci.yaml, CLAUDE.md). The reapply commit `e3618d92` is exactly the diff of the original `882dbab5`.

## Cross-repo notes

- `mysticat-data-service` migration `20260528000000_brand_to_semrush_projects.sql`: already merged + deployed pre-revert. Schema is in place.
- `@adobe/spacecat-shared-data-access` `3.68.0`: shipped, pinned by this PR.
- Operator runbook for incident mitigation: [adobe/spacecat-api-service#2462](https://github.com/adobe/spacecat-api-service/pull/2462) (ramboz).
- Follow-up: [adobe/spacecat-api-service#2467](https://github.com/adobe/spacecat-api-service/pull/2467) makes `SEMRUSH_PROJECTS_BASE_URL` strict / Vault-backed (no source default).

## Follow-ups (out of scope)

- Lift the `bundle-build` job into `adobe/mysticat-ci`'s reusable workflow so every spacecat/mysticat service inherits the gate.
- Phase 7 (dev smoke tests against `https://spacecat.experiencecloud.live/api/ci/`) — runs after this lands and deploys.


aliciadriani:
Clean re-land. Every issue in the #2456 review is addressed, the root cause is properly fixed (not papered over), and the controller has been materially strengthened with a brand-org binding check in authorize() that wasn't in the original. Quality is high.

 **Must Fix** - FIXED

 bundle-build is not wired as a required status check — the gate may be unenforced.

The CI YAML adds bundle-build as a parallel standalone job with no needs: relationship and no other job gating on it. If it isn't listed in the repo's branch protection required checks, PRs can be merged despite a failing build — the job runs, turns red, and is ignored. YAML itself provides no enforcement. Before merging, confirm bundle-build appears in the required status checks list. If it doesn't, the entire point of the CI gate is lost.

*Note for whoever configures branch protection required checks after this lands: the required check must be either ci (the wrapper) or bundle-build directly, NOT the inner rendered names like ci / build or ci / it-postgres. The needs: chain only enforces transitively through the wrapper job; listing inner names instead silently breaks the gate.

**Nits**

  - bundle-build runs checkout@v6 — the rest of the repo uses v4. Inconsistency, not a blocker.
  - The docs/plans/ files from #2456 are absent here. Good cleanup.
  - api.d.ts is also absent (was 30k lines in #2456). Good.

**What's Good**

  - locations.js comment explicitly references SITES-45260 — future contributors will know exactly why this is a .js file instead of .json.
  - The TOCTOU race handling returns 409 with the winner's semrushProjectId — clients get a useful, actionable response instead of an opaque 500.
  - Publish-failure structured logging includes every field an operator needs to find and delete the orphaned Semrush project.
  - 403 on workspace mismatch is the right status code and is tested.
  - The CI YAML comment is unusually thorough — names the ticket, explains the gap in mysticat-ci, and describes what hedy --test-bundle actually does. Good institutional memory.

MysticatBot-Dev (approved):
Based on my thorough analysis of the full diff, existing discussion, and PR context, I'll write the consolidated review directly given the clear quality of this PR.

Hey @rainer-friederich,

## Strengths

- **Root cause properly fixed, not papered over** - converting `data/locations.json` to `src/support/semrush/data/locations.js` (`locations.js:18-94`) eliminates the entire class of bundling failure by making the data a normal ESM import. The comment explicitly referencing SITES-45260 is excellent institutional memory.
- **Defense-in-depth error handling** - `src/controllers/semrush.js:157-179` (`mapError`) correctly separates `ErrorWithStatusCode`, `SemrushTransportError`, and unexpected errors into distinct response envelopes. Upstream bodies are deliberately NOT leaked to clients (logged server-side only) - this is the right security posture for a proxy.
- **TOCTOU race handling in createProject** (`src/support/semrush/handlers/projects.js:361-397`) - the 409 gate before upstream calls plus the unique-constraint catch after DB write covers both the happy-path duplicate and the concurrent-create race. Returning the winner's `semrushProjectId` gives clients an actionable 409.
- **Transport security** - `rest-transport.js:54-66` validates HTTPS-only on the base URL override, preventing a compromised env var from redirecting IMS bearers to an attacker host. Path segments routed through `encodeURIComponent` uniformly (`rest-transport.js:139`).
- **Query/body separation** - `src/controllers/semrush.js:111-123` (`extractQuery`) explicitly refuses to fall back to `context.data` on parse failure, preventing silent attribute confusion on GETs. The test at line 4134-4144 codifies this contract.
- **`verifyPathWorkspace`** (`src/controllers/semrush.js:265-277`) is a clean defense-in-depth check ensuring the path `:workspaceId` matches the org's resolved workspace, preventing cross-tenant access even if Semrush's ACL has gaps.
- **CI gate with real validation** - `bundle-build` runs `hedy --test-bundle` which invokes the bundled Lambda against a healthcheck event, not just a build-succeeds check. This catches the exact failure class that caused the incident.
- **Comprehensive test coverage** - 525-line controller test with parameterized error-path sweep across all 9 handlers, plus dedicated handler and transport test suites. Codecov confirms full coverage on modified lines.

## Issues

### Minor (Nice to Have)

1. **`actions/checkout@v6` vs repo standard `v4`** - `.github/workflows/ci.yaml:35` uses `checkout@v6` while the rest of the repo's workflows use `v4`. Not a blocker (v6 is the newer version), but the inconsistency means Dependabot or Renovate will eventually bump them separately, creating unnecessary churn.

   *How to fix:* Either align to v4 for consistency now, or bump the mysticat-ci reusable workflow's checkout version in a follow-up so both converge on v6. Low priority.

2. **`handleUpdatePrompt` silently continues after delete failure** - `src/support/semrush/handlers/prompts.js:368-374` logs a warning when deleting the old prompt fails but proceeds to create the new one. This is intentional PATCH semantics (the old prompt may already be gone), but it means a transient upstream error could leave both old and new prompts coexisting until the next publish. The warning log is sufficient for now; a follow-up could verify the old prompt still exists before deciding whether to propagate.

## Recommendations

- Once this lands, confirm `bundle-build` is listed in the repo's branch protection required status checks (as @aliciadriani noted in the PR discussion). The YAML `needs: bundle-build` enforces ordering within the workflow, but branch protection enforcement is what prevents merge-on-red.

## Assessment
Ready to merge? Yes
Reasoning: This is a clean re-land of a previously-reviewed feature with the root cause properly fixed (ESM import instead of `readFileSync`), a load-bearing CI gate that catches the exact failure class, and comprehensive test coverage. The code quality is high - error handling is explicit, security boundaries are well-drawn, and race conditions are handled correctly. The one operational concern (required status check configuration) is outside the PR's scope and already tracked in the discussion.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 35s | <b>Cost:</b> $1.44 | <b>Commit:</b> <code>f63bcf4a8139d070a406a234a9be9a7eb69a22b0</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Re: the `actions/checkout@v6` nit — keeping it at `@v6` rather than downgrading.

The "rest of the repo uses v4" framing isn't current any more. Inventory on `main`:

```
.github/workflows/ci.yaml::bundle-build  @v6  (this PR — flagged)
.github/workflows/e2e-tests.yaml         @v6  (already on main, bumped by Renovate in #2384)
.github/workflows/semver-check.yaml      —    (no checkout step)
```

`e2e-tests.yaml` was bumped `@v4 → @v5 → @v6` over time by Renovate (#1080, #2384). Locally we're already on `@v6`; the comparison anchor is the `mysticat-ci` reusable workflow, which lives in a different repo.

Downgrading this file to `@v4` would *create* the inconsistency it's supposedly fixing — `ci.yaml @v4` vs `e2e-tests.yaml @v6`. Keeping `@v6` is the locally-consistent choice and matches where Renovate is taking us anyway.

The `handleUpdatePrompt` warn-then-continue nit and the branch-protection-required-checks note: both deliberately out of scope (intentional PATCH semantics + ops config respectively).

PR #2467, opened 2026-05-22
feat(serenity): rename /semrush → /serenity + post-launch hardening

rainer-friederich:
## Summary

Follow-up to the AIO proxy re-land in #2466. Bundles four work-streams:

1. **Rename URL surface back to `/serenity/*`** (the team-agreed product name) — endpoint paths + matching code; the upstream vendor (Semrush) keeps its name where it belongs (env vars, DB columns, shared-lib entity).
2. **Strict env/Vault for `SEMRUSH_PROJECTS_BASE_URL`** — no source default; missing config now fails fast as a typed `503 configurationError` rather than a generic 500.
3. **Real-upstream alignment** — fixes for the language-catalog field shape and the project `type` enum that the original implementation got wrong against the actual `adobe-hackathon.semrush.com` runtime.
4. **Drop the hand-curated `locations.js` table** — `location_id` is derived from ISO 3166-1 numeric via the `2000 + numeric` Google Ads geo-target prefix formula, validated against every live project in the Adobe LLMO-Dev workspace.

Plus the post-merge review fixes (baseUrl origin-strip security fix, typed 503, validator alignment, expanded test coverage, doc polish — see "Review follow-ups" below).

### 1. Rename `/semrush/*` → `/serenity/*` (commit `8e2e5ed5`)

| Layer | Change |
|---|---|
| Routes | 9 paths: `/v2/orgs/.../semrush/*` → `/v2/orgs/.../serenity/*` |
| Controller | `src/controllers/semrush.js` → `serenity.js`; `SemrushController` → `SerenityController` |
| Handler module | `src/support/semrush/` → `src/support/serenity/`; `createSemrushTransport` → `createSerenityTransport`; `SemrushTransportError` → `SerenityTransportError` |
| OpenAPI | `docs/openapi/semrush-api.yaml` → `serenity-api.yaml`; tag `semrush` → `serenity`; `listSemrush*` operationIds → `listSerenity*` |
| Docs | `docs/semrush.md` → `docs/serenity.md` |
| Tests | mirror source paths/names |

DB columns (`semrush_location_id`, `semrush_workspace_id`, …), `BrandSemrushProject` entity in `spacecat-shared`, and `SEMRUSH_PROJECTS_BASE_URL` env var stay as-is — those name the upstream vendor, not our local endpoint surface.

### 2. Strict env config — `SEMRUSH_PROJECTS_BASE_URL`

| File | Change |
|---|---|
| `src/support/serenity/rest-transport.js` | Drop `DEFAULT_BASE_URL`. `baseUrl(env)` validates https + URL-parseability + non-empty, returns `${parsed.protocol}//${parsed.host}` (strips path/query/userinfo). Throws typed `ErrorWithStatusCode(msg, 503)` on every failure. |
| `src/controllers/serenity.js` | `errorTokenForStatus` maps `503 → 'configurationError'` so the envelope distinguishes ops failures from runtime bugs. |
| `.env.example`, `docs/serenity.md` | Env table flipped from "override / default" to "required / Vault-backed". Operator runbook for the `vault kv patch` writes. |

Vault is populated in dev/stage/prod (`https://adobe-hackathon.semrush.com` today). Lazy resolution inside `buildTransport` — failure surface limited to `/serenity/*` routes.

### 3. Real-upstream alignment

Empirical findings from running against `adobe-hackathon.semrush.com`:

| Original (didn't work) | Reality (verified) | Fix |
|---|---|---|
| `GET /v1/languages` returns `{id, code}` | Actually `{id, name}` — no BCP-47 code field | Index catalog by English name. Use `Intl.DisplayNames(['en'], {type:'language'})` to map caller's BCP-47 primary subtag → English name → catalog UUID. |
| `POST /v1/.../projects` body `type: 'aio'` | Rejected with `ProjectRequest.Type ... 'oneof'`. Real type is `'ai'` | Send `type: 'ai'`. Validator now hard-requires the client to send `'ai'` too (no more legacy `'aio'`). |

### 4. Drop `locations.js` — derive `location_id` from ISO 3166-1 numeric

Verified empirically against every live project in the Adobe LLMO-Dev Semrush workspace:

```
US → 2 + 840 = 2840 ✓    DE → 2 + 276 = 2276 ✓
FR → 2 + 250 = 2250 ✓    AU → 2 + 36  = 2036 ✓
…
```

Google Ads "Geo Target ID" for **countries** = `2000 + ISO 3166-1 numeric`. The leading `2` is Google's country-type prefix; the trailing digits are the country's UN M.49 numeric code. Semrush echoes the same `location.id` back on read, so it's stable.

| File | Change |
|---|---|
| `src/support/serenity/handlers/projects.js` | `resolveLocation()` now reads `iso31661Alpha2ToNumeric[a2]` + `Intl.DisplayNames(['en'], {type: 'region'})`. Doc block explains the formula, links to Google's canonical CSV, leaves a TODO for cities/ZIPs (which don't follow the formula). |
| `src/support/semrush/data/locations.js` | **Deleted** (94 lines). |
| `package.json` | Add `iso-3166` dep (~5 KB, same author as our existing `iso-639-3`). |

**Out of scope:** cities, postal codes, airports. Those don't follow the formula — when sub-national geo lands in the UX, this single function gets a new lookup path (Semrush location-search endpoint if exposed, or S3-backed Google CSV). DB slice key stays `semrush_location_id`, no schema churn needed. **Kosovo (XK) limitation** documented in Troubleshooting — XK has no official ISO numeric so the formula can't be applied; its real Geo Target ID needs the sub-national path.

### 5. Review follow-ups (commit `f3295012`)

| Item | Action |
|---|---|
| `baseUrl()` returns `candidate` (security) | Now returns `${parsed.protocol}//${parsed.host}` — strips path, query, fragment, userinfo. Two new tests assert path/query/fragment + userinfo strip-down. |
| Missing env → 500 not 503 | `baseUrl()` throws `ErrorWithStatusCode(msg, 503)`; `errorTokenForStatus` adds `503 → 'configurationError'`. Controller test verifies the envelope shape. |
| `validateCreateBody` projectType stale | Hard-requires `'ai'`. The previous `'aio'` accept was an artifact of confusing the GET-listing `?type=AIO` collection filter with the CREATE body type. |
| Dual-vocabulary contract untested | Sweep test over 5 bad values + success test asserting upstream body carries `type: 'ai'`. |
| Cache TTL expiry untested | Sinon fake timers: within TTL = 1 `listLanguages` call, past 1h = 2 calls. |
| Troubleshooting + `.env.example` | New 503 + Kosovo XK troubleshooting rows. `.env.example` value commented out so `cp .env.example .env` doesn't silently aim a fresh local stack at the live upstream. |

### 6. Other cleanup

- **Bump `@adobe/spacecat-shared-data-access` 3.69.0 → 3.70.1** — picks up [adobe/spacecat-shared#1617](https://github.com/adobe/spacecat-shared/pull/1617) which fixes the `BrandSemrushProject` `belongs_to: 'Brand'` reference causing `Collection BrandCollection not found` at runtime.
- **Remove openapi typegen plumbing** (`gen:types:semrush`, `openapi-typescript` devDep, generated `.d.ts` files). Not adopted, was adding bundle/build complexity without payoff.
- **Drop dead-defensive guard** in `isoToEnglishName` — `LANGUAGE_TAG_REGEX` upstream guarantees the input is non-empty.

## Verification

- Smoke-test cycle against `https://adobe-hackathon.semrush.com` with stage IMS tokens — auth, list, create, duplicate detection, language resolution, location resolution all working.
- `npx mocha test/support/serenity/**/*.test.js test/controllers/serenity.test.js test/openapi-contract/serenity-api.test.js` → **185 passing**.
- c8 coverage on touched serenity files: **100% line / func / stmt** on controller + transport + both handlers.
- `npm run lint` → clean.
- `npm run docs:lint` → clean.
- `npm run build` → exit 0 (`hedy --test-bundle` healthcheck 200).

## Dependencies / sequencing

- Now stacks directly on `main`. Conflicts from the prior #2466 merge resolved by keeping the further-evolved versions on this branch (formula-based locations, `type: 'ai'`, English-name catalog index, strict env, no typegen, `/serenity/*` URLs).
- Deploy sequence: this lands → semantic-release ships **minor** → smoke `/serenity/*` against dev → done. Vault is already populated; no coordinated changes at merge time.
- elmo-ui follow-up needed: client API methods + hook + type names still say `Semrush*` and call `/semrush/*` URLs. They'll be renamed in the same elmo PR that's already in flight for the new prompt-management page.


rainer-friederich:
## Local multi-agent review

Ran a multi-agent review (backend-api, nodejs, security, configuration, testing, documentation experts). Node.js verdict is clean — `iso-3166` bundles correctly with esbuild, `Intl.DisplayNames` is safe at module load in Lambda's Node 24 runtime. Issues below sorted by confidence score.

---

### FIX (score ≥ 75)

**[security] `baseUrl()` returns `candidate` instead of `parsed.origin`** — a value like `https://host/path-prefix` silently prepends that path to every upstream request; `https://user:pass@host` leaks credentials in every fetch. The comment claims this check prevents IMS bearer redirect but `candidate` retains path/userinfo/query. Fix: `return \`${parsed.protocol}//${parsed.host}\`` instead of `return candidate`.
_File: `src/support/semrush/rest-transport.js:56-66`_

**[docs] Missing migration guidance** — `docs/semrush.md` doesn't call out that operators who relied on the previous `DEFAULT_BASE_URL` must write to Vault before deploying. First `/semrush/*` request in prod 500s after upgrade with no pre-deploy signal.
_File: `docs/semrush.md:11`_

**[docs] Broken markdown table** — the `|---|---|---|---|` separator row appears after the data row, breaking rendering in GitHub and all markdown viewers.
_File: `docs/semrush.md:13-15`_

**[docs] Missing Troubleshooting entry** — no row for the new required env var: what error string appears in logs, what HTTP response users see, how to resolve it. All other failure modes in the existing table have this.
_File: `docs/semrush.md` (Troubleshooting section)_

**[backend] Empty language catalog silently pins empty cache for 1h** — the warn at line 137 only fires when `items.length > 0`. A transient upstream blank response causes all `POST /semrush/projects` to return 400 `unknownLanguage` for an hour with no observable log entry.
_File: `src/support/semrush/handlers/projects.js:128-143`_

**[config] No pre-deploy validation gate** — no CI/deploy step asserts `SEMRUSH_PROJECTS_BASE_URL` is present in Secrets Manager before the Lambda goes live. A missed Vault write only surfaces as a prod 500.

**[config] No rotation/TTL story** — `secrets` middleware injects env on cold start only. Rotating the value (e.g. prod cutover to a real Semrush host) leaves warm containers on the old URL indefinitely. No documented max-TTL or force-recycle procedure.

**[config] `.env.example` ships production hackathon URL** — `cp .env.example .env` silently points local dev at the real upstream. Comment says "Required" but the populated value masks that requirement. Use a non-routable placeholder or explicitly comment it out.
_File: `.env.example`_

**[testing] Dual-vocabulary contract untested** — no test explicitly passes `projectType: 'aio'` and asserts the upstream body gets `type: 'ai'`. The happy-path test omits `projectType` entirely.
_File: `test/support/semrush/handlers/projects.test.js:271-308`_

**[testing] Language cache TTL expiry untested** — tests verify "within TTL only one `listLanguages` call" but no test advances time past TTL and asserts a refresh. A bug that pinned `expiresAt` permanently would not fail CI.
_File: `test/support/semrush/handlers/projects.test.js:433-450`_

---

### CONSIDER (score 50–74)

**[bug] `validateCreateBody` stale validator value** — accepts only `projectType: 'aio'` while the handler always sends `type: 'ai'` upstream. Future callers passing `'ai'` (the correct upstream value) will be rejected by the validator.
_File: `src/support/semrush/handlers/projects.js:292`_

**[security] Protocol-only check overstated** — `baseUrl()` enforces `https:` but the comment claims it prevents IMS bearer redirect to attacker hosts. Any `https://attacker.com` passes; the check prevents accidental http redirect, not deliberate https redirect from a compromised env.
_File: `src/support/semrush/rest-transport.js:46-67`_

**[backend] Missing env → 500 not 503** — plain `Error` from `baseUrl()` maps to generic `500 internalServerError` via `mapError`. A configuration failure is indistinguishable from a runtime bug. Typed `503 configurationError` matches the existing pattern for missing PostgREST.
_File: `src/support/semrush/rest-transport.js:50-55`_

**[config] No `vault kv patch` vs `put` warning** — the runbook uses `patch` (correct) but doesn't warn that `put` would wipe all sibling secrets in `dx_mysticat/<env>/api-service`.
_File: `docs/semrush.md` (Vault writes section)_

**[testing] Formula expansion coverage gap** — formula now accepts ~250 ISO codes but only US→2840 is tested. A regression in `iso-3166` numeric mappings for DE→2276, FR→2250, AU→2036, GB→2826 would not fail CI.
_File: `test/support/semrush/handlers/projects.test.js:74-86`_

**[testing] Kosovo XK silent 400** — `iso31661Alpha2ToNumeric['XK']` is `undefined` (no official ISO numeric), so `resolveLocation('XK')` returns `null` → 400. Kosovo has a real Google Geo Target ID (2061632); worth documenting as a known limitation.
_File: `test/support/semrush/handlers/projects.test.js:74-86`_

**[testing] Language subtag-stripping untested** — `'pt-BR'` → strips to `'pt'` → `isoToEnglishName` → catalog lookup is a real code path for any caller using regional language tags, but has no dedicated test.
_File: `test/support/semrush/handlers/projects.test.js:259-308`_

**[docs] Duplicate `VAULT_ADDR` export** — appears before both `vault login` and the verify loop; the inline comment on the second ("CLI default is 127.0.0.1:8200") implies the verify block needs it but the login block doesn't, which is wrong.
_File: `docs/semrush.md:31-32`_

**[docs] Architecture sentence contradicts per-env framing** — opening sentence hardcodes `adobe-hackathon.semrush.com` as the upstream while the env-config section explicitly says the value is per-environment.
_File: `docs/semrush.md:7`_

---

🤖 Generated with [Claude Code](https://claude.ai/code)

rainer-friederich:
## Review items intentionally not addressed

For traceability, here's why each unaddressed item from the multi-agent review was left out of `f3295012`.

### From the FIX (≥75) list

**#2 — Missing migration guidance in `docs/serenity.md`**
Skipped. The previous `DEFAULT_BASE_URL` only ever pointed at the same hackathon host; there are no callers relying on the old default-fallback semantics in any environment. No back-compat surface to migrate, so no migration guide needed. Vault writes are already done in dev/stage/prod.

**#5 — Empty language catalog silently pins empty cache for 1h**
Skipped. The concern requires the upstream to return a structurally-valid response with `items.length === 0`, which would only happen if Semrush briefly served a degraded catalog. In practice we've never seen this; the Vault-backed config is correct and the catalog has been populating consistently. The `length > 0` warn-log will still fire if items arrive but the field shape drifts (the more likely failure mode). Filing this as a future-hardening note rather than addressing inline.

**#6 — No pre-deploy validation gate that `SEMRUSH_PROJECTS_BASE_URL` is in Secrets Manager**
Skipped (out of code scope). This would be a CI workflow change, not an api-service code change, and the typed 503 we now emit gives operators a clear signal at first request rather than a generic 500. If we add a pre-deploy gate later it'll be in `mysticat-ci`, not here.

**#7 — No rotation/TTL story for warm containers**
Skipped (out of code scope). Lambda's `secrets` middleware injects env on cold start; rotation requires forcing a redeploy or alias flip, which is the existing operational pattern for every other Vault-backed config in the repo. No new pattern to invent for this one variable.

**#8 — `.env.example` ships production hackathon URL**
Addressed differently. Instead of replacing the value with a placeholder, the line is now commented out and the hackathon URL is preserved in the surrounding comment block as a reference. `cp .env.example .env` no longer silently activates the URL; local devs still see the example value for orientation.

### From the CONSIDER (50–74) list

**#12 — Protocol-only check overstated in `baseUrl()`**
Skipped (framing nit). The `https://` check exists to prevent the accidental `http://` fall-back; the doc comment was misleading about deliberate-attacker scenarios. The new path/userinfo strip-down in `f3295012` covers the *real* attack surface (silent path-prefix injection, userinfo credential leak). Updated the doc comment in that commit to drop the overstated framing.

**#14 — No `vault kv patch` vs `put` warning**
Skipped (low-leverage doc nit). The runbook already shows the correct `patch` invocation; if someone runs `put` instead and wipes sibling secrets, the deploy will obviously fail and Vault has an audit log. Not worth bloating the doc.

**#15 — Formula expansion coverage gap (only US→2840 tested)**
Skipped (low risk). The formula `2000 + iso31661Alpha2ToNumeric[a2]` is mechanical and the `iso-3166` package is a stable upstream lib. A regression would have to come from a bug in the lib itself, which would surface broader than just our formula. The new Kosovo XK test (#16) covers the *interesting* boundary (no-numeric case); the spot-check pattern in the existing test is enough for the assigned-country happy path.

**#17 — Language subtag-stripping untested (`'pt-BR'` → `'pt'`)**
Skipped (minor gap). The `LANGUAGE_TAG_REGEX` already accepts the `<2-3 letters>-<2-4 letters>` shape, and the existing `isoToEnglishName` runs `.split('-')[0]` on every input — the path is exercised by the broader create-project happy path tests. A dedicated unit test would be 5 lines of additional coverage on already-100%-line-covered code.

**#18 — Duplicate `VAULT_ADDR` export in docs/serenity.md**
Skipped (minor doc nit). Both occurrences are intentional — one before the `vault login` block (some shells lose env between paragraphs in copy-paste), one in the verify loop because the original draft was ambiguous about whether the verify loop runs in the same shell. The duplication is a defensive choice for runbook copy-pasters, not a mistake.

**#19 — Architecture sentence contradicts per-env framing**
Skipped (minor doc nit). The opening sentence names `adobe-hackathon.semrush.com` because that's the literal value currently in Vault for dev/stage/prod. When the prod target changes, the opener gets updated then. Per-env framing is preserved further down.

---

Happy to revisit any of these if you'd rather see them addressed inline.

aniham (approved):

PR #2468, opened 2026-05-22
fix(fixes): load suggestions in getAllForOpportunity without fixCreatedDate (SITES-45274)

abhishekgarg18:
## Summary

- **Deployed tab shows count=2 but only renders 1 row** — the initial fetch path (`getAllForOpportunity` without `fixCreatedDate`) returns fix entities WITHOUT suggestions attached. The UI renders rows by iterating `fixEntity.suggestions`, so missing suggestions = missing rows.

**Tracking:** [SITES-45274](https://jira.corp.adobe.com/browse/SITES-45274)

## Root cause

`src/controllers/fixes.js:99-149` has two code paths:

| Path | When | Loads suggestions? |
|---|---|---|
| WITH `fixCreatedDate` (lines 111-136) | User clicks a date group | **YES** — `getAllFixesWithSuggestionByCreatedAt` + attaches `_suggestions` |
| WITHOUT `fixCreatedDate` (lines 139-149) | Initial load (count + overview) | **NO** — `allByOpportunityId` only |

`FixDto.toJSON` (line 56-61) only includes `suggestions` when `fix._suggestions` is truthy. The UI's `DeployedViewTable.tsx:98` renders via `fixEntity.suggestions.forEach()` — no suggestions = 0 rows, but the fix entity still counts toward the tab badge.

## Fix

Added `getSuggestions()` loading to the no-filter path (line 139), matching the existing pattern in the `fixCreatedDate` path.

## Test plan

- [ ] CI tests
- [ ] Verify on Maurice Blackburn prod: Deployed tab count matches rendered rows
- [ ] Verify on JLR prod: backfilled FixEntities render correctly

### DIFF
--- a/src/controllers/fixes.js
+++ b/src/controllers/fixes.js
@@ -145,6 +145,18 @@ export class FixesController {
       return res;
     }
 
+    // SITES-45274: attach suggestions to each fix entity so FixDto.toJSON
+    // includes them in the response. Without this, the UI's DeployedViewTable
+    // renders 0 rows (it iterates fixEntity.suggestions) even though the tab
+    // counter correctly counts the fix entities.
+    // The fixCreatedDate path above (lines 112-126) already does this via
+    // getAllFixesWithSuggestionByCreatedAt; this path was missing it.
+    await Promise.all(fixEntities.map(async (fixEntity) => {
+      const suggestions = await fixEntity.getSuggestions();
+      // eslint-disable-next-line no-underscore-dangle,no-param-reassign
+      fixEntity._suggestions = suggestions;
+    }));
+
     fixes = fixEntities.map((fix) => FixDto.toJSON(fix));
     return ok(fixes);
   }


abhishekgarg18:
## Why this change is needed here and only here

### The bug
The `getAllForOpportunity` endpoint in `fixes.js` has two code paths:

1. **WITH `fixCreatedDate`** (lines 111-136): calls `getAllFixesWithSuggestionByCreatedAt` → attaches `_suggestions` to each fix entity → `FixDto.toJSON` includes `suggestions[]` in the response
2. **WITHOUT `fixCreatedDate`** (lines 139-149): calls `allByOpportunityId` → does **NOT** attach `_suggestions` → `FixDto.toJSON` **omits** `suggestions` field entirely

The UI's `DeployedViewTable.tsx:98` renders rows via `fixEntity.suggestions.forEach()`. When `suggestions` is missing/undefined from the API response, 0 rows render — but the fix entity still counts toward the tab badge ("Deployed (2)" but only 1 row visible).

### Why fix it in the API, not the UI
The API is returning **incomplete data** — same endpoint, same entity, but different fields depending on whether a query param is present. This violates the principle of consistent API responses. The UI shouldn't have to handle "sometimes suggestions are included, sometimes not" — the DTO should always include the relationship data.

The `fixCreatedDate` path already does this correctly (line 124: `fixEntity._suggestions = item.suggestions`). This fix simply applies the same pattern to the no-filter path.

### Why NOT fix it in Mystique
Mystique creates FixEntities correctly — `_create_fix_entity` at `autofix_handler.py:1217` passes `suggestionIds` to `spacecat.create_fixes()`. The FixEntity-to-Suggestion relationship is stored properly in SpaceCat. The bug is purely in the read path (this endpoint), not the write path.

### Why NOT fix it in the autofix-worker
The autofix-worker creates FixEntities with `suggestions: [suggestion.getId()]` (line 251 in `alt-text/handler.js`). Again, the write is correct — the read is the problem.

### Edge cases handled
- **Empty fix entities list** (`allByOpportunityId` returns `[]`): The existing `checkOwnership(fixEntities[0], ...)` on line 143 already handles this (returns 404)
- **Fix entity with 0 suggestions** (e.g., manual-fix path that doesn't link suggestions): `getSuggestions()` returns `[]`, `_suggestions` is set to `[]`, `FixDto.toJSON` includes `suggestions: []` — the UI renders a row with empty suggestion data instead of dropping the row entirely
- **Large number of fix entities**: `Promise.all` parallelizes the suggestion loading. For N fix entities with M suggestions each, this is N+1 queries (1 for fixes + N for suggestions). Acceptable for the Deployed tab which typically has <50 fix entities per opportunity.

### Companion PR (different bug, same tab)
[Mystique PR #2347](https://git.corp.adobe.com/experience-platform/mystique/pull/2347) (SITES-45273) enriches FixEntity `changeDetails` with AEM page links. That's a separate issue — rows are present but lack clickable links. This PR makes the rows appear; that PR makes them useful.

mimchome (commented):
## Review

**The controller fix is correct** -- it mirrors the existing `fixCreatedDate` pattern and solves the Deployed tab rendering gap. Two things to address:

### 1. Missing test (blocker)

The existing test for the no-`fixCreatedDate` path doesn't stub `getSuggestionsByFixEntityId`, so it doesn't verify suggestions are attached. Here's a replacement that covers the new behavior:

```js
it('can get all fixes for an opportunity with suggestions attached', async () => {
  const suggestion1 = await suggestionCollection.create({
    opportunityId,
    type: Suggestion.TYPES.CONTENT_UPDATE,
  });
  const suggestion2 = await suggestionCollection.create({
    opportunityId,
    type: Suggestion.TYPES.REDIRECT_UPDATE,
  });

  const fixEntities = await Promise.all([
    fixEntityCollection.create({
      type: Suggestion.TYPES.CONTENT_UPDATE,
      opportunityId,
      changeDetails: { arbitrary: 'value 1' },
    }),
    fixEntityCollection.create({
      type: Suggestion.TYPES.REDIRECT_UPDATE,
      opportunityId,
      changeDetails: { arbitrary: 'value 2' },
    }),
  ]);
  fixEntityCollection.allByOpportunityId.resolves(fixEntities);

  fixEntityCollection.getSuggestionsByFixEntityId
    .withArgs(fixEntities[0].getId()).resolves([suggestion1]);
  fixEntityCollection.getSuggestionsByFixEntityId
    .withArgs(fixEntities[1].getId()).resolves([suggestion2]);

  const response = await fixesController.getAllForOpportunity(requestContext);

  expect(response).includes({ status: 200 });
  const result = await response.json();
  expect(result).to.have.lengthOf(2);
  expect(result[0].suggestions).to.have.lengthOf(1);
  expect(result[0].suggestions[0].id).to.equal(suggestion1.getId());
  expect(result[1].suggestions).to.have.lengthOf(1);
  expect(result[1].suggestions[0].id).to.equal(suggestion2.getId());
});
```

Verified locally: fails without the controller fix (`suggestions` is undefined), passes with it.

### 2. N+1 query cost (follow-up, not a blocker for this PR)

The `fixCreatedDate` path uses `getAllFixesWithSuggestionByCreatedAt` -- a batch method that issues **3 PostgREST calls** total regardless of how many fix entities exist. This PR's path calls `getSuggestions()` per fix entity, which is **2N+1 calls**.

Concrete example: an opportunity with 5 fix entities, each with 3 suggestions:
- **This PR (N+1):** 1 call for fix entities + 5x(1 junction query + 1 suggestions batch-get) = **11 PostgREST calls**
- **Batch path:** 1 junction query + 1 fix entities batch-get + 1 suggestions batch-get = **3 PostgREST calls**

At worst case (~200 fix entities for a large customer site), that's 401 calls at 10-50ms each = 4-20s inside a 29s Lambda timeout. It works but is slow.

**Follow-up fix in spacecat-shared** (`fix-entity.collection.js`): add `getAllFixesWithSuggestionsByOpportunityId(opportunityId)`. It's nearly identical to `getAllFixesWithSuggestionByCreatedAt` (lines 174-243), with one change at the junction query:

```js
// Current (queries by opportunityId + date):
const fixEntitySuggestions = await fixEntitySuggestionCollection
  .allByOpportunityIdAndFixEntityCreatedDate(opportunityId, fixEntityCreatedDate);

// New (queries by opportunityId alone -- index already exists, schema line 51):
const fixEntitySuggestions = await fixEntitySuggestionCollection
  .allByIndexKeys({ opportunityId });
```

The junction table already has an index with `opportunityId` as partition key (`{ composite: ['opportunityId'] }` in the schema), so this query is supported without schema changes. The rest of the method (group by fix entity, batch-get entities, batch-get suggestions, correlate in memory) is identical.

Then in the api-service controller, replace the `Promise.all` with the same pattern used at lines 112-126:

```js
const fixEntitiesWithSuggestions = await this.#FixEntity
  .getAllFixesWithSuggestionsByOpportunityId(opportunityId);

fixEntities = fixEntitiesWithSuggestions.map((item) => {
  const { fixEntity } = item;
  fixEntity._suggestions = item.suggestions;
  return fixEntity;
});
```

This makes both code paths O(3) calls. Suggest filing a follow-up ticket for the batch method.

mimchome (approved):
Approve with suggestions -- the controller fix is correct and bounded N+1 is acceptable short-term. Please add the test from the review comment before merging. See the review thread for the batch method follow-up in spacecat-shared.

PR #2471, opened 2026-05-22
feat: add sorting parameters for FTS topic research endpoints | LLMO-4963

calvarezg:
This update introduces new sorting parameters for the FTS topic research endpoints in the AI Visibility API. The following parameters have been added:

- `sortBy` options for topics, prompts, brands, and source domains, allowing sorting by relevance score, volume, prompt, mentioned brands count, sources count, and domain.
- `sortDirection` options to specify ascending or descending order.

These changes enhance the API's flexibility in handling search queries and improve the user experience by providing more control over the results returned. Additionally, the sorting logic has been implemented in the corresponding handler functions, ensuring consistent behavior across different endpoints.

The change is fully non-breaking: both query params are optional, and their defaults reproduce the pre-existing server-side behaviour for every endpoint (`RELEVANCE_SCORE` for topics, `MENTIONED_BRANDS_COUNT` for prompts, `MENTIONS` for brands and source domains, `DESC` for all). Invalid `sortBy` / `sortDirection` values return `400 invalid_sort_by` / `400 invalid_sort_direction` rather than silently defaulting.

## Related Issues
[LLMO-4963](https://jira.corp.adobe.com/browse/LLMO-4963)


calvarezg:
Thanks for the review — addressed all five points in the latest push. Quick walkthrough:

### 1. `example: DESC` misplaced in YAML ✅
Fixed in `docs/openapi/ai-visibility-api.yaml`. The stray `example: DESC` was deleted from `x-ai-vis-fts-source-domains-sort-by` (where `DESC` isn't in the enum and was failing as an invalid OpenAPI example) and moved up to `x-ai-vis-fts-sort-direction`, where it actually belongs. `npm run docs:lint` is clean for these schemas.

### 2. Field-name mismatch in `promptsResearchSortKey` — no bug, but worth hardening 🟡
Wanted to push back on this one carefully. There isn't a bug — but the test was weak enough that a hypothetical bug would have slipped through, so you were right to flag it.

The data flow:

1. The gRPC client returns prompts with the proto field `sourcesCount`.
2. `mapPromptRow` (line ~281) renames it on the way out: `citedPages: num(p.sourcesCount)`. So the mapped row exposes `citedPages`, not `sourcesCount`.
3. `promptsResearchSortKey` is called on the **mapped** row (see `handleTopicsResearchPrompts` line ~585: `const row = mapPromptRow(p); ... withKeys.mentionSortKey = sortKey(withKeys);`).
4. The test stub feeds `sourcesCount: 99` into the *gRPC mock*, which is correct because that's the proto-side contract; `mapPromptRow` then converts it to `citedPages: 99` before the sort key accessor sees it.

So the SOURCES_COUNT sort does fire correctly today.

**However**, the test wasn't proving that. The original prompts were `'low sources'` / `'high sources'`, and because `'high' < 'low'` alphabetically, the comparator's tiebreak (`cmpStr(a.norm, b.norm)`) would put `'high sources'` first **even if `promptsResearchSortKey` silently returned `undefined` for every row**. The assertion would still pass on a regression — which is exactly the failure mode you were worried about, just by a different mechanism.

Two changes to close the gap:

- Renamed the test prompts to `'aaa low sources'` (sourcesCount=1) and `'zzz high sources'` (sourcesCount=99). The alphabetical tiebreak now *disagrees* with the SOURCES_COUNT order — if the sort accessor ever stops returning the right field, the test flips and fails.
- Added a comment block on `mapPromptRow` and `promptsResearchSortKey` documenting the proto→UI rename (`sourcesCount`→`citedPages`, `mentionedBrandsCount`→`mentions`) so the next reader doesn't trip on the same field-name question.

### 3a. `cmpNum` coerces undefined to 0 ✅
Added an inline comment explaining the intent (defensive against gRPC nulls / non-finite stringy values, avoids NaN making `Array.prototype.sort` non-deterministic) and flagging that "missing" and "genuine zero" are indistinguishable in the resulting order — useful context if anyone later wants a sentinel-aware variant.

### 3b. No tiebreak for DOMAIN sort ✅
Added a comment in `sourceDomainsResearchComparator` noting that the upstream `agg` Map is keyed by `sourceDomain`, so the sort key is already unique by the time the comparator runs and a tiebreak literally couldn't fire.

### 3c. PR checklist + issue link ✅
Updated the PR description: ticked the applicable boxes, added a Related Issues section explaining the motivation (`FT_LLMO-4799` flag, server-side sort for SR Prompt Research) and linking the consuming UI PR ([adobe/project-elmo-ui#1812](https://github.com/adobe/project-elmo-ui/pull/1812)).

Tests still pass (108/108 in `topics.test.js`) and `npm run lint` is clean.

dmitriigulkin-semrush (approved):

PR #2472, opened 2026-05-22
chore(it): diagnostic — dump open handles after IT teardown (DO NOT MERGE)

akshaymagapu:
## Purpose

Diagnostic-only probe to identify which handle is keeping the Node event loop alive after the PostgreSQL integration suite finishes, which intermittently causes \`ci/it-postgres\` to hit the ~15-minute step timeout and get cancelled on \`main\` (e.g. runs [26282298612](https://github.com/adobe/spacecat-api-service/actions/runs/26282298612), [26229325483](https://github.com/adobe/spacecat-api-service/actions/runs/26229325483), [26216823004](https://github.com/adobe/spacecat-api-service/actions/runs/26216823004)).

In every cancelled run, the log shows \`491 passing\` and successful container teardown, then ~11 minutes of silence with orphan \`npm exec mocha\` reported at cleanup. Mocha doesn't exit because something is holding the loop open.

## What this PR does

In \`test/it/postgres/harness.js\` \`afterAll\`, after the existing teardown:

1. Waits 5s for the loop to drain on its own.
2. If still alive, calls \`why-is-node-running\` to dump active handles with file:line.
3. \`process.exit(0)\` so the job ends in ~3 min instead of timing out at 15.

\`why-is-node-running\` added as a dev dependency.

## Expected outcome

CI logs will name the leaked handle. Hypothesis: the module-scope \`http.Agent({ keepAlive: true })\` inside \`@adobe/spacecat-shared-data-access\`'s \`@adobe/fetch\` context. Verified or refuted by the dump.

## After this PR

Close as not-for-merge. Open the real fix:
- Export \`resetDataAccessConnections\` from \`@adobe/spacecat-shared-data-access\` (calls \`fetchContext.reset()\`).
- Call it from the IT \`afterAll\` before \`stopPostgres()\`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -102,6 +102,7 @@
         "semantic-release": "^25.0.3",
         "sinon": "22.0.0",
         "sinon-chai": "4.0.1",
+        "why-is-node-running": "3.2.2",
         "yaml": "2.8.4"
       },
       "engines": {
@@ -704,7 +705,6 @@
       "resolved": "https://registry.npmjs.org/@adobe/helix-universal/-/helix-universal-5.4.1.tgz",
       "integrity": "sha512-wgmjwo0xJkYhFQUmv6GTPvCFjDYBoT7zP3OuAxLN+FlHgS6kDkbJOtKxwQn9SrWbhoIfM8GdCnRDpBn6BmkASw==",
       "license": "Apache-2.0",
-      "peer": true,
       "dependencies": {
         "@adobe/fetch": "4.3.0",
         "aws4": "1.13.2"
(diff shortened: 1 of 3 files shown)


akshaymagapu:
Closing — diagnostic served its purpose.

## What we learned

\`why-is-node-running\` (imported at module top so its \`async_hooks\` listener was in place from process start) reported **270 active handles** keeping the loop alive after \`afterAll\` completed. The vast majority traced to \`@adobe/fetch\`:

\`\`\`
node_modules/@adobe/fetch/src/core/request.js:91   - socket = tls.connect(...)
node_modules/@adobe/fetch/src/core/h1.js:192       - req = request(url, opts)
node_modules/@adobe/fetch/src/core/h2.js:178       - session = connect(...)
node_modules/@adobe/fetch/src/fetch/body.js:60     - return streamToBuffer(stream)
... (270 total)
\`\`\`

Source: the module-scope \`fetchContext = keepAliveNoCache()\` in \`@adobe/spacecat-shared-data-access/src/service/index.js\`. The pool's keep-alive Agent prevents the Node event loop from going idle. In Lambda that's intentional (warm-start connection reuse). In tests it's a leak — which is why \`ci/it-postgres\` intermittently hangs for 11 minutes and gets cancelled at the 15-min step timeout.

The "MainThread" orphan-process name from cancelled runs turned out to be a red herring — it's just Node 24's main thread label (\`prctl(PR_SET_NAME)\`), not a Python process.

## Proper fix

- **adobe/spacecat-shared#1621** — adds \`resetDataAccessConnections\` export (one-line wrapper around \`fetchContext.reset()\`).
- **#2475** — consumes it from \`test/it/postgres/harness.js\`. Draft until #1621 publishes.

No \`--exit\` flag, no \`process.exit\` hacks — the process exits cleanly because no handle holds the loop.

PR #2473, opened 2026-05-22
feat(llmo): add PATCH endpoint to update a row in an LLMO XLSX data sheet

dzehnder:
## Summary

Adds a `PATCH` endpoint on the LLMO `data` route so the elmo UI can update a single row in the SharePoint XLSX file that backs an LLMO data sheet — concretely, to mark one Strategic Recommendations prompt as `deleted: "true"` from the UI.

This is what enables the [project-elmo-ui PR #1804](https://github.com/adobe/project-elmo-ui/pull/1804) Strategic Recommendations feature to persist deletions back to the source of truth (SharePoint), rather than holding them only in local UI state.

## Routes

| Method | Path | Capability |
| --- | --- | --- |
| `PATCH` | `/sites/:siteId/llmo/data/:dataSource/row` | `site:write` |
| `PATCH` | `/sites/:siteId/llmo/data/:sheetType/:dataSource/row` | `site:write` |

### Body

```json
{
  "sheet": "Semrush",
  "match": { "topic_id": "14239648398373927587", "prompt": "What's the best free online photo editor for non-designers?" },
  "values": { "deleted": "true" }
}
```

- `sheet` — worksheet name inside the XLSX (`Semrush` / `Synthetic Personas` / `Citation Attempt` for the SR workbook)
- `match` — column-value pairs that must uniquely identify exactly one row (no match → 404, multiple matches → 409)
- `values` — column-value pairs to write to the matched row
- All keys in `match` and `values` must exist as columns in the worksheet header row (else 400)

### Success response (200)

```json
{
  "siteId": "…",
  "sheetType": "strategic-recommendations",
  "dataSource": "strategic-recommendations-template",
  "sheet": "Semrush",
  "rowNumber": 2,
  "updated": { "deleted": "true" }
}
```

## Implementation notes

- **Authentication / authorization**: standard LLMO chain — `getSiteAndValidateLlmo` (site exists, has `llmoConfig.dataFolder`, caller has access via LLMO entitlement) + `site:write` capability.
- **SharePoint path**: `${SHAREPOINT_URL}/${dataFolder}[/${sheetType}]/${dataSource}.xlsx`, using the existing `createSharePointClient` helper from `llmo-onboarding.js` (no new auth surface).
- **Workbook mutation**: download → load with `exceljs` → find unique matching row → write cells → upload via `uploadRawDocument` → publish `preview` and `live` via `admin.hlx.page` so the Helix CDN serves the new JSON.
- **Multi-sheet workbooks**: `workbook.getWorksheet(sheet)` targets by name, which the existing SDK's `findRowInSheet` / `updateSheetRowAt` cannot (they always pick the first worksheet). This is why this PR uses `exceljs` directly rather than the SDK's spreadsheet helpers.
- **Concurrency**: read-modify-write at the workbook level (last-writer-wins). Acceptable for the low-frequency UI soft-delete use case but documented in the JSDoc on `patchSheetRow`.
- **HLX publish failures**: logged as warnings and swallowed. The SharePoint write has already succeeded; the next scheduled republish cycle will pick the change up.

## Why a new endpoint (not direct UI → SharePoint)

The UI authenticates against SpaceCat with an IMS bearer; it has no path to SharePoint, and giving end-users SharePoint write access to a customer-shared sheet would be the wrong trust boundary. The same shape was used previously for `updateRedirects` / `updateMetadata` on `spacecat-shared-content-client`.

## Test plan

- [x] `npm test` — full `test/controllers/**` suite passes (4854 tests)
- [x] `npm run lint` — clean
- [x] `npm run docs:lint` — clean (no new warnings/errors)
- [x] Unit tests on `llmo-sheet-write.js` cover: path building, body validation, HLX publish branches, happy path (round-trips workbook via `exceljs`), 404 (workbook missing / worksheet missing / no row matches), 400 (unknown columns), 409 (ambiguous match), missing `ADMIN_HLX_API_KEY`, and `createSharePointClient` fallback when no client is injected
- [x] Unit tests on the new `patchLlmoDataRow` factory method cover: 200, 400 invalid body, 404, 409, 500
- [x] `test/routes/index.test.js` snapshot updated to include the new dynamic routes
- [x] `test/routes/required-capabilities.test.js` `route coverage` check passes (every new route is mapped)
- [ ] **Manual / dev**: post-deploy, hit the dev endpoint with a real site to confirm a SR prompt round-trips through SharePoint and re-appears as `deleted: "true"` in the JSON.

## Files touched

```
src/controllers/llmo/llmo-sheet-write.js        (new)
src/controllers/llmo/llmo.js                    (+import, +patchLlmoDataRow handler, +export)
src/routes/index.js                             (+2 routes)
src/routes/required-capabilities.js             (+2 site:write entries)
docs/openapi/api.yaml                           (+2 path refs)
docs/openapi/llmo-api.yaml                      (+ llmo-data-row-with-datasource, llmo-data-row-with-type)
docs/openapi/schemas.yaml                       (+ LlmoSheetRowPatch, LlmoSheetRowPatchResult)
package.json                                    (+ exceljs 4.4.0)
test/controllers/llmo/llmo-sheet-write.test.js  (new — 19 cases)
test/controllers/llmo/llmo.test.js              (+ patchLlmoDataRow describe block — 5 cases)
test/routes/index.test.js                       (+ snapshot updates)
```


jjenscodee (approved):

MysticatBot-Dev (changes requested):
Hey @dzehnder,

## Strengths

- **Clean separation of concerns** - `llmo-sheet-write.js` isolates the workbook read-modify-write logic from the controller glue, making it independently testable and keeping the already-large `llmo.js` from growing more complex internally.
- **Thorough error taxonomy** - The code distinguishes workbook-not-found (404), worksheet-not-found (404), unknown-columns (400), no-match (404), and ambiguous-match (409). Each gets a statusCode that the controller maps cleanly to HTTP responses with actionable messages.
- **Dependency injection for testability** - The `deps` parameter pattern (SharePoint client, ExcelJS, fetch, adminKey) makes the function trivially testable. The happy-path test actually round-trips through real ExcelJS workbook buffers and verifies the specific cell value changed.
- **Correct trust boundary reasoning** - Routing writes through the API rather than direct UI-to-SharePoint access is the right architectural call, and the PR description articulates why clearly.
- **Good test breadth** - 19 test cases for the core module plus 8 handler-level tests covering every error code path, including the esmock fallback for SharePoint client creation.

## Issues

### Important (Should Fix)

**1. No validation of `dataSource`/`sheetType` path parameters for path traversal**

`src/controllers/llmo/llmo-sheet-write.js` - `buildSharePointPath` and `buildPublishPath`

Both `dataSource` and `sheetType` are URL path parameters concatenated directly into file paths. A value like `../../other-tenant/secrets` would produce `/sites/elmo-ui-data/folder/../../other-tenant/secrets.xlsx`. While the Express-style router likely strips slashes and SharePoint may reject traversal sequences, the HLX admin publish URL (`admin.hlx.page`) is less certain to sanitize. Since this is a write endpoint (unlike the existing read-only `queryFiles` which follows the same pattern), the blast radius of traversal is higher.

Fix: Add a guard in the controller that rejects `dataSource` and `sheetType` values containing `/`, `\`, or `..` with a 400 - a simple regex like `/^[a-zA-Z0-9_-]+$/` would suffice.

**2. The 500 fallback in `patchLlmoDataRow` exposes raw internal error messages**

`src/controllers/llmo/llmo.js` - the catch block's final fallback:
```javascript
return internalServerError(cleanupHeaderValue(error.message));
```

If `createSharePointClient` or `getDocumentContent()` throws (e.g., missing env vars, network errors), the message will contain internal details (SDK errors, connection info). The error is already logged with `log.error` on the preceding line.

Fix: Return a generic message for the 500 case: `internalServerError('Failed to patch sheet row')`.

**3. `values` entries are not validated as strings despite the OpenAPI schema declaring `additionalProperties: { type: string }`**

`src/controllers/llmo/llmo-sheet-write.js` - the `Object.entries(values).forEach` block

If a caller sends `{ "deleted": true }` (boolean) instead of `{ "deleted": "true" }` (string), ExcelJS writes a boolean cell value. When the XLSX is projected to JSON via the Helix CDN (which expects string cell values for .json sheets), the result may differ from expectations. The `match` logic handles this with `String(expected)` coercion, but `values` are written verbatim.

Fix: Add type validation in `validateSheetRowPatch` that all values in `match` and `values` are strings, matching the OpenAPI contract.

### Minor (Nice to Have)

**4. `publishToHlx` runs preview and live sequentially when they are independent**

`src/controllers/llmo/llmo-sheet-write.js` - the `for...of` loop with `await`

The two publish calls are independent and could use `Promise.allSettled` to halve publish latency. Low priority given the stated low-frequency use case.

**5. `findRowMatching` will produce confusing 404s for formula or rich-text cells**

`src/controllers/llmo/llmo-sheet-write.js` - line where `String(cell.value)` is called

ExcelJS returns rich-text cells as objects (`{ richText: [...] }`) and formula cells as `{ formula: '...', result: '...' }`). `String()` on these produces `[object Object]`, which would never match the caller's plain-string expectation, resulting in a misleading 404. For the current use case (Strategic Recommendations sheets with simple string cells), this is unlikely but worth a defensive note.

**6. No handler-level test exercises the route variant without `sheetType`**

`test/controllers/llmo/llmo.test.js` - all tests use `baseParams` which always includes `sheetType: 'strategic-recommendations'`

The route `PATCH /sites/:siteId/llmo/data/:dataSource/row` leaves `sheetType` as `undefined` in `context.params`. The guard in `buildSharePointPath` (`if (sheetType)`) handles this correctly, but a test proving this would protect against regressions.

## Recommendations

- Add a workbook buffer size check after `getDocumentContent()` - if it exceeds a configured threshold (e.g., 50MB), return 400/413 rather than attempting to parse. This protects the Lambda from OOM on unexpectedly large workbooks.
- Cap `Object.keys(match).length` and `Object.keys(values).length` to a reasonable upper bound (e.g., 20-30 columns) in `validateSheetRowPatch` to prevent unbounded iteration in the row-matching loop.
- Consider including a `published: true/false` field in the success response so the caller has observability into whether the HLX CDN invalidation completed.

## Assessment

Ready to merge? With fixes

The design is architecturally sound: proper auth gating, clean separation of concerns, thorough error handling, and good test coverage. The three Important findings are about defensive hardening (path validation, error message sanitization, type contract enforcement) rather than correctness bugs. None are likely to cause incidents at the current scale and usage patterns, but they represent gaps worth closing before broader adoption of this write endpoint pattern. The path traversal guard (finding 1) and error message sanitization (finding 2) are the most impactful and lowest-effort to address.

## Next Steps

1. Add path-parameter validation for `dataSource` and `sheetType` (reject traversal characters).
2. Replace the raw error message in the 500 response with a generic string.
3. Validate that `match` and `values` entries are all strings.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 8m 13s | <b>Cost:</b> $3.87 | <b>Commit:</b> <code>a878d924e6e0b9a9c812b79d2a532b6f46449977</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

dzehnder:
Thanks @MysticatBot — addressed all three **Important** items plus the two **Minor** ones that were cheap to land. Pushed in `f5fdcd51`.

### Important
1. **Path traversal** — `isSafePathSegment` (regex `^[A-Za-z0-9_-]+$`) gates `dataSource` and `sheetType` in `patchLlmoDataRow` before any path concatenation. A value like `../../other-tenant/secrets` now returns 400 and never reaches the SharePoint client. Tests cover both the `dataSource` and `sheetType` rejection paths.
2. **500 leakage** — Unexpected errors now respond `internalServerError('Failed to patch sheet row')`. Full detail is still logged with `log.error` (line `1189`); only the response body changed. Test asserts the SharePoint-secret error message does **not** appear in the response.
3. **String enforcement on `match` / `values`** — `validateSheetRowPatch` now rejects any non-string entry with a 400 that names the offending key (e.g. `values.deleted must be a string`). Matches the OpenAPI `additionalProperties: { type: string }` contract at runtime.

### Minor
- **(5) Rich-text / formula cells** — Added `cellValueAsString` which handles `{ richText: [...] }`, `{ formula, result }`, and `{ text, hyperlink }` shapes, plus null/undefined and the `String()` fallback. Used in `findRowMatching` so a `match` of `"first prompt"` still hits a cell stored as rich text. Helper is exported and unit-tested directly (rich-text round-tripping through ExcelJS write/load doesn't reliably preserve the object shape, so indirect coverage isn't dependable).
- **(6) No-`sheetType` route variant** — Added a handler test that hits `PATCH /sites/:siteId/llmo/data/:dataSource/row` (no `sheetType`) and asserts the response has `sheetType: null` and the path built for `patchSheetRow` skips the type segment.

### Deferred (not in this PR)
- **(4) Parallel `preview` + `live` publish** — Left sequential. Low-frequency endpoint, and serializing matches the pattern in `spacecat-audit-worker/src/utils/report-uploader.js` for easier comparison. Happy to flip to `Promise.allSettled` in a follow-up.
- **Workbook size cap / column-count cap** — Real exposure is low (workbooks are small, callers send 1–4 columns), but I agree both are sensible. Tracking as a follow-up so this PR stays scoped.
- **`published: true/false` in the response** — Worth doing but ties into deciding whether HLX failure should be 502 vs 200; will pair with that decision.

Local checks: `npm test` for affected suites green (430 passing on the two LLMO test files), `npm run lint` clean, `npm run docs:lint` valid. Patch coverage back to 100% on both touched source files.

MysticatBot-Dev (approved):
Hey @dzehnder,

## Strengths

- **All prior Important findings addressed cleanly.** Path traversal protection via `isSafePathSegment` (`llmo-sheet-write.js:23-24`), generic 500 response (`llmo.js:1219`), and string-type enforcement in `validateSheetRowPatch` (`llmo-sheet-write.js:139-145`) are all properly implemented. The no-sheetType route variant and rich-text/formula cell matching (prior Minor items) are also resolved with dedicated tests and the `cellValueAsString` helper.
- **Correct trust boundary decision** - routing writes through the API rather than giving the UI direct SharePoint access maintains the same trust model as the existing read path. No new auth surface introduced.
- **Clean module separation** - `llmo-sheet-write.js` is self-contained with no reverse dependencies on the controller, keeping the already-large `llmo.js` from growing in internal complexity.
- **Dependency injection via `deps` parameter** makes the module trivially testable without heavy mocking infrastructure. Consistent with patterns elsewhere in the codebase.
- **Thorough error taxonomy** - 404 for missing workbook/worksheet/row, 400 for unknown columns and type mismatches, 409 for ambiguous match. Each message tells the caller exactly what to fix.
- **Strong test suite** - 30+ test cases including a round-trip verification that reloads the uploaded buffer with ExcelJS and asserts the specific cell value changed. Error mapping, path traversal rejection, generic 500 non-leaking, and the no-sheetType variant are all covered.
- **No SSRF vector in publish path** - the HLX publish URL is constructed from validated path segments and server-side constants. No user-controlled URL component reaches `fetch` unvalidated.

## Issues

### Minor (Nice to Have)

**1. Deprecated transitive dependencies in the `exceljs` tree**

`package-lock.json` - `exceljs@4.4.0` pulls in `fstream@1.0.12` (deprecated, had CVE-2019-13173), `inflight@1.0.6` (deprecated, memory leak), `glob@7.2.3` (deprecated), and `rimraf@2.7.1` (deprecated) via `unzipper@0.10.14`.

Risk is low in this context: XLSX files originate from authenticated, tenant-scoped SharePoint, not arbitrary user uploads. The vulnerable code paths in `fstream` relate to tar extraction with crafted filenames, which would require a malicious XLSX already planted in the customer's SharePoint. No immediate action required, but worth tracking for a future `exceljs` release that drops `unzipper`.

## Recommendations

- Consider adding a workbook buffer size check after `getDocumentContent()` in `patchSheetRow` - if the buffer exceeds a reasonable threshold (e.g., 50MB), return 413 rather than attempting to parse. This protects the Lambda from OOM on unexpectedly large workbooks.
- Consider pinning `exceljs` to exact version (`"4.4.0"` instead of `"^4.4.0"`) to match the pinning style used by other dependencies in the same package.json (e.g., `"csv": "6.5.1"`). Deliberate upgrades give more control over supply chain changes in a dependency with many transitive packages.
- Consider adding timing logs around the `getDocumentContent()` and `uploadRawDocument()` calls - these external I/O operations will dominate latency and the timing would be valuable during future investigations.

## Assessment

Ready to merge? Yes

All prior review findings have been properly addressed. The design is architecturally sound: proper auth gating, clean separation of concerns, tight input validation (allowlist regex on path segments, string-type enforcement on body fields), thorough error handling with no internal detail leakage, and comprehensive test coverage. The endpoint is well-scoped for its stated purpose (low-frequency UI soft-delete) and the concurrency limitation is documented. Ship it.

jjenscodee:
Is it possible to support updating multiple rows?

dzehnder:
@jjenscodee Yes, easy to add — and actually a strict win on this endpoint shape. The expensive parts are the SharePoint download/upload (~1–2s each) and the HLX `preview`+`live` republish; the row mutation itself is microseconds. Batching N updates into one call collapses N×(download + upload + republish) into 1× — much cheaper, and it also narrows the read-modify-write race window vs. N sequential single-row calls.

Repo conventions already favour this shape (CLAUDE.md "Bulk operations: prefer bulk endpoints accepting arrays" + existing `POST /v2/orgs/.../prompts/delete`).

Proposed body for a batched variant:

```json
{
  "updates": [
    { "sheet": "Semrush",            "match": { "topic_id": "111", "prompt": "…" }, "values": { "deleted": "true" } },
    { "sheet": "Semrush",            "match": { "topic_id": "222", "prompt": "…" }, "values": { "deleted": "true" } },
    { "sheet": "Citation Attempt",   "match": { "source_url": "…" },               "values": { "deleted": "true" } }
  ]
}
```

Semantics:
- Each entry independently validated and matched (same 400 / 404 / 409 rules per entry).
- All applied to one in-memory workbook; one upload; one republish.
- All-or-nothing: if any entry fails to match uniquely, no rows are written. Response lists each entry's `rowNumber` / `updated`, or which entries failed and why.

Two options for the route shape:

1. **Reuse the existing endpoint, accept either body** (`{ sheet, match, values }` *or* `{ updates: [...] }`). Pro: one path, gradual UI migration. Con: union body type is mildly less clean.
2. **Add `PATCH .../row` (singular) and `PATCH .../rows` (batch)** as siblings. Pro: cleaner contract. Con: two paths to authorize, doc, and test.

I'd lean (1). Either way I'd cap `updates.length` (e.g. 100) so a runaway batch can't time out the Lambda.

**Do you want this in this PR or a follow-up?** The UI in #1804 deletes prompts one at a time today, so single-row covers the immediate need. If a bulk-delete affordance is coming soon I'm happy to add it here — adds maybe ~50 lines + tests.

dzehnder:
@jjenscodee Batch shape pushed in `cca0f5cd`.

**Body:** either the existing single-update shape, or `{ updates: [{ sheet, match, values }, ...] }` (1..100 entries).

**Behaviour:**
- All-or-nothing — every entry must validate AND match exactly one row before any cells are written. First failing entry aborts the batch with `updates[i]: ...` in the message so you can pinpoint it.
- One SharePoint download + one upload + one preview/live publish per call regardless of batch size.
- Worksheet header maps cached per-sheet across entries (one header scan per worksheet, not per entry).
- Capped at 100 to bound Lambda compute / memory.

**Response:**
- Single-update body → unchanged top-level `{ siteId, sheetType, dataSource, sheet, rowNumber, updated }` (back-compat).
- Batch body → `{ siteId, sheetType, dataSource, updates: [{ sheet, rowNumber, updated }] }` in input order.

**OpenAPI:** request + response schemas updated to `oneOf` and visible in the regenerated docs.

**Local verification:** 2811 affected unit tests pass (incl. happy-path batch round-tripping a real ExcelJS workbook with 3 updates across 2 worksheets, all-or-nothing semantics, ambiguous/unknown-column error messages with the right `updates[i]` index). Patch line coverage 100% on both touched source files. Lint + OpenAPI lint clean.

jjenscodee:
lgtm, thank you

PR #2474, opened 2026-05-22
fix(fixes): replace N+1 getSuggestions loop with batch method (SITES-45307)

mimchome:
[SITES-45307](https://jira.corp.adobe.com/browse/SITES-45307)

## Summary

- Replace per-entity `getSuggestions()` loop (2N+1 PostgREST calls) with `getAllFixesWithSuggestionsByOpportunityId` (3 calls total via batch)
- Dependency: [spacecat-shared#1619](https://github.com/adobe/spacecat-shared/pull/1619) -- merged, published as `@adobe/spacecat-shared-data-access@3.71.1`
- Dep bump to 3.71.1 picked up via merge from main

## Test plan

- [x] 102 fixes controller tests passing
- [ ] Verify GET /sites/:siteId/opportunities/:opportunityId/fixes returns suggestions on dev


mimchome:
Replaced by a clean branch to remove accidentally committed .http files from git history.

MysticatBot-Dev (changes requested):
Hey @mimchome,

Good performance fix - eliminating the N+1 `getSuggestions()` loop in favor of the batch method is the right call, and the consistency with the `fixCreatedDate` path makes the controller easier to reason about. A couple of issues need addressing before this can merge.

## Strengths

- **Core optimization is well-targeted** (`src/controllers/fixes.js:151-163`): Replacing unbounded per-entity `getSuggestions()` calls with `getAllFixesWithSuggestionsByOpportunityId` reduces PostgREST round-trips from 2N+1 to 3. Meaningful latency improvement on a user-facing path.
- **Structural consistency**: Both the `fixCreatedDate` path and the non-filtered path now use the same data access pattern, reducing cognitive overhead and divergence bugs.
- **Good test coverage**: New unit test for the empty-array case, updated existing tests to match the new contract, and integration seed data expanded to validate junction table queries end-to-end through real PostgREST.
- **`test/e2e/local.http` follows secure patterns**: Uses `{{$dotenv ...}}` placeholders exclusively. This is the correct pattern.

## Issues

### Critical (Must Fix)

**1. Hardcoded credentials committed to repository** - `.rest/test-autofix.http:22,28`

The file contains a plaintext API key (`aab2c470-7708-013c-664a-7ea0e1f7f5af`) and a full IMS staging access token (JWT with broad scopes including `session`, `aem.assets.author`, `service_principals.write`, etc.). Once merged, these persist in git history permanently.

This is especially notable because `test/e2e/local.http` in the same PR demonstrates the correct approach using `{{$dotenv ...}}` references.

**How to fix**: Either remove `.rest/test-autofix.http` from the PR entirely, or convert it to use `{{$dotenv API_KEY}}` / `{{$dotenv IMS_TOKEN}}` placeholders matching the pattern in `test/e2e/local.http`. The committed API key should be rotated regardless. Consider adding `.rest/` to `.gitignore` if these are personal developer scratch files.

### Important (Should Fix)

**2. Behavioral change: empty results bypass ownership check** - `src/controllers/fixes.js:155-157`

The early return `if (fixEntitiesWithSuggestions.length === 0) return ok([])` skips `checkOwnership` entirely. Previously, the empty-array case still called `checkOwnership(undefined, opportunityId, siteId, ...)`, which verified the opportunity exists and belongs to the site (returning 404 if not). Now `GET /sites/:siteId/opportunities/:opportunityId/fixes` returns `200 []` even when `opportunityId` does not belong to `siteId`.

This is mitigated by the fact that the `fixCreatedDate` path already has the same early-return pattern, so the PR is at least consistent. But this should be a conscious decision. If intentional, a brief comment or PR description note would help. If not, move the early return after an ownership check, or call `checkOwnership(null, opportunityId, siteId, this.#Opportunity)` before the early return.

**3. Version specifier changed from exact pin to caret range** - `package.json:88`

`@adobe/spacecat-shared-data-access` changed from `"3.71.2"` to `"^3.71.2"`. Every other `@adobe/spacecat-shared-*` dependency (15 of 16) uses exact pinning. The caret gives no operational benefit in a deployed service with a lockfile, but widens the window for non-deterministic installs if the lockfile is regenerated.

**How to fix**: Revert to exact pinning to match the established pattern. If a newer version is required, pin to that specific version.

### Minor (Nice to Have)

**4. Unit test happy path does not assert suggestions appear in response** - `test/controllers/fixes.test.js:149-157`

The "can get all fixes" test stubs `suggestions: []` for every fix entity. Since the whole point of this PR is ensuring suggestions are attached, a test with non-empty suggestions asserting they appear in the serialized output would catch a regression where the attachment fails silently.

## Recommendations

- Extract the `fixEntitiesWithSuggestions.map(item => { fixEntity._suggestions = item.suggestions; ... })` block into a small helper. This pattern now appears identically in both the `fixCreatedDate` path and the non-filtered path - a helper would keep them in sync and reduce duplication.
- Rotate the committed API key (`aab2c470-...`) regardless of environment. Treat credential exposure as a standard incident.

## Assessment

Ready to merge? No - requires fixes.

The hardcoded credentials must be removed before merge. The dependency pin should be reverted. The ownership-bypass behavioral change should be explicitly acknowledged as intentional (consistent with the adjacent code path) or fixed.

## Next Steps

1. Remove or convert `.rest/test-autofix.http` to use dotenv placeholders, and rotate the exposed credentials.
2. Revert `@adobe/spacecat-shared-data-access` to exact pinning.
3. Decide on the ownership-check bypass - document the decision or add the check back.

PR #2475, opened 2026-05-22
fix(it): force HTTP/1.1 early in IT harness to fix it-postgres timeouts

akshaymagapu:
## Problem

The `ci/it-postgres` job intermittently hangs after `491 passing` prints and gets cancelled at the 15-minute step timeout (e.g. [run 26282298612](https://github.com/adobe/spacecat-api-service/actions/runs/26282298612), [26229325483](https://github.com/adobe/spacecat-api-service/actions/runs/26229325483)). Even on passing runs, the job spends ~4 minutes idle after the last test before mocha exits.

The cause: ~9 spacecat-shared libraries each construct a module-scope `@adobe/fetch` context at import time. By default on Node ≥19 these contexts hold HTTP/1.1 keep-alive Agents with pooled sockets that keep the Node event loop alive after `afterAll` completes. In Lambda warm starts that's by design; in tests it's a leak.

## Why doesn't `test/setup-env.js` already fix this?

[`test/setup-env.js:22`](https://github.com/adobe/spacecat-api-service/blob/main/test/setup-env.js#L22) sets `process.env.HELIX_FETCH_FORCE_HTTP1 = 'true'` and is wired via `package.json` `mocha.require`. So in principle the env var should already be set before tests run.

The catch: **mocha CLI `--require` runs before `package.json` `mocha.require`**. The IT command from the shared `mysticat-ci` workflow is:

```
npx mocha --require test/it/postgres/harness.js --timeout 30000 'test/it/postgres/**/*.test.js'
```

So `harness.js` loads first → its static imports trigger the chain that loads `spacecat-shared-data-access`, `spacecat-shared-rum-api-client`, and the other ~7 fetch-context-holding libraries → each constructs its keep-alive context with `HELIX_FETCH_FORCE_HTTP1` still undefined → pools created. Then `setup-env.js` finally runs and sets the env var, but it's too late — the contexts already exist and read the env only at construction time.

Verified in the diagnostic run (closed PR #2476): `[proof] HELIX_FETCH_FORCE_HTTP1 was: undefined` was logged at the top of `harness.js`, *before* `Forcing HTTP/1.1 for Adobe Fetch` from `setup-env.js`.

## Solution

Set `HELIX_FETCH_FORCE_HTTP1=1` at the top of `harness.js` — the earliest deterministic load point in the IT process.

## How it fixes

Each shared library's fetch context reads the env var at its own module-load time:

```js
const fetchContext = process.env.HELIX_FETCH_FORCE_HTTP1 ? h1() : h2();
```

When set, every library chooses `h1()` instead of `h2()`. `h1()` on Node ≥19 builds an HTTP agent with `keepAlive: false` — sockets are destroyed when the request ends, the pool stays empty, no handles persist after tests complete, Node exits cleanly.

One additional constraint: ES static imports hoist before code in the same file, so simply setting `process.env` at the top doesn't help unless the imports that depend on it are made **dynamic**. Hence the `await import(...)` pattern for everything `harness.js` pulls in.

`test/setup-env.js` is left alone — it still works correctly for unit tests, which are invoked via `npm test` (where mocha's CLI `--require` order isn't an issue because there's no CLI `--require`).

## Result

`ci/it-postgres` job total time: **~8m → ~4m**. Zero handles pin the loop after teardown; process exits naturally within 1s of the last test. No more 15-minute cancellations.

## Scope

- One file changed: `test/it/postgres/harness.js`.
- Test-only. `harness.js` is loaded only by mocha for the IT suite; not bundled into the Lambda. No production code or library changes. No new dependencies.

### DIFF
--- a/test/it/postgres/harness.js
+++ b/test/it/postgres/harness.js
@@ -18,13 +18,21 @@
  *
  * Starts Docker Compose (Postgres + PostgREST) and the dev server once
  * before all test files, and tears both down after all test files complete.
+ *
+ * Force HTTP/1.1 (no keep-alive on Node ≥19) for every shared library's
+ * @adobe/fetch context. Static ES imports hoist before code, and mocha CLI
+ * --require this file before package.json `mocha.require` runs, so this is
+ * the earliest point in process startup where we can set the env var before
+ * any module-level fetch context is constructed. Subsequent imports must be
+ * dynamic so they evaluate AFTER this line.
  */
+process.env.HELIX_FETCH_FORCE_HTTP1 = '1';
 
-import { initAuth, createAllTokens } from '../shared/auth.js';
-import { buildEnv } from '../env.js';
-import { startServer, stopServer } from '../server.js';
-import { createHttpClient } from '../shared/http-client.js';
-import { startPostgres, stopPostgres } from './setup.js';
+const { initAuth, createAllTokens } = await import('../shared/auth.js');
+const { buildEnv } = await import('../env.js');
+const { startServer, stopServer } = await import('../server.js');
+const { createHttpClient } = await import('../shared/http-client.js');
+const { startPostgres, stopPostgres } = await import('./setup.js');
 
 /** Shared state populated during beforeAll, consumed by test files. */
 export const ctx = {};


akshaymagapu:
Closing in favor of the proper fix: **[adobe/mysticat-ci#15](https://github.com/adobe/mysticat-ci/pull/15)**.

That PR sets `HELIX_FETCH_FORCE_HTTP1=1` on the workflow-step env for `Run PostgreSQL Integration Tests`, which propagates to every consumer of the shared workflow with zero consumer-repo changes. No `harness.js` modifications, no dynamic-import gymnastics, no test-only knob duplicated across repos — just the right environment in the right place.

This PR's harness.js approach was a workaround for not having control over the shared workflow command. With the workflow fix, the harness can stay exactly as it is on `main`.

PR #2476, opened 2026-05-22
chore(it): proof — verify resetDataAccessConnections drains @adobe/fetch leak (DO NOT MERGE)

akshaymagapu:
## Purpose

Prove in CI that calling \`fetchContext.reset()\` actually drains the 270 \`@adobe/fetch\` handles identified by the earlier diagnostic (#2472), **before** we land the real fix as a shared-library PR.

## How

- \`patches/@adobe+spacecat-shared-data-access+3.69.0.patch\` adds the \`resetDataAccessConnections\` export locally (mirrors adobe/spacecat-shared#1621). Applied via \`postinstall: patch-package\`.
- \`test/it/postgres/harness.js\` afterAll:
  1. Counts active handles BEFORE reset.
  2. Calls \`resetDataAccessConnections()\`.
  3. Counts again AFTER reset + runs \`why-is-node-running\`.
  4. Tears down Postgres.
  5. 5s backstop force-exit so the job doesn't sit at the 15-min timeout if the fix doesn't fully drain.

## Expected outcome

\`\`\`
[proof] handles BEFORE reset: ~270  [PipeWrap, TCPSocketWrap, ...]
[proof] handles AFTER reset:  ~3-4  [PipeWrap stdio only]
[proof] wirn — handles still keeping loop alive: (empty)
\`\`\`

Process exits naturally within ~1-2s (no backstop firing). Total job time drops from ~8m to ~4m.

If the backstop fires or the AFTER count is still high, we learn what else is leaking and adjust.

## After this PR

- If the proof works as expected → close this PR, let adobe/spacecat-shared#1621 + #2475 land normally with high confidence.
- If it doesn't → use the wirn output to find the next leak source, update the fix accordingly.


akshaymagapu:
Closing — proof complete.

## What this PR established

Two CI runs with progressively more invasive diagnostics produced the conclusive evidence:

**Run 1** (patch-package + `resetDataAccessConnections()`): wirn count 270 → 177. Showed the data-access fix was *partial* — remaining handles traced to `@adobe/spacecat-shared-rum-api-client`'s separate `@adobe/fetch` context. Surveying spacecat-shared found ~9 module-scope fetch contexts across 8 libraries; per-library teardown exports were untenable.

**Run 2** (`HELIX_FETCH_FORCE_HTTP1=1` set at top of harness.js + dynamic imports): job total dropped from ~8m to **4m 9s**, `process.getActiveResourcesInfo()` reported 6 handles (stdio + the backstop timer itself, none pinning the loop), backstop force-exit never fired, process exited naturally 1s after the last log line. Conclusion: setting the env var before any `@adobe/fetch` context is constructed causes all ~9 shared libraries to choose `h1()` mode (no keep-alive on Node ≥19), eliminating the leak across all of them with a single one-line knob.

Also incidentally proved that **mocha CLI `--require` runs before `package.json` `mocha.require`** — explains why the existing `HELIX_FETCH_FORCE_HTTP1='true'` in `test/setup-env.js` was being applied too late.

## Where the real fix lives

[#2475](https://github.com/adobe/spacecat-api-service/pull/2475). Single file change to `harness.js`. No library modifications, no new dependencies, no patch-package.

PR #2478, opened 2026-05-22
feat: add site-scoped preflight REST endpoints | SITES-44686

GeezerPelletier:
## Summary

- Adds three new site-scoped REST endpoints for the Preflight feature:
  - `POST /sites/:siteId/preflights` — creates a `Preflight` record + `AsyncJob`, fires Mysticat analyze, returns 202 with `Location` header
  - `GET /sites/:siteId/preflights` — lists all preflights for a site (optional `?url` filter)
  - `GET /sites/:siteId/preflights/:preflightId` — returns full preflight detail
- Adds `src/dto/preflight.js` with `PreflightDto.toJSON` (list view) and `toDetailJSON` (detail view)
- Wires `preflightId` UUID validation in `src/index.js` (alongside existing `jobId` validation)
- Routes registered in `src/routes/index.js` and capability requirements in `src/routes/required-capabilities.js`
- Removes internal `/preflight/beta/*` endpoints (replaced by the new site-scoped endpoints)
- Existing `/preflight/jobs` endpoint is **not** modified
- Adds [ADR-003](docs/decisions/003-preflight-entity-design.md) — extracts the Preflight entity design into its own decision record (per retrospective review feedback on #2377); ADR-002 updated to reference it

## Mysticat call shape

The POST body sent to mystique's `/v1/preflight/analyze` is intentionally minimal:

```json
{
  "site_id": "<uuid>",
  "url": "<page-url>",
  "scan_id": "<async-job-id>",
  "persist": true
}
```

No `audits` or `mode`/`step` fields are sent:
- **No `audits`** — per the [Preflight Opportunity Enablement ADR](https://github.com/adobe/mysticat-architecture/blob/main/platform/decisions/preflight-opportunity-enablement.md), mystique resolves the goal set itself via three-gate filtering (tier enablement → `run_in_preflight=true` fact registry → per-site `goal_overrides`). Caller-supplied audit lists are no longer required or used.
- **No `mode`/`step`** — per [ADR-002](docs/decisions/002-preflight-api-rest-redesign.md) ("Key changes"): *"`step` is removed. Mysticat's agent always performs both identify and suggest as a single flow."*

`scan_id` is the `AsyncJob.id` minted on this side before the call. mystique writes resulting facts tagged with that scan_id; the projector picks them up via SNS and flips `Preflight.status` / `AsyncJob.status` to terminal state ([mysticat-projector-service #196](https://github.com/adobe/mysticat-projector-service/pull/196)).

## Design decisions

- New endpoints are undocumented (not in the public OpenAPI spec) intentionally — they will remain that way until the feature is ready for GA
- Access control uses `AccessControlUtil.fromContext(ctx)` with `hasAccess` check against the resolved site
- Entitlement check via `TierClient.createForSite` → `checkValidEntitlement('preflight')` gates the create endpoint
- Cross-site isolation: `getPreflightById` returns 404 (not 403) when `preflight.getSiteId() !== siteId` to prevent site enumeration
- `createdBy` is derived server-side from the IMS profile, never accepted from the client
- Errors use the `{ errorCode, message }` contract via `preflightError(...)` across all new endpoints; `errorCode` is the stable machine-readable identifier
- Promise-token resolution uses the existing `resolvePromiseToken` helper (cookie + IMS fallback), matching the established pattern in `createPreflightJob`. The `x-promise-token` header support that #2480 was building toward was reverted by #2544; if/when that work re-lands, the team behind it owns updating this endpoint's auth path and corresponding tests.

## Test plan

- [x] 64 unit tests in `test/controllers/preflight.test.js` (all passing)
- [x] Full coverage of happy path and error paths for all three new endpoints
- [x] Access denied (403), site not found (404), preflight not found (404), entitlement not enabled (403), invalid URL (400), upstream timeout / 5xx (502), internal errors (500)
- [x] ESLint clean on all changed files


GeezerPelletier:
**Note on CI failures:** The failing checks in this PR are pre-existing failures on `main` and are unrelated to these changes. Verified locally — our diff against `main` is exactly the 8 files in this PR (preflight controller, DTO, routes, index, tests, and the two ADR docs). The `PlgOnboardingController` failures came in via `88edf952` on main, and the original ~51 failures (organizations, configurations, AI visibility, etc.) pre-date this branch entirely. All preflight tests pass cleanly.

src/controllers/preflight.js
```diff
@@ -357,256 +365,243 @@ function PreflightController(ctx, log, env) {
...
+  };
+
+  /**
+   * Returns all preflights for a site, optionally filtered by URL.
+   * @param {Object} context - The request context
+   * @returns {Promise<Object>} 200 OK with array of preflights
+   */
+  const getAllPreflights = async (context) => {
+    const siteId = context.params?.siteId;
+    const urlFilter = context.params?.url;
```
clotton: 🚨 **`context.params?.url` is always `undefined` for this route.**

`context.params` is populated by `matchPath` from path segments — `GET /sites/:siteId/preflights` has no `:url` segment, so query strings never land here. The test at `preflight.test.js:1418` doesn't catch this because it injects `params: { url }` directly.

Query strings need to be parsed from `context.invocation.event.rawQueryString` — see `feature-flags.js:40` or `brands.js:129` for the established pattern.

Suggested fix:
```js
function getQueryParams(context) {
  const raw = context.invocation?.event?.rawQueryString;
  if (!raw) return {};
  const out = {};
  raw.split('&').forEach((p) => {
    const [k, v] = p.split('=');
    if (k) out[decodeURIComponent(k)] = v !== undefined ? decodeURIComponent(v) : '';
  });
  return out;
}

// in getAllPreflights:
const urlFilter = getQueryParams(context).url;
```

And add a test that constructs a real `rawQueryString` so this can't regress.

src/controllers/preflight.js
```diff
@@ -357,256 +365,243 @@ function PreflightController(ctx, log, env) {
...
+      );
+    } catch (mysticatError) {
+      log.error(`Mysticat analyze failed: ${mysticatError.message}`);
+      preflight.setStatus('FAILED');
+      preflight.setError({ code: 'MYSTICAT_ERROR', message: mysticatError.message });
+      preflight.setEndedAt(new Date().toISOString());
+      await preflight.save().catch(() => {});
+      asyncJob.setStatus(AsyncJob.Status.FAILED);
+      await asyncJob.save().catch(() => {});
+      return preflightError('PREFLIGHT_UPSTREAM_ERROR', `Mysticat analyze failed: ${mysticatError.message}`, 502);
```
clotton: `mysticatError.message` may include the response body from Mysticat (see line 363 in the unchanged helper), which could contain stack traces, internal IDs, or other internal info we don't want to leak to API clients. Suggest logging the full error and returning a generic upstream message:

```js
log.error(`Mysticat analyze failed: ${mysticatError.message}`);
return preflightError('PREFLIGHT_UPSTREAM_ERROR', 'Upstream analyze service failed', 502);
```

src/controllers/preflight.js
```diff
@@ -357,256 +365,243 @@ function PreflightController(ctx, log, env) {
...
+        siteId,
+        asyncJobId: asyncJob.getId(),
+        url,
+        status: 'IN_PROGRESS',
+        createdBy,
+        startedAt: new Date().toISOString(),
+      });
+    } catch (e) {
+      log.error(`Failed to create Preflight: ${e.message}`);
+      await asyncJob.remove().catch(() => {});
```
clotton: `.catch(() => {})` silently swallows the rollback failure. If `asyncJob.remove()` fails here, the AsyncJob is orphaned and ops has no visibility. The TTL will eventually clean it up, but a `.catch((e) => log.warn(\`Failed to roll back AsyncJob ${asyncJob.getId()}: ${e.message}\`))` would surface drift.

src/controllers/preflight.js
```diff
@@ -357,256 +365,243 @@ function PreflightController(ctx, log, env) {
...
+        'suggest',
+        authorizationHeader,
+        preflightAudits,
+      );
+    } catch (mysticatError) {
+      log.error(`Mysticat analyze failed: ${mysticatError.message}`);
+      preflight.setStatus('FAILED');
+      preflight.setError({ code: 'MYSTICAT_ERROR', message: mysticatError.message });
+      preflight.setEndedAt(new Date().toISOString());
+      await preflight.save().catch(() => {});
```
clotton: Same silent-rollback concern at lines 512 and 514 — if either `preflight.save()` or `asyncJob.save()` fails while persisting the FAILED state, the DB ends up inconsistent and we never log it. A `.catch((e) => log.warn(...))` on each would help operability.

src/controllers/preflight.js
```diff
@@ -357,256 +365,243 @@ function PreflightController(ctx, log, env) {
...
-          jobId: job.getId(),
-          status: job.getStatus(),
-          createdAt: job.getCreatedAt(),
-          pollUrl: `https://spacecat.experiencecloud.live/api/${isDev ? 'ci' : 'v1'}`
-            + `/preflight/beta/jobs/${job.getId()}`,
-        });
-      }
+    try {
+      await callMysticatAnalyze(
+        env.MYSTIQUE_API_BASE_URL,
```
clotton, line 517: Two things on this call:

1. **No timeout on the fetch inside `callMysticatAnalyze`.** If Mysticat hangs, this request hangs until Lambda's invocation timeout. Suggest adding an `AbortController` with ~10-15s in `callMysticatAnalyze` so caller can surface upstream timeouts cleanly.

2. **`env.MYSTIQUE_API_BASE_URL` not validated.** If unset, the URL becomes `"undefined/v1/preflight/analyze"` and fetch fails with a confusing error. The constructor already validates `env` is a non-empty object — recommend adding a specific check for `MYSTIQUE_API_BASE_URL` (and any other required env keys) at controller construction so misconfiguration fails loud at startup, not silently mid-request.

src/controllers/preflight.js
```diff
@@ -357,256 +365,243 @@ function PreflightController(ctx, log, env) {
...
-            + `/preflight/beta/jobs/${job.getId()}`,
-        });
-      }
+    let preflight;
+    try {
+      preflight = await dataAccess.Preflight.create({
+        siteId,
+        asyncJobId: asyncJob.getId(),
+        url,
+        status: 'IN_PROGRESS',
```
clotton: Status string mismatch with line 469 — there it's `AsyncJob.Status.IN_PROGRESS` (typed enum), here it's the raw string `'IN_PROGRESS'`. If `Preflight` has its own `Status` enum, use it for consistency. If not, worth adding one to `spacecat-shared-data-access` so we don't end up with typos like `'IN_PROGESS'` later.

src/controllers/preflight.js
```diff
@@ -357,256 +365,243 @@ function PreflightController(ctx, log, env) {
...
 
-    const mysticatBaseUrl = (isDev && hasText(normalizedMystiqueUrl))
-      ? normalizedMystiqueUrl
-      : env.MYSTIQUE_API_BASE_URL;
+    // Validate the URL belongs to this site
+    let siteByUrl;
+    try {
+      siteByUrl = await dataAccess.Site.findByPreviewURL(previewBaseURL);
+    } catch (e) {
+      // URL doesn't match a known site — fall through to invalid request
```
clotton: Silent error swallow — if `findByPreviewURL` throws (vs. returns null), the subsequent `!siteByUrl` check handles it correctly, so the behavior is fine. But adding a `log.debug(\`findByPreviewURL failed for ${previewBaseURL}: ${e.message}\`)` would help ops spot if this starts firing unexpectedly.

clotton (changes requested):
Strong overall — the ADR is excellent, cross-site isolation is done right (404 not 403 in `getPreflightById`), `createdBy` is derived from the authenticated profile rather than client input, and `mode: 'suggest'` is correctly pinned server-side. One shippable bug and a few smaller concerns below.

## 🚨 Critical — `?url=` filter never works in production

[`src/controllers/preflight.js:532`](src/controllers/preflight.js#L532) reads `urlFilter` from `context.params?.url`. `context.params` is populated by `matchPath` from **path** segments — the route `GET /sites/:siteId/preflights` has no `:url` segment, so `params.url` is always `undefined`. Query strings live at `context.invocation.event.rawQueryString` and need to be parsed explicitly (see [`feature-flags.js:40-49`](src/controllers/feature-flags.js#L40) and [`brands.js:129`](src/controllers/brands.js#L129) for the pattern).

The existing test at [`preflight.test.js:1418`](test/controllers/preflight.test.js#L1418) doesn't catch this because it injects `params: { url }` directly, bypassing the routing layer.

**Suggested fix:** add a `getQueryParams(context)` helper, pull `url` from there, and add a test that constructs `context.invocation.event.rawQueryString = 'url=...'`.

## Concerns worth addressing

1. **Mysticat error messages leak to API clients** (line 515). The Mysticat response body (caught at 361-364 in the unchanged helper) can contain stack traces / internal IDs. Log the full error; return a generic upstream error to clients.
2. **Silent rollback failures** at lines 493, 512, 514 — `.catch(() => {})` hides DB inconsistencies from ops. Use `.catch((e) => log.warn(...))`.
3. **No timeout on Mysticat fetch + `env.MYSTIQUE_API_BASE_URL` not validated** (line 499 / call site). If Mysticat hangs, this request hangs until Lambda timeout. If the env var is unset, fetch URL becomes `"undefined/v1/preflight/analyze"`. Add an `AbortController` (~10-15s) and validate required env vars at controller construction.
4. **Status enum inconsistency** — `AsyncJob.Status.IN_PROGRESS` (line 469) vs raw string `'IN_PROGRESS'` (line 487).
5. **Silent `findByPreviewURL` error swallow** at line 405. Behavior is correct (falls through to `!siteByUrl`), but a `log.debug` would give ops visibility if this starts firing.
6. **Synchronous HEAD to customer URL** in `checkEnableAuthentication` (unchanged helper) — pre-existing pattern, also in legacy `createPreflightJob`. Not a regression, but worth a follow-up to bound it.

## Strengths

- ADR-003 is genuinely well-written — clearly states why a dedicated `Preflight` entity beats extending `AsyncJob`, and was pre-aligned with @ekdogan. This is what an ADR should look like.
- Authorization ordering (`!site → !hasAccess → URL validation`) is correct — never disclose data before access verification.
- `preflightId` UUID validation centralized in the router ([`index.js:356`](src/index.js#L356)) follows the existing `jobId`/`siteId` pattern.
- Required capabilities registered for all three new routes with the right read/write semantics.
- 61 tests with broad coverage of error paths (access denied, not found, internal errors, entitlement gate).

Happy to discuss any of these — the URL-filter fix is the only one I'd consider a hard block.

src/controllers/preflight.js
```diff
@@ -357,256 +373,255 @@ function PreflightController(ctx, log, env) {
...
 
-    if (![AUDIT_STEP_IDENTIFY, AUDIT_STEP_SUGGEST].includes(data?.step?.toLowerCase())) {
-      return badRequest(
-        `Invalid request: step must be either ${AUDIT_STEP_IDENTIFY} or ${AUDIT_STEP_SUGGEST}`,
-      );
-    }
-
-    const step = data.step.toLowerCase();
     const { url } = data;
+    const previewBaseURL = `${new URL(url).protocol}//${new URL(url).hostname}`;
```
johobot: Nit: `new URL(url)` is parsed twice here. Parse once:
```js
const { protocol, hostname } = new URL(url);
const previewBaseURL = `${protocol}//${hostname}`;
```

src/controllers/preflight.js
```diff
@@ -357,256 +373,255 @@ function PreflightController(ctx, log, env) {
...
 
-    const mysticatBaseUrl = (isDev && hasText(normalizedMystiqueUrl))
-      ? normalizedMystiqueUrl
-      : env.MYSTIQUE_API_BASE_URL;
+    // Validate the URL belongs to this site
+    let siteByUrl;
+    try {
+      siteByUrl = await dataAccess.Site.findByPreviewURL(previewBaseURL);
+    } catch (e) {
+      log.debug(`findByPreviewURL failed for ${previewBaseURL}: ${e.message}`);
```
johobot: If `findByPreviewURL` throws (e.g. a DB error), the exception is swallowed at `debug` level and execution falls through to the `if (!siteByUrl || ...)` check below, which returns `400 PREFLIGHT_INVALID_REQUEST` — "URL does not belong to this site" — when the real cause is an infrastructure fault.

Suggest treating the throw as a 500:
```js
} catch (e) {
  log.error(`findByPreviewURL failed for ${previewBaseURL}: ${e.message}`);
  return preflightError('PREFLIGHT_INTERNAL_ERROR', 'Internal error', 500);
}
```

src/controllers/preflight.js
```diff
@@ -357,256 +373,255 @@ function PreflightController(ctx, log, env) {
...
-      }
+    let preflight;
+    try {
+      preflight = await dataAccess.Preflight.create({
+        siteId,
+        asyncJobId: asyncJob.getId(),
+        url,
+        status: AsyncJob.Status.IN_PROGRESS,
+        createdBy,
+        startedAt: new Date().toISOString(),
```
johobot, line 507: `startedAt` is set here, before the `callMysticatAnalyze` call below. ADR-003 defines it as "When processing began", but actual processing begins when Mysticat accepts the POST. If Mysticat ever queues requests (or is slow to respond), the recorded `startedAt` will be earlier than real processing. Worth moving this to after `callMysticatAnalyze` succeeds, or clarifying in ADR-003 that it means "when the request was dispatched".

src/controllers/preflight.js
```diff
@@ -357,256 +373,255 @@ function PreflightController(ctx, log, env) {
...
+  /**
+   * Returns all preflights for a site, optionally filtered by URL.
+   * @param {Object} context - The request context
+   * @returns {Promise<Object>} 200 OK with array of preflights
+   */
+  const getAllPreflights = async (context) => {
+    const siteId = context.params?.siteId;
+    const rawQueryString = context.invocation?.event?.rawQueryString;
+    const urlFilter = rawQueryString
+      ? Object.fromEntries(
```
johobot: Hand-rolled query-string parsing can be replaced with the built-in `URLSearchParams`:
```js
const urlFilter = new URLSearchParams(context.invocation?.event?.rawQueryString ?? '').get('url') ?? undefined;
```
The current `split('=')` approach would split on every `=` in the key-value pair (edge case: bare `=` in an un-encoded value), and `Object.fromEntries` silently discards duplicate keys.

src/controllers/preflight.js
```diff
@@ -357,256 +373,255 @@ function PreflightController(ctx, log, env) {
...
     try {
-      const job = await dataAccess.AsyncJob.findById(jobId);
+      const preflight = await dataAccess.Preflight.findById(preflightId);
 
-      if (!job) {
-        log.error(`Job with ID ${jobId} not found`);
-        return notFound(`Job with ID ${jobId} not found`);
+      // Treat a siteId mismatch identically to not found — no cross-site probing
+      if (!preflight || preflight.getSiteId() !== siteId) {
+        return notFound(`Preflight with ID ${preflightId} not found`);
```
johobot: This 404 response uses `notFound(...)` which returns `{ message }` only. Every other error path in the three new endpoints goes through `preflightError(...)` and returns `{ errorCode, message }`. The inconsistency will make MFE error-handling harder — suggest:
```js
return preflightError('PREFLIGHT_NOT_FOUND', `Preflight with ID ${preflightId} not found`, 404);
```

test/controllers/preflight.test.js
```diff
@@ -855,36 +944,27 @@ describe('Preflight Controller', () => {
...
-      getStatus: () => betaJobStatus,
-      setStatus: sandbox.stub().callsFake((s) => { betaJobStatus = s; }),
-      getMetadata: () => betaJobMetadata,
-      setMetadata: sandbox.stub().callsFake((m) => { betaJobMetadata = m; }),
-      setError: sandbox.stub(),
-      setEndedAt: sandbox.stub(),
-      save: sandbox.stub().resolves(),
-    };
-
+  describe('createPreflight', () => {
```
johobot, line 955: The old `createBetaPreflightJob` suite had 5 auth-path tests (CS+promiseToken, AMS, CS_CW IMS fallback, SP+Secrets Manager, `ErrorWithStatusCode` propagation). `createPreflight` runs the same `resolvePromiseToken` → `retrievePageAuthentication` → `authorizationHeader` path, but none of those scenarios are covered here — only the no-auth happy path is tested. A regression in the auth branch would go undetected.

johobot (approved):
Comments are non-blocking suggestions — nothing here should hold up the merge.

src/controllers/preflight.js
GeezerPelletier: Fixed - `getAllPreflights` now parses from `context.invocation.event.rawQueryString` following the pattern from `feature-flags.js` and `brands.js`. The test has also been updated to construct a real `rawQueryString` with URL-encoded values so this can't regress.

src/controllers/preflight.js
GeezerPelletier: Fixed - the full Mysticat error is logged but the client response now returns the generic `'Upstream analyze service failed'` message so internal details aren't leaked.

src/controllers/preflight.js
GeezerPelletier: Fixed - rollback now uses `.catch((re) => log.warn(`Failed to roll back AsyncJob \: \`))` so orphaned jobs surface in logs.

src/controllers/preflight.js
GeezerPelletier: Fixed - both `preflight.save()` and `asyncJob.save()` now have `.catch((e) => log.warn(...))` so inconsistent DB state is surfaced in logs rather than swallowed.

src/controllers/preflight.js
GeezerPelletier: Fixed both points:
1. Added `AbortController` with a 15s timeout in `callMysticatAnalyze`, cleared in a `finally` block so it doesn't leak.
2. `MYSTIQUE_API_BASE_URL` validation moved into `createPreflight` at call time (returns 500) rather than the constructor, so `getAllPreflights` and `getPreflightById` can still be instantiated without it - the misconfiguration fails loud when it matters.

src/controllers/preflight.js
GeezerPelletier: Fixed - `Preflight.create` now uses `AsyncJob.Status.IN_PROGRESS` consistently throughout, matching line 469.

src/controllers/preflight.js
GeezerPelletier: Fixed - added `log.debug(`findByPreviewURL failed for \: \`)` in the catch block so unexpected failures surface in logs.

src/controllers/preflight.js
GeezerPelletier: Fixed - destructuring on the first parse: `const { protocol, hostname } = new URL(url); const previewBaseURL = \\\//\\\\`.

src/controllers/preflight.js
GeezerPelletier: Good catch - fixed. A throw from `findByPreviewURL` now returns `PREFLIGHT_INTERNAL_ERROR` 500 rather than falling through to the `!siteByUrl` check and returning a misleading 400. Added a test for the throw case as well.

src/controllers/preflight.js
GeezerPelletier: Fair point on the semantics. `startedAt` is set at record creation because the Preflight row must exist before we call Mysticat (so we can persist FAILED state if it throws). Moving it to post-call would require a second save on the happy path. Clarified the field description in ADR-003 to read: ''When the request was dispatched to the analyze service (set at record creation, before the upstream call completes)''.

src/controllers/preflight.js
GeezerPelletier: Good suggestion - replaced with `new URLSearchParams(rawQueryString ?? '').get('url') ?? undefined`. Handles the duplicate-key and bare-`=` edge cases correctly.

src/controllers/preflight.js
GeezerPelletier: Fixed - now uses `preflightError('PREFLIGHT_NOT_FOUND', ..., 404)` for consistency. Updated both the not-found and siteId-mismatch paths, and tightened the tests to assert `errorCode` rather than just status.

test/controllers/preflight.test.js
GeezerPelletier: Valid gap - we've added two `createPreflight` tests for the `x-promise-token` path (missing header -> 400, valid header -> 202 with token forwarded to `retrievePageAuthentication`). These will activate fully once PR #2480 (which makes `x-promise-token` the required mechanism) is merged and we rebase. The remaining cookie/IMS scenarios from the old flow will be superseded by that change.

clotton (dismissed):
## Review: feat: add site-scoped preflight REST endpoints (SITES-44686)

The architectural direction is solid — the `Preflight` entity extraction is well-reasoned, the auth guard ordering is correct, the rollback on `Preflight.create` failure prevents orphaned `AsyncJob` records, and the structured `errorCode` contract is a meaningful improvement over freeform 500 bodies. The test improvements are real. Three issues need fixing before this ships.

---

### Blocking

**1. URL filter in `getAllPreflights` is silently broken**

`src/controllers/preflight.js:532` — `context.params?.url` reads path parameters, not query parameters. `context.params` will never contain `url`. Every `GET /sites/X/preflights?url=...` request from the MFE returns an unfiltered list. The test passes only because it stubs `allBySiteIdAndUrl` directly without exercising query string parsing.

Fix: use `rawQueryString` to extract the URL filter:
```js
const rawQueryString = context.invocation?.event?.rawQueryString;
const urlFilter = new URLSearchParams(rawQueryString ?? '').get('url') ?? undefined;
```
(This was correct in the diff; it appears to have been reverted.)

**2. `findByPreviewURL` catch block converts DB errors into misleading 400s**

`src/controllers/preflight.js:402-409` — the catch comment says "URL doesn't match a known site" but `findByPreviewURL` doesn't throw on null results — it returns `null`. A thrown exception means a real DB error (connection failure, timeout, etc.). Swallowing that and returning `PREFLIGHT_INVALID_REQUEST / 400` gives callers an incorrect error code for what is a 5xx failure. Log the error and return a 500 to match every other DB call in these handlers.

**3. `getPreflightById` uses `notFound()` instead of `preflightError()` — breaks the error contract**

`src/controllers/preflight.js:589` — `notFound()` produces `{ message }` only. Every other new endpoint uses `preflightError(code, message, status)` which produces `{ errorCode, message }`. The test at line ~2321 in the diff expects `result.errorCode === 'PREFLIGHT_NOT_FOUND'`, which this call does not satisfy.

Fix:
```js
return preflightError('PREFLIGHT_NOT_FOUND', `Preflight with ID ${preflightId} not found`, 404);
```

---

### Important

**4. Empty `preflightAudits` passes through to Mysticat with `audits: []` — behavior undefined**

`src/controllers/preflight.js:427-430` — when no `*-preflight` audit types are enabled for a site, `preflightAudits` is an empty array and the code continues to `callMysticatAnalyze`. Whether Mysticat treats `audits: []` as "run all" or returns an error is untested. Either add a guard that short-circuits with a `PREFLIGHT_NOT_ENABLED / 403` here, or add a test proving Mysticat handles `[]` correctly.

**5. Raw upstream error message leaked to callers in the 502 response**

`src/controllers/preflight.js:515` — `mysticatError.message` contains the raw body from Mysticat (including status codes, stack traces, internal hostnames). Leaking this to API callers is a security and privacy concern. Use a generic message in the external response and log the full error server-side. The diff version had this right: `'Upstream analyze service failed'`.

**6. No OpenAPI spec for the three new endpoints**

`docs/openapi/` has no entries for `POST /sites/:siteId/preflights`, `GET /sites/:siteId/preflights`, or `GET /sites/:siteId/preflights/:preflightId`. The repo's CLAUDE.md workflow checklist requires an OpenAPI spec entry. The `PreflightDto` fields are well-defined in code but missing from the contract document.

---

### Minor

**7. `new URL(url)` parsed twice** — `src/controllers/preflight.js:382`. Destructure once: `const { protocol, hostname } = new URL(url)`.

**8. AMS and SP auth coverage dropped from tests** — the old beta suite tested CS Bearer, AMS token, CS_CW IMS fallback, and SP Secrets Manager paths. The new suite covers "CS Bearer present" and "auth not required." Low regression risk since the auth path is unchanged, but worth noting for follow-up.

---

Fixes to #1, #2, #3 are needed before merge. #1 makes a documented feature completely non-functional. #2 misreports error class. #3 breaks the error contract the test suite itself verifies.

GeezerPelletier:
Thanks for the second pass @clotton — but I think this review is looking at a stale view of the diff. All three Blocking items are already fixed in HEAD (commit `0ed4fbe0`), and the two Important items are non-issues for reasons that aren't visible from inside this PR. Walking through:

## Blocking — all three already fixed in HEAD

Your line numbers match the pre-fix snapshot. Current state on the branch tip:

**#1 — URL filter** — fixed at [src/controllers/preflight.js:546-547](https://github.com/adobe/spacecat-api-service/blob/sites-44686-preflight-api/src/controllers/preflight.js#L546-L547):
```js
const rawQueryString = context.invocation?.event?.rawQueryString;
const urlFilter = new URLSearchParams(rawQueryString ?? '').get('url') ?? undefined;
```
Matches the johobot suggestion exactly, with a real `rawQueryString` test added in `preflight.test.js` so this can't regress.

**#2 — `findByPreviewURL` catch returns 500, not 400** — fixed at [src/controllers/preflight.js:415-420](https://github.com/adobe/spacecat-api-service/blob/sites-44686-preflight-api/src/controllers/preflight.js#L415-L420):
```js
try {
  siteByUrl = await dataAccess.Site.findByPreviewURL(previewBaseURL);
} catch (e) {
  log.error(`findByPreviewURL failed for ${previewBaseURL}: ${e.message}`);
  return preflightError('PREFLIGHT_INTERNAL_ERROR', 'Internal error', 500);
}
```
With a test for the throw case.

**#3 — `getPreflightById` uses `preflightError`, not `notFound`** — fixed at [src/controllers/preflight.js:604](https://github.com/adobe/spacecat-api-service/blob/sites-44686-preflight-api/src/controllers/preflight.js#L604):
```js
return preflightError('PREFLIGHT_NOT_FOUND', `Preflight with ID ${preflightId} not found`, 404);
```
Tests assert `errorCode === 'PREFLIGHT_NOT_FOUND'`, not just status.

The parenthetical on your #1 ("This was correct in the diff; it appears to have been reverted") was the giveaway — `git show HEAD:src/controllers/preflight.js` confirms the fix is still in place; nothing was reverted. I'd guess the review was generated against a cached older diff snapshot. A diff refresh should clear all three.

## Important — both invalid in the broader system context

**#4 — empty `preflightAudits` passing through to Mysticat** — invalid because spacecat no longer needs to send an audit list at all once mystique [PR #2405](https://git.corp.adobe.com/experience-platform/mystique/pull/2405) lands. That PR implements the [Preflight Opportunity Enablement ADR](https://github.com/adobe/mysticat-architecture/blob/main/platform/decisions/preflight-opportunity-enablement.md): the mystique executor resolves the goal set itself via three-gate filtering (tier → `run_in_preflight=true` fact registry → per-site `goal_overrides`) when no `audits` field is sent on the request. The follow-up cleanup on this PR's side is to drop the `preflightAudits` computation + the `audits` parameter from `callMysticatAnalyze`; with that, `request.audits` is absent and the empty-array edge case can't arise because the field isn't there. (Mystique still accepts an `audits` field as an explicit override for debugging, but the API isn't expected to use it.)

**#6 — no OpenAPI spec** — intentional, and called out in the PR body under **Design decisions**: *"New endpoints are undocumented (not in the public OpenAPI spec) intentionally — they will remain that way until the feature is ready for GA."* The MFE is the only consumer pre-GA and ships with internal contracts; we'd rather not publish a public contract we know will change before launch.

## Minor

**#7 — `new URL` parsed twice** — fixed at [src/controllers/preflight.js:394](https://github.com/adobe/spacecat-api-service/blob/sites-44686-preflight-api/src/controllers/preflight.js#L394) via destructure.

**#8 — AMS / SP auth coverage** — superseded. PR [#2480](https://github.com/adobe/spacecat-api-service/pull/2480) makes `x-promise-token` the required auth mechanism for these flows; the AMS / SP / cookie / IMS-fallback scenarios from the old beta suite no longer apply. The two new `x-promise-token` tests in `preflight.test.js` cover the new path (and will activate fully on the rebase after #2480 lands — flagged in the PR body).

Happy to walk through any of this if I've misunderstood the concerns. But unless I'm missing something, all the Blocking items are landed and the Important items are by design / out of scope.

MysticatBot (changes requested):
Hey @GeezerPelletier,

## Strengths

- **Previously flagged issues all resolved.** URL filter now reads from `rawQueryString` via `URLSearchParams` (src/controllers/preflight.js:546-547), `findByPreviewURL` catch returns 500 (not 400), `getPreflightById` uses `preflightError()` consistently, and the 502 response uses a generic message. Well done addressing the prior round.
- **Consistent structured error contract** - `preflightError(errorCode, message, status)` (src/controllers/preflight.js:319) is used across all error paths, giving callers a predictable `{ errorCode, message }` shape.
- **Cross-site isolation done right** - returning 404 (not 403) on siteId mismatch in `getPreflightById` (src/controllers/preflight.js:709) prevents site enumeration.
- **Authorization ordering is correct** - the guard sequence (site exists -> hasAccess -> URL ownership -> entitlement) never discloses data before access verification.
- **Rollback on Preflight.create failure** (src/controllers/preflight.js:496) with logged `.catch()` on the rollback itself is the right pattern.
- **AbortController with 15s timeout on Mysticat fetch** (src/controllers/preflight.js:341-344) with proper `clearTimeout` in `finally`. Addresses the prior concern about hanging requests.
- **Domain model separation is well-reasoned** - ADR-003 clearly justifies extracting Preflight from AsyncJob. The dual-store boundary with natural TTL expiry is clean (no migration, no backfill).
- **Simplified Mysticat call surface** - removing `mode`, `step`, and `audits` reduces coupling and aligns with the Preflight Opportunity Enablement ADR.
- **Thorough test coverage** - 64 tests covering happy paths, error paths, rollback, cross-site isolation, query string routing, and the createdBy fallback ladder.
- **DTO layer follows the documented Request -> Controller -> DTO -> Response pattern** with proper field projection between list and detail views.

## Issues

### Important (Should Fix)

**1. `checkEnableAuthentication` can throw unhandled from `createPreflight`**

src/controllers/preflight.js:436 - `checkEnableAuthentication` does a bare `fetch(url, { method: 'HEAD' })` with no error handling. If the customer URL is unreachable (DNS failure, connection refused, TLS error), `fetch()` throws and there is no try/catch around this call. The old `createBetaPreflightJob` had one giant outer try/catch that caught this; the new `createPreflight` has granular try/catch blocks around each operation but misses this one. The result is an unstructured error response that breaks the `{ errorCode, message }` contract this PR establishes.

Fix:
```js
let enableAuthentication;
try {
  enableAuthentication = await checkEnableAuthentication(previewBaseURL);
} catch (e) {
  log.warn(`checkEnableAuthentication failed for ${previewBaseURL}: ${e.message}`);
  enableAuthentication = false;
}
```

Defaulting to `false` is safe - if the URL is unreachable, Mysticat will also fail and report back through the FAILED status path.

**2. Missing integration tests for new endpoints**

The repo's CLAUDE.md ("Test Requirements" section) states that new or modified endpoints must include integration tests in `test/it/`. This PR adds three new endpoints with unit tests only. No `test/it/` changes are present. The unit tests mock all data access calls, which means the full request lifecycle (middleware auth -> route matching -> controller -> data access -> response) is not exercised end-to-end.

Fix: Add integration tests (can be a follow-up PR if Preflight seed data requires changes in spacecat-shared-data-access first, but should be tracked).

**3. No test for AbortController timeout behavior**

src/controllers/preflight.js:348-349 - The 15s AbortController timeout on the Mysticat fetch throws an `AbortError`, which has a different shape than a standard HTTP failure. No test verifies that this specific error path produces the expected 502 `PREFLIGHT_UPSTREAM_ERROR` response and marks both Preflight and AsyncJob as FAILED. If the catch block at line 508 doesn't handle `AbortError` correctly, the timeout could produce an unstructured response.

Fix: Add a test where `fetchStub` rejects with an `AbortError` (or a `DOMException` with name `AbortError`) and assert the 502 + FAILED state behavior.

### Minor (Nice to Have)

**4. `badRequest(e.message)` breaks the structured error contract**

src/controllers/preflight.js:444 - When `resolvePromiseToken` throws an `ErrorWithStatusCode`, the handler returns `badRequest(e.message)` which produces `{ message }` only (no `errorCode`). Every other error path uses `preflightError()`. This means callers cannot rely on `errorCode` being present on all 4xx responses.

Fix: `return preflightError('PREFLIGHT_INVALID_REQUEST', e.message, 400);`

**5. Protocol restriction on user-provided URL**

src/controllers/preflight.js:394 - `isValidUrl()` accepts any protocol that `new URL()` parses (including `file://`, `ftp://`). While `fetch()` typically rejects non-HTTP protocols and the URL must match a registered site preview URL (limiting exploitability), adding an explicit protocol check is defense-in-depth:
```js
if (!['http:', 'https:'].includes(protocol)) {
  return preflightError('PREFLIGHT_INVALID_REQUEST', 'url must use http or https', 400);
}
```

## Recommendations

- Log Mysticat call duration. A simple `const start = Date.now()` before and `log.info({ duration: Date.now() - start })` after would be the first thing needed during a slow-Mysticat investigation.
- Document the projector dependency in ADR-003. The PR description mentions "the projector picks them up via SNS and flips Preflight.status" but the ADR Consequences section doesn't state that the projector must be updated to write to the Preflight entity. A one-line note would help future readers understand the lifecycle completion path.
- Consider standardizing the repeated site-lookup + access-check pattern into a route-level middleware in a follow-up. All three handlers repeat the same seven lines.

## Assessment

Ready to merge? With fixes
Reasoning: The `checkEnableAuthentication` unhandled throw is a regression from the old code (which had an outer try/catch). It's a one-line fix that prevents unstructured errors from reaching callers. The other Important items (integration tests, abort timeout test) can be tracked as follow-ups if needed, but the unhandled throw should be fixed before merge.

## Next Steps

1. Wrap `checkEnableAuthentication` call in a try/catch (one-line fix, prevents contract-breaking errors).
2. Add a test for the AbortController timeout path.
3. Track integration tests as a follow-up (may depend on shared-data-access Preflight seed data).
4. The Minor items are optional but quick to address in the same commit.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 43s | <b>Cost:</b> $7.65 | <b>Commit:</b> <code>c9540d13d56ef1f1ab3f388b0fc7808158cddc29</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

GeezerPelletier:
Thanks @MysticatBot-Dev — fresh pass picked up two real contract defects worth fixing now. Landed in `5b5ade81`. Deferring the rest deliberately to keep this PR's diff focused on the original SITES-44686 scope.

## Implemented

**Important #1 — `checkEnableAuthentication` unhandled throw**

Wrapped in try/catch with `log.warn` and `enableAuthentication = false` on throw, exactly as you proposed. An unreachable customer URL now defaults to auth-not-required; if the URL is genuinely broken, the downstream Mysticat call also fails and surfaces a structured 502 via the existing FAILED-status path. Added one test verifying HEAD throw → 202 with no Authorization header.

**Minor #4 — `badRequest(e.message)` breaks the structured-error contract**

Swapped to `preflightError('PREFLIGHT_INVALID_REQUEST', e.message, 400)`. Test strengthened to assert `errorCode === 'PREFLIGHT_INVALID_REQUEST'` and the exact message, not just status code.

## Deferred (with rationale)

**Important #2 — Integration tests in `test/it/`**

Real gap per CLAUDE.md's Test Requirements. The Preflight entity needs seed data in `test/it/postgres/seed-data/` plus shared test factories in `test/it/shared/tests/preflight.js` plus a postgres wiring file — non-trivial scope and meaningfully separate from the controller-level work in this PR. Tracking as a follow-up ticket; happy to take the next pass on it once this lands.

**Important #3 — AbortError-specific test for the 15s Mysticat timeout**

Real test gap, but the **functional** path works: any throw from `callMysticatAnalyze` (network error, AbortError, non-2xx) lands in the same catch block at [preflight.js:508](src/controllers/preflight.js#L508) and produces the same 502 + FAILED-state response. The catch doesn't discriminate on error type, so AbortError specifically behaves identically to any other rejection. Worth a dedicated test for documentation purposes, but the path isn't broken — deferring to a coverage-improvement follow-up.

**Minor #5 — Protocol whitelist (`http:` / `https:`)**

Defense-in-depth, real but low risk: the URL must match a registered site preview URL via `findByPreviewURL`, and registered sites are http(s) only. An attacker can't get a `file://` URL through that gate. Worth adding but not blocking; deferring.

## Recommendations

- **Log Mysticat call duration** — agreed it'd help slow-Mysticat investigations; would slot naturally next to the existing `log.error` on Mysticat failure. Deferring as a small operability follow-up.
- **Document projector dependency in ADR-003 Consequences** — fair point; the lifecycle bridge through the projector matters for future readers. Deferring to a doc-only follow-up.
- **Middleware refactor for repeated site-lookup + access-check** — real abstraction opportunity but cross-handler; not appropriate to bundle into this PR.

Net: 78 unit tests pass, coverage stays at 100% lines / 99.4% branches (remaining uncovered branch is in the legacy `createPreflightJob`'s `traceId` path, not in this diff). Ready for another look.

MysticatBot (commented):
Hey @GeezerPelletier,

## Strengths

- **Previously flagged issues all resolved.** `checkEnableAuthentication` wrapped in try/catch with `log.warn` and `enableAuthentication = false` default (src/controllers/preflight.js:434-439); `badRequest(e.message)` replaced with `preflightError('PREFLIGHT_INVALID_REQUEST', e.message, 400)` for consistent structured error contract (src/controllers/preflight.js:449). Both fixes are clean and well-tested.
- **Error handling is granular and recoverable throughout `createPreflight`.** Each I/O operation has its own try/catch with a specific error code and appropriate log level. Textbook pattern.
- **Authorization ordering is correct across all three endpoints** - site exists, hasAccess, URL ownership/entitlement checks never disclose data before access verification.
- **Cross-site isolation done right** - returning 404 (not 403) on siteId mismatch in `getPreflightById` (src/controllers/preflight.js:604) prevents site enumeration.
- **SSRF surface is well-gated** - user-provided URLs must pass `findByPreviewURL` ownership validation, restricting Mysticat calls to registered site domains only.
- **Domain model separation is well-reasoned** - ADR-003 clearly justifies the dedicated `Preflight` entity. The dual-store boundary with natural TTL expiry via `ON DELETE CASCADE` avoids migration overhead.
- **Simplified Mysticat call surface** - removing `mode`, `step`, and `audits` reduces coupling and aligns with the Preflight Opportunity Enablement ADR. The four-field request body is minimal.
- **Lifecycle completion via projector + SNS** keeps the Lambda invocation short without long-poll or webhook callback in the request path.
- **AbortController with 15s timeout** (src/controllers/preflight.js:341-344) with proper `clearTimeout` in `finally` prevents hanging requests and timer leaks.
- **Rollback on Preflight.create failure** (src/controllers/preflight.js:496) with logged `.catch()` on the rollback itself gives ops visibility into cleanup failures.
- **URL filter correctly reads from rawQueryString** (src/controllers/preflight.js:546-547) via `URLSearchParams` - the critical bug from the first review round is fixed.
- **DTO separation is clean** - `toJSON` for lightweight list view, `toDetailJSON` for full detail including result/error.
- **createdBy is server-derived** from the authenticated IMS profile, never accepted from client input.
- **Thorough test coverage** - 78 tests with 100% line coverage per Codecov, spanning happy paths, error paths, auth flows, rollback, cross-site isolation, and the createdBy fallback ladder.

## Issues

### Minor (Nice to Have)

**1. Mysticat error body stored verbatim in the Preflight record, visible to API callers**

src/controllers/preflight.js:509 - `preflight.setError({ code: 'MYSTICAT_ERROR', message: mysticatError.message })` stores the full upstream error text, which is constructed at line 360 as `Mysticat returned ${response.status}: ${text}` (the raw response body). While the API response on the 502 path correctly uses a generic message, any caller who later reads `GET /sites/:siteId/preflights/:preflightId` via `PreflightDto.toDetailJSON` sees the stored `error.message` - potentially exposing internal hostnames, stack traces, or debugging info from Mysticat.

Fix: Store a sanitized message matching the external response:
```js
preflight.setError({ code: 'MYSTICAT_ERROR', message: 'Upstream analyze service failed' });
```
The full error is already logged at line 508 for operational debugging.

**2. Mysticat 502 test does not assert asyncJob FAILED state**

test/controllers/preflight.test.js:1168 - The test asserts `mockPreflight.setStatus('FAILED')` and `mockPreflight.save()` but does not verify `mockJob.setStatus('FAILED')` or `mockJob.save()`. The controller explicitly updates both records on failure (src/controllers/preflight.js:525-526). If a future refactor removed the asyncJob update, this test would still pass.

Fix: Add assertions:
```js
expect(mockJob.setStatus).to.have.been.calledWith('FAILED');
expect(mockJob.save).to.have.been.called;
```

## Recommendations

- Consider dropping the AsyncJob backing entirely in a future iteration. Since `asyncJobId` is never exposed to consumers and the projector writes directly to `Preflight.status`, the AsyncJob row is write-only overhead. A `record_expires_at` column on `preflights` would replace the CASCADE expiry. Not a blocker - the natural next simplification once the projector migration is proven.
- Document the projector dependency in ADR-003 Consequences. The lifecycle path (Mysticat -> SNS -> projector -> Preflight.status flip) is in the PR body but absent from the ADR. A one-line note helps future readers.
- The `getAllPreflights` list view test could assert that detail-only fields (`result`, `error`, `updatedAt`, `startedAt`, `endedAt`) are absent from response items, verifying the DTO projection contract.

## Assessment

Reasoning: All prior blocking issues are resolved. The architecture is sound - clean entity separation, minimal Mysticat coupling, correct auth ordering, and solid test coverage. The two minor findings (stored error message visibility, test assertion gap) are low-risk improvements that can be addressed in-place or as follow-ups.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 1s | <b>Cost:</b> $6.58 | <b>Commit:</b> <code>5b5ade81c3f40b47edbe13e4b157699a1518b2d2</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

GeezerPelletier:
Thanks @MysticatBot-Dev — both minor items landed in `7939f429`.

## Implemented

**Minor #1 — Stored `Preflight.error.message` exposed via GET detail**

Good catch — same class of leak as the 502 response we sanitized in an earlier round, just stored on the row. Now `preflight.setError({ code: 'MYSTICAT_ERROR', message: 'Upstream analyze service failed' })`, matching the external 502 message. Full upstream error stays in `log.error` for ops.

**Minor #2 — 502 test missing AsyncJob FAILED assertions**

Added `expect(mockJob.setStatus).to.have.been.calledWith('FAILED')` + `expect(mockJob.save).to.have.been.called`, and tightened the `setError` matcher to assert the sanitized message — so a regression on Minor #1 would also be caught here.

## Recommendations — deferred

- **Drop AsyncJob entirely in a future iteration** — agree this is the natural next simplification once the projector path is proven in production. Its own PR.
- **Document projector dependency in ADR-003 Consequences** — fair; we noted this in the last round too. A doc-only follow-up keeps this PR focused on the controller work.
- **`getAllPreflights` list view: assert detail-only fields are absent** — covered by the `PreflightDto` unit tests today (`toJSON` vs `toDetailJSON` projection); adding a controller-level assertion would be duplicative.

78 unit tests pass.

MysticatBot (approved):
Hey @GeezerPelletier,

## Strengths

- **Previously flagged: stored error message sanitized** (src/controllers/preflight.js:604) - The raw Mysticat response body is no longer persisted on the `Preflight` record. The stored message now mirrors the external 502 response (`'Upstream analyze service failed'`), with the inline comment clearly explaining why the raw upstream body could leak internal hostnames/stack traces via the GET detail endpoint. Full error remains in `log.error` for ops. Clean fix.
- **Previously flagged: 502 test now asserts AsyncJob FAILED state** (test/controllers/preflight.test.js) - The test verifies both `mockJob.setStatus('FAILED')` and `mockJob.save()`, plus tightens the `setError` matcher to assert the sanitized message. A future refactor that drops the AsyncJob update or reverts the sanitization will be caught.
- **Focused commit** - Touches exactly the two locations flagged in the prior review, with no unrelated changes mixed in.
- **Error message sanitization closes an information-disclosure seam** between the write path (POST 502 response already used generic message) and the read path (GET detail now also sees the generic message). Consistent treatment across both surfaces.
- **All project conventions satisfied** - routes in both `src/routes/index.js` and `src/index.js`, `AccessControlUtil` for tenant access, `PreflightDto` for response transformation, UUID validation for `preflightId`, structured error responses, required capabilities registered with appropriate read/write semantics.
- **Overall architecture remains sound** - dedicated Preflight entity with FK to AsyncJob, lifecycle completion via SNS projector, dual-store boundary, simplified Mysticat call surface (four fields, no caller-supplied audit list), cross-site isolation via 404 on mismatch, AbortController timeout.
- **Thorough test coverage** - 78 unit tests with 100% line coverage per Codecov, spanning happy paths, error paths, auth flows, rollback, cross-site isolation, query string routing, and the createdBy fallback ladder.

## Assessment

Ready to merge? Yes
Reasoning: Both prior minor findings are cleanly resolved. The 1-commit delta is tight and correct. The broader architecture has been validated across 4 review rounds with comprehensive test coverage. No new issues introduced.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 29s | <b>Cost:</b> $2.55 | <b>Commit:</b> <code>7939f429683c993508bb993fde09b43454ff1aab</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

clotton (approved):
All blocking issues from my earlier review are addressed:

- URL filter uses `rawQueryString` correctly — filter is functional
- `findByPreviewURL` catch logs the error and returns a 500 instead of a misleading 400
- `getPreflightById` uses `preflightError('PREFLIGHT_NOT_FOUND', ...)` — error contract consistent
- Upstream Mysticat error sanitized before returning to caller, full message logged server-side

Good to merge.

PR #2480, opened 2026-05-25
feat: resolve promise token from the header required to create a preflight job

radhikagpt1208:
## Summary

**JIRA:** [SITES-45229](https://jira.corp.adobe.com/browse/SITES-45229)

Requires clients to supply the IMS promise token on the **`x-promise-token`** request header when creating preflight jobs for promise-based authoring types (**CS, CS_CW, AMS**). Removes the previous cookie-based (`promiseToken`) and IMS session-JWT mint fallback (`getIMSPromiseToken`).

This is the **API half** of the promise-token header migration; the MFE client changes are in [OneAdobe/aem-sites-optimizer-preflight-mfe#204](https://github.com/OneAdobe/aem-sites-optimizer-preflight-mfe/pull/204). Tokens are obtained client-side via **`POST /auth/v2/promise`** (auth-service).

### What changed

- **`getPromiseTokenHeader(context)`** — case-insensitive read of `x-promise-token` from `context.pathInfo.headers`; trims value; `decodeURIComponent` with safe fallback on malformed encoding.
- **`resolvePromiseToken(site, context)`** — for CS / CS_CW / AMS, returns `{ promise_token }` from the header or throws **`ErrorWithStatusCode` 400** with `Invalid request: missing required header: x-promise-token`; non-promise site types return `null` (unchanged).
- **`createPreflightJob` / `createBetaPreflightJob`** — use header-based resolution only; SQS payload still uses `{ promise_token: value }` shape expected by downstream audit workers (inherited from prior IMS helper — follow-up cleanup suggested in review).
- **OpenAPI / ADR** — `docs/openapi/parameters.yaml`, `docs/decisions/002-preflight-api-rest-redesign.md`, and related decision doc updated for header-based flow.
- **Tests** — coverage for missing/empty/whitespace/non-string headers, URL-encoded values, malformed `%` encoding fallback, beta + legacy job paths, and traceId on SQS when present in controller ctx.

### Behavior

| Site type | `x-promise-token` | Result |
|-----------|-------------------|--------|
| CS / CS_CW / AMS | present | Job created; token forwarded to SQS |
| CS / CS_CW / AMS | absent / empty | **400** — missing required header |
| Other (e.g. EDS) | ignored | No promise token required |

## Test plan

- [x] `npm test test/controllers/preflight.test.js` — new/updated promise-token cases pass
- [x] Deploy to dev branch; create preflight job for CS site **with** `x-promise-token` → **202**
- [ ] Same request **without** header → **400** `Invalid request: missing required header: x-promise-token`
- [ ] EDS site job create without header → still **202** (unchanged)
- [ ] End-to-end with MFE PR [#204](https://github.com/OneAdobe/aem-sites-optimizer-preflight-mfe/pull/204) once both are deployed

## Related PRs

- [OneAdobe/aem-sites-optimizer-preflight-mfe#204](https://github.com/OneAdobe/aem-sites-optimizer-preflight-mfe/pull/204) — MFE: `/auth/v2/promise` + send `x-promise-token` on job create
- https://github.com/adobe/spacecat-auth-service/pull/550/ - add v2/promise API to return promise token in response body

---

## PR checklist

- [x] Linked related issue (SITES-45229) and described the problem being solved
- [x] Updated API specification / ADR for implementation change
- [ ] Regenerated OpenAPI HTML docs (if required by repo process)
- [x] Added/updated tests for changed behavior

Thanks for contributing!


GeezerPelletier:
LGTM - holding approval pending coverage fix.

The `getPromiseTokenHeader` helper is well-structured: case-insensitive header lookup, `decodeURIComponent` with a safe fallback, and proper `hasText` guard. The new `resolvePromiseToken` is cleaner and the intent is explicit - CS/CS_CW/AMS sites must supply `x-promise-token`; the IMS fallback is gone.

A few notes from review:

- The `ErrorWithStatusCode(400)` throw for a missing header is the right shape - callers that wrap `resolvePromiseToken` (like `createPreflight` in our implementation branch) already handle it and return `badRequest`, so no change needed there.
- ADR-002 update is accurate and consistent with the implementation.
- 7 new test cases give good coverage of the edge cases (empty string, whitespace, encoded value, wrong-case header name).

One minor nit (non-blocking): the `decodeURIComponent` fallback silently swallows a malformed URI component by returning the raw trimmed value. That's probably the right call for a bearer token, just worth noting in case there's ever a need to distinguish "bad encoding" from "valid token that happens to contain %".

**Blocked on coverage**: the Codecov patch check is at 88.24% with 6 uncovered lines - below threshold. The implementation looks correct, but this should be resolved before merging.

docs/openapi/parameters.yaml
```diff
@@ -388,9 +388,8 @@ cursor:
 xPromiseToken:
   name: x-promise-token
   description: |
-    Optional session-based promise token. When provided, this token is used directly
-    instead of creating one via IMS. Only relevant for promise-based authoring types
-    (CS, CS_CW, AMS); ignored for other site types.
+    Optional session-based promise token. When provided, this token is used directly for operations like creating a preflight job for authenticated CMS pages.
```
bhellema: Generating the openapi requires to rebuild the docs again.  You should get a html file out of that process.

src/controllers/preflight.js
```diff
@@ -158,17 +147,55 @@ function PreflightController(ctx, log, env) {
...
+      return null;
+    }
+    const trimmed = String(raw).trim();
+    try {
+      return decodeURIComponent(trimmed);
+    } catch {
+      return trimmed;
+    }
+  }
+
```
bhellema: I'm surprised that this isn't in a some shared utility.

src/controllers/preflight.js
```diff
@@ -158,17 +147,55 @@ function PreflightController(ctx, log, env) {
...
+   * Reads `x-promise-token` from pathInfo.headers.
+   * @param {Object} context - Request context
+   * @returns {string|null} Trimmed decoded token, or null if absent / empty
+   */
+  function getPromiseTokenHeader(context) {
+    const headers = context.pathInfo?.headers;
+    if (!isNonEmptyObject(headers)) {
+      return null;
+    }
+    const entry = Object.entries(headers).find(([name]) => name.toLowerCase() === 'x-promise-token');
```
bhellema: Do we not have constants defined for these headers somewhere?

src/controllers/preflight.js
```diff
@@ -158,17 +147,55 @@ function PreflightController(ctx, log, env) {
...
+    const trimmed = String(raw).trim();
+    try {
+      return decodeURIComponent(trimmed);
+    } catch {
+      return trimmed;
+    }
+  }
+
   async function resolvePromiseToken(site, context) {
     if (!PROMISE_BASED_TYPES.includes(site.getAuthoringType())) {
```
bhellema: fwiw - this constant does not say "what" it is.  PROMISE_BASED_AUTHORING_TYPES would be more accurate.

src/controllers/preflight.js
```diff
@@ -158,17 +147,55 @@ function PreflightController(ctx, log, env) {
...
+    }
+    const trimmed = String(raw).trim();
+    try {
+      return decodeURIComponent(trimmed);
+    } catch {
+      return trimmed;
+    }
+  }
+
   async function resolvePromiseToken(site, context) {
```
bhellema, line 156: JS Docs that describe the return object and the error would be good.

src/controllers/preflight.js
bhellema, line 207: Why do we not always use the siteId here.  Seems like a direct link to the correct site.

src/controllers/preflight.js
```diff
@@ -158,17 +147,55 @@ function PreflightController(ctx, log, env) {
...
     if (!PROMISE_BASED_TYPES.includes(site.getAuthoringType())) {
       return null;
     }
-    const cookieToken = getCookieValue(context, 'promiseToken');
-    if (hasText(cookieToken)) {
-      return { promise_token: cookieToken };
+    const promiseTokenHeader = getPromiseTokenHeader(context);
+    if (hasText(promiseTokenHeader)) {
+      return { promise_token: promiseTokenHeader };
```
bhellema, line 170: Why is this an object and not just the value of the promise token?

radhikagpt1208:
> LGTM - holding approval pending coverage fix.
> 
> The `getPromiseTokenHeader` helper is well-structured: case-insensitive header lookup, `decodeURIComponent` with a safe fallback, and proper `hasText` guard. The new `resolvePromiseToken` is cleaner and the intent is explicit - CS/CS_CW/AMS sites must supply `x-promise-token`; the IMS fallback is gone.
> 
> A few notes from review:
> 
> * The `ErrorWithStatusCode(400)` throw for a missing header is the right shape - callers that wrap `resolvePromiseToken` (like `createPreflight` in our implementation branch) already handle it and return `badRequest`, so no change needed there.
> * ADR-002 update is accurate and consistent with the implementation.
> * 7 new test cases give good coverage of the edge cases (empty string, whitespace, encoded value, wrong-case header name).
> 
> One minor nit (non-blocking): the `decodeURIComponent` fallback silently swallows a malformed URI component by returning the raw trimmed value. That's probably the right call for a bearer token, just worth noting in case there's ever a need to distinguish "bad encoding" from "valid token that happens to contain %".
> 
> **Blocked on coverage**: the Codecov patch check is at 88.24% with 6 uncovered lines - below threshold. The implementation looks correct, but this should be resolved before merging.

Thanks for the detailed feedback, @GeezerPelletier.
For the coverage concern, I recently pushed some commits to increase the test coverage but that commits haven't been built and deployed yet. It seems to be stuck at the Kodiak check itself. Will check again in some time.

src/controllers/preflight.js
bhellema, line 261: It's odd that we set a property called promiseToken to an object { promise_token: value } is it not?  Why is this not just .promiseToken = value? I know the ripple effect of this would require a change down the road in the other lambdas, so it might not be worth the effort, but it does seem odd.

GeezerPelletier:
Minor nit: the new ADR file is named `pf-request-id-spacecat-design.md` which doesn't match the naming convention used by the other decisions in this folder (`002-preflight-api-rest-redesign.md`, `003-preflight-entity-design.md`). Suggest renaming to `004-preflight-promise-token-header.md` (or similar) to keep things consistent and easy to scan in sequence.

src/controllers/preflight.js
GeezerPelletier: +1 on the rename. Note this constant exists with the same name (`PROMISE_BASED_TYPES`) in our implementation branch (PR #2478), so if this is renamed here we'll need to pick it up after rebasing.

src/controllers/preflight.js
GeezerPelletier: Agreed the shape is a bit odd. The `{ promise_token: value }` wrapping is inherited from the existing pattern used by `getIMSPromiseToken` which returns the full IMS token object - so the downstream lambdas already expect that shape. Changing it now would be a cross-repo ripple. Worth a follow-up ticket to clean up but not a blocker here.

radhikagpt1208:
> > LGTM - holding approval pending coverage fix.
> > The `getPromiseTokenHeader` helper is well-structured: case-insensitive header lookup, `decodeURIComponent` with a safe fallback, and proper `hasText` guard. The new `resolvePromiseToken` is cleaner and the intent is explicit - CS/CS_CW/AMS sites must supply `x-promise-token`; the IMS fallback is gone.
> > A few notes from review:
> > 
> > * The `ErrorWithStatusCode(400)` throw for a missing header is the right shape - callers that wrap `resolvePromiseToken` (like `createPreflight` in our implementation branch) already handle it and return `badRequest`, so no change needed there.
> > * ADR-002 update is accurate and consistent with the implementation.
> > * 7 new test cases give good coverage of the edge cases (empty string, whitespace, encoded value, wrong-case header name).
> > 
> > One minor nit (non-blocking): the `decodeURIComponent` fallback silently swallows a malformed URI component by returning the raw trimmed value. That's probably the right call for a bearer token, just worth noting in case there's ever a need to distinguish "bad encoding" from "valid token that happens to contain %".
> > **Blocked on coverage**: the Codecov patch check is at 88.24% with 6 uncovered lines - below threshold. The implementation looks correct, but this should be resolved before merging.
> 
> Thanks for the detailed feedback, @GeezerPelletier. For the coverage concern, I recently pushed some commits to increase the test coverage but that commits haven't been built and deployed yet. It seems to be stuck at the Kodiak check itself. Will check again in some time.

@GeezerPelletier Pushed additional preflight controller tests (header edge cases, traceId on SQS, non-ErrorWithStatusCode paths). Codecov is now updated on the latest deploy with 100% coverage of diff hit.

docs/openapi/parameters.yaml
radhikagpt1208: Done.

src/controllers/preflight.js
radhikagpt1208: Resolved, refer to the above comment.

src/controllers/preflight.js
radhikagpt1208: Renamed, shifted PROMISE_BASED_AUTHORING_TYPES constant in `src/utils/constants.js` for all consumers (preflight and scrapeJob controller) to use.

GeezerPelletier (approved):
Approved. Coverage is green now (codecov/patch passing on the latest commit), all 17 checks pass, and the new test cases hit exactly the edge cases I called out in the earlier round — missing/empty/whitespace/non-string header, base64 token with `=` characters preserved, decode-fallback path, beta path, AMS/CS/CS_CW coverage, plus the non-`ErrorWithStatusCode` 500 paths I'd asked about.

Final pass on the diff:
- `getPromiseTokenHeader` looks good: case-insensitive header lookup, `hasText` guard, trimmed `decodeURIComponent` with a safe raw-string fallback. Clean and self-contained.
- `resolvePromiseToken` now throws `ErrorWithStatusCode(STATUS_BAD_REQUEST, ...)` explicitly when the header is missing for CS/CS_CW/AMS — exactly what downstream callers like `createPreflight` expect.
- ADR-002 wording matches the implementation (header-based, no IMS fallback).
- Cookie + IMS fallback paths are cleanly removed; no dead-code residue.

Non-blocking follow-up (same nit as before): the new decision file is still named `pf-request-id-spacecat-design.md` — the other decisions in this folder follow `NNN-<slug>.md` (`002-preflight-api-rest-redesign.md`, `003-preflight-entity-design.md`). Worth renaming to `004-…` either in this PR before you merge or in a small cleanup right after. Either way, this approval doesn't depend on it.

@radhikagpt1208 — ready to merge whenever you're ready. We need this in to unblock [#2478](https://github.com/adobe/spacecat-api-service/pull/2478), so a nudge from this side: please go ahead and merge when you have a moment. Thanks for the quick turnaround on the coverage!

src/controllers/preflight.js
radhikagpt1208: Most controllers read headers inline from `context.pathInfo.headers` (like `context.pathInfo.headers['x-api-key']`). A few purpose-specific helpers live in `src/support/utils.js`, but nothing for “get any header from pathInfo.” 
I added a shared helper in `src/support/http-headers.js` to get the header from the context and updated `preflight.js` and `consumers.js` controllers to use the shared helper instead of its local duplicate.

Broader cleanup of inline header reads and the remaining auth/promise-token duplication across controllers is tracked in [SITES-45204](https://jira.corp.adobe.com/browse/SITES-45204).

src/controllers/preflight.js
radhikagpt1208: Added.

src/controllers/preflight.js
radhikagpt1208: @GeezerPelletier I have renamed PROMISE_BASED_TYPES to PROMISE_BASED_AUTHORING_TYPES (in `src/utils/constants.js`). Please update the same in PR https://github.com/adobe/spacecat-api-service/pull/2478.

radhikagpt1208:
> Minor nit: the new ADR file is named `pf-request-id-spacecat-design.md` which doesn't match the naming convention used by the other decisions in this folder (`002-preflight-api-rest-redesign.md`, `003-preflight-entity-design.md`). Suggest renaming to `004-preflight-promise-token-header.md` (or similar) to keep things consistent and easy to scan in sequence.

The ADR was added earlier as `pf-request-id-spacecat-design.md`. I’ve renamed it to `docs/decisions/003-preflight-request-id-header.md` and updated the title to `ADR-003: Preflight Request-ID Header` so it lines up with the other decisions (001-…, 002-…, etc.).

src/controllers/preflight.js
radhikagpt1208: We return an object instead of a plain string to preserve the same shape the rest of the pipeline already expects, i.e., to keep the SQS promiseToken payload shape unchanged from the previous flow, so downstream consumers can keep reading `promiseToken.promise_token` without changes.

GeezerPelletier (approved):

src/controllers/preflight.js
radhikagpt1208: Previously, `getIMSPromiseToken()` returned the full IMS response object `({ promise_token, expires_in, token_type })`, and that object was passed through to SQS. Downstream consumers (spacecat-audit-worker, retrievePageAuthentication in scrape/autofix paths, etc.) already read `promiseToken.promise_token`.

With this PR, the token comes from the `x-promise-token` header (a plain string from POST `/auth/v2/promise`), but we still wrap it as `{ promise_token: value }` so the SQS contract stays unchanged and we don’t need a coordinated cross-repo worker change in the same release.

Agree it would be cleaner long term to either:
1. have resolvePromiseToken return the string and wrap at the SQS call site, or
2. rename the SQS field to match the shape
I’ve left that as-is in this PR to limit scope.

radhikagpt1208: This pattern predates this PR. On POST /preflight/jobs, siteId is optional - curl/Postman callers typically send only urls + step (and now `x-promise-token`), with no siteId.
The MFE does resolve siteId client-side when it can (match page URL against org sites), but only includes it when found `(...(siteId && { siteId }))`. When it can’t match, it sends URL only and relies on the API to resolve the site.
Hence: `findById` when a valid UUID siteId is present, otherwise `findByPreviewURL` from the request URL. ADR-002 / #2478 moves to site-scoped routes where siteId is required on the path and this fallback goes away.

bhellema (approved):

PR #2481, opened 2026-05-25
feat(auth): add route-scoped legacy API key handler for S2S-exempt endpoints

ravverma:
## Summary

Two admin endpoints — `POST /event/fulfillment` and `POST /slack/channels/invite-by-user-id` — are called by external systems that cannot be provisioned as IMS S2S consumers. This PR adds a route-scoped subclass of `LegacyApiKeyHandler` that limits legacy API key auth to exactly those two routes, following the same pattern as `ApiKeyImsHandler` from the IMS-to-JWT migration.

**The parent `LegacyApiKeyHandler` is intentionally kept in `AUTH_HANDLERS` for now.** Removal is a planned follow-up once all other consumers are confirmed migrated to S2S.

### `src/support/route-scoped-legacy-api-key-handler.js` (new)
- Subclasses `LegacyApiKeyHandler`; inherits the `'legacyApiKey'` handler name so `authInfo.getType()` is unchanged downstream
- `SCOPED_ROUTES` set contains only the two exempt routes
- Returns `null` immediately for any other route; delegates to `super.checkAuth()` for the two scoped routes
- Emits an info log on success as an operational signal (mirrors `ApiKeyImsHandler` pattern)

### `src/index.js`
- Import added for `RouteScopedLegacyApiKeyHandler`
- Handler inserted into `AUTH_HANDLERS` immediately before `LegacyApiKeyHandler`
- Comment updated to document both handlers and the planned removal of the parent

### `test/support/route-scoped-legacy-api-key-handler.test.js` (new)
- 14 tests: both scoped routes delegate to super, success log emitted, null result suppresses log, all out-of-scope routes return null (including wrong-method variants for the two scoped paths), edge cases (missing context, undefined route)

### `test/auth-handlers-order.test.js`
- New contract test asserts `RouteScopedLegacyApiKeyHandler` appears before `LegacyApiKeyHandler` in the parsed `AUTH_HANDLERS` array

## Decision doc

[adobe/mysticat-architecture#86](https://github.com/adobe/mysticat-architecture/pull/86) — `platform/decisions/route-scoped-legacy-api-key-handler.md`

## Test plan
- [ ] `test/support/route-scoped-legacy-api-key-handler.test.js` — 14 unit tests, all passing
- [ ] `test/auth-handlers-order.test.js` — order contract test passes (3 tests total)
- [ ] Verify `POST /event/fulfillment` and `POST /slack/channels/invite-by-user-id` authenticate with `ADMIN_API_KEY` in a dev environment
- [ ] Verify other routes (e.g. `GET /sites`) are unaffected by the new handler

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -107,6 +107,7 @@ import AiVisibilityController from './controllers/ai-visibility.js';
 import SemrushController from './controllers/semrush.js';
 import GitHubWebhookHmacHandler from './support/github-webhook-hmac-handler.js';
 import ApiKeyImsHandler from './support/api-key-ims-handler.js';
+import RouteScopedLegacyApiKeyHandler from './support/route-scoped-legacy-api-key-handler.js';
 
 const uuidRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
 
@@ -394,10 +395,17 @@ const { WORKSPACE_EXTERNAL } = SLACK_TARGETS;
 //    working without re-introducing a global IMS auth backdoor.
 //  - AdobeImsHandler: legacy global IMS path; kept for routes still on IMS auth
 //    (e.g. Auto-Fix). To be removed once all consumers are JWT-migrated.
-//  - ScopedApiKeyHandler / LegacyApiKeyHandler: standard API-key auth paths.
+//  - ScopedApiKeyHandler: scoped API-key auth for Import-as-a-Service.
+//  - RouteScopedLegacyApiKeyHandler: route-scoped legacy API key handler for
+//    POST /event/fulfillment and POST /slack/channels/invite-by-user-id — the two
+//    admin endpoints whose callers cannot migrate to S2S. Returns null for all
+//    other routes. Must run BEFORE LegacyApiKeyHandler so the scoped handler owns
+//    those two routes explicitly. LegacyApiKeyHandler is kept for now and will be
+//    removed in a follow-up once all other consumers are confirmed migrated.
+//  - LegacyApiKeyHandler: legacy catch-all; to be removed in follow-up.
 // When adding a new path-scoped handler, place it in the same position (after
 // SkipAuthHandler, before the path-agnostic handlers) to preserve early-bail.
-// AUTH_HANDLERS order is enforced by test/auth-handlers.test.js.
+// AUTH_HANDLERS order is enforced by test/auth-handlers-order.test.js.
 const AUTH_HANDLERS = [
   SkipAuthHandler,
   GitHubWebhookHmacHandler,
@@ -405,6 +413,7 @@ const AUTH_HANDLERS = [
   ApiKeyImsHandler,
   AdobeImsHandler,
   ScopedApiKeyHandler,
+  RouteScopedLegacyApiKeyHandler,
   LegacyApiKeyHandler,
 ];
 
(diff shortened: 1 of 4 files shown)


MysticatBot-Dev (approved):
Hey @ravverma,

## Strengths

- **Clean route-matching design** (`src/support/route-scoped-legacy-api-key-handler.js:17-19`): The `SCOPED_ROUTES` Set with exact `"METHOD /path"` composite keys gives O(1) lookup with zero regex-bypass risk. Method-mismatch and prefix-confusion attacks are impossible by construction.
- **Minimal, readable handler logic** (`src/support/route-scoped-legacy-api-key-handler.js:42-51`): Guard clause, delegate, conditional log. The entire file is understandable in under a minute. No new credential-handling logic introduced - actual key validation remains in the parent.
- **Incremental auth narrowing strategy**: Scoping first, then removing the catch-all is the right decomposition for this migration. It turns a risky flag-day removal into a safe two-step narrowing.
- **Handler ordering contract test** (`test/auth-handlers-order.test.js:108-124`): Protecting the positional-dispatch invariant with an explicit test prevents silent regressions if someone reorders the array in a future refactor.
- **Thorough test coverage** (`test/support/route-scoped-legacy-api-key-handler.test.js`): 14 tests covering both scoped routes, wrong-method variants, out-of-scope routes, edge cases (missing context, undefined route), and log behavior. Good test-to-code ratio and behavior-focused assertions.
- **Null-safe route extraction** (`src/support/route-scoped-legacy-api-key-handler.js:43`): `context?.pathInfo?.route` gracefully handles missing context without throwing, returning null and falling through safely.

## Issues

### Minor (Nice to Have)

1. `src/support/route-scoped-legacy-api-key-handler.js:47` - **Log message lacks the authenticated route.** The info log says `'request authenticated via route-scoped legacy API key handler'` but does not include which of the two scoped routes was matched. When investigating an auth issue in production, you will grep this message and then need to cross-reference with the request log to identify the endpoint. Including the route costs nothing and saves a log-correlation hop:
   ```js
   this.log(`request authenticated via route-scoped legacy API key handler [${context.pathInfo.route}]`, 'info');
   ```

2. `test/support/route-scoped-legacy-api-key-handler.test.js` - **No test for exception propagation from `super.checkAuth`.** The production code has no try/catch around `await super.checkAuth(request, context)`. If the parent throws (e.g., missing API key config), the exception propagates to the auth framework. This is almost certainly correct behavior, but there is no test documenting that contract. If someone later adds a try/catch with a fallback, the suite would not catch the regression. Suggestion:
   ```js
   it('propagates exceptions from super.checkAuth without catching', async () => {
     superCheckAuthStub.rejects(new Error('key validation failed'));
     const ctx = { pathInfo: { route: 'POST /event/fulfillment' } };
     await expect(handler.checkAuth({}, ctx)).to.be.rejectedWith('key validation failed');
   });
   ```

## Recommendations

- When the follow-up removes `LegacyApiKeyHandler` from AUTH_HANDLERS, evaluate whether `RouteScopedLegacyApiKeyHandler` should stand on its own (implementing `checkAuth` with a direct key comparison) rather than subclassing a parent that no longer participates in the chain. The inheritance is the right call today (behavioral parity, zero risk of drift), but once the parent is gone the subclass relationship becomes pure coupling with no benefit. Consider documenting this in the existing TODO block.

## Assessment

Ready to merge? Yes

The implementation is clean, well-tested, and architecturally sound. It narrows the legacy auth surface incrementally, follows an established team pattern (ApiKeyImsHandler), and is protected by order-contract tests. No new failure modes, no new attack surface, no breaking changes. The two minor findings are quality-of-life improvements, not blockers.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 46s | <b>Cost:</b> $4.30 | <b>Commit:</b> <code>cdbce6ca4d3a40161aa6ec4b3a58f8f1083e6927</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2482, opened 2026-05-25
feat(slack): add add-oae-stage-domain command for OAE stage domain on…

nit23uec:
…boarding

Adds a new internal Slack command that onboards customer stage domains for Optimize at Edge testing, bypassing the same-registered-domain restriction enforced by the REST API to support cross-domain staging scenarios.

## Related Issues

Thanks for contributing!


src/support/slack/commands/add-oae-stage-domain.js
```diff
@@ -0,0 +1,169 @@
...
+      /* eslint-disable no-await-in-loop */
+      for (const { domain, stageBaseURL } of domainEntries) {
+        let stageSite = await Site.findByBaseURL(stageBaseURL);
+        if (!stageSite) {
+          stageSite = await Site.create({
+            baseURL: stageBaseURL,
+            organizationId,
+          });
+        } else if (stageSite.getOrganizationId() !== organizationId) {
+          throw new Error(`Stage domain \`${domain}\` already belongs to a different organization.`);
```
jindaliiita: Site.create() commits to the DB immediately inside the loop. If a later iteration throws (org mismatch on domain B, or Tokowaka returns no API keys), execution jumps to the catch block and
   site.save() is never called. Domain A's site record is now permanently in the DB with no corresponding stagingDomains entry on the prod site

src/support/slack/commands/add-oae-stage-domain.js
```diff
@@ -0,0 +1,169 @@
...
+            throw new Error(`Failed to provision API key for stage domain \`${domain}\`: createMetaconfig returned no API keys.`);
+          }
+        } else {
+          await tokowakaClient.updateMetaconfig(
+            stageBaseURL,
+            stageSite.getId(),
+            {},
+            { lastModifiedBy, isStageDomain: true },
+          );
+          metaconfig = await tokowakaClient.fetchMetaconfig(stageBaseURL);
```
jindaliiita, line 147: can we add this same check here also
```
 if (!metaconfig?.apiKeys?.length) {
            throw new Error(`Failed to provision API key for stage domain \`${domain}\`: createMetaconfig returned no API keys.`);
          }
```

src/support/slack/commands/add-oae-stage-domain.js
```diff
@@ -0,0 +1,169 @@
...
+    if (!siteInput || !domainsArg) {
+      await say(baseCommand.usage());
+      return;
+    }
+
+    if (extraArgs.length > 0) {
+      await say(`:warning: Too many arguments. Use comma-separated domains, not spaces.\n${baseCommand.usage()}`);
+      return;
+    }
+
```
jindaliiita, line 61: Should we add a validation to ensure that the stage domain is not equal to the prod domain?

jindaliiita:
LGTM. Thanks

jindaliiita (approved):

PR #2485, opened 2026-05-25
fix(auth): use context.log in auth handler subclasses for per-request logging

ravverma:
this.log() in AbstractHandler uses the logger captured at handler construction time (first request), causing all subsequent requests to log with a stale context. Switch to context.log in ApiKeyImsHandler and RouteScopedLegacyApiKeyHandler so the per-request logger is used.

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/support/api-key-ims-handler.js
+++ b/src/support/api-key-ims-handler.js
@@ -77,7 +77,7 @@ export default class ApiKeyImsHandler extends AdobeImsHandler {
       // claim for exactly this reason). Normalising to trial_email (the real
       // address) would create a format mismatch that makes existing keys
       // invisible on GET and breaks cross-path consistency (ASO-607).
-      this.log('api-key request authenticated via scoped IMS handler - JWT migration pending', 'info');
+      context.log.info('[ims] api-key request authenticated via scoped IMS handler - JWT migration pending');
     }
     return result;
   }
--- a/src/support/route-scoped-legacy-api-key-handler.js
+++ b/src/support/route-scoped-legacy-api-key-handler.js
@@ -45,7 +45,7 @@ export default class RouteScopedLegacyApiKeyHandler extends LegacyApiKeyHandler 
     }
     const result = await super.checkAuth(request, context);
     if (result) {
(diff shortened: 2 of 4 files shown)


MysticatBot-Dev (approved):
Hey @ravverma,

## Strengths

- **Correct fix for a real per-request logging bug** (`src/support/api-key-ims-handler.js:80`, `src/support/route-scoped-legacy-api-key-handler.js:48`): `this.log()` captures the logger at handler construction time, so in a long-lived handler instance subsequent requests log with stale request context (wrong trace ID, wrong invocation metadata). Switching to `context.log` ensures each request uses its own logger. Meaningful correctness improvement for observability.
- **Added structured log prefixes** (`[ims]`, `[legacyApiKey]`): Makes it trivial to filter auth log lines by mechanism in log aggregators.
- **Minimal, focused diff**: Two lines changed, both doing the same thing for the same reason. No scope creep.

## Issues

### Minor (Nice to Have)

**Log prefix naming inconsistency** - `src/support/api-key-ims-handler.js:80`, `src/support/route-scoped-legacy-api-key-handler.js:48`

The prefix `[ims]` is lowercase while `[legacyApiKey]` uses camelCase. Not blocking, but aligning on one convention (e.g., all lowercase like `[legacy-api-key]`, or all camelCase) would make grep-based log filtering more predictable.

## Recommendations

- Consider updating the test suites for both handlers to assert that `context.log.info()` is invoked with the expected message, rather than relying on the old `this.log()` stubs. This would catch regressions if the logging mechanism is changed again.

## Assessment

Ready to merge? Yes

Clean, well-scoped fix for a real per-request logging bug. The change is obviously correct - `context.log` is the per-request logger, `this.log()` is the construction-time logger - and the risk surface is negligible (log calls on a success path only).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 40s | <b>Cost:</b> $1.81 | <b>Commit:</b> <code>87d21ac21f0701980fc55711bbc95392276ad1fe</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2486, opened 2026-05-26
feat(serenity): denormalize semrushLocationName + local-dev env docs

rainer-friederich:
## Summary

Follow-up to #2467 (Semrush AIO REST proxy). Two independent changes, both surfaced while validating the proxy end-to-end against the elmo UI:

- **`handleListProjects` now denormalizes `semrushLocationName`** onto each row (e.g. `2840 → "United States"`). The UI was previously falling back to displaying the raw Semrush `location_id` integer in the Market combobox because the controller only returned the numeric id. Reverse-lookup uses the existing `iso31661NumericToAlpha2` table + `Intl.DisplayNames` — no new deps. Includes a zero-pad fix for 2-digit ISO numerics like `036` (Australia).
- **`.env.example` expanded** with the local-dev requirements that today's session uncovered as undocumented or actively misleading:
  - `IMS_HOST` corrected from `https://dummy-ims.example.com` to the bare hostname `dummy-ims.example.com`. `ImsClient` prepends `https://` internally, so a scheme-prefixed value yields `https://https://.../ims/profile/v1` → `getaddrinfo ENOTFOUND https`.
  - `AUTH_PUBLIC_KEY_B64` + `AUTH_HANDLER_IMS` documented as commented stubs with `vault kv get` commands — without these the entire auth chain silently 401s.
  - `ENABLE_CORS` + `CORS_ALLOWED_ORIGINS` documented for elmo at `https://localhost:3000`.
  - `LLMO_HLX_API_KEY` and `TOKOWAKA_SITE_CONFIG_BUCKET` documented (used by `/sites/:id/llmo/sources/*` and any controller through `tokowakaClient`).

Paired with elmo PR adobe/project-elmo-ui#1815 which switches `/serenity/*` calls to send the raw IMS bearer.

## Test plan

- [x] `npx mocha test/support/serenity/handlers/projects.test.js test/controllers/serenity.test.js` — 109/109 passing, includes new denormalization test (2-digit/3-digit ISO numerics + out-of-range locationId)
- [x] `npm run lint` — exit 0
- [x] Local E2E verified: `curl /v2/orgs/:org/brands/:brand/serenity/projects` returns rows with `semrushLocationName` populated (`"United States"`, `"Australia"`, `"France"`)
- [x] No public API contract regressions — response shape gains one field; existing fields unchanged

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -26,10 +26,44 @@ USER_API_KEY=api_key_for_user_requests
 ADMIN_API_KEY=api_key_for_admin_requests
 
 # ── IMS client (eagerly initialized, dummy values are fine) ───────────────────
-IMS_HOST=https://dummy-ims.example.com
+# IMS_HOST is just the hostname — NO scheme. ImsClient builds URLs as
+# `https://${IMS_HOST}${endpoint}`, so a value like `https://ims-...`
+# yields `https://https://ims-.../...` and fails with `ENOTFOUND https`.
+# Real value for stage: `ims-na1-stg1.adobelogin.com` (sourced from Vault in prod).
+IMS_HOST=dummy-ims.example.com
 IMS_CLIENT_ID=dummy-client-id
 IMS_CLIENT_CODE=dummy-client-code
 IMS_CLIENT_SECRET=dummy-client-secret
+
+# ── Auth handler config (REQUIRED for any authenticated request) ──────────────
+# Without these, JwtHandler / AdobeImsHandler return null on every request and
+# the auth chain 401s with a generic "Unauthorized". Both are Vault-backed in
+# prod (dx_mysticat/<env>/api-service); fetch values for local dev with:
+#   vault kv get -field=AUTH_PUBLIC_KEY_B64 dx_mysticat/dev/api-service
+#   vault kv get -field=AUTH_HANDLER_IMS    dx_mysticat/dev/api-service
+# Uncomment + paste the values to use real elmo IMS / JWT tokens locally.
+# AUTH_PUBLIC_KEY_B64=
+# AUTH_HANDLER_IMS={"name":"ims-na1-stg1","discovery":{"jwks_uri":"https://ims-na1-stg1.adobelogin.com/ims/keys"}}
+
+# ── Local CORS (needed for elmo at https://localhost:3000) ────────────────────
+# `localCORSWrapper` is a no-op unless ENABLE_CORS=true. CORS_ALLOWED_ORIGINS is
+# a comma-separated list — must include the elmo dev URL exactly (including
+# scheme + port).
+# ENABLE_CORS=true
+# CORS_ALLOWED_ORIGINS=https://localhost:3000
+
+# ── LLMO Helix API (sheets API, used by /sites/:id/llmo/sources/*) ────────────
+# Bearer token for hlx.live admin operations against the LLMO data folder.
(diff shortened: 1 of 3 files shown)


MysticatBot-Dev (approved):
Hey @rainer-friederich,

## Strengths

- **`src/support/serenity/handlers/projects.js:89-99`** - `locationNameFromId` is a clean inverse of the existing `resolveLocation`, reusing the same `iso-3166` tables and `ENGLISH_REGION_NAMES` formatter already in scope. No new dependencies, consistent style, and the defensive guards (`Number.isInteger`, range check, null-safe returns) handle edge inputs gracefully.

- **`src/support/serenity/handlers/projects.js:95`** - The `padStart(3, '0')` for ISO numeric codes (catching the 2-digit case like AU = 036) is the key correctness detail here, and it is both implemented and tested.

- **`test/support/serenity/handlers/projects.test.js:115-134`** - Good boundary coverage: 2-digit ISO numeric needing zero-pad (AU), standard 3-digit (US, FR), and an unmapped ID returning `null`. The test asserts on observable behavior (the response shape) rather than implementation details.

- **`.env.example`** - The IMS_HOST correction from `https://dummy-ims.example.com` to `dummy-ims.example.com` with the explanation of the double-scheme failure mode is genuinely useful onboarding documentation. The vault commands for auth config will save developers real debugging time.

## Issues

No issues found. The change is additive, non-breaking, well-scoped, and correctly tested.

## Recommendations

1. **`Intl.DisplayNames.of()` edge behavior** - On some Node/ICU versions, `Intl.DisplayNames.of(alpha2)` may return the input code itself (e.g. `"BQ"`) rather than `undefined` for structurally valid but uncommon region codes. The `|| null` fallback only catches falsy returns. Practical risk is very low since `iso-3166` only contains officially assigned codes with full ICU coverage, but a `result === alpha2 ? null : result` guard would make it airtight if you ever want to harden it.

2. **Consider testing with non-integer input** - `locationNameFromId` correctly handles `null`/`undefined`/string via the `Number.isInteger` guard, but there is no test exercising this path through `handleListProjects`. A single additional case (e.g., `semrushLocationId: null`) would document the contract for future maintainers.

## Assessment

Ready to merge? Yes

Well-scoped feature addition that eliminates a client-side lookup table dependency, uses existing infrastructure with no new deps, and includes appropriate test coverage for the interesting edge cases.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 44s | <b>Cost:</b> $2.42 | <b>Commit:</b> <code>8e52211c97cb43438ec52c0af0d912ad34122860</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Superseded by #2488 (docs-only change). 

Walking back the server-side `semrushLocationName` denormalization: derivation belongs in the client, not on this surface. The `2000 + ISO 3166-1 numeric` formula only holds for countries — moving it server-side would lock us into an API contract that doesn't extend to sub-national markets (cities, regions, postal codes) without the official Google Ads geotargets dataset. The lookup now lives in elmo's `useSerenityProjects.ts` (adobe/project-elmo-ui#1815), with an explicit TODO captured for the sub-national rework.

The `.env.example` cleanup from this PR is preserved in #2488 (now the only commit there).

PR #2487, opened 2026-05-26
fix(slack): correct agentic-traffic db-status (backfills + raw evidence) and cdn-logs-status region

akshaymagapu:
Fixes two Slack status commands that gave false results for non-us-east-1 sites and for backfilled dates.

## `check agentic traffic db status`
1. **Backfills were reported missing.** The fetch used `projected_at < date+7`, so a projection that ran weeks after the traffic day (a backfill) fell outside the window and the date showed `NO_DB_ROWS_FOR_DATE` even though the data was fully imported. Now we match the exact date via jsonb metadata containment — `metadata->dailyRefreshDates cs ["<date>"]` (daily) and `weeklyRefreshWeeks cs ["<weekStart>"]` (weekly) — with only a `projected_at >= date` lower bound. Backfills are credited, and result sets stay small for all-sites sweeps (server-side filter, still chunked by `scope_prefix`). Validated the `cs` filter against the live prod PostgREST before implementing.
2. **Raw per-date evidence** (original scope): a raw `projection_audit` row carries no per-date metadata, so it can't prove the requested date was covered. Raw is credited via daily-implies-raw (a daily-for-X row proves raw-for-X), plus a bounded day-after fallback (`projected_at >= date+1`, positive `output_count`) for the "raw OK, daily not run yet" case.

## `check cdn logs status`
The per-region aggregate buckets (`spacecat-{env}-cdn-logs-aggregates-{region}`) were listed with the Lambda's us-east-1 S3 client, so eu-west-1 sites `ListObjectsV2` 301'd and showed as error/empty. Now reuses the default client for the runtime region and lazily creates one client **per region** (not per site).

## Validation
- 58 tests passing across both commands (incl. a backfill regression test); lint clean.
- Coverage above the 90% gate: cdn-logs-status 100%, db-status 98.8% lines / 94% branch.


MysticatBot (changes requested):
Hey @akshaymagapu,

## Strengths

- **Correct root-cause fix for backfills** (`check-agentic-traffic-db-status.js`): Switching from a bounded `projected_at` time window to jsonb metadata containment (`contains 'metadata->dailyRefreshDates'`) is the right structural approach. It matches on semantic evidence (which dates were actually refreshed) rather than temporal proximity, so late backfills are no longer invisible.
- **"Daily implies raw" shortcut** (`check-agentic-traffic-db-status.js:195-199`): Deriving raw success from a daily-refresh row is sound given the stated invariant that daily refresh is a post-success downstream of raw import. This eliminates a class of false-positive "raw missing" reports.
- **Per-region S3 client cache** (`check-cdn-logs-status.js:267-275`): Lazy instantiation of one `S3Client` per region, cached in a Map and seeded with the runtime-region client, is clean and proportionate. Not per-site, not pre-created for all known regions.
- **Comprehensive edge-case tests**: The backfill regression test, the "raw before dateStr+1 does not count" boundary test, the "raw OK but daily missing" scenario, and the CDN region-reuse verification all exercise the critical behavioral boundaries that motivated this fix.
- **PostgREST mock redesign** (`check-agentic-traffic-db-status.test.js:36-97`): The chainable-thenable mock now mirrors actual PostgREST builder semantics (supporting `contains`, `gt`, deferred resolution) more faithfully than the prior immediate-resolve approach.

## Issues

### Important (Should Fix)

**1. Comment/code discrepancy on raw fallback window span**

`src/support/slack/commands/check-agentic-traffic-db-status.js` - the comment at the raw fallback query says "credit it only within the day-after window with positive output," but the actual window spans from `dateStr + 1 day` to `dateStr + 7 days` (via `PROJECTION_LOOKAHEAD_DAYS`). A future reader will interpret "day-after window" as a 1-day span and may "fix" the code to match the comment, breaking the fallback for weekend delays or slow pipelines.

Either:
- (a) Update the comment to say something like "credit it within a 7-day post-traffic window (covers weekend and pipeline delays)" if the wide window is intentional, or
- (b) Tighten the upper bound to `addUtcDays(targetDate, 2)` if a 1-day window was truly the intent.

The daily-implies-raw shortcut covers the common backfill case regardless, so this raw fallback is only relevant for the "raw OK, daily not yet run" scenario - but the intent should be unambiguous.

### Minor (Nice to Have)

**2. Sequential queries where parallel is possible**

`src/support/slack/commands/check-agentic-traffic-db-status.js` - The three per-chunk PostgREST queries (daily, weekly, raw) are independent and issued sequentially. For a typical fleet (~250 sites, 2 chunks), this is 6 round-trips instead of 2 parallel batches. Since this is a Slack command with a human waiting, the latency multiplier is noticeable.

Fix: `const [dailyRows, weeklyRows, rawRows] = await Promise.all([query(...), weeklyExpected ? query(...) : Promise.resolve([]), query(...)]);`

**3. `PROJECTION_LOOKAHEAD_DAYS` naming is a holdover**

`src/support/slack/commands/check-agentic-traffic-db-status.js` - This constant now only bounds the raw fallback window, not a general "lookahead." Renaming to something like `RAW_FALLBACK_WINDOW_DAYS` would communicate what it controls and why 7 was chosen.

## Recommendations

- Parallelize the per-chunk queries (daily/weekly/raw are data-independent within a chunk) via `Promise.all` to recover the latency cost of splitting one query into three.
- Consider adding a test for the `rawWindowEnd` upper boundary (a raw row with `projected_at >= dateStr + 7 days` should be excluded) to lock down the intent of the window.

## Assessment

Ready to merge? With fixes
Reasoning: The correctness fixes are sound, well-tested (98%+ coverage, regression tests, boundary cases), and the CDN region fix is clean. The one Important finding is a comment-vs-code discrepancy that creates maintenance risk - a future contributor could incorrectly tighten the raw fallback window by reading the "day-after" comment literally. Clarifying the intent is a small change that protects the fix long-term.

Note: CI checks are currently failing (branch-deploy, deploy-stage, it-postgres, semantic-release) - this should be investigated before merge.

## Next Steps

1. Clarify the raw fallback window comment (the Important item).
2. Optionally parallelize the per-chunk queries and rename the constant (Minor items, can be follow-up).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 31s | <b>Cost:</b> $4.78 | <b>Commit:</b> <code>5db3da32c5a08396656110a2a3ee4f07b885eedc</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

akshaymagapu:
Thanks! Addressed in `e2518f88`:

- **Raw fallback window comment (Important):** clarified it's a 7-day post-traffic window (covers weekend/pipeline delays), and renamed `PROJECTION_LOOKAHEAD_DAYS` → `RAW_FALLBACK_WINDOW_DAYS` so the constant says what it bounds.
- **Parallelized the per-chunk queries:** daily/weekly/raw now run via `Promise.all`. The chunk loop stays sequential, so peak concurrency is fixed at **3 in-flight queries** regardless of fleet size (500 sites = 4 sequential chunks × 3) — no pool risk.
- **Boundary test added:** a raw row at `dateStr + RAW_FALLBACK_WINDOW_DAYS` is excluded from the fallback.

16 tests passing, lint clean, coverage above gate (db-status 98.8%/94%, cdn-logs-status 100%).

On the failing CI checks (branch-deploy/deploy-stage/it-postgres/semantic-release): those are deploy/infra stages, not lint/unit/coverage, and don't appear related to these command changes (likely tied to the large `main` merge/env). Will confirm before merge.

MysticatBot (approved):
Hey @akshaymagapu,

## Strengths

- **All prior findings addressed cleanly** (commit `e2518f88`): Raw fallback window comment clarified, constant renamed to `RAW_FALLBACK_WINDOW_DAYS`, queries parallelized via `Promise.all` with peak concurrency fixed at 3 per chunk. Proportional and precise responses to each item.
- **jsonb containment over time-window heuristic** (`check-agentic-traffic-db-status.js:88-108`): Switching from a `projected_at` time window to `contains('metadata->dailyRefreshDates', ...)` is the correct structural fix. It queries by semantic evidence (which dates were actually refreshed) rather than temporal proximity, decoupling correctness from pipeline scheduling latency.
- **Clean query abstraction** (`check-agentic-traffic-db-status.js:179-185`): The `query(decorate)` helper factors out the common PostgREST base and lets each handler add domain-specific predicates via a callback. Easy to read and extend.
- **Narrowed select clause** (line 181): Only fetching `scope_prefix` instead of the full row - less data over the wire for what is essentially a set-membership check across hundreds of sites.
- **Per-region S3 client cache** (`check-cdn-logs-status.js:267-275`): Lazy instantiation keyed by region, seeded with the runtime client. Bounded by the number of distinct AWS regions in the fleet (single digits). Proportionate - not per-site, not pre-allocated.
- **Robust input validation chain**: Date input passes through regex validation, `new Date()` parsing, and a round-trip format check before reaching the query builder. No injection vector into jsonb containment.
- **Thorough boundary testing** (`check-agentic-traffic-db-status.test.js`): Four new boundary tests (backfill regression, raw before dateStr+1, raw at dateStr+7, raw with positive output_count in-window) form a complete specification of the raw fallback logic. Each documents the domain invariant it exercises.
- **Thenable mock redesign** (`check-agentic-traffic-db-status.test.js:36-87`): Faithfully mirrors PostgREST builder semantics (deferred resolution, per-handler filtering, jsonb containment parsing) without over-coupling to implementation internals.

## Assessment

Ready to merge? Yes
Reasoning: The correctness fixes are architecturally sound (semantic date matching over temporal heuristics), well-tested (98.8% lines, 94% branch coverage, boundary tests on both sides of the raw fallback window), and all prior review findings have been cleanly addressed. No security concerns - the attack surface is narrow (internal Slack command) with properly parameterized queries and trusted config data. No convention violations.

Note: CI build/validate steps are still pending - confirm they pass before merging. The previously-failing deploy/infra stages appear unrelated to these changes.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 46s | <b>Cost:</b> $3.48 | <b>Commit:</b> <code>e2518f8810060be2b5bc05f8bed0c19d283fdda3</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2488, opened 2026-05-26
docs(env): document local-dev auth handler, CORS, and IMS_HOST format

rainer-friederich:
## Summary

Local-dev validation of #2467 (Semrush AIO REST proxy) surfaced four undocumented (or actively misleading) requirements in `.env.example` — all of which silently 401 the entire auth chain when missing or malformed. None of these are introduced by this PR; the gap predates the proxy work but the proxy was the first thing to exercise the local stack end-to-end against a real elmo client.

- **`IMS_HOST` corrected** from `https://dummy-ims.example.com` (misleading) to the bare hostname `dummy-ims.example.com`. The `@adobe/spacecat-shared-ims-client` builds URLs as `https://${IMS_HOST}${endpoint}`, so a scheme-prefixed value yields `https://https://.../ims/profile/v1` → DNS resolver tries `https` as the hostname → `getaddrinfo ENOTFOUND https` → `AdobeImsHandler` silently returns null → generic 401. A 4-line comment block above the value explains the trap.
- **`AUTH_PUBLIC_KEY_B64` + `AUTH_HANDLER_IMS`** documented as commented stubs with `vault kv get -field=... dx_mysticat/dev/api-service` commands. Without these `JwtHandler` and `AdobeImsHandler` return null on every request.
- **`ENABLE_CORS` + `CORS_ALLOWED_ORIGINS`** documented for browser-based callers like elmo at `https://localhost:3000`. `localCORSWrapper` is a no-op unless `ENABLE_CORS=true`.
- **`LLMO_HLX_API_KEY`** + **`TOKOWAKA_SITE_CONFIG_BUCKET`** documented — required by `/sites/:id/llmo/sources/*` and any controller using `tokowakaClient` (sites, suggestions, llmo, edge-routing-utils).

Each section includes the Vault-fetch command inline so a new developer can bootstrap their `.env` without spelunking through Vault paths.

This supersedes #2486 — that PR also included a server-side `semrushLocationName` denormalization which we decided to compute client-side in elmo instead (see adobe/project-elmo-ui#1815). Closing #2486 in favor of this docs-only change.

## Test plan

- [x] `npm run lint` — exit 0
- [x] No runtime behavior changes — `.env.example` is documentation only

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -39,6 +39,7 @@ IMS_CLIENT_SECRET=dummy-client-secret
 # Without these, JwtHandler / AdobeImsHandler return null on every request and
 # the auth chain 401s with a generic "Unauthorized". Both are Vault-backed in
 # prod (dx_mysticat/<env>/api-service); fetch values for local dev with:
+#   (one-time: vault login -method=oidc  — opens browser for Okta FastPass)
 #   vault kv get -field=AUTH_PUBLIC_KEY_B64 dx_mysticat/dev/api-service
 #   vault kv get -field=AUTH_HANDLER_IMS    dx_mysticat/dev/api-service
 # Uncomment + paste the values to use real elmo IMS / JWT tokens locally.
--- a/.env.example
+++ b/.env.example
@@ -26,10 +26,44 @@ USER_API_KEY=api_key_for_user_requests
 ADMIN_API_KEY=api_key_for_admin_requests
 
 # ── IMS client (eagerly initialized, dummy values are fine) ───────────────────
-IMS_HOST=https://dummy-ims.example.com
+# IMS_HOST is just the hostname — NO scheme. ImsClient builds URLs as
+# `https://${IMS_HOST}${endpoint}`, so a value like `https://ims-...`
+# yields `https://https://ims-.../...` and fails with `ENOTFOUND https`.
+# Real value for stage: `ims-na1-stg1.adobelogin.com` (sourced from Vault in prod).
+IMS_HOST=dummy-ims.example.com
 IMS_CLIENT_ID=dummy-client-id
 IMS_CLIENT_CODE=dummy-client-code
 IMS_CLIENT_SECRET=dummy-client-secret
+
+# ── Auth handler config (REQUIRED for any authenticated request) ──────────────
+# Without these, JwtHandler / AdobeImsHandler return null on every request and
+# the auth chain 401s with a generic "Unauthorized". Both are Vault-backed in
+# prod (dx_mysticat/<env>/api-service); fetch values for local dev with:
+#   vault kv get -field=AUTH_PUBLIC_KEY_B64 dx_mysticat/dev/api-service
+#   vault kv get -field=AUTH_HANDLER_IMS    dx_mysticat/dev/api-service
+# Uncomment + paste the values to use real elmo IMS / JWT tokens locally.
+# AUTH_PUBLIC_KEY_B64=
+# AUTH_HANDLER_IMS={"name":"ims-na1-stg1","discovery":{"jwks_uri":"https://ims-na1-stg1.adobelogin.com/ims/keys"}}
+
+# ── Local CORS (needed for elmo at https://localhost:3000) ────────────────────
+# `localCORSWrapper` is a no-op unless ENABLE_CORS=true. CORS_ALLOWED_ORIGINS is
+# a comma-separated list — must include the elmo dev URL exactly (including
+# scheme + port).
+# ENABLE_CORS=true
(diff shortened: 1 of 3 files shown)


MysticatBot-Dev (approved):
Hey @rainer-friederich,

## Strengths

- **.env.example:29-32** - The IMS_HOST correction is excellent. The inline explanation of how `ImsClient` builds URLs as `https://${IMS_HOST}${endpoint}` makes the "no scheme" requirement immediately clear, and documenting the exact failure mode (`ENOTFOUND https`) saves developers significant debugging time.
- **.env.example:38-42** - Vault fetch commands inline with each variable is the right call. New developers can bootstrap their `.env` without spelunking through Vault paths or guessing field names.
- **.env.example:35-41** - All sensitive credentials are documented as commented-out stubs, reducing the risk of accidental commits with real values while still making the required shape visible.
- **.env.example:48-52** - Tying CORS configuration to the specific `localCORSWrapper` no-op behavior bridges config to code clearly.
- **Overall** - The progressive disclosure pattern (section header, behavior explanation, vault command, commented stub) is consistent across all new sections and scales well.

## Issues

### Minor (Nice to Have)

1. **.env.example:37** - Test file inconsistency. `test/it/env.js` (and a few unit test files) still use `IMS_HOST: 'https://dummy-ims.example.com'` with the scheme. The tests likely pass because the HTTP layer is mocked and the malformed URL never hits DNS resolution, so this isn't a bug in this PR - but the inconsistency could confuse developers who read both files. Worth a follow-up to align test env values with the documented format.

2. **.env.example:36-37** - The section header says "REQUIRED for any authenticated request" but the values are commented out. This is standard `.env.example` practice, but adding a parenthetical like "(uncomment for real IMS/JWT auth locally)" would reduce the cognitive gap between "REQUIRED" and "commented out by default".

3. **.env.example:38-42** - Vault commands lack a prerequisite note. Developers unfamiliar with Vault will hit auth errors before discovering they need `vault login -method=oidc` first. A single line like `# (Requires vault login first; see workspace README)` above the first vault command would save a round-trip.

## Recommendations

- Consider adding startup validation in `src/index.js` that checks whether `IMS_HOST` starts with `http` and logs a clear error, since the current failure mode (silent 401 via `ENOTFOUND`) is notoriously hard to debug.
- As `.env.example` grows, a `docs/local-development.md` with a higher-level bootstrapping flow (linking back to `.env.example` for variable definitions) would keep the example file from becoming a full getting-started guide.

## Assessment

Ready to merge? Yes
Reasoning: This is a documentation-only change that significantly improves the local development experience by surfacing four previously undocumented requirements that cause silent 401s. All documented behaviors are accurate. The minor items are polish - none block merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-sonnet-4-5-20250929-v1:0 | <b>Duration:</b> 1m 13s | <b>Cost:</b> $2.23 | <b>Commit:</b> <code>3555374866c357958658a2f6a1f4ef0ce5d483bb</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Thanks for the review @MysticatBot — addressed in commit d1cc734a.

**Finding 1 — test file inconsistency**

Fixed in `test/it/env.js` and `test/support/email-service.test.js`. Both now use the bare hostname format. The email-service test's roundtrip assertion on line 126 updates in lockstep with the setup. Confirmed `npx mocha test/support/email-service.test.js` still 13/13 passing.

Left `test/controllers/brands.test.js:151` alone — that's a different variable (`BRAND_IMS_HOST`), not `IMS_HOST`.

**Finding 2 — "REQUIRED" wording vs commented-out values**

Acknowledged but not taking. Standard `.env.example` practice carries enough convention that "section header explains importance, body shows shape, commented to prevent accidental copy" is fluent on its own. A parenthetical risks reading as redundant — the next maintainer to skim through will treat it as noise.

**Finding 3 — missing `vault login` prereq**

Fixed. Added a one-line note above the first `vault kv get` command:

```
#   (one-time: vault login -method=oidc  — opens browser for Okta FastPass)
```

Saves the round-trip you identified.

**Recommendation: startup validation for `IMS_HOST` scheme**

Good idea, deferred to a separate PR. It crosses the boundary from `.env.example` documentation into runtime code in `src/index.js` — different reviewer audience and different blast radius. Captured as a follow-up.

**Recommendation: `docs/local-development.md`**

Same call — defer until `.env.example` warrants it. At current size the example is still readable end-to-end; promoting to a separate doc costs a discovery hop without earning anything yet.

PR #2489, opened 2026-05-26
feat: add POST /prompts/check endpoint for batch existence check | LLMO-5156

jjenscodee:
## Summary

- Adds `checkPromptsExist` storage function in `prompts-storage.js` that calls the new `rpc_check_prompts_exist` PostgREST RPC
- Adds `checkPromptsByBrand` controller handler with full validation (org access, brand resolution, 500-item cap per request)
- Wires up `POST /v2/orgs/:spaceCatId/brands/:brandId/prompts/check` route with `organization:read` capability

## Related issue
https://jira.corp.adobe.com/browse/LLMO-5156

## Context

Part of a fix for a brandalf UI bug where prompt suggestions weren't being filtered for prompts on page 2+ of the configured-prompts table. The UI sends all suggestion `{text, region}` pairs to this endpoint and receives back only the ones that already exist (status `active`/`deleted`/`ignored`), so it can accurately hide them without paginating everything client-side.

**Depends on:** `mysticat-data-service` PR with the `rpc_check_prompts_exist` migration.

## Test plan

- [ ] Happy path: returns matching `{text, region}` pairs
- [ ] Empty/missing `prompts` body → 400
- [ ] >500 pairs → 400
- [ ] Brand not found → 404
- [ ] Org access check enforced

### DIFF
--- a/docs/openapi/prompts-v2-api.yaml
+++ b/docs/openapi/prompts-v2-api.yaml
@@ -373,6 +373,7 @@ v2-prompts-by-brand-check:
                   properties:
                     text:
                       type: string
+                      maxLength: 2000
                       description: Prompt text to look up
                     region:
                       type: string
--- a/docs/openapi/prompts-v2-api.yaml
+++ b/docs/openapi/prompts-v2-api.yaml
@@ -322,3 +322,84 @@ v2-prompts-by-brand-bulk-delete:
         description: PostgREST unavailable
       '500':
         $ref: './responses.yaml#/500'
+
+v2-prompts-by-brand-check:
+  parameters:
+    - name: spaceCatId
+      in: path
+      required: true
+      description: SpaceCat Organization ID (UUID)
+      schema:
+        type: string
+        format: uuid
+    - name: brandId
+      in: path
+      required: true
+      description: Brand ID (UUID, config id, or brand name)
+      schema:
+        type: string
+  post:
+    tags:
+      - customer-config
+    summary: Batch prompt existence check
+    description: |
+      Given a list of {text, region} pairs, returns the subset that already
+      exist in the prompts table for the given brand with status active,
+      deleted, or ignored. Used by the UI to filter suggestion panels without
+      fetching all paginated prompt pages client-side. Max 500 pairs per request.
+      Region values are normalized to lowercase 2-char codes before comparison.
+    operationId: checkPromptsByBrand
+    security:
+      - ims_key: [ ]
+      - api_key: [ ]
+    requestBody:
+      required: true
+      content:
+        application/json:
+          schema:
+            type: object
+            required:
+              - prompts
+            properties:
+              prompts:
+                type: array
+                minItems: 1
+                maxItems: 500
+                items:
+                  type: object
+                  required:
+                    - text
+                    - region
+                  properties:
+                    text:
+                      type: string
+                      description: Prompt text to look up
+                    region:
+                      type: string
+                      description: Region code (e.g. "US", "gb"). Normalized to lowercase 2-char before comparison.
+    responses:
+      '200':
+        description: Existence check completed
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                results:
+                  type: array
+                  description: Subset of the input pairs that already exist in the DB
+                  items:
+                    type: object
+                    properties:
+                      text:
+                        type: string
+                      region:
+                        type: string
+      '400':
+        $ref: './responses.yaml#/400'
+      '404':
+        description: Organization or brand not found
+      '503':
+        description: PostgREST unavailable
+      '500':
+        $ref: './responses.yaml#/500'
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -629,8 +629,8 @@ function BrandsController(ctx, log, env) {
       if (prompts.length > 500) {
         return badRequest('Maximum 500 prompt pairs per request');
       }
-      if (prompts.some((p) => !p.text || !p.region)) {
-        return badRequest('Each prompt must have "text" and "region"');
+      if (prompts.some((p) => !p || typeof p !== 'object' || !p.text || !p.region || p.text.length > 2000)) {
+        return badRequest('Each prompt must have "text" (max 2000 chars) and "region"');
       }
 
       const organization = await getOrganizationOrNotFound(spaceCatId);
@@ -656,7 +656,7 @@ function BrandsController(ctx, log, env) {
       const results = await checkPromptsExist({ brandUuid, prompts, postgrestClient });
       return ok({ results });
     } catch (error) {
-      log.error(`Error checking prompts existence for brand ${brandId}:`, error);
+      log.error('Error checking prompts existence', { brandId, error });
       return createErrorResponse(error);
     }
   };
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -40,6 +40,7 @@ import {
   updatePromptById,
   deletePromptById,
   bulkDeletePrompts,
+  checkPromptsExist,
   resolveBrandUuid,
 } from '../support/prompts-storage.js';
 import {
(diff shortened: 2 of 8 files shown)


MysticatBot-Dev (changes requested):
Hey @jjenscodee,

## Strengths

- Clean, well-scoped change that does exactly one thing. The controller follows the same validation-then-IO structure as its siblings, making it immediately recognizable to anyone working in this file.
- Correct architectural decision to push the set-intersection into a database RPC (`src/support/prompts-storage.js:800-817`) rather than fetching all prompts into memory. The DB uses indexes and the app stays stateless.
- Thorough input validation with the 500-item cap, UUID enforcement on `spaceCatId`, org access check, and PostgREST availability guard - all executed before any expensive I/O (`src/controllers/brands.js:615-647`).
- OpenAPI spec (`docs/openapi/prompts-v2-api.yaml`) ships alongside the implementation with accurate constraints (minItems, maxItems, required fields).
- Capability correctly assigned to `organization:read` (`src/routes/required-capabilities.js:265`) - this is a pure query with no side effects.
- Good test coverage across all documented validation paths, happy path, and error cases at both controller and storage layers (19 test cases total).

## Issues

### Important (Should Fix)

1. **`src/controllers/brands.js:632` - Null/undefined array items crash with 500 instead of 400**

   The validation `prompts.some((p) => !p.text || !p.region)` throws `TypeError: Cannot read properties of null` if any item in the array is `null` or `undefined`. This is caught by the outer try/catch and surfaces as a generic 500 error, polluting error logs with a stack trace for what is fundamentally a validation failure.

   Fix: Add a null/object check before property access:
   ```js
   if (prompts.some((p) => !p || typeof p !== 'object' || !p.text || !p.region)) {
   ```

2. **`src/controllers/brands.js:632` - No `maxLength` constraint on individual `text` values**

   The 500-item cap bounds cardinality but not total payload size. An authenticated user could send 500 prompts each with multi-MB `text` values, which are passed directly to the PostgREST RPC. This creates a resource exhaustion vector at the database layer.

   Fix: Add a length check (e.g., 2000 chars to match typical prompt lengths):
   ```js
   if (prompts.some((p) => !p || typeof p !== 'object' || !p.text || !p.region || p.text.length > 2000)) {
     return badRequest('Each prompt must have "text" (max 2000 chars) and "region"');
   }
   ```
   Also add `maxLength: 2000` to the `text` property in the OpenAPI spec.

### Minor (Nice to Have)

3. **`src/controllers/brands.js:659` - User-controlled `brandId` interpolated into log template**

   `brandId` accepts arbitrary strings (UUID, config id, or brand name) and is interpolated directly into the log template literal. Passing it as a structured field is more defensive against log injection:
   ```js
   log.error('Error checking prompts existence', { brandId, error });
   ```

## Recommendations

- The OpenAPI spec mentions "Region values are normalized to lowercase 2-char codes before comparison" but the controller passes values through unchanged. If the RPC handles normalization server-side, a brief comment noting this delegation would prevent future contributors from adding a redundant normalization loop.

## Assessment

Ready to merge? With fixes

The architecture is clean and follows established patterns. The two Important items are both straightforward fixes in the same validation block. The CI failure (bundle-build) should be investigated separately - it may be unrelated to this PR's changes.

## Next Steps

1. Fix the null-item validation crash and add a `maxLength` constraint on `text`. 2. Consider the structured log message improvement as optional polish.

MysticatBot-Dev (approved):
Hey @jjenscodee,

## Strengths

- All three prior findings addressed cleanly in one focused commit (`e63dc0c5`). The validation at `src/controllers/brands.js:631` now handles null items, non-objects, missing fields, and text length in a single pass - concise and correct.
- Correct computation boundary: pushing the set-intersection into a database RPC (`src/support/prompts-storage.js:800-817`) keeps the app layer stateless and avoids fetching all prompts into memory. The storage function is a thin RPC wrapper with a well-defined contract.
- Solid auth layering: route-level capability (`organization:read`), org-level access control (`accessControlUtil.hasAccess`), and brand scoped to org via `resolveBrandUuid` - three distinct authorization gates before any data access.
- Parameterized database access: user-supplied prompts array passed as a structured argument to `postgrestClient.rpc()`, not interpolated into a query string. No injection surface.
- OpenAPI spec and controller validation in lockstep: `minItems: 1`, `maxItems: 500`, `maxLength: 2000` on text, required fields. No spec/implementation drift.
- Controller follows the same structure as its siblings: validate params, validate body, check org access, check postgrest availability, resolve brand, call storage, return result.
- Thorough test coverage: 16 controller tests covering every validation branch, auth, not-found paths, happy path, and error propagation. 4 storage tests covering null client, null data, success path with argument verification, and RPC error.

## Issues

### Minor (Nice to Have)

1. **`src/controllers/brands.js:631` - `region` field accepts unbounded strings despite spec declaring 2-char codes**

   The validation checks that `region` is truthy but does not enforce length. The OpenAPI spec documents "Normalized to lowercase 2-char codes before comparison," but a caller can send arbitrarily long region strings in the 500-item array. Real-world risk is low given HTTP body parser limits and PostgREST payload handling, but for defense-in-depth consistency with the `text` validation, consider a generous cap:
   ```js
   if (prompts.some((p) => !p || typeof p !== 'object' || !p.text || !p.region || p.text.length > 2000 || p.region.length > 10)) {
   ```

## Recommendations

- The `POST .../prompts/check` pattern (batch existence check with array body, subset response, cap enforcement) may be useful to other domains in the platform. If it proves successful, consider documenting it as a reusable pattern so future endpoints follow the same shape.

## Assessment

Ready to merge? Yes

This is a well-scoped, architecturally sound addition that follows established patterns. The hardening commit resolved all previously flagged issues. Authentication, authorization, input validation, and parameterized database access are all in place. The single Minor item is cosmetic polish, not a blocker.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 33s | <b>Cost:</b> $2.47 | <b>Commit:</b> <code>e63dc0c572071d350a486e5741904c689082f9d8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

JayKid:
## Multi-persona PR review (Architect · DBA · Security · Tester + Junior Dev)

Both PRs were re-reviewed end-to-end alongside each other. Prior bot-flagged issues are resolved (cardinality guard, `CONCURRENTLY` index, null-item guard, text `maxLength`, structured logging, integration tests). Below are findings not previously raised.

### Major

1. **Btree index on `prompts.text` may exceed Postgres' ~2704-byte page entry limit at the new 2000-char OpenAPI cap** — `db/migrations/20260530000000_idx_prompts_brand_text.sql:15` (mysticat). UTF-8 multibyte content (CJK, emoji, accented Latin) at 2000 chars can produce index entries >2704 bytes, causing runtime INSERT failures (`index row size NNN exceeds maximum 2704 for index "idx_prompts_brand_text"`). Consider indexing `(brand_id, md5(text))` and rewriting the RPC join, or adding `CHECK (octet_length(text) <= 2700)`.

2. **`resolveBrandUuid` tenant scoping not verifiable from this diff** — `src/controllers/brands.js:150` (spacecat). `accessControlUtil.hasAccess(organization)` confirms the caller's access to *their* org; the SQL `rpc_check_prompts_exist` takes a bare `p_brand_id` with no org-scoped predicate. If `resolveBrandUuid` resolves brands by global lookup, an authenticated user in org A could submit org B's brand UUID and enumerate its prompts (text-oracle, including soft-deleted entries). Please confirm `resolveBrandUuid` filters by `organization_id = spaceCatId`, or add an explicit ownership assertion after resolution.

### Minor

- **Whitespace-only `text`/`region` bypass validation** (`brands.js:131`): `!p.text` doesn't catch `"   "`. Use `!hasText(p.text)` / `!hasText(p.region)` (already imported); add corresponding tests.
- **Missing controller-layer boundary tests**: no tests for `prompts.length === 500` (only 501 → 400) and `text.length === 2000` (only 2001 → 400). Pins the `> N` vs `>= N` invariant.
- **"Storage throws" test asserts only status code** (`brands.test.js:475-485`) — should also verify `loggerStub.error` was called with `{ brandId, error }`.
- **Magic numbers 500/2000 duplicated across both repos** with no named constants. Drift between repos = silent contract divergence.
- **Compound validation predicate** at `brands.js:131` is hard to scan (5 conditions, mixed negations, one error message). Extract `isValidPromptPair(p)`.
- **Response envelope `{ results: [...] }`** diverges from sibling v2 prompt endpoints which return flat arrays. SDK consumers will see inconsistent types.
- **`p_prompts = NULL` silently returns empty** instead of erroring (`jsonb_array_length(NULL) <= 500` is `NULL` → falls to `'[]'::jsonb`). Masks caller bugs in non-API callers of the RPC.
- **PostgREST `error.message` is interpolated into the thrown Error** (`prompts-storage.js:218`) → flows into `createErrorResponse`. Verify the shared error util strips messages from 5xx bodies, else Postgres schema/constraint details can leak.
- **Deleted/ignored prompts confirmable** to any `organization:read` caller. Documented as intentional for dedup, but worth an explicit OpenAPI security note or restricting that visibility to `organization:write`.
- **No GIN index on `prompts.regions`** — `= ANY(p.regions)` is a per-row array scan after the btree narrow. Fine today; worth a follow-up if this becomes a hot read path.
- **Silent region normalization** (`USA` → `us`) — documented in OpenAPI, invisible in JS code. Add a one-line JSDoc on `checkPromptsExist`, or normalize at the edge so request/response shapes match.
- **No `test/it/postgres/` integration tests** added on the spacecat side. PR 629 has thorough DB-level integration tests; spacecat covers unit only.
- **Non-array `prompts` and partial `postgrestClient`** (`.from` without `.rpc`) not exercised by tests.

### Questions

- Is `postgrest_anon` the literal unauthenticated PostgREST role for this deployment, or the post-JWT authenticated role? Other RPCs in the repo grant to the same role, suggesting convention, but if it's truly anonymous, anyone reaching PostgREST bypasses the spacecat controller's org check.
- Is there route-level rate limiting on `POST /v2/orgs/:spaceCatId/brands/:brandId/prompts/check`? Combined with the IDOR question, an unrate-limited 500-text-per-call endpoint enables enumeration.
- Should `{ results: [...] }` be the v2 query-endpoint convention, or align with sibling flat arrays?

### What's working well

- Layered test strategy (DB-level integration in mysticat + unit branches in spacecat) forms a clean pyramid.
- `LANGUAGE sql STABLE PARALLEL SAFE` + explicit `search_path` + typed JSONB params — correct pattern for a read-only PostgREST RPC. No injection surface.
- Prior bot findings cleanly resolved in one follow-up commit each, with surprising behaviors (silent cardinality cap, region normalization) documented in the SQL `COMMENT ON FUNCTION` block where future readers will find them.

### Recommendation

**Approve with reservations** — address the two Major items (or confirm the IDOR concern is already mitigated upstream) before merge. Minors can ship as follow-ups.

---

<sub>Cross-posted to companion PR. Reviewed via parallel persona agents (Architect · DBA · Security · Tester+JuniorDev) over the full diffs of both PRs together.</sub>

JayKid (approved):
Approving.

The `df313d88` commit resolves all six Minor items I flagged in the previous review (whitespace-only `text`/`region` rejection via `?.trim()`, boundary tests at exactly-500 prompts and exactly-2000-char text, and the structured `loggerStub.error` assertion).

I also verified the Major IDOR concern from my earlier review against `resolveBrandUuid` on `main` (`src/support/prompts-storage.js:26`): both the UUID branch and the name-lookup branch filter `.eq('organization_id', organizationId)`, so brand resolution is tenant-scoped. No IDOR.

Remaining open items are all Minor and can be deferred or addressed in follow-ups:
- Magic numbers 500/2000 hardcoded on both sides of the wire
- `{ results: [...] }` envelope diverges from sibling v2 prompt endpoints (decide convention)
- PostgREST `error.message` flows into `createErrorResponse` — verify the shared util strips message bodies on 5xx
- `deleted`/`ignored` status visibility is intentional; worth an OpenAPI security note
- Silent region normalization not surfaced in JS code (one-line JSDoc on `checkPromptsExist` would help)
- No `test/it/postgres/` integration tests
- Compound validation predicate is dense (now denser with `?.trim()` chains)

None of these block merge.

PR #2491, opened 2026-05-26
fix(llmo): guard createOrFindSite from moving site with active enroll…

Kanishkavijay39:
## Summary:
  - createOrFindSite was silently reassigning a site to a new org whenever the org IDs differed, regardless of whether the site had active products in its current org
  - This matches the same guard already present in PLG onboarding (plg-onboarding.js:1050-1058)
  - Now: if getSiteEnrollments() returns any enrollments, we throw an error instead of moving the site
  - Sites with no active enrollments continue to be re-parented as before (LLMO-4176 behaviour preserved)

## Local testing
image

## Related Issues

Thanks for contributing!


MysticatBot-Dev (changes requested):
Hey @Kanishkavijay39,

## Strengths

- **Guard placement is correct** (`src/controllers/llmo/llmo-onboarding.js:999-1002`): The enrollment check is inserted before `setOrganizationId` and `save`, so the guard prevents any partial mutation. A failed check leaves the site object completely unmodified.
- **Well-shaped test coverage** (`test/controllers/llmo/llmo-onboarding.test.js:1260-1291`): The new test asserts both the positive (exception thrown with meaningful message) and the negative (`setOrganizationId` and `save` were NOT called). The existing re-parent test is updated to stub `getSiteEnrollments` returning empty, confirming the happy path still works.
- **Consistency with existing platform behavior**: This mirrors the guard already present in PLG onboarding (`plg-onboarding.js:1050-1058`), aligning both onboarding paths to the same invariant.
- **Well-scoped change**: One behavioral fix, one new test, one existing-test fixup. No unrelated refactoring mixed in.

## Issues

### Important (Should Fix)

**Optional chaining on `enrollments` treats unexpected undefined/null as "no enrollments"** - `src/controllers/llmo/llmo-onboarding.js:1000`

`enrollments?.length > 0` - if `getSiteEnrollments()` resolves to `undefined` or `null` (e.g., a bug in the data-access layer where a model method returns nothing instead of an empty array), the optional chaining silently evaluates to false and the re-parent proceeds. This is the exact scenario the guard is meant to prevent: a site with active products gets moved because the enrollment check was inconclusive.

For a guard whose explicit purpose is to prevent data corruption, the failure mode should be loud (fail closed), not permissive (fail open). Suggested fix:

```js
const enrollments = await site.getSiteEnrollments();
if (!Array.isArray(enrollments)) {
  throw new Error(`Unable to verify enrollments for site ${baseURL}; aborting org move.`);
}
if (enrollments.length > 0) {
  throw new Error(`Site ${baseURL} is already associated with another org that has active products and cannot be moved.`);
}
```

If the team has established that `getSiteEnrollments` is contractually guaranteed to always return an array, this becomes less pressing - but given that this guard exists specifically to prevent data corruption, failing closed is the safer posture.

## Recommendations

- **Include org IDs in the error message for operational debuggability**: The error currently includes `baseURL` but not the current or requested org IDs. When this surfaces in logs or Slack notifications, operators need to look up both orgs separately. Consider: `Site ${baseURL} belongs to org ${site.getOrganizationId()} with active enrollments and cannot be moved to org ${organizationId}.`
- **Consider a typed/coded error**: Callers further up the stack may want to distinguish "site has enrollments" from other errors programmatically. A custom error code (e.g., `err.code = 'SITE_HAS_ACTIVE_ENROLLMENTS'`) allows upstream handlers to present appropriate user-facing feedback without string-matching.

## Assessment

Ready to merge? With fixes
Reasoning: The core logic is correct, the change is minimal and focused, and the test coverage is solid. The optional-chaining concern is worth addressing because a safety guard that fails open under unexpected data-layer behavior contradicts the PR's stated intent. The fix is a one-line addition.

## Next Steps

1. Address the optional chaining concern to ensure the guard fails closed on unexpected return values.
2. The recommendations (richer error message, error code) are optional improvements.

src/controllers/llmo/llmo-onboarding.js
```diff
@@ -996,6 +996,10 @@ export async function createOrFindSite(baseURL, organizationId, context, deliver
   const site = await Site.findByBaseURL(baseURL);
   if (site) {
     if (site.getOrganizationId() !== organizationId) {
+      const enrollments = await site.getSiteEnrollments();
+      if (enrollments?.length > 0) {
```
MysticatBot-Dev: **Important**: `enrollments?.length > 0` - if `getSiteEnrollments()` resolves to `undefined` or `null` (data-access bug, model method returning nothing), the optional chaining silently evaluates to false and the re-parent proceeds. For a guard whose purpose is preventing data corruption, the failure mode should be loud:

```js
const enrollments = await site.getSiteEnrollments();
if (!Array.isArray(enrollments)) {
  throw new Error(`Unable to verify enrollments for site ${baseURL}; aborting org move.`);
}
if (enrollments.length > 0) {
  throw new Error(`Site ${baseURL} is already associated with another org that has active products and cannot be moved.`);
}
```

If the contract guarantees an array, the optional chaining can be dropped in favor of plain `.length > 0` so a contract violation surfaces loudly.

Kanishkavijay39: fixed

MysticatBot-Dev (approved):
Hey @Kanishkavijay39,

## Strengths

- **Prior finding fully addressed** (`src/controllers/llmo/llmo-onboarding.js:999-1001`): The optional chaining fail-open concern from the prior review is resolved with an explicit `!Array.isArray(enrollments)` guard that throws with a clear, actionable error message. The guard fails closed on unexpected return types - the correct defensive pattern for a data integrity check.
- **Error messages are operationally useful** (`src/controllers/llmo/llmo-onboarding.js:1000, 1003`): Both error paths now include base URL, current org, and requested org. When this fires in production, the on-call engineer can immediately understand what happened.
- **Thorough test coverage** (`test/controllers/llmo/llmo-onboarding.test.js:1260-1323`): Three distinct scenarios tested - empty enrollments (re-parent proceeds), active enrollments (blocks move), and non-array response (blocks move). The negative assertions (`setOrganizationId` and `save` not called) verify side-effect prevention, not just exception propagation.
- **Minimal blast radius**: The guard is surgically placed before the existing `setOrganizationId` call, preserving all existing behavior for the happy path. Consistent with the pattern already in PLG onboarding.
- **Integration test updated** (`test/support/slack/actions/onboard-llmo-modal.test.js:58`): The Slack modal mock was updated to include `getSiteEnrollments`, preventing regressions in the broader onboarding flow.

## Recommendations

- **Consider wrapping `getSiteEnrollments()` rejection for consistent error context** (`src/controllers/llmo/llmo-onboarding.js:999`): If `getSiteEnrollments()` rejects (network error, DB timeout), the exception propagates with a raw stack trace rather than the structured "Unable to verify enrollments" message. The behavior is already fail-closed (the move is still prevented), but wrapping in try/catch would produce a consistent, debuggable error for all "can't verify" scenarios. A matching test case (`getSiteEnrollments` rejects with an error) would document the behavior either way.

## Assessment

Ready to merge? Yes
Reasoning: The fix is minimal, correct, and well-tested. It closes a real data integrity gap with a fail-closed approach, addresses all prior review findings, and aligns with the existing PLG onboarding pattern.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 48s | <b>Cost:</b> $2.44 | <b>Commit:</b> <code>2da80ac7b3a78758d7362967084e9533d1fc63a4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

vivesing (approved):

PR #2493, opened 2026-05-27
fix(auth): use method+suffix for route matching in RouteScopedLegacyApiKeyHandler

ravverma:
## Summary

- `enrichPathInfo` sets `context.pathInfo.route` to the **first path segment only** (e.g. `'event'` for `/event/fulfillment`), not the full `'METHOD /path'` string
- `RouteScopedLegacyApiKeyHandler` was checking `SCOPED_ROUTES.has(context.pathInfo.route)`, so `SCOPED_ROUTES.has('event')` always returned `false`
- The two-route scope guard was completely bypassed: the handler returned `null` for every request, and `LegacyApiKeyHandler` (the catch-all) authenticated legacy API key requests on **all routes** not just the two S2S-exempt ones
- The migration-signal log never fired (which is what surfaced this bug)

Fix: derive the route key from `method + suffix` (the actual shape `enrichPathInfo` produces) and update the tests to reflect reality. The tests were also using the wrong `pathInfo` shape (`route: 'POST /event/fulfillment'` instead of `method: 'POST', suffix: '/event/fulfillment'`), which is why the bug was not caught earlier.

## Test plan
- [ ] `npx mocha test/support/route-scoped-legacy-api-key-handler.test.js` — all 15 tests pass
- [ ] Verify `POST /event/fulfillment` and `POST /slack/channels/invite-by-user-id` now emit the `[legacyApiKey] request authenticated via route-scoped legacy API key handler` log
- [ ] Verify other routes authenticated with a legacy API key no longer pass through this handler (they should 401 once `LegacyApiKeyHandler` is removed in the planned follow-up)

### DIFF
--- a/src/support/route-scoped-legacy-api-key-handler.js
+++ b/src/support/route-scoped-legacy-api-key-handler.js
@@ -40,12 +40,14 @@ const SCOPED_ROUTES = new Set([
  */
 export default class RouteScopedLegacyApiKeyHandler extends LegacyApiKeyHandler {
   async checkAuth(request, context) {
-    if (!SCOPED_ROUTES.has(context?.pathInfo?.route)) {
+    const { method, suffix } = context?.pathInfo || {};
+    const routeKey = method && suffix ? `${method} ${suffix}` : null;
+    if (!SCOPED_ROUTES.has(routeKey)) {
       return null;
     }
     const result = await super.checkAuth(request, context);
     if (result) {
-      context.log.info(`[legacyApiKey] request authenticated via route-scoped legacy API key handler [${context.pathInfo.route}]`);
+      context.log.info(`[legacyApiKey] request authenticated via route-scoped legacy API key handler [${routeKey}]`);
     }
     return result;
   }
--- a/test/support/route-scoped-legacy-api-key-handler.test.js
+++ b/test/support/route-scoped-legacy-api-key-handler.test.js
@@ -66,7 +66,7 @@ describe('RouteScopedLegacyApiKeyHandler', () => {
 
   it('delegates to LegacyApiKeyHandler.checkAuth for POST /event/fulfillment', async () => {
     const req = {};
-    const ctx = { pathInfo: { route: 'POST /event/fulfillment' }, log: logStubs };
+    const ctx = { pathInfo: { method: 'POST', suffix: '/event/fulfillment' }, log: logStubs };
     const result = await handler.checkAuth(req, ctx);
     expect(result).to.equal('SUPER_RESULT');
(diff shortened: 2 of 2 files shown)


MysticatBot-Dev (approved):
Hey @ravverma,

## Strengths

- **Correct root-cause identification and targeted fix** (`src/support/route-scoped-legacy-api-key-handler.js:42-44`): The fix correctly derives `routeKey` from `method` and `suffix` - the actual fields `enrichPathInfo` populates on `pathInfo`. The constructed key (`'POST /event/fulfillment'`) matches the entries in `SCOPED_ROUTES`. Clean one-line-of-logic fix for a bypass that was completely disabling the route-scoping guard.
- **Fail-closed design** (`src/support/route-scoped-legacy-api-key-handler.js:42-43`): The `context?.pathInfo || {}` fallback and the `method && suffix` guard ensure that missing or malformed `pathInfo` produces `null` (not in `SCOPED_ROUTES`) rather than throwing or accidentally matching. Ambiguous state results in denial, not accidental grant.
- **Tests now reflect reality** (`test/support/route-scoped-legacy-api-key-handler.test.js`): All fixtures updated to use the actual `{ method, suffix }` shape instead of the imaginary `{ route: 'METHOD /path' }` format. This is how the bug went undetected - the tests encoded a contract that never existed.
- **Well-scoped change discipline**: Only the broken handler and its tests are touched. No unrelated cleanup, no scope creep.

## Issues

### Minor (Nice to Have)

**Missing tests for partial pathInfo** - `test/support/route-scoped-legacy-api-key-handler.test.js`: The test for `'returns null when method and suffix are missing'` covers `pathInfo: {}`, but there are no tests for `{ method: 'POST' }` (suffix missing) or `{ suffix: '/event/fulfillment' }` (method missing). The `method && suffix` guard is load-bearing for safety, and a future refactor could inadvertently remove it. Two small test cases would lock this behavior in.

## Assessment

Ready to merge? Yes
Reasoning: This is a targeted, correct fix for a security-relevant bypass where the route-scoped guard was entirely non-functional. The fix is fail-closed, minimal, and correctly aligns with the actual data shape produced by the upstream `enrichPathInfo` middleware.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 39s | <b>Cost:</b> $3.20 | <b>Commit:</b> <code>1a1b3e2785b19e8ff77e9adf8e34dc50ffd4f945</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2494, opened 2026-05-27
feat: classify gapKind per topic/prompt | LLMO-4959

calvarezg:
## Description

The proto response from the gRPC gap-topics / gap-prompts services does not expose the per-row GAP_KIND, so every row arrived at the UI without a kind field and badges defaulted to WEAK regardless of the active filter. Compute the kind here from `mentions` and `gapMentions` using the proto enum semantics (MISSING / WEAK / SHARED / STRONG / UNIQUE) so the server-side filter and the UI badge column stay in sync.

## Related Issues
https://jira.corp.adobe.com/browse/LLMO-4959

### DIFF
--- a/src/support/ai-visibility/grpc-utils.js
+++ b/src/support/ai-visibility/grpc-utils.js
@@ -598,6 +598,69 @@ export function parseGapKindEnumList(sp) {
   return result.length ? result : [1];
 }
 
+/**
+ * Classify a topic or prompt entry by its mention gap between the target brand and
+ * its competitors. Returns one of GAP_KIND enum names (string): 'MISSING' | 'WEAK'
+ * | 'SHARED' | 'STRONG' | 'UNIQUE'. Mirrors the proto enum semantics defined in
+ * `semrush.services.ai_seo.common.v1.GAP_KIND.ENUM`:
+ *   - MISSING : target has 0 mentions AND every competitor has > 0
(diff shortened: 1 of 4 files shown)


src/support/ai-visibility/grpc-utils.js
```diff
@@ -598,6 +598,69 @@ export function parseGapKindEnumList(sp) {
...
+  const competitorMentions = gapMentions
+    .filter((m) => {
+      const d = m?.brand?.domain;
+      return typeof d === 'string' && d.toLowerCase() !== target;
+    })
+    .map((m) => num(m?.mentions));
+
+  if (competitorMentions.length === 0) {
+    return 'SHARED';
+  }
```
dmitriigulkin-semrush, line 638: if there are no competitors mentioned, it should be UNIQUE

dmitriigulkin-semrush:
hey Char! I thought before I start texting where the mapping can go wrong, could you please advice why do we want to display kind type at all? 
it's fairly technical thing that we use to separate different relations of your brand vs your competitor' brand, we use it to display something like that (see screenshot)
Screenshot 2026-05-27 at 12 53 53

since we're looking for relevant opportunities here, we can use gapKind: [WEAK, SHARED, STRONG] and default sorting  by total amount of competitors mentioned

my point being, we might not need to display gap-kind at all(i myself don't see any value for that, but correct me if i'm wrong), that would also mean less code branches, less client specificity and smaller surface to support

PR #2495, opened 2026-05-27
fix(ci): add exit:true to mocha config to fix it-postgres timeout

ravverma:
## Summary

- The `ci / it-postgres` job was hitting the 15-minute wall even though all 499 tests pass in ~3 minutes
- After tests complete, mocha hangs on open handles: the `@adobe/fetch` HTTP agent used by the data-access layer for PostgREST requests holds sockets alive after the PostgREST container stops, keeping the Node.js event loop from emptying
- `test/it/postgres/.mocharc.postgres.yml` already had `exit: true`, but the CI command (hardcoded in the `mysticat-ci` reusable workflow as `npx mocha --require test/it/postgres/harness.js --timeout 30000 ...`) passes no `--config` flag — so that file is never loaded
- Mocha reads `package.json`'s `mocha` section instead; adding `exit: true` there ensures the CI invocation picks it up

## Evidence

From the timed-out run (CI job `78015006847`, branch `feat-LLMO-4959-topic-opps-gap`):
- Tests finish at `05:44:47Z` (`499 passing (3m)`)
- Process killed at `05:56:04Z` (`The operation was canceled`)
- `Terminate orphan process: pid (2578) (npm exec mocha)` confirms mocha is the hanging process

## Test plan
- [ ] Confirm `ci / it-postgres` completes in under 5 minutes on next run

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
   "mocha": {
     "require": "test/setup-env.js",
     "recursive": "true",
+    "exit": true,
     "reporter": "mocha-multi-reporters",
     "reporter-options": "configFile=.mocha-multi.json"
   },


MysticatBot-Dev (approved):
Hey @ravverma,

## Strengths

- **Correct root-cause analysis** (`package.json:71`): The PR description clearly identifies why the existing `.mocharc.postgres.yml` config was not being loaded (CI workflow passes no `--config` flag, so mocha falls back to the `package.json` `mocha` section). Adding `exit: true` here is the right fix for the immediate problem.
- **Minimal, focused change**: One line, one file, one purpose. Good change discipline with minimal blast radius.

## Issues

### Minor (Nice to Have)

1. **`package.json:71` - Boolean vs string inconsistency**: The existing `recursive` field uses `"true"` (a string), while the new `exit` field uses `true` (a boolean). Mocha handles both, but the inconsistency is a minor readability wart. Not worth blocking on since mocha normalizes both forms identically.

## Recommendations

- The underlying issue is that the `mysticat-ci` reusable workflow hardcodes the mocha invocation without `--config`, making `.mocharc.postgres.yml` dead configuration for CI runs. Consider filing a follow-up to have the workflow pass `--config` explicitly - that would let the postgres-specific config be the single source of truth and avoid duplicating intent across two files. Out of scope for this PR since the workflow lives in a different repo.

## Assessment

Ready to merge? Yes

The fix is correct, proportional, and solves the immediate CI timeout. The `exit` flag tells mocha to force-quit after tests complete rather than waiting for the event loop to drain - appropriate behavior for CI where lingering handles from stopped containers are expected and harmless. The global application of `exit: true` to all mocha runs (not just postgres integration tests) is harmless since test results are fully reported before the flag fires.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 26s | <b>Cost:</b> $0.96 | <b>Commit:</b> <code>c762b616b75e44c0dbb7db9c3b7bb31c4f8b7ef8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2497, opened 2026-05-27
feat: cursor-based pagination for GET /sites | SITES-41367

martinst06:
## What

Adds cursor-based pagination to `GET /sites` so consumers can page through the full result set instead of relying on a single capped flat-array response.

The endpoint now supports two response shapes for backwards compatibility:

- **Legacy (no query params):** unchanged flat array of sites. Still uses the existing default + friends-and-family org exclusion to stay under the 6 MB Lambda response limit. Deprecated — see follow-ups.
- **Paginated (`limit` and/or `cursor` present):** returns the `{ sites, pagination }` envelope over the full unfiltered result set. Iterate using `pagination.cursor` until `pagination.hasMore` is `false`.

Validation on the paginated path: `limit` defaults to 100, is clamped to a max of 500, and non-integer/non-positive values (`abc`, `0`, `-1`) return 400; `cursor` is opaque, must be a string, and is rejected with 400 if longer than 256 characters.

## Behaviour changes to be aware of

- **Polymorphic response:** `GET /sites` now returns either a flat array (legacy) or the `{ sites, pagination }` envelope, depending on whether `limit`/`cursor` is supplied. Modelled as a `oneOf` in the OpenAPI spec.
- **Paginated path returns more orgs:** the legacy org exclusion is intentionally **not** applied on the paginated path, so paging returns sites from the two non-customer orgs that the legacy array hides. "All sites" therefore means different sets on the two paths.
- **`fetchConfig` trimmed in the list DTO:** `ConfigDto.toListJSON` now exposes only `fetchConfig.overrideBaseURL` (and omits `fetchConfig` entirely when there is no `overrideBaseURL`). Because the DTO is shared, this also applies to the legacy path, so `fetchConfig.headers` is no longer returned by `GET /sites`. This is intentional — `headers` can carry credentials (`Authorization`, `Cookie`, etc.) that should not be returned in bulk. The full `fetchConfig` remains available via `GET /sites/:siteId`.
- **Sunset instrumentation:** the legacy path emits a `[sites][legacy-shape]` log on every hit, which is the signal we will use to retire it.

## Backwards compatibility

Consumers that call `GET /sites` with no query params keep getting a flat array and the same set of sites. The only shape change on that path is the trimmed `fetchConfig` described above.

## Follow-ups (tracked separately, not blockers)

- **Caller inventory** — use the `[sites][legacy-shape]` Coralogix marker to enumerate every consumer still calling `GET /sites` without `limit`/`cursor`.
- **Migration plan** — give those consumers a concrete path to the paginated envelope, noting the unfiltered org behaviour and the trimmed `fetchConfig`.
- **Shutdown date** — agree and document a hard date for removing the legacy branch, gated on the marker showing zero hits for the agreed window (currently 30 consecutive days).
- **Product/entitlement filtering** (e.g. all ASO sites / all LLMO sites) — out of scope here; `productCode` lives two hops away (`Site` → `SiteEnrollment` → `Entitlement`), so it needs new data-access query support and a cursor design of its own.

## Testing

- Unit tests in `test/controllers/sites.test.js` cover the paginated envelope, legacy flat array, limit defaulting/clamping/validation, cursor validation and boundaries, slim DTO shape, S2S `site:readAll` on the paginated path, non-admin denial, and the unexpected-shape fallback. The `[sites][legacy-shape]` log format is pinned by a test.
- Integration tests in `test/it/shared/tests/sites.js` exercise the real DAL: paginated envelope shape and a two-page cursor round-trip with no overlap.
- OpenAPI specs (`schemas.yaml`, `sites-api.yaml`) and generated docs updated to reflect both response shapes.

## Related Issues

SITES-41367


MysticatBot-Dev (changes requested):
Hey @martinst06,

## Strengths

- `src/controllers/sites.js:411-413` - Clean backward-compatible branching. The `hasText(limitParam) || hasText(cursor)` gate gives a tidy opt-in to pagination while preserving the legacy contract for existing consumers with zero response shape change.
- `src/controllers/sites.js:419-422` - Solid limit validation. `parseInt` with radix, `Number.isInteger` positivity check, and clamping to `MAX_LIMIT` correctly handle NaN, floats, negatives, zero, and over-large values.
- `test/controllers/sites.test.js:623-744` - Thorough unit test suite covering the paginated envelope, legacy flat array, limit clamping, invalid input, cursor passthrough, DTO shape, and auth denial. Good boundary condition coverage.
- Removing the org-exclusion workaround in favor of proper pagination is the correct long-term solution. The old approach was data-dependent and would silently break as the dataset grew.
- Auth check is preserved and runs before any pagination logic on both paths.
- Integration tests updated to reflect actual behavior (6 sites returned vs the old filtered 4).

## Issues

### Important (Should Fix)

**1. Legacy path lost its protection against the 6MB Lambda response size limit**
`src/controllers/sites.js:436` - The old code excluded `DEFAULT_ORGANIZATION_ID` and `ORGANIZATION_ID_FRIENDS_FAMILY` specifically to avoid 413 responses from exceeding the 6MB Lambda response limit. The legacy path (no `limit`/`cursor` params) now calls `Site.all({}, { fetchAllPages: true })` with no filtering, returning strictly more data than before. Existing consumers hitting `GET /sites` without pagination params will receive a larger payload than yesterday.

Why it matters: The org exclusion existed because this endpoint already hit the 6MB limit in production. Removing it without protecting the legacy path re-exposes the original problem before consumers have had time to adopt pagination.

How to fix: Either (a) cap the legacy path with a warning log when the result set is large, (b) restore the org exclusion on the legacy path only, or (c) at minimum, add a deprecation log on the legacy path and document a sunset timeline so the risk is explicitly acknowledged. The comment on `MAX_LIMIT` ("lower to 500 if 6mb lambda limit is hit again") should note that the legacy path is the unprotected surface.

**2. Cursor passed to data access layer without validation**
`src/controllers/sites.js:412` - The `cursor` parameter is taken directly from user input and forwarded to `Site.all()` without type or format validation. While this endpoint requires admin or S2S readAll capability (limiting the attacker pool), the cursor is still untrusted input passed to the data layer. A malformed cursor could trigger unexpected errors that leak internal state in error responses.

Why it matters: Defense in depth. Even behind auth, validating input shape at the controller boundary prevents classes of issues in downstream layers.

How to fix: Add basic validation before passing to the DAL - validate type is string and apply a reasonable length limit:
```js
if (cursor !== null && (typeof cursor !== 'string' || cursor.length > 1024)) {
  return badRequest('Invalid cursor parameter');
}
```

**3. No integration test for the paginated path**
`test/it/shared/tests/sites.js` - The integration tests verify only the legacy flat-array response. There is no integration test calling `GET /sites?limit=2` to verify the paginated envelope shape, cursor behavior, and `hasMore` field work end-to-end against real data access.

Why it matters: Unit tests mock `Site.all` to return `{ data, cursor }`, but the integration between the controller, the actual data access layer's `limit`/`cursor`/`returnCursor` options, and the response serialization is only proven via integration tests. If `Site.all` returns a different shape under real conditions, the unit tests would still pass.

How to fix: Add at least one integration test that calls `GET /sites?limit=2` as admin, asserts the response has `sites` and `pagination` keys, and verifies `hasMore` behaves correctly when results exceed the requested limit.

### Minor (Nice to Have)

**4. Silent empty-array fallback on contract violation**
`src/controllers/sites.js:425` - `(results.data || [])` will silently return an empty `sites` array if `Site.all` with `returnCursor: true` returns an unexpected shape (missing `data` property). This masks a data-access contract violation from the caller, who gets `{ sites: [], pagination: { hasMore: false } }` with no signal that something went wrong.

How to fix: Consider logging a warning or failing explicitly if `results` lacks a `data` property.

**5. Test claims negative-limit coverage but only tests zero**
`test/controllers/sites.test.js:695` - The test is named "falls back to DEFAULT_LIMIT when limit is zero or negative" but only sends `limit: '0'`. A negative value (`'-5'`) exercises a subtly different code path. The logic is correct, but the test does not prove it for negatives as claimed.

How to fix: Add a second assertion or separate test case with `limit: '-5'`.

## Recommendations

- Log a deprecation notice on the legacy path so you can track which consumers haven't migrated. The existing S2S log covers S2S consumers, but admin callers on the legacy path are invisible. This gives the data needed to set a removal timeline.
- Consider adding a `total` or `totalCount` field to the pagination envelope if the data access layer can provide it cheaply. If it is expensive, document explicitly why it is absent - that decision will be re-asked.

## Assessment

Ready to merge? With fixes

The pagination implementation is clean, well-tested at the unit level, and architecturally sound - cursor-based pagination is the right pattern. The backward-compatible branching is well-structured. The main concern is that the legacy path lost its 6MB protection (the org exclusion) without gaining an alternative. If you are confident the current production dataset fits in 6MB with the previously-excluded orgs included, this could merge as-is with a documented follow-up to deprecate the legacy path - but the risk should be explicitly acknowledged.

## Next Steps

1. Address the legacy path 6MB protection gap - either restore a safeguard or explicitly document the risk and deprecation plan.
2. Add cursor validation at the controller boundary.
3. Add an integration test for the paginated path.
4. Minor items are optional.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 28s | <b>Cost:</b> $4.65 | <b>Commit:</b> <code>4f4e8785f43f6af024e50ad8a1fce82f945154c5</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot-Dev (changes requested):
Hey @martinst06,

## Strengths

- `src/controllers/sites.js:430-443` - Previously flagged issue now addressed: the legacy path correctly retains the `DEFAULT_ORGANIZATION_ID` and `ORGANIZATION_ID_FRIENDS_FAMILY` exclusion to stay under the 6MB Lambda response limit. The comment explaining the rationale is a nice touch for future maintainers.
- `src/controllers/sites.js:411-413` - Clean backward-compatible branching. The `hasText(limitParam) || hasText(cursor)` gate provides a tidy opt-in to pagination while preserving the legacy contract for existing consumers with zero response shape change.
- `src/controllers/sites.js:419-422` - Solid limit validation. `parseInt` with radix, `Number.isInteger` positivity check, and clamping to `MAX_LIMIT` correctly handle NaN, floats, negatives, zero, and over-large values.
- `test/controllers/sites.test.js:625-744` - Thorough unit test suite covering the paginated envelope, legacy flat array, limit clamping, invalid input, cursor passthrough, DTO shape, and auth denial. Good boundary condition coverage.
- Auth check is preserved and runs before any pagination logic on both paths.
- `test/it/shared/tests/sites.js:90-96` - Integration test updated to accurately reflect the restored exclusion behavior.

## Issues

### Important (Should Fix)

**1. Cursor passed to data-access layer without validation**
`src/controllers/sites.js:412` - The `cursor` parameter is taken directly from user input and forwarded to `Site.all()` without type, format, or length validation. While this endpoint requires admin or S2S readAll capability (limiting the attacker pool), if the underlying data layer does not defensively validate cursor format, a malformed cursor could trigger unexpected query behavior, error-based information disclosure, or resource exhaustion. The `limit` parameter gets thorough validation; the cursor gets none.

How to fix - add a basic guard before passing to the data layer:
```js
if (cursor !== null && (typeof cursor !== 'string' || cursor.length > 2048)) {
  return badRequest('Invalid cursor parameter');
}
```

If the data access layer already validates cursors and returns safe errors, a brief comment documenting that assumption would suffice as an alternative.

**2. No integration test for the paginated path**
`test/it/shared/tests/sites.js` - The integration tests only verify the legacy flat-array response. There is no integration test calling `GET /sites?limit=2` to verify the paginated envelope shape, cursor behavior, and `hasMore` field work end-to-end against the real data access layer.

Why it matters: unit tests mock `Site.all` to return `{ data, cursor }`, but the integration between the controller, the actual data access layer's `limit`/`cursor`/`returnCursor` options, and the response serialization is only proven via integration tests. The `ci / it-postgres` check is currently failing, which makes this gap more concerning.

How to fix: add at least one integration test that calls `GET /sites?limit=2` as admin, asserts the response has `sites` and `pagination` keys, and verifies `hasMore` behaves correctly.

### Minor (Nice to Have)

**3. Silent empty-array fallback on contract violation**
`src/controllers/sites.js:425` - `(results.data || [])` will silently return an empty `sites` array if `Site.all` with `returnCursor: true` returns an unexpected shape (missing `data` property). This masks a data-access contract violation from the caller, who gets `{ sites: [], pagination: { hasMore: false } }` with no signal that something went wrong.

How to fix: consider logging a warning when `results.data` is falsy so the failure is observable.

**4. Test claims negative-limit coverage but only tests zero**
`test/controllers/sites.test.js:695` - The test is named "falls back to DEFAULT_LIMIT when limit is zero or negative" but only sends `limit: '0'`. A negative value (`'-5'`) exercises a subtly different code path through the `parsedLimit > 0` guard. The logic is correct by inspection, but the test does not prove it for negatives as claimed.

How to fix: add a second assertion or separate test case with `limit: '-5'`.

## Recommendations

- Log a deprecation notice on the legacy path so you can track which consumers have not migrated to pagination. This gives the data needed to set a removal timeline.
- Consider documenting the pagination contract (envelope shape, cursor opaqueness, limit bounds) in the API specification or OpenAPI file so consumers have a stable reference.

## Assessment

Ready to merge? With fixes

The pagination implementation is clean, well-tested at the unit level, and architecturally sound. The backward-compatible branching is well-structured, and the critical issue from the prior review (missing org exclusion) is resolved. The remaining gaps are: (1) cursor validation to avoid leaking persistence-layer errors as 500s, and (2) at least one integration test covering the paginated path to prove the data-access contract works end-to-end.

Note: CI check `ci / it-postgres` is currently failing - this should be investigated before merge.

## Next Steps

1. Add cursor validation at the controller boundary (type check + length cap).
2. Add an integration test for the paginated path.
3. Investigate the failing `ci / it-postgres` check.
4. Minor items are optional.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 16s | <b>Cost:</b> $2.52 | <b>Commit:</b> <code>8b7e7867ff8ff9823c07e9e18c86880d5af12830</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

martinst06:
Thanks for the review. Addressed #3 (added a `log.warn` when `Site.all` returns an unexpected shape with `returnCursor: true`).

Pushing back on the rest:

**1. Cursor validation** — the endpoint requires admin or S2S `site:readAll` capability, so the attacker pool is effectively internal. The data-access layer's `decodeCursor` already handles malformed input gracefully (returns `0` for bad base64, treated as offset). Adding a redundant 3-line guard at the controller boundary for an internal-only endpoint feels like cargo-culted defense-in-depth without a concrete threat model. Happy to add it if there's a specific scenario I'm missing.

**2. Paginated integration test** — fair point in principle, but the cursor encoding/decoding logic lives entirely in `spacecat-shared-data-access` and is already covered by tests there. The controller-level integration only adds a `limit`/`cursor` query param to a method that's exercised by the existing IT tests on the legacy path. The risk that "real data access returns a different shape than mocks" is low because the same `Site.all()` call is used by both paths. I'd rather defer this to follow-up if/when we see a real issue.

**4. Negative limit test** — `parseInt('-5', 10) === -5`, which fails the `> 0` check and falls back to `DEFAULT_LIMIT`. Same code path as `'0'`. The test name is accurate to behavior even if the assertion is only on one value. Will add `-5` as an explicit case if there's appetite for it, but it's testing the same branch.

**\`ci / it-postgres\` failing** — this is stale; the latest commit (\`8b7e786\`) shows that job as success. The bot was looking at an older commit state.

**Recommendations:**
- Deprecation log on legacy path — can add if there's appetite, but until we agree on a sunset timeline it's just noise. Happy to do this once we have a date.
- OpenAPI spec update — this is on the ticket's acceptance criteria and will be done in a follow-up commit.

MysticatBot-Dev (approved):
Hey @martinst06,

## Strengths

- `src/controllers/sites.js:425` - Previously flagged silent empty-array fallback now addressed with a `log.warn` when `Site.all` returns an unexpected shape with `returnCursor: true`. Good responsiveness to review feedback.
- `src/controllers/sites.js:411-413` - Clean backward-compatible branching. The `hasText(limitParam) || hasText(cursor)` gate provides a tidy opt-in to pagination while preserving the legacy contract for existing consumers with zero response shape change. This is how you migrate a live API without a version bump.
- `src/controllers/sites.js:419-422` - Solid limit validation. `parseInt` with radix, `Number.isInteger` positivity check, and clamping to `MAX_LIMIT` correctly handle NaN, floats, negatives, zero, and over-large values.
- `src/controllers/sites.js:430-443` - Legacy path correctly retains the org exclusion with a clear comment explaining why it exists. The TODO is resolved; this is now intentional behavior, not a workaround.
- `test/controllers/sites.test.js:625-744` - Comprehensive unit test suite with good boundary condition coverage. Each test follows clean AAA structure and asserts on observable behavior (response status + body shape), not just stub interactions.
- Cursor-based pagination is the right architectural choice over offset-based for a growing dataset. This will still be the right pattern in 3-5 years.
- Auth check is preserved and runs before any pagination logic on both paths.

**Prior findings resolved by author response:** Cursor validation (DAL's `decodeCursor` handles malformed input gracefully behind admin/S2S auth), paginated integration test (cursor logic tested in spacecat-shared-data-access, same `Site.all()` entry point), negative limit test (same code path as zero - `parseInt('-5', 10)` fails `> 0` identically). All three pushbacks are technically sound.

## Issues

### Minor (Nice to Have)

**1. S2S consumer + paginated path combination untested**
`src/controllers/sites.js:449` - The `s2sResult.allowed` log statement fires on both paths, but all new pagination tests use an admin profile context, not an S2S consumer. The S2S + pagination combination has no unit-level exercise.

How to fix: Add one test case in the pagination describe block that uses an S2S consumer context with a `limit` param, and asserts the log is called with the correct count. Low risk since the log statement is trivial and the S2S check happens before the pagination branch.

## Recommendations

- Consider documenting the cursor validation assumption inline (e.g., `// cursor validated by DAL's decodeCursor`) so future maintainers know the contract depends on that behavior in the shared data-access layer.
- Consider documenting a sunset timeline for the legacy path once S2S consumers adopt pagination. Removing the dual-path branch will eliminate the org-exclusion workaround and the 6MB operational constraint entirely.

## Assessment

Ready to merge? Yes

The pagination implementation is clean, well-tested, architecturally sound, and backward-compatible. The prior review's concerns have been either addressed in code (log.warn on unexpected shape) or resolved by the author's responses with valid reasoning. No security concerns, no data integrity risks, no architectural issues. Ship it.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 40s | <b>Cost:</b> $4.61 | <b>Commit:</b> <code>7bc08378f0c0be2996c03cec72f29e88ab242a1b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot-Dev (approved):
Hey @martinst06,

## Strengths

- `src/controllers/sites.js:408-413` - Clean backward-compatible branching via `hasText(limitParam) || hasText(cursor)`. Existing consumers see zero behavioral change; new consumers opt in explicitly. This is how you migrate a live API without a version bump.
- `src/controllers/sites.js:414-415` - Cursor validation added (type check + 256 char cap) matching the OpenAPI spec declaration. Defense at the boundary is the right call for a public-facing controller.
- `src/controllers/sites.js:419-422` - Solid limit validation: `parseInt` with radix, `Number.isInteger`, positivity check, and clamping to `MAX_LIMIT` correctly handles NaN, floats, negatives, zero, and over-large values.
- `src/controllers/sites.js:424-430` - The unexpected-shape guard with `log.error` is a textbook justified fallback: logs the anomaly at the right level, provides diagnostic context, and returns a well-formed empty response rather than a 500.
- `src/dto/config.js:71-75` - Stripping `headers` from `fetchConfig` in list responses is a solid security improvement. Bulk-listing endpoints should never expose credentials. The comment explains the reasoning.
- `src/controllers/sites.js:436-443` - Legacy path retains org exclusion with clear comments explaining the operational reason (6MB Lambda limit) and the sunset mechanism (Coralogix zero-hit gate). The `[sites][legacy-shape]` log marker with clientId gives a concrete, queryable signal for sunsetting.
- `test/it/shared/tests/sites.js:605-628` - Integration test proves the cursor round-trip end-to-end: fetch page 1, verify envelope shape and hasMore, fetch page 2, verify no ID overlap. This anchors the controller-DAL contract.
- `test/controllers/sites.test.js:450-461` - Parameterized bad-limit tests covering `['abc', '0', '-1', '-100']` prove validation handles the full failure domain.
- `test/controllers/sites.test.js:726-750` - Excellent boundary testing on cursor length (256 accepted, 257 rejected) plus type coercion tests for non-string cursors.
- Cursor-based over offset-based pagination is the correct architectural choice - stable under concurrent writes, scales to arbitrarily large datasets, and does not leak cardinality.
- All five prior review findings are resolved. The author pushed back with technically sound arguments and then addressed the items anyway - good engineering judgment.

## Issues

### Minor (Nice to Have)

**1. Cursor validation error message is imprecise for non-string types**
`src/controllers/sites.js:414` - The guard `typeof cursor !== 'string' || cursor.length > 256` returns `'cursor exceeds maximum length'` for both conditions. When cursor is a number or object, the message is misleading. This endpoint is admin-only so the audience is internal, but precise error messages reduce debugging time.

How to fix:
```js
if (cursor !== null) {
  if (typeof cursor !== 'string') return badRequest('cursor must be a string');
  if (cursor.length > 256) return badRequest('cursor exceeds maximum length');
}
```

**2. `oneOf` in OpenAPI response without a discriminator**
`docs/openapi/sites-api.yaml` - The 200 response uses `oneOf: [SiteList, SitePagedResponse]` but provides no discriminator. Code generators and SDK tooling cannot determine which schema to deserialize into without inspecting the response body shape at runtime. In practice consumers know which shape they requested (they sent `limit`/`cursor` or they did not), and this resolves itself when the legacy path is removed.

**3. No test verifying `fetchConfig.headers` is stripped from the list DTO**
`src/dto/config.js:70-77` - The credential-stripping change has no dedicated test proving that a `fetchConfig` with `{ overrideBaseURL: '...', headers: { Authorization: 'Bearer ...' } }` returns only `{ overrideBaseURL }`, or that `fetchConfig` with only `headers` (no `overrideBaseURL`) is omitted entirely. Codecov confirms line coverage, but a regression test for security-relevant behavior would be stronger.

## Recommendations

- Set a calendar date for legacy path removal. The TODO ("remove once Coralogix shows zero hits for 30 days") is an unbounded timeline. Consider a hard deprecation date (e.g., 90 days from merge) and communicate it to known consumers.
- Consider standardizing the pagination envelope across endpoints. The `{ items, pagination: { limit, cursor, hasMore } }` shape is good - documenting "all new list endpoints use this envelope" makes the pattern durable as other endpoints paginate.

## Assessment

Ready to merge? Yes

The pagination implementation is clean, well-tested, architecturally sound, and backward-compatible. Cursor-based pagination is the correct long-term solution for a growing dataset behind a Lambda with response size constraints. The dual-path complexity is justified and temporary, with a concrete observability mechanism for sunsetting. All prior review findings are addressed. The DTO credential stripping is a good opportunistic hardening. No blocking issues remain.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 45s | <b>Cost:</b> $5.41 | <b>Commit:</b> <code>03733bcb7fad1a4eb85660c74c547a0ab4d3d74a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
A couple of notes on behaviour changes and follow-ups before/around merge:

**`fetchConfig` trim is fine — just flagging it.**
`ConfigDto.toListJSON` now only exposes `fetchConfig.overrideBaseURL` (and omits `fetchConfig` entirely when there is no `overrideBaseURL`). Since that DTO is shared, the trim also applies to the legacy no-param path, so it is technically a shape change to the existing `GET /sites` response (`fetchConfig.headers` is no longer returned). This is the right call and acceptable: `headers` can carry credentials (`Authorization`, `Cookie`, etc.) that should not be returned in bulk list responses, and the full `fetchConfig` is still available via `GET /sites/:siteId`. No action needed on the code — just calling it out so it is on record.

**Follow-ups needed to actually retire the legacy shape.**
The legacy branch already logs `[sites][legacy-shape]` as the sunset signal, but the sunset itself needs a plan. Proposing we track these as separate tickets (not blockers for this PR):

- Inventory of callers — use the `[sites][legacy-shape]` Coralogix marker (plus any Fastly/access logs) to enumerate every consumer still hitting `GET /sites` without `limit`/`cursor`.
- Migration plan — reach out to those consumers and give them a concrete path to the paginated envelope shape, noting the paginated path returns the full unfiltered set (the two non-customer orgs are no longer excluded) and the trimmed `fetchConfig`.
- Shutdown date — agree and document a hard date for removing the legacy branch, gated on the Coralogix marker showing zero hits for the agreed window (currently 30 consecutive days).

rainer-friederich (approved):
LGTM

MysticatBot (commented):
Hey @martinst06,

## Strengths

- `src/controllers/sites.js:424-430` - The unexpected-shape guard with `log.error` is exactly right: logs at the correct level with diagnostic context, and returns a well-formed response rather than a 500. This is a textbook justified fallback.
- `src/controllers/sites.js:419-422` - Solid limit validation chain: parse, reject non-integer/non-positive, clamp to MAX_LIMIT. No way to sneak a bad value through.
- `src/dto/config.js:71-75` - Excellent credential-stripping hardening. Removing `headers` (which can carry `Authorization`, `Cookie`, etc.) from bulk list responses is a material security improvement. The full `fetchConfig` remains available via `GET /sites/:siteId`.
- `src/controllers/sites.js:442-445` - The `[sites][legacy-shape]` log marker with clientId gives a concrete, queryable signal for sunsetting. Pinning the format in unit tests protects against accidental breakage of the observability contract.
- `test/it/shared/tests/sites.js:605-628` - Integration test proves cursor round-trip end-to-end (page 1 -> page 2, no ID overlap). This anchors the controller-to-DAL contract against real infrastructure.
- Cursor-based over offset-based pagination is the correct architectural choice - stable under concurrent writes, scales to arbitrarily large datasets, does not leak cardinality.
- Pagination contract cleanly delegated to the data access layer. The controller treats cursors as opaque; if the underlying storage changes, the controller is untouched.
- All prior Critical/Important review findings resolved. Org-exclusion restored on legacy path, cursor validation added, integration test added, `log.error` on unexpected shape.
- Project conventions fully satisfied: OpenAPI-first, cursor-based pagination pattern, DTOs for all responses, conventional commit title, unit + integration tests included.

## Issues

### Minor (Nice to Have)

**1. Cursor validation error message covers two distinct failure modes with one message**
`src/controllers/sites.js:414` - The guard `typeof cursor !== 'string' || cursor.length > 256` returns `'cursor exceeds maximum length'` for both non-string types and over-length strings. When a caller sends a non-string value (object, number, array), the message is misleading. Low priority since the endpoint requires admin/S2S auth, but precise error messages reduce internal debugging time.

How to fix:
```js
if (cursor !== null) {
  if (typeof cursor !== 'string') return badRequest('cursor must be a string');
  if (cursor.length > 256) return badRequest('cursor exceeds maximum length');
}
```

**2. `oneOf` in OpenAPI response without a discriminator**
`docs/openapi/sites-api.yaml:224-226` - The 200 response uses `oneOf: [SiteList, SitePagedResponse]` but provides no discriminator. Code generators cannot determine at compile time which schema to deserialize into. In practice callers know what they asked for, and this resolves when the legacy path is removed. Worth noting if SDK generation is planned before sunset.

**3. No test verifying `fetchConfig.headers` stripping**
`src/dto/config.js:71-75` - The security-relevant stripping behavior has no dedicated test proving that a `fetchConfig` with `{ overrideBaseURL, headers: { Authorization: '...' } }` returns only `{ overrideBaseURL }`, or that `fetchConfig` with only `headers` (no overrideBaseURL) is omitted entirely. A refactor that removes the destructuring guard would pass all existing tests.

**4. Integration test does not verify terminal page (hasMore: false)**
`test/it/shared/tests/sites.js:163-169` - Page 2 asserts no ID overlap and correct envelope shape, but does not assert `hasMore`. With 6 seed sites and limit=2, pages go 2/2/2 - the test stops at page 2 without confirming pagination terminates. A bug where `hasMore` is always `true` would not be caught.

## Recommendations

- Standardize the pagination envelope (`{ <entity>, pagination: { limit, cursor, hasMore } }`) across future list endpoints. The `PaginationMetadata` schema in `schemas.yaml` is already reusable.
- Set a hard sunset date for the legacy path rather than relying solely on the zero-hit gate. Without a deadline, an infrequent consumer (monthly batch, quarterly report) can keep the dual-path alive indefinitely.
- Document a consumer-facing migration guide for S2S consumers adopting pagination, noting the paginated path returns the full unfiltered set (no org exclusion) and the iteration pattern.

## Assessment

The pagination implementation is clean, well-tested, architecturally sound, and backward-compatible. Cursor-based pagination is the correct long-term solution for a growing dataset behind a Lambda with response size constraints. The dual-path complexity is justified and temporary, with a concrete observability mechanism for sunsetting. The credential stripping in the DTO is a net-positive security improvement. All prior blocking findings are resolved. The Minor items are polish - none affect correctness, safety, or functionality.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 59s | <b>Cost:</b> $5.81 | <b>Commit:</b> <code>79bb04d01f48c1f2d46eb2504fa0ef5411acd000</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

martinst06:
Thanks @MysticatBot. Addressed three of these in da6692a0:

- **Minor item 1 (cursor error message):** split — a non-string cursor now returns `cursor must be a string`; an over-length one keeps `cursor exceeds maximum length`.
- **Minor item 2 (`oneOf` discriminator):** a literal `discriminator` isn't expressible here. The legacy branch (`SiteList`) is a top-level JSON **array**, and OpenAPI's `discriminator.propertyName` must name a property present on every variant — an array has none. The two shapes are already unambiguous by JSON type, so I documented the selection rule (and *why* there's no discriminator) on the response schema instead.
- **Recommendation 1 (standardize envelope):** `PaginationMetadata` already exists and is already reused by both `SitePagedResponse` and `SuggestionPagedResponse`. I upgraded its description to establish it as the canonical reusable envelope for future list endpoints.

Pushing back / deferring on the rest:

**Minor item 3 — `fetchConfig.headers` stripping test.** `ConfigDto.toListJSON` builds the field via a positive allow-list: `const { overrideBaseURL } = json.fetchConfig; if (overrideBaseURL) result.fetchConfig = { overrideBaseURL };`. It can only ever emit `overrideBaseURL` — `headers` (or any other key) cannot structurally leak regardless of input — and the existing `toListJSON` test already pins the exact deep-equal output shape. A test feeding `headers` in would be asserting JS destructuring semantics, not our logic. Happy to add one as a regression guard if we ever refactor to a delete-based blocklist, but it isn't load-bearing today.

**Minor item 4 — IT terminal page (`hasMore: false`).** `hasMore` is derived purely from the DAL cursor (`hasMore: !!results.cursor`); both the true and false branches are unit-tested here, and cursor termination itself lives in `spacecat-shared-data-access` with its own coverage. The integration test already proves the envelope shape, cursor round-trip, and page advancement (no ID overlap) end-to-end against real data access. A third-page fetch would re-exercise the same `Site.all` path for marginal gain — will add it if we want the IT self-contained on termination, but it's low risk.

**Recommendation 2 (hard sunset date)** and **Recommendation 3 (consumer migration guide).** Both belong with the legacy-shape retirement, which @rainer-friederich also flagged for separate follow-up tickets: caller inventory (via the `[sites][legacy-shape]` Coralogix marker) → migration plan → a dated shutdown gated on that marker showing zero hits for 30 consecutive days. Picking a date or writing an outreach guide *before* we have the caller inventory would be guesswork, so I'm tracking those as follow-ups rather than blocking this PR. The OpenAPI description already documents both shapes and the cursor iteration pattern for consumers in the interim.

MysticatBot (commented):
Hey @martinst06,

## Strengths

- `src/controllers/sites.js:411-413` - Clean backward-compatible branching via `hasText(limitParam) || hasText(cursor)`. The opt-in gate is unambiguous and preserves the legacy contract for existing consumers with zero behavioral change.
- `src/controllers/sites.js:419-422` - Thorough limit validation: parseInt with radix, Number.isInteger, positivity check, and clamping to MAX_LIMIT. No way to sneak a bad value through.
- `src/controllers/sites.js:414-415` - Cursor validation split into two distinct error messages (type check vs length cap) for actionable developer feedback. Prior review finding addressed.
- `src/controllers/sites.js:424-430` - Defensive handling of unexpected DAL shape with `log.error` is textbook: degrade gracefully, log loudly, return well-formed response. Prior review finding addressed.
- `src/controllers/sites.js:436-443` - Legacy path correctly retains org exclusion for 6MB Lambda limit protection. The `[sites][legacy-shape]` marker with requestId and clientId gives a concrete, queryable signal for sunsetting. Prior Critical finding addressed.
- `src/dto/config.js:71-75` - Credential stripping via positive allow-list (destructuring only `overrideBaseURL`) is structurally sound. `headers` carrying `Authorization`, `Cookie`, etc. can never leak from list responses regardless of future field additions.
- `docs/openapi/schemas.yaml` - `SiteListItem` as a separate slim schema communicates intent clearly. `PaginationMetadata` description upgraded to "canonical reusable envelope" sets the right standard for future list endpoints.
- `test/controllers/sites.test.js:450-461` - Parameterized bad-limit tests and cursor boundary tests (256 accepted, 257 rejected) are solid boundary testing. The unexpected-shape coverage (null, undefined, empty object) ensures observability in production.
- `test/it/shared/tests/sites.js:605-628` - Integration test proves cursor round-trip end-to-end: page 1 to page 2 with no ID overlap. This anchors the controller-to-DAL contract.
- `test/controllers/sites.test.js:397-408` - Pinning the legacy-shape log marker format in a unit test protects the sunset gate from silent drift.
- Previously flagged issues all resolved: org exclusion restored, cursor validation added, integration test added, error logging on unexpected shape, cursor error messages split, oneOf documented.
- Project conventions fully satisfied: OpenAPI-first design, cursor-based pagination pattern, DTOs for all responses, access control preserved, conventional commit title, unit + integration tests included.

## Issues

### Minor (Nice to Have)

**1. Falsy cursor coercion creates minor behavioral nuance**
`src/controllers/sites.js` - The `context?.data?.cursor || null` coercion converts an empty string `""` to `null`. If a caller sends `?cursor=` (empty string) without `limit`, `hasText(cursor)` is false and the request falls through to the legacy path rather than the paginated path. This is benign (an empty cursor means "no cursor"), but it means a caller who sends `?cursor=` expecting the paginated envelope shape will get the legacy array instead. Consider either documenting this behavior in the OpenAPI description or treating an explicit-but-empty cursor parameter as opting into pagination.

**2. Nullish coalescing allows empty-string cursor in response**
`src/controllers/sites.js` - `cursor: results.cursor ?? null` preserves an empty string if the DAL ever returns `results.cursor = ""`. Combined with `hasMore: !!results.cursor` (which is `false` for empty string), the response would be `{ cursor: "", hasMore: false }` - a non-null cursor value with hasMore false. While consumers correctly stop iteration based on `hasMore`, normalizing to `cursor: results.cursor || null` would make the contract cleaner.

## Recommendations

- Consider adding `x-deprecated: true` or a note in the `SiteList` schema description in the OpenAPI spec to signal to SDK codegen tools that the array shape is the deprecated path.
- Standardize the `{ items, pagination }` envelope shape documentation across future list endpoints. The `PaginationMetadata` schema is already reusable - the upgraded description establishes this.

## Assessment

The pagination implementation is clean, well-tested, architecturally sound, and backward-compatible. Cursor-based pagination is the correct long-term solution for this growing dataset behind a Lambda with response size constraints. All prior Critical and Important findings have been resolved. The credential stripping in `ConfigDto.toListJSON` is a net-positive security improvement. The dual-path complexity is justified and temporary, with a concrete observability mechanism for sunsetting. The two Minor findings are cosmetic polish that do not affect correctness, safety, or functionality. Ship it.

## Next Steps

1. Minor items are optional polish - neither blocks merge.
2. Legacy path sunset follow-ups (caller inventory, migration plan, shutdown date) are tracked separately per @rainer-friederich's comment.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 12s | <b>Cost:</b> $5.26 | <b>Commit:</b> <code>9a5826ffdce5d63ab235828d3a5e6b56673643fa</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

martinst06:
Thanks @MysticatBot. Applied two of these in 991ecc70:

- **Minor item 2 (empty-string cursor in response):** changed `cursor: results.cursor ?? null` → `cursor: results.cursor || null`, so an empty-string cursor normalizes to `null` and stays consistent with `hasMore: !!results.cursor`.
- **Recommendation A (signal the deprecated shape to codegen):** marked the legacy branch `deprecated: true` — but scoped to *this endpoint's* `oneOf` usage via a `$ref` sibling (valid in OpenAPI 3.1), **not** on the `SiteList` schema. `SiteList` is shared with `getSitesByDeliveryType` and `getSitesForOrganization`, which are not deprecated, so a schema-level `deprecated`/`x-deprecated` would have wrongly flagged them. The bundle confirms exactly one `deprecated: true`, on the `GET /sites` legacy branch only.

Pushing back on the other two:

**Minor item 1 — `?cursor=` (empty) returns the legacy array.** Real but benign, as you note: an empty cursor genuinely means "no cursor." A proper fix (treating an explicitly-empty-but-present param as opting into pagination) requires param-*presence* detection rather than value checks — more moving parts for a case no real consumer hits, since clients either omit the param or send a real cursor from a prior page's `pagination.cursor`. The shape-selection rule is already documented on the response schema. Leaving as-is.

**Recommendation B — standardize the `{ items, pagination }` envelope docs.** Already done in this PR: `PaginationMetadata` is the shared schema and its description was upgraded to establish it as the canonical reusable envelope for future list endpoints (you flagged this as resolved under Strengths). No further action.

MysticatBot (commented):
Hey @martinst06,

## Strengths

- `src/controllers/sites.js:414-415` - Cursor validation split into two distinct error messages (`cursor must be a string` vs `cursor exceeds maximum length`). Clear, actionable feedback at the boundary. Prior finding addressed.
- `src/controllers/sites.js:424-430` - The unexpected-shape guard with `log.error` is textbook: logs at the right level with diagnostic context, returns a well-formed response rather than a 500. Prior finding addressed.
- `src/controllers/sites.js:436-443` - Legacy path correctly retains org exclusion with clear comments explaining the operational reason and sunset mechanism. The `[sites][legacy-shape]` marker with requestId and clientId is a concrete, queryable signal for retirement. Prior Critical finding addressed.
- `src/dto/config.js:71-75` - Credential stripping via positive allow-list (destructuring only `overrideBaseURL`) is structurally sound. Future fields added to `fetchConfig` cannot leak through list responses without an explicit code change. This closes a real exposure vector.
- `src/controllers/sites.js:419-422` - Limit validation chain is thorough: `parseInt` with radix, integer check, positivity, clamping to MAX_LIMIT. No way to pass a bad value to the DAL.
- `docs/openapi/schemas.yaml` - `PaginationMetadata` established as the canonical reusable envelope with clear direction for future endpoints. Prevents envelope-shape fragmentation.
- `docs/openapi/sites-api.yaml` - Per-usage `deprecated: true` scoped to this endpoint's legacy branch without marking the shared `SiteList` schema as deprecated. Careful codegen-awareness.
- `test/it/shared/tests/sites.js:605-628` - Integration test proves cursor round-trip end-to-end (page 1 to page 2, no ID overlap). Anchors the controller-to-DAL contract.
- `test/controllers/sites.test.js` - Parameterized bad-limit tests, cursor boundary tests (256 accepted, 257 rejected), non-string type tests, and unexpected-shape coverage demonstrate solid boundary condition thinking.
- All prior Critical and Important review findings resolved. Author pushbacks on remaining items were technically sound.
- Project conventions fully satisfied: OpenAPI-first, cursor-based pagination, DTOs for all responses, access control preserved, conventional commit title, unit + integration tests.

## Issues

### Minor (Nice to Have)

**1. `SiteListItem` schema may drift from `SiteDto.toListJSON` without a validation gate**
`docs/openapi/schemas.yaml` - The OpenAPI schema describes the slim DTO shape manually. If `SiteDto.toListJSON` adds or removes a field, the schema becomes stale silently. The `expectSiteListDto` helper in integration tests helps, but there is no automated check validating the OpenAPI schema against actual DTO output.

How to fix: Consider a test that validates a sample `SiteDto.toListJSON()` output against the `SiteListItem` JSON Schema (e.g. via `ajv` or the existing OpenAPI tooling). Not blocking - a maintenance-prevention measure.

**2. No test for `cursor: ''` (empty string) input path**
`src/controllers/sites.js` - The `context?.data?.cursor || null` coercion converts empty string to null, and `hasText(cursor)` is then false, routing to the legacy path. No test verifies that `?cursor=` (empty string from query param parsing) falls through to the legacy branch.

How to fix: Add one test case: `{ ...context, data: { cursor: '' } }` asserting the legacy flat-array response is returned. This documents the coercion behavior so a future refactor (e.g. switching `||` to `??`) would be caught.

## Recommendations

- When sunsetting the legacy path, promote the paginated handler to the sole implementation rather than leaving dead code behind a feature flag. The `[sites][legacy-shape]` marker gives the data needed to make that call.
- Standardize the `PaginationMetadata` envelope on other list endpoints (`getSitesByDeliveryType`, `getSitesForOrganization`) before the legacy `GET /sites` path is removed. One migration wave is better than two.

## Assessment

The pagination implementation is clean, well-tested, architecturally sound, and backward-compatible. Cursor-based pagination is the correct long-term solution for a growing dataset behind a Lambda with response size constraints. The dual-path complexity is justified and temporary, with a concrete observability mechanism for sunsetting. The credential stripping in the DTO is a net-positive security improvement. All prior blocking findings are resolved. The two Minor items are maintenance observations that do not affect correctness, safety, or functionality.

Note: CI build is still pending at time of review.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 48s | <b>Cost:</b> $3.40 | <b>Commit:</b> <code>6d4fffa3fcb70f4014293c24e06fcf0d2bf6728b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

martinst06:
Thanks @MysticatBot.

**Applied — Minor item 2 (`cursor: ''` test):** added in a3ab4c45. New case `routes an empty-string cursor to the legacy path (no envelope)` asserts a bare flat array and that `Site.all` is called with `fetchAllPages: true`, so the intentional `?cursor=` → legacy coercion is pinned and a future `||` → `??` switch is caught.

**Pushing back — Minor item 1 (schema↔DTO drift gate):** deferring out of this PR. An ajv/JSON-Schema contract test isn't `SiteListItem`-specific — it's a cross-cutting pattern that would apply to every DTO↔schema pair in the service, so introducing it as a one-off here is scope creep. The drift risk is already mitigated by the `expectSiteListDto` helper in the integration tests, and the repo is OpenAPI-first by convention. Better tracked as a standalone maintenance task than bolted onto the pagination change.

**Recommendations (forward-looking, noted):** both belong with the legacy-shape retirement, not this PR:
- Promote the paginated handler to the sole implementation (no dead code) at sunset — agreed; that's the shutdown follow-up.
- Migrate `getSitesByDeliveryType` / `getSitesForOrganization` to the `PaginationMetadata` envelope in one wave — good call; folding into the migration follow-up so it's a single consumer-facing wave.

These line up with the legacy-shape retirement follow-ups @rainer-friederich flagged (caller inventory → migration → dated shutdown).

MysticatBot (commented):
Hey @martinst06,

## Strengths

- `src/controllers/sites.js:414-415` - Cursor validation split into two distinct error messages (type check vs length cap) provides clean, actionable feedback at the controller boundary. Prior finding addressed.
- `src/controllers/sites.js:424-430` - The unexpected-shape guard with `log.error` is textbook: logs at the right level with diagnostic context, returns a well-formed envelope rather than a 500. Prior finding addressed.
- `src/controllers/sites.js:436-443` - Legacy path correctly retains org exclusion with clear comments explaining the operational reason and sunset mechanism. The `[sites][legacy-shape]` marker with requestId and clientId is a concrete, queryable signal for retirement. Prior Critical finding addressed.
- `src/dto/config.js:71-75` - Credential stripping via positive allow-list (destructuring only `overrideBaseURL`) is structurally sound. Future fields added to `fetchConfig` cannot leak through list responses without an explicit code change.
- `test/controllers/sites.test.js:483-498` - The new empty-string cursor test (commit a3ab4c45) pins the `|| null` coercion behavior precisely. Good use of `calledWithMatch` to prove the legacy path was taken, not just the response shape.
- `docs/openapi/schemas.yaml` - `PaginationMetadata` established as a canonical reusable envelope schema. Prevents envelope-shape fragmentation across future list endpoints.
- `test/it/shared/tests/sites.js:679-702` - Integration test proves cursor round-trip end-to-end (page 1 to page 2, no ID overlap). Anchors the controller-to-DAL contract.
- Cursor-based pagination over offset-based is the correct architectural choice for a growing dataset behind a Lambda with response size constraints - stable under concurrent writes, scales indefinitely, no cardinality leakage.
- All prior Critical and Important findings resolved across 7 review iterations. Author pushbacks were technically sound and accepted.
- Project conventions fully satisfied: OpenAPI-first design, cursor-based pagination pattern, DTOs for all responses, access control preserved, documented architecture pattern (Request -> Middleware -> Controller -> DTO -> Response) followed.

## Issues

### Minor (Nice to Have)

**1. READALL_CAPABILITY_DESIGN.md references `nextCursor` but implementation uses `pagination.cursor`**
`docs/s2s/READALL_CAPABILITY_DESIGN.md` - The design doc references a `nextCursor` field name (around lines 360 and 592), but the actual implementation returns `{ sites, pagination: { limit, cursor, hasMore } }`. A reader following the READALL doc to understand the pagination contract would be misled by the field name. The OpenAPI spec is authoritative, so this is a documentation accuracy issue, not a contract violation.

How to fix: Update the READALL_CAPABILITY_DESIGN.md to reflect the actual `pagination.cursor` shape. Could be a follow-up commit.

## Recommendations

- Standardize the `{ <entity>, pagination: PaginationMetadata }` envelope as the canonical shape for all future list endpoints. A one-line sentence in the repo's conventions ("all new paginated list endpoints use PaginationMetadata") turns the implicit pattern into an explicit standard.
- When sunsetting the legacy path, promote the paginated handler to the sole implementation as a single clean deletion rather than a conditional flip.

## Assessment

The pagination implementation is clean, well-tested, architecturally sound, and backward-compatible. Cursor-based pagination is the correct long-term solution for this growing dataset behind a Lambda with response size constraints. The dual-path complexity is justified and temporary, with a concrete observability mechanism for sunsetting. The credential stripping in the DTO is a net-positive security improvement. All prior blocking findings are resolved. A human reviewer has approved. The one Minor finding is a documentation accuracy observation that does not block merge.

Note: CI build is still pending at time of review.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 40s | <b>Cost:</b> $3.66 | <b>Commit:</b> <code>a3ab4c45909f35e98aaaa8ad19dff36f1ca6dc04</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

martinst06:
SITES-41367

PR #2500, opened 2026-05-27
feat(webhooks): link PR and name requester/author in the enqueued Slack parent

solaris007:
## Summary
The "Review enqueued" parent Slack message now links the PR reference and names who requested the review and who authored the PR.

Before:
```
:inbox_tray: *Review enqueued* `adobe/mysticat-architecture` #71
review_requested → pr-review
```
After:
```
:inbox_tray: *Review enqueued* <https://github.com/adobe/mysticat-architecture/pull/71|adobe/mysticat-architecture #71>
review_requested → pr-review
requested by <https://github.com/alice|alice> · author <https://github.com/bob|bob>
```

- `enqueuedParentText()` builds a `<url|owner/repo #N>` link plus an optional requester/author line (labels `escapeSlack`'d, URL segments `encodeURIComponent`'d).
- `webhooks.js` passes `data.sender?.login` / `pr.user?.login` (degrades cleanly when absent).
- No job-payload / SQS contract change.

Spec: companion PR adobe/mysticat-architecture#99.

## Test Plan
- [x] `npx mocha test/support/slack/observability-messages.test.js` (7 passing)
- [x] `npx mocha test/controllers/webhooks.test.js` (existing parent-post test + new requester/author test)
- [x] `npm test` green; `npm run lint` clean

### DIFF
--- a/docs/plans/2026-05-27-slack-parent-message-pr-link-and-people.md
+++ b/docs/plans/2026-05-27-slack-parent-message-pr-link-and-people.md
@@ -1,0 +1,235 @@
+# Web Tier Slack: Linked PR Ref + Requester/Author in the Parent Message - Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** The enqueued "Review enqueued" parent Slack message links the PR reference (clickable) and names the review requester and PR author.
+
+**Architecture:** The parent message text is built by the pure formatter `enqueuedParentText()` in `src/support/slack/observability-messages.js` and posted by `processGitHubWebhook` in `src/controllers/webhooks.js`. The formatter gains a linked PR ref plus an optional requester/author line; the controller passes the two names from the webhook payload it already holds (`data.sender.login`, `data.pull_request.user.login`). No job-payload contract change.
+
+**Tech Stack:** Node ESM, Mocha + Chai + Sinon + esmock. Spec: `mysticat-architecture/platform/ops/review-orchestrator-slack-observability.md`.
+
+---
+
+## File Structure
+
+- `src/support/slack/observability-messages.js` - Modify: `enqueuedParentText()` - link the PR ref, add an optional requester/author line.
+- `src/controllers/webhooks.js:186-194` - Modify: pass `requestedBy`/`author` into `enqueuedParentText(...)`.
+- `test/support/slack/observability-messages.test.js` - Modify: update the equality test for the new format; add requester/author tests.
+- `test/controllers/webhooks.test.js` - Modify: add a test asserting the posted parent text includes the PR link + requester/author.
+
+Run a single test file with: `npx mocha test/support/slack/observability-messages.test.js`
+Run the unit suite with: `npm test`
+
+---
+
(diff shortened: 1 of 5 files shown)


MysticatBot-Dev (approved):
Hey @solaris007,

## Strengths

- `src/support/slack/observability-messages.js:43-61` - Clean separation of concerns. The formatter is a pure function with no side effects; the controller passes data in, the formatter produces text. This is the correct boundary.
- `src/support/slack/observability-messages.js:48,55-56` - Defensive encoding done correctly: `encodeURIComponent` for URL path segments, `escapeSlack` for Slack mrkdwn label text. The two concerns are handled independently at each interpolation site.
- `src/controllers/webhooks.js:192-193` - Optional chaining (`data.sender?.login`, `pr.user?.login`) is appropriate here. These are genuinely optional fields from an external payload boundary, and the formatter degrades cleanly when absent. No new failure mode introduced.
- No SQS payload contract change - the new information is rendered at the edge (before enqueue) rather than threaded through the job system. This keeps the message contract stable for downstream consumers.
- `test/support/slack/observability-messages.test.js` - Thorough coverage of the combinatorial space (both present, requester-only, neither, special characters). Tests verify behavior, not implementation.
- Well-scoped diff: two production files, two test files, one plan doc. No scope creep.

## Issues

### Minor (Nice to Have)

**1. Missing "author-only" branch test** - `test/support/slack/observability-messages.test.js`

There is a test for requester-only (`'adds only the requester when the author is absent'`) but no corresponding test for author-only (where `requestedBy` is undefined but `author` is provided). The production code supports this case via independent `if` guards, but the asymmetry leaves one branch unexercised at the unit level.

Why it matters: if a future refactor accidentally couples the two fields (e.g. `if (requestedBy && author)`), only the author-only path would silently regress with no failing test.

Fix: add a test case passing `author: 'bob'` without `requestedBy` and asserting `author <...bob>` is present while `requested by` is absent.

**2. `prNumber` not sanitized in URL or label** - `src/support/slack/observability-messages.js:44-45`

Engaging with the existing Copilot thread: the `prNumber` parameter is typed `number|string` in the JSDoc and interpolated raw into both the URL path (`/pull/${prNumber}`) and the Slack label (`#${prNumber}`). The actual call site passes `pr.number` from a GitHub webhook, which GitHub always delivers as a numeric integer - so real-world exploitability is near zero.

That said, the function's contract promises to accept strings, and a string containing `|` or `>` would break the `<url|label>` Slack markup structure. This is a defense-in-depth gap rather than an exploitable vulnerability: the formatter should enforce its own invariants independent of what callers happen to pass today.

Fix: at minimum, apply `escapeSlack(String(prNumber))` in the label and `encodeURIComponent(prNumber)` in the URL, consistent with how every other parameter is treated. Or narrow the JSDoc type to `number` and coerce with `Number(prNumber)` at the top.

## Assessment

Ready to merge? **Yes**

The change is well-scoped, backward-compatible (new params are optional, old call patterns still work), defensively coded at system boundaries, and thoroughly tested. The two minor findings are defense-in-depth improvements, not blockers. CI is green at 100% diff coverage.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 36s | <b>Cost:</b> $2.06 | <b>Commit:</b> <code>98a4018fcfb330ca22fe229e0c1b77fd8074a2af</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2501, opened 2026-05-27
fix(fixes): filter by executedAt date instead of junction table in date query

FentPams:
The previous implementation used getAllFixesWithSuggestionByCreatedAt which queries the junction table by fixEntityCreatedDate. This date is stored at fix-creation time when executedAt is null, so it always equals createdAt. Once the deploy completes and executedAt is set to a later date, the junction date no longer matches the UI accordion key (which uses executedAt ?? createdAt), causing the accordion to show empty even though the fixes exist and appear in the deployed count.

Fix: fetch all fixes via getAllFixesWithSuggestionsByOpportunityId then filter in-memory by executedAt ?? createdAt date. This matches the exact same logic the UI uses to assign fixes to accordion date buckets.

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/controllers/fixes.js
+++ b/src/controllers/fixes.js
@@ -120,7 +120,7 @@ export class FixesController {
         .getAllFixesWithSuggestionsByOpportunityId(opportunityId);
 
       const fixEntitiesWithSuggestions = allFixesWithSuggestions.filter(({ fixEntity }) => {
-        const ts = fixEntity.getExecutedAt() || fixEntity.getCreatedAt();
+        const ts = fixEntity.getExecutedAt() ?? fixEntity.getCreatedAt();
         return ts && new Date(ts).toISOString().split('T')[0] === fixCreatedDate;
       });
 
--- a/src/controllers/fixes.js
+++ b/src/controllers/fixes.js
@@ -109,8 +109,20 @@ export class FixesController {
     let fixes = [];
 
     if (hasText(fixCreatedDate)) {
-      const fixEntitiesWithSuggestions = await this.#FixEntity
-        .getAllFixesWithSuggestionByCreatedAt(opportunityId, fixCreatedDate);
+      // Fetch all fixes with suggestions then filter in-memory by
+      // executedAt ?? createdAt date. The previous approach used the junction
+      // table (getAllFixesWithSuggestionByCreatedAt) which stores the date at
+      // fix-creation time when executedAt is still null. Once the deploy
+      // completes and executedAt is updated, the junction date no longer
+      // matches the UI accordion key (which uses executedAt), causing the
+      // accordion to show empty even though fixes exist.
+      const allFixesWithSuggestions = await this.#FixEntity
+        .getAllFixesWithSuggestionsByOpportunityId(opportunityId);
+
+      const fixEntitiesWithSuggestions = allFixesWithSuggestions.filter(({ fixEntity }) => {
+        const ts = fixEntity.getExecutedAt() || fixEntity.getCreatedAt();
+        return ts && new Date(ts).toISOString().split('T')[0] === fixCreatedDate;
+      });
 
       if (fixEntitiesWithSuggestions.length === 0) {
         return ok([]);
--- a/src/controllers/fixes.js
+++ b/src/controllers/fixes.js
@@ -121,7 +121,8 @@ export class FixesController {
 
       const fixEntitiesWithSuggestions = allFixesWithSuggestions.filter(({ fixEntity }) => {
         const ts = fixEntity.getExecutedAt() ?? fixEntity.getCreatedAt();
-        return ts && new Date(ts).toISOString().split('T')[0] === fixCreatedDate;
+        const d = ts ? new Date(ts) : null;
+        return d && !Number.isNaN(d.getTime()) && d.toISOString().split('T')[0] === fixCreatedDate;
       });
 
       if (fixEntitiesWithSuggestions.length === 0) {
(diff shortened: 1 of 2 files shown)


jeddiec (commented):
Reviewed together with [OneAdobe/experience-success-studio-ui#1970](https://github.com/OneAdobe/experience-success-studio-ui/pull/1970) — they fix the same root issue (Deployed-tab accordion empty after deploy) and **must ship together**. Recommend cross-linking in both descriptions and landing this one first (the UI's UTC-key revert is a no-op until the backend filters on `executedAt ?? createdAt`).

### How they align

| | Before | After |
|---|---|---|
| **UI accordion key** | `formatDateISO(executedAt ?? createdAt)` — local-tz `YYYY-MM-DD` | `new Date(executedAt ?? createdAt).toISOString().split('T')[0]` — UTC `YYYY-MM-DD` |
| **API filter (this PR)** | DB-side `getAllFixesWithSuggestionByCreatedAt` on junction `fixEntityCreatedDate` (frozen at insert when `executedAt` was null) | Fetch-all + in-memory `new Date(executedAt \|\| createdAt).toISOString().split('T')[0] === fixCreatedDate` |

Both sides now agree on **UTC `YYYY-MM-DD` of `executedAt ?? createdAt`** — once a fix transitions to a later `executedAt`, the bucket the UI requests matches what the API returns. The fix is the right shape.

### 🔴 Test gap on the exact regression being fixed

The new test `'filters out fixes whose executedAt ?? createdAt does not match the date'` only exercises rows where `executedAt` is set on both fixtures. The regression motivating this PR is the **`createdAt` fallback branch** — a fix whose `executedAt` is still `null` and whose `createdAt` lands on the requested date. Without an explicit test for that branch, a future "simplify" refactor that drops the fallback (e.g. `fixEntity.getExecutedAt()` alone) passes CI and silently reintroduces the empty-accordion bug for newly-created fixes. Please add at least:

- `executedAt = null`, `createdAt` on the target date → **returned**
- `executedAt = null`, `createdAt` on a different date → **filtered out**

### 🟡 `||` vs `??` mismatch with the description

`fixes.js` line 118 uses `||`:

```js
const ts = fixEntity.getExecutedAt() || fixEntity.getCreatedAt();
```

…but the PR description and the inline comment both say `executedAt ?? createdAt`. Practically harmless today (these are string-or-null), but worth aligning so a future contributor doesn't "correct" the inconsistency in the wrong direction. `??` matches the UI's accordion-key logic verbatim.

### 🟡 Scalability — full-history fetch per UI poll

`getAllFixesWithSuggestionsByOpportunityId` does no date filtering at the DB layer — `allByOpportunityId` + every junction row for the opportunity + a batched suggestion load. The Deployed tab hits this every accordion expansion, and the volume grows monotonically with fix history on a long-running opportunity. Not a merge blocker, but please file a follow-up: ideally write `fixEntityExecutedDate` to the junction when `executedAt` flips and add an index, so we can restore DB-side date filtering once volume hurts. Otherwise large alt-text opportunities will quietly start paying for a full-history scan per click.

### 🟢 Minor cleanup

`fixEntityCollection.getAllFixesWithSuggestionByCreatedAt` is still stubbed in `beforeEach` (`fixes.test.js`). The method isn't called anymore — drop the stub.

---

**Summary**: right fix, ship after adding the `executedAt=null` test cases. Other items are non-blocking but worth tightening before merge.

audrey-kho:
Adding one regression-risk note on top of @jeddiec's review:

**`new Date(ts).toISOString()` can 500 the whole request on a single malformed timestamp.**

The old code path (`getAllFixesWithSuggestionByCreatedAt`) was a DB-side string equality check — a malformed `fixEntityCreatedDate` would just silently never match. This PR moves the comparison into JS:

```js
const ts = fixEntity.getExecutedAt() || fixEntity.getCreatedAt();
return ts && new Date(ts).toISOString().split('T')[0] === fixCreatedDate;
```

`ts &&` guards null/empty, but `new Date("garbage").toISOString()` throws `RangeError: Invalid time value`. If a single fix in the result set has a malformed `executedAt` or `createdAt` (data migration, partial backfill, manual DB edit), the whole accordion request 500s — strictly worse than the old code's silent-skip behavior.

One-line defensive fix:

```js
const ts = fixEntity.getExecutedAt() || fixEntity.getCreatedAt();
const d = ts ? new Date(ts) : null;
return d && !Number.isNaN(d.getTime()) && d.toISOString().split('T')[0] === fixCreatedDate;
```

Probably never fires in practice (ORM writes timestamps consistently), but cheap insurance against making this endpoint less robust than what it replaces.

jeddiec (approved):
Both review threads addressed cleanly — thanks for the quick turnaround.

**Verified in latest diff:**

- ✅ `fixes.js:122` uses `??` to match the UI accordion key verbatim
- ✅ `fixes.js:123` guards malformed timestamps with `!Number.isNaN(d.getTime())`
- ✅ `fixes.test.js:412-457` covers the `executedAt=null` regression branch (the original bug) in both the *returned* and *filtered out* directions
- ✅ `fixes.test.js:460-477` proves a malformed-timestamp row returns 200 with skip, not 500
- ✅ Orphan `getAllFixesWithSuggestionByCreatedAt` stub removed from `beforeEach`

**Two suggestions, neither blocking:**

1. **Please file a follow-up ticket for the scalability concern.** `getAllFixesWithSuggestionsByOpportunityId` does no DB-side date filtering — every Deployed-tab accordion expansion now pulls the opportunity's full fix history. A ticket to write `fixEntityExecutedDate` to the junction when `executedAt` flips (+ index, restore DB-side filtering) would let us track this before large alt-text opportunities quietly start paying a full-history scan per click. Link it back here when filed.

2. **Cross-link `OneAdobe/experience-success-studio-ui#1970` in the description.** They must ship together (the UI's UTC-key revert is a no-op without this backend filter), and the dependency is currently only documented in this review thread. Worth making it explicit for whoever lands them.

Approving — these are both follow-ups, not merge gates.

PR #2505, opened 2026-05-28
feat: slack alert on skipping a suggestion in PLG

anshulk-public:
## Summary

- Adds a Slack alert whenever a PLG-tier customer transitions a suggestion to `SKIPPED`
  status, fired from both `PATCH /suggestions/:id` and `PATCH /suggestions/status`
- Alert is sent to `SLACK_PLG_SKIP_CHANNEL_ID` using `SLACK_BOT_TOKEN`; both must be
  set or the function exits early without side effects
- Message includes site URL, site ID, opportunity type/ID, suggestion ID, skip reason,
  and skip detail
- An `isNewSkipTransition` flag ensures the alert only fires on the first transition to
  `SKIPPED` — not when skip reason/detail are later updated on an already-skipped suggestion
- Tier detection via `isSitePlgTier()`: resolves the site's ASO enrollment entitlement
  and returns `true` only when `getTier() === 'PLG'`; PAID sites and sites without an
  ASO entitlement are silently skipped
- Alert failures are caught and logged via `context.log.error` — they never propagate
  as HTTP errors
  
  **jira:**https://jira.corp.adobe.com/browse/SITES-44711
  
image

  
  

## Environment variables required

| Variable | Description |
|---|---|
| `SLACK_PLG_SKIP_CHANNEL_ID` | Slack channel ID to post PLG skip alerts to |
| `SLACK_BOT_TOKEN` | Bot token with `chat:write` scope |

## Test plan

- [ ] 8 new unit tests in `PLG skip Slack alert` describe block cover:
  - PLG site → alert sent with correct channel, message, and skip reason
  - PAID site → no alert
  - Missing `SLACK_PLG_SKIP_CHANNEL_ID` → no alert
  - Already-SKIPPED suggestion updated → no alert (idempotent)
  - Slack API throws → error logged, response still 200/207
  - Bulk `patchSuggestionsStatus` with two new SKIPPED → two alerts fired
  - Bulk with already-SKIPPED suggestion → no alert
- [ ] All 411 existing suggestions controller tests pass
- [ ] `npm run lint` passes

## Related Issues

Thanks for contributing!


MysticatBot-Dev (dismissed):
Hey @anshulk-public,

## Strengths

- **Clean separation of concerns** (`src/controllers/suggestions.js:57-65, 67-105`): The alert logic is extracted into two focused helper functions (`isSitePlgTier`, `postPlgSuggestionSkipAlert`) rather than being inlined into the controller flow.
- **Correct transition-only gating** (`src/controllers/suggestions.js:727, 934`): The `isNewSkipTransition` flag ensures alerts fire only on the first transition to SKIPPED, not on re-saves of already-skipped suggestions. This is a common alerting mistake handled correctly upfront.
- **Graceful degradation** (`src/controllers/suggestions.js:71-73`): Early-returns when env vars are missing, so non-PLG environments skip alerting without errors.
- **Error isolation** (`src/controllers/suggestions.js:102-104`): The try/catch ensures Slack failures never break the core suggestion update response.
- **Thorough test coverage** (`test/controllers/suggestions.test.js`): 10 test cases cover both endpoints, PLG vs PAID filtering, missing config, already-skipped idempotency, enrollment errors, alert failure resilience, and skipDetail inclusion.

## Issues

### Important (Should Fix)

**1. `skipDetail` injected unescaped into Slack mrkdwn message** - `src/controllers/suggestions.js:98`

```js
message += `\n• *Skip Detail:* ${skipDetail}`;
```

`skipDetail` is arbitrary user-supplied text. Unlike `skipReason` (which is validated against an allowlist and wrapped in backticks on line 95), `skipDetail` is interpolated raw into a Slack `mrkdwn` block. A PLG customer could inject `@here`/`@channel` mentions to ping an entire channel, craft misleading link text via `<http://evil.com|Looks Fine>`, or break message formatting with `*`, `~`, `_` characters.

**Fix:** Wrap in backticks like the other fields:
```js
message += `\n• *Skip Detail:* \`${skipDetail}\``;
```

---

**2. `isSitePlgTier` silently swallows all errors without logging** - `src/controllers/suggestions.js:62-64`

```js
} catch {
  return false;
}
```

If `getSiteEnrollments()` or `getEntitlement()` throws due to a transient DB failure, the function returns `false` and the alert is silently skipped with no log trail. If the enrollments service is intermittently failing, PLG skip alerts will silently stop arriving and there will be no signal to diagnose why.

**Fix:** Pass `log` (or `context`) and log at warn level:
```js
async function isSitePlgTier(site, log) {
  try {
    // ...
  } catch (err) {
    log.warn(`Failed to determine PLG tier for site ${site.getId()}: ${err.message}`);
    return false;
  }
}
```

---

**3. Alert awaited on HTTP critical path adds latency** - `src/controllers/suggestions.js:767, 977-978`

`postPlgSuggestionSkipAlert` is awaited before returning the response to the caller. This means every PLG skip request is held open waiting on: the tier check (2+ DB calls), plus the Slack HTTP call. On Slack degradation, user-facing latency increases even though the DB write already succeeded.

In the bulk endpoint (`patchSuggestionsStatus`), this compounds: each newly-SKIPPED suggestion sequentially awaits a Slack call AND a redundant `isSitePlgTier` check (the site is the same across the entire batch).

**Fix (minimal):** Fire-and-forget since the alert is non-critical:
```js
postPlgSuggestionSkipAlert(site, opportunity, updatedSuggestion, context)
  .catch((err) => log.error(`PLG skip alert failed: ${err.message}`));
```

For the bulk path specifically, hoist the `isSitePlgTier` check outside the per-suggestion loop since the site does not change between iterations.

---

**4. Missing error context in the catch block** - `src/controllers/suggestions.js:103`

```js
log.error(`Failed to send PLG suggestion skip Slack alert: ${alertError.message}`);
```

The log contains only the error message string with no suggestion ID, site ID, or stack trace. When this fires in production, the operator cannot correlate which suggestion triggered the failure or reproduce the issue.

**Fix:** Include structured context:
```js
log.error('Failed to send PLG suggestion skip Slack alert', {
  error: alertError,
  suggestionId: suggestion.getId?.(),
  siteId: site.getId?.(),
});
```

### Minor (Nice to Have)

**5. Inconsistent formatting between `skipReason` and `skipDetail`** - `src/controllers/suggestions.js:95,98`

`skipReason` is wrapped in backticks (code formatting), `skipDetail` is bare text. This looks like an oversight. Wrapping `skipDetail` in backticks (as suggested in finding 1) also resolves this inconsistency.

## Recommendations

- For the bulk endpoint, consider collecting alert payloads and firing them after the loop with `Promise.allSettled` to avoid serialized Slack calls blocking the response.
- Consider extracting `isSitePlgTier` and `postPlgSuggestionSkipAlert` to a dedicated module (e.g., `src/support/plg-alert.js`). The suggestions controller is already large (2600+ lines) and the alert logic is self-contained.

## Assessment

Ready to merge? No - with fixes

Address findings 1-3 before merge. Finding 1 (mrkdwn injection) is a one-line fix. Finding 2 (silent swallow) and 3 (latency on critical path) are slightly more involved but straightforward. Finding 4 (error context) is a strong recommendation.

## Next Steps

1. Wrap `skipDetail` in backticks to prevent Slack mrkdwn injection (one-line fix).
2. Add warn-level logging to the `isSitePlgTier` catch block.
3. Make the alert fire-and-forget (or at minimum hoist the tier check out of the bulk loop).
4. Add structured context to the error log in the alert catch block.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 52s | <b>Cost:</b> $3.79 | <b>Commit:</b> <code>92e91dfd7f8d65395872e58abc60d3c6b8b98cb5</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Kanishkavijay39 (approved):
As discussed lets add a ternary operator and add one more channel where we see skip suggestion done by customer not on plg site as well.

PR #2506, opened 2026-05-28
feat(serenity): hide Semrush as implementation detail (LLMO-5190)

rainer-friederich:
## Summary

**Phase 1** of the LLMO-5190 three-phase cut-over described in [`docs/specs/2026-05-28-prompts-api-abstraction.md`](docs/specs/2026-05-28-prompts-api-abstraction.md). The Serenity API surface no longer leaks Semrush-specific identifiers (`semrushLocationId`, `semrushId`, base64 `id`), nor does it pretend a value isn't Semrush's when it actually is (`semrushPromptId`, `semrushProjectId` — kept). Adds a DELETE endpoint for markets; removes the zero-caller `/serenity/workspaces` resource.

Builds on **Phase 0** ([adobe/spacecat-shared#1635](https://github.com/adobe/spacecat-shared/pull/1635), merged 2026-05-28). The new accessor names land via `@adobe/spacecat-shared-data-access@3.73.1`, pinned in this PR.

### What changed on the public API

| Surface | Before | After |
|---|---|---|
| Resource | `/serenity/projects`, `/serenity/workspaces/:ws/projects` | `/serenity/markets` (single canonical resource) |
| Field | `semrushLocationId` (number) | `geoTargetId` (number) |
| Field | `language` (string) | `languageCode` (string) |
| Field | `semrushId` (uuid) | `semrushPromptId` (uuid) — honest name |
| Field | `id` (base64 envelope) | _removed_; PATCH/DELETE address by `semrushPromptId` directly, slice in body/path |
| Field | `semrushProjectId` | `semrushProjectId` (kept — the value genuinely IS Semrush's project UUID) |
| Bulk-delete body | `{ semrushIds: [uuid] }` | `{ prompts: [{semrushPromptId, geoTargetId, languageCode}] }` |
| Filters | optional `geoTargetId`+`languageCode` | **required** on `GET /prompts`, `/tags`, `/models` (400 instead of fan-out + upstream 404) |
| New | _none_ | `DELETE /serenity/markets/:geoTargetId/:languageCode` (upstream-first/DB-second, idempotent on 404) |
| Brand guard | upstream returns 404 on non-UUID | controller returns 400 (`isValidUUID` gate before `resolveBrandUuid`) |

### What did NOT change

- **DB layout**: the table `brand_to_semrush_projects` and its columns `semrush_location_id` / `language` stay. The shared `postgrestField` overrides on the schema attributes do the column-name mapping. The table is subject to change for unrelated reasons (Phase 0 PR explains in detail); renaming columns now would be wasted work or would conflict with that pending change.
- **Internal identifiers**: `BrandSemrushProject` entity name and `semrushProjectId` JS accessor remain — they refer to the actual upstream identifier, and honest naming prefers a truthful name over a fake-abstract one.
- **mysticat-projector pipeline**: reads `brand_to_semrush_projects` directly via PostgREST; out of scope and unaffected.

### Why no `@deprecated` aliases / parallel routes / fallback decoder

The routes are gated only by the existing `organization:read` / `organization:write` capabilities in `required-capabilities.js` — there is no per-route feature flag. (The elmo consumer gates the *UI* surfacing client-side via the `FT_LLMO-5005` / `FT_LLMO-4799` LaunchDarkly flags — that is UI gating, not API gating.) The single consumer is project-elmo-ui (PR adobe/project-elmo-ui#1815, Phase 2). Single coordinated cut-over per spec §6.

**Rollback plan** if the new shape regresses after deploy:
1. Revert this PR's merge commit on api-service main; CD re-deploys the prior surface within minutes.
2. Then revert elmo PR #1815 if it merged in the same window.
Order matters: api-service first, then elmo — otherwise elmo on the new shape sees the old api-service and breaks.

### DELETE market semantics

`DELETE /serenity/markets/:geoTargetId/:languageCode`:
1. Resolve the `(brandId, geoTargetId, languageCode)` row.
2. **Upstream first**: `DELETE /v1/workspaces/{ws}/projects/{pid}`. 404 is treated as idempotent success.
3. **DB second**: remove the `BrandSemrushProject` row.
4. Return 204.

Upstream non-404 failure → 502, row stays (operator can retry idempotently). DB write failure after a successful upstream delete → 500, operator re-runs (idempotent again, because upstream now returns 404).

Pre-flight (spec §3.5, 2026-05-28): upstream `OPTIONS` returned `Allow: DELETE, GET, PATCH`; DELETE on a bogus UUID returned 404 not 405 — Option A (no orphan-cleanup sweep) ships.

Requires `organization:write` capability.

## Review fixes (since the initial open)

Hardening that landed across the review rounds and is in the current code:

- **`handleUpdatePrompt`**: text-based fallback removed (was a wrong-prompt data-corruption vector); PATCH now requires all fields and runs explicit DELETE-old + POST-new full-replace.
- **Idempotent-DELETE predicate unified** via `isUpstreamGone` in `src/support/serenity/errors.js` — strict `e instanceof SerenityTransportError && e.status === 404` at all three sites (market DELETE, PATCH delete-then-create, bulk-delete).
- **`handleCreateMarket` orphan handling**: `findBySlice` 409 gate before any upstream call; best-effort `transport.deleteProject` cleanup in the `publishProject` catch (swallows cleanup errors, logs a greppable orphan line, re-throws the original error).
- **`languageCode` acceptance unified** through `normalizeLanguageCode` across every handler (including `POST /markets`).
- **Missing-slice contract**: `handleListPrompts` returns `404 marketNotFound` (aligned with PATCH) via a `.code`-aware `mapError`.
- **Controller boundary**: strict `/^\d+$/` digit gate on the `:geoTargetId` path segment; `messages[]` collapsed into the schema-correct single `message`.
- **Bulk limits**: `BULK_PROMPTS_MAX_ITEMS = 500` cap enforced on bulk create + delete.
- **Tests**: OpenAPI contract test restored and rewritten against the new `Serenity*` schemas; `test/support/serenity/validation.test.js` added; tag-cache invalidation, upstream-body-shape, language-catalog TTL, and bulk mixed-failure tests added; IT 401-contract anchors in `test/it/shared/tests/serenity.js`.
- **Tag cache**: per-mutation invalidation + 60s TTL; 50-page truncation warn.

## Related

- LLMO-5190 — Refactor Serenity API: hide Semrush as an implementation detail
- Epic: LLMO-5005
- Phase 0: [adobe/spacecat-shared#1635](https://github.com/adobe/spacecat-shared/pull/1635) — `@adobe/spacecat-shared-data-access@3.73.1` (merged)
- Phase 2: [adobe/project-elmo-ui#1815](https://github.com/adobe/project-elmo-ui/pull/1815) — UI cut-over (draft; built against this PR's surface, flips after this deploys)

## Test plan

- [x] `npm run lint` — clean
- [x] `npm test` — passing (full suite, exit 0)
- [x] `npm run docs:lint` — clean (errors 0; pre-existing warnings unchanged)
- [x] Banned-name grep across `src` + `docs` (excluding the gitignored generated `docs/index.html`):
  ```bash
  git grep -nE 'semrushLocationId|semrushIds|\bsemrushId\b|encodeLogicalId|decodeLogicalId' \
    -- src docs ':!docs/index.html'
  ```
  empty. `semrushPromptId` and `semrushProjectId` allowed (honest names).
- [x] Phase 0 dependency pinned exactly: `package.json` `dependencies` → `"@adobe/spacecat-shared-data-access": "3.73.1"`; lockfile resolves to the published tarball.
- [x] CI green — `build`, `it-postgres`, `codecov/patch`, `validate-pr-title`, `protect-nyc-config`, `branch-deploy`, and Adobe CLA all passing. (Bundle Build is opt-in and does not run on this PR.)
- [x] Reviewed — MysticatBot multi-agent findings addressed across the review rounds; MysticatBot then reached its cost limit, so re-review moved local and @byteclimber approved on the HEAD commit (`93207f3`) to unblock.


MysticatBot-Dev (changes requested):
## Code Review — PR #2506

**feat(serenity): hide Semrush as implementation detail (LLMO-5190)**

### Strengths

- Clean abstraction: Semrush is fully hidden behind the "Serenity" facade — field renames (`semrushLocationId` → `geoTargetId`, `language` → `languageCode`), resource rename (`/projects` → `/markets`), and the fan-out removal all move in the same direction.
- `handleDeleteMarket` implements upstream-first/DB-second ordering correctly, with thorough test coverage across all five distinct paths (idempotent no-row, happy path, upstream 404 as success, non-404 propagation, half-delete 500).
- DTO leakage assertions in tests explicitly verify that `semrushProjectId`, `semrushLocationId`, and synthetic `id` fields never appear in responses — exactly the contract this PR establishes.
- UUID-only brand guard at the controller boundary prevents non-UUID brand identifiers from reaching handlers.
- OpenAPI specs (`schemas.yaml`, `serenity-api.yaml`, `api.yaml`) updated in sync with implementation; `docs/serenity.md` operator runbook reflects the new surface.
- Routing consistency satisfied: routes in `src/routes/index.js`, controller wired via `getRouteHandlers` in `src/index.js`.
- `rest-transport.test.js` (513 lines) thoroughly covers the new transport methods including `deleteProject`, `createProject`, `listLanguages`, URL sanitization, and timeout handling.

---

### Issues

#### Important

**1. OpenAPI contract test deleted without replacement**
`test/openapi-contract/serenity-api.test.js` (298 lines) is deleted entirely. This was the only automated check that verified endpoint responses conform to declared OpenAPI schemas via AJV validation. Given the heavy schema renaming in this PR (`SemrushPrompt` → `SerenityPrompt`, field renames, required-field changes), a mismatch between `docs/openapi/schemas.yaml` and actual handler responses will go undetected until a client hits it.

The PR's own spec (`docs/specs/2026-05-28-prompts-api-abstraction.md`) lists "OpenAPI contract test — green" as a Phase 1 validation gate, and the CLAUDE.md rule states "Specification Sync: Keep OpenAPI specs and implementation in sync."

**How to fix:** Rewrite the contract test against the new operationIds and schema names. The old structure (fixture table + `operationsForTag` + AJV validation) is reusable; only fixtures and handler imports need updating.

**2. No integration tests for new/modified endpoints**
CLAUDE.md rule: "New or modified endpoints must include integration tests in `test/it/`." This PR adds `DELETE /serenity/markets/:geoTargetId/:languageCode` and substantially modifies 8 other endpoints. No `test/it/` file is touched, and no pre-existing serenity integration test exists.

**How to fix:** Add a shared test factory in `test/it/shared/tests/` (e.g., `serenity-markets.js`) covering at minimum: list markets, create market (409 on dup), delete market (204 + idempotent retry), and the required-filter 400 on prompts/tags/models.

**3. `handleUpdatePrompt` text-based fallback can silently delete the wrong prompt**
When the paginated search doesn't find a prompt by ID, the handler falls back to matching by `name` (text). If two prompts have the same text in different tags, the fallback matches the first one found — the handler then deletes it and recreates with new properties. This is a data-corruption vector: the caller intended to update prompt A but the handler deleted prompt B.

**How to fix:** Remove the text-based fallback entirely, or require an exact single match (reject with 409/conflict if multiple prompts share the same text).

**4. Performance regression in `handleUpdatePrompt` — paginating all prompts for ID lookup**
The handler calls `listPromptsByTags` with no tag filter and paginates through all prompts to find one by ID. For brands with hundreds of prompts, this is O(n) API calls to upstream for a single-item lookup. The old approach decoded the logical ID directly.

**How to fix:** Pass the prompt ID to an upstream endpoint that supports direct lookup, or at minimum add a `page_size=100` with early termination when the item is found (current code already does this but without tag filtering, so the first page may not contain the target).

---

### Minor

**5. Stale comment in `docs/openapi/serenity-api.yaml`** — `# enumerate them. See test/openapi-contract/serenity-api.test.js.` references the file this PR deletes.

**6. `handleCreateMarket` returns `status: 'pending'` in the 201 response** — but the handler calls `publishProject` synchronously before returning. If publish succeeds, the status is already `live` by the time the response arrives; `pending` is misleading.

**7. Concurrent DELETE race condition** — Two simultaneous `DELETE /markets` for the same slice can both pass the `findBySlice` check and both attempt upstream deletion. The upstream 404-as-success pattern handles this gracefully, but both callers will also call `row.remove()`, and the second will fail on a stale row. Low probability but worth a comment.

---

### Recommendations

- Add a test for `handleUpdatePrompt` where the paginated search finds no item matching the `:semrushPromptId` and text fallback also fails — assert 404 with `error: 'promptNotFound'`.
- Consider a lightweight "wiring" test in the controller suite that verifies each method calls `authorize()` — even a shallow assertion that `resolveBrandUuidStub` was called for each handler catches future regressions.

---

### Assessment

**With fixes** — The abstraction design is solid and the unit test coverage for new handler code is strong. However, the deleted OpenAPI contract test removes an important drift-detection layer, the CLAUDE.md-mandated integration tests are absent, and the text-based fallback in `handleUpdatePrompt` introduces a data-integrity risk.

| Agent | Verdict |
|-------|---------|
| Staff Engineer | With fixes |
| Architect | With fixes |
| Security | Ready |
| QA | With fixes |
| Conventions | With fixes |

MysticatBot-Dev:
Mysticat review failed: Claude CLI crashed (exit 1): stderr= stdout={"type":"result","subtype":"error_max_budget_usd","duration_ms":123165,"duration_api_ms":825191,"is_error":true,"num_turns":5,"stop_reason":"tool_use","session_id":"8630b56e-ad7b-45fc-b5b8-571e20724915","total_cost_usd":10.082367249999997,"usage":{"input_tokens":25976,"cache_creation_input_tokens":5

MysticatBot-Dev (dismissed):
Hey @rainer-friederich,

This is a re-review following 9 new commits since the prior review. The text-fallback removal, PATCH simplification (requiring all fields), and documented orphan gap are solid improvements that address the most critical prior findings.

## Strengths

- Previously flagged issues now addressed: text-based fallback removed (data-corruption vector gone), PATCH slow path dropped in favor of explicit full-replace semantics, `status: 'pending'` misleading response resolved, concurrent DELETE race documented in code.
- Clean provider abstraction: Semrush routing details (workspace IDs, internal project IDs) are fully hidden behind the Serenity facade. The `rest-transport.js` file is now the single integration seam for a future provider swap.
- DELETE market implements upstream-first/DB-second ordering correctly with thorough coverage of all five distinct paths (idempotent no-row, happy path, upstream 404 as success, non-404 propagation, half-delete 500).
- Required filters on GET endpoints eliminate the expensive fan-out + merged-pagination pattern. One slice = one upstream call = real pagination. This is the right simplification.
- UUID-only brand guard at the controller boundary prevents name-based brand lookup from silently resolving to wrong rows on rename.
- Orphan handling in `handleCreateMarket` is well-documented with explicit rationale for deferring auto-rollback. The log line is greppable for operator recovery.
- Tag cache with per-mutation invalidation + 60s TTL is the right trade-off for same-container freshness and bounded cross-container staleness.

## Issues

### Important (Should Fix)

**1. Spec vs. implementation drift on Market DTO shape**
`docs/specs/2026-05-28-prompts-api-abstraction.md` section 3.3 defines the Market DTO with `status` and `name` fields. The implementation (`handleListMarkets` in `src/support/serenity/handlers/markets.js`) returns only `{ brandId, geoTargetId, languageCode, createdAt, updatedAt }`. The test explicitly asserts their absence. The project-elmo-ui team could build against the spec and find these fields missing at runtime.

**How to fix:** Update section 3.3 of the spec to mark `status` and `name` as deferred/future, or note them as "not shipped in Phase 1". The spec included in this PR should match what this PR ships.

**2. OpenAPI contract test deleted without replacement (prior finding, still unresolved)**
`test/openapi-contract/serenity-api.test.js` (298 lines) is still deleted with no replacement. The spec's own validation gate (section 7, Phase 1) lists "OpenAPI contract test - green" as a block-merge criterion. The CLAUDE.md rule states "Specification Sync: Keep OpenAPI specs and implementation in sync." Given the heavy schema renaming in this PR, a mismatch between `docs/openapi/schemas.yaml` and actual handler responses will go undetected.

**How to fix:** Rewrite the contract test against the new operationIds and schema names. The old structure (fixture table + AJV validation) is reusable; only fixtures and handler imports need updating.

**3. No integration tests for new/modified endpoints (prior finding, still unresolved)**
CLAUDE.md rule: "New or modified endpoints must include integration tests in `test/it/`." This PR adds `DELETE /serenity/markets/:geoTargetId/:languageCode` and substantially modifies 8 other endpoints. No `test/it/` file is touched.

**How to fix:** Add shared test logic in `test/it/shared/tests/` covering at minimum: list markets, create market (409 on dup), delete market (204 + idempotent retry), and the required-filter 400 on prompts/tags/models.

**4. Stale OpenAPI comment references deleted test file (prior finding, still unresolved)**
`docs/openapi/serenity-api.yaml` line 10 still contains `# enumerate them. See test/openapi-contract/serenity-api.test.js.` - this file no longer exists.

**How to fix:** Remove or update the comment to reference the replacement test location.

**5. `markets.js` coverage gap at 74% (156 lines missing)**
Codecov reports `src/support/serenity/handlers/markets.js` at 74% patch coverage with 156 lines uncovered. For a new file implementing a DELETE endpoint with data-integrity implications, this is below the bar for production readiness.

**How to fix:** Add unit tests for the uncovered paths in `handleCreateMarket` (publish failure orphan logging, DB race catch, upstream returning no project id) and `handleDeleteMarket` (half-delete state).

### Minor (Nice to Have)

**6. Tag pagination ceiling silently truncates for large projects**
`handleListTags` caps at `page <= 50` (10k prompts). Projects exceeding this limit will have an incomplete tag set with no warning to the caller.

**How to fix:** Add a log.warn when `page === 50` and items are still being returned, so operators know the tag set is truncated.

**7. Spec section 3.2 says `brandDomain`/`brandNames` are "read server-side from the brands row"**
The implementation still requires them in the request body (`validateCreateBody` enforces both as mandatory). The spec document is aspirational on this point.

**How to fix:** Either update the spec to match the implementation (fields required in body), or note this as a future simplification.

## Recommendations

- Consider adding a TODO comment in `handleListTags` noting the pagination walk is an O(N) approximation; a dedicated upstream tags endpoint would eliminate it.
- The tag cache TTL was reduced from 5 min to 60s alongside the per-mutation invalidation. This is the right call for freshness but increases upstream load on cold containers. Monitor after deploy.

## Assessment

Ready to merge? With fixes
Reasoning: The abstraction design is solid, the prior critical findings (text fallback, PATCH semantics) are resolved, and the orphan handling is well-documented. The remaining gaps are the unresolved OpenAPI contract test deletion (drift detection), missing integration tests (CLAUDE.md mandate), spec-to-implementation drift on the Market DTO, and below-bar coverage on the new markets handler. Items 1-5 should be addressed before merge; items 6-7 are optional.

## Next Steps

1. Address the spec drift (item 1) and stale comment (item 4) - quick doc fixes.
2. Restore the OpenAPI contract test (item 2) - reuse old structure with new schema names.
3. Add integration tests (item 3) - at minimum for DELETE market and required-filter 400s.
4. Improve markets.js coverage (item 5) - cover the orphan and half-delete paths.

MysticatBot-Dev:
Mysticat review failed: Claude CLI crashed (exit 1): stderr= stdout={"type":"result","subtype":"error_max_budget_usd","duration_ms":7,"duration_api_ms":940073,"is_error":true,"num_turns":1,"stop_reason":null,"session_id":"ecab042c-977d-4293-84e7-29dc279b4ef9","total_cost_usd":10.106775499999996,"usage":{"input_tokens":0,"cache_creation_input_tokens":0,"cache_read_in

rainer-friederich:
Thanks for the re-review @MysticatBot. Walking through each finding with what's now on `feat/semrush-proxy-2`:

**1 — Spec drift on Market DTO** — fixed in `d4dbcdb5`. `docs/specs/2026-05-28-prompts-api-abstraction.md` §3.3 now matches the shipped DTO (`{brandId, geoTargetId, languageCode, createdAt, updatedAt}`) and adds an explicit "Deferred fields" rationale section explaining why `status` and `name` were dropped post-design. §9 follow-up note also updated.

**2 — OpenAPI contract test deleted** — **invalid finding**: `test/openapi-contract/serenity-api.test.js` is present in the branch (304 lines, restored in commit `1cf090f1` after the prior review's prompt). The contract test was deleted briefly during the initial refactor and put back the same day. `git log -- test/openapi-contract/serenity-api.test.js | head` on `34de1ddc` confirms the file is in tree. No action needed; please re-verify on the latest commit.

**3 — No integration tests** — addressed within the IT-harness constraint in `00c37562`. Added `test/it/shared/tests/serenity.js` + `test/it/postgres/serenity.test.js` covering 7 contracts: 400 on non-UUID spaceCatId, 400 on non-UUID brandId, 401 on JWT-authenticated GET /markets / /prompts / /tags / /models / DELETE /markets.

The deeper "list/create/delete market" and "required-filter 400 on prompts/tags/models" IT coverage you asked for is blocked by a structural gap: the serenity controller enforces IMS-only authentication (`requireImsBearer`), and the shared IT harness (`test/it/shared/auth.js`) mints JWT tokens. JWT-authenticated requests deterministically 401 before any handler ever runs. The full IT-replacement path requires either minting IMS-shaped tokens in the harness or stubbing `requireImsBearer` at test time — both ship test-only code into production paths or duplicate auth harnessing. I filed this as follow-up rather than worked around it. The handler-side coverage of those scenarios is at 100% in the unit suite (see #5 below) plus AJV-validated in the contract test you flagged as deleted.

**4 — Stale OpenAPI comment line 10** — **invalid finding**: the comment "`See test/openapi-contract/serenity-api.test.js.`" references a file that exists at exactly that path (see #2 above). The comment is accurate. No action.

**5 — markets.js coverage gap at 74%** — fixed in `d4dbcdb5`. Coverage on the changed files is now:

| File | Lines | Branches | Functions |
|---|---|---|---|
| `src/support/serenity/handlers/markets.js` | 100% | 100% | 100% |
| `src/support/serenity/handlers/prompts.js` | 100% | 100% | 100% |
| `src/support/serenity/validation.js` | 100% | 100% | 100% |

New tests cover `handleCreateMarket` orphan paths (publish-failure log + throw, DB-race log + 409, missing project id 502, default name generation, empty name 400), `handleDeleteMarket` half-delete state (already present), `handleListTags` cache hit + truncation warn + object-form tag dedup, `handleListModels` empty-page / multi-page / missing items field, and complete `handleCreatePrompts` / `handleBulkDeletePrompts` failure-bucket coverage. Two `/* c8 ignore */` markers on truly defensive short-circuits (LRU eviction loop on a 512-slot cache; `items[0] || {}` adversarial-shape guard) with reasons documented inline.

**6 — Tag pagination ceiling silently truncates** — fixed in `d4dbcdb5`. `handleListTags` now logs at `warn` when the 50-page (10,000-prompt) ceiling is hit AND the last page was full (more pages exist that we never read). Log carries `projectId, pagesWalked, pageSize, tagsFound` for operator triage. Plus a TODO comment pointing at the upstream dedicated-tags endpoint as the future O(1) replacement.

**7 — Spec §3.2 brandDomain/brandNames drift** — fixed in `d4dbcdb5`. Spec now matches implementation: both fields stay in the request body, with the rationale captured inline (client has them loaded; server-side read would be a second DB lookup with no behaviour change). Bonus: §3.2 also now documents `POST /serenity/prompts` body shape which was missing.

---

**Recommendations:**
- "TODO on handleListTags O(N)" — done as part of 6.
- "Monitor tag cache TTL on cold containers after deploy" — accepted; will keep an eye on it.

PR is ready for another pass when you have a minute.

MysticatBot-Dev:
Mysticat review failed: Claude CLI crashed (exit 1): stderr= stdout={"type":"result","subtype":"error_max_budget_usd","duration_ms":6,"duration_api_ms":904921,"is_error":true,"num_turns":1,"stop_reason":null,"session_id":"35820eed-4409-4fe5-9d4f-be0456ca334e","total_cost_usd":10.041010899999996,"usage":{"input_tokens":0,"cache_creation_input_tokens":0,"cache_read_in

rainer-friederich (commented):
Hey @rainer-friederich,

Multi-agent review on top of the work since the MysticatBot passes. The architectural framing (honest naming, single-slice simplification, narrow integration seam at `rest-transport.js`) is the right design, and the regression-guard work in `f21083c2` / `d4dbcdb5` materially closed the prior reviews. The findings below are net-new — they are not duplicates of the MysticatBot items you already addressed.

## Strengths

- The "Serenity" facade is honest in the right places: `geoTargetId` and `languageCode` hide a Semrush-specific encoding that wasn't ours, while `semrushPromptId` and `semrushProjectId` admit they ARE upstream UUIDs. Fake-abstract names would have lied without making provider swap any easier.
- The fan-out + merged-pagination + partial-failure envelope on `GET /prompts` collapses to one upstream call per slice. Type-enforced at the OpenAPI level via required filters. A real simplification, not a rename.
- `404-as-success` is the right idempotency invariant for a thin proxy without distributed locks, applied to market DELETE, prompt bulk-delete, and PATCH delete-then-create.
- `handleCreateMarket` runs `findBySlice` before any upstream call, so the common 409 case never produces an orphan. The two orphan-producing edges each have a regression test asserting the exact grep-able log shape an operator would need.
- The new `handleUpdatePrompt` (DELETE-then-CREATE, abort on any DELETE error other than 404, no fallback CREATE) is materially safer than the prior `findSemrushPromptByText` walk. The "no warn-and-create" regression test in `prompts.test.js` locks the contract.
- `handleListTags` truncation warn at the 50-page ceiling (with `pagesWalked` / `pageSize` / `tagsFound` in the log context) turns "the dropdown is missing a tag" into a one-search diagnosable symptom rather than a mystery.
- `src/index.js` deletion of the `:workspaceId` UUID guard is consistent: no route in `src/routes/index.js` or `src/routes/required-capabilities.js` declares `:workspaceId` anymore. Confirmed dead, not just untested.
- Security posture is strong: `rest-transport.js` `baseUrl()` parses with `new URL()`, enforces https, strips path/userinfo; `enc()` `encodeURIComponent`s every interpolated path segment; UUID gate at the controller boundary applies uniformly to every handler; `mapError` for `SerenityTransportError` returns hardcoded `{error: 'serenityUpstreamError'}` with the full upstream body logged server-side only; the IMS bearer is required by `buildHeaders` (throws 401 if absent); the lockfile pin on `@adobe/spacecat-shared-data-access@3.73.1` matches the published tarball integrity.
- OpenAPI contract test restored (commit `1cf090f1`) and rewritten against the new `Serenity*` schemas. AJV validates each operationId's response against the documented schema, with the 204 DELETE path correctly skipping body validation. Drift-detection guarantee intact.
- IT suite at `test/it/shared/tests/serenity.js` correctly anchors the IMS-vs-JWT 401 contract for the six new/renamed routes.

## Issues

### Important (Should Fix)

1. **Inconsistent 404-recognition predicate across the three idempotent-DELETE sites.** `src/support/serenity/handlers/markets.js:2646` uses the strict form `e instanceof SerenityTransportError && e.status === 404`. `src/support/serenity/handlers/prompts.js:4096` (PATCH delete-then-create) and `prompts.js:4249` (bulk-delete per-target) use the loose form `e?.status === 404`. Same architectural invariant, three slightly different guards. The loose form will treat any throwable whose `.status` happens to equal 404 — a future custom error, a downstream library that decorates with `status: 404`, an ad-hoc test stub — as upstream-idempotent-success. That silently swallows real failures and turns "upstream gone" from a guarded predicate into "anything that looks like 404". Fix: extract an `isUpstreamGone(e)` helper into `src/support/serenity/validation.js` (or a new `errors.js`) and use it in all three sites. The strict variant is the safe one. One line per site.

2. **`languageCode` acceptance contract diverges between `POST /markets` and every other endpoint.** `src/support/serenity/handlers/markets.js` `validateCreateBody` runs `LANGUAGE_TAG_REGEX.test(body.languageCode)` directly on the raw input, then lowercases post-validation. Every other handler that takes `languageCode` (`handleDeleteMarket`, `handleListPrompts`, `handleListTags`, `handleListModels`, `handleCreatePrompts`, `handleUpdatePrompt`, `handleBulkDeletePrompts`) routes through `normalizeLanguageCode`, which lowercases BEFORE the regex test. User-visible effect: `POST /markets {"languageCode":"EN"}` returns 400, but `DELETE /markets/2840/EN`, `GET /markets?languageCode=EN`, `GET /tags?languageCode=EN` silently accept and lowercase. Same input field, two different acceptance rules on the same surface. Fix: replace the inline regex in `validateCreateBody` with `normalizeLanguageCode(body.languageCode) === null` and drop the redundant post-validation `.toLowerCase()` (the normalizer already returns lowercase).

3. **`handleCreateMarket` retry contract is documented but not implemented.** The spec (§3.5 narrative) and `docs/serenity.md` tell operators they can "safely retry with the same body" because the 409 gate catches the case where a previous attempt succeeded both upstream calls but failed the DB write. The code does not match that contract: every retry generates a fresh `defaultMarketName` via `crypto.randomBytes(3)`, and the upstream `createProject` body has no idempotency key. A retry after a `publishProject` failure creates a SECOND upstream project, not a recovery of the first. The 409 gate only fires if a DB row exists; it never sees orphan upstream projects. The `b4348219` JSDoc deferral ("the storage shape is in flux, no auto-rollback") is a reasonable argument for not building a saga, but the current path silently violates the spec's retry contract. Fix: in the `publishProject` catch block, call `transport.deleteProject(semrushWorkspaceId, semrushProjectId)` best-effort, swallow its errors, log both outcomes, then re-throw. The transport already exposes `deleteProject` as of this PR. The retry contract then holds. The DB-row-create-failure branch (line ~2570) is harder to reason about because a concurrent create could be in flight; keeping that as "log only" with the 409 is defensible.

4. **"Missing slice" produces three different observable behaviors across the surface.** `handleListPrompts` returns `200 {items: [], total: 0}` when `findBySlice` is null. `handleCreatePrompts` lands the same condition in `skipped[]` with `reason: "No market for slice (geoTargetId, languageCode)"`. `handleUpdatePrompt` throws `ErrorWithStatusCode(404, 'marketNotFound')`. Same precondition, three different contracts. A renamed-then-stale market between page load and a list call silently renders an empty prompts list when the right answer is "this slice no longer exists." Fix: unify on the PATCH precedent (`404 marketNotFound`). The slice not existing is structurally different from the slice being empty; clients deserve to be able to distinguish.

5. **Feature-flag rollout claim is not corroborated by the diff.** PR body and spec §6 cite `FT_LLMO-5005` and `FT_LLMO-4799` as the gates that make "single coordinated cut-over, no parallel routes, no @deprecated aliases" safe. Grep across the diff for `FT_LLMO`, `featureFlag`, or any flag-gating wrapper returns zero matches; `src/routes/required-capabilities.js` shows ordinary `organization:read` / `organization:write` and nothing else. If the routes are live to anyone with `organization:write`, rolling api-service forward and elmo backward leaves elmo broken on the old shape — the rollback story the spec promises does not exist. Fix: either (a) wrap the new routes (or all of `/serenity/*`) in the existing feature-flag mechanism if the team relies on flag gating for this rollout, or (b) update spec §6 and the PR description to drop the "feature-flagged surface" claim and document the actual rollback plan ("revert this PR's merge before reverting elmo").

6. **Tag-cache invalidation contract has no end-to-end test.** `src/support/serenity/handlers/prompts.js` POST `/prompts`, PATCH, and bulk-delete each call `invalidateTagCacheForProject(semrushWorkspaceId, projectId)`. That is the contract that lets you tighten the TTL from 5 minutes to 60 seconds without "writes invisible until cache expires" becoming a user-visible bug. No test exercises the full path (mutation → invalidation → subsequent `handleListTags` re-walks upstream). A regression that drops the invalidate call from any of the three mutation handlers would not trip either test suite. Fix: add one test per mutating handler that (a) populates the tag cache via `handleListTags` with stubbed upstream returning set A, (b) runs the mutation with a successful upstream stub, (c) reconfigures the upstream stub to return set B and asserts `handleListTags` returns set B (proving cache was invalidated). Three short tests sharing fixtures.

7. **Upstream `createProject` body shape is not asserted in any test.** `handleCreateMarket` constructs the upstream body with `type: 'ai'`, `brand_name_display`, `brand_names`, `domain`, `country_code` (lowercased ISO-2), `location_id`, `location_name`, `language_id`. The deleted `projects.test.js` had a "type 'ai' forwarded" assertion; the new `markets.test.js` only asserts `body.name` against the default-name regex. Everything else upstream-side is untested. A regression that flips `country_code` to uppercase, drops `language_id`, or sends `type: 'ai_overview'` would not be caught — and the upstream contract is the most load-bearing thing this handler does. Fix: in the existing happy-path test, add a block of assertions on `transport.createProject.firstCall.args[1]` for the full upstream body shape (`type: 'ai'`, `country_code: 'us'`, `location_id: 2840`, `location_name: 'United States'`, `language_id: ...`, `brand_name_display`, `brand_names`, `domain`). No new test file needed.

8. **Language-catalog cache TTL/refresh behavior is unverified.** `resolveLanguageId` caches the upstream language catalog for 1h. The deleted `projects.test.js` had explicit tests for "caches across calls within TTL" and "refreshes after TTL expires"; the new `markets.test.js` calls `clearLanguageCache()` in `beforeEach` and never asserts a single `listLanguages.callCount(1)` across two consecutive `handleCreateMarket` calls within one test, never advances a fake clock past 1h. A regression that resets `expiresAt` against a stale `now` would silently hit the upstream catalog on every request, adding latency under load with no test failure. Fix: one test asserting `transport.listLanguages.callCount(1)` across two back-to-back `handleCreateMarket` calls (no `clearLanguageCache` between); one test with `sinon.useFakeTimers()` advancing 1h+1ms and asserting `callCount(2)`.

9. **Bulk-delete partial-failure mixes are not tested.** `handleBulkDeletePrompts` buckets targets by project; per-bucket upstream errors propagate into each target's `failed` entry. Current tests cover the homogeneous cases (all 200, all 404, all 503, all unmapped). The bucketing logic — the only actually load-bearing thing in this handler — is not exercised by mixed-mix scenarios: (slice A 200 + slice B 503), (slice A 200 + slice B 404 idempotent), (mix of valid + invalid item shapes in same body). Fix: one mixed-mix `it()` block per important pairing, sharing fixtures with the existing homogeneous tests. Three short tests.

10. **`deleteMarket` controller boundary on malformed path params is undertested and is likely the source of the codecov-missing 3 lines.** `src/controllers/serenity.js` does `parseInt(pGeo, 10)` then `Number.isFinite(geoTargetId) ? geoTargetId : null` and `pLang ? String(pLang).toLowerCase() : null`. The controller test only covers the happy path (`'2840'` + `'EN'`). The `geoTargetId: 'abc'` / `geoTargetId: ''` / `languageCode: ''` paths through the `parseInt → null` short-circuit at the controller boundary are not asserted. Codecov reports 97.00% on `src/controllers/serenity.js` with 3 lines missing — confirm via the codecov report that these are the 3 lines, then add coverage instead of accepting the patch-coverage failure.

### Minor (Nice to Have)

1. **`parseInt` is too forgiving on `geoTargetId` path segments.** `parseInt('2840abc', 10) === 2840`, so `/markets/2840abc/en` is silently routed to the legitimate slice (2840, en); same for `'2840.5'`. OpenAPI declares `geoTargetId: { type: integer, minimum: 1 }` — at the contract level these inputs should 400. Low blast radius (still resolves to the right slice or 204-idempotent), so polish rather than correctness. Fix: replace `parseInt(pGeo, 10)` with `/^\d+$/.test(pGeo) ? Number(pGeo) : null`.

2. **`validateCreateBody` returns `messages: [...]` (plural), but `SerenityErrorResponse` schema declares `message` (singular).** `POST /markets` with bad input returns a body that does not match the documented OpenAPI error schema. Either join with `; ` into a single `message`, or extend the schema to declare an optional `messages` array alongside `message`.

3. **Module-level cross-import: `prompts.js` reaches into `markets.js` for `invalidateTagCacheForProject`.** The tag cache logically belongs to "tags" (a markets concern) but is now manipulated by three prompts handlers. The coupling works, but `markets.js` exports an internal cache-control function used by a sibling. When handlers split (per the spec hint about a future `handlers/tags.js`), this naturally lifts into a `tags-cache.js` module. Worth noting now so it does not solidify.

4. **Bulk-delete and bulk-create accept arrays of unbounded size.** OpenAPI declares `maxItems: 500` on both `SerenityBulkDeletePromptsRequest.prompts` and `SerenityCreatePromptsRequest.prompts`, but this codebase does not run OpenAPI request validation as middleware. Within API Gateway's request cap, an authenticated caller can submit tens of thousands of items. Pre-existing pattern (prior `semrushIds` array also had no cap), so not a regression — but the spec rewrite here is the clean place to enforce it. Add `if (targets.length > 500) throw new ErrorWithStatusCode('prompts array exceeds maxItems=500', 400);` after the empty-check in both handlers.

5. **`src/support/serenity/validation.js` has no dedicated unit-test file.** Coverage is transitive through `markets.test.js` and `prompts.test.js`, which exercise the callers, not the validator contract. `normalizeLanguageCode` is not directly tested against surrounding whitespace, non-string types, `null`, `undefined`; `normalizeGeoTargetId` is not directly tested against `NaN`, negative numbers, zero (only via the handler boundary), float, Infinity, string-of-digits. A small `test/support/serenity/validation.test.js` with ~8 it-blocks would lock the validator contract independently of caller behavior.

6. **No regression test for the malformed-languageCode 400 path on prompts handlers.** The author's commit note for `b8281e81` says prompts handlers "now regex-validate languageCode, so malformed values produce 400 instead of 'no market for slice' 404s downstream." The markets DELETE test covers this for the delete handler (`'1z'`). The same behavior change in `handleListPrompts` / `handleListTags` / `handleListModels` has no dedicated regression test — each handler's "missing filters" test passes `languageCode: undefined`, which exits via `hasText` short-circuit. Add one test per handler that passes a syntactically malformed languageCode and asserts `ErrorWithStatusCode(400)`.

7. **`failed[]` publish-failure entries carry `text: ''`.** A `failed` entry with empty `text` is structurally ambiguous: which prompt failed? Schema accepts it (no `required` on the inner objects), but a UI iterating `failed[]` to highlight per-row errors cannot map this row back to anything. Reasonable for MVP, but a `kind: 'publish'` discriminator would make the response easier to consume without breaking back-compat.

## Recommendations

- File the orphan-reconciliation follow-up ticket now while the design rationale is fresh. The JSDoc on `handleCreateMarket` says "via a one-off script reading the same log stream" — operators will not write that script until something breaks. A scheduled Lambda that diffs `listWorkspaceProjects` against `BrandSemrushProject.allByBrandId` per org is roughly 50 lines and converts an operator runbook into self-healing.
- Add a 4-line file-header comment in `rest-transport.js` ("this is the ONLY file in `src/` that knows the upstream is Semrush; add new Semrush-aware code here or extend this file"). Today the integration seam is real but tribal; a header comment cements the architectural invariant. Optionally pair it with a `npm test` lint that greps `src/**/*.js` for `semrush.com|adobe-hackathon|/v1/workspaces|/v2/workspaces` and asserts the only matches are in `rest-transport.js`. Cheap, catches the failure mode the rename was designed to prevent.
- The `e?.status === 404` vs `e instanceof SerenityTransportError && e.status === 404` split (Important 1) is one symptom of a more general pattern — there is no single typed-error-mapping helper. Each handler open-codes its own 404 recognition. The `isUpstreamGone(e)` fix from Important 1 is the seed of a small `errors.js` module that gives a single place to grow the matrix later (e.g., 410 if Semrush ever adds it).
- File the LLMO ticket for the upstream dedicated-tags endpoint that the `TODO` at the truncation-warn site references. Today the TODO is anonymous; without a ticket it gets lost.
- Consider a one-time Stryker mutation-test run against `markets.js` and `prompts.js`. Both are at "100% coverage" but the gaps in Important 6/7/8/9 show that line/branch coverage is misleading on this surface. Cheaper than re-reading manually next time someone touches it.

- `mysticat-projector` reads `brand_to_semrush_projects` columns directly via PostgREST. The clean public surface in this PR masks a hidden cross-repo coupling on the column names. If the table is ever renamed, both api-service (`postgrestField` overrides in spacecat-shared) and mysticat-projector must change atomically. Belongs in a spacecat-shared schema comment naming projector as a direct reader.
- Phase 0 (`adobe/spacecat-shared#1635`) shipped an accessor rename without a `BREAKING CHANGE:` footer because api-service is the only known consumer. If any other team has pinned to the unrenamed accessors, they will bump and break silently. Cross-repo concern, not fixable here.
- The IT-harness structural blocker the author documented (IMS-only auth vs JWT-minting harness) should be tracked as a real Jira ticket against the harness rather than an indefinite carry-forward in the IT comment. The deeper end-to-end IT scenarios depend on it.

## Assessment

The architecture and security posture are strong. The Important items above are addressable inside this PR: 1 (404 predicate unification) and 2 (languageCode normalizer split) are one-line consistency fixes; 3 (best-effort upstream cleanup on publish-failure) is a small addition to a catch block using a transport method that already exists; 4 (unify missing-slice contract) is picking the PATCH precedent in two more handlers; 5 (feature-flag claim vs reality) is either adding the flag or updating the spec; 6/7/8/9/10 are five targeted test additions that close real regression-guard gaps without inventing new infrastructure. Once they land, this is a strong piece of work.

## Next Steps

1. Address the 5 consistency/contract Important items (1, 2, 3, 4, 5) — these are surface-visible to clients and the smallest in code-size.
2. Address the 5 test-coverage Important items (6, 7, 8, 9, 10) — these are regression guards for behaviors the PR explicitly tightened.
3. Minor items are optional polish; pick whichever land naturally during the Important fixes.

rainer-friederich (commented):
Hey @rainer-friederich,

Re-review on commit `2f8aaa2c` (the only commit since the prior multi-agent review). The 10 Important and 7 Minor findings from the prior round are addressed or deliberately deferred, and the fix commit is materially correct - `isUpstreamGone` is strict-by-type, the best-effort cleanup in `handleCreateMarket` preserves the publish-error propagation, the `.code` plumbing through `mapError` is single-callsite with a clean fallback, and the test additions for the prior Important 6-10 are state-observation tests rather than interaction-verification dressed up as E2E. Three small polish items and one spec-sync note below - none block merge.

## Strengths

- Previously flagged issues now addressed:
  - 404 predicate unified via the new `src/support/serenity/errors.js::isUpstreamGone` - strict `e instanceof SerenityTransportError && e.status === 404`. All three call sites (`markets.js:455`, `prompts.js:391`, `prompts.js:507`) use the helper; no `e?.status === 404` loose form remains anywhere on the serenity surface.
  - `languageCode` acceptance contract collapsed onto a single rule. `validateCreateBody` routes raw input through `normalizeLanguageCode` so `POST /markets`, `DELETE /markets/:geo/:lang`, and every `GET` filter agree.
  - `handleCreateMarket` retry contract honored. Best-effort `transport.deleteProject` in the `publishProject` catch block uses an inner try/catch that swallows cleanup errors, emits a separate orphan log on cleanup failure, and re-throws the original publish error unchanged. No error shadowing, no double-log.
  - `handleListPrompts` missing-slice contract unified with PATCH via `ErrorWithStatusCode(404).code = 'marketNotFound'` + the new `.code`-aware branch in the controller's `mapError`.
  - PR body updated to drop the `FT_LLMO-5005`/`FT_LLMO-4799` feature-flag claim and document the real rollback plan (revert api-service first, then elmo).
  - `parseInt` path-segment parser replaced with strict `/^\d+$/` digit-gate at `src/controllers/serenity.js:363`.
  - `messages: [...]` (plural) joined into the schema-correct `message: errors.join('; ')` string.
  - `BULK_PROMPTS_MAX_ITEMS = 500` cap enforced at the request boundary in both bulk handlers (before `allByBrandId` and before Map build).
  - `test/support/serenity/validation.test.js` added - 15 it-blocks locking `normalizeLanguageCode` + `normalizeGeoTargetId` independent of caller behavior.
  - Malformed-`languageCode` 400 tests added for `handleListPrompts`, `handleListTags`, `handleListModels`.

- Test additions for the prior Important 6-10 are real behavioral pins, not coverage padding:
  - Tag-cache E2E (`test/support/serenity/handlers/prompts.test.js`, the "tag cache invalidation" describe block) populates the cache via the real `handleListTags`, mutates via the real handler, and asserts the next `handleListTags` returns set B. State observation, not interaction verification.
  - `createProject` upstream-body shape pin at `test/support/serenity/handlers/markets.test.js:201-214` covers all the regression scenarios the prior round called out (`country_code: 'us'`, `type: 'ai'`, `language_id` present, `brand_names` deep-equal).
  - Language-catalog cache TTL tests correctly do NOT clear the cache between the two `handleCreateMarket` calls in the cache-hit test, and use `sinon.useFakeTimers().tick(60 * 60 * 1000 + 1)` with `clock.restore()` in `finally` for the TTL refresh test.
  - Bulk-delete mixed-mix tests assert response shape (`result.deleted`, `result.failed[]` entries with `semrushPromptId`/`geoTargetId`/`languageCode`/`status`), not just stub call counts.
  - Controller-boundary tests at `test/controllers/serenity.test.js:284-325` assert the handler stub args (`args[4]`/`args[5]` is `null`), proving the controller short-circuited - a regression resurrecting `parseInt` would not pass these.

- Stub-migration discipline: the two pre-existing tests at `prompts.test.js:619` and `:822` that exercised the idempotent-404 path were migrated from `Object.assign(new Error, { status: 404 })` to `new SerenityTransportError(404, ...)`. Non-404 sites correctly retain the bare-Error form. Grep confirms no leftover `status: 404` ducks.

## Issues

### Minor (Nice to Have)

1. **`.code` field on `ErrorWithStatusCode` has no enumeration or schema** - `src/support/serenity/handlers/prompts.js:113`. Today exactly one site sets it (hardcoded literal `'marketNotFound'`), and `mapError` guards with `hasText`. The risk is the next site: if the next author picks `'market_not_found'` or `'MarketNotFound'`, the response token vocabulary diverges silently because nothing centralizes the names. Why it matters: API error tokens are a contract clients pattern-match on; drift here breaks them silently. How to fix (not blocking, but the right moment is the second `.code` token, not the tenth): introduce a frozen map next to `isUpstreamGone` in `src/support/serenity/errors.js`, e.g. `export const ERROR_CODES = Object.freeze({ MARKET_NOT_FOUND: 'marketNotFound' });` and have handlers assign `err.code = ERROR_CODES.MARKET_NOT_FOUND`. Cheaper than a subclass hierarchy and keeps the existing constructor signature untouched.

2. **Loose range assertion in the mixed-mix valid+invalid bulk-delete test** - `test/support/serenity/handlers/prompts.test.js:1036`. The 4-item input is deterministically 1 valid + 3 invalid (missing `semrushPromptId`, unparseable `geoTargetId`, malformed `languageCode`), so the handler produces exactly `failed.length === 3`. The current assertion `expect(result.failed.length).to.be.at.least(3)` is wider than the contract and would silently accept a regression that double-pushes the same target into `failed`. The sibling tests at lines 988 and 1004 use the tighter `lengthOf(...)` form. How to fix: replace `.to.be.at.least(3)` with `.to.have.lengthOf(3)`.

3. **`LANGUAGE_TAG_REGEX` direct test partially mirrors `normalizeLanguageCode`** - `test/support/serenity/validation.test.js:71-80`. The regex IS the only validator inside `normalizeLanguageCode` (verified at `src/support/serenity/validation.js`), so a drift between the two is either deliberate (and surfaces in the 7 surrounding `normalizeLanguageCode` tests) or structurally impossible. The block leans slightly per-symbol-mirror. The genuine behavior the regex test catches that the normalizer tests do not is case-sensitivity (raw regex rejects `'EN'`, normalizer accepts it via lowercase pre-pass). How to fix (optional): drop this block, or rephrase the comment to call out the case-sensitivity asymmetry as the actual contract being locked.

## Recommendations

- **One-line spec sync on the `handleListPrompts` 404 contract.** `docs/specs/2026-05-28-prompts-api-abstraction.md` section 3.2 covers request shapes and the GET filter section covers the required-filter 400, but neither mentions what `GET /serenity/prompts` returns when filters are valid but the `(brandId, geoTargetId, languageCode)` slice has no row. The commit changes that behavior from "200 with empty page" to "404 `marketNotFound`" - a strictly better contract (alignment with PATCH semantics, distinguishable from "exists, empty"), but the spec does not mention it. Add one line under the GET filter table: "Missing slice (filters valid, no `BrandSemrushProject`) returns 404 `marketNotFound`. Single-slice handlers (GET, PATCH) emit 404; bulk handlers carry per-item `skipped` entries." Lands the spec in sync with the shipped code instead of forcing the next reader to derive the contract from grep.

- Consider one additional controller-boundary test for `geoTargetId: '0'`. Strict-digit regex `/^\d+$/` accepts `'0'` and forwards `Number('0') === 0` to the handler, where `normalizeGeoTargetId(0)` returns `null` per the new normalizer contract. This is a different code path than `'2840abc'` (regex-reject) and `''` (regex-reject). Out of scope for the current commit if you would rather ship; worth a follow-up.

- Prior Minor (cross-module import `prompts.js -> markets.js::invalidateTagCacheForProject`) and prior Minor (`failed[]` publish-failure entries with empty `text`) remain deferred by stated design. The new tag-cache invalidation tests significantly lower the risk of the future tags-handler refactor - they lock the contract behaviorally, so the import direction can flip without test rewrites.

## Assessment

Ready to merge? **Yes.**

Reasoning: The fix commit addresses prior Important 1-10 and Minor 1-2+4-6 with the right structural shape (strict typed-error gate, single normalizer, best-effort cleanup that preserves the original error, state-observation tests). The three Minor items above are polish - the `.code` vocabulary is a follow-up the second site will surface, the loose range assertion is a one-character tighten, and the regex-direct test is a 10-line trim. The spec-sync recommendation is the one item I would land in this PR rather than later, because the next reader of the spec will otherwise believe the empty-200 behavior still holds.

## Next Steps

1. Tighten the `at.least(3)` to `lengthOf(3)` in the bulk-delete mixed-mix test (one-line change).
2. Add the one-line spec update on the `handleListPrompts` 404 contract.
3. The `.code` enumeration and the regex-mirror trim can land with the next change to either file.

byteclimber (approved):
So you are unblocked since the reviews needs to happen locally for cost limit of MysticatBot

PR #2507, opened 2026-05-28
feat(suggestions): prefer x-promise-token header over IMS token creation in auto-fix flow

ravverma:
## Summary

- Adds `x-promise-token` request header as the primary source for the promise token in the `autofixSuggestions` flow
- Removes the intermediate cookie (`promiseToken`) check — flow is now: header → IMS token creation (fallback)
- Adds `x-promise-token` to `localCORSWrapper` allowed headers in `src/index.js` to match the existing OPTIONS preflight configuration

## Test plan

- [x] Updated `'uses promiseToken cookie when present instead of IMS'` → `'uses x-promise-token header when present instead of IMS'`
- [x] Updated `'falls back to IMS when promiseToken cookie is absent'` → `'falls back to IMS when x-promise-token header is absent'`
- [x] Removed now-redundant `'falls back to IMS when promiseToken cookie is not present among other cookies'` test
- [x] All updated tests pass locally

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,10 +16,7 @@ admin-idp-p*.json
 .vscode/settings.json
 
 # Cursor
-.cursor/*
-!.cursor/rules/
-!.cursor/commands/
-.cursor/commands/speckit.*
+.cursor/
 
 # V1 to V2 migration scripts and test data (local only)
 scripts/
--- a/src/controllers/suggestions.js
+++ b/src/controllers/suggestions.js
@@ -40,7 +40,6 @@ import { FixDto } from '../dto/fix.js';
 import { GeoExperimentDto } from '../dto/geo-experiment.js';
 import {
   sendAutofixMessage,
-  getCookieValue,
   getIMSPromiseToken,
   ErrorWithStatusCode,
   getHostName,
@@ -1256,9 +1255,9 @@ function SuggestionsController(ctx, sqs, env) {
     let promiseTokenResponse;
     const skipPromiseToken = isAssessAction && precheckOnly === true;
     if (!skipPromiseToken) {
-      const cookieToken = getCookieValue(context, 'promiseToken');
-      if (hasText(cookieToken)) {
-        promiseTokenResponse = { promise_token: cookieToken };
+      const headerToken = context.pathInfo?.headers?.['x-promise-token'];
+      if (hasText(headerToken)) {
+        promiseTokenResponse = { promise_token: headerToken };
       } else {
         try {
           promiseTokenResponse = await getIMSPromiseToken(context);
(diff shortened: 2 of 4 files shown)


MysticatBot-Dev (approved):
Hey @ravverma,

## Strengths

- **Well-scoped change with correct CORS alignment** - three files, one concern, no unrelated cleanup. The CORS `Access-Control-Allow-Headers` update in `src/index.js:151` is easy to forget and good to see included alongside the controller change.
- **Net security improvement** (`src/controllers/suggestions.js:1257-1259`) - moving from cookie-sourced token to an explicit request header eliminates the automatic-attachment behavior browsers apply to cookies. Custom headers require explicit JavaScript to set, which removes the CSRF surface for this flow.
- **Clean fallback preservation** - the IMS token creation fallback is retained, so existing callers that haven't adopted the header degrade gracefully rather than failing.
- **Appropriate optional chaining at the request boundary** (`src/controllers/suggestions.js:1257`) - `context.pathInfo?.headers?.['x-promise-token']` is correct defensive access for request pipeline data, with `hasText()` plus IMS fallback forming the resolution chain.
- **Justified test removal** (`test/controllers/suggestions.test.js:5807-5835`) - the deleted test verified cookie-parsing semantics (other cookies present but not `promiseToken`). With direct header lookup this scenario is structurally impossible - a header is present or it isn't. The remaining two tests cover both branches.

## Issues

### Minor (Nice to Have)

1. `test/controllers/suggestions.test.js` - **Empty-string header value untested.** The "falls back to IMS when x-promise-token header is absent" test omits the header entirely. A case where the header is present but empty (`'x-promise-token': ''`) would confirm `hasText` rejects it and the IMS fallback fires. Low risk since `hasText` was already the guard in the old code, but it would document the boundary explicitly.

## Recommendations

- Confirm that the production Fastly VCL configuration allows `x-promise-token` through in its CORS `Access-Control-Allow-Headers` response. The diff only touches `localCORSWrapper` (Lambda-level CORS). If Fastly strips or omits this header in preflight responses, the browser will refuse to send it on the actual request.

## Assessment

Ready to merge? Yes

The change is architecturally sound, well-tested, and a net security improvement. It simplifies the token resolution from a three-tier lookup (cookie parsed from header, then IMS) to a clean two-tier path (explicit header, then IMS). No new failure modes introduced, no breaking changes for existing callers.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 43s | <b>Cost:</b> $5.22 | <b>Commit:</b> <code>28b646fab728fde75a11d1a1d2302cae718ece39</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2511, opened 2026-05-28
feat: add GET /sites/:siteId/config/scraper

ymadobe:
## Summary
The existing scraper-config endpoints have a write but no read. To inspect the persisted \`scraperConfig\` today, callers have to hit \`GET /sites/:siteId\` and dig into \`.config.scraperConfig\` — which dumps the entire site config (\`imports\`, \`llmo\`, \`slack\`, \`tokowakaConfig\`, etc.) and risks echoing unrelated secrets to a caller that only needs scraper config.

This PR adds \`GET /sites/:siteId/config/scraper\`:
- Returns \`scraperConfig: {}\` when nothing is persisted (no separate "missing" case)
- \`site:read\` capability 

## Test plan
- [x] 5 new unit tests (returns persisted, returns empty, 400 malformed UUID, 404 missing site, 403 no access)
- [x] 5 new IT tests including a writer-then-reader check that locks GET ≡ PATCH envelope
- [x] 295 sites-controller tests pass locally
- [x] ESLint clean
- [x] \`npm run docs:lint\` validates

LLMO-4975


MysticatBot-Dev (commented):
Hey @ymadobe,

## Strengths

- **Symmetric GET/PATCH response shape** (docs/openapi/schemas.yaml:867-872, src/controllers/sites.js:1186-1189): Reusing `ScraperConfigResponse` for both endpoints means a caller that writes then reads sees an identical envelope. This eliminates a class of client-side deserialization bugs and is the right long-term shape for per-subresource config endpoints.
- **Secret-boundary respect** (docs/openapi/site-api.yaml:159-161): The design deliberately avoids echoing the full site config, preventing accidental leakage of `tokowakaConfig.apiKey` and similar secrets to callers that only need scraper config.
- **Writer-then-reader integration test** (test/it/shared/tests/sites.js:556-570): The PATCH-then-GET test that asserts `getRes.body` deep-equals `patchRes.body` is the single most valuable test in this PR - it locks the symmetry contract.
- **Thorough defensive null handling** (src/controllers/sites.js:1190-1197): Differentiating `null` (explicitly cleared) from `undefined` (never written) via `log.warn` gives operators a forensics breadcrumb without changing the API contract.
- **Proper access control and capability gating** (src/routes/required-capabilities.js:341): The `site:read` capability is least-privilege, correctly separated from the `site:write` required by PATCH.
- **Comprehensive test coverage**: 8 unit tests covering all code paths (happy path, empty config, null config, 400/403/404/5xx) plus 5 integration tests including the critical writer-then-reader check.

## Issues

### Minor (Nice to Have)

1. **src/controllers/sites.js:1200-1203 - Catch-log-throw produces duplicate error logging.** The handler logs with `log.error(...)` then re-throws. The framework's top-level error handler will log the same error again, producing two log entries per failure during incidents. However, this matches the established convention in this controller (the sibling `updateScraperConfig` uses the same pattern), so fixing this one instance without addressing the controller-wide pattern would create inconsistency. Worth addressing codebase-wide if the team decides to change the convention.

2. **src/controllers/sites.js:1148-1161 - Comment density is high relative to the code's self-evidence.** The 14-line JSDoc block largely restates what the function name, route path, and return statement already communicate. The inline comments at lines 1183-1197 explaining the null-vs-undefined distinction are valuable, but the preamble could be trimmed to just the non-obvious bits (the security motivation for not echoing full config).

## Recommendations

- **Consider a generic config-subresource GET helper if the pattern repeats.** This PR adds `GET .../config/scraper` alongside `PATCH .../config/scraper` and `PATCH .../config/cdn-logs`. If a `GET .../config/cdn-logs` follows, a shared helper that takes a config accessor function and produces the narrow `{ siteId, <configKey> }` response would reduce per-endpoint boilerplate. Not a blocker for this PR.

## Assessment

Ready to merge? Yes

This is a clean, well-tested, well-scoped addition that follows established patterns precisely. The narrow response shape, proper auth/authz, and symmetric GET/PATCH contract are all done right. No security concerns, no architectural issues, and comprehensive test coverage.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 8s | <b>Cost:</b> $3.14 | <b>Commit:</b> <code>5bb5bca1d87d90ff5c6b99cac4cac4426d64e373</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anagarwa (approved):

PR #2512, opened 2026-05-28
fix(fixes): replace N+1 getSuggestions loop with batch query

mimchome:
[SITES-45307](https://jira.corp.adobe.com/browse/SITES-45307)

## Summary

- Replace per-fix `getSuggestions()` calls with single `getAllFixesWithSuggestionsByOpportunityId` batch query (2-3 DB calls instead of N+1)
- Fixes without mapping table entries now return with empty suggestions instead of being silently dropped (58% of fixes on dev were affected)

## Test plan

- [ ] Unit tests pass (new empty-array test + updated stubs)
- [ ] Integration tests pass (added FIX_2 junction seed record)
- [ ] codecov/patch green
- [ ] After deploy: sunstar.com opp `78c9837b` returns 3 fixes (was 2)


src/controllers/fixes.js
```diff
@@ -149,27 +149,25 @@ export class FixesController {
       return ok(fixes);
     }
 
-    fixEntities = await this.#FixEntity.allByOpportunityId(opportunityId);
+    const fixEntitiesWithSuggestions = await this.#FixEntity
+      .getAllFixesWithSuggestionsByOpportunityId(opportunityId);
+
+    if (fixEntitiesWithSuggestions.length === 0) {
```
MysticatBot-Dev, line 174: **Important:** This early return bypasses the `checkOwnership` call below. If an opportunity belongs to a different site but has zero fixes, the caller now gets `200 []` instead of the 404 that `checkOwnership` would produce.

Move the ownership check above this guard, or validate the opportunity-site relationship independently of fix results.

MysticatBot-Dev (dismissed):
Hey @mimchome,

## Strengths

- **N+1 elimination follows existing pattern** (`src/controllers/fixes.js:150-151`): Aligns with `getAllFixesWithSuggestionByCreatedAt` already used on the `fixCreatedDate` path, reducing the number of distinct query shapes the team needs to reason about.
- **Corrects silent data loss**: Fixes without junction table entries (58% on dev per PR description) now return with empty suggestions instead of being silently dropped. The right contract for a list endpoint.
- **Integration tests assert on actual IDs** (`test/it/shared/tests/fixes.js:64-71`): Verifying specific suggestion IDs per fix, not just array lengths. This is the kind of test that would have caught the original silent-drop bug.
- **Tightly scoped change**: One controller method, its tests, and seed data. No unrelated cleanup mixed in.
- **New unit test for early-return path** (`test/controllers/fixes.test.js:194-203`): Clean AAA structure, asserts on both status and body.
- **Conventions compliance**: PR title follows conventional commit format, unit and integration tests are present, DTOs are used correctly.

## Issues

### Important (Should Fix)

1. **`src/controllers/fixes.js:153-155` - Early return on empty results bypasses ownership check**

   ```javascript
   if (fixEntitiesWithSuggestions.length === 0) {
     return ok([]);
   }
   ```

   The old code called `checkOwnership(fixEntities[0], opportunityId, siteId, this.#Opportunity)` regardless of whether fixes existed. With the new early return, if an opportunity has zero fixes but belongs to a *different* site, the caller gets `200 []` instead of the 404 that `checkOwnership` would have produced.

   **Why it matters:** A caller with a valid `siteId` they own and a guessed `opportunityId` from a different site now gets 200 instead of 404 when that opportunity has no fixes. Returning 200 for a resource the caller does not own is incorrect regardless of what data it carries. The unit test at `test/controllers/fixes.test.js:176` previously tested this scenario (empty result + wrong site = 404) but was changed to return a non-empty array, masking the behavioral change.

   **How to fix:** Validate the opportunity-to-site relationship before the early return. One approach - move the ownership check above the empty guard:

   ```javascript
   const fixEntitiesWithSuggestions = await this.#FixEntity
     .getAllFixesWithSuggestionsByOpportunityId(opportunityId);

   res = await checkOwnership(
     fixEntitiesWithSuggestions[0]?.fixEntity,
     opportunityId, siteId, this.#Opportunity
   );
   if (res) {
     return res;
   }

   if (fixEntitiesWithSuggestions.length === 0) {
     return ok([]);
   }
   ```

   If `checkOwnership` requires a non-null first argument to trace the relationship, validate the opportunity's `siteId` directly before the early return instead.

## Recommendations

- The integration test verifies the happy path (2 fixes, each with 1 suggestion). Consider adding a seed fixture where a fix has no junction entry to validate the stated behavior change at the integration level ("fixes without mapping table entries now return with empty suggestions").

## Assessment

Ready to merge? **With fixes**
Reasoning: The N+1 elimination and the silent-drop bug fix are both correct and well-tested. The ownership bypass on the empty-results path is a behavioral regression in access control that should be addressed before merge - it is a reordering fix, not a redesign.

## Next Steps

1. Address the ownership check ordering so the empty-results path still validates opportunity-site ownership.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 42s | <b>Cost:</b> $5.81 | <b>Commit:</b> <code>3efd6010aa2796dce436c3e614475380e70be5ca</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

mimchome:
@MysticatBot-Dev  issues are addressed

MysticatBot (approved):
Hey @mimchome,

## Strengths

- **Previously flagged ownership bypass resolved** (`src/controllers/fixes.js:122-131`, `src/controllers/fixes.js:156-165`): Both code paths now validate opportunity-site ownership before any early return on empty results. The optional chaining with `checkOwnership` fallback to `Opportunity.findById` ensures the authorization check fires even when no fix entities exist. The unit test at `test/controllers/fixes.test.js:205-215` directly regresses the prior vulnerability.
- **N+1 elimination converges both branches onto a single query shape** (`src/controllers/fixes.js:153-155`): The `fixCreatedDate` branch and the no-filter branch now both use `getAllFixesWithSuggestionsByOpportunityId`, reducing the number of distinct query patterns the data layer must optimize and monitor.
- **Silent data loss corrected**: Fixes without junction table entries now return with empty suggestions instead of being silently dropped. The PR description quantifies the impact (58% of fixes on dev), and the integration test at `test/it/shared/tests/fixes.js:75-77` explicitly asserts on FIX_3 with no junction entry.
- **Comprehensive ownership test matrix** (`test/controllers/fixes.test.js:176-215`): Four distinct ownership/empty scenarios - wrong opportunity, wrong site with fixes present, empty results with correct ownership (200), and empty results with wrong site (404). Clean AAA structure throughout.
- **Integration tests assert on specific IDs** (`test/it/shared/tests/fixes.js:64-80`): Verifying exact suggestion IDs per fix rather than just array lengths catches the class of bug this PR fixes.
- **Prior recommendation addressed**: Seed fixture for junction-less fix (FIX_3) validates the behavioral change at the integration level.
- **Tightly scoped change**: One controller method, its tests, and seed data. No unrelated cleanup mixed in.

## Issues

### Minor (Nice to Have)

1. **`src/controllers/fixes.js:121-131` and `src/controllers/fixes.js:153-166` - Duplicated ownership validation pattern across two branches**

   Both the `fixCreatedDate` branch and the no-filter branch contain an identical ownership validation block (call `checkOwnership` with `[0]?.fixEntity`, return early if truthy). Today it is two copies; if a third query shape is added, a third copy will appear.

   **How to fix:** Extract a shared helper (e.g., `validateOwnershipOrFail(results, opportunityId, siteId)`) that both branches call. Follow-up concern, not a merge blocker.

## Recommendations

- The controller is converging on a consistent fetch-validate-transform pattern. If additional filter dimensions are planned (by status, by type), consider whether the filtering belongs in the query layer (a parameterized `getAllFixesWithSuggestions(opportunityId, filters)`) rather than as additional in-memory branches. The current approach is fine for two branches but will fragment as the filter surface grows.
- Consider adding unit test coverage for the `fixCreatedDate` path where `allFixesWithSuggestions` is non-empty but the date filter narrows to zero results - the ownership check still fires correctly, but this scenario is currently only covered by the integration suite.

## Assessment

Ready to merge? **Yes**
Reasoning: The N+1 elimination is correct, the ownership check ordering regression from the first review is resolved, and both unit and integration tests cover the key behavioral change. CI is green. No security, correctness, or architectural concerns remain.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 35s | <b>Cost:</b> $2.70 | <b>Commit:</b> <code>f1560a46c1e3d411c79428a17665abfeb4757eda</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2513, opened 2026-05-29
feat(onboarding): Serenity cohort gate + markets[] DTO — DB feature-flag gate & /projects→/markets (LLMO-5201/5202/5493)

luis6156:
## Update (since opening)

This branch now carries the full "cohort gate + markets API" scope. Two changes were folded in after the original T1/T2 write-up (retained below for history):

- **Cohort gate migrated env → DB feature flag (LLMO-5493).** `isSerenityOnboardingEnabled` is now an **async, per-org read** of the `serenity` row in the `feature_flags` table (product LLMO), seedable at runtime via `PUT /organizations/:id/feature-flags/LLMO/serenity`. **Transitional dual-read:** the DB flag is authoritative; when no row exists for the org the legacy `SERENITY_SITE_ALLOWLIST` env arm is still honored (logged as deprecated), and an explicit DB `false` wins over the env arm. Remove the env fallback once every cohort org has a DB row. The legacy env check is renamed `isInSerenityAllowlist`; `readSerenityFlagOverride` + `LLMO_SERENITY_FLAG` are added (mirroring the brandalf flag). The caller now `await`s the gate and passes `context`.
- **Dropped the redundant `GET /serenity/projects` endpoint; `market` folded into `/serenity/markets`.** The markets list now carries the derived ISO `market` field, so the client has a single list endpoint. OpenAPI (`api`/`schemas`/`serenity-api`), the route-list test, and the contract test are updated accordingly.
- **`markets[].language` renamed to `markets[].languageCode`** to align the DTO with the upstream Semrush schema.

> ⚠️ The "Env-driven — no DB flag required for the prototype phase" line in the original T1 write-up below is **superseded** by the DB-flag migration above.

---

## Summary

Implements T1 (LLMO-5201) and T2 (LLMO-5202) from the Semrush onboarding orchestration epic (LLMO-5007).

- **T1** — Add `SERENITY_SITE_ALLOWLIST` cohort gate: orgs in the allowlist enter the Semrush provisioning path (M5–M8); all others fall through to the existing v1/v2 flow unchanged. Env-driven — no DB flag required for the prototype phase.
- **T2** — Accept optional `markets[{market, language}]` in `POST /llmo/onboard`. Validates market (`^[A-Z]{2}$`) and language (BCP-47 subtag) per entry. Backward-compat: synthesizes `[{market: region, language: 'en'}]` when `region` is supplied without `markets`. When both are present, `markets` wins. `cadence` preserved unchanged. `markets` is threaded into `performLlmoOnboarding` for the T3b fan-out.

## What changed

**`src/support/llmo-onboarding-mode.js`**
- Export `SERENITY_SITE_ALLOWLIST = 'SERENITY_SITE_ALLOWLIST'` constant (consistent with `LLMO_BRANDALF_FLAG` pattern)
- Add `isSerenityOnboardingEnabled(organizationId, imsOrgId, env)` — checks the comma-separated allowlist against both SpaceCat org ID and IMS org ID; whitespace-tolerant

**`src/controllers/llmo/llmo.js`**
- Destructure `markets` from request body
- Validate each entry: `market` must be `^[A-Z]{2}$`, `language` must be BCP-47 (`^[a-z]{2,3}(-[a-z]{2,4})?$`)
- Backward-compat fallback: synthesize from `region` when `markets` absent; `markets` wins when both present
- Pass `resolvedMarkets` into `performLlmoOnboarding`

**`src/controllers/llmo/llmo-onboarding.js`**
- Import `isSerenityOnboardingEnabled`
- Destructure `markets` from params
- After `upsertBrand` in the v2 path, enter the serenity gate block when the org is allowlisted — placeholder for T3a/b/c (LLMO-5203–5205)

**`docs/openapi/llmo-api.yaml`**
- `markets[]` request field with per-item validation patterns
- Updated `region` deprecation note
- `requested`/`succeeded`/`failed` fields on 200 response
- New 207 response schema

## Branch note for T3

T3a (LLMO-5203), T3b (LLMO-5204), T3c (LLMO-5205) should branch from this branch (`feat/LLMO-5201-5202-cohort-gate-markets`), not from `main`. The `markets` variable and the serenity gate block are already in place — T3 only needs to fill in the M5/M7/M8 logic inside the gate.

## Test plan

- [ ] `isSerenityOnboardingEnabled`: DB-flag true/false, env fallback when no row, explicit DB false beats env, whitespace trimming, undefined env
- [ ] `performLlmoOnboarding` gate: serenity log fires when enabled; skipped when not
- [ ] `onboardCustomer` markets[]: valid markets forwarded, region synthesized, markets wins over region, invalid market code → 400, invalid language → 400, empty array → 400, neither supplied → no markets key
- [ ] `/serenity/markets` returns the derived `market` field; `/serenity/projects` is gone (openapi-contract + route-list tests)
- [ ] Non-cohort orgs: all pre-existing tests pass unchanged


MysticatBot (dismissed):
Hey @luis6156,

## Strengths

- **Well-structured error isolation in `performSerenityFanOut`** (src/controllers/llmo/llmo-onboarding.js:1254-1290): The `failAll` helper, sequential pre-condition checks, and per-tuple error isolation are clean. Each failure path logs with context and returns a structured result rather than throwing.
- **Trust-but-verify M8 read-back** (src/controllers/llmo/llmo-onboarding.js:401-457): Treating the DB as the single source of truth rather than trusting upstream API responses catches the 201-returned-but-row-missing case. Correct architecture for a system where silent write failures are possible.
- **Idempotent re-invocation by design** (src/controllers/llmo/llmo-onboarding.js:349-354): 409-as-success plus M8 read-back means callers can safely retry the same `markets[]` without creating duplicates.
- **Input validation at the boundary** (src/controllers/llmo/llmo.js:968-1013): Thorough per-entry validation with clear error messages, duplicate collapse, and a cap at 50 entries.
- **Proper 207 Multi-Status semantics** (src/controllers/llmo/llmo.js:759-761): Using proper HTTP semantics for partial failures rather than jamming errors into a 200.
- **SerenityTransportError sanitization** (src/controllers/llmo/llmo-onboarding.js:1295-1302): Messages containing full upstream URLs with workspace IDs are deliberately sanitized to the generic token `semrushUpstreamError`.
- **Preflight error separation** (src/controllers/llmo/llmo-onboarding.js:528-530): The `preflight` flag cleanly distinguishes "nothing was created yet" from "partial state exists and needs cleanup," keeping the existing cleanup logic untouched.
- **Comprehensive test coverage**: The test suite covers every early-exit, per-tuple partial failures, idempotency (409), security boundary (URL sanitization), and structured metric assertions including property completeness checks.
- **Conventions compliance**: No CLAUDE.md violations. New code follows the JS-module import pattern for static data, respects the controller factory pattern, and does not expose raw database models.

## Issues

### Important (Should Fix)

**1. Non-SerenityTransportError exceptions leak raw error messages to API callers**

src/controllers/llmo/llmo-onboarding.js:1344

When a non-`SerenityTransportError` is thrown (e.g., a TypeError, DB driver error, or JSON parse failure inside `handleCreateMarket`), its raw `.message` is placed into the `failed[]` array and returned to the API caller in the 200/207 response body. The `SerenityTransportError` path is correctly sanitized to a generic token, but this else-branch does not receive the same treatment.

**Why it matters**: Internal error messages can contain file paths, internal hostnames, stack fragments, or other implementation details. This contradicts the explicit sanitization already applied to `SerenityTransportError`.

**Fix**:
```js
error: thrown instanceof SerenityTransportError ? 'semrushUpstreamError' : 'internalError',
```
Log `thrown.message` server-side for debugging (it is already visible via the structured metric), but do not echo it to the caller.

**2. Sequential fan-out has no timeout guard - Lambda time bomb**

src/controllers/llmo/llmo-onboarding.js:317-319

The code acknowledges this in a TODO: "worst case 50 tuples x ~30s = 1500s, well past Lambda timeout." Even with adobe.com's ~19 real tuples at worst-case latency, this exceeds typical Lambda timeouts. When the Lambda dies mid-loop, the caller gets no structured response - just an abrupt timeout with no visibility into which tuples succeeded.

**Why it matters**: The system will observe partial provisioning with no way to distinguish a timeout boundary from a genuine failure.

**Fix**: Add a deadline check at the top of each loop iteration. The Lambda context provides `getRemainingTimeInMillis()` - if remaining time drops below a safety margin (e.g. 10s for the M8 read-back), break the loop, mark remaining tuples as failed with a `'timeout'` error token, and proceed to M8. This gives the caller a structured 207 instead of an abrupt timeout.

**3. User input reflected in error messages without length bounds**

src/controllers/llmo/llmo.js:990-993

The `entry.market` and `entry.language` values are interpolated directly into the 400 error message. While the regex will reject them, a caller can send a multi-MB string as `entry.market` which gets reflected into the response body and potentially into server-side logs. This enables log flooding and response body amplification.

**Fix**: Truncate the reflected value before interpolation:
```js
const displayMarket = String(entry.market).slice(0, 20);
return badRequest(`Invalid market "${displayMarket}". Must be an ISO 3166-1 alpha-2 uppercase code (e.g. US, DE)`);
```

**4. `extractImsBearer` can return an empty string as a valid token**

src/controllers/llmo/llmo-onboarding.js:227

If `authorization` header is exactly `"Bearer "` (trailing space, no token), `hasText` passes, `startsWith('Bearer ')` passes, and `.substring(7)` returns `""`. This empty string is then passed to `createSerenityTransport` and forwarded upstream as an empty bearer.

**Fix**: Validate the extracted token is non-empty:
```js
const token = header.substring('Bearer '.length);
return hasText(token) ? token : null;
```

### Minor (Nice to Have)

**5. OpenAPI spec missing `maxItems` on `markets` array**

docs/openapi/llmo-api.yaml (markets definition)

The controller enforces `MAX_MARKETS = 50`, but the OpenAPI spec does not declare `maxItems: 50`. Consumers reading the spec will not know the cap until they hit the 400.

**Fix**: Add `maxItems: 50` to the `markets` array definition.

**6. Internal org ID exposed in 404 error message**

src/controllers/llmo/llmo-onboarding.js:1540-1544

The 404 response includes the internal SpaceCat organization UUID and an explicit internal API path (`PATCH /organizations/:id`). While the caller is authenticated, this leaks internal system topology.

**Fix**: Return a generic message to the API caller and log the detailed operator guidance server-side.

**7. `reconcileSerenityProjects` silently picks last-write-wins on duplicate DB rows**

src/controllers/llmo/llmo-onboarding.js:421-424

The `rowBySlice` Map uses `set()` in a `forEach` - if `allByBrandId` returns multiple rows with the same `(geoTargetId, languageCode)` key, the last row silently wins. Unlikely in practice (the proxy deduplicates via `findBySlice`), but the function is described as "authoritative" yet does not assert row uniqueness.

**Fix**: Log a warning when a duplicate key is encountered.

## Recommendations

- The `performSerenityFanOut` function is 120+ lines. Consider extracting the pre-condition validation (brand, IMS bearer, transport) into a helper that returns either a ready-to-use context object or the `failAll` result, keeping the fan-out loop focused.
- Consider adding a test for the `brandName = " "` case (whitespace-only). The controller's `!brandName` check passes for `" "` (truthy), then `brandName.trim()` produces `""`, and `generateBrandId("")` may produce unexpected results.
- Document the env-var allowlist as a transitional mechanism with a sunset point (e.g., "once cohort exceeds 10 orgs, move to DB flag"). Env-var changes require deployment, cannot be audited in real-time, and have no rollback semantics.

## Assessment

Ready to merge? With fixes
Reasoning: The architecture is well-reasoned (trust-but-verify, idempotent retry, clean preflight separation, proper 207 semantics) and test coverage is comprehensive. The four Important findings are information disclosure (findings 1, 3), an operational risk (finding 2), and a functional edge case (finding 4). Finding 1 (error message leak) is the most immediately actionable since it contradicts the explicit sanitization already applied elsewhere in this same PR.

## Next Steps

1. Address finding 1 (error message leak) - quick one-line fix that aligns with the existing sanitization pattern.
2. Address finding 4 (empty bearer) - another quick fix.
3. Address finding 3 (input truncation in error messages) - straightforward.
4. Finding 2 (timeout guard) may be deferred to the T4/GA phase if the team prefers, given the existing TODO, but should be tracked.
5. Minor items are optional.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 9s | <b>Cost:</b> $5.92 | <b>Commit:</b> <code>f372b320f5b6182c51a664b51fc13acf7453642f</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

aenascut (commented):
## Code Review — Semrush onboarding orchestration (LLMO-5201/5202)

All 7 MysticatBot findings confirmed still present. Additional findings below.

---

### 🔴 High — Fan-out timeout guard (TODO still open)

`src/controllers/llmo/llmo-onboarding.js` ~L1346

The TODO is acknowledged but unfixed. At adobe.com's ~19 real market tuples × p99 ~30s upstream = ~570s — already 63% of the 15-min Lambda ceiling with no headroom for retries or upstream slowness. A 50-tuple customer hits ~25 min guaranteed timeout.

Suggested fix before GA: track elapsed time between iterations and fail remaining tuples with `{ status: 504, error: 'fanoutBudgetExceeded' }` once a budget (e.g. 600s) is exceeded. Alternatively switch to `Promise.allSettled` for parallel fan-out now that the AIO Proxy handles its own idempotency.

---

### 🟠 Medium — Exception `.message` leaks to API caller

`src/controllers/llmo/llmo-onboarding.js` ~L1388

```js
error: thrown instanceof SerenityTransportError ? 'semrushUpstreamError' : thrown.message,
```

The comment says internal errors are safe to echo, but `handleCreateMarket` calls external libs (node fetch, URL parser, JSON parse) that can throw with messages containing env vars, file paths, or upstream response bodies.

```js
error: thrown instanceof SerenityTransportError
  ? 'semrushUpstreamError'
  : (log.error(`Serenity fan-out unexpected: ${thrown.message}`), 'internalError'),
```

---

### 🟠 Medium — User input reflected in 400 errors without length bounds

`src/controllers/llmo/llmo.js` ~L690, L693

`entry.market` / `entry.language` come from the raw request body — no type check, no length cap. A caller can send a 10 MB string or an object (`String(obj)` → `[object Object]`), inflating the response and polluting error-tracker grouping.

```js
const safeMarket = typeof entry.market === 'string' ? entry.market.slice(0, 16) : '[invalid type]';
if (!ISO_ALPHA2_UPPER_REGEX.test(entry.market)) {
  return badRequest(`Invalid market "${safeMarket}". Must be ISO 3166-1 alpha-2 uppercase.`);
}
```

Same for `entry.language` (cap at 16).

---

### 🟠 Medium — SpaceCat org UUID + internal API path leaked in 404 body

`src/controllers/llmo/llmo-onboarding.js` M5 block  
`src/controllers/llmo/llmo.js` L1131

```js
// onboarding.js
const workspaceError = new Error(
  `Semrush workspace not bound to org ${spaceCatId}. `
  + `Run PATCH /organizations/${spaceCatId} with { semrushWorkspaceId: ... } ...`
);
// llmo.js
return notFound(cleanupHeaderValue(error.message)); // full string reaches caller
```

`cleanupHeaderValue` strips header-unsafe chars but not the UUID or path content.

```js
workspaceError.publicMessage = 'Semrush workspace not bound to this org. Contact support.';
workspaceError.operatorHint  = `Run PATCH /organizations/${spaceCatId} …`;
// llmo.js: return notFound(error.publicMessage);
// log includes error.operatorHint for operators in CloudWatch
```

---

### 🟠 Medium — `new URL(baseURL)` throws outside per-tuple guard

`src/controllers/llmo/llmo-onboarding.js` ~L1325

```js
const brandDomain = new URL(baseURL).hostname; // ← before the for-loop, no try/catch
```

An invalid `baseURL` throws a `TypeError` that escapes `performSerenityFanOut` entirely — caller gets an unhandled throw instead of a structured `{ requested, succeeded: [], failed }`. `baseURL` is pre-validated upstream so real-world risk is low, but the exported function's standalone contract is broken.

```js
let brandDomain;
try {
  brandDomain = new URL(baseURL).hostname;
} catch {
  return failAll(400, 'invalidBaseURL');
}
```

---

### 🟡 Low — `extractImsBearer` returns empty string as valid token

When `Authorization` is exactly `"Bearer "` (trailing space, no token), `header.substring('Bearer '.length)` returns `""` and passes the null-check downstream, forwarding an empty IMS token to the upstream.

```js
const token = header.substring('Bearer '.length).trim();
return token.length > 0 ? token : null;
```

---

### 🟡 Low — OpenAPI spec missing `maxItems: 50` on markets array

`docs/openapi/llmo-api.yaml`

The controller enforces `MAX_MARKETS = 50` and returns 400 on overflow, but the spec says unlimited. SDK generators and callers learn the cap from a 400 round-trip.

```yaml
markets:
  type: array
  minItems: 1
  maxItems: 50
```

---

### 🟡 Low — `reconcileSerenityProjects` silent last-write-wins on duplicate DB rows

`src/controllers/llmo/llmo-onboarding.js` — `reconcileSerenityProjects`

If two rows share the same `(geoTargetId, languageCode)` for the same brand (concurrent re-onboards, race in the 409 path), the second silently overwrites the first in the Map with no log or diagnostic.

```js
rows.forEach((row) => {
  const key = `${row.getGeoTargetId()}::${row.getLanguageCode()}`;
  if (rowBySlice.has(key)) {
    log.warn(`Duplicate BrandSemrushProject row for brand ${brandId}, slice ${key}`);
    return;
  }
  rowBySlice.set(key, row);
});
```

MysticatBot (approved):
Hey @luis6156,

## Strengths

- **All 7 prior review findings addressed comprehensively** - error message sanitization (opaque `internalError` token), deadline guard, input truncation (`.slice(0, 16)`), empty bearer (`hasText(token)` guard), OpenAPI `maxItems: 50`, generic 404 message (operator detail logged server-side only), duplicate row handling (first-write-wins with warn log). Each fix is targeted and correct.
- **Trust-but-verify M8 reconciliation** (src/controllers/llmo/llmo-onboarding.js:421-483): Reading the authoritative DB state after the fan-out rather than trusting upstream 201 responses catches the silent-write-failure class of bug. Graceful degradation on read-back failure is the right call.
- **Defense-in-depth auth boundary** (src/controllers/llmo/llmo-onboarding.js:215-230): `extractImsBearer` validates three independent conditions (auth type === 'ims', header prefix, non-empty token) before forwarding. Stricter than the existing proxy pattern by design.
- **Error isolation in `performSerenityFanOut`** (src/controllers/llmo/llmo-onboarding.js:256-398): The failAll/per-tuple pattern with explicit pre-condition checks (brand, token, transport, URL), per-tuple error containment, and sanitized error tokens to callers is well-structured.
- **Preflight fail-fast before state creation** (src/controllers/llmo/llmo-onboarding.js:517-559): Evaluating the cohort gate and workspace pre-condition immediately after the org lookup but before site/brand/entitlement creation means a missing workspace leaves zero orphaned state.
- **Input validation at the boundary** (src/controllers/llmo/llmo.js:711-752): Type check, format regex, cap at 50, and duplicate collapse - all before handing to business logic. The `typeof` guard before regex prevents `.test()` on non-strings.
- **Idempotent re-invocation by design** (src/controllers/llmo/llmo-onboarding.js:370-375): 409-as-success combined with M8 read-back makes the entire fan-out naturally idempotent.
- **Deadline guard as structural safety net** (src/controllers/llmo/llmo-onboarding.js:318-331): Self-protects against Lambda timeout with a structured 207 response on deadline exhaustion rather than an abrupt invocation death.
- **Structured metrics with no PII** (serenity_onboarding_start, serenity_project_fanout, serenity_onboarding_complete): Clean field naming, explicit denominator/numerator pattern, and test assertions on exact key sets to catch future PII introduction.
- **Comprehensive test coverage**: 14 tests for performSerenityFanOut, 8 for reconcileSerenityProjects, full M7-M8 integration seam, controller validation edge cases, and metric property completeness assertions.
- **Conventions compliance**: JS-module imports for static data, controller factory pattern, DTO discipline (model getters, not raw objects), OpenAPI updated in lockstep.

## Issues

### Minor (Nice to Have)

**1. Composite key separator inconsistency between layers**

src/controllers/llmo/llmo.js:739 uses `:` (`${market}:${language}`) while src/controllers/llmo/llmo-onboarding.js:443,451 uses `::` (`${geoTargetId}::${languageCode}`). These operate on different data at different layers so it is not a bug, but it creates a maintenance trap when grepping for tuple-identity patterns.

**Fix**: Use `::` consistently, or extract a `tupleKey(a, b)` helper.

**2. `outcome.body?.error` may forward upstream-originated content in the `failed[]` response**

src/controllers/llmo/llmo-onboarding.js:379

When `handleCreateMarket` returns a non-2xx/non-409 status, `outcome.body?.error` is placed directly into the response. If the upstream ever echoes request-body content in its error field, caller-supplied strings could round-trip. Real-world exploitability is low (upstream is a controlled Adobe proxy, inputs are regex-validated), but it differs from the thrown-error path which correctly sanitizes.

**Fix**: Apply the same opaque-token pattern: `error: 'upstreamRejected'`, or allowlist known safe tokens.

**3. Missing test for deadline guard firing mid-loop (partial success + timeout)**

test/controllers/llmo/llmo-onboarding.test.js, performSerenityFanOut suite

The tests cover "deadline imminent from start" (all fail) and "no deadline" (guard skipped), but not the path where the first N tuples succeed and then the deadline guard fires mid-loop. The `requested.slice(succeeded.length + failed.length)` calculation in the timeout branch is correct but non-obvious and would benefit from regression protection.

**Fix**: Add a test with a fake clock where the first call succeeds and the deadline crosses the 15s threshold before the second iteration.

## Recommendations

- The `performSerenityFanOut` function is ~140 lines. Extracting the pre-condition block (brand, IMS, transport, URL) into a helper returning either a context object or the `failAll` result would improve readability without changing behavior.
- The sequential-to-async migration path is clear: `performSerenityFanOut`'s `{requested, succeeded, failed}` interface is already the right shape for an SQS job result. Consider a one-line comment noting the SQS migration is the GA path, so a future reader does not attempt to "fix" the sequential loop with in-process `Promise.allSettled`.

## Assessment

Ready to merge? Yes
Reasoning: The architecture is well-considered for its stated purpose (cohort prototype). Every concern from the prior review has been addressed. The sequential fan-out is bounded by the deadline guard, the trust-but-verify pattern catches silent write failures, error isolation and sanitization are thorough, and the test suite covers every significant code path. The three Minor findings are defense-in-depth observations, not blockers.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 59s | <b>Cost:</b> $5.82 | <b>Commit:</b> <code>d969fefa09a5ebd85c307397c402082ce92f36ab</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/serenity/handlers/markets.js
```diff
@@ -71,6 +71,33 @@ export function resolveLocation(market) {
...
+ * @returns {string|null} Uppercase alpha-2 code (e.g. 2840 → 'US'), or null.
+ */
+export function marketFromGeoTargetId(geoTargetId) {
+  const n = Number(geoTargetId);
+  if (!Number.isInteger(n) || n <= 2000) {
+    return null;
+  }
+  // ISO numeric codes are 3-digit, zero-padded (e.g. US=840, AF=004).
+  const numeric = String(n - 2000).padStart(3, '0');
+  return numericToAlpha2[numeric] ?? null;
```
MysticatBot, line 98: issue (blocking): `handleListProjects` maps every DB row without filtering, but `marketFromGeoTargetId` returns `null` for non-country geo target IDs (region/metro/city ranges). If such a row exists in `BrandSemrushProject`, the response includes `{ market: null, ... }`. The OpenAPI schema declares `market` as required with `type: string` and `pattern: `^[A-Z]{2}$`` - a null violates the contract and breaks typed SDK clients.

Fix: filter out rows where the reverse-mapping returns null:
```js
items: rows
  .map((row) => ({
    market: marketFromGeoTargetId(row.getGeoTargetId()),
    language: row.getLanguageCode(),
    status: 'live',
  }))
  .filter((item) => item.market !== null),
```
Add a test with a non-country geoTargetId (e.g. 1234) to verify the filtering.

test/controllers/serenity.test.js
```diff
@@ -79,6 +79,7 @@ describe('SerenityController', () => {
     handleUpdatePrompt: sinon.stub(),
```
MysticatBot, line 79: issue (blocking): CLAUDE.md rule: "New or modified endpoints must include integration tests in `test/it/` - add shared test logic in `shared/tests/`, seed data in `postgres/seed-data/`, and a wiring file in `postgres/`".

The new `GET /v2/orgs/:spaceCatId/brands/:brandId/serenity/projects` endpoint has unit tests and an OpenAPI contract fixture, but no integration test in `test/it/`. The existing `test/it/shared/tests/serenity.js` already tests the 401-on-JWT contract for other serenity endpoints - extending it for this route is straightforward.

Fix: add to the existing serenity IT suite:
```js
it('401s on GET projects with JWT auth (IMS-only contract)', async () => {
  const res = await http.admin.get(
    `/v2/orgs/${ORG_1_ID}/brands/${BRAND_1_ID}/serenity/projects`
  );
  expect(res.status).to.equal(401);
});
```

MysticatBot (changes requested):
Hey @luis6156,

**Verdict: Request changes** - two gaps in the new GET /serenity/projects endpoint need addressing before merge.
_Changes: Adds SERENITY_SITE_ALLOWLIST cohort gate, markets[] DTO with validation on POST /llmo/onboard, sequential Semrush project fan-out (M5-M8) with deadline guard and trust-but-verify read-back, structured metrics, and a new GET /serenity/projects read-back endpoint for the onboarding UI (18 files)._

## Must fix before merge
1. [Important] `handleListProjects` can return `market: null` for non-country geo targets, violating the OpenAPI schema - `src/support/serenity/handlers/markets.js` (details inline)
2. [Important] Missing integration test for `GET /v2/orgs/:spaceCatId/brands/:brandId/serenity/projects` per CLAUDE.md rule - `test/it/` (details inline)

<details>

- nit: `listProjects` controller passes `log` as 5th arg but `handleListProjects` signature does not declare it; conversely accepts `transport` and `semrushWorkspaceId` it never uses - `src/controllers/serenity.js` / `src/support/serenity/handlers/markets.js`
- suggestion: Consider a `brandName.length > 256` early-return in the controller, consistent with the 16-char truncation on other reflected values - `src/controllers/llmo/llmo.js`
</details>

- Composite key separator inconsistency (`:` vs `::`) - unified
- `outcome.body?.error` forwarding upstream content - now uses opaque `upstreamRejected` token
- Missing test for deadline guard firing mid-loop - added in d4512197

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 25s | <b>Cost:</b> $7.65 | <b>Commit:</b> <code>3d1609b52c3f53a9ebd640b1a958dcf87ec56f0d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @luis6156,

**Verdict: Approve** - clean consolidation that resolves both prior findings by eliminating the redundant endpoint.
_Changes: Removes the GET /serenity/projects endpoint entirely and folds the derived ISO `market` field into the existing /serenity/markets list response with correct nullable semantics (11 files)._

- `handleListProjects` returning `market: null` for non-country geo targets - endpoint removed; market added to /markets with honest `oneOf: [string, null]` schema
- Missing integration test for GET /serenity/projects - endpoint removed; /markets already has IT coverage in `test/it/shared/tests/serenity.js`

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 27s | <b>Cost:</b> $2.71 | <b>Commit:</b> <code>112364b7e52c1bcee014613da829efe98f3eb024</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

andreeastroe96:
**Supersession note (audit 2026-07-21)**

`main` now implements the Serenity cohort gate this PR added: `src/support/serenity/serenity-active.js` → `isSerenityActiveForOrg` reads the same `LLMO`/`serenity` `feature_flags` row (a superset — it adds a bounded module cache + fail-safe-to-inactive on read error).

More broadly, Serenity onboarding on `main` runs entirely through the **sub-workspace surface** (brand-create → `POST /serenity/activate` → `POST /serenity/markets`), not the `POST /llmo/onboard` `markets[]` fan-out this PR extends — `/llmo/onboard` has **no** Serenity references on `main`. The `markets[]` DTO validation and `MAX_MARKETS` cap also already exist on the sub-workspace `/serenity/markets` path (`src/controllers/serenity.js`).

Net: the concepts here (DB cohort gate, markets DTO) are already on `main` under the sub-workspace model, and the fan-out path is retired. Recommend closing as superseded — nothing here needs porting. (Deferred-publish/finalize from the stacked #2584 is the one piece not on main; it's being re-targeted onto the sub-workspace publish path separately.)

PR #2514, opened 2026-05-29
feat: lazy-load tab data, eager badge counts via totals endpoints | LLMO-5218

calvarezg:
- [x] Related issue linked below
- [ ] When merging/squashing, make sure the fixed issue references are visible in the commits

- [x] New endpoints added to `ai-visibility-api.yaml` with full parameter specs
- [x] New response schemas added to `schemas.yaml`
- [x] Endpoints are fully implemented (no "Not implemented yet" note needed)

- [x] `ai-visibility-api.yaml` and `schemas.yaml` updated to reflect the separation of data vs. totals

## Summary
Separates the per-page data fetch from the totals count on the AI Visibility v1 topic endpoints, so the UI can load tab badge counts cheaply without fetching full paginated data.

- `GET /llmo/ai-visibility/v1/topic/brand-topics-totals` — new handler returning `{ total }` only (calls `brandTopicsTotals`, no pagination)
- `GET /llmo/ai-visibility/v1/topic/gap-topics-totals` — new handler returning `{ totals: [...] }` only (calls `gapTopicsTotals`, no pagination)
- `brand-topics.js` and `gap-topics.js` simplified: totals call removed, each now returns only `{ data }`
- Filter builder functions exported from both handlers so the new totals handlers can reuse them without duplication

## Deploy order
These two PRs ([UI](https://github.com/adobe/project-elmo-ui/pull/1874) + [API](https://github.com/adobe/spacecat-api-service/pull/2514)) can be released in either order without breaking the tables:

- **UI first**: the new totals endpoints don't exist yet → badge counts stay blank/hidden, tables fully functional.
- **API first**: the data endpoints no longer return `total`/`totals`, but the UI no longer reads those fields → badges stay blank/hidden until the UI ships, tables fully functional.

## Related Issues
https://jira.corp.adobe.com/browse/LLMO-5218

Thanks for contributing!


calvarezg:
## Self-review (fresh-context)

**[Intent]** `gap-topics-totals.js` accepts `gapKinds` as a documented query parameter (OpenAPI spec line 1262 lists `*v1GapTopicsGapKindsParam`, and the description says the endpoint accepts the same filters as gap-topics "including `gapKinds`"), but the value is never read or forwarded in the `GapTopicsTotalsRequest`. `GapTopicsTotalsRequest` in the proto intentionally has no `kind` field — the service returns a per-kind breakdown regardless — so the server-side behavior is correct. However, a caller sending `gapKinds=STRONG` will get back all kinds anyway, which could be surprising. Worth a clarifying note in the endpoint description (e.g., "`gapKinds` is accepted for API consistency but has no effect — the response always includes totals for all kinds") or removing the parameter from the spec if it is never intended to filter.

dmitriigulkin-semrush (approved):

dmitriigulkin-semrush:
great job 👍 🔥

PR #101, opened 2024-01-19
feat: 404 report for customers

alinarublea:

PR #2515, opened 2026-05-29
docs(plans): per-target reviewer login (web tier) implementation plan

solaris007:
## Summary

Bite-sized TDD implementation plan for the **web-tier half** of per-target reviewer login. Docs-only (adds `docs/plans/2026-05-29-per-target-reviewer-login.md`).

Spec: [`per-target-reviewer-login.md`](https://github.com/adobe/mysticat-architecture/blob/main/platform/decisions/per-target-reviewer-login.md) (adobe/mysticat-architecture#101), extending the multi-destination ADR (#94 / web #2503). Companion worker plan: adobe/mysticat-github-service#31.

## What the plan implements

- **`parseTargets`** validates a per-entry `reviewerLogin`: **required** for non-`default` (enterprise-matched) entries, **optional** on the `default` catch-all; charset `^[A-Za-z0-9][A-Za-z0-9-_]*(\[bot\])?$`, max 64.
- **`github-webhook-hmac-handler.js`** attaches the matched target's `reviewerLogin` to the post-HMAC auth profile as `reviewer_login` (legacy path attaches nothing).
- **`webhooks.js`** resolves `profile.reviewer_login || env.GITHUB_REVIEWER_LOGIN` and passes it to `getSkipReason`.
- **`getSkipReason`** gains a 5th param `reviewerLogin = env.GITHUB_REVIEWER_LOGIN`; expected reviewer becomes `reviewerLogin?.trim() || ${appSlug}[bot]`. 3/4-arg callers unchanged.

4 tasks, each TDD (failing test -> impl -> pass -> commit). The plan also handles updating the existing `VALID_TARGETS` / `noDefault` test fixtures (whose non-`default` `ghec` entries now require a `reviewerLogin`), plus a Validation section (`npm test`, `npm run lint`) and the `GITHUB_TARGETS` config example.

## Note

Plan only - implementation lands in follow-up PR(s) per the plan's tasks.


MysticatBot-Dev (commented):
Hey @solaris007,

Clean, well-structured implementation plan. A couple of design clarifications worth addressing before an agent executes this verbatim.

## Strengths

- **Backward-compatibility by default parameter** (line 7, Architecture section): Threading `reviewerLogin` through the existing target/profile/controller chain and defaulting the 5th arg to `env.GITHUB_REVIEWER_LOGIN` means all existing callers are unaffected. This is the right way to extend a multi-caller API surface without a flag day.
- **Fail-fast validation at config-parse time** (Task 1): Making `reviewerLogin` required on non-default targets with a strict regex means misconfigurations blow up at startup rather than silently falling back to the wrong reviewer identity at runtime. Correct failure mode for Vault-provisioned config.
- **Explicit fixture maintenance callouts** (Task 1, Step 5): Noting that existing `VALID_TARGETS` and `noDefault` fixtures need updating before the full-file run prevents the "works in isolation, breaks in CI" failure.
- **Conventions compliance**: Correctly placed in `docs/plans/` per the Information Hierarchy, validation gates at every step, PR title follows conventional commits.

## Issues

### Important (Should Fix)

1. **Dual-source fallback creates a subtle null/undefined mismatch** - `docs/plans/2026-05-29-per-target-reviewer-login.md`:7

   The plan says: (a) the controller resolves `profile.reviewer_login || env.GITHUB_REVIEWER_LOGIN` and passes it as arg 5, AND (b) `getSkipReason` defaults param 5 to `env.GITHUB_REVIEWER_LOGIN`. This means the env-fallback lives in two places. If the controller's `||` resolves to `null` (profile was unset, env was unset), the function-level default does NOT fire (null is not undefined). If a future caller passes `undefined` explicitly, the default fires correctly.

   The plan should pick one owner for the fallback and be explicit: either the controller passes `profile.reviewer_login || undefined` (letting the function default handle env resolution in one place), or the controller is the sole resolver and the function default is documented as a safety net for legacy callers only. One sentence of clarification prevents a subtle bug during implementation.

### Minor (Nice to Have)

2. **Regex hyphen placement is ambiguous** - `docs/plans/2026-05-29-per-target-reviewer-login.md`:27

   `^[A-Za-z0-9][A-Za-z0-9-_]*(\[bot\])?$` has an unescaped hyphen between `9` and `_` in the character class. In V8 this works, but it is a lint magnet and confusing to read. Since this plan will be copied verbatim by an agent, placing the hyphen at the end of the class (`[A-Za-z0-9_-]`) eliminates the ambiguity.

3. **Task 1 Step 5 fixture-update instructions appear after the run command** - `docs/plans/2026-05-29-per-target-reviewer-login.md`:146

   The critical instruction to patch `VALID_TARGETS` and `noDefault` is buried below the `npx mocha` invocation. An agent following steps sequentially will hit a red test before reading the fix. Moving the fixture-update instructions above the run command (or into a "Step 4b: patch existing fixtures") avoids a wasted debug cycle.

## Recommendations

- Consider whether the controller should pass `profile.reviewer_login || undefined` (single fallback owner) rather than `profile.reviewer_login || env.GITHUB_REVIEWER_LOGIN` (dual). This is a design decision worth documenting explicitly in the plan so the implementer does not have to infer it.

## Assessment

Ready to merge? Yes (with minor clarification)

Reasoning: This is a well-structured TDD implementation plan that correctly identifies threading points, maintains backward compatibility, and respects existing patterns. The dual-fallback ambiguity is worth a one-sentence clarification before an agent executes the plan, but it does not block merging the document itself.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 22s | <b>Cost:</b> $2.31 | <b>Commit:</b> <code>8b4004cab8522a3c580e8320c5c09631db458b00</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

solaris007:
Addressed in aa4328db.

- **#2 (regex hyphen) - fixed.** Reordered to `^[A-Za-z0-9][A-Za-z0-9_-]*(\[bot\])?$` (trailing hyphen; identical character set, unambiguous when copied verbatim).
- **#3 (fixture update buried after the run command) - fixed.** Task 1 now patches the `VALID_TARGETS` / `noDefault` fixtures in a dedicated **Step 5, before** the full-file run (Step 6), so an executor won't hit a red suite before reading the fix.
- **#1 (dual-source fallback) - kept the form, clarified rather than changed.** This mirrors the adjacent `appSlug` handling (`profile.app_slug || env.GITHUB_APP_SLUG` + a `getSkipReason` default) - the established convention; switching only `reviewerLogin` to `|| undefined` would make the two siblings inconsistent and diverge from the merged spec (#101). The `null` path isn't reachable here: env vars are `undefined` when unset (never `null`), so the controller's `||` yields `undefined` and the function-level default fires correctly - no null/undefined gap. Added a sentence to the Architecture and Task 4 sections documenting this ownership explicitly (controller resolves the webhook path; the `getSkipReason` default covers direct/legacy callers).

MysticatBot-Dev (approved):
Hey @solaris007,

## Strengths

- **All prior findings addressed cleanly.** The dual-fallback null/undefined concern is now explicitly documented in the Architecture section with the reasoning that unset env vars are `undefined` (never `null`), so the `||` chain works correctly. The regex hyphen is trailing, and fixture updates precede the full-file run. Good responsiveness to review feedback.
- **Backward-compatible API extension** (docs/plans/2026-05-29-per-target-reviewer-login.md:13): The 5th param defaulting to `env.GITHUB_REVIEWER_LOGIN` means all existing 3/4-arg callers are unchanged. This mirrors the sibling `appSlug` pattern exactly, so the next developer sees one consistent shape.
- **Fail-fast validation at config-parse time** (lines 38-173): Making `reviewerLogin` required on non-default targets with a strict regex means misconfigurations blow up at startup (during Vault secret resolution), not at webhook-receive time under load. Error messages include the target id for immediate context.
- **TDD sequencing is disciplined**: The pattern of "write failing test, verify it fails for the RIGHT reason, implement, verify green, run full file for regressions" is exactly the rigor needed when an agent will execute this plan verbatim.
- **Fixture maintenance called out explicitly** (Task 1 Step 5): Identifies which existing fixtures need patching and which do not (with reasoning), preventing the "works in isolation, breaks in CI" failure mode.
- **Conventions compliance**: Correctly placed in `docs/plans/`, PR title follows conventional commits, validation gates at every step with concrete runnable checks.

## Issues

### Minor (Nice to Have)

1. **Task 2 Step 1 fixture update is implicit** - docs/plans/2026-05-29-per-target-reviewer-login.md:187-203

   The plan says "First add `reviewerLogin` to BOTH the `ghec` entry (required) inside that `describe`'s `TARGETS`" but embeds the fixture fix mid-sentence in the test-writing step. Unlike Task 1 which explicitly calls out Step 5 as "patch existing fixtures FIRST (before the full-file run)", here the fixture modification is less visible.

   **Why it matters:** Consistency with Task 1's explicit "patch fixtures first, then run" pattern would make this more reliably executable by an agent. An agent following steps literally may write test assertions first and only then realize the fixture needs updating.

   **Fix:** Add a brief callout (similar to Task 1's Step 5) noting that the registry-path `TARGETS` fixture is being modified to add the required `reviewerLogin` field so `parseTargets` won't throw during handler test setup.

## Recommendations

- **Cross-tier config invariant** (line 509): The plan notes that the web tier's `reviewerLogin` must match the worker's `targets.ghec.reviewer_login` and that this is "enforced operationally, not by the web tier in isolation." If the Vault secret template for both services draws from the same source variable, this invariant becomes impossible to violate. If the two secrets are provisioned independently, a config-validation CI step that diffs the two Vault paths would catch drift before production. Future-hardening opportunity, not a blocker for this plan.

## Assessment

Ready to merge? Yes

Reasoning: This is a well-structured, implementation-ready plan that follows established patterns (backward-compatible default params, fail-fast config validation, consistent dual-fallback shape). All prior review findings have been addressed with clear reasoning. The single Minor finding is a polish item for agent executability.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 27s | <b>Cost:</b> $1.98 | <b>Commit:</b> <code>a3ebcd186a234e5342632a1d6e3620e2556d1aa9</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2517, opened 2026-05-29
fix(serenity): propagate upstream 401/403 instead of mapping to 502

rainer-friederich:
## Summary

- When Semrush returns 403 (user has no ACL access to the workspace), the serenity controller was returning 502 `serenityUpstreamError` — a misleading server-error signal for what is actually an access-denied condition
- Same problem for 401 upstream responses
- `SerenityTransportError` with status 401 now maps to 401 `authenticationRequired`; status 403 maps to 403 `forbidden`, using the same `errorTokenForStatus` path already in place for `ErrorWithStatusCode`
- All other upstream errors (5xx etc.) continue to surface as 502 `serenityUpstreamError`

**Root cause found via:** Vlad getting 502 on `GET /v2/orgs/.../brands/.../serenity/prompts?geoTargetId=2036&languageCode=en` — Semrush returned 403 `invalid access attempt` because his IMS user was not in the workspace ACL. The controller's `mapError` branch for `SerenityTransportError` had no status discrimination and always returned 502.

## Test plan

- New unit tests in `test/controllers/serenity.test.js` covering the 403 and 401 propagation cases, verifying no upstream body leakage
- Existing 502 test for `SerenityTransportError(503, ...)` continues to pass unchanged
- All 24 serenity controller tests pass


MysticatBot-Dev (approved):
Hey @rainer-friederich,

## Strengths

- **Clean reuse of existing infrastructure** (`src/controllers/serenity.js:111-115`): The fix reuses the established `errorTokenForStatus` lookup and `safeError` sanitization rather than introducing parallel logic, keeping the error-mapping consistent between `ErrorWithStatusCode` and `SerenityTransportError` for auth-class failures.
- **Leak-safe by construction** (`test/controllers/serenity.test.js:365-389`): Both 401 and 403 test cases construct a `MockTransportError` with a `{ secret: 'leak' }` payload and explicitly assert it does not appear in the serialized response body.
- **Well-scoped change**: A single behavioral fix in one function with no unrelated modifications mixed in. The early-return placement is correct (before the generic 502 fallthrough), and the existing 502 path remains undisturbed.

## Issues

### Minor (Nice to Have)

**Upstream message passthrough** - `src/controllers/serenity.js:113`

The `safeError(e.message)` output is returned as the `message` field in the HTTP response. If the upstream service (Semrush) ever returns verbose error messages containing internal details (endpoint paths, account IDs, rate limit metadata), those would flow through to the caller. This is low risk given that `safeError()` is the established pattern in this codebase for exactly this purpose, and the existing `ErrorWithStatusCode` path does the same thing - just worth being aware of as the integration matures.

## Assessment

Ready to merge? Yes

The fix is minimal, correct, well-tested, and consistent with existing error-mapping conventions. It eliminates a genuinely misleading 502 for what is a client-side access problem, giving API consumers actionable signals instead of a blanket server error.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 33s | <b>Cost:</b> $2.02 | <b>Commit:</b> <code>49e9340997537e121e3a1d7b27380341054982b7</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2518, opened 2026-05-29
fix(auth): extend RouteScopedLegacyApiKeyHandler to match sub-paths of scoped routes

ravverma:
POST /event/fulfillment/xxxx was falling through to LegacyApiKeyHandler because the handler used exact suffix matching. Replace with prefix-aware matching so any sub-path (and query-string variants) of a scoped route is handled correctly.

## Related Issues

Thanks for contributing!


MysticatBot-Dev (approved):
Hey @ravverma,

## Strengths

- Well-scoped fix that does exactly one thing correctly. The `isScopedRoute` function cleanly separates matching logic from auth delegation (`src/support/route-scoped-legacy-api-key-handler.js:22-30`).
- The prefix match with the trailing `/` delimiter (`routeKey.startsWith(`${prefix}/`)`) prevents false positives like `/event/fulfillment-extra` matching against the `/event/fulfillment` prefix. This is the detail that matters and it is handled correctly.
- Query string stripping before matching (`suffix.split('?')[0]`) ensures query parameters cannot influence route matching decisions.
- Tests cover both the sub-path case and the query-string case, with clear AAA pattern and behavior-focused assertions (`test/support/route-scoped-legacy-api-key-handler.test.js:137-149`).
- JSDoc comment updated to reflect the new sub-path semantics.
- Conventional commit format and behavior change accompanied by tests.

## Issues

### Minor (Nice to Have)

1. **Duplicated query-string stripping** - `src/support/route-scoped-legacy-api-key-handler.js:26` and `:58`

   The `suffix.split('?')[0]` logic appears once inside `isScopedRoute` and again in `checkAuth` to build the log message `routeKey`. If the stripping logic ever needs to change, there are two sites to update. Consider having `isScopedRoute` return the cleaned route key (or `null` for no-match) so `checkAuth` can reuse it directly.

2. **Missing negative boundary test for prefix-without-slash-delimiter** - `test/support/route-scoped-legacy-api-key-handler.test.js`

   There is no test asserting that a path like `/event/fulfillmentevil` (prefix match without a `/` separator) returns `null`. The production code handles this correctly, but for an auth-scoped handler a regression test protects the security invariant. If a future refactor simplifies to `.startsWith(prefix)` without the trailing slash check, the test suite would still pass while silently expanding the auth boundary.

   Suggested addition:
   ```js
   it('returns null for POST /event/fulfillmentevil (prefix match without slash delimiter)', async () => {
     const ctx = { pathInfo: { method: 'POST', suffix: '/event/fulfillmentevil' } };
     const result = await handler.checkAuth({}, ctx);
     expect(result).to.be.null;
     expect(superCheckAuthStub).to.not.have.been.called;
   });
   ```

## Assessment

Ready to merge? Yes
Reasoning: The fix is correct, architecturally sound, well-tested, and narrowly scoped. The prefix matching with the `/` boundary delimiter is the right approach for hierarchical route matching and will not need revisiting as additional sub-paths appear. The two minor findings are hardening suggestions, not blockers.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 36s | <b>Cost:</b> $2.16 | <b>Commit:</b> <code>3f78028183262011bbff01a8d822deb7d8f3137d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2520, opened 2026-05-30
feat: per-target reviewer login (web tier)

solaris007:
## Summary

Implements the **web-tier half** of per-target reviewer login, per the plan in #2515 and the spec [`per-target-reviewer-login.md`](https://github.com/adobe/mysticat-architecture/blob/main/platform/decisions/per-target-reviewer-login.md) (adobe/mysticat-architecture#101; extends #94 / #2503). Companion worker PR: adobe/mysticat-github-service#33.

Threads a per-destination `reviewerLogin` from the `GITHUB_TARGETS` registry through the HMAC auth profile into the `getSkipReason` trigger gate, so a destination like `ghec` reacts to its own reviewer account instead of the global `MysticatBot`. Additive and backward-compatible.

- **`parseTargets`** (`github-targets.js`): validates a `reviewerLogin` per entry - **required** on non-`default` (enterprise-matched) entries (throws at config-parse), **optional** on the `default` catch-all; charset `^[A-Za-z0-9][A-Za-z0-9_-]*(\[bot\])?$`, max 64.
- **`github-webhook-hmac-handler.js`**: attaches the matched target's `reviewerLogin` to the post-HMAC auth profile as `reviewer_login`; legacy/no-match paths attach nothing.
- **`getSkipReason`** (`github-trigger-rules.js`): new 5th param `reviewerLogin = env.GITHUB_REVIEWER_LOGIN`; `expectedReviewer = reviewerLogin?.trim() || ${appSlug}[bot]`. `isMysticatTargetedSkip` unchanged; 3/4-arg callers unaffected.
- **`webhooks.js`**: resolves `profile.reviewer_login || env.GITHUB_REVIEWER_LOGIN` and passes it as the 5th arg.

TDD, one commit per task:
```
179c72b0 feat(github-targets): validate per-target reviewerLogin (required for non-default)
36df15e9 feat(github-webhook): attach per-target reviewer_login to auth profile
4dd62b75 feat(github-trigger-rules): add per-target reviewerLogin 5th arg to getSkipReason
b792c83f feat(webhooks): resolve per-target reviewer_login and pass to getSkipReason
```

## Test plan
- [x] `npm test` → **11030 passing**, 0 failing. New cases across `parseTargets` (required/optional/charset/length), HMAC profile (carries / legacy-omits), `getSkipReason` (5-arg match / mismatch / back-compat default / `[bot]` tail), controller (per-target gate / mismatch-skip / legacy-fallback).
- [x] `npm run lint` clean.
- [x] Independent spec-compliance + code-quality review: clean (all requirements met, nothing extra; existing `appSlug` test still green via the unchanged default path).

## Notes
- Three small adaptations vs the plan, all reviewed: the `dup`/`noDefault` fixtures gained a `reviewerLogin` (their first non-`default` entry now hits the new check before the duplicate-id/exactly-one check), and two lint fixes (multi-line signature for `max-len`; a test-description string for `no-template-curly-in-string`). No assertions weakened.
- **Deploy/config:** `github-public` (the `default` entry) omits `reviewerLogin` and falls back to `env.GITHUB_REVIEWER_LOGIN` per env. As the per-env reviewer model rolls out, dev's `GITHUB_REVIEWER_LOGIN` should be `MysticatBot-Dev` and prod's `MysticatBot`. `ghec` supplies its own `reviewerLogin` at cutover.

### DIFF
--- a/src/controllers/webhooks.js
+++ b/src/controllers/webhooks.js
@@ -100,6 +100,9 @@ function WebhooksController(context) {
     const profile = ctx.attributes?.authInfo?.getProfile?.() || {};
     const targetId = profile.target_id;
     const appSlug = profile.app_slug || env.GITHUB_APP_SLUG;
+    // Per-target reviewer-gate identity. Set for non-default targets by the HMAC
+    // handler; the default/legacy path falls back to the global env knob.
+    const reviewerLogin = profile.reviewer_login || env.GITHUB_REVIEWER_LOGIN;
 
     // Security-relevant: which bot can trigger automated runs. In registry mode
     // this comes from the target's appSlug; in legacy mode from env. A missing
@@ -145,7 +148,7 @@ function WebhooksController(context) {
     }
 
     // Apply trigger rules (returns skip reason string or null)
-    const skipReason = getSkipReason(data, action, env, appSlug);
+    const skipReason = getSkipReason(data, action, env, appSlug, reviewerLogin);
     if (skipReason) {
       log.info('Skipping webhook', {
         skipReason,
--- a/src/support/github-targets.js
+++ b/src/support/github-targets.js
@@ -75,6 +75,23 @@ export function parseTargets(env) {
     if (isDefault && i !== parsed.length - 1) {
       throw new Error(`GITHUB_TARGETS default entry "${t.id}" must be last`);
     }
+    // reviewerLogin is the trigger-gate identity (the user we react to as the
+    // requested reviewer). REQUIRED on non-default (enterprise-matched) entries
+    // so a destination missing its reviewer fails loudly at config-load rather
+    // than silently falling back to the global GITHUB_REVIEWER_LOGIN (wrong for
+    // that destination). OPTIONAL on the default entry, which keeps the global
+    // fallback. When present, accept plain users (MysticatBot), EMU users
+    // (handle_shortcode), and App bots (slug[bot]).
+    if (!isDefault && (typeof t.reviewerLogin !== 'string' || !t.reviewerLogin.trim())) {
+      throw new Error(`GITHUB_TARGETS["${t.id}"] is missing a string "reviewerLogin" (required for non-default targets)`);
+    }
+    if (t.reviewerLogin !== undefined) {
+      if (typeof t.reviewerLogin !== 'string'
+        || t.reviewerLogin.length > 64
+        || !/^[A-Za-z0-9][A-Za-z0-9_-]*(\[bot\])?$/.test(t.reviewerLogin)) {
+        throw new Error(`GITHUB_TARGETS["${t.id}"].reviewerLogin must match ^[A-Za-z0-9][A-Za-z0-9_-]*(\\[bot\\])?$ and be at most 64 chars`);
+      }
+    }
   });
   const defaults = parsed.filter((t) => t.match?.default === true);
   if (defaults.length !== 1) {
--- a/src/support/github-webhook-hmac-handler.js
+++ b/src/support/github-webhook-hmac-handler.js
@@ -141,7 +141,14 @@ class GitHubWebhookHmacHandler extends AbstractHandler {
     }
     return new AuthInfo()
       .withAuthenticated(true)
-      .withProfile({ user_id: 'github-webhook', target_id: result.id, app_slug: result.appSlug })
+      .withProfile({
+        user_id: 'github-webhook',
+        target_id: result.id,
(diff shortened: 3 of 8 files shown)


MysticatBot-Dev (approved):
Hey @solaris007,

## Strengths

- **Fail-loud on misconfiguration** (`src/support/github-targets.js:78-80`): Requiring `reviewerLogin` on non-default entries and validating charset/length at config-parse time is the right posture. A missing reviewer would silently fall back to the wrong identity, producing hard-to-diagnose "why didn't it trigger" tickets. Making it blow up at boot is a mature choice.
- **Clean backward compatibility** (`src/utils/github-trigger-rules.js:33-38`): The default parameter `reviewerLogin = env.GITHUB_REVIEWER_LOGIN` preserves the 3/4-arg call signature for existing callers. Zero risk of breaking anything outside this PR's scope.
- **Proper trust boundary layering** (`src/support/github-webhook-hmac-handler.js:141-152`): The `reviewer_login` is attached to the auth profile only after HMAC verification succeeds. An attacker cannot influence the profile without a valid webhook secret.
- **Well-structured TDD with behavioral test names**: Each test describes observable behavior, not implementation details. The coverage spans the priority-order fallback chain thoroughly (explicit arg > env > appSlug[bot]).
- **Upstream spec compliance**: The implementation faithfully follows the upstream spec (`adobe/mysticat-architecture/platform/decisions/per-target-reviewer-login.md`) across all four touchpoints (parseTargets, HMAC profile, controller resolution, getSkipReason). Three minor adaptations are explicitly documented in the PR body with rationale.
- **Excellent test-to-code ratio**: ~135 lines of test code for ~36 lines of production code. Codecov confirms 100% line coverage on modified files.

## Issues

### Minor (Nice to Have)

1. **`src/utils/github-trigger-rules.js:33-39` - Positional parameter growth approaching ergonomic boundary.** `getSkipReason` is now at 5 positional parameters. Tolerable today, but if the next target-scoped property also threads through here, a 6th/7th positional arg will make call sites unreadable. The next change that would add a 6th parameter should refactor to an options object (e.g., `getSkipReason(data, action, { env, appSlug, reviewerLogin })`). Worth noting in the team's decision log so the growth is intentional, not accidental.

2. **`test/utils/github-trigger-rules.test.js` - Whitespace-only `reviewerLogin` not tested at the trigger-rules level.** The production code does `reviewerLogin?.trim() || ...`, meaning a whitespace-only string like `'  '` would fall through to the `appSlug[bot]` fallback. While `parseTargets` rejects whitespace-only values at config-load, `getSkipReason` is a public utility that can be called independently. A test with `reviewerLogin = '  '` confirming the fallback behavior would document this edge case for future callers.

3. **`src/support/github-targets.js:87-92` - Redundant `typeof` check in the second validation branch.** The second `if` block (`if (t.reviewerLogin !== undefined)`) re-checks `typeof t.reviewerLogin !== 'string'`. For non-default entries, this is always true by the time execution reaches here (the first check at line 82 already confirmed it). The re-check only provides value for the default entry path (e.g., `reviewerLogin: 123`). Not a bug - the logic is correct - but a brief comment or factoring the regex/length validation into a small helper would make the two cases (required vs optional) structurally distinct for future readers.

## Recommendations

- Consider adding the resolved `reviewerLogin` value (or its source: "profile" / "env" / "appSlug-fallback") to the existing `log.info('Skipping webhook', { ... })` context in the controller. When debugging why a webhook was skipped in production, the first question is always "what reviewer did it expect?"
- Document the per-target identity model in one place. The system now has three identity facets per target: `appSlug` (the GitHub App that acts), `webhookSecretEnvVar` (the HMAC trust anchor), and `reviewerLogin` (the trigger-gate persona). A short section in the `GITHUB_TARGETS` config docs would prevent confusion for the next person adding a target.

## Assessment

Ready to merge? Yes

The change is focused, additive, well-tested, and backward-compatible. The trust model is sound (HMAC guards the boundary, reviewer identity is a second authorization layer behind it), validation is strict and fails loudly, and the fallback chain is clean. The three Minor items are non-blocking observations for future improvement.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 42s | <b>Cost:</b> $3.11 | <b>Commit:</b> <code>b792c83f0458b003559edd0932e75328a2b30d7e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2526, opened 2026-06-01
fix: update plg onboarding existing delivery type

anshulk-public:
## Summary

- **Reject frescopa domains**: Any PLG onboarding request for a domain containing `frescopa` (e.g. `frescopa.com`, `shop.frescopa.com`) is rejected early with a `400` response and a Slack notification posted to `SLACK_PLG_ONBOARDING_CHANNEL_ID`, consistent with how `internal-org` and `paid-customer` rejections are handled.

- **Alert on delivery type mismatch (Step 5a)**: After a site is resolved or created, `findDeliveryType` is called for existing sites (when no preset delivery type is provided) to detect the site's actual delivery type. If the detected type differs from what is stored and is not `OTHER`, a Slack alert is posted to `SLACK_DELIVERY_TYPE_ALERT_CHANNEL_ID` with the site ID, domain, org, and both delivery type values. The site is **not mutated** — operators can investigate and correct manually since `findDeliveryType` is not 100% reliable.

**Jira:**
- https://jira.corp.adobe.com/browse/SITES-45739
- https://jira.corp.adobe.com/browse/SITES-45737

### `src/controllers/plg/plg-onboarding.js`
- Added `'frescopa-domain'` entry to `PLG_REJECTION_MESSAGES`
- Added frescopa domain check in `onboard()` after `isValidDomain` validation, before the org lookup

### `src/controllers/plg/plg-onboarding/onboarding-flow.js`
- Added **Step 5a** in `performAsoPlgOnboarding` between site resolution (`steps.siteResolved`) and canonical URL resolution (Step 5b)
- Skips newly created sites (delivery type was just set in Step 5)
- Wraps `findDeliveryType` in a try/catch so network failures do not block onboarding
- Posts a Slack alert to `SLACK_DELIVERY_TYPE_ALERT_CHANNEL_ID` including site ID, domain, org ID, org name/IMS org ID, stored type, and detected type

## New environment variable

| Variable | Purpose |
|---|---|
| `SLACK_DELIVERY_TYPE_ALERT_CHANNEL_ID` | Slack channel for delivery type mismatch alerts |

## Related Issues

Thanks for contributing!


src/controllers/plg/plg-onboarding.js
```diff
@@ -1115,6 +1115,24 @@ async function performAsoPlgOnboarding({
...
+      const detectedDeliveryType = await findDeliveryType(baseURL);
+      if (
+        detectedDeliveryType
+        && detectedDeliveryType !== SiteModel.DELIVERY_TYPES.OTHER
+        && detectedDeliveryType !== existingDeliveryType
+      ) {
+        site.setDeliveryType(detectedDeliveryType);
+        if (existingDeliveryType) {
+          site.setDeliveryConfig(null);
+          site.setHlxConfig(null);
```
habansal: just to clarify, is Hlxconfig the only sub-config related to deliverType for different opportunities and features?

cc @sandsinh @tarunsinghdev

src/controllers/plg/plg-onboarding.js
```diff
@@ -1115,6 +1115,24 @@ async function performAsoPlgOnboarding({
...
+      const existingDeliveryType = site.getDeliveryType();
+      const detectedDeliveryType = await findDeliveryType(baseURL);
+      if (
+        detectedDeliveryType
+        && detectedDeliveryType !== SiteModel.DELIVERY_TYPES.OTHER
+        && detectedDeliveryType !== existingDeliveryType
+      ) {
+        site.setDeliveryType(detectedDeliveryType);
+        if (existingDeliveryType) {
+          site.setDeliveryConfig(null);
```
habansal: for this removal, I think, we should have some logging or a way to figure out which config was removed.

src/controllers/plg/plg-onboarding.js
```diff
@@ -1115,6 +1115,24 @@ async function performAsoPlgOnboarding({
     onboarding.setSiteId(site.getId());
     steps.siteResolved = true;
 
+    // Step 5a: Validate and correct delivery type for both new and existing sites
+    if (!presetDeliveryType) {
```
habansal: what is this presetDeliveryType and how is that set? why not update in that case?

src/controllers/plg/plg-onboarding.js
```diff
@@ -1115,6 +1115,24 @@ async function performAsoPlgOnboarding({
     onboarding.setSiteId(site.getId());
     steps.siteResolved = true;
 
+    // Step 5a: Validate and correct delivery type for both new and existing sites
```
habansal: wasn't this being set by the code if it didn't exist or something?
is this flow totally independent of that?

e.g. https://github.com/adobe/spacecat-api-service/blob/main/src/controllers/plg/plg-onboarding.js#L1106

src/controllers/plg/plg-onboarding.js
```diff
@@ -1115,6 +1115,24 @@ async function performAsoPlgOnboarding({
...
+    if (!presetDeliveryType) {
+      const existingDeliveryType = site.getDeliveryType();
+      const detectedDeliveryType = await findDeliveryType(baseURL);
```
habansal: are we 100% confident on our discovery or is there a possibility of a false case there?
cc @Kanishkavijay39

src/controllers/plg/plg-onboarding.js
MysticatBot: **Important:** When `deliveryConfig` and `hlxConfig` are set to null here, there is no record of what they contained. Log the previous values before clearing so operators can audit and recover if needed. This aligns with @habansal's earlier comment - the concern is valid from both an observability and recoverability standpoint.

src/controllers/plg/plg-onboarding.js
MysticatBot: **Important:** This `await findDeliveryType(baseURL)` is now unconditional for every PLG onboarding without a preset type. If it rejects (network timeout, DNS failure), the entire onboarding fails. Since this is a correction mechanism rather than a prerequisite, consider wrapping in try/catch and falling back gracefully:
```javascript
let detectedDeliveryType;
try {
  detectedDeliveryType = await findDeliveryType(baseURL);
} catch (e) {
  log.warn(`Failed to detect delivery type for ${baseURL}: ${e.message}`);
}
```

test/controllers/plg/plg-onboarding.test.js
```diff
@@ -3366,8 +3381,9 @@ describe('PlgOnboardingController', function describePlgOnboarding() {
 
       expect(res.status).to.equal(200);
       expect(mockOnboarding.setStatus).to.have.been.calledWith('ONBOARDED');
-      expect(findDeliveryTypeStub).to.not.have.been.called;
-      expect(mockLog.info).to.have.been.calledWithMatch(/Using existing site delivery type aem_cs/);
+      expect(findDeliveryTypeStub).to.have.been.calledOnceWith(TEST_BASE_URL);
+      expect(existingSite.setDeliveryType).to.have.been.calledWith('aem_edge');
```
MysticatBot: **Important:** This test verifies the delivery type update but does not assert that config clearing also happened (`setDeliveryConfig(null)` and `setHlxConfig(null)`). Since config clearing on type change is the highest-risk behavior in this PR, add assertions to prevent silent regression.

MysticatBot (changes requested):
Hey @anshulk-public,

## Strengths

- `src/controllers/plg/plg-onboarding.js:1118-1135`: Step 5a placement between site resolution and canonical URL resolution is architecturally correct - detecting and correcting delivery type before downstream steps depend on it prevents cascading stale-config propagation.
- `src/controllers/plg/plg-onboarding.js:1120-1131`: The guard conditions are solid - skips when a preset is provided, ignores OTHER detection results, only clears config when an existing type was present, and checks for inequality before mutating.
- `src/controllers/plg/plg-onboarding.js:1554-1556`: Frescopa rejection is placed early in the validation chain (after domain format validation, before any DB work), which is the right ordering for short-circuit rejections. The case-insensitive `includes()` check prevents bypass via mixed-case input.
- `test/controllers/plg/plg-onboarding.test.js:921-933`: Frescopa tests cover both bare domain and subdomain cases with assertions on both status code and response message.
- `test/controllers/plg/plg-onboarding.test.js:3373-3389`: The delivery type correction test asserts on the state mutation (`setDeliveryType`) and the log message pattern, verifying observable behavior.

## Issues

### Important (Should Fix)

**1. Config clearing without logging the cleared values** - `src/controllers/plg/plg-onboarding.js:1129-1130`

When `deliveryConfig` and `hlxConfig` are set to null, the log message at line 1133 only records the delivery type transition, not what configuration was destroyed. This is a one-way door: once cleared, the old config values are unrecoverable without a database backup. If `findDeliveryType` produces an incorrect result (even once, even for one site), the site's accumulated configuration is permanently lost.

This aligns with @habansal's comment on line 1129. From an incident-response perspective, if this causes unexpected behavior in production, operators have no way to audit what was cleared.

**Fix:** Log the previous config values before clearing:
```javascript
if (existingDeliveryType) {
  log.info(`Clearing stale config for site ${site.getId()}`, {
    previousDeliveryConfig: site.getDeliveryConfig(),
    previousHlxConfig: site.getHlxConfig(),
  });
  site.setDeliveryConfig(null);
  site.setHlxConfig(null);
}
```

**2. No error handling around `findDeliveryType` in Step 5a** - `src/controllers/plg/plg-onboarding.js:1121`

If `findDeliveryType(baseURL)` throws (network timeout, DNS failure, unexpected response shape), it bubbles up and fails the entire onboarding flow. This call is now unconditional for every PLG onboarding without a preset type, making the blast radius larger than before. A transient failure in delivery type detection should not block the entire onboarding - the site already has a type (or null for new sites), and detection is a correction mechanism, not a prerequisite.

**Fix:** Wrap in try/catch that logs and continues:
```javascript
let detectedDeliveryType;
try {
  detectedDeliveryType = await findDeliveryType(baseURL);
} catch (e) {
  log.warn(`Failed to detect delivery type for ${baseURL}: ${e.message}`);
}
```

**3. Missing test assertion for config clearing** - `test/controllers/plg/plg-onboarding.test.js:3373-3389`

The test `'calls findDeliveryType for existing site and updates delivery type when detected type differs'` asserts that `setDeliveryType` was called with the new value but does not assert that `setDeliveryConfig(null)` and `setHlxConfig(null)` were also called. This is the highest-risk behavior in the PR (destructive config wipe on existing sites), and it is currently unverified by tests. A regression that removes the clearing logic would pass silently.

**Fix:** Add assertions:
```javascript
expect(existingSite.setDeliveryConfig).to.have.been.calledWith(null);
expect(existingSite.setHlxConfig).to.have.been.calledWith(null);
```

### Minor (Nice to Have)

**4. Test comment references wrong step** - `test/controllers/plg/plg-onboarding.test.js:3367`

The comment says "once in Step 5b.5 to validate delivery type" but the source code labels the block as "Step 5a". This will confuse anyone tracing test assertions back to the implementation.

## Recommendations

- Add a test for the no-op path: when `findDeliveryType` returns the same value as `existingDeliveryType`, assert that `setDeliveryType` is NOT called. This documents the contract and protects against regressions where the inequality condition is accidentally removed.
- Add a test for the new-site path: when `existingDeliveryType` is falsy and `findDeliveryType` returns a valid type, the delivery type should be set but `setDeliveryConfig(null)` and `setHlxConfig(null)` should NOT be called (the `if (existingDeliveryType)` guard prevents it).

## Assessment

Ready to merge? With fixes

The delivery type auto-correction is architecturally sound in intent and the frescopa rejection follows established patterns cleanly. The three Important items are about production resilience: the missing config-value logging creates a recoverability gap, the unhandled `findDeliveryType` rejection puts a correction mechanism on the critical path without a safety net, and the missing test assertion leaves the highest-risk behavior unverified. All three are straightforward to address.

## Next Steps

1. Add logging of cleared config values before nulling them (addresses both this review and @habansal's inline feedback).
2. Wrap `findDeliveryType` in a try/catch so detection failures do not block onboarding.
3. Add test assertions for the config-clearing behavior.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 44s | <b>Cost:</b> $2.92 | <b>Commit:</b> <code>3816f0ba812d348bcdab578de915373bda6ea94d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/controllers/plg/plg-onboarding.js
anshulk-public: preset delivery type is the delivery type we can recieve from ESEs from Backoffice if the onboarding stuck in "Domain is not an AEM site" issue.
Screenshot 2026-06-02 at 2 55 20 PM

src/controllers/plg/plg-onboarding.js
anshulk-public: Yes, If there is creation of a new site, then it will be already updated. I will update the code for this.

MysticatBot (commented):
Hey @anshulk-public,

**Verdict: Comment** - clean rework that addresses all prior concerns; a few minor improvements noted below.
_Changes: Rejects frescopa domains early in PLG onboarding and adds a Slack alert (no site mutation) when detected delivery type differs from stored type (3 files)._

<details>

- nit: `src/controllers/plg/plg-onboarding/onboarding-flow.js:660` - consider logging when `channelId` or `token` is falsy so operators can distinguish "no mismatch" from "mismatch detected but alert skipped due to missing config"
- nit: `src/controllers/plg/plg-onboarding/onboarding-flow.js:647` - `findDeliveryType(baseURL)` is called a second time for existing sites that already resolved it in Step 2; consider reusing the cached result to avoid the extra network round-trip
- nit: `src/controllers/plg/plg-onboarding.js:295` - `domain.toLowerCase().includes('frescopa')` matches any domain containing the substring (e.g. `notfrescopa.com`); consider matching on domain-segment boundaries if only `*.frescopa.*` domains are intended
- suggestion: `src/controllers/plg/plg-onboarding/onboarding-flow.js:643` - extend the Step 5a comment to document why `presetDeliveryType` also skips the check (admin explicitly overrode delivery type)
- suggestion: document `SLACK_DELIVERY_TYPE_ALERT_CHANNEL_ID` in deployment config or default to `SLACK_PLG_ONBOARDING_CHANNEL_ID` so the feature is operational without a separate infra change
</details>

- Config clearing without logging cleared values - no longer applicable (alert-only approach, no mutation)
- Missing error handling on findDeliveryType - now wrapped in try/catch with log.warn
- Missing test assertion for config clearing - no longer applicable; tests now verify site is NOT mutated

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 0s | <b>Cost:</b> $3.93 | <b>Commit:</b> <code>1be012f379957a89fd72de24615afa1429723908</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @anshulk-public,

**Verdict: Approve** - all prior blocking concerns addressed; alert-only approach is sound.
_Changes: Rejects frescopa domains early in PLG onboarding and adds a Slack alert (no site mutation) when detected delivery type differs from stored type (3 files)._

Note: CI build is currently pending - ensure it passes before merge. The non-blocking nits from the prior review (frescopa substring broadness, redundant `findDeliveryType` call for RUM-fail path) are still applicable if you want to address them in a follow-up.

- Config clearing without logging cleared values - no longer applicable (alert-only approach, no mutation)
- Missing error handling on findDeliveryType - now wrapped in try/catch with log.warn
- Missing test assertion for config clearing - no longer applicable; tests verify site is NOT mutated

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 28s | <b>Cost:</b> $3.35 | <b>Commit:</b> <code>db7eb56f050e9fd4b58e1fe4e0afaa4dac5be3fe</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/controllers/plg/plg-onboarding.js
anshulk-public: Now we are not updating the delivery type and delivery config, we will see some intial cases through slack alert and udpate them manually if needed as we are not 100% sure about the delivery type detected through Ravi AEM detection script.

src/controllers/plg/plg-onboarding/onboarding-flow.js
```diff
@@ -637,6 +637,43 @@ export async function performAsoPlgOnboarding({
...
+      } catch (e) {
+        log.warn(`Failed to detect delivery type for ${baseURL}: ${e.message}`);
+      }
+      if (
+        detectedDeliveryType
+        && detectedDeliveryType !== SiteModel.DELIVERY_TYPES.OTHER
+        && detectedDeliveryType !== existingDeliveryType
+      ) {
+        log.warn(`Delivery type mismatch for site ${site.getId()} (${baseURL}): stored=${existingDeliveryType} detected=${detectedDeliveryType}`);
+        const channelId = env.SLACK_DELIVERY_TYPE_ALERT_CHANNEL_ID;
```
Kanishkavijay39: Do you think we should create a separate channel for this? My only concern is that having too many channels can also become a problem.

Alternatively, we could add a single line in the PLG onboarding message mentioning the incorrect configuration and how to fix it. That might be sufficient as well.

I'm okay with either approach. What are your thoughts?

Kanishkavijay39 (approved):
LGTM

src/controllers/plg/plg-onboarding/onboarding-flow.js
habansal: I agree this should just be added to the onboarding alert that we already have

PR #2527, opened 2026-06-01
fix: force download on agentic traffic export URLs | LLMO-4888

calvarezg:
- [x] Related issue linked below (LLMO-4888)
- [x] Fix commit includes the issue reference for release notes

- N/A — no API contract change; presigned URL generation is internal.

- N/A — `ResponseContentDisposition` / `ResponseContentType` are `Response*` override parameters baked into the presigned URL query string. The stored S3 object is unchanged.

## Related Issues

LLMO-4888: https://jira.corp.adobe.com/browse/LLMO-4888

Chrome opens agentic traffic export CSVs inline as text instead of downloading them. Firefox and Safari handle it correctly. Root cause: presigned S3 URLs for `spacecat-prod-reports` (no CORS) lacked `Content-Disposition: attachment`, so Chrome renders the response inline.

Adds `ResponseContentDisposition: 'attachment; filename="urls.csv"'` and `ResponseContentType: 'text/csv; charset=utf-8'` to `GetObjectCommand` in `buildExportReadyResponse`. For multi-part exports the filename becomes `urls_part1.csv`, `urls_part2.csv`, etc. No import changes needed — these are first-class `GetObjectCommand` options.

Thanks for contributing!

### DIFF
--- a/src/controllers/llmo/llmo-agentic-traffic.js
+++ b/src/controllers/llmo/llmo-agentic-traffic.js
@@ -769,7 +769,11 @@ export function createAgenticTrafficUrlsExportHandler(getSiteAndValidateAccess) 
           // Fast path: sign worker-written files[] directly. Validated first.
           if (metadata?.status === 'success'
             && Array.isArray(metadata.files) && metadata.files.length > 0) {
-            const safeFiles = validateFilesAgainstPrefix(metadata.files, siteId, exportId);
+            const safeFiles = validateFilesAgainstPrefix(metadata.files, siteId, exportId)
+              .sort((a, b) => {
+                const partNum = (k) => Number(k.match(PART_SUFFIX_PATTERN)?.[1] ?? 0);
+                return partNum(a) - partNum(b);
+              });
             if (safeFiles.length === metadata.files.length) {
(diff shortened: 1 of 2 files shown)


calvarezg:
## Self-review (fresh-context)

**[Correctness]** In the fast-path (metadata.files), `safeFiles` is the filtered `metadata.files` array with no sort applied, so array index `i` may not match the intended part number. `listExportCsvObjects` sorts by part number before returning, but `validateFilesAgainstPrefix` only filters — it does not sort. If the worker writes `metadata.files` in a different order, `urls_part1.csv` in `Content-Disposition` could be attached to the wrong S3 key. `src/controllers/llmo/llmo-agentic-traffic.js:772`. Fix: sort `safeFiles` by part number (same comparator used in `listExportCsvObjects`) before passing to `buildExportReadyResponse`.

**[Test gap]** No test asserts that `GetObjectCommand` is constructed with `ResponseContentDisposition` and `ResponseContentType`. The existing multi-part test at line 1687 only checks `downloadUrls` values; a future refactor could silently drop the response-override params. Consider adding a check on `ctx.s3.GetObjectCommand` call args (accessible via the constructor stub) that the new fields are present and correctly valued for both the single-file and multi-part cases.

calvarezg:
Not required

PR #2528, opened 2026-06-01
feat(prompts): add GET /v2/orgs/:spaceCatId/brands/:brandId/prompts/stats endpoint

nsdere:
## Summary

Adds a new aggregate endpoint that returns prompt counts per brand for the `project-elmo-ui` Prompts Management tab status bar.

- `GET /v2/orgs/:spaceCatId/brands/:brandId/prompts/stats`
- Returns `{ branded, unbranded, intents: { informational, instructional, comparative, transactional, planning, delegation } }`
- Backed by the `rpc_brand_prompt_stats` PostgREST RPC in `mysticat-data-service` - all classification logic stays in SQL, no write-path changes or backfill required on the API side

## Design note

This replaces the originally-spec'd approach of storing a derived `branded` column on `prompts` and reclassifying inside brand-mutation transactions. The RPC computes counts at read time using the current brand name + aliases, which removes the consistency model complexity (transactional reclassify, 250K row-count gate, backfill script) at the cost of slightly slower reads. Partial index on `(organization_id, brand_id, status) WHERE status = 'active'` keeps the query fast.

Intent counts read 0 until the DRS to Postgres ingestion workstream forwards `intent` on prompts upsert (separate, prerequisite workstream).

## Test plan

- [ ] Unit tests for `getPromptStats` (RPC happy path, error path, missing intents hydration)
- [ ] Unit tests for `getPromptStatsByBrand` controller (auth, org not found, brand not found, postgrest unavailable)
- [ ] Integration test in `test/it/postgres/`
- [ ] Manual smoke against staging: `curl /v2/orgs/<orgId>/brands/<brandId>/prompts/stats` returns flat shape with all 6 intent keys

Generated with [Claude Code](https://claude.com/claude-code)


MysticatBot (commented):
Hey @nsdere,

## Strengths

- **Clean architectural decision**: Computing branded/unbranded counts at read time via an RPC removes write-path consistency complexity. The trade-off is well-articulated in the PR description and is the correct choice for aggregate stats that tolerate staleness.
- **Defensive response shaping** (`src/support/prompts-storage.js:832-847`): Whitelisting known intent keys, coercing values to numbers with `|| 0` fallback, and handling both array and single-object RPC responses means the API contract is stable regardless of what the SQL function returns.
- **Consistent access control and validation pattern** (`src/controllers/brands.js:660-690`): The controller follows the same guard sequence (param validation, org lookup, access check, postgrest availability check, brand resolution) as sibling prompt endpoints.
- **Correct route ordering** (`src/routes/index.js:216`): The `/prompts/stats` static segment is registered before `/prompts/:promptId`, so the router will not accidentally capture "stats" as a `:promptId` parameter.
- **Thorough storage unit tests** (`test/support/prompts-storage.test.js`): Good breadth covering null data, array response, missing intents, unknown intent filtering, and error paths.
- **Output sanitization**: The hardcoded allowlist of intent keys prevents unexpected RPC keys from leaking to API consumers.

## Issues

### Minor (Nice to Have)

**1. Missing `spaceCatId` in error log context** - `src/controllers/brands.js:707`

```js
log.error('Error fetching prompt stats', { brandId, error });
```

Sibling handlers in this file typically include `spaceCatId` in error context. Without it, correlating a production error to a specific org requires tracing upstream. Adding `spaceCatId` to the log payload improves operational debugging.

**2. Partial assertion in happy-path controller test** - `test/controllers/brands.test.js`

```js
expect(body).to.have.property('intents').that.includes({ informational: 410, delegation: 15 });
```

Chai's `.includes()` on an object performs a partial deep-equal, checking only 2 of 6 intent keys. If the wiring accidentally zeroed out four intents, this assertion would still pass. A `.deep.equal(STATS_RESPONSE)` on the full body would be more rigorous - the storage unit tests already prove correctness, but the controller test should confirm passthrough fidelity.

**3. Route ordering is load-bearing but undocumented** - `src/routes/index.js:216`

The `/prompts/stats` route must precede `/prompts/:promptId` or the framework captures "stats" as a promptId. The current placement is correct, but there is no comment indicating this constraint. If someone later sorts the route table alphabetically, the endpoint breaks silently. Consider a brief inline comment.

## Recommendations

- The PR's own test plan mentions an integration test in `test/it/postgres/` (still unchecked). Worth completing before or shortly after merge to exercise the real PostgREST RPC contract.
- Consider a brief JSDoc on the `getPromptStats` export documenting the return shape (`{ branded: number, unbranded: number, intents: Record<Intent, number> }`). Saves a future reader from tracing the full function body.

## Assessment

Ready to merge? Yes

The change is well-scoped, follows established patterns, has solid test coverage (codecov confirms all modified lines covered), and the architectural decision to compute stats at read-time via RPC is sound. No blocking issues. The minor findings above are polish items that could land as a follow-up commit on this PR.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 18s | <b>Cost:</b> $3.11 | <b>Commit:</b> <code>52223e2b12f8e5e3a5809bd944f2cc4074a6eeee</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/routes/index.js
```diff
@@ -213,6 +213,7 @@ export default function getRouteHandlers(
     'GET /v2/orgs/:spaceCatId/brands/:brandId/serenity/tags': serenityController.listTags,
     'GET /v2/orgs/:spaceCatId/brands/:brandId/serenity/models': serenityController.listModels,
     'GET /v2/orgs/:spaceCatId/brands/:brandId/prompts': brandsController.listPromptsByBrand,
+    'GET /v2/orgs/:spaceCatId/brands/:brandId/prompts/stats': brandsController.getPromptStatsByBrand,
```
dzehnder, line 216: **Important (item 2): route-ordering dependency is untested.**

Both `/prompts/stats` and `/prompts/:promptId` are dynamic routes with the same segment count. Tracing `src/utils/route-utils.js`, the matcher iterates `Object.entries(dynamicRoutes)` and returns the first segment-count match; literal `stats` vs `:promptId` are not differentiated. The route only resolves correctly because it appears here first.

If anyone alphabetizes, regroups, or moves this line in a later refactor, `GET .../prompts/stats` will silently dispatch to `getPromptByBrandAndId` with `promptId="stats"` and return a misleading 404. The existing presence-only assertion in `test/routes/index.test.js` does not catch this.

Add a unit test that resolves the stats path through the actual matcher and asserts the returned handler is `getPromptStatsByBrand`. Two lines, future-proof.

src/support/prompts-storage.js
```diff
@@ -815,3 +815,36 @@ export async function checkPromptsExist({ brandUuid, prompts, postgrestClient })
...
+  const { data, error } = await postgrestClient.rpc('rpc_brand_prompt_stats', {
+    p_organization_id: organizationId,
+    p_brand_id: brandUuid,
+  });
+
+  if (error) {
+    throw new Error(`getPromptStats RPC failed: ${error.message}`);
+  }
+
+  const row = Array.isArray(data) && data.length > 0 ? data[0] : (data ?? {});
```
dzehnder, line 835: **Important (item 3): empty-array RPC response is an untested branch.**

`Array.isArray(data) && data.length > 0 ? data[0] : (data ?? {})` does not behave the way a casual reader expects when `data === []`:

- `Array.isArray([]) && [].length > 0` is false
- `[] ?? {}` is `[]` (nullish coalescing does not treat `[]` as nullish)
- So `row` becomes an array, not a plain object

The output happens to be correct here (all-zero shape via `Number(undefined) || 0`), but any future code that does `row.someField` assuming `row` is an object misbehaves silently. PostgREST commonly returns empty arrays for set-returning RPCs with no rows, so this is realistic.

Fix: add a test asserting the all-zero shape when `data: []`. Optional but recommended: tighten the fallback to `Array.isArray(data) ? (data[0] ?? {}) : (data ?? {})` so `row` is always an object.

src/support/prompts-storage.js
```diff
@@ -815,3 +815,36 @@ export async function checkPromptsExist({ brandUuid, prompts, postgrestClient })
...
+  if (row.intents && typeof row.intents === 'object') {
+    for (const [k, v] of Object.entries(row.intents)) {
+      if (INTENT_VALUES.includes(k)) {
+        intents[k] = Number(v) || 0;
+      }
+    }
+  }
+
+  return {
+    branded: Number(row.branded) || 0,
```
dzehnder, line 846: **Minor (item 5): stringified bigint coverage gap.**

PostgREST returns `bigint` columns as strings by default in v9+ (`branded: "42"`). `Number(row.branded) || 0` handles this correctly, but the contract is only implicit: existing tests pass numeric values.

A single storage test with `{ branded: "42", unbranded: "1208", intents: { informational: "410" } }` asserting numeric output would lock the contract against regression if anyone replaces `Number(...) || 0` with `?? 0`.

src/controllers/brands.js
```diff
@@ -657,6 +658,53 @@ function BrandsController(ctx, log, env) {
...
+
+      const stats = await getPromptStats({
+        organizationId: spaceCatId,
+        brandUuid,
+        postgrestClient,
+      });
+
+      return ok(stats);
+    } catch (error) {
+      log.error('Error fetching prompt stats', { brandId, error });
```
dzehnder, line 703: **Minor (item 7): missing org context plus user-influenced log field.**

`log.error('Error fetching prompt stats', { brandId, error })` omits `spaceCatId`, which the prior reviewer already flagged. Worth combining with one related observation: `brandId` accepts brand-name slugs (see `resolveBrandUuid` in `prompts-storage.js`), so this field is user-influenced text in structured logs.

Not a vulnerability because structured logging escapes correctly, but the one-line fix `{ spaceCatId, brandId, error }` covers both the org-context gap and the awareness of where this value comes from. Update the matching test assertion to include `spaceCatId`.

dzehnder (commented):
Hey @nsdere,

Following up on the prior review. The implementation itself is clean and mirrors sibling patterns well, but a few contract-and-test gaps surfaced on a closer read that are worth addressing before merge.

## Strengths

- Controller (`src/controllers/brands.js:661-707`) mirrors `listPromptsByBrand` and `getPromptByBrandAndId` line-for-line on validation, auth, and postgrest-availability ordering. New team members will recognize the shape immediately.
- Storage helper (`src/support/prompts-storage.js:821-849`) is small, pure, and well-isolated. Both single-object and array RPC envelopes are handled without leaking that nuance to the controller.
- Allowlist-filtered, type-coerced response shape: `INTENT_VALUES`-driven `Object.fromEntries` produces a stable schema-aligned response even when the RPC returns null/empty or unknown future keys.
- Capability mapping (`src/routes/required-capabilities.js:260`) correctly maps to `organization:read`, consistent with sibling read endpoints. No capability drift.
- Horizontal escalation is mechanically prevented at the resolver layer: `resolveBrandUuid(spaceCatId, brandId, ...)` scopes every brand lookup with `organization_id`, so even probing another org's brand UUID hits the `notFound` branch unless `hasAccess` has already granted access to that org.
- Storage unit tests cover the unusual envelope cases (null data, error envelope, array vs single-object, missing intents, unknown-key filtering) that protect against future RPC drift.

## Issues

### Important (Should Fix)

**1. New endpoint is missing from the OpenAPI spec.** `docs/openapi/api.yaml` lines 140-145 register `/prompts`, `/prompts/{promptId}`, and `/prompts/delete` against fragments in `docs/openapi/prompts-v2-api.yaml`. The new `/prompts/stats` path appears in neither file. project-elmo-ui (the documented consumer) loses the contract to code against, and generated client SDKs will not see the route at all. The response shape, including the intent enum, also effectively lives in `INTENT_VALUES` rather than a published contract. Fix: add a `v2-prompts-by-brand-stats` fragment in `prompts-v2-api.yaml` matching the existing sibling fragments (path params, 200/400/403/404/503 responses, schema) and wire it under `paths:` in `api.yaml`. Same PR.

**2. Route-ordering dependency is real and untested.** `src/routes/index.js:216` places `/prompts/stats` before `/prompts/:promptId`. Tracing `src/utils/route-utils.js`, the matcher iterates `Object.entries(dynamicRoutes)` and returns the first segment-count match; literal `stats` vs `:promptId` are not differentiated, so the route only wins because of insertion order. If anyone alphabetizes, regroups, or moves the line in a later refactor, `GET .../prompts/stats` silently dispatches to `getPromptByBrandAndId` with `promptId="stats"` and returns a misleading 404. The existing route test at `test/routes/index.test.js:743` asserts presence via map lookup but says nothing about which handler answers a specific path. Fix: add a unit test that resolves the stats path through the actual matcher and asserts the returned handler is `getPromptStatsByBrand`. Two lines, future-proof. Prior reviewer flagged the missing comment as Minor; the structural gap is the missing test, which is what makes the invariant survive future refactors.

**3. Empty-array RPC response is an untested branch with a load-bearing JS quirk.** In `src/support/prompts-storage.js:842`, the fallback `Array.isArray(data) && data.length > 0 ? data[0] : (data ?? {})` does not behave the way a casual reader expects when `data === []`. `Array.isArray([]) && [].length > 0` is false, then `[] ?? {}` is `[]` (nullish coalescing does not treat `[]` as nullish), so `row` becomes an array. The output happens to be correct (all-zero shape via `Number(undefined) || 0`), but the path is fragile: any future code that does `row.someField` on the assumption `row` is a plain object misbehaves silently. PostgREST often returns empty arrays for RPCs returning a set with no rows, so this is a realistic case. Fix: add a test asserting the all-zero shape when `data: []`. Optional but recommended: tighten the fallback to always normalize to an object, e.g., `Array.isArray(data) ? (data[0] ?? {}) : (data ?? {})`.

**4. The 404-brand-not-found controller test does not assert the RPC was skipped.** `test/controllers/brands.test.js` (the `returns 404 when brand is not found` case) stubs `rpc` to resolve with `STATS_RESPONSE`. If a future refactor accidentally calls `getPromptStats` before checking `brandUuid`, the test still passes. Add `expect(mockDataAccess.services.postgrestClient.rpc).to.not.have.been.called` to that case. Same negative assertion would harden the 403 and 503 cases too, though those are recommendations rather than blockers.

### Minor (Nice to Have)

**5. Storage tests do not cover stringified bigints from PostgREST.** PostgREST returns `bigint` columns as strings by default in v9+ (`branded: "42"`). The code handles this correctly via `Number(row.branded) || 0`, but the contract is only validated implicitly: existing tests pass numeric values. A single test with `{ branded: "42", unbranded: "1208", intents: { informational: "410" } }` asserting numeric output would lock the contract against regression if anyone later replaces `Number(...) || 0` with `?? 0`.

**6. The 500-path controller test pins to two failure surfaces ambiguously.** `test/controllers/brands.test.js` (the `returns 500 when storage throws` case) makes `rpc` reject, but the controller flow has two PostgREST surfaces: `from(...)` for `resolveBrandUuid`, then `rpc(...)` for `getPromptStats`. The test relies on the fact that `resolveBrandUuid` uses `from`, not `rpc`. If anyone refactors `resolveBrandUuid` to use an rpc, this test starts failing on the wrong stage and the log assertion becomes ambiguous. Add `expect(postgrestClient.rpc).to.have.been.calledOnce` to pin the failure to the right call site.

**7. Error log omits org context, and the included `brandId` is user-influenced text.** `src/controllers/brands.js:707` writes `log.error('Error fetching prompt stats', { brandId, error })`. The prior reviewer flagged the missing `spaceCatId`; combine that fix with the observation that `brandId` accepts brand-name slugs per `resolveBrandUuid` (so it is user-influenced text in structured log fields). Not a vulnerability since structured logging escapes correctly, but the one-line fix `{ spaceCatId, brandId, error }` covers both concerns in the same edit. Update the matching test assertion to include `spaceCatId`.

## Recommendations

- Acknowledging the prior reviewer's third Minor finding: the happy-path partial assertion `expect(body).to.have.property('intents').that.includes({ informational: 410, delegation: 15 })` in `test/controllers/brands.test.js` should be `.deep.equal(STATS_RESPONSE)`. The shape-stability contract is the whole point of the `INTENT_VALUES` filter; if you do not assert the full shape, a regression that drops `comparative` would not be caught.
- The intents block returns all zeros until the DRS to Postgres ingestion workstream lands. The PR description acknowledges this, but consumers (project-elmo-ui) will render "0 of everything" with no signal that the field is degraded vs genuinely empty. One line of contract documentation in the OpenAPI fragment (item 1) covers this naturally.
- Consider importing `INTENT_VALUES` into the test file rather than redeclaring the keys in `ALL_ZERO_INTENTS`. If a seventh intent is ever added, both currently must update independently.
- PostgREST error envelopes can include `details` and `hint` fields that occasionally contain SQL or query fragments. In `src/support/prompts-storage.js:835`, `getPromptStats RPC failed: ${error.message}` is generally safe, but be aware that `error` itself, logged downstream in the controller catch, can leak that metadata. Worth a follow-up to use a structured error mapper for the v2 RPC family.
- `INTENT_VALUES` will likely become a shared contract with the OpenAPI fragment, the RPC, and the DRS ingestion path. No need to hoist it now (YAGNI), but a shared constants module is the natural home once a second consumer in this repo references it.

- The `rpc_brand_prompt_stats` SQL implementation and the partial index on `(organization_id, brand_id, status) WHERE status = 'active'` live in `mysticat-data-service`. Confirm the index exists in the migration before this ships to prod; not reviewable here.
- Intent taxonomy is now embedded across three places (the RPC, this API filter, the DRS ingestion path). A single source of truth becomes worth the effort once the enum grows beyond six values.
- Integration test in `test/it/postgres/` was mentioned in the PR test plan but not added. The RPC contract (column names, return shape, bigint string vs number) is asserted nowhere in this repo. A follow-up integration test would catch RPC drift.

## Assessment

Ready to merge? With fixes.

Reasoning: The implementation follows established patterns cleanly and the architectural simplification (read-time aggregation over derived-column maintenance) is the right call. The four Important findings are all small additive commits within this PR's scope, and they protect contracts this PR is establishing: the public response shape (OpenAPI), the dispatch behavior of a segment-count-ambiguous route (test), and two test cases that would silently miss real regressions.

## Next Steps

1. Add OpenAPI fragment and path entry for `/prompts/stats` (Important item 1).
2. Add a matcher-level test pinning `/prompts/stats` to `getPromptStatsByBrand` (Important item 2).
3. Add an empty-array RPC test and harden the 404 test with a negative RPC assertion (Important items 3 and 4).
4. Roll the four Minor and the prior reviewer's happy-path assertion fix into the same commit.

dzehnder (approved):

PR #2531, opened 2026-06-01
feat(routes): expose consent-banner routes to S2S consumers via organization capabilities

ravverma:
Move POST /consent-banner and GET /consent-banner/:jobId from INTERNAL_ROUTES into routeRequiredCapabilities, gated by organization:write and organization:read respectively, so S2S consumers can be granted access.

### DIFF
--- a/src/routes/required-capabilities.js
+++ b/src/routes/required-capabilities.js
@@ -57,10 +57,6 @@ export const INTERNAL_ROUTES = [
   'POST /slack/events',
   'POST /slack/channels/invite-by-user-id',
 
-  // Consent banner - screenshot tooling, end-user/internal use only
-  'POST /consent-banner',
-  'GET /consent-banner/:jobId',
-
   // Brand Presence stats - org-scoped, LLMO product; not yet required by S2S consumers
   'GET /org/:spaceCatId/brands/all/brand-presence/stats',
   'GET /org/:spaceCatId/brands/:brandId/brand-presence/stats',
@@ -220,6 +216,10 @@ export const INTERNAL_ROUTES = [
 const routeRequiredCapabilities = {
   // Audits
   'GET /audits/latest/:auditType': 'audit:read',
+
+  // Consent Banner
+  'POST /consent-banner': 'organization:write',
+  'GET /consent-banner/:jobId': 'organization:read',
 
   // Configuration
   'GET /configurations/latest': 'configuration:read',


MysticatBot (changes requested):
Hey @ravverma,

## Strengths

- `src/routes/required-capabilities.js`: Surgically scoped change - two routes removed from one list, added to another, nothing else touched. Good change discipline.
- The move from INTERNAL_ROUTES (broad admin/IMS access) to capability-gated routes strictly tightens access control. S2S consumers must now present a scoped API key with an explicit grant.
- Correct read/write separation: POST (mutating) requires `organization:write`, GET (read-only) requires `organization:read`, consistent with the convention used throughout this file.
- The existing test infrastructure (`required-capabilities.test.js`) already validates that every registered route is categorized, the two lists are disjoint, and every capability value follows the `entity:action` format - so this change is covered by structural tests automatically.

## Issues

### Important (Should Fix)

**`src/routes/required-capabilities.js:220-221` - Capability choice creates an implicit cross-cutting grant with no org-scoping in the URL**

Every other route gated by `organization:read` or `organization:write` has an org identifier in the URL path (`/organizations/:organizationId`, `/org/:spaceCatId`, `/v2/orgs/:spaceCatId/...`). The consent-banner routes have no such scoping (`POST /consent-banner`, `GET /consent-banner/:jobId`). This means any S2S consumer granted `organization:write` for legitimate org-management use cases will also gain access to `POST /consent-banner` as an unintended side effect.

This is not a security vulnerability (the change is strictly an improvement over the INTERNAL_ROUTES state), but it degrades the authorization model: the capability acts as a coarse-grained "role" here rather than a resource permission. If a consumer needs only org management and not consent-banner, there is no way to grant one without the other.

**How to fix (pick one):**
1. Introduce dedicated `consent-banner:write` and `consent-banner:read` capabilities to keep grants narrow - this follows the principle of least privilege demonstrated by the rest of the capabilities map.
2. If consent-banner genuinely requires the caller to hold org-level authority (because it operates on org data), add a brief comment explaining the relationship so future maintainers understand the coupling is intentional.

## Assessment
Ready to merge? With fixes
Reasoning: The migration pattern is correct and the change tightens security. The capability choice deserves a brief justification - either a dedicated capability or a code comment explaining why `organization:*` is appropriate for a non-org-scoped path.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 43s | <b>Cost:</b> $2.53 | <b>Commit:</b> <code>d78201a897608b21cf0da29fa87bd5be693d86bf</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ravverma:
@MysticatBot 


The `organization:*` capability choice is intentional and correct here. `consent-banner` is not a domain entity in SpaceCat — there is no stored consent-banner resource. `ConsentBannerController` is purely a scraping utility that creates ephemeral scrape jobs via `ScrapeClient` and returns S3-presigned screenshot URLs.

Every capability in this file follows the `entity:action` convention where the entity is a real data model type (`site`, `organization`, `audit`, `configuration`, etc.). Introducing `consent-banner:read/write` would be the only capability without a corresponding domain entity, which is inconsistent with the established pattern.

`organization:*` is the correct authority level because the consent-banner tool is part of the org-level operator workflow — the S2S consumers being granted org capabilities are exactly the callers who need this tooling.

Regarding Option 2 (add a comment): this is already satisfied by the `// Consent Banner` section header in the file.

PR #2532, opened 2026-06-01
fix(prompts): transform flat intent_* RPC fields into nested intents object

nsdere:
## Summary

Follow-up fix to #2528. The `rpc_brand_prompt_stats` RPC returns flat `intent_informational`, `intent_instructional` etc. fields, but the endpoint was returning them as-is instead of transforming into the spec's nested `intents` shape.

**Before:**
```json
{ "branded": 2129, "unbranded": 57029, "intent_informational": 0, ... }
```

**After:**
```json
{
  "branded": 2129,
  "unbranded": 57029,
  "intents": { "informational": 0, "instructional": 0, ... }
}
```

Also updates tests to mock the actual RPC flat-field shape and assert the correct nested output.

## Test plan

- [x] Unit tests pass locally (360 passing)
- [x] Lint clean

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -832,15 +832,10 @@ export async function getPromptStats({ organizationId, brandUuid, postgrestClien
     throw new Error(`getPromptStats RPC failed: ${error.message}`);
   }
 
-  const row = Array.isArray(data) && data.length > 0 ? data[0] : (data ?? {});
-  const intents = Object.fromEntries(INTENT_VALUES.map((k) => [k, 0]));
-  if (row.intents && typeof row.intents === 'object') {
-    for (const [k, v] of Object.entries(row.intents)) {
-      if (INTENT_VALUES.includes(k)) {
-        intents[k] = Number(v) || 0;
-      }
-    }
-  }
+  const row = Array.isArray(data) ? (data[0] ?? {}) : (data ?? {});
+  const intents = Object.fromEntries(
+    INTENT_VALUES.map((k) => [k, Number(row[`intent_${k}`]) || 0]),
+  );
 
   return {
     branded: Number(row.branded) || 0,
--- a/test/controllers/brands.test.js
+++ b/test/controllers/brands.test.js
@@ -2062,17 +2062,16 @@ describe('Brands Controller', () => {
 
   describe('getPromptStatsByBrand', () => {
     const BRAND_UUID = 'd1111111-1111-4111-b111-111111111111';
-    const STATS_RESPONSE = {
+    // RPC returns flat intent_* fields; the storage layer transforms them to nested intents
+    const STATS_RPC_ROW = {
       branded: 42,
       unbranded: 1208,
-      intents: {
-        informational: 410,
-        instructional: 180,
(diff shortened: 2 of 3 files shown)


MysticatBot (approved):
Hey @nsdere,

## Strengths

- **src/support/prompts-storage.js:834-836** - The new implementation is significantly more concise. Mapping directly from `intent_${k}` fields to the nested shape in a single `Object.fromEntries` call replaces a nested conditional + loop that was harder to reason about. Net reduction of 10 lines to 4.
- **src/support/prompts-storage.js:834** - The simplified array check (`Array.isArray(data) ? (data[0] ?? {}) : (data ?? {})`) correctly drops the redundant `data.length > 0` guard since `data[0] ?? {}` handles the empty-array case via nullish coalescing.
- **test/support/prompts-storage.test.js** - Tests now mock the actual RPC flat-field shape rather than the desired output shape. This means the tests genuinely exercise the transformation logic rather than tautologically asserting the mock back. Correct testing pattern for a data-mapping layer.
- **test/support/prompts-storage.test.js (new tests)** - Good defensive additions: empty-array RPC response and stringified bigint coercion cover real PostgREST edge cases the old tests missed.
- **test/controllers/brands.test.js:2184** - Switching from `.that.includes()` to `.to.deep.include()` fixes a subtle issue where the old chai assertion was not doing deep comparison on objects, making it weaker than intended.
- **Security** - The transformation reads only from a controlled allowlist (`INTENT_VALUES`) via `.map()`, preventing any unexpected RPC fields from leaking into the response. Safe-by-construction.

## Issues

### Minor (Nice to Have)

**src/support/prompts-storage.js:835** - `Number(row[...]) || 0` makes "field absent", "field is zero", and "field is NaN" indistinguishable in the output. This is correct for non-negative count data and is the inherited convention across the function (`Number(row.branded) || 0` uses the same pattern). No action needed - just noting that if the RPC contract ever expands to include signed deltas, this would silently mask negatives. Awareness item only.

## Recommendations

- In the "ignores unknown intent_* fields" test, consider also asserting that the returned `intents` object has exactly the expected number of keys (`expect(Object.keys(result.intents)).to.have.lengthOf(6)`). This would catch a hypothetical regression if someone switched to iterating `row` keys instead of the `INTENT_VALUES` allowlist.

## Assessment

Ready to merge? Yes

Well-scoped bug fix that correctly aligns the transformation layer with the actual RPC wire format. Tests cover the key scenarios (single object, array-wrapped, empty array, absent fields, unknown fields, string coercion), CI is green, and Codecov confirms full coverage. The change respects the existing system boundary where the RPC owns its return shape and the storage layer owns the mapping to the API contract.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 37s | <b>Cost:</b> $2.75 | <b>Commit:</b> <code>1d764503ae4f749e5f9e8bd632511e3907f8b6b8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2536, opened 2026-06-02
feat(agentic-rules): customer-edit endpoints for URL classification rules (LLMO-5037)

akshaymagapu:
## Summary

PR 3 of the 3-PR **customer-edit** feature. Adds customer-facing endpoints that let LLMO customers review, add, edit, and delete the auto-derived URL classification rules surfaced by the upstream audit-worker work.

- **8 endpoints** (4 per dimension × categories + page-types) under `/sites/{siteId}/agentic-categories/*` and `/sites/{siteId}/agentic-page-types/*`
- Customers paste **sample URLs**; the server computes the matching regex via `src/support/regex-from-urls.js` (4-strategy ladder). Power users may pass a raw regex instead.
- Direct PostgREST writes — no new RPC
- `AccessControlUtil`-based auth: `site:read` for GET, `site:write` for all mutations
- Responses are mapped through a DTO (`src/dto/agentic-rule.js`) so the API returns **camelCase** (`siteId`, `sortOrder`, `sampleUrls`, `derivationMethod`, `createdAt`, …) regardless of the snake_case DB columns

### Endpoints (per dimension)

| Method | Path | Body | Purpose |
|---|---|---|---|
| `GET` | `/sites/{siteId}/agentic-categories` | — | List rules with metadata |
| `POST` | `/sites/{siteId}/agentic-categories` | `{ name, urls[] }` *(or `{ name, regex }`)* | Create rule (`source='human'`) |
| `PATCH` | `/sites/{siteId}/agentic-categories/{name}` | `{ newName?, urls?, newRegex? }` | Edit; recomputes regex if `urls` given; marks `source='human'` |
| `DELETE` | `/sites/{siteId}/agentic-categories/{name}` | — | Remove rule |

Same shape for `/sites/{siteId}/agentic-page-types/*`.

### `regex-from-urls` strategy ladder
Returns the first strategy whose regex matches **every** input URL, plus an `evidence` string (logged once per create/update). All comparisons and emitted regexes are case-insensitive (`(?i)` prefix), and every strategy is anchored/bounded so a rule can't accidentally match a longer sibling path:
1. **common-prefix** — longest shared prefix, snapped to a segment boundary. A prefix that ends mid-segment (no trailing `/`) is bounded with `(/|\.|$)` so `/products/foo` can't match `/products/foobar`. Locale-only roots (`/en-us/`, `/de/`) are rejected so a rule can't silently match the whole site.
2. **universal-token** — a whole path segment present in every URL; file extension stripped from the last segment only (dotted middle segments like `node.js` and versions like `v1.2` keep their dot), emitted anchored as `/<token>(/|\.|$)`. Locale-only tokens are skipped.
3. **disjoint-cover** — per-URL keyword alternation, anchored `/(a|b|c)(/|\.|$)` (no substring matches). Locale-only tokens skipped here too.
4. **literal-fallback** — escaped full-path alternation, anchored at start **and** bounded at end `^(…)(/|\.|$)` (worst case).

`validateUserRegex` (the raw-regex paste path) normalizes input to case-insensitive `(?i)` so a pasted rule behaves like a derived one; power users who need case-sensitivity can opt out with an inline `(?-i)`. Input is validated for compile + length cap (512 chars).

### Validation & error handling
- Rejects missing / empty / whitespace-only URL entries (trimmed before extraction)
- Empty `PATCH` body (and blank `newName`) → `400`
- Duplicate rule name → `409` (maps PG unique-constraint `23505`)
- Null row from an unexpected PostgREST result → `500` guard
- Oversized or uncompilable regex → `400`

### Data model
- Customer-created → `source='human'`, `derivation_method='customer'`, `sample_urls` = pasted URLs
- Any edit of an auto rule (`source='ai'`) flips it to `source='human'`
- Writes go to `agentic_url_category_rules` / `agentic_url_page_type_rules` (unique on `(site_id, name)`)

### Files
**New**
- `src/controllers/agentic-categories.js` / `agentic-page-types.js` — thin wrappers over a shared factory
- `src/controllers/agentic-rules-factory.js` — `createRulesController({ tableName, dimensionLabel })`: `list`, `create`, `update`, `remove`
- `src/dto/agentic-rule.js` — snake_case row → camelCase response mapper
- `src/support/regex-from-urls.js` — pasted-URLs → regex helper (+ `validateUserRegex`)
- `docs/openapi/agentic-rules-api.yaml` — OpenAPI spec (camelCase schema)
- Tests: `test/controllers/agentic-{categories,page-types}.test.js`, `test/support/regex-from-urls.test.js`

**Modified**
- `src/routes/index.js`, `src/routes/required-capabilities.js`, `src/index.js`, `docs/openapi/api.yaml`, `docs/index.html`, `test/routes/index.test.js`

### Predecessor
Builds on spacecat-audit-worker PR #2539, which derives the candidate rules this PR lets customers edit.

## Validation
- `npm run lint`: clean for all changed files
- `npm test`: **11271 passing**, coverage gate green (`codecov/patch` ✓)
- `npm run docs:lint && docs:build`: API description valid

## Out of scope (follow-ups)
- Structured includes/excludes/contains rule builder (compose regex from declared predicates) — better authoring UX, exclusion as a first-class concept
- Raw-regex paste-path ReDoS hardening (real mitigation is Athena query timeout / tenant isolation, not code) — the derived-regex path is ReDoS-safe by construction

## Related Issues
LLMO-5037

### DIFF
--- a/docs/openapi/agentic-rules-api.yaml
+++ b/docs/openapi/agentic-rules-api.yaml
@@ -52,7 +52,7 @@ agentic-categories-list:
           schema:
             $ref: '#/AgenticRuleCreateRequest'
     responses:
-      '200':
+      '201':
         description: Rule created
         content:
           application/json:
@@ -185,7 +185,7 @@ agentic-page-types-list:
           schema:
             $ref: '#/AgenticRuleCreateRequest'
     responses:
-      '200':
+      '201':
         description: Rule created
         content:
           application/json:
--- a/docs/openapi/agentic-rules-api.yaml
+++ b/docs/openapi/agentic-rules-api.yaml
@@ -1,0 +1,359 @@
+# Customer-edit endpoints for auto-derived URL classification rules.
+# Two parallel dimensions (categories + page-types) sharing one shape:
+#   - `agentic_url_category_rules`
+#   - `agentic_url_page_type_rules`
+# Customers paste sample URLs; the server computes a matching regex via
+# the regex-from-urls 4-strategy ladder (common-prefix → universal-token
+# → disjoint-cover → literal-fallback). See src/support/regex-from-urls.js.
+
+# ───────────────────────────── categories ─────────────────────────────
+
+agentic-categories-list:
+  parameters:
+    - $ref: './parameters.yaml#/siteId'
+  get:
+    tags:
+      - site
+      - agentic rules
+    summary: List URL category rules for a site
+    description: Returns auto-derived + customer-edited URL category rules.
+    operationId: listAgenticCategories
+    responses:
+      '200':
+        description: List of category rules
+        content:
+          application/json:
+            schema:
+              $ref: '#/AgenticRuleListResponse'
+      '400': { $ref: './responses.yaml#/400' }
+      '401': { $ref: './responses.yaml#/401' }
+      '403': { $ref: './responses.yaml#/403' }
+      '404': { $ref: './responses.yaml#/404-site-not-found-with-id' }
+      '500': { $ref: './responses.yaml#/500' }
+    security:
+      - ims_key: [ ]
+      - scoped_api_key: [ ]
+  post:
+    tags:
+      - site
+      - agentic rules
+    summary: Create a URL category rule from sample URLs
+    description: |
+      Computes a matching regex from the supplied sample URLs and persists
+      it with `source='human'`. Returns 409 if a sample URL already belongs
+      to another active rule on the same site, if the rule name collides
+      with an existing active rule, or if the site already has the maximum of
+      20 active category rules.
+    operationId: createAgenticCategory
+    requestBody:
+      required: true
+      content:
+        application/json:
+          schema:
+            $ref: '#/AgenticRuleCreateRequest'
+    responses:
+      '200':
+        description: Rule created
+        content:
+          application/json:
+            schema:
+              $ref: '#/AgenticRule'
+      '400': { $ref: './responses.yaml#/400' }
+      '401': { $ref: './responses.yaml#/401' }
+      '403': { $ref: './responses.yaml#/403' }
+      '404': { $ref: './responses.yaml#/404-site-not-found-with-id' }
+      '409': { $ref: '#/AgenticRuleConflictResponse' }
+      '500': { $ref: './responses.yaml#/500' }
+    security:
+      - ims_key: [ ]
+      - scoped_api_key: [ ]
+
+agentic-categories-one:
+  parameters:
+    - $ref: './parameters.yaml#/siteId'
+    - name: name
+      in: path
+      required: true
+      schema:
+        type: string
+      description: Rule name (unique per site).
+  patch:
+    tags:
+      - site
+      - agentic rules
+    summary: Edit a URL category rule
+    description: |
+      Edit `name`, `urls`, or `newRegex`. Any customer edit marks the rule
+      `source='human'`; `updated_at` is bumped automatically. Returns 409 if
+      a supplied sample URL already belongs to a different active rule.
+    operationId: updateAgenticCategory
+    requestBody:
+      required: true
+      content:
+        application/json:
+          schema:
+            $ref: '#/AgenticRuleUpdateRequest'
+    responses:
+      '200':
+        description: Rule updated
+        content:
+          application/json:
+            schema:
+              $ref: '#/AgenticRule'
+      '400': { $ref: './responses.yaml#/400' }
+      '401': { $ref: './responses.yaml#/401' }
+      '403': { $ref: './responses.yaml#/403' }
+      '404': { $ref: './responses.yaml#/404' }
+      '409': { $ref: '#/AgenticRuleConflictResponse' }
+      '500': { $ref: './responses.yaml#/500' }
+    security:
+      - ims_key: [ ]
+      - scoped_api_key: [ ]
+  delete:
+    tags:
+      - site
+      - agentic rules
+    summary: Delete a URL category rule
+    description: |
+      Soft delete — flips the rule's `status` to `deleted` and stamps
+      `updated_by`, retaining the row for audit. The deleted name becomes
+      free to re-create. Runtime classification and the list endpoint only
+      return active rules.
+    operationId: deleteAgenticCategory
+    responses:
+      '200':
+        description: Rule deleted
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                deleted: { type: boolean }
+                name: { type: string }
+      '400': { $ref: './responses.yaml#/400' }
+      '401': { $ref: './responses.yaml#/401' }
+      '403': { $ref: './responses.yaml#/403' }
+      '404': { $ref: './responses.yaml#/404' }
+      '500': { $ref: './responses.yaml#/500' }
+    security:
+      - ims_key: [ ]
+      - scoped_api_key: [ ]
+
+# ───────────────────────────── page-types ─────────────────────────────
+# Identical shape to categories — re-uses the same schemas.
+
+agentic-page-types-list:
+  parameters:
+    - $ref: './parameters.yaml#/siteId'
+  get:
+    tags:
+      - site
+      - agentic rules
+    summary: List URL page-type rules for a site
+    operationId: listAgenticPageTypes
+    responses:
+      '200':
+        description: List of page-type rules
+        content:
+          application/json:
+            schema:
+              $ref: '#/AgenticRuleListResponse'
+      '400': { $ref: './responses.yaml#/400' }
+      '401': { $ref: './responses.yaml#/401' }
+      '403': { $ref: './responses.yaml#/403' }
+      '404': { $ref: './responses.yaml#/404-site-not-found-with-id' }
+      '500': { $ref: './responses.yaml#/500' }
+    security:
+      - ims_key: [ ]
+      - scoped_api_key: [ ]
+  post:
+    tags:
+      - site
+      - agentic rules
+    summary: Create a URL page-type rule from sample URLs
+    description: |
+      Computes a matching regex from the supplied sample URLs and persists
+      it with `source='human'`. Returns 409 if a sample URL already belongs
+      to another active rule on the same site, if the rule name collides
+      with an existing active rule, or if the site already has the maximum of
+      20 active page-type rules.
+    operationId: createAgenticPageType
+    requestBody:
+      required: true
+      content:
+        application/json:
+          schema:
+            $ref: '#/AgenticRuleCreateRequest'
+    responses:
+      '200':
+        description: Rule created
+        content:
+          application/json:
+            schema:
+              $ref: '#/AgenticRule'
+      '400': { $ref: './responses.yaml#/400' }
+      '401': { $ref: './responses.yaml#/401' }
+      '403': { $ref: './responses.yaml#/403' }
+      '404': { $ref: './responses.yaml#/404-site-not-found-with-id' }
+      '409': { $ref: '#/AgenticRuleConflictResponse' }
+      '500': { $ref: './responses.yaml#/500' }
+    security:
+      - ims_key: [ ]
+      - scoped_api_key: [ ]
+
+agentic-page-types-one:
+  parameters:
+    - $ref: './parameters.yaml#/siteId'
+    - name: name
+      in: path
+      required: true
+      schema:
+        type: string
+      description: Rule name (unique per site).
+  patch:
+    tags:
+      - site
+      - agentic rules
+    summary: Edit a URL page-type rule
+    description: |
+      Edit `name`, `urls`, or `newRegex`. Any customer edit marks the rule
+      `source='human'`; `updated_at` is bumped automatically. Returns 409 if
+      a supplied sample URL already belongs to a different active rule.
+    operationId: updateAgenticPageType
+    requestBody:
+      required: true
+      content:
+        application/json:
+          schema:
+            $ref: '#/AgenticRuleUpdateRequest'
+    responses:
+      '200':
+        description: Rule updated
+        content:
+          application/json:
+            schema:
+              $ref: '#/AgenticRule'
+      '400': { $ref: './responses.yaml#/400' }
+      '401': { $ref: './responses.yaml#/401' }
+      '403': { $ref: './responses.yaml#/403' }
+      '404': { $ref: './responses.yaml#/404' }
+      '409': { $ref: '#/AgenticRuleConflictResponse' }
+      '500': { $ref: './responses.yaml#/500' }
+    security:
+      - ims_key: [ ]
+      - scoped_api_key: [ ]
+  delete:
+    tags:
+      - site
+      - agentic rules
+    summary: Delete a URL page-type rule
+    description: |
+      Soft delete — flips the rule's `status` to `deleted` and stamps
+      `updated_by`, retaining the row for audit. The deleted name becomes
+      free to re-create. Runtime classification and the list endpoint only
+      return active rules.
+    operationId: deleteAgenticPageType
+    responses:
+      '200':
+        description: Rule deleted
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                deleted: { type: boolean }
+                name: { type: string }
+      '400': { $ref: './responses.yaml#/400' }
+      '401': { $ref: './responses.yaml#/401' }
+      '403': { $ref: './responses.yaml#/403' }
+      '404': { $ref: './responses.yaml#/404' }
+      '500': { $ref: './responses.yaml#/500' }
+    security:
+      - ims_key: [ ]
+      - scoped_api_key: [ ]
+
+# ─────────────────────────────── schemas ──────────────────────────────
+
+AgenticRule:
+  type: object
+  description: |
+    A single URL classification rule. The same shape is used for both
+    categories and page-types.
+  properties:
+    id: { type: string, format: uuid }
+    siteId: { type: string, format: uuid }
+    name: { type: string }
+    regex: { type: string }
+    sortOrder: { type: [integer, 'null'] }
+    source:
+      type: string
+      enum: [ai, human]
+      description: Provenance — `ai` (audit-worker auto-derivation) or `human` (customer create, edit, or confirm).
+    sampleUrls:
+      type: array
+      items: { type: string }
+      description: Representative URLs that define the rule (hard cap 50, each ≤ 2048 chars).
+    derivationMethod:
+      type: [string, 'null']
+      enum: [llm, common-prefix, universal-token, disjoint-cover, literal-fallback, customer, null]
+      description: How the regex was produced. `llm` for auto-derived; customer-edit API writes `customer` (or a regex-from-urls strategy). NULL for legacy rules.
+    createdAt:
+      type: string
+      format: date-time
+    updatedAt:
+      type: string
+      format: date-time
+    createdBy:
+      type: [string, 'null']
+      description: Identifier of the user or system that first created the rule. Set once on create and never modified.
+    updatedBy:
+      type: [string, 'null']
+      description: Identifier of the user or system that last wrote the rule. Rewritten on every edit.
+
+AgenticRuleListResponse:
+  type: object
+  description: |
+    List of active rules. v1 has no cursor pagination; the result is capped at
+    1000 rules per site (well above expected cardinality). A truncated response
+    is logged server-side.
+  properties:
+    items:
+      type: array
+      items: { $ref: '#/AgenticRule' }
+
+AgenticRuleConflictResponse:
+  description: |
+    Conflict — the rule name already exists for an active rule on the site, a
+    supplied sample URL already belongs to another active rule, or the site has
+    reached the maximum of 20 active rules for this dimension.
+  content:
+    application/json:
+      schema:
+        type: object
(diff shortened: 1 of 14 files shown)


MysticatBot (changes requested):
Hey @akshaymagapu,

## Strengths

- Excellent use of the factory pattern (`src/controllers/agentic-rules-factory.js`) - both dimensions share identical CRUD logic differing only by table name and label. This eliminates duplication and will trivially accommodate future classification dimensions.
- Thorough `authorize()` function (factory:86-108) validates UUID, checks dataAccess/postgrestClient availability, loads site, and enforces org-level access control. Every failure returns an actionable error.
- The 4-strategy regex ladder (`src/support/regex-from-urls.js`) is deterministic, pure, and independently testable with no DB/network dependencies. Strong separation of concern.
- Soft-delete pattern with `status` column keeps audit trail while the partial unique index allows name re-creation - sound design for customer-facing data.
- `logAndFail()` (factory:137-150) correctly maps Postgres 23505 to a user-friendly 409 without leaking internal details.
- DTO (`src/dto/agentic-rule.js`) correctly omits `status` from the API surface, keeping soft-delete as an internal concern.
- `decodeRuleName()` (factory:43-57) correctly URL-decodes the path param with error handling - a detail many would miss.
- The empty-PATCH guard (factory:254-259) prevents silently laundering provenance from `ai` to `human` without content change. Thoughtful domain awareness.
- All 8 routes have explicit capability requirements in `required-capabilities.js` with correct `site:read`/`site:write` scoping.
- Unit tests cover every code path through the factory (auth gates, validation, PostgREST errors, URL-decoding edge cases, cross-rule dedup conflicts, soft-delete payload verification).

## Issues

### Important (Should Fix)

**1. Missing input size limits on `urls` array and `name` field**
`src/controllers/agentic-rules-factory.js:155-156` (create) and `:257` (update)

The OpenAPI spec documents "hard cap 50, each <= 2048 chars" for `sampleUrls`, but the controller only validates that `urls` is a non-empty array of non-empty strings. Similarly, `name` has no length cap. A caller can submit thousands of URLs (each megabytes long) or a 10KB rule name.

Why it matters: `findSampleUrlConflict()` loads all active rules then iterates the submitted URLs against them. `regexFromUrls()` processes every path through multiple strategies. The MAX_REGEX_LEN guard eventually rejects the result, but not before doing significant CPU work. On a customer-facing endpoint, this is a resource-exhaustion vector.

How to fix:
```javascript
const MAX_SAMPLE_URLS = 50;
const MAX_URL_LEN = 2048;
const MAX_NAME_LEN = 200;

function isValidUrlListBody(urls) {
  return isArray(urls) && urls.length > 0
    && urls.length <= MAX_SAMPLE_URLS
    && urls.every((u) => hasText(u) && u.length <= MAX_URL_LEN);
}
// Add in create/update: if (body.name.length > MAX_NAME_LEN) return badRequest(...)
```

**2. `findSampleUrlConflict` fetches all active rules for client-side dedup**
`src/controllers/agentic-rules-factory.js:52-71`

The conflict check issues `SELECT name, sample_urls WHERE site_id=X AND status='active'` without a LIMIT, pulling every active rule for the site into the Lambda and iterating all their `sample_urls` arrays into a Map. This runs on every create/update call.

Why it matters: When the upstream audit-worker populates these tables with hundreds of auto-derived rules per site, this becomes an O(rules * urls_per_rule) in-memory operation per request. No error, just increasing latency and memory pressure until it hits Lambda limits.

How to fix: For v1, add a defensive `.limit(500)` or similar safety net so the query cannot return unbounded data. Long-term, push the overlap check into a DB query (GIN index on `sample_urls` supports containment queries, or use an RPC function with `unnest`).

**3. Missing integration tests for 8 new endpoints**
`test/it/` (absent)

The project's CLAUDE.md explicitly requires: "New or modified endpoints must include integration tests in `test/it/` - add shared test logic in `shared/tests/`, seed data in `postgres/seed-data/`, and a wiring file in `postgres/`." The repo has ~30 existing integration test suites following this pattern.

Why it matters: Unit tests mock PostgREST entirely, leaving untested: the UNIQUE(site_id, name) constraint producing actual SQLSTATE 23505; the partial unique index allowing name re-creation after soft-delete; RLS policy interactions; sort_order ordering with NULLs; and column type coercion for the JSONB sample_urls array.

How to fix: Add `test/it/shared/tests/agentic-rules.js`, seed data in `test/it/postgres/seed-data/`, and a wiring file. At minimum test: list returns ordered rows; create with duplicate name yields 409; soft-delete then re-create same name succeeds; cross-org access denied.

### Minor (Nice to Have)

**4. `badRequest` for missing postgrestClient is misleading**
`src/controllers/agentic-rules-factory.js:119-122`

The PostgREST client should always exist in a properly configured environment - this is a server configuration issue, not a client input error. Should be `internalServerError` to match the `dataAccess` check two lines above.

**5. No unit test for the 23505-to-409 mapping path**
`test/controllers/agentic-categories.test.js`

The "create returns 500 on PostgREST error" test uses a generic error. No test simulates `{code: '23505', message: '...'}` and asserts the 409 response and message shape. This user-facing contract deserves its own assertion.

**6. `getUserIdentifier` fallback to 'system' masks a broken invariant**
`src/controllers/agentic-rules-factory.js:33-37`

Since `authorize()` gates on `hasAccess()`, the fallback branch should be unreachable. If auth middleware behavior changes and requests arrive without profile data, mutation rows would silently lose attribution. Consider logging a warning when the fallback fires.

## Recommendations

- The list endpoint has no pagination. Rule counts per site are expected to be low for v1, and the response shape (`{ items: [...] }`) is pagination-ready. Consider adding a `.limit()` guard even without full cursor support, or document the intentional omission in the OpenAPI spec with a cardinality assumption.
- Consider trimming `body.name` (and `body.newName`) before validation. A name with leading/trailing whitespace creates "invisible" uniqueness conflicts.
- The regex utility (`regex-from-urls.js`) is a pure function with no service dependencies. If the audit-worker or future bulk-import tools need URL-to-regex, extraction to `spacecat-shared-utils` would prevent duplication.

## Assessment

Ready to merge? With fixes
Reasoning: The implementation is well-structured with clean factory abstraction, thorough auth enforcement, and solid test coverage within unit-test scope. The three Important items are: (1) missing server-side enforcement of limits already promised by the OpenAPI spec - a resource-exhaustion surface on a customer-facing endpoint; (2) an unbounded query that will degrade as rule counts grow; and (3) a documented project requirement for integration tests that validates the DB-layer behaviors unit tests cannot reach.

## Next Steps

1. Add input size validation for urls array (count + per-URL length) and rule name length to match the OpenAPI spec contract.
2. Add a defensive limit on the findSampleUrlConflict query or push the check into the database.
3. Add integration tests following the test/it/ pattern documented in CLAUDE.md.
4. Minor items are optional but straightforward.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 20s | <b>Cost:</b> $7.55 | <b>Commit:</b> <code>1d9c5765e3dbcbe75d14fc15add5b7d0246b01ea</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

akshaymagapu:
Thanks for the review. Addressed in `feaa991`:

**Important**
1. **Input size limits** — `urls` now capped at 50 items / 2048 chars each, and `name`/`newName` at 200 chars, enforced server-side in the factory (matching the OpenAPI contract). Added `maxItems`/`maxLength` to the request schemas too.
2. **Unbounded `findSampleUrlConflict`** — added a defensive `.limit(1000)` to both the dedup scan and the `list` query so a request can't pull/iterate unbounded rows as rule counts grow. (DB-side containment check noted as the long-term follow-up.)
3. **Integration tests** — ⚠️ deferred: the `test/it/` harness boots the `mysticat-data-service` image pinned at `v5.13.0`, whose migrations predate the `20260602` customer-edit schema (`source`/`sample_urls`/`status`/`created_by`). The tables can't be exercised until the IT image is bumped to a data-service release that includes those columns. Will add the shared test + seed + wiring once that image is available.

**Minor**
4. Missing PostgREST client now returns **500** (server misconfig), not 400.
5. The 23505→409 mapping already has a dedicated unit test (`create returns 409 on a duplicate-name unique violation`).
6. `getUserIdentifier` now logs a **warning** when it falls back to `'system'`.

**Recommendations**
- Whitespace is now **trimmed** from `name`/`newName` before validation (no invisible conflicts).
- `list` carries a `.limit()` guard; full cursor pagination remains a future item.
- Extracting `regex-from-urls` to `spacecat-shared-utils` noted for when a 2nd consumer needs it.

Full unit suite green (11339 passing), coverage gate passing.

MysticatBot (commented):
Hey @akshaymagapu,

## Strengths

- **All prior Important findings addressed comprehensively** (commit feaa991): Input size limits (MAX_SAMPLE_URLS=50, MAX_URL_LEN=2048, MAX_NAME_LEN=200) enforced server-side and aligned with OpenAPI schema constraints. Defensive `.limit(MAX_ACTIVE_RULES_SCAN)` cap on both the list endpoint and `findSampleUrlConflict`. PostgREST-missing case now correctly returns 500. Identity fallback now logs a warning. Whitespace trimming on names prevents invisible conflicts. All boundaries exercised by new tests.
- **OpenAPI spec and runtime enforcement are in sync** (`docs/openapi/agentic-rules-api.yaml` + `src/controllers/agentic-rules-factory.js`): `maxLength`, `maxItems`, and `minItems` in the schema match the runtime constants exactly. Clients get schema-level feedback AND the server enforces independently.
- **Thorough unit test coverage for the new validation paths**: URL count cap, per-URL length cap, name length cap, whitespace-only rejection, limit assertion, update-side caps, trim behavior, and identity fallback warning are all explicitly tested.
- **Factory pattern well-applied**: Single implementation shared across two dimensions with only table name and label diverging, avoiding the duplication that typically plagues paired-endpoint features.

## Issues

### Minor (Nice to Have)

**1. Missing URL-length cap test on the update path**
`test/controllers/agentic-categories.test.js`

There is a test for "create returns 400 when a url exceeds the 2048-char cap" but no corresponding test for the update path. The update path reuses `isValidUrlListBody` (so the production code is covered), but since create and update take different code paths to reach the validation (update must first fetch the existing rule), an explicit assertion for this boundary on update would make the test suite more robust against reordering regressions.

Suggested addition:
```js
it('update returns 400 when a url exceeds the 2048-char cap', async () => {
  const client = buildClient({ resolveWith: { data: { name: 'foo', source: 'human' }, error: null } });
  const { controller } = loadController();
  const longUrl = `/${'a'.repeat(2048)}`;
  const res = await controller.update(buildContext({
    client, params: { name: 'foo' }, data: { urls: [longUrl] },
  }));
  expect(res.status).to.equal(400);
});
```

**2. `list bounds the scan with a limit` test asserts type but not value**
`test/controllers/agentic-categories.test.js`

The test asserts `limitOp.args[0]` is `a('number')` but does not assert the actual value (1000). A regression that changes the limit to 1 would pass this test. Consider pinning: `expect(limitOp.args[0]).to.equal(1000)`.

**3. List endpoint silently truncates at 1000 rows**
`src/controllers/agentic-rules-factory.js` (list handler)

The `.limit(MAX_ACTIVE_RULES_SCAN)` silently truncates the response with no indicator to the client that more rows exist. The code comment ("v1 has no cursor") acknowledges this intentionally. For a future iteration, consider returning a `truncated: true` flag or logging a warning when the cap is hit: `if (data.length === MAX_ACTIVE_RULES_SCAN) context.log.warn(...)`.

## Recommendations

- Simplify `newName` trimming in the update handler - trim unconditionally before validation rather than branching through `hasText` first, making the data flow linear (coerce, trim, validate).
- Consider documenting the 1000-rule list cap in the OpenAPI spec as a behavior note, even without full pagination support.
- Track the integration test addition once the data-service IT image is bumped to include the 20260602 schema columns. The deferral is reasonable given the technical blocker.

## Assessment

Reasoning: The increment cleanly addresses all prior Important findings. OpenAPI spec and runtime enforcement are aligned. Input validation is thorough and well-tested. The remaining items are test-hardening suggestions and a documentation note - none block merge. The integration test deferral has a legitimate technical justification (IT image predates the new schema).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 1s | <b>Cost:</b> $4.27 | <b>Commit:</b> <code>feaa991b5ea761743d350078291630970735231f</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

akshaymagapu:
Addressed the minor items in `bada4fa`:

1. **Update-path URL-length test** — added an explicit `update returns 400 when a url exceeds the 2048-char cap` case.
2. **Pinned the list-limit assertion** — now `expect(limitOp.args[0]).to.equal(1000)` instead of a type-only check.
3. **Silent truncation** — `list` now logs a warning when the result hits the 1000 cap, with a new test asserting it. Also documented the cap as a behavior note on `AgenticRuleListResponse` in the OpenAPI spec.

Integration tests remain tracked as a follow-up pending the IT data-service image bump (predates the `20260602` schema). Full suite green.

MysticatBot (changes requested):
Hey @akshaymagapu,

## Strengths

- **All three prior Minor findings addressed cleanly** (commits `bada4fa` and `0dd08c7`): list limit test pinned to value 1000, truncation warning added with test coverage, and update url-length cap test added.
- Per-site rule cap piggybacks on the existing dedup scan (`src/controllers/agentic-rules-factory.js:118`) - no additional DB round-trip, clean reuse of already-fetched data.
- OpenAPI spec updated in sync with runtime enforcement: the 409 conflict response now documents all three conflict scenarios (name collision, URL overlap, cap), and the list response documents the 1000-row ceiling with server-side logging.
- Cap test (`test/controllers/agentic-categories.test.js`) correctly isolates the count check by using rules with empty `sample_urls`, preventing interference from the conflict path.
- Truncation warning uses optional chaining (`context.log?.warn?.(...)`) for safe degradation in contexts where the logger is absent.

## Issues

### Important (Should Fix)

**1. `ruleCount` includes AI-derived rules, contradicting stated customer-only cap**
`src/controllers/agentic-rules-factory.js:46-48` (comment) and `:230-235` (check)

The `findSampleUrlConflict` query selects ALL active rules regardless of `source` and returns `rules.length` as `ruleCount`. The cap check (`dedup.ruleCount >= MAX_RULES_PER_SITE`) therefore rejects the customer when the total active rules (AI + human) reaches 20. The code comment says "Max customer-created rules per site+dimension" and notes "auto-derivation is uncapped", implying the intent is to cap customer additions only.

Why it matters: The upstream audit-worker (PR 2539) is designed to auto-derive candidate rules. Sites with rich URL taxonomies will easily exceed 20 auto-derived rules. Those sites become the ones where customers want to add their own corrections, yet this cap silently blocks them with a 409 before a single customer rule exists.

How to fix - decide the intended behavior, then either:
- (a) Filter the count to human-source rules only:
```javascript
const humanRuleCount = rules.filter((r) => r.source === SOURCE_HUMAN).length;
return { conflict, ruleCount: humanRuleCount };
```
Update the query's `select` to include `source`: `.select('name,sample_urls,source')`.
- (b) If the combined cap is intentional, rename `MAX_RULES_PER_SITE` to `MAX_ACTIVE_RULES_PER_SITE`, update the comment to remove "customer-created", and update the OpenAPI description and error message to say "total active rules" so customers get a clear explanation.

### Minor (Nice to Have)

**2. Truncation warning test asserts `called` but not message content**
`test/controllers/agentic-categories.test.js` (the "list warns when the result hits the scan cap" test)

The assertion `expect(ctx.log.warn).to.have.been.called` passes if any code path calls `log.warn`, not necessarily the truncation path. A future change adding a different warning earlier in the flow would pass spuriously.

How to fix:
```javascript
expect(ctx.log.warn).to.have.been.calledWithMatch(/truncated at 1000/);
```

**3. No boundary test at exactly 19 rules (cap minus one)**
`test/controllers/agentic-categories.test.js`

The cap test verifies 20 rules triggers 409. No companion test confirms that 19 existing rules still allows creation. A regression from `>=` to `>` (or vice versa) would not be caught.

How to fix: Add a test with 19 existing rules asserting success, confirming the boundary is `>= 20`.

## Recommendations

- Consider whether the 20-rule cap should be per-site configurable (e.g., a column on the `sites` table with the constant as default) so different customer tiers can have different limits without a code change.
- Add a brief comment in `findSampleUrlConflict` noting its dual purpose (dedup + count) since the `select('name,sample_urls')` clause now serves both functions.

## Assessment

Ready to merge? With fixes
Reasoning: The increment is well-structured with clean factory reuse, thorough input validation, and properly documented API contract. The one Important finding is a behavioral ambiguity where the cap counts all active rules (AI + human) against a stated intent to cap only customer-created rules. On sites with active auto-derivation, customers could be permanently locked out of manual rule creation. The fix is either a one-line filter or a documentation correction depending on product intent.

## Next Steps

1. Clarify the cap semantics: filter to `source='human'` if capping customer rules only, or update comments/docs/error messages if capping total active rules.
2. Minor test-hardening items are optional.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 17s | <b>Cost:</b> $3.29 | <b>Commit:</b> <code>0dd08c73db5d80fc85506df81cea9040327e6b3e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

akshaymagapu:
Addressed in `13dd4a5`:

**Important — cap semantics (1):** Confirmed the intent is a **total** ceiling of 20 active rules per site+dimension (auto + customer), not customer-only. Went with option (b): renamed `MAX_RULES_PER_SITE` → `MAX_ACTIVE_RULES_PER_SITE`, removed the "customer-created" wording from the comment, and the 409 message now reads "maximum of 20 active … rules". The OpenAPI spec already described it as "active rules", so docs and runtime now agree.

**Minor (2):** Truncation-warning test now asserts `calledWithMatch(/truncated at 1000/)` instead of just `called`.

**Minor (3):** Added a 19-rule boundary test confirming create still succeeds below the cap (locks the `>= 20` boundary).

Also added a comment noting `findSampleUrlConflict`'s dual purpose (dedup + count). Full suite green.

Re the recommendation to make the cap per-site configurable (sites-table column): noted as a sensible future enhancement, out of scope for this PR.

MysticatBot (approved):
Hey @akshaymagapu,

## Strengths

- **All prior review findings cleanly addressed** (commit `13dd4a5`): `MAX_RULES_PER_SITE` renamed to `MAX_ACTIVE_RULES_PER_SITE` with updated comment and error message confirming the total-active-rules intent. The dual-purpose comment on `findSampleUrlConflict` (`src/controllers/agentic-rules-factory.js:110-111`) makes the non-obvious coupling between dedup and count explicit for future readers.
- **Test hardening is well-targeted**: `calledWithMatch(/truncated at 1000/)` (`test/controllers/agentic-categories.test.js:286`) pins the contract rather than just asserting invocation. The 19-rule boundary test correctly isolates the count check by using empty `sample_urls: []` so it does not accidentally trigger the dedup conflict path.
- **Factory pattern is a genuine architectural asset** (`src/controllers/agentic-rules-factory.js`): two dimensions share identical CRUD logic differing only by table name and label. Adding a third classification dimension requires one thin controller file, a route entry, and a capability declaration.
- **Security posture is solid**: auth enforced on all 8 endpoints (site:read for GET, site:write for mutations), input validation comprehensive and aligned with OpenAPI spec (50 URLs / 2048 chars / 200 char name), PostgREST client handles query parameterization, error responses do not leak internals.
- **Conventions fully followed**: factory function pattern, DTO layer, access control checks, declarative routing, no Lambda bundle violations, OpenAPI spec in sync with runtime enforcement.

## Issues

### Minor (Nice to Have)

**1. User-supplied URL reflected verbatim in conflict error message**
`src/controllers/agentic-rules-factory.js:240-243`

The 409 conflict response includes the raw user-supplied URL: `URL "${dedup.conflict.url}" already belongs to...`. In the current JSON API context this is not exploitable (no HTML rendering). However, if any downstream consumer (monitoring dashboard, Slack notification) renders this message as HTML without escaping, the reflected string becomes a stored XSS vector. Acceptable as-is given the JSON-only contract - noting for awareness.

## Recommendations

- The factory pattern is ready to standardize. When a second consumer of `regexFromUrls` appears (e.g. audit-worker), extraction to `spacecat-shared-utils` would prevent cross-service duplication.
- The 20-rule cap is hardcoded. Consider making it a site-level or organization-level config column with 20 as the default, so different customer tiers can have different limits without a code deploy.

## Assessment

Ready to merge? Yes
Reasoning: The increment cleanly resolves all prior review findings with focused, correct changes. Naming, comments, error messages, and tests are now consistent with confirmed product semantics. Architecture is sound (factory pattern, stateless regex, direct PostgREST, soft-delete lifecycle). Security posture is solid. All CI gates pass.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 39s | <b>Cost:</b> $3.66 | <b>Commit:</b> <code>13dd4a57fcee5aceab4c40fa3d70bc75f4ce4aad</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2541, opened 2026-06-03
feat: update onboarding flow to align with audit-worker config-free execution 

tkotthakota-adobe:
https://jira.corp.adobe.com/browse/SITES-45860
                                                                                                                                                                                                        
Following the recent audit-worker change that allows audits to run regardless of configuration state, the onboarding flow no longer needs to enable audits in the global configuration before triggering them.                                                                       
                                                                                                                                             
  - Audit enable is now gated to scheduledRun=true only — enabling in config is only needed for recurring scheduled runs; one-shot onboards rely on the audit-worker firing without a config entry
  - Fixed auditTypes in disableImportAndAuditJob taskContext — was hardcoded to [], now correctly passes the full profile audit type list so the task processor can clean up any audits that were enabled                                                                                                                                                                                 
  - Added exceljs and iso-3166 to ESLint import/no-unresolved ignore list — these installed packages cannot be statically resolved by the ESLint import resolver (same pattern as @octokit/rest)


tathagat2241 (commented):
Issues
  
  Bug — Dead code in else if (scheduledRun) (Critical)

  File: src/support/utils.js, around line 1814

  // Before
  } else {
    log.debug(`All the required audits...`);
  }
  
  // After
  } else if (scheduledRun) {   // ← always false here
    log.debug(`All the required audits...`);
  }

  This block is unreachable. We're in the else branch of if (scheduledRun), so scheduledRun is guaranteed to be false — the else if (scheduledRun) condition can never be true. The debug log will never fire.

  Intent is unclear — was the goal to remove the log entirely, or to replace it with a different message for the non-scheduled case? Either way, else if (scheduledRun) here is logically impossible and should be resolved.

src/support/utils.js
```diff
@@ -1780,33 +1780,34 @@ export const onboardSingleSite = async (
...
-        if (!isEnabled) {
-          latestConfiguration.enableHandlerForSite(auditType, site);
-          auditsEnabled.push(auditType);
-        }
+    for (const auditType of auditTypes) {
+      /* eslint-disable no-await-in-loop */
+      const isEnabled = latestConfiguration.isHandlerEnabledForSite(auditType, site);
+      if (!isEnabled) {
+        latestConfiguration.enableHandlerForSite(auditType, site);
+        auditsEnabled.push(auditType);
```
rpapani: @tkotthakota-adobe we  should move this logic under if scheduledRun or paid profile. We don't need to maintain and pass the auditsEnabled list to task processor, as it doesn't need to disable any. All the enable/disable logic can be run in api service itself.

rpapani: please note that keep the imports enabling logic as is, as we need enabling the imports, which will add entries in the site config, but those won't be used for scheduling the imports, so we only have to handle schedule for audits.

rpapani (changes requested):

src/support/utils.js
tkotthakota-adobe: Hi @rpapani Both points are valid. We do not actually need to send to task processor. Changes incorporated.

tathagat2241 (commented):
## Overview
Now that audit-worker runs audits without a config entry, this stops the onboarding flow from enabling audits in global config for one-shot onboards. onboardSingleSite (src/support/utils.js) now enables+saves audit handlers only when scheduledRun || profile.protected; free-trial onboards just get audits triggered once. It also drops auditTypes: [] from the disableImportAndAuditJob taskContext and adds exceljs/iso-3166 to the ESLint import/no-unresolved ignore list. I reviewed it through two lenses: (1) the dropped auditTypes field is a cross-repo contract with the task-processor, and (2) the PR description makes claims that needed checking against the diff.

## Blockers
None. The one change with real breakage potential — removing auditTypes from the taskContext — is verified safe: adobe/spacecat-task-processor src/tasks/disable-import-audit-processor/handler.js destructures { auditTypes = [] } = taskContext, so a missing field behaves identically to the old explicit [] (the disable loop iterates nothing). No downstream crash.

## Should-fix
- PR description bullet #2 is inverted. It says auditTypes "now correctly passes the full profile audit type list so the task processor can clean up any audits that were enabled." The code does the opposite — it removes the field (task-processor defaults it to [] → cleans up nothing), which matches the new in-code comment at src/support/utils.js:2005. The code is self-consistent and correct; the description will mislead reviewers/future readers (looks like a stale artifact from before "addressed review comments"). Please fix the description.

- test/support/utils-scheduledrun.test.js — the PR's core behavior change is not directly tested. This file only re-implements the scheduledRun ternary inline and asserts on the copy; it never calls onboardSingleSite. Add an assertion that enableHandlerForSite/Configuration.save fire for a scheduledRun/protected profile and are skipped for a free-trial one — that's the whole point of the change and it's currently unguarded against regression.

- package-lock.json — ~30 peer: true flag toggles with no version changes and package.json untouched in this PR. This is incidental lockfile regeneration (likely a different local npm version), unrelated to the stated scope. Revert it or regenerate with the project's pinned npm, otherwise it's merge-conflict noise.

## Nits
- src/support/utils.js:1937 — the non-scheduled success message still reads "Enabled … audits: <list>", but on that path audits are now only triggered once, not enabled in config. Slightly inaccurate now that enabling is gated; consider "Triggered audits" for the non-scheduled branch.

- PR body bullet #1 says gating is "scheduledRun=true only", but the code is scheduledRun || profile.protected (utils.js:1902). Minor wording.

## Strengths
- Correctly recognized and preserved the cross-repo contract — the change is genuinely behavior-neutral on the task-processor side.

- Nice incidental win: Configuration.findLatest() (a DB read) now runs only for scheduled/protected profiles instead of every onboard.

- Clean refactor — latestConfiguration/auditsEnabled are fully scoped inside the guard with no dangling references, and the added comments explaining why no disable step is needed are genuinely helpful.

- exceljs/iso-3166 are real, declared deps (package.json ^4.4.0) actually imported in serenity/locations.js and llmo/llmo-sheet-write.js, so the ESLint ignore is justified.

## Tests
Existing onboard.test.js / onboard-modal.test.js / utils-paid-profile-guard.test.js still exercise onboardSingleSite, but focus on the paid/tier guards, not the new enable-gating. I could not run the suite here, but I verified the task-processor contract directly via gh api rather than trusting it. The gating and the dropped-field behavior are the coverage gaps (see should-fix [#2](https://github.com/ramboz/jig/issues/2)).

## Verdict
Approve with minor notes — the code is correct and the risky part is verified safe; fixing the misleading description, dropping the lockfile churn, and adding one targeted test would make this solid.

tathagat2241 (approved):

tkotthakota-adobe:
Hi @tathagat2241 thanks for the additional review. Comments are addressed
 1. src/support/utils.js:1937 — message for non-scheduled branch now reads "triggered (not scheduled) audits: ${auditsMessage}" instead of "Enabled … audits". Accurate to the actual behavior.                                                    
                                                                                                                                                                                                                                                    
  2. test/support/utils-scheduledrun.test.js — added a new describe block (onboardSingleSite — audit handler enable/disable gating) with four tests that call the actual function via esmock and assert:                                            
  - Free-trial: enableHandlerForSite and Configuration.save are NOT called, disableHandlerForSite is NOT called (nothing to disable)                                                                                                                
  - scheduledRun=true: enableHandlerForSite and Configuration.save ARE called                                                                                                                                                                       
  - profile.protected=true (paid): enableHandlerForSite and Configuration.save ARE called                                                                                                                                                           
  - Free-trial re-onboard with previously-enabled audits: disableHandlerForSite IS called, enableHandlerForSite is NOT                                                                                                                              
                                                                                                                                                                                                                                                    
  3. PR description (https://github.com/adobe/spacecat-api-service/pull/2541) — updated:                                                                                                                                                            
  - Bullet #1: now says scheduledRun=true or profile.protected                                                                                                                                                                                      
  - Bullet #2: now accurately says auditTypes was dropped (not set to full list), and explains the task-processor defaults { auditTypes = [] } → disable loop is a no-op — consistent with in-process handling                                      
                                                                                                                                                                                                                                                    
  4. package-lock.json — reverted to origin/main via git checkout origin/main -- package-lock.json. Run npm install if you need the lockfile updated with any new deps.                                                                             
                                                                                                                                                                                                                                                    
  One note on the new tests: the makeContext mock stubs Organization.findByImsOrgId to reject — which is fine since the tests use findByBaseURL to return an existing site (skips the org-creation path). But updateRumConfig is mocked via esmock  
  on ./rum-config-service.js — you may need to verify the relative import path matches your esmock setup. Run npm test -- --grep "audit handler enable/disable" to confirm before committing.

src/support/utils.js
```diff
@@ -1904,45 +1904,63 @@ export const onboardSingleSite = async (
...
-    const auditsEnabled = [];
-    if (scheduledRun) {
-      for (const auditType of auditTypes) {
-        /* eslint-disable no-await-in-loop */
-        const isEnabled = latestConfiguration.isHandlerEnabledForSite(auditType, site);
-        if (!isEnabled) {
-          latestConfiguration.enableHandlerForSite(auditType, site);
-          auditsEnabled.push(auditType);
-        }
+    const wantEnabled = scheduledRun || profile.protected;
```
rpapani, line 1922: why `protected`, why not `paid` ?

rpapani (approved):

src/support/utils.js
tkotthakota-adobe: `profile.protected` is used instead of a name check (`=== "paid"`) for forward compatibility — if a future tier (e.g. `paid-enterprise`) is added with `"protected": true` in profiles.json, this logic picks it up automatically without a code change. Hardcoding `"paid"` would require updating this condition every time a new protected tier is introduced. The same flag already drives the downgrade guard at line 1648 for the same reason.

tkotthakota-adobe: @rpapani

PR #2542, opened 2026-06-03
feat(s2s): expose auto-fix endpoint to S2S consumers via fixEntity:create

ravverma:
## Summary

- Moves `PATCH /sites/:siteId/opportunities/:opportunityId/suggestions/auto-fix` out of `INTERNAL_ROUTES` and maps it to the new `fixEntity:create` capability in `required-capabilities.js`
- Adds `CAP_FIX_ENTITY_CREATE = 'fixEntity:create'` to `capability-constants.js` as the single source of truth
- In `autofixSuggestions` controller, adds a Layer 2 `hasS2SCapability(CAP_FIX_ENTITY_CREATE)` check before the existing `hasAccess(site, 'auto_fix')` check — S2S consumers with this capability bypass the IMS user profile gate; non-S2S callers are unaffected

## Root cause

S2S consumers calling `PATCH /suggestions/auto-fix` were getting 403 because the endpoint was in `INTERNAL_ROUTES` (blocked at Layer 1), and even when reached, `hasAccess(site, 'auto_fix')` checks for an IMS user scope that service accounts structurally cannot hold.

## Test plan

- [ ] All existing `auto-fix suggestions` tests pass unchanged (non-S2S path unaffected)
- [ ] New test: S2S consumer with `fixEntity:create` → 207 (bypasses profile check)
- [ ] New test: S2S consumer without `fixEntity:create` + no `auto_fix` access → 403
- [ ] `required-capabilities.test.js`: disjointness invariant passes, route coverage passes
- [ ] `capability-constants.test.js`: Layer 2 call-site check passes (suggestions.js added to scan)
- [ ] Register the S2S consumer with `fixEntity:create` capability via the standard S2S admin process

### DIFF
--- a/docs/openapi/site-opportunities.yaml
+++ b/docs/openapi/site-opportunities.yaml
@@ -873,6 +873,11 @@ site-opportunity-suggestions-auto-fix:
       is true (for assess or assess-urls), the worker runs only checks that do not require
       AEM CS access and does not need a valid promise token; when omitted or false, behavior
       is unchanged (full assessment with AEM when possible).
+
+      **S2S access:** S2S consumers registered with the `fixEntity:create` capability may
+      call this endpoint. The `auto_fix` IMS user-profile scope check is bypassed for S2S
+      callers; all other validations (site ownership, opportunity lookup, handler enablement)
+      apply unchanged.
     tags:
       - opportunity-suggestions
     requestBody:
--- a/src/controllers/suggestions.js
+++ b/src/controllers/suggestions.js
@@ -1056,7 +1056,12 @@ function SuggestionsController(ctx, sqs, env) {
     }
 
     const s2sResult = await accessControlUtil.hasS2SCapability(CAP_FIX_ENTITY_CREATE);
-    if (!s2sResult.allowed && !await accessControlUtil.hasAccess(site, 'auto_fix')) {
+    if (s2sResult.allowed) {
+      ctx.log?.info(`[acl] S2S auto-fix granted - clientId=${s2sResult.clientId} consumerId=${s2sResult.consumerId}`);
+    } else if (!await accessControlUtil.hasAccess(site, 'auto_fix')) {
+      if (s2sResult.reason !== 'not-s2s') {
+        ctx.log?.info(`[acl] Denied PATCH auto-fix - reason=${s2sResult.reason} clientId=${s2sResult.clientId || 'n/a'} consumerId=${s2sResult.consumerId || 'n/a'}`);
+      }
       return forbidden('User does not belong to the organization or does not have sufficient permissions');
     }
 
--- a/src/controllers/suggestions.js
+++ b/src/controllers/suggestions.js
@@ -47,6 +47,7 @@ import {
   isViewAsTrialRequest,
 } from '../support/utils.js';
 import AccessControlUtil from '../support/access-control-util.js';
+import { CAP_FIX_ENTITY_CREATE } from '../routes/capability-constants.js';
 import { grantSuggestionsForOpportunity } from '../support/grant-suggestions-handler.js';
 import { createAtomicStrategy, deleteAtomicStrategy } from '../support/atomic-strategy-helper.js';
 
@@ -1054,7 +1055,8 @@ function SuggestionsController(ctx, sqs, env) {
       return notFound('Site not found');
     }
 
-    if (!await accessControlUtil.hasAccess(site, 'auto_fix')) {
+    const s2sResult = await accessControlUtil.hasS2SCapability(CAP_FIX_ENTITY_CREATE);
+    if (!s2sResult.allowed && !await accessControlUtil.hasAccess(site, 'auto_fix')) {
       return forbidden('User does not belong to the organization or does not have sufficient permissions');
     }
 
--- a/src/routes/capability-constants.js
+++ b/src/routes/capability-constants.js
@@ -23,3 +23,5 @@ export const CAP_SITE_READ_ALL = 'site:readAll';
 export const CAP_ORG_READ_ALL = 'organization:readAll';
 
 export const CAP_SITE_CREATE = 'site:create';
+
+export const CAP_FIX_ENTITY_CREATE = 'fixEntity:create';
--- a/src/routes/required-capabilities.js
+++ b/src/routes/required-capabilities.js
@@ -10,7 +10,12 @@
  * governing permissions and limitations under the License.
  */
 
-import { CAP_ORG_READ_ALL, CAP_SITE_CREATE, CAP_SITE_READ_ALL } from './capability-constants.js';
+import {
+  CAP_FIX_ENTITY_CREATE,
+  CAP_ORG_READ_ALL,
+  CAP_SITE_CREATE,
+  CAP_SITE_READ_ALL,
+} from './capability-constants.js';
 
 /**
  * Routes that are intentionally excluded from S2S consumer access.
@@ -38,9 +43,7 @@ export const INTERNAL_ROUTES = [
   // GitHub App webhook - authenticated by HMAC-SHA256 signature, not S2S JWT
   'POST /webhooks/github',
 
-  // Suggestion edge ops (auto-fix, edge-deploy, etc.): not yet required by S2S
-  // TODO: Add these back in when we have a S2S consumer that needs them
-  'PATCH /sites/:siteId/opportunities/:opportunityId/suggestions/auto-fix',
+  // Suggestion edge ops (edge-deploy, etc.): not yet required by S2S
   'POST /sites/:siteId/opportunities/:opportunityId/suggestions/edge-deploy',
   'POST /sites/:siteId/opportunities/:opportunityId/suggestions/edge-rollback',
   'POST /sites/:siteId/opportunities/:opportunityId/suggestions/edge-preview',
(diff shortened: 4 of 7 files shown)


MysticatBot (changes requested):
Hey @ravverma,

## Strengths

- Well-scoped change that does exactly three things (remove from INTERNAL_ROUTES, map capability, add Layer 2 check) with no unrelated noise.
- `src/routes/required-capabilities.js:390` - the route entry for `auto-fix` is placed above the wildcard `:suggestionId` pattern, avoiding ambiguity in route-matching order.
- `src/routes/capability-constants.js:27` - introducing a dedicated `fixEntity:create` capability rather than reusing `suggestion:write` follows the principle of least privilege.
- `src/controllers/suggestions.js:1055-1056` - the short-circuit evaluation preserves the existing IMS user path unchanged. Non-S2S callers see zero behavioral change.
- The disjointness invariant (route cannot appear in both `INTERNAL_ROUTES` and `routeRequiredCapabilities`) is maintained. The INTERNAL_ROUTES comment is updated to reflect remaining entries.
- `test/routes/capability-constants.test.js:77` - adding `suggestions.js` to the drift-coverage scan keeps the existing validation gate intact.

## Issues

### Important (Should Fix)

1. **`src/controllers/suggestions.js:1055-1056` - Missing audit log on S2S capability check**

   The established pattern in `sites.js` logs denial details (`[acl] Denied ... reason=... clientId=... consumerId=...`) when an S2S capability check fails. The new code discards the structured `s2sResult` (reason, clientId, consumerId) without logging on either the grant or denial path.

   Why it matters: Auto-fix is a write operation that mutates suggestions. If a compromised or misconfigured S2S consumer triggers auto-fixes, the SOC team needs to correlate the action to a specific clientId/consumerId. Additionally, when S2S denial occurs, operators cannot distinguish "IMS user lacks profile scope" from "S2S consumer is registered but missing the capability."

   How to fix:
   ```javascript
   const s2sResult = await accessControlUtil.hasS2SCapability(CAP_FIX_ENTITY_CREATE);
   if (s2sResult.allowed) {
     log.info(`[acl] S2S auto-fix granted - clientId=${s2sResult.clientId} consumerId=${s2sResult.consumerId}`);
   }
   if (!s2sResult.allowed && !await accessControlUtil.hasAccess(site, 'auto_fix')) {
     if (s2sResult.reason !== 'not-s2s') {
       log.info(`[acl] Denied PATCH auto-fix - reason=${s2sResult.reason} clientId=${s2sResult.clientId || 'n/a'} consumerId=${s2sResult.consumerId || 'n/a'}`);
     }
     return forbidden('User does not belong to the organization or does not have sufficient permissions');
   }
   ```

2. **`test/controllers/suggestions.test.js` - Missing test: S2S denied but user HAS auto_fix access**

   The diff tests two paths: (a) S2S allowed -> 207, (b) S2S denied + user denied -> 403. But the production logic has a third path where `s2sResult.allowed` is false but the user does have `auto_fix` access, which should still succeed (return 207). This is the regression path for existing non-S2S users.

   Why it matters: If someone later refactors the condition (e.g., flips `&&` to `||`), only this test would catch the regression for existing users who already have access.

   How to fix: Add a test where `hasS2SCapability` returns `{ allowed: false, reason: 'not-s2s' }` and `hasAccess` returns `true`, asserting the response is 207.

3. **`docs/openapi/` - OpenAPI spec not updated for changed access semantics**

   CLAUDE.md rule: "Specification Sync: Keep OpenAPI specs and implementation in sync." This PR changes the access control semantics of `PATCH .../suggestions/auto-fix` from internal-only to S2S-accessible with the `fixEntity:create` capability. No `docs/openapi/` files are modified.

   How to fix: Update the relevant OpenAPI path spec to document that the endpoint now accepts S2S authentication with the `fixEntity:create` capability.

4. **`test/it/` - Missing integration tests for the modified endpoint**

   CLAUDE.md rule: "New or modified endpoints must include integration tests in test/it/." The endpoint's access control behavior is materially modified. No integration test additions appear.

   How to fix: Add integration tests in the `test/it/` shared test structure exercising the S2S capability path (S2S consumer with `fixEntity:create` -> 207, S2S consumer without it and no profile access -> 403).

### Minor (Nice to Have)

5. **`test/controllers/suggestions.test.js:6222` - `hasAccess` stubbed with `.returns(false)` instead of `.resolves(false)`**

   In production, `hasAccess` is awaited (confirming it returns a Promise). The test stubs it with `.returns(false)` (synchronous). This works because `await false` yields `false`, but it's inconsistent with how `hasS2SCapability` is stubbed (`.resolves(...)`) in the same test and misleading about the method's contract.

   How to fix: Change `.returns(false)` to `.resolves(false)`.

## Recommendations

- The successful S2S test verifies 207 but does not assert that `hasAccess` was not called. A stub like `sandbox.stub(AccessControlUtil.prototype, 'hasAccess').rejects(new Error('should not be reached'))` would make the bypass contract ironclad and catch regressions where the IMS check is accidentally reached for S2S consumers.
- The Layer 1 + Layer 2 pattern (route-level capability gate + controller-level check) will likely repeat for `edge-deploy`, `edge-rollback`, and `edge-preview` (the remaining INTERNAL_ROUTES with similar TODO shapes). Consider documenting this as the standard approach so future authors replicate it consistently.

## Assessment

Ready to merge? With fixes
Reasoning: The authorization logic is sound and well-tested for the core paths. The missing audit logging, the regression test gap for existing users, and the CLAUDE.md compliance items (OpenAPI sync, integration tests) should be addressed before merge.

## Next Steps

1. Add audit logging on the S2S grant and denial paths to match the established pattern.
2. Add the "S2S denied + user allowed" test case for regression protection.
3. Update OpenAPI spec and add integration tests per CLAUDE.md requirements.
4. The minor `.returns`/`.resolves` fix can be addressed alongside the other test additions.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 31s | <b>Cost:</b> $4.58 | <b>Commit:</b> <code>4af789aacfce2f6a7aa80aa0cd3d24b7bf74c87c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @ravverma,

## Strengths

- All five prior review findings addressed cleanly in a single follow-up commit (9e9904b) - audit logging, regression test, OpenAPI update, integration tests, and the `.resolves(false)` fix are all present.
- `src/controllers/suggestions.js:1055-1061` - The three-path control flow (S2S granted / S2S denied + user denied / non-S2S + user granted) is clearly expressed without unnecessary nesting. The two-layer authorization model (Layer 1 route gate + Layer 2 controller check) provides defense-in-depth.
- `src/controllers/suggestions.js:1059` - The `reason !== 'not-s2s'` guard on the denial log avoids noise from regular IMS users while surfacing misconfigured S2S consumers. Good operational design choice.
- `test/controllers/suggestions.test.js:6196-6199` - Stubbing `hasAccess` with `.rejects(new Error('hasAccess must not be called...'))` as a trap assertion proves the bypass contract, not just the status code. Future refactors that accidentally evaluate the IMS check on the S2S path will trip this immediately.
- `src/routes/required-capabilities.js:390` - The `auto-fix` entry is placed above the wildcard `:suggestionId` pattern, correctly preventing route-matching ambiguity.
- `src/routes/capability-constants.js:27` - Introducing a dedicated `fixEntity:create` capability rather than overloading `suggestion:write` respects least-privilege at the authorization boundary.
- CLAUDE.md compliance satisfied: OpenAPI spec in sync, integration tests in `test/it/` present, controller follows factory function convention, route follows declarative `METHOD /path` pattern.

## Issues

### Minor (Nice to Have)

1. **`test/controllers/suggestions.test.js:6222` - Denial test does not verify audit log fires**

   The test for "S2S consumer missing fixEntity:create when user lacks auto_fix access" stubs `hasS2SCapability` with `reason: 'missing-capability'`, which should trigger the `ctx.log?.info('[acl] Denied PATCH auto-fix...')` path. However, the test invocation uses `suggestionsController.autofixSuggestions(...)` without `...context` (which carries the log), so `ctx.log` is likely undefined and the log line silently no-ops via optional chaining.

   The functional behavior (403) is correctly verified, so this is not a correctness gap, but it leaves the audit contract untested.

   How to fix: Either pass `...context` and assert `ctx.log.info` was called with the expected structured fields, or add a dedicated test that verifies the denial audit log fires with correct clientId/consumerId/reason interpolation.

## Recommendations

- The remaining `INTERNAL_ROUTES` entries for `edge-deploy`, `edge-rollback`, and `edge-preview` carry the same TODO shape this PR resolved. Consider documenting the Layer 1 + Layer 2 pattern once (even as a short comment block above `INTERNAL_ROUTES`) so future authors replicate the approach without reverse-engineering it from the `auto-fix` precedent.
- A positive-path integration test (S2S consumer WITH `fixEntity:create` reaching Layer 2 and getting 207) would complete the IT story, but likely requires registering a new test consumer fixture. Worth tracking as follow-up.

## Assessment

Ready to merge? Yes
Reasoning: The authorization logic is architecturally sound with proper defense-in-depth, audit logging on both grant and denial paths, and unchanged behavior for existing non-S2S callers. Test coverage spans all three behavioral paths plus Layer 1 integration. The single minor finding is a test robustness improvement, not a functional risk.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 38s | <b>Cost:</b> $3.12 | <b>Commit:</b> <code>9e9904b31cb0e587e2b090382c32fe477a143ff6</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2543, opened 2026-06-03
fix(topics): return 422 instead of 500 on foreign-key violation (23503)

rainer-friederich:
## Summary

- `createTopic` already mapped Postgres error `23505` (unique constraint) to a typed 409. A stale or nonexistent `brandId` passed to `POST /v2/orgs/:id/topics` produces error `23503` (FK violation), which fell through to a generic `throw new Error(...)` with no `.status` — causing `createErrorResponse` to return 500.
- Added a `23503` branch that throws a 422 so callers receive a typed client error rather than an opaque 500. The original PostgREST error is preserved as `.cause` for operator triage.
- Added a matching unit test in `topics-storage.test.js` following the same pattern as the existing 23505 test.

## Root cause

DRS scheduler schedules can carry a stale `brand_id` that no longer exists in the Postgres `brands` table (brand deleted after the schedule was created). When `ensure_topics_v2` posts a topic with that orphaned UUID, Postgres rejects it with `23503`. The immediate fix is a data cleanup on the DRS side; this PR hardens the API layer so the failure is always a 422 instead of a 500.

## Test plan

- Run `npm test -- --grep "createTopic"` — all tests pass including the new 23503 case
- Confirm `POST /v2/orgs/:id/topics` with a nonexistent `brandId` returns 422


MysticatBot (approved):
Hey @rainer-friederich,

## Strengths

- **Pattern consistency** (`src/support/topics-storage.js:143-153`): The new 23503 branch mirrors the existing 23505 handler exactly - same catch-block location, same structure (check `error.code`, construct typed Error with `.status`, preserve `.cause`). Predictable and easy to maintain.
- **Correct layer placement**: Postgres error-to-HTTP-status translation lives in the storage layer, which is the right boundary for this concern.
- **Error chaining** via `{ cause: error }`: Preserves the original PostgREST error for operator debugging without leaking it directly in the response.
- **Well-structured test** (`test/support/topics-storage.test.js:488-515`): Asserts on four distinct observable properties (instanceof, status code, message substring, cause identity), mocks at the correct boundary (PostgREST client), and follows the existing test pattern.
- **Change discipline**: Two files, one behavior, no scope creep.

## Issues

### Minor (Nice to Have)

1. **Information disclosure via Postgres internals in error message** - `src/support/topics-storage.js:148`

   The interpolated `error.message` contains internal schema details (table names, constraint names like `topics_brand_id_fkey`). This leaks database schema information to the API consumer.

   **Why it matters:** An attacker enumerating the API can learn table and relationship structures from error responses.

   **Mitigating context:** The existing 23505 handler uses the same pattern, and the callers are currently internal (DRS scheduler). This is a pre-existing cross-cutting concern for the error response serialization layer, not something to fix in a single storage function.

   **If addressed in a future pass:** Use a generic message and keep Postgres details only in `.cause`:
   ```javascript
   const fk = new Error(
     'Topic references a non-existent related entity',
     { cause: error },
   );
   ```

## Recommendations

- **Consider centralizing Postgres error-code-to-HTTP-status mapping.** The `if (error.code === 'XXXXX') { wrap; set .status; throw }` pattern is now repeated twice in this function and likely exists (or will exist) in other storage modules. A small utility like `throwIfPgConstraintViolation(error, { '23505': 409, '23503': 422 })` would make the mapping declarative and prevent each storage function from growing its own catch-block chain. Not urgent for this PR, but worth considering as a pattern.

## Assessment

Ready to merge? Yes

This is a clean, well-scoped bug fix that converts an opaque 500 into a meaningful 422 for a known production scenario. The code follows existing patterns, includes a proper unit test, and preserves error context for operators. No blocking issues.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 28s | <b>Cost:</b> $1.83 | <b>Commit:</b> <code>3b9658cd49bb28695fc967dcbe23cc6656badf6d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (changes requested):
Hey @rainer-friederich,

## Strengths

- **Prior information disclosure concern addressed** (`src/support/errors.js:24-45`): The 23503 FK violation now uses a generic message ('Topic references a non-existent related entity') with no table names, constraint names, or schema details leaking to the client. Postgres internals are preserved exclusively in `.cause` for operator triage.
- **Prior centralization recommendation addressed** (`src/support/errors.js:37-44`): `throwOnPgConstraintViolation` is a clean, focused utility - 6 lines of logic, no dependencies, no framework. Callers declare a `codeMap` and get typed HTTP errors without repeating if/throw chains. The JSDoc explicitly warns against exposing Postgres internals in client-facing messages.
- **Thorough test coverage** (`test/support/errors.test.js`): Edge cases (null error, missing `.code`, unmatched code) are covered alongside the happy paths. The integration test in `test/support/topics-storage.test.js:488-515` exercises the full `createTopic` flow with a realistic PostgREST error shape.
- **Defensive coding** (`src/support/errors.js:39`): `error?.code` handles null/undefined inputs gracefully without requiring callers to guard beforehand.

## Issues

### Important (Should Fix)

1. **Constraint name still leaks in the 23505 client-facing message** - `src/support/topics-storage.js:133-135`

   The regex extracts the Postgres constraint name from `error.message` and interpolates it into the client response: `Topic conflicts with ${constraint} for this organization`. This surfaces schema internals (e.g. `topics_organization_id_name_key`) to API consumers.

   **Why it matters:** This contradicts both the utility's own JSDoc ("Postgres internals (table names, constraint names) should not appear in `message`") and the repo's CLAUDE.md rule ("Never expose stack traces or internal details to clients"). The 23503 branch correctly uses a fully generic message - the 23505 branch should follow the same pattern for consistency.

   **How to fix:** Use a generic message and remove the regex extraction entirely:
   ```javascript
   throwOnPgConstraintViolation(error, {
     '23505': { status: 409, message: 'A topic with these attributes already exists for this organization' },
     '23503': { status: 422, message: 'Topic references a non-existent related entity' },
   });
   ```
   The constraint name is already available in `.cause` for operator triage.

### Minor (Nice to Have)

2. **Regex runs unconditionally for all error codes** - `src/support/topics-storage.js:133-134`

   The constraint-name regex now executes on every error (including 23503 and unmatched codes), but its result is only used in the 23505 map entry. If the Important finding above is addressed (static message for 23505), this regex can be removed entirely. If not, consider moving it inside a message factory to avoid misleading future readers into thinking it applies to all codes.

## Recommendations

- The fallback `throw new Error(\`Failed to create topic: ${error.message}\`)` on line 136 still interpolates the raw Postgres error for any code not in the map. For consistency with the hardening goal, consider a generic message like `'Failed to create topic'` with `{ cause: error }` for operator detail. This is pre-existing behavior and outside this PR's immediate scope, but worth a follow-up.

## Assessment

Ready to merge? With fixes

The centralization is well-executed and the utility is cleanly designed. The one Important finding (constraint name in the 23505 message) contradicts the contract the PR's own JSDoc establishes - a one-line fix that aligns the code with its documented intent. CI checks are currently failing and should be resolved before merge.

## Next Steps

1. Replace the constraint-name interpolation with a generic message for the 23505 entry.
2. Investigate and fix the CI failure.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 24s | <b>Cost:</b> $3.10 | <b>Commit:</b> <code>61eb3b85013c8bf44283aec62e96c70ea8148b10</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the review feedback in d6217475:

- Important (23505 constraint-name leak): the `23505` branch now returns a static generic message (`A topic with these attributes already exists for this organization`) instead of interpolating the Postgres constraint name. The constraint/table names stay in `.cause` for operator triage only, matching the `23503` branch and the `throwOnPgConstraintViolation` JSDoc contract.
- Minor (unconditional regex): removed the unique-constraint regex entirely, since it no longer feeds any message.
- Recommendation (fallback interpolates raw message): the fallback is now `new Error('Failed to create topic', { cause: error })` — no raw Postgres message in the client-facing string.
- CI failure: the `ci / build` lint failure was `quote-props` flagging the quoted numeric keys. Unquoted the `23505`/`23503` keys in both `topics-storage.js` and `errors.test.js` (object keys are coerced to strings, so `codeMap[error.code]` lookups are unaffected).
- Updated the 23505 unit test to assert the generic message and that the constraint name does not leak; dropped the now-obsolete "lacks constraint clause" test since the regex is gone.

Local: `npm run lint` clean on the touched files, `topics-storage.test.js` + `errors.test.js` green (36 passing).

MysticatBot (approved):
Hey @rainer-friederich,

## Strengths

- **Prior information-disclosure concern addressed** (`src/support/topics-storage.js:128-131`): Both 23505 and 23503 now use static, generic messages with no interpolation of Postgres internals. Constraint names stay exclusively in `.cause` for operator triage, satisfying the repo's "never expose internal details to clients" rule.
- **Prior unconditional-regex concern addressed**: The regex is gone entirely - replaced by the declarative `codeMap` approach.
- **Secure-by-design utility** (`src/support/errors.js:24-31`): `throwOnPgConstraintViolation` is 6 lines of logic, zero dependencies, with JSDoc that explicitly warns against leaking Postgres internals in the `message` field. The pattern makes it structurally difficult to accidentally interpolate raw DB error content into client responses.
- **Fallback error hardened** (`src/support/topics-storage.js:133`): The catch-all `throw new Error('Failed to create topic', { cause: error })` no longer interpolates `error.message`, closing the schema-leakage vector for unmapped error codes too.
- **Negative assertion codifies the no-leak contract** (`test/support/topics-storage.test.js:465`): `expect(err.message).to.not.include('uq_topic_per_org')` serves as a regression guard against future re-introduction of constraint name interpolation.
- **Thorough edge-case coverage** (`test/support/errors.test.js:59-69`): null input, missing `.code`, and unmatched codes are all covered - the failure modes future callers will hit.
- **Correct architectural boundary**: Database-to-HTTP error translation lives in the storage-support layer. Storage functions own the DB dialect; controllers stay ignorant of Postgres error codes. The declarative map approach means adding a new PG error code in the future is a one-line entry, not a new control-flow branch.

## Issues

### Minor (Nice to Have)

1. **Numeric keys in `codeMap` vs string `.code` property** - `src/support/topics-storage.js:129-130`

   The `codeMap` is declared with numeric keys (`23505`, `23503`) but PostgREST returns `.code` as a string (`'23505'`). The lookup works because JavaScript object property access coerces both to strings, but this relies on an implicit coercion. The tests confirm the behavior works (they use string codes against numeric map keys), but the asymmetry could surprise a future contributor.

   **Why it matters:** If someone later converts the map to a `Map` object (for type safety or iteration order), numeric keys would no longer match string lookups without explicit coercion.

   **How to fix:** Use string keys in the map (`'23505'`, `'23503'`) to make the match explicit. Alternatively, add a one-line note in the JSDoc that keys are coerced to strings during lookup.

2. **Missing integration test for the new 422 path** - `test/it/`

   The CLAUDE.md states: "New or modified endpoints must include integration tests in `test/it/`." The PR modifies the error response of `POST /v2/orgs/:id/topics` (500 to 422 for FK violations, changed 409 message text). No integration test was added.

   **Mitigating context:** This is a storage-layer error-path change, not a new endpoint or request shape. The unit tests exercise the behavior thoroughly, and the brands controller test confirms propagation through the request lifecycle. The rule could reasonably be read as targeting new route definitions rather than error-path refinements on existing routes.

## Recommendations

- **Adopt `throwOnPgConstraintViolation` across other storage modules.** The PR description notes that the if/throw-on-PG-error-code pattern likely exists in other storage files (e.g., `categories-storage.js` referenced in the old diff comments). A follow-up pass to migrate those callers would make the error-translation pattern uniform across the service.

## Assessment

Ready to merge? Yes

This is a clean, well-executed iteration on the prior review feedback. Both previously flagged issues are fully resolved. The utility is minimal, well-documented, and placed at the correct architectural boundary. The security posture is strictly improved - two information disclosure paths removed. The two Minor observations are cosmetic and do not block merge. CI build should be confirmed green before merging.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 44s | <b>Cost:</b> $2.44 | <b>Commit:</b> <code>eeeb18ed983861d852c8c8d68b9c633352f31c8a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2544, opened 2026-06-03
fix: Revert "feat: resolve promise token from the header required to create a preflight job"

radhikagpt1208:
Reverts adobe/spacecat-api-service#2480


Kanishkavijay39 (approved):

MysticatBot (dismissed):
Hey @radhikagpt1208,

## Strengths

- Clean revert that restores cross-controller consistency between `preflight.js` and `scrapeJob.js` - both now use identical `resolvePromiseToken` logic (cookie first, IMS fallback).
- The cookie-first-then-IMS-fallback approach (`src/controllers/preflight.js:162-169`) is a graceful degradation pattern that serves both browser clients (cookie set by `/auth/promise`) and programmatic clients (Authorization header, IMS mints the token). Preferable to the hard 400 on missing header.
- Tests are well-structured: cookie path, IMS fallback, base64 token edge cases, and error propagation (both `ErrorWithStatusCode` and generic errors) are all covered.
- Security posture is neutral-to-positive: cookies carry `HttpOnly`/`Secure`/`SameSite` attributes that custom headers cannot leverage. The IMS fallback uses the already-authenticated Bearer token, introducing no new trust boundary.
- Reduces surface area by removing the single-purpose `src/support/http-headers.js` module and associated constants.

## Issues

### Important (Should Fix)

1. **OpenAPI spec documents a header the code no longer reads** - `docs/openapi/parameters.yaml:185-197`, `docs/openapi/preflight-api.yaml`

   The spec declares `x-promise-token` as an optional header parameter for the preflight endpoint. The implementation now ignores this header entirely and reads from the `promiseToken` cookie. API consumers reading the spec will believe sending `x-promise-token` does something; it does not. This violates the repo's CLAUDE.md "Specification Sync: Keep OpenAPI specs and implementation in sync" rule.

   **How to fix**: Remove the `xPromiseToken` parameter reference from the preflight endpoint in `preflight-api.yaml`, or document the cookie-based mechanism and deprecate the header parameter.

2. **traceId test removed but production code remains** - `test/controllers/preflight.test.js` (removed lines 205-237)

   The test "includes traceId on SQS message when controller ctx includes traceId" was removed, but the source code that spreads `traceId` into the SQS message remains at `src/controllers/preflight.js:239`: `...(ctx.traceId && { traceId: ctx.traceId })`. This means a production behavior (forwarding traceId to SQS) now has zero test coverage. Either restore the test (if traceId was intentionally kept) or remove the source line (if the revert should have removed it too).

3. **Orphaned JSDoc comment above wrong function** - `src/controllers/preflight.js:95-109`

   The `createPreflightJob` JSDoc block (describing cookie-based promise token resolution) is positioned above `isAuditEnabledForSite`, not above `createPreflightJob`. A revert artifact where the comment got displaced. A reader will see a 14-line JSDoc describing cookie resolution immediately followed by an unrelated function about audit handler enablement.

   **How to fix**: Move the JSDoc block to immediately above `const createPreflightJob = async (context) => {`.

### Minor (Nice to Have)

1. **Inlined `getHeaderCaseInsensitive` drops null-safety guard** - `src/controllers/consumers.js:40-44`

   The inlined version omits the `isNonEmptyObject(headers)` check that the deleted shared version had. The call site provides `|| {}` so this is safe in context, but the function is less defensive standalone. A one-line `if (!headers) return undefined;` guard would match the predecessor's robustness.

## Assessment

Ready to merge? With fixes
Reasoning: The revert is architecturally sound and restores a consistent, well-tested pattern. The three Important items are straightforward to address: reconcile the OpenAPI spec with the actual implementation, resolve the traceId test/code mismatch, and reposition the misplaced JSDoc.

## Next Steps

1. Decide whether traceId is intentional (restore test) or accidental (remove source line).
2. Remove or update the `xPromiseToken` parameter in the OpenAPI spec.
3. Move the orphaned JSDoc to its correct position.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 50s | <b>Cost:</b> $8.49 | <b>Commit:</b> <code>3fa4dfcdab5daff2a365ceb97199f0e86391d682</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2547, opened 2026-06-03
feat(serenity): server-side tag filtering for prompts list

rainer-friederich:
## What

Adds server-side tag filtering to `GET …/serenity/prompts` and wires the OpenAPI contract to match.

- **`tagIds` query param** — repeated keys (`?tagIds=a&tagIds=b`, max 50) are collected in `extractQuery` via `getAll()` and forwarded to the upstream `by_tags` call as `tag_ids` (previously hardcoded `[]`). Semrush applies OR semantics across multiple tag IDs.
- **`tagMap` DTO field** — each prompt now carries a `tagMap` (tag name → upstream tag UUID) so the MFE can render tag names and send the IDs back as `tagIds` to filter.
- **Accurate last-page total** — when the returned item count is below the page limit we compute the exact filtered total as `(page-1)*limit + items.length` instead of trusting `resp.total`, which Semrush may report as the unfiltered project count.
- **OpenAPI wiring** — `tagIds` declared as `style: form, explode: true` array query param on `listSerenityPrompts`; `tagMap` added to `SerenityPrompt` schema.

## Review fixes

- **`MAX_TAG_IDS = 50` cap** — `tagIds` is now sliced before `map/filter` to prevent an unbounded upstream payload from a misbehaving IMS-authenticated caller.
- **`extractQuery` multi-value guard** — `tagIds` is excluded from the `for…of` loop (last-write-wins) and collected exclusively via `getAll()`. Comment documents the pattern for future multi-value params.

## Why the OpenAPI commit is separate

The feature commits added the `tagIds` param and `tagMap` field without touching the contract, which violates the repo's OpenAPI-first / spec-sync rule. The follow-up commit closes the spec gap and makes the drift observable by adding `tagMap` + `tagIds` to the `listSerenityPrompts` fixture.

## Tests

- Handler unit tests: `tagIds` forwarded when present, empty `tag_ids` when absent, `tagMap` in the DTO, `MAX_TAG_IDS` slice.
- Controller unit tests: repeated-key `tagIds` parsing in `extractQuery` (present → array; absent → omitted).
- Contract suite: `listSerenityPrompts` fixture exercises `tagMap`/`tagIds` against the documented schema.
- `npm run docs:lint` passes.


MysticatBot (approved):
Hey @rainer-friederich,

## Strengths

- Well-scoped, focused change: one feature (tag filtering) wired end-to-end from OpenAPI spec through controller, handler, DTO, and tests with no unrelated changes mixed in.
- OpenAPI contract updated in lockstep (`docs/openapi/schemas.yaml`, `docs/openapi/serenity-api.yaml`). The `style: form, explode: true` declaration correctly describes the repeated-key wire format.
- Clean passthrough architecture: filtering is delegated to the upstream `listPromptsByTags` call rather than reimplemented locally, keeping the API service as a thin orchestration layer.
- Input sanitization at the handler boundary (`src/support/serenity/handlers/prompts.js:102-104`): `query.tagIds.map(String).filter(Boolean)` normalizes and strips empty strings before forwarding downstream.
- `buildTagMapOf` is appropriately defensive (`src/support/serenity/handlers/prompts.js:50-58`) - guards against non-array `tags`, non-object entries, and missing `name`/`id` fields.
- Test coverage is layered well: controller tests verify `extractQuery` multi-value parsing, handler tests verify `tag_ids` forwarding, and the contract test exercises `tagMap`/`tagIds` against the documented schema.

## Issues

### Minor (Nice to Have)

1. **No upper bound on `tagIds` array length** - `src/support/serenity/handlers/prompts.js:102-104`

   A client can send an unbounded number of repeated `?tagIds=x` query params. While the endpoint is IMS-authenticated (limiting the attacker population), a misbehaving client could send thousands of values that flow to the upstream service uncapped.

   This matches a pattern already in the file where `limit` is capped via `Math.min(requestedLimit, MAX_PAGE_LIMIT)`. Consider:
   ```javascript
   const MAX_TAG_IDS = 50;
   const tagIds = Array.isArray(query?.tagIds)
     ? query.tagIds.slice(0, MAX_TAG_IDS).map(String).filter(Boolean)
     : [];
   ```

2. **`extractQuery` hard-codes `tagIds` as the only multi-value param** - `src/controllers/serenity.js:64-68`

   The existing `for...of` loop yields only the last value per repeated key, then `tagIds` is special-cased with `getAll()`. If a second multi-value param is added later, a developer will hit the last-write-wins bug unless they know to add another `getAll()` block. A brief comment noting this contract (or skipping `tagIds` in the loop with `if (k === 'tagIds') continue;`) would make the override order explicit.

## Assessment

Ready to merge? Yes
Reasoning: The code is clean, well-tested, and the API contract is properly wired. No bugs, no data loss risk, no security concerns. The minor findings are hardening opportunities, not blockers.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 49s | <b>Cost:</b> $2.51 | <b>Commit:</b> <code>d170e77aa9c52604e6f5977add11111c3f08d6ba</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (changes requested):
Hey @rainer-friederich,

## Strengths

- Previously flagged issues now addressed: `MAX_TAG_IDS = 50` cap (`src/support/serenity/handlers/prompts.js:22`) cleanly mirrors the existing `MAX_PAGE_LIMIT` pattern, and the `extractQuery` guard (`src/controllers/serenity.js:62-65`) with its comment makes the multi-value contract explicit for future maintainers.
- Clean boundary between wire format and upstream API: `tagIds` (camelCase, API convention) mapped to `tag_ids` (snake_case, Semrush convention) at a single translation point (`src/support/serenity/handlers/prompts.js:143`). No leakage of upstream naming into the public contract.
- Input sanitization at `src/support/serenity/handlers/prompts.js:112-114` chains `.slice(0, MAX_TAG_IDS).map(String).filter(Boolean)` in correct order - cap, normalize, clean.
- Thin orchestration layer preserved: filtering is delegated to the upstream Semrush `by_tags` endpoint rather than reimplemented locally. The API service stays a passthrough, which is the correct boundary for this tier.
- `tagMap` in the DTO is a good contract choice (`docs/openapi/schemas.yaml`): exposing name-to-id mapping lets the MFE render human-readable tags and send back machine IDs without a separate lookup call.
- Last-page total correction (`src/support/serenity/handlers/prompts.js:146-151`) protects the MFE from pagination drift when the upstream total reports unfiltered counts. Comment explains both what and why.
- `buildTagMapOf` (`src/support/serenity/handlers/prompts.js:50-58`) is appropriately defensive at the system boundary - guards non-array, non-object entries, and missing fields.
- OpenAPI wiring uses `style: form, explode: true` which is the correct declaration for repeated query keys.
- Test coverage is layered at the right boundaries: controller tests verify `extractQuery` parsing, handler tests verify forwarding semantics, contract tests verify schema compliance.

## Issues

### Important (Should Fix)

1. **Missing integration test for modified endpoint behavior** - `test/it/`

   The repo's CLAUDE.md rule states: "New or modified endpoints must include integration tests in `test/it/`". This PR adds a `tagIds` filter parameter that changes which data the endpoint returns. No files under `test/it/` are in the changeset.

   Note: since the filtering happens upstream at Semrush (not in the local PostgreSQL layer), the value of an IT here is debatable - it would largely verify the same passthrough mechanics the unit tests already cover. However, the rule is explicit. If an existing serenity IT already exercises `listPrompts` and can be extended with a `tagIds` parameter, that would satisfy the rule with minimal effort.

### Minor (Nice to Have)

1. **No unit test for `MAX_TAG_IDS` cap enforcement** - `src/support/serenity/handlers/prompts.js:112-113`

   The `.slice(0, MAX_TAG_IDS)` guard is untested. A test passing 51+ tagIds and asserting only 50 reach `transport.listPromptsByTags` would verify the cap survives refactoring. Without it, someone could remove the slice without a test failure.

2. **No dedicated test for last-page total computation** - `src/support/serenity/handlers/prompts.js:146-151`

   The new logic `(page - 1) * limit + items.length` when `items.length < limit` is a behavioral change (previously it trusted `resp.total`). No test exercises a response where items returned < limit and verifies the computed total differs from `resp.total`. If this logic regresses, filtered results would show incorrect pagination totals.

## Assessment

Ready to merge? With fixes
Reasoning: The code is clean, well-designed, and the OpenAPI contract is properly wired. The Important finding is a documented convention that may not apply cleanly to this case (upstream filtering vs local DB), but should be consciously addressed or explicitly waived. The two Minor items are hardening tests that reduce future regression risk. CI failures (deploy-stage, it-postgres) appear infrastructure-related rather than caused by this PR.

## Next Steps

1. Evaluate the integration test requirement - if an existing serenity IT covers `listPrompts`, extend it with `tagIds`; otherwise, document why the convention does not apply here (upstream-only filtering).
2. Consider adding the two minor test cases for cap enforcement and last-page total logic.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 27s | <b>Cost:</b> $2.79 | <b>Commit:</b> <code>1e56a953c68274275f75090777342420a49bfe70</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (changes requested):
Hey @rainer-friederich,

## Strengths

- Previously flagged issues now fully addressed: `MAX_TAG_IDS = 50` cap with test (`test/support/serenity/handlers/prompts.test.js` "slices tagIds to MAX_TAG_IDS"), dedicated last-page total tests ("uses resp.total when returned item count equals the page limit" / "computes exact total from items when on the last (partial) page"), and integration test (`test/it/shared/tests/serenity.js` "401s on prompts endpoint with tagIds params").
- Clean pass-through architecture: filtering is delegated to the upstream Semrush `by_tags` endpoint rather than reimplemented locally. This service stays a thin orchestration layer, which is the correct boundary for this tier.
- Defensive input bounding (`src/support/serenity/handlers/prompts.js:112-114`): `slice(0, MAX_TAG_IDS).map(String).filter(Boolean)` caps, coerces, and cleans in correct order - bounds upstream payload size even with garbage input.
- `buildTagMapOf` (`src/support/serenity/handlers/prompts.js:50-58`) is pure, null-safe, and guards against non-array, non-object, and missing-field cases without throwing.
- Multi-value param extraction (`src/controllers/serenity.js:62-71`) uses `getAll()` with explicit exclusion from the single-value loop - the correct pattern, with a comment guiding future multi-value params.
- Last-page total heuristic (`src/support/serenity/handlers/prompts.js:146-151`) is well-commented and solves a real Semrush quirk (unfiltered total leaking through on filtered queries).
- `tagMap` DTO design eliminates a client round-trip for tag ID resolution - simple flat string-to-string object, JSON-serializable, stable.
- Well-scoped change: exactly the files required for this feature touched, no unrelated drive-by changes.
- OpenAPI contract uses `style: form, explode: true` which correctly describes the repeated-key wire format.
- Tests verify behavior (asserting on `result.total`, `body.tag_ids`) rather than just stub interactions.

## Issues

### Important (Should Fix)

1. **OpenAPI schema missing `maxItems: 50` on `tagIds` array** - `docs/openapi/serenity-api.yaml:72-76`

   The code enforces `MAX_TAG_IDS = 50` by silently truncating, but the OpenAPI schema declares `tagIds` as `type: array, items: { type: string }` with no `maxItems`. Consumers reading the spec have no way to know the limit exists or that exceeding it produces silent truncation rather than a 400.

   Why it matters: if the MFE or another integrator sends 60 tag IDs expecting all 60 to apply, they get silently incomplete results with no indication their filter was partially ignored. This is a one-line schema fix that prevents a class of consumer-side confusion.

   Fix:
   ```yaml
   schema:
     type: array
     maxItems: 50
     items: { type: string }
   ```

   Also consider whether the server should return 400 on over-limit rather than silently truncating - truncation with success is an anti-pattern for filter params (it misrepresents what the response contains). Either approach is fine, but document the chosen behavior in the `tagIds` description.

### Minor (Nice to Have)

2. **`docs/serenity.md` endpoint table incomplete** - The endpoint surface table lists `/serenity/prompts?geoTargetId=&languageCode=&page=&limit=&search=` without the new `tagIds` parameter. Update the table to include `&tagIds=` so operator runbooks reflect the new filter.

3. **No test exercises `buildTagMapOf` defensive guards** - `src/support/serenity/handlers/prompts.js:50-58`

   The function handles malformed tag objects (missing `name`, missing `id`, non-object entries), but existing tests only exercise the happy path. A single test with `tags: [null, {name: 'x'}, {id: 'y'}, {name: 'valid', id: 42}]` asserting `tagMap === {valid: '42'}` would lock in the filtering contract and prevent someone from removing the guards later.

4. **No format validation on tagIds values** - `src/support/serenity/handlers/prompts.js:112-114`

   Values are coerced to String and filtered for emptiness, but no regex ensures they look like Semrush tag UUIDs. The risk is low (JSON-encoded body, IMS auth gate, mature upstream SaaS), but a `/^[\w-]{1,128}$/` check would be trivial defense-in-depth. Optional hardening, not blocking.

## Assessment

Ready to merge? With fixes
Reasoning: The code is clean, well-tested, and architecturally sound. The one blocking item is the OpenAPI `maxItems` gap - a one-line fix that documents an enforced contract limit consumers cannot otherwise discover. The minor items are optional hardening.

## Next Steps

1. Add `maxItems: 50` to the `tagIds` schema in `docs/openapi/serenity-api.yaml`.
2. Optionally update `docs/serenity.md` endpoint table and add a `buildTagMapOf` edge-case test.


---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 8s | <b>Cost:</b> $4.38 | <b>Commit:</b> <code>92b308fc328d1a11adeb34df2b7697109084797b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

## Strengths

- Previously flagged issues now fully addressed: `maxItems: 50` on the `tagIds` schema (`docs/openapi/serenity-api.yaml:72-73`), `docs/serenity.md` endpoint table updated with OR semantics and max-50 constraint, and `buildTagMapOf` edge-case test added exercising null, numeric, missing-name, string fallback, and id-coercion paths.
- Clean delegation pattern preserved: tag filtering is correctly delegated to the upstream Semrush `by_tags` endpoint. The orchestration layer stays thin (parse, cap, forward, map response) with no local filtering or caching of tag state.
- Schema evolution from `tags: string[]` to `tagMap: object` (`docs/openapi/schemas.yaml:10069-10077`) is the right structural move. Exposing both display name (key) and filterable UUID (value) in a single response field eliminates a client round-trip for tag ID resolution.
- Input bounding is well-implemented: `slice(0, MAX_TAG_IDS).map(String).filter(Boolean)` caps cardinality, normalizes type, and removes empty strings. Transport layer forwards as a JSON array in a POST body with no injection grammar.
- Defensive total computation (`src/support/serenity/handlers/prompts.js:146-151`) distrusts the upstream `total` on partial pages - sound engineering for a known Semrush quirk where unfiltered counts leak through on filtered queries.
- Good test assertion specificity: the new `buildTagMapOf` test uses `deep.equal` on the exact expected output, proving both inclusion of valid entries and exclusion of invalid ones in a single pass. Behavior-level assertion, not an interaction check.
- CLAUDE.md rules respected: integration test added per "New or modified endpoints must include integration tests in test/it/" requirement, OpenAPI contract updated, DTO layer properly used.

## Issues

### Minor (Nice to Have)

1. **`tagMap` key collision on duplicate tag names** - `src/support/serenity/handlers/prompts.js:50-58`

   `buildTagMapOf` uses `t.name` as the object key. If Semrush ever returns two tags with the same display name but different IDs (e.g. two categories both named "Brand" in different tag groups), the second silently overwrites the first. Risk is low today (tag names appear globally unique within a project), but consider adding a one-line JSDoc note in the schema description: "tag names are assumed unique per project; if duplicates exist, only the last is retained."

2. **Empty-string tag case untested** - `test/support/serenity/handlers/prompts.test.js`

   The production code guards `typeof t === 'string' && t`, meaning an empty string is filtered out. No test exercises this path. Adding `''` to the tags array in the `buildTagMapOf` test fixture and confirming it does not appear in the output would close the gap cheaply with no change to the expected result.

3. **No format validation on tagIds values** (carried from prior review, unresolved) - `docs/openapi/serenity-api.yaml:74`

   The schema declares `items: { type: string }` with no `pattern`. Since these are Semrush UUIDs per the `tagMap` description, a `pattern` or `maxLength` would reject garbage at the OpenAPI validation layer rather than passing arbitrary strings upstream. Practical exploitability is negligible given IMS auth gating and JSON transport, so this remains optional hardening.

## Assessment

Ready to merge? Yes
Reasoning: All three previously flagged findings are addressed. The code is clean, well-tested, and architecturally sound. The remaining minor items are optional hardening and documentation opportunities, not blockers. CI pending checks appear unrelated to this change.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 0s | <b>Cost:</b> $2.48 | <b>Commit:</b> <code>e423970123a1dd8737208892ade709ef995aeba3</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

## Strengths

- Previously flagged items now fully addressed: `maxLength: 128` on tagIds items (`docs/openapi/serenity-api.yaml:73`), duplicate-key documentation note on `tagMap` (`docs/openapi/schemas.yaml`), and empty-string edge case added to the `buildTagMapOf` test fixture (`test/support/serenity/handlers/prompts.test.js`).
- Clean input pipeline: `slice(0, MAX_TAG_IDS).map(String).filter(Boolean)` caps cardinality, normalizes type, and removes empty strings in correct order. JSON serialization of the resulting array to the upstream Semrush POST body has no injection surface.
- Well-documented API contract: the OpenAPI `tagIds` parameter includes `maxItems: 50`, `maxLength: 128`, `style: form`, `explode: true`, and a description referencing the server-side constant. Consumers know exactly what to send and what happens on overflow.
- Thin orchestration layer preserved: filtering is delegated to upstream Semrush `by_tags` endpoint rather than reimplemented locally. The API service stays a passthrough at the correct boundary.
- `buildTagMapOf` is pure, null-safe, and guards against non-array, non-object, and missing-field cases without throwing. The `reduce` semantics match the documented last-write-wins contract for duplicate tag names.
- Last-page total heuristic is well-tested with dedicated assertions for both branches (full page trusts `resp.total`, partial page computes from offset).
- Multi-value param integration test proves repeated `tagIds` params survive the full controller pipeline without crashing before auth.
- All CLAUDE.md conventions respected: integration test added, OpenAPI contract updated in lockstep, docs/serenity.md endpoint table updated, no bundle constraint violations.

## Assessment

Ready to merge? Yes
Reasoning: The incremental commit is a clean response to prior review findings - schema documentation, input length bound, and test coverage for an edge case. The full diff is well-scoped, well-tested (all lines covered per Codecov), and follows established codebase patterns throughout.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 39s | <b>Cost:</b> $2.82 | <b>Commit:</b> <code>70a2820f44a8181d7a8d904140474a29d6cc7658</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2555, opened 2026-06-04
fix(slack/run-audit): set onDemand bypass and align entitlement to siteEnrollment

tathagat2241:
## Summary
`run audit` is an explicit, one-off Slack command — it should run regardless
of whether the site appears in the handler's enabled-list, and without
mutating that list. Today, even after the audit-worker enabled-list gate is
removed (companion PR), api-service still has two issues that prevent the
end-to-end fix:
1. The prerender CSV path still pre-checks `isHandlerEnabledForSite` and
   short-circuits with `:x: Will not audit site '<url>' because audits of
   type '<type>' are disabled for this site.`.
2. The entitlement check uses `tierResult.entitlement` (org-level), while
   the audit-worker downstream gate (`checkProductCodeEntitlements`) checks
   `tierResult.siteEnrollment` (site-level). A site can pass the
   api-service check and then be silently dropped in audit-worker, producing
   a confusing user experience.
This PR fixes both issues and emits the `onDemand: true` flag in the SQS
message so the audit-worker bypasses its enabled-list gate.

## What changed
- `src/support/slack/commands/run-audit.js`
  - Pass `auditContext: { onDemand: true }` to every `triggerAuditForSite`
    call emitted by `run-audit` — single audit, `audit:all`, and the
    prerender CSV path.
  - Drop the redundant `configuration.isHandlerEnabledForSite` precheck from
    the prerender CSV path so it behaves consistently with the single-audit
    path.
  - Switch entitlement check from `tierResult.entitlement` →
    `tierResult.siteEnrollment` in both `runAuditForSite` and
    `runPrerenderAuditForUrls`. This matches the audit-worker's downstream
    gate and avoids the previously silent failure when org-level
    entitlement existed but the specific site was not enrolled.
- `test/support/slack/commands/run-audit.test.js`
  - Updated existing positive-path mocks from `entitlement` →
    `siteEnrollment`.
  - Replaced the obsolete "blocks prerender CSV audits when the prerender
    handler is disabled" test with a positive test that verifies the run
    proceeds via the `onDemand` bypass.
  - New assertions that the SQS payload's `auditContext` contains
    `onDemand: true` for single, `audit:all`, and prerender flows.
  - New regression test: org has `entitlement` but no `siteEnrollment` ⇒
    audit is blocked (parity with audit-worker).
## Behavior matrix after this PR
| Path                    | Enabled-list check (api-service) | Entitlement key checked | Sets `onDemand` |
|-------------------------|:--------------------------------:|:-----------------------:|:---------------:|
| Slack `run audit`       | ❌ (removed; bypassed downstream) | `siteEnrollment`        | ✅              |
| Slack `run audit` (CSV prerender) | ❌ (removed; bypassed downstream) | `siteEnrollment` | ✅              |
| Slack `add site` / `add repo`     | ✅ (pre-filter)              | n/a (delegated)         | ❌              |
| Scheduled trigger (`triggerFromData`) | ✅ (pre-filter via `Configuration`) | n/a (delegated) | ❌              |
## Test plan
- [x] `npm test` — 11298 passing, 0 failing.
- [x] `npm run lint` clean.
- [ ] Manual verification in stage with the companion audit-worker PR
      deployed:
  - `@spacecat run audit <site-not-in-enabled-list>` for an entitled site →
    audit runs to completion; the site is **not** added to any
    enabled/disabled list.
  - `@spacecat run audit <site>` for an unentitled site →
    `:x: Will not audit site '<url>' because site is not entitled for this audit.`
  - Scheduled audit for the same not-in-enabled-list site → still skips
    (unchanged).
## Risk
- Low, contingent on the companion audit-worker PR being deployed first or
  at the same time. If api-service is deployed alone, the `onDemand` flag
  is simply ignored downstream and behavior is identical to today (the
  enabled-list check still runs in audit-worker for sites in the list, and
  one-off runs still fail for sites that are not — i.e. no regression, just
  no fix yet).
- The `entitlement → siteEnrollment` change tightens the api-service
  precheck. Any site that previously passed via org-level entitlement but
  had no site enrollment was already failing silently downstream, so this
  only changes _where_ the rejection happens (api-service now returns a
  clear Slack message instead of a silent worker skip).
## Notes for reviewers
- Scheduled audit triggers (`src/controllers/trigger/common/trigger.js`)
  remain unchanged and still pre-filter by `isHandlerEnabledForSite`. They
  do not set `onDemand`, so the audit-worker bypass does not apply to them.
- `triggerAuditForSite` in `src/support/utils.js` already accepts and
  forwards an `auditContext`, so no signature change was needed there.

## Related Issues

Thanks for contributing!


src/support/slack/commands/run-audit.js
```diff
@@ -155,7 +155,17 @@ function RunAuditCommand(context) {
...
+              // `run audit` is an explicit one-off command — pass onDemand so the
+              // audit worker bypasses the per-site handler enabled-list check
+              // (entitlement is still enforced downstream).
+              await triggerAuditForSite(
+                site,
+                enabledAuditType,
+                undefined,
+                slackContext,
+                context,
+                { onDemand: true },
```
rpapani: we shouldn't be sending any additional flags, audit-worker should simply check, if its not disabled, then just run it.

src/support/slack/commands/run-audit.js
```diff
@@ -170,27 +180,38 @@ function RunAuditCommand(context) {
...
 
-        await triggerAuditForSite(site, auditType, auditData, slackContext, context);
+        // One-off run: bypass enabled-list check downstream (entitlement already verified here).
+        await triggerAuditForSite(
+          site,
+          auditType,
+          auditData,
+          slackContext,
+          context,
+          { onDemand: true },
```
rpapani: same as above

src/support/slack/commands/run-audit.js
```diff
@@ -210,36 +231,41 @@ function RunAuditCommand(context) {
         return;
       }
 
-      if (!configuration.isHandlerEnabledForSite(auditType, site)) {
-        await say(`:x: Will not audit site '${baseURL}' because audits of type '${auditType}' are disabled for this site.`);
```
rpapani, line 340: we should use isHandlerDisabledForSite (which checks only disabled list) for genuine cases where we want to disable a particular audit for a specific site.

rpapani (changes requested):

src/support/slack/commands/run-audit.js
tathagat2241: Fixed: removed onDemand flag.

src/support/slack/commands/run-audit.js
tathagat2241: Fixed: onDemand flag.

src/support/slack/commands/run-audit.js
tathagat2241: Created isHandleDisabledForSite in spaceat-shared and used it here.

rpapani (approved):

PR #2556, opened 2026-06-04
feat(bot-blocker): forward scraperConfig.headers to detectBotBlocker probe

ymadobe:
## Summary
The `/sites/:siteId/bot-blocker` controller now reads `site.getConfig()?.getScraperConfig?.()?.headers` and forwards them to `detectBotBlocker`. The probe now mirrors what the real scraper sends, so sites whose WAF allowlists scraper traffic via a custom header (e.g. Akamai Bot Manager requiring `Accept-Language`) report `crawlable: true` correctly.

Defensive `?.` chain matches the pattern in [spacecat-scrape-job-manager#183](https://github.com/adobe/spacecat-scrape-job-manager/pull/183) — older `@adobe/spacecat-shared-data-access` versions without `getScraperConfig` silently fall back to `{}`, preserving today's behavior for sites without custom headers.

Depends on `@adobe/spacecat-shared-utils@1.117.0` ([adobe/spacecat-shared#1650](https://github.com/adobe/spacecat-shared/pull/1650)) which added the optional `headers` param to `detectBotBlocker`.

## Related Issues
- LLMO-4975 — careinsurance.com "Unlock Opportunities" card stuck in error state even though scrapes succeed after `scraperConfig.headers` allowlist

## Test plan
- [x] 2 new unit tests: scraperConfig.headers passthrough + old data-access fallback
- [x] Existing 9 tests updated for new call signature
- [x] All 11 controller tests pass
- [x] ESLint clean
- [x] `npm run docs:lint` validates

## Verification after merge
1. Hit `GET /sites/{careHealthSiteId}/bot-blocker` directly — should return `crawlable: true`
2. Refresh `llmo.now/.../brands-management?...&brandTab=cdn` in elmo UI — "Unlock Opportunities" card flips from Error → green
3. Spot-check 1-2 sites *without* `scraperConfig.headers` to confirm no regression

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,7 +33,7 @@
         "@adobe/spacecat-shared-slack-client": "1.6.7",
         "@adobe/spacecat-shared-tier-client": "1.5.1",
         "@adobe/spacecat-shared-tokowaka-client": "1.19.0",
-        "@adobe/spacecat-shared-utils": "1.116.6",
+        "@adobe/spacecat-shared-utils": "1.117.0",
         "@adobe/spacecat-shared-vault-secrets": "1.3.5",
         "@aws-sdk/client-s3": "3.1045.0",
         "@aws-sdk/client-secrets-manager": "3.1045.0",
@@ -10441,9 +10441,9 @@
       }
     },
     "node_modules/@adobe/spacecat-shared-utils": {
-      "version": "1.116.6",
-      "resolved": "https://registry.npmjs.org/@adobe/spacecat-shared-utils/-/spacecat-shared-utils-1.116.6.tgz",
-      "integrity": "sha512-yPohm6sZrLCnkinD6jgtYGMdFoJYy+fHM+MUkCXUzlVVgG+Rw0mpVjzGeZC+xH8hBUQ5gPYoUSP5fKHcaIT1Uw==",
+      "version": "1.117.0",
+      "resolved": "https://registry.npmjs.org/@adobe/spacecat-shared-utils/-/spacecat-shared-utils-1.117.0.tgz",
+      "integrity": "sha512-oQU+1pYaGHE9mbA432lEuNXAfF8QaqN0RnM6ox9lxovIhVmRy5kCDwbI+Vm8gLnRP//tD+QqCWFUafZlpBsjjw==",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/fetch": "4.3.0",
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "@adobe/spacecat-shared-slack-client": "1.6.7",
     "@adobe/spacecat-shared-tier-client": "1.5.1",
     "@adobe/spacecat-shared-tokowaka-client": "1.19.0",
-    "@adobe/spacecat-shared-utils": "1.116.6",
+    "@adobe/spacecat-shared-utils": "1.117.0",
     "@adobe/spacecat-shared-vault-secrets": "1.3.5",
     "@aws-sdk/client-s3": "3.1045.0",
     "@aws-sdk/client-secrets-manager": "3.1045.0",
--- a/src/controllers/bot-blocker.js
+++ b/src/controllers/bot-blocker.js
@@ -72,10 +72,18 @@ function BotBlockerController(ctx, log) {
         return internalServerError('Site has no baseURL configured');
       }
 
-      log.debug(`Checking bot blocker for site ${siteId} with baseURL: ${baseURL}`);
+      // Mirror what the real scraper sends so the probe is a faithful predictor
+      // of scrape success. Without this, sites that allowlist scraper traffic via
+      // a custom header (e.g. Akamai Bot Manager requiring `Accept-Language`)
+      // probe as blocked even though scrapes succeed.
+      // Defensive `?.` chain matches spacecat-scrape-job-manager — older
+      // `@adobe/spacecat-shared-data-access` versions without `getScraperConfig`
+      // silently fall back to `{}`, preserving today's behavior.
+      const customHeaders = site.getConfig()?.getScraperConfig?.()?.headers || {};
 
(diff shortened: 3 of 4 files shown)


MysticatBot (commented):
Hey @ymadobe,

## Strengths

- **Well-scoped, focused change** - the diff does exactly one thing: thread custom headers from site config into the bot-blocker probe. No unrelated cleanup or scope creep.
- **Backward-compatible by design** (`src/controllers/bot-blocker.js:78`): The defensive `?.` chain with `|| {}` fallback means older `@adobe/spacecat-shared-data-access` versions continue to work identically. No cross-service coordination required for safe deployment.
- **Cross-service consistency**: The probe now mirrors the real scraper's request shape (matching spacecat-scrape-job-manager), closing the gap between "can we crawl this?" and "will the scraper succeed?". A probe that diverges from the real operation is a false signal.
- **Safe logging** (`src/controllers/bot-blocker.js:81`): Only header *names* are logged (`Object.keys(customHeaders).join(',')`) - not values. Avoids leaking potentially sensitive header values into debug logs.
- **Thorough test coverage** (`test/controllers/bot-blocker.test.js:120-155`): Two behavior-named tests exercise the passthrough and old-data-access fallback. All existing tests updated for the new call signature.
- **Follows project conventions**: conventional commit title, factory function controller pattern, uses `site.getConfig()` per documented patterns, shared-utils usage, tests included for behavior change.

## Issues

### Minor (Nice to Have)

1. **test/controllers/bot-blocker.test.js - Missing edge case: `getConfig()` returns null/undefined**

   The production code uses `site.getConfig()?.getScraperConfig?.()?.headers || {}`. The "old data-access" test covers a config object missing `getScraperConfig`, but does not cover `getConfig()` returning `null` or `undefined`. The `?.` chain handles this correctly at runtime, but a test would document the contract explicitly and guard against future refactors.

   Fix: Add a case where `getConfig: () => null` and assert `detectBotBlockerStub` is still called with `{ baseUrl, headers: {} }`.

2. **test/controllers/bot-blocker.test.js - Missing edge case: `getScraperConfig()` returns object without `headers` key**

   If a site has a scraperConfig with other fields (e.g. `{ timeout: 5000 }`) but no `headers` property, the production code yields `undefined || {}`. This is a realistic configuration shape not exercised by any test.

   Fix: Add a case where `getScraperConfig` returns `{ timeout: 5000 }` (no `headers`) and verify the fallback to `{}`.

3. **src/controllers/bot-blocker.js:81 - Header key logging could grow unwieldy**

   `Object.keys(customHeaders).join(',')` will produce an arbitrarily long string if a site config carries many headers. This is debug-level so the blast radius is low, but worth noting if header sets grow over time. Fine as-is given it is appropriately gated behind `debug`.

## Assessment

Reasoning: Clean, focused feature addition. The defensive chaining is correct, the fallback to `{}` is well-motivated and tested, the dependency bump is intentional and explained, and no security concerns arise (headers flow from admin-managed config, same trust level as the existing `baseURL`). The minor test gaps are low-risk since the `?.` chain handles them correctly by construction.

Note: CI checks are currently failing (deploy-stage and branch-deploy jobs) - this should be investigated but appears unrelated to the code changes.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 39s | <b>Cost:</b> $2.03 | <b>Commit:</b> <code>9fcd527c5940dc26379de13a83cd882c97d0f4af</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anagarwa (approved):

MysticatBot (approved):
Hey @ymadobe,

## Strengths

- **Prior review feedback addressed promptly** (`test/controllers/bot-blocker.test.js`): Both edge-case tests requested in the previous review - `getConfig()` returning null and `scraperConfig` missing the `headers` key - are now present with clear comments explaining the defensive intent.
- **Defensive chain is well-calibrated** (`src/controllers/bot-blocker.js:79`): `site.getConfig()?.getScraperConfig?.()?.headers || {}` handles four distinct null/undefined states in a single expression. The `?.` on a method call is the correct pattern for a method that may not exist on older data-access versions, and the `|| {}` tail ensures the probe always receives a plain object.
- **Architecturally sound alignment with the scraper pipeline**: The probe now mirrors what the actual scraper sends, closing the gap between "can we crawl this?" and "will the scraper succeed?". A probe that diverges from the real request produces false positives that erode trust in the system.
- **Safe logging** (`src/controllers/bot-blocker.js:81`): Only header keys are logged (`Object.keys(customHeaders).join(',')`) - not values. Avoids leaking potentially sensitive header values into debug logs.
- **Sound trust model**: The `customHeaders` originate from admin-managed site configuration (same trust level as `site.getBaseURL()` already in use). No path for external/unauthenticated user input to influence these headers.
- **Thorough test coverage**: 4 new test cases exercise every meaningful branch of the single conditional chain - happy path passthrough, old data-access compat, null config, and missing headers key. Each test combines state assertions with boundary-interaction verification.
- **Follows project conventions**: Controller pattern, Lambda bundle constraints, and testing conventions all respected. No CLAUDE.md violations.

## Issues

No issues identified.

## Assessment

Ready to merge? Yes
Reasoning: Clean, focused feature addition. The defensive chaining is correct, the fallback to `{}` is well-motivated and tested, the dependency bump is intentional and explained, no security concerns arise, and all prior review feedback has been addressed. Codecov confirms all modified and coverable lines are covered.

Note: CI build is currently pending - confirm it passes before merging.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 32s | <b>Cost:</b> $2.65 | <b>Commit:</b> <code>90dc859cabc5fc37ad38c559369bc85d300d0bd3</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2558, opened 2026-06-05
feat(suggestions): allow S2S consumers with suggestion:write to reject suggestions

ravverma:
## Summary

- Adds `hasS2SCapability(CAP_SUGGESTION_WRITE)` check inside the `REJECTED` status gate in `patchSuggestionsStatus` — only that block is touched; `hasAccess` and the rest of the flow are unchanged
- Decision tree for `REJECTED` transitions:
  - **S2S with `suggestion:write`** → granted, info-level audit log
  - **S2S missing `suggestion:write`** → error-level audit log + immediate 403 with S2S-specific message (admin gate is never consulted)
  - **Non-S2S caller** → original `hasAdminAccess()` gate, completely untouched
- Introduces `CAP_SUGGESTION_WRITE = 'suggestion:write'` constant so the Layer 1 route guard (`required-capabilities.js`) and the new Layer 2 `hasS2SCapability` call reference the same string, preventing silent drift
- Updates all five `'suggestion:write'` string literals in `required-capabilities.js` to use the constant

## Test plan

- [ ] `npx mocha test/controllers/suggestions.test.js --grep "REJECTED"` — all 6 tests pass (4 existing + 2 new S2S cases)
- [ ] `npx mocha test/controllers/suggestions.test.js test/routes/capability-constants.test.js` — 445 passing, 0 failing

### DIFF
--- a/docs/openapi/site-opportunities.yaml
+++ b/docs/openapi/site-opportunities.yaml
@@ -478,6 +478,14 @@ site-opportunity-suggestions-status:
     operationId: updateSiteOpportunitySuggestionsStatus
     summary: |
       Update the status of one or multiple suggestions in one transaction
+    description: |
+      Updates the status of one or more suggestions in a single bulk transaction. Returns
+      a 207 multi-status response with per-item success or failure details.
+
+      **S2S access:** S2S consumers registered with the `suggestion:write` capability may
+      call this endpoint. The admin-only gate for `REJECTED` status transitions is bypassed
+      for S2S callers holding this capability; all other validations (site membership,
+      opportunity lookup, valid status transition) apply unchanged.
     tags:
       - opportunity-suggestions
     requestBody:
--- a/src/controllers/suggestions.js
+++ b/src/controllers/suggestions.js
@@ -927,16 +927,19 @@ function SuggestionsController(ctx, sqs, env) {
             const s2sResult = await accessControlUtil.hasS2SCapability(CAP_SUGGESTION_WRITE);
             if (s2sResult.allowed) {
               context.log.info(`[acl] S2S REJECT granted - suggestionId=${id} clientId=${s2sResult.clientId} consumerId=${s2sResult.consumerId}`);
-            } else if (!accessControlUtil.hasAdminAccess()) {
+            } else {
               if (s2sResult.reason !== 'not-s2s') {
-                context.log.info(`[acl] Denied REJECTED - reason=${s2sResult.reason} clientId=${s2sResult.clientId || 'n/a'} consumerId=${s2sResult.consumerId || 'n/a'}`);
+                // S2S call but missing the required capability — audit trail
+                context.log.error(`[acl] S2S REJECT denied - reason=${s2sResult.reason} clientId=${s2sResult.clientId || 'n/a'} consumerId=${s2sResult.consumerId || 'n/a'}`);
               }
-              return {
-                index,
-                uuid: id,
-                message: 'Only admins can reject suggestions',
-                statusCode: 403,
-              };
+              if (!accessControlUtil.hasAdminAccess()) {
+                return {
+                  index,
+                  uuid: id,
+                  message: 'Only admins can reject suggestions',
+                  statusCode: 403,
+                };
+              }
             }
 
             // Only allow REJECTED from PENDING_VALIDATION
--- a/src/controllers/suggestions.js
+++ b/src/controllers/suggestions.js
@@ -47,7 +47,7 @@ import {
   isViewAsTrialRequest,
 } from '../support/utils.js';
 import AccessControlUtil from '../support/access-control-util.js';
-import { CAP_FIX_ENTITY_CREATE } from '../routes/capability-constants.js';
+import { CAP_FIX_ENTITY_CREATE, CAP_SUGGESTION_WRITE } from '../routes/capability-constants.js';
 import { grantSuggestionsForOpportunity } from '../support/grant-suggestions-handler.js';
 import { postSlackMessage } from '../utils/slack/base.js';
 import { createAtomicStrategy, deleteAtomicStrategy } from '../support/atomic-strategy-helper.js';
@@ -923,8 +923,14 @@ function SuggestionsController(ctx, sqs, env) {
         if (currentStatus !== status) {
           // Validate REJECTED status transition
           if (status === SuggestionModel.STATUSES.REJECTED) {
-            // Check admin access for REJECTED status
-            if (!accessControlUtil.hasAdminAccess()) {
+            // S2S consumers with suggestion:write capability may also reject suggestions
+            const s2sResult = await accessControlUtil.hasS2SCapability(CAP_SUGGESTION_WRITE);
+            if (s2sResult.allowed) {
+              context.log.info(`[acl] S2S REJECT granted - suggestionId=${id} clientId=${s2sResult.clientId} consumerId=${s2sResult.consumerId}`);
+            } else if (!accessControlUtil.hasAdminAccess()) {
+              if (s2sResult.reason !== 'not-s2s') {
+                context.log.info(`[acl] Denied REJECTED - reason=${s2sResult.reason} clientId=${s2sResult.clientId || 'n/a'} consumerId=${s2sResult.consumerId || 'n/a'}`);
+              }
               return {
                 index,
                 uuid: id,
--- a/src/controllers/suggestions.js
+++ b/src/controllers/suggestions.js
@@ -992,6 +992,9 @@ function SuggestionsController(ctx, sqs, env) {
           };
         }
       } catch (e) {
+        if (e?.name !== VALIDATION_ERROR_NAME) {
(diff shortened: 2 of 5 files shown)


MysticatBot (changes requested):
Hey @ravverma,

## Strengths

- **Well-structured decision tree** (`src/controllers/suggestions.js:925-937`): The three-branch conditional (S2S allowed, S2S denied, non-S2S fallback to admin) is clear and easy to follow. Each branch has a distinct, documented outcome. A new team member will understand the authorization logic without needing to trace through multiple files.
- **Two-layer capability architecture**: Layer 1 (route guard in `required-capabilities.js`) gates HTTP access; Layer 2 (controller) enforces finer-grained business rules within the endpoint's action space. Neither layer alone is sufficient, and neither is redundant.
- **Constant extraction eliminates drift** (`src/routes/capability-constants.js:28`): Replacing five string literals with `CAP_SUGGESTION_WRITE` means a future rename of the capability string is a one-line change.
- **Existing tests updated** (`test/controllers/suggestions.test.js:3898, 3935, 3973, 4011`): All four pre-existing REJECTED tests now stub `hasS2SCapability` with the `not-s2s` response, preventing them from accidentally entering the new S2S branch.
- **Negative assertion on the denial path** (`test/controllers/suggestions.test.js:4095`): The test explicitly asserts `hasAdminStub` was never called, validating the documented behavior that "admin gate is never consulted" for S2S callers.

## Issues

### Important (Should Fix)

1. **`context.log.error` used for an expected authorization rejection** - `src/controllers/suggestions.js:932`

   An S2S consumer presenting valid credentials but missing a specific capability is not an unexpected error condition - it is a normal authorization denial, similar to a user without admin access. Using `log.error` here means on-call will see these in error dashboards and investigate them as system faults.

   **Why it matters**: A misconfigured consumer will generate a steady stream of ERROR-level logs that look like incidents but are actually expected rejections. This trains on-call to ignore errors from this service.

   **Fix**: Change to `context.log.warn`. If the team's convention is that all 403s from S2S are operationally significant enough to warrant error-level, add a brief comment explaining why this differs from the existing admin-denial path (which does not log at all).

2. **Missing test: `hasS2SCapability` throws/rejects** - `test/controllers/suggestions.test.js`

   The production code `await`s `accessControlUtil.hasS2SCapability(...)` but there is no test covering what happens when this call rejects (network error, unexpected exception from the access-control layer). If `hasS2SCapability` throws, the error will bubble up unhandled from within the per-suggestion loop. Whether that crashes the entire bulk operation or is caught by an outer handler is untested.

   **Why it matters**: A transient auth-service failure could turn a partial-success bulk operation into a 500 for the entire batch.

   **Fix**: Add a test where `hasS2SCapability` rejects with an Error, and assert the behavior matches expectations (either per-item error in the 207 response, or a top-level 500 - whichever is intended).

## Recommendations

- The info-level log on the granted path (`src/controllers/suggestions.js:929`) includes `suggestionId`, `clientId`, and `consumerId`. Consider adding `suggestionId` to the denied log (line 932) for symmetry - having both paths log the same field set makes log correlation easier during incident investigation.

## Assessment

Ready to merge? With fixes
Reasoning: The authorization logic is correct, the security model is sound (no new attack surface introduced), and the change is well-scoped. The two Important items are about operational quality (log level that will degrade monitoring signal) and resilience confidence (untested exception path in a bulk operation). Neither is a correctness bug, but both are worth addressing before merge.

## Next Steps

1. Consider changing `log.error` to `log.warn` for the S2S denial path (item 1).
2. Add a test for the `hasS2SCapability` rejection/throw scenario (item 2).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 18s | <b>Cost:</b> $2.87 | <b>Commit:</b> <code>37e92104d20dcdec21f4812fcc4703391f34b409</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (commented):
Hey @ravverma,

## Strengths

- **Prior findings resolved cleanly** (`1c66001`): The log level for S2S denial is now `context.log.warn` (previously `error`), and the test for `hasS2SCapability` throwing unexpectedly is in place with proper 500/failed-metadata assertions. Both prior Important findings are addressed.
- **Clear three-branch decision tree** (`src/controllers/suggestions.js:925-940`): S2S-allowed, S2S-denied, non-S2S-fallback. Each branch has a distinct documented outcome and the admin gate is never consulted for S2S consumers - fail-closed by design.
- **Correct error handling improvement** (`src/controllers/suggestions.js:995`): The catch block now returns 500 for non-validation errors instead of incorrectly masking them as 400 client errors. The `e?.name === VALIDATION_ERROR_NAME ? 400 : 500` pattern matches the three other catch blocks in this file (lines 653, 785, 1007).
- **Constant extraction eliminates drift** (`src/routes/capability-constants.js:29`): `CAP_SUGGESTION_WRITE` shared between Layer 1 (route guard) and Layer 2 (controller check) ensures a future rename is a single-line change.
- **Test coverage is thorough**: 3 new tests covering granted/denied/error paths, 4 existing tests updated to stub `hasS2SCapability` preventing false passes into the new S2S branch. Negative assertion on `hasAdminStub` validates the short-circuit semantics. Codecov confirms full line coverage.
- **Security model is sound**: S2S consumers missing the required capability get an immediate 403 without falling through to the admin check. No new attack surface introduced. Fresh DB lookup in `hasS2SCapability` prevents stale grants.

## Issues

### Minor (Nice to Have)

1. **OpenAPI description could document the new S2S authorization path** - `docs/openapi/site-opportunities.yaml`

   The CLAUDE.md "Specification Sync" rule asks that specs and implementation stay in sync. The auto-fix endpoint in the same spec file documents S2S access in its description ("S2S access: S2S consumers registered with the `fixEntity:create` capability may call this endpoint..."). Adding similar documentation for the REJECTED status transition would help S2S integrators discover the authorization rules without reading controller code.

   **Suggestion**: Add a note to the endpoint description following the existing auto-fix precedent. Not blocking since the API contract (request/response shape) is unchanged.

## Recommendations

- Consider adding `context.log.error` in the catch block (`src/controllers/suggestions.js:992-999`) for non-validation errors. Currently, when `hasS2SCapability` throws a transient error, the per-item 500 response is returned but the failure is not logged - making it invisible in monitoring unless the caller reports it.
- The `hasS2SCapability` call sits inside the per-suggestion loop (`src/controllers/suggestions.js:926`). Since the caller's S2S capability status is invariant within a single request, hoisting the check above the loop would avoid redundant DB lookups on bulk-reject payloads. Worth considering in a follow-up if bulk rejection volume grows.

## Assessment

Reasoning: The authorization logic is correct and well-tested, the security model is sound, both prior review findings are addressed, and the error handling improvement is a net positive. The minor suggestion about OpenAPI documentation is a convention alignment item that does not block merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 18s | <b>Cost:</b> $3.37 | <b>Commit:</b> <code>1c66001f39d94dfc7f8dabd2757ccd2820223587</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @ravverma,

## Strengths

- **Prior review findings cleanly addressed** (`7dae1db`): Both the OpenAPI documentation gap (Minor) and the error logging recommendation from the previous review are resolved in this commit.
- **OpenAPI description documents the S2S authorization path** (`docs/openapi/site-opportunities.yaml:481-485`): Clear, concise, and correctly scopes what the capability bypass covers versus what validations remain enforced. S2S integrators can now discover the authorization rules without reading controller code.
- **Error logging with validation guard** (`src/controllers/suggestions.js:994-996`): The `e?.name !== VALIDATION_ERROR_NAME` check avoids noisy error logs for expected validation rejections while ensuring genuinely unexpected failures are surfaced at the appropriate level. Good signal-to-noise discipline for a bulk 207 endpoint.
- **Test assertion for error log** (`test/controllers/suggestions.test.js:4126`): Uses `calledWithMatch` with a regex, which is appropriately loose (resilient to message wording changes) while still confirming the log path fires for the right code path.
- **Security model remains sound**: Two-layer authorization (route guard + controller-level `hasS2SCapability`) with explicit deny-by-default for the REJECTED transition. No new attack surface introduced by this commit.

## Issues

No issues found.

## Assessment

Ready to merge? Yes
Reasoning: The incremental commit directly addresses both prior review findings (OpenAPI documentation and error logging) without introducing any new concerns. The authorization logic, error handling, test coverage, and documentation are all correct and well-scoped. CI should be green before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 27s | <b>Cost:</b> $2.02 | <b>Commit:</b> <code>7dae1db3de73aa5ab9c9346e50deaee92039fd32</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2559, opened 2026-06-05
fix(configuration): allow S2S consumers with configuration:read to read configurations

rainer-friederich:
## Problem

`GET /configurations/latest` and `GET /configurations/:version` are gated solely by `accessControlUtil.hasAdminReadAccess()` in the controller. For IMS/S2S tokens that only passes when the token carries the `admin` scope. So an S2S consumer that holds the route's **declared** `configuration:read` capability still gets:

```
403 {"message":"Only admins can view configurations"}
```

The route already declares `configuration:read` in `required-capabilities.js` and `s2sAuthWrapper` (Layer 1) enforces it — but the controller's admin gate (Layer 2) is independent of that capability, so the capability grant is effectively dead for these read routes. The two layers are inconsistent.

Legacy api-key auth bypassed the admin gate (api-key is neither IMS nor JWT, so `hasAdminAccess()` returns `true`). So this was invisible until the **api-key → S2S migration** (api-key deprecation + S2S enablement): once a consumer authenticates with an S2S Bearer instead of the api-key, it loses the admin-equivalent treatment and hits the long-standing admin gate.

### Impact
This silently broke the Adobe LLMO **Data Retrieval Service** schedule-sync, which calls `GET /configurations/latest` to read brand-presence handler→site assignments. With both legacy auths dead for it (api-key 401, S2S 403), no new schedules were reconciled for ~11 days — new customer onboardings got no data-retrieval / brand-presence schedule. The DRS S2S consumer was registered with `configuration:read` in its access request (SITES-43722), but the controller admin gate ignored it.

Verified in prod (3-way auth test against both `llmo` and `spacecat` hosts):

| Auth | Result |
|---|---|
| legacy x-api-key | 401 (deprecated) |
| platform (internal-org) S2S Bearer | 403 `Only admins can view configurations` |
| org-scoped S2S Bearer | 403 `Only admins can view configurations` |

## Fix

Add a Layer-2 `hasS2SCapability('configuration:read')` opt-in on both read endpoints, mirroring the established readAll pattern in the `sites` / `organizations` controllers, via a shared `authorizeConfigRead` helper:

- Full and read-only admins pass exactly as before (admin bypass).
- Otherwise, an S2S consumer holding `configuration:read` is allowed (fresh consumer re-fetch via `hasS2SCapability`, consistent with the other controllers' Layer-2 re-check).
- The denial message (`Only admins can view configurations`) and all write-endpoint gating are **unchanged**.
- Introduces `CAP_CONFIGURATION_READ` as the single-source-of-truth constant shared by `required-capabilities.js` (Layer 1) and the controller (Layer 2), and wires it into the existing drift coverage.

Write endpoints (`PATCH/POST/PUT/DELETE /configurations/...`) remain admin-only — only the two `configuration:read` GET routes change.

## Tests

- New cases in `test/controllers/configurations.test.js`: S2S consumer with `configuration:read` is granted on `getLatest` and `getByVersion`; denied when missing the capability or revoked (with the `[acl]`/`[s2s]` log assertions).
- Existing admin / read-only-admin / non-admin denial tests pass unchanged.
- `test/routes/capability-constants.test.js` updated to include `configuration.js` as a known `hasS2SCapability` call site.
- Full suite green locally (`2598 passing`), lint clean.

## Context

Root-cause analysis and the prod evidence are tracked in LLMO-5341. This is the durable, least-privilege fix (honor the declared capability) rather than granting the DRS consumer a full `admin` scope.

### DIFF
--- a/src/controllers/configuration.js
+++ b/src/controllers/configuration.js
@@ -27,6 +27,7 @@ import { checkConfiguration } from '@adobe/spacecat-shared-data-access';
 
 import { ConfigurationDto } from '../dto/configuration.js';
 import AccessControlUtil from '../support/access-control-util.js';
+import { CAP_CONFIGURATION_READ } from '../routes/capability-constants.js';
 
 function ConfigurationController(ctx) {
   if (!isNonEmptyObject(ctx)) {
@@ -52,13 +53,40 @@ function ConfigurationController(ctx) {
   };
 
   /**
+   * Authorizes a configuration read. Full and read-only admins pass as before.
+   * Additionally, an S2S consumer holding the `configuration:read` capability is
+   * allowed — the capability is also enforced at Layer 1 by `s2sAuthWrapper`, so
+   * this re-check (a fresh consumer fetch via `hasS2SCapability`) is the Layer-2
+   * guard that mirrors the readAll pattern in the sites/organizations controllers.
+   * @param {UniversalContext} context - Request context.
+   * @param {string} route - Route label used in structured logs.
+   * @returns {Promise<Response|null>} A `forbidden` response when denied, else null.
+   */
+  const authorizeConfigRead = async (context, route) => {
+    const requestId = context?.invocation?.id || 'unknown';
+    const isAdmin = accessControlUtil.hasAdminReadAccess();
+    const s2sResult = isAdmin
+      ? { allowed: false, reason: 'admin-bypass' }
+      : await accessControlUtil.hasS2SCapability(CAP_CONFIGURATION_READ);
+    if (!isAdmin && !s2sResult.allowed) {
+      log.info(`[acl] Denied ${route} - reason=${s2sResult.reason} clientId=${s2sResult.clientId || 'n/a'} consumerId=${s2sResult.consumerId || 'n/a'} requestId=${requestId}`);
+      return forbidden('Only admins can view configurations');
+    }
+    if (s2sResult.allowed) {
+      log.info(`[s2s] ${route} granted clientId=${s2sResult.clientId} consumerId=${s2sResult.consumerId} capability=${CAP_CONFIGURATION_READ} requestId=${requestId}`);
+    }
+    return null;
+  };
+
+  /**
    * Retrieves the configuration identified by the given version.
    * @param {UniversalContext} context - Context of the request.
    * @return {Promise<Response>} Configuration response.
    */
   const getByVersion = async (context) => {
-    if (!accessControlUtil.hasAdminReadAccess()) {
-      return forbidden('Only admins can view configurations');
+    const denied = await authorizeConfigRead(context, 'GET /configurations/:version');
+    if (denied) {
+      return denied;
     }
     const configurationVersion = context.params?.version;
 
@@ -76,11 +104,13 @@ function ConfigurationController(ctx) {
 
   /**
    * Retrieves the latest configuration with schema validation.
+   * @param {UniversalContext} context - Context of the request.
    * @return {Promise<Response>} Configuration response.
    */
-  const getLatest = async () => {
-    if (!accessControlUtil.hasAdminReadAccess()) {
-      return forbidden('Only admins can view configurations');
+  const getLatest = async (context) => {
+    const denied = await authorizeConfigRead(context, 'GET /configurations/latest');
+    if (denied) {
+      return denied;
     }
     const configuration = await Configuration.findLatest();
     if (!configuration) {
--- a/src/routes/capability-constants.js
+++ b/src/routes/capability-constants.js
@@ -22,6 +22,8 @@
 export const CAP_SITE_READ_ALL = 'site:readAll';
 export const CAP_ORG_READ_ALL = 'organization:readAll';
 
+export const CAP_CONFIGURATION_READ = 'configuration:read';
+
 export const CAP_SITE_CREATE = 'site:create';
 
 export const CAP_FIX_ENTITY_CREATE = 'fixEntity:create';
--- a/src/routes/required-capabilities.js
+++ b/src/routes/required-capabilities.js
@@ -11,6 +11,7 @@
  */
 
 import {
+  CAP_CONFIGURATION_READ,
   CAP_FIX_ENTITY_CREATE,
   CAP_ORG_READ_ALL,
   CAP_SITE_CREATE,
@@ -223,10 +224,10 @@ const routeRequiredCapabilities = {
   'GET /consent-banner/:jobId': 'organization:read',
 
   // Configuration
-  'GET /configurations/latest': 'configuration:read',
+  'GET /configurations/latest': CAP_CONFIGURATION_READ,
   'PATCH /configurations/latest': 'configuration:write',
   'POST /configurations/:version/restore': 'configuration:write',
-  'GET /configurations/:version': 'configuration:read',
+  'GET /configurations/:version': CAP_CONFIGURATION_READ,
   'POST /configurations/audits': 'configuration:write',
   'DELETE /configurations/audits/:auditType': 'configuration:write',
   'PUT /configurations/latest/queues': 'configuration:write',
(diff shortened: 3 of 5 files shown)


MysticatBot (changes requested):
Hey @rainer-friederich,

## Strengths

- **Well-extracted shared helper** (`src/controllers/configuration.js:55-79`): `authorizeConfigRead` cleanly consolidates the two-layer authorization logic into one place, preventing future drift between `getLatest` and `getByVersion`. The pattern mirrors what exists in `sites`/`organizations` controllers, making this immediately recognizable to anyone familiar with the codebase.
- **Structured audit logging** (`src/controllers/configuration.js:67-77`): Both the denial and grant paths log clientId, consumerId, capability, route, and requestId. The `[acl]`/`[s2s]` prefixes make log filtering trivial. This is exactly what you need when debugging access issues in production.
- **Defense in depth maintained**: The Layer-2 re-check via `hasS2SCapability` performs a fresh DB fetch of the consumer record rather than trusting the in-context `s2sConsumer` object, so a revoked or deactivated consumer cannot ride a cached auth context. Write endpoints remain admin-only.
- **Single-source-of-truth constant** (`src/routes/capability-constants.js:24`): `CAP_CONFIGURATION_READ` eliminates string-literal duplication between Layer 1 and Layer 2, and the drift-coverage tests enforce the wiring immediately.
- **Drift-prevention tests** (`test/routes/capability-constants.test.js:68-80`): The new "every exported CAP_ constant is used by at least one route" test generalizes coverage and catches dead constants the moment they appear.
- **Change discipline**: The diff is tightly scoped - only read endpoints touched, write endpoints remain admin-only, no unrelated refactoring mixed in.
- **Conventions alignment**: The PR follows the established two-layer validation pattern documented in `docs/s2s/READALL_CAPABILITY_DESIGN.md` including the controller opt-in, fresh DB fetch, structured logging, and capability-constant import.

## Issues

### Important (Should Fix)

1. **`getByVersion` test passes a bare context object instead of the shared test context** - `test/controllers/configurations.test.js:358-366`

   The `getByVersion` grant test passes `{ params: { version: '1' }, invocation: { id: 'req-cfg-1' } }` directly, rather than using the `context` object configured in `beforeEach` (which sets `context.s2sConsumer`, `context.attributes.authInfo`, etc.). The test passes because `accessControlUtil` reads from the closure-captured `ctx`, not from the argument - but this is passing by accident rather than by design.

   **Why it matters:** If the controller is ever refactored to read auth state from the passed-in context argument (which is the natural expectation given the function signature accepts `context`), this test would silently pass via the wrong path or break unexpectedly. The `getLatest` tests already use the shared `context` correctly - this inconsistency is confusing.

   **Fix:** Use the shared `context` with overrides:
   ```js
   context.params = { version: '1' };
   const result = await configurationsController.getByVersion(context);
   ```

### Minor (Nice to Have)

2. **Drift test name no longer reflects its scope** - `test/routes/capability-constants.test.js:86`

   The test is named `'every readAll constant has an actual hasS2SCapability(CONST) call site'` but now scans `configuration.js` which uses `CAP_CONFIGURATION_READ` (a read capability, not readAll). Consider updating the name to reflect its broader scope (e.g., "every capability constant with an S2S opt-in has an actual hasS2SCapability call site").

3. **No test for the `requestId` fallback to `'unknown'`** - `src/controllers/configuration.js:62`

   The `authorizeConfigRead` helper falls back to `'unknown'` when `context?.invocation?.id` is missing. No test exercises this path. This only affects log output and is not a correctness issue, but a quick test would confirm the fallback works and document the expected behavior.

## Assessment

Ready to merge? With fixes

The authorization logic is correct, well-scoped, and follows established patterns. The one material concern is the `getByVersion` test using a bare context object - it works today by coincidence (closure-captured state) but is fragile and inconsistent with the adjacent `getLatest` tests. Worth a quick fix before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 4s | <b>Cost:</b> $3.20 | <b>Commit:</b> <code>be8c1e786b8eeef320fa99587cb1ef308a7ace0c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the review in 576bcfbf:

- **Important — `getByVersion` grant test used a bare context object:** the test now drives the shared `context` (with a `context.params = { version: '1' }` override) instead of `{ params, invocation }`, so it exercises the same `s2sConsumer` / `authInfo` auth state as the adjacent `getLatest` tests rather than passing via closure-captured `ctx`.
- **Minor — drift test name no longer matched its scope:** renamed to `every CAP_ constant with an S2S opt-in has an actual hasS2SCapability(CONST) call site (Layer 2 opt-in)`, reflecting that it now scans all `CAP_` constants (incl. `configuration:read`), not just `:readAll`.
- **Minor — no test for the `requestId` fallback:** added a case asserting `authorizeConfigRead` logs `requestId=unknown` when `context.invocation.id` is absent.

`test/controllers/configurations.test.js` + `test/routes/capability-constants.test.js` green (72 passing), lint clean.

MysticatBot (approved):
Hey @rainer-friederich,

## Strengths

- **All three prior findings addressed cleanly.** The `getByVersion` test now uses the shared `context` object (`test/controllers/configurations.test.js:362-364`), the drift test name accurately reflects its broader scope (`test/routes/capability-constants.test.js:81`), and the `requestId=unknown` fallback path has explicit coverage (`test/controllers/configurations.test.js:398-407`).
- **Well-extracted shared helper** (`src/controllers/configuration.js:64-80`): `authorizeConfigRead` consolidates the two-layer authorization logic cleanly, mirroring the established readAll pattern in sibling controllers. Immediately recognizable to anyone familiar with the codebase.
- **Defense in depth maintained**: Layer-2 re-check via `hasS2SCapability` performs a fresh consumer DB fetch rather than trusting the in-context `s2sConsumer` object. A revoked consumer cannot ride a stale auth context - this closes the TOCTOU window between Layer 1 and Layer 2.
- **Single-source-of-truth constant** (`src/routes/capability-constants.js:24`): `CAP_CONFIGURATION_READ` eliminates string-literal drift between Layer 1 and Layer 2. The new generalized drift-coverage test (`test/routes/capability-constants.test.js:68-80`) catches dead constants immediately and extends protection to all future capability additions.
- **Structured audit logging** (`src/controllers/configuration.js:72-77`): Both denial and grant paths log clientId, consumerId, capability, route, and requestId at INFO level. The `[acl]`/`[s2s]` prefixes make log filtering trivial. No user-controlled input reaches the log interpolation (route is hardcoded, IDs come from DB-fetched records).
- **Minimal blast radius**: Write endpoints remain admin-only. The change is strictly additive - a new grant path for a previously-dead capability without removing any existing gate.
- **Pattern consistency**: The approach follows the established two-layer authorization architecture documented in `docs/s2s/READALL_CAPABILITY_DESIGN.md`, adapted appropriately for a single-resource read capability. No new coupling, no new failure modes.
- **Behavior-first test assertions**: Each test asserts on observable outcomes (HTTP status, response body, structured log output) rather than purely on stub interactions. The `calledOnce` assertion on the DB lookup verifies the fresh-fetch contract at a real boundary.

## Issues

No issues found.

## Assessment

Ready to merge? Yes

The authorization logic is correct, well-scoped, and follows established patterns. All prior review findings have been addressed with minimal, well-targeted changes. The test suite covers the grant path (both endpoints), denial paths (missing capability, revoked consumer), and the requestId fallback. The drift-coverage tests prevent future regressions. This is a clean fix for a real production issue with minimal blast radius - read endpoints only, write endpoints unchanged.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 29s | <b>Cost:</b> $2.35 | <b>Commit:</b> <code>576bcfbf17aab16f35acac72afd4203380e5f7b1</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1674, opened 2026-01-06
feat: decouple Slack one-off audit execution from site enablement

tathagat2241:
## 🎯 Objective

Allow users to run one-off audits via Slack without permanently enabling sites for scheduled audits. This addresses the operational issue where Customer Success teams forget to disable sites after running one-off audits, leading to:
- Unnecessary scheduled audit runs
- Increased infrastructure costs

#### `src/support/slack/commands/run-audit.js`
- **Removed**: `isHandlerEnabledForSite` check (lines 144-147)
- **Removed**: Filtering for `enabledAudits` when `auditType === 'all'` (lines 124-126)
- **Result**: Slack-triggered audits now run without checking site enablement
- **Unchanged**: Entitlement checks, product code validation, site lookup

#### `src/support/slack/commands/toggle-site-audit.js`
- **Added**: Deprecation message for `audit enable` command (returns early)
⚠️ **`audit enable` command**: Now shows deprecation message and returns
- **Removed**: All enable-specific logic (lines 228-298):
  - Preflight configuration validation
  - `promptPreflightConfig()` function (60 lines of dead code)
  - Enable logic in single-site workflow
  - Enable logic in CSV bulk workflow
- **Simplified**: All error messages now say "disable" instead of "enable or disable"
- **Unchanged**: `audit disable` command works exactly as before
- **Unchanged**: API enablement and scheduled audit logic

## 🔗 Related Changes

This PR works in conjunction with:
- **spacecat-audit-worker PR**: [https://github.com/adobe/spacecat-audit-worker/pull/1820]
  - Adds `auditContext` parameter to `isAuditEnabledForSite()`
  - Bypasses enabled check when `slackContext` is present

## 🚀 Deployment Notes

1. **Deploy spacecat-audit-worker FIRST**, then deploy this PR
2. **Rollback plan**: If issues occur, revert both PRs in reverse order

Please ensure your pull request adheres to the following guidelines:
- [X] make sure to link the related issues in this description. Or if there's no issue created, make sure you 
  describe here the problem you're solving.
- [X] when merging / squashing, make sure the fixed issue references are visible in the commits, for easy compilation of release notes

If the PR is changing the API specification:
- [ ] make sure you add a "Not implemented yet" note the endpoint description, if the implementation is not ready 
  yet. Ideally, return a 501 status code with a message explaining the feature is not implemented yet.
- [ ] make sure you add at least one example of the request and response.

If the PR is changing the API implementation or an entity exposed through the API:
[...]

PR #2561, opened 2026-06-05
feat: adding new `ensure preflight` slack command

radhikagpt1208:
Related Issue: https://jira.corp.adobe.com/browse/SITES-45215

## Summary

Adds a new Slack command `@spacecat ensure preflight <baseURL>` that validates site preflight configuration and enables the preflight audit. This ports the preflight enablement logic removed from `audit enable` in #1674 into a dedicated command, without reintroducing bulk enable.

### New command: `ensure preflight`
- Validates the site exists and that `preflight` is registered in configuration
- Checks site readiness (authoring type, preview URL / delivery config, etc.)
- If config is missing → posts a Slack message with **Configure & Enable** button
- If only `contentSourcePath` is missing (CS sibling disambiguation case) → posts a tailored message explaining shared program/environment
- If ready → enables preflight immediately via `Configuration.enableHandlerForSite`

### Shared preflight helpers (`src/support/slack/preflight/preflight-config.js`)
- `getPreflightMissingConfigLabels()` — DA / CS / CS-Crosswalk / AMS missing config detection
- `isContentSourcePathRequired()` — requires path when another site in the **same org** shares the same program, environment, and **authoring type** (`cs` vs `cs/crosswalk` are separate pools; aligned with preflight MFE `findMatchingSite`)
- `isPreflightSiteConfigReady()`, `promptPreflightConfig()`, `enablePreflightAuditForSite()`

### Modal updates (reused existing handlers)
- `open-preflight-config.js` — added **Content Source Path** field
- `preflight-config-modal.js` — validates and persists `deliveryConfig.contentSourcePath`; refactored to use shared helpers

### Deprecation message
- `toggle-site-audit.js` — `audit enable` deprecation text now points to `ensure preflight <site>`

## Out of scope (follow-ups)
- Audit-worker `isAuditDisabledForSite` change (deferred)
- UE as a separate authoring type in the modal (MFE maps UE → `cs/crosswalk`)
- Uniqueness validation of `contentSourcePath` across siblings
- Onboarding wiki update (`audit enable … preflight` → `ensure preflight <site>`)

## Test plan

- [x] Unit tests: `ensure-preflight.test.js`, `preflight-config.test.js`, updated modal/toggle tests
- [x] Manual dev Slack validation:
  - [ ] Invalid URL / site not found errors
  - [ ] Ready CS site → immediate enable
  - [ ] Missing base config → generic prompt → modal → enable
  - [ ] CS sibling (same authoring type) → custom content-source-path message → path required in modal
  - [ ] Cross-type sibling (`cs` + `cs/crosswalk`) → path **not** required
  - [ ] `get site-audits` shows `preflight` enabled after success
  - [ ] `audit enable … preflight` shows deprecation message

Thanks for contributing!

### DIFF
--- a/src/support/slack/commands/ensure-preflight.js
+++ b/src/support/slack/commands/ensure-preflight.js
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { isValidUrl } from '@adobe/spacecat-shared-utils';
+import BaseCommand from './base.js';
+import { extractURLFromSlackInput, postErrorMessage } from '../../../utils/slack/base.js';
+import {
+  enablePreflightAuditForSite,
+  ERROR_MESSAGE_PREFIX,
+  isPreflightSiteConfigReady,
+  PREFLIGHT_AUDIT_TYPE,
+  promptPreflightConfig,
+  SUCCESS_MESSAGE_PREFIX,
+} from '../preflight/preflight-config.js';
+
+const PHRASES = ['ensure preflight'];
+
+export default (context) => {
+  const baseCommand = BaseCommand({
+    id: 'ensure-preflight',
+    name: 'Ensure Preflight for Site',
+    description: 'Validates site preflight configuration and enables the preflight audit for a site.',
+    phrases: PHRASES,
+    usageText: `${PHRASES[0]} {site}`,
+  });
+
+  const { log, dataAccess } = context;
+  const { Configuration, Site } = dataAccess;
+
+  /**
+   * Executes the command to ensure preflight configuration for a site.
+   *
+   * @param {Array<string>} args - The arguments provided to the command.
+   * @param {Object} slackContext - The Slack context object.
+   * @param {Function} slackContext.say - The Slack say function.
+   * @returns {Promise} A promise that resolves when the operation is complete.
+   */
+  const handleExecution = async (args, slackContext) => {
+    const { say } = slackContext;
+    let baseURL;
+
+    try {
+      const [baseURLInput] = args;
+      baseURL = extractURLFromSlackInput(baseURLInput);
+
+      if (!isValidUrl(baseURL)) {
+        await say(`${ERROR_MESSAGE_PREFIX}Please provide a valid site baseURL.`);
+        return;
+      }
+
+      const site = await Site.findByBaseURL(baseURL);
+      if (!site) {
+        await say(`${ERROR_MESSAGE_PREFIX}Cannot update site with baseURL: "${baseURL}", site not found.`);
+        return;
+      }
+
+      const configuration = await Configuration.findLatest();
+      const registeredAudits = configuration.getHandlers();
+      if (!registeredAudits[PREFLIGHT_AUDIT_TYPE]) {
+        await say(`${ERROR_MESSAGE_PREFIX}The "${PREFLIGHT_AUDIT_TYPE}" audit is not present in the configuration.`);
+        return;
+      }
+
+      const { ready, missingLabels } = await isPreflightSiteConfigReady(site, context);
+      if (!ready) {
+        if (missingLabels.length === 1 && missingLabels[0] === 'Content Source Path') {
+          await say({
+            text: `:warning: Preflight audit requires additional configuration for \`${site.getBaseURL()}\``,
+            blocks: [
+              {
+                type: 'section',
+                text: {
+                  type: 'mrkdwn',
+                  text: `:warning: *Preflight audit requires additional configuration for:*\n\`${site.getBaseURL()}\`\n\n*Missing:*\n• Content Source Path\n\nMultiple sites in this organization share the same AEM CS program and environment. Please provide a content source path to distinguish this site.`,
+                },
+              },
+              {
+                type: 'actions',
+                elements: [
+                  {
+                    type: 'button',
+                    text: {
+                      type: 'plain_text',
+                      text: 'Configure & Enable',
+                    },
+                    style: 'primary',
+                    action_id: 'open_preflight_config',
+                    value: JSON.stringify({
+                      siteId: site.getId(),
+                      auditType: PREFLIGHT_AUDIT_TYPE,
+                    }),
+                  },
+                ],
+              },
+            ],
+          });
+          return;
+        }
+
+        await promptPreflightConfig(slackContext, site, PREFLIGHT_AUDIT_TYPE);
+        return;
+      }
+
+      await enablePreflightAuditForSite(site, dataAccess);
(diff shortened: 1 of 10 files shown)


src/support/slack/commands/ensure-preflight.js
```diff
@@ -0,0 +1,130 @@
...
+      const configuration = await Configuration.findLatest();
+      const registeredAudits = configuration.getHandlers();
+      if (!registeredAudits[PREFLIGHT_AUDIT_TYPE]) {
+        await say(`${ERROR_MESSAGE_PREFIX}The "${PREFLIGHT_AUDIT_TYPE}" audit is not present in the configuration.`);
+        return;
+      }
+
+      const { ready, missingLabels } = await isPreflightSiteConfigReady(site, context);
+      if (!ready) {
+        if (missingLabels.length === 1 && missingLabels[0] === 'Content Source Path') {
```
noozoo: Not a typescript expert, but wouldn't ('Content Source Path' === missingLabels?.[0]) work here?  Or without the ?. if you're sure missingLabels is not null or undefined?

src/support/slack/commands/ensure-preflight.js
```diff
@@ -0,0 +1,130 @@
...
+        await promptPreflightConfig(slackContext, site, PREFLIGHT_AUDIT_TYPE);
+        return;
+      }
+
+      await enablePreflightAuditForSite(site, dataAccess);
+      await say(`${SUCCESS_MESSAGE_PREFIX}Preflight audit has been enabled for "${site.getBaseURL()}".`);
+    } catch (error) {
+      log.error(error);
+      await say(
+        `${ERROR_MESSAGE_PREFIX}An error occurred while trying to ensure preflight for site "${baseURL}": ${error.message}`,
```
noozoo: Should this be 'enable' instead of 'ensure'?

radhikagpt1208: The old `audit enable <site> preflight` was removed in https://github.com/adobe/spacecat-api-service/pull/1674/. After that change, audits are no longer gated behind a per-site enablement check at execution time - execution was decoupled from enablement. So "enable" no longer describes a meaningful state toggle the way it used to.

What this command actually does is it checks the site has the required preflight config (authoring type, preview URL / delivery config, content source path when needed), and only then ensure the preflight handler for the site. So it "ensures" the site is set up for preflight rather than flipping an enable flag. It also keeps us consistent with the existing `ensure entitlement site`/`ensure entitlement imsorg` commands.

src/support/slack/commands/ensure-preflight.js
radhikagpt1208: Simplified this. `isPreflightSiteConfigReady` already returns a `needsContentSourcePath` boolean that's set exactly in the "only content source path is missing" case, so I switched the check to:
`if (needsContentSourcePath)`. It reads as intent and avoids re-deriving the same state from the missingLabels array.

src/support/slack/commands/ensure-preflight.js
atopper: I get it, but I share Mike's first reaction.  Ensure is used throughout 'ensure-preflight" but to maybe add a few words for anyone who reads that error message without the whole context or "...ensure Preflight was configured properly for site..."
Or maybe it is ok as is - "ensure" to me first makes me think it'll only validate, but when I read it differently, it can convey that it'll be changed if the validation fails.
Little nit-picky but...

src/support/slack/commands/ensure-preflight.js
```diff
@@ -0,0 +1,130 @@
...
+      );
+      await postErrorMessage(say, error);
```
atopper: Does this line also report the message, doubling it up?

src/support/slack/commands/ensure-preflight.js
```diff
@@ -0,0 +1,130 @@
...
+        await promptPreflightConfig(slackContext, site, PREFLIGHT_AUDIT_TYPE);
+        return;
+      }
+
+      await enablePreflightAuditForSite(site, dataAccess);
+      await say(`${SUCCESS_MESSAGE_PREFIX}Preflight audit has been enabled for "${site.getBaseURL()}".`);
+    } catch (error) {
+      log.error(error);
+      await say(
+        `${ERROR_MESSAGE_PREFIX}An error occurred while trying to ensure preflight for site "${baseURL}": ${error.message}`,
```
atopper: `baseUrl` could be undefined, if extractURLFromSlackInput() fails.

src/support/slack/preflight/preflight-config.js
```diff
@@ -0,0 +1,280 @@
...
+  const configuration = await Configuration.findLatest();
+  configuration.enableHandlerForSite(PREFLIGHT_AUDIT_TYPE, site);
+  await configuration.save();
+}
+
+export {
+  ERROR_MESSAGE_PREFIX,
+  SUCCESS_MESSAGE_PREFIX,
+  PREFLIGHT_AUDIT_TYPE,
+};
```
atopper: If using the 'export' block at the end, I thought, for consistency, everything would be exported in it (functions, variables, etc.).
Or, more simply, just export the 3 variables immediately on lines 15-17.

src/support/slack/commands/toggle-site-audit.js
```diff
@@ -124,7 +124,7 @@ export default (context) => {
         ? auditTypeOrProfileInput.toLowerCase() : null;
 
       if (isEnableAudit) {
-        await say(`${ERROR_MESSAGE_PREFIX}The \`audit enable\` command is deprecated. Use one-off \`run audit\` from Slack without enabling the site. To disable audits, use \`audit disable\`.`);
+        await say(`${ERROR_MESSAGE_PREFIX}The \`audit enable\` command is deprecated. Use one-off \`run audit\` from Slack without enabling the site. To enable Preflight, use \`ensure preflight <site>\`. To disable audits, use \`audit disable\`.`);
```
atopper, line 127: This is a little confusing of a message... maybe?  I don't really understand "**Use one-off \`run audit\` from Slack without enabling the site.**"  Will the typical reader understand it?

atopper (approved):
A few comments to consider (some just thinking out loud really), but looks good.

src/support/slack/commands/ensure-preflight.js
radhikagpt1208: Removed the `postErrorMessage(say, error)` call from the catch block so the error is reported once (contextual :x: message), and removed the now-unused `postErrorMessage` import.

src/support/slack/commands/ensure-preflight.js
radhikagpt1208: added `const [baseURLInput] = args` out of the try, and made the catch message fall back: `"${baseURL ?? baseURLInput ?? 'unknown'}"` so it never renders `undefined`.

src/support/slack/preflight/preflight-config.js
radhikagpt1208: Changed `ERROR_MESSAGE_PREFIX, SUCCESS_MESSAGE_PREFIX, PREFLIGHT_AUDIT_TYPE` to inline export const on lines 15–17 and deleted the trailing export { … } block.

src/support/slack/commands/toggle-site-audit.js
radhikagpt1208: This was carried over from when the audit enable command was removed in https://github.com/adobe/spacecat-api-service/pull/1674 - it's the same messaging that was already communicated to users at that time. 
"Use one-off run audit from Slack without enabling the site" means that: after the decoupling, you can just run the audit directly via run audit without having to enable the site first (enablement is no longer a prerequisite for execution). Happy to expand the phrasing if you think it still reads ambiguously, but wanted to keep it consistent with what users have already seen.

PR #2562, opened 2026-06-05
feat(prompts): persist intent on prompt upsert/update

dzehnder:
## Problem

DRS already classifies a search-intent bucket for every prompt and forwards it to the API (`V2Prompt.intent` in `llmo-data-retrieval-service/.../spacecat_v2_prompts_sync.py`). The Postgres `prompts.intent` column exists (mysticat-data-service migration `20260601000000_prompts_branded_intent.sql`) and the live stats endpoint (`GET /v2/orgs/:spaceCatId/brands/:brandId/prompts/stats`) aggregates over it.

But the API **dropped `intent` at ingestion**:
- `upsertPrompts()` in `src/support/prompts-storage.js` built the inserted row without `intent`.
- `updatePromptById()` did not handle `intent` in its update patch.
- The `V2PromptInput` OpenAPI schema had no `intent` property.

Result: every prompt landed with `intent = NULL`, so the stats intent breakdown was all zeros.

## Change

Persist the `intent` callers already send. **No LLM classification here** — classifying human-added prompts that arrive without an intent is a separate follow-up PR.

### Normalization (on write)
A small, unit-tested `normalizeIntent()` helper:
- lowercases + trims the value;
- applies the legacy remap mirroring DRS `INTENT_REMAP`: `statistical`/`navigational` -> `informational`, `commercial` -> `transactional`;
- validates against the 6 canonical buckets: `informational, instructional, comparative, transactional, planning, delegation`;
- if absent, empty, or invalid after remap -> stores `NULL`.

`NULL` (rather than coercing to `informational`) is deliberate: gap-filling is handled elsewhere, and silently defaulting would pollute the intent breakdown with prompts whose intent is genuinely unknown.

### Wiring
- Threads `intent` through the create path (`createPromptsByBrand`) and update path (`updatePromptByBrandAndId`) — the controllers already pass the full request body through, so the change lands in `upsertPrompts()` (insert row) and `updatePromptById()` (update patch).
- Surfaces `intent` on read: added to the `listPrompts`/`getPromptById` selects and `mapRowToPrompt`.
- Documents `intent` (allowed values + normalization rules) in the `V2PromptInput` and `V2Prompt` OpenAPI schemas.

## Tests
- `test/support/prompts-storage.test.js`: `normalizeIntent` unit tests (canonical passthrough, uppercase lowercased, legacy remap, absent/empty/invalid -> null); `upsertPrompts` captures normalized intent on the inserted row; `updatePromptById` captures normalized intent on the patch and null-on-invalid.
- `test/controllers/brands.test.js`: POST and PATCH carrying `intent` thread the normalized value down to the DB layer.

Commands run:
- `npx mocha test/support/prompts-storage.test.js test/controllers/brands.test.js` -> 370 passing.
- `npm test` -> 11249 passing, 1 failing (pre-existing, unrelated: `fixes.test.js` stubs a non-existent data-access method due to a stale local `@adobe/spacecat-shared-data-access`; fails identically on clean `origin/main`).
- `npm run docs:lint` -> valid (pre-existing warnings only).
- `npx eslint` on changed files -> clean.

### CI root cause (`ci/it-postgres`) — correcting the earlier note

The earlier claim that the *only* CI failure is the pre-existing, unrelated `fixes.test.js` was wrong. The `ci/it-postgres` job is genuinely red, and it is caused by this PR's `intent` work interacting with the integration environment:

- The integration PostgREST image is pinned to `mysticat-data-service:v5.13.0` (`test/it/postgres/docker-compose.yml`).
- `v5.13.0` **predates** the `prompts.intent` migration (`20260601000000_prompts_branded_intent.sql`, first released in data-service `v5.27.0`).
- So in that environment, writing/selecting `prompts.intent` makes PostgREST return 500 — insert: `PGRST204` "Could not find the 'intent' column of 'prompts' in the schema cache"; select: `42703` "column prompts.intent does not exist" — failing `test/it/shared/tests/categories-prompts.js`.

Bumping the pinned image is **not** viable: a newer image introduces a `chk_active_brand_has_site_id` constraint that rejects the IT seed brand and pulls in ~23 unrelated migrations.

### Best-effort missing-column fallback (`src/support/prompts-storage.js`)

The storage layer now self-defends so it works whether or not the `intent` column exists:

- A module-level `WeakMap` (`intentColumnSupported`), keyed by the PostgREST client, caches per-client support. Fresh clients (and unit-test mocks) start unknown; production detects once per client.
- `isMissingIntentColumnError(error)` recognizes the missing-`intent`-column case (lowercased `message`/`details`/`hint`/`code` contains `intent` and either `column` or `schema cache`), covering both the insert `PGRST204` and the select `42703`.
- The four touchpoints (`listPrompts`, `getPromptById`, `upsertPrompts` insert + update, `updatePromptById`) try **with** intent and, on a missing-column error, set the WeakMap to `false`, strip/omit `intent`, and retry that one operation. Once a client is known-unsupported, subsequent calls skip intent up front (no extra error/retry).
- **The happy path (column present) is byte-for-byte unchanged**: intent is still written/selected, no extra DB round-trips, no retry. All existing unit tests pass untouched.

Net effect: in the pinned-image IT environment, prompts are written/read **without** intent (201/200) instead of 500; in environments where the column exists, intent is persisted exactly as before.

### OpenAPI input-enum relaxation

`V2PromptInput.intent` previously pinned an `enum` of only the 6 canonical buckets, but the code accepts and normalizes legacy aliases (`statistical`/`navigational`/`commercial`). Removed the `enum` on the **input** schema (kept `type: string` + a description listing the canonical values and noting legacy aliases are accepted and normalized on write). The `V2Prompt` **output** schema is unchanged (still the canonical enum + nullable).

### Tests / gates (latest run)

- New `intent column best-effort fallback` suite in `test/support/prompts-storage.test.js`: missing-column error on insert -> retries and still inserts (without intent); update retries; `listPrompts`/`getPromptById` retry the select and succeed omitting intent; second call on the same client skips intent up front (WeakMap); a non-intent error still throws with no spurious retry.
- `npm test` -> 11256 passing, 1 failing — the single failure remains the pre-existing, unrelated `fixes.test.js` (stale local `@adobe/spacecat-shared-data-access` stub), identical on `origin/main`.
- `npm run docs:lint` -> valid (pre-existing warnings only). `npx eslint` on changed files -> clean.
- `it-postgres` cannot be run locally (no Docker/AWS); the fallback is validated by the pinned-image `ci/it-postgres` job plus the new unit tests.

### DIFF
--- a/docs/openapi/schemas.yaml
+++ b/docs/openapi/schemas.yaml
@@ -6263,6 +6263,21 @@ V2Prompt:
       description: The source system or process that created the prompt
       example: "config"
       default: "config"
+    intent:
+      type: [string, 'null']
+      description: >-
+        Canonical search-intent bucket persisted for the prompt, or null when
+        unknown/unclassified. One of the canonical buckets (mirrors DRS prompt
+        generation).
+      enum:
+        - informational
+        - instructional
+        - comparative
+        - transactional
+        - planning
+        - delegation
+        - null
+      example: informational
     updatedAt:
       $ref: '#/DateTime'
     updatedBy:
@@ -6342,6 +6357,24 @@ V2PromptInput:
       type: string
       description: The source system or process that created the prompt
       default: "config"
+    intent:
+      type: string
+      description: >-
+        Optional search-intent bucket for the prompt. Normalized on write:
+        the value is lowercased, legacy labels are remapped
+        (`statistical`/`navigational` -> `informational`,
+        `commercial` -> `transactional`), and the result is validated against
+        the canonical buckets below. An absent, empty, or (post-remap) invalid
+        value is stored as null (NULL) — it is NOT coerced to a default
+        bucket. Canonical buckets mirror DRS prompt generation.
+      enum:
+        - informational
+        - instructional
+        - comparative
+        - transactional
+        - planning
+        - delegation
+      example: informational
 
 V2PromptBulkDeleteRequest:
   type: object
--- a/docs/openapi/schemas.yaml
+++ b/docs/openapi/schemas.yaml
@@ -6361,19 +6361,16 @@ V2PromptInput:
       type: string
       description: >-
         Optional search-intent bucket for the prompt. Normalized on write:
-        the value is lowercased, legacy labels are remapped
+        the value is lowercased, legacy aliases are remapped
         (`statistical`/`navigational` -> `informational`,
         `commercial` -> `transactional`), and the result is validated against
-        the canonical buckets below. An absent, empty, or (post-remap) invalid
+        the canonical buckets. An absent, empty, or (post-remap) invalid
         value is stored as null (NULL) — it is NOT coerced to a default
-        bucket. Canonical buckets mirror DRS prompt generation.
-      enum:
-        - informational
-        - instructional
-        - comparative
-        - transactional
-        - planning
-        - delegation
+        bucket. Canonical buckets (mirror DRS prompt generation):
+        `informational`, `instructional`, `comparative`, `transactional`,
+        `planning`, `delegation`. No `enum` constraint is applied to the
+        input: legacy aliases (`statistical`, `navigational`, `commercial`)
+        are accepted and normalized on write.
       example: informational
 
 V2PromptBulkDeleteRequest:
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -15,6 +15,47 @@ import crypto from 'node:crypto';
 import { hasText, isValidUUID } from '@adobe/spacecat-shared-utils';
 
 /**
+ * The 6 canonical intent buckets persisted in `prompts.intent`. Mirrors the
+ * buckets DRS emits (see DRS `prompt_generation_agentic_traffic` generation
+ * prompts) and that the stats intent breakdown aggregates over.
+ */
+const INTENT_VALUES = ['informational', 'instructional', 'comparative', 'transactional', 'planning', 'delegation'];
+const CANONICAL_INTENTS = new Set(INTENT_VALUES);
+
+/**
+ * Legacy intent labels remapped onto the canonical buckets. Mirrors DRS
+ * `INTENT_REMAP` (src/providers/prompt_generation_agentic_traffic/utils/
+ * hard_validate.py) so values produced by older generations or external
+ * callers collapse onto the supported set instead of dropping to NULL.
+ */
+const INTENT_REMAP = {
+  statistical: 'informational',
+  navigational: 'informational',
+  commercial: 'transactional',
+};
+
+/**
+ * Normalizes a caller-supplied intent for persistence into `prompts.intent`.
+ *
+ * Lowercases the value, applies the legacy remap, then validates against the
+ * 6 canonical buckets. Absent, empty, or values that are still invalid after
+ * remapping yield `null` — gap-filling (e.g. LLM classification of
+ * human-added prompts) is handled elsewhere, so we never coerce to a default
+ * bucket here.
+ *
+ * @param {*} intent - Raw intent value from the request body
+ * @returns {string|null} Canonical lowercase intent, or null
+ */
+export function normalizeIntent(intent) {
+  if (!hasText(intent)) {
+    return null;
+  }
+  const lowered = intent.trim().toLowerCase();
+  const remapped = INTENT_REMAP[lowered] || lowered;
+  return CANONICAL_INTENTS.has(remapped) ? remapped : null;
+}
+
+/**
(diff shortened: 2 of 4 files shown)


MysticatBot (changes requested):
Hey @dzehnder,

**Verdict: Request changes** - one tightenable finding around the error-detection heuristic that could silently drop data via false-positive.
_Changes: Persists the `intent` field on prompt upsert/update with normalization, legacy remapping, and a best-effort fallback for environments missing the column (4 files)._

## Must fix before merge
1. [Important] `isMissingIntentColumnError` uses broad string-matching instead of gating on known error codes - `src/support/prompts-storage.js:isMissingIntentColumnError`

   The detector joins `message`, `details`, `hint`, and `code` into a single haystack and checks for both `"intent"` AND (`"column"` OR `"schema cache"`). This will false-positive on any PostgREST error that happens to mention `intent` alongside `column` - for example, a future check-constraint violation like `"column intent violates check constraint"` would be swallowed, the WeakMap would latch to `false`, and all subsequent operations on that client would silently discard the intent the caller sent.

   This is the exact data-loss pattern the PR exists to fix, reintroduced via a different mechanism.

   **Suggested fix** - gate on the two known error codes first, then confirm `intent` is mentioned:

   ```javascript
   export function isMissingIntentColumnError(error) {
     if (!error) return false;
     const code = (error.code || '').toUpperCase();
     if (code === '42703' || code === 'PGRST204') {
       const haystack = [error.message, error.details, error.hint]
         .filter(Boolean).join(' ').toLowerCase();
       return haystack.includes('intent');
     }
     return false;
   }
   ```

   This preserves the `intent` substring check (so a missing `status` column does not trigger the fallback) but only fires on the two specific codes the PR identifies. Robust against PostgREST wording changes and future column additions.

<details>

- suggestion: Add a `TODO(SITES-XXXX)` comment at the `intentColumnSupported` WeakMap declaration linking to a tracked backlog item for removal once the IT PostgREST image is bumped past v5.27.0. Without this, the ~200 lines of fallback code will sit indefinitely as unexplained complexity. - `src/support/prompts-storage.js:intentColumnSupported`
- suggestion: If the retry/fallback pattern persists beyond the IT image bump, consider extracting the try/detect/retry/cache logic into a shared helper to avoid the four near-identical copies drifting independently.
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 8m 59s | <b>Cost:</b> $4.39 | <b>Commit:</b> <code>973a939f24355e615ac042b48ae3ae4fef63a2ec</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

dzehnder:
Thanks for the review — addressed in `3c46d5ae`:

**Must-fix — `isMissingIntentColumnError` broad matching → false-positive data loss:** now gates on the error `code` (`PGRST204` / `42703`) **first**, then checks for the `intent` substring in message/details/hint. A check-constraint violation like `column intent violates check constraint` (code `23514`) no longer matches, so the fallback can't latch off and silently drop intent. Added regression tests for that case and for a correct missing-column code on a *different* column.

**Suggestion 1 — TODO/tracking:** added a TODO at the `intentColumnSupported` WeakMap with the removal condition (IT PostgREST image bumped to a data-service version ≥ v5.27.0 with the `intent` column), referencing SITES-39521.

**Suggestion 2 — de-duplicate the fallback:** extracted the try/detect/cache/retry into a single `withMissingIntentFallback(client, run)` helper plus a module-level `stripIntent()`, replacing the four near-identical copies in `listPrompts`/`getPromptById`/`upsertPrompts`/`updatePromptById`. Behavior is unchanged (happy path still includes intent, no extra round-trips); `prompts-storage.js` patch coverage stays at 100%.

MysticatBot (approved):
Hey @dzehnder,

_Changes: Persists the `intent` field on prompt upsert/update with normalization, legacy remapping, and a best-effort fallback for environments missing the column (4 files)._

- `isMissingIntentColumnError` broad string-matching now gated on error codes (PGRST204/42703) first, preventing false-positive data loss
- TODO added at WeakMap with removal condition and SITES-39521 reference
- Fallback logic extracted into `withMissingIntentFallback` helper, eliminating four near-identical copies

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 26s | <b>Cost:</b> $3.63 | <b>Commit:</b> <code>d427ddf524e1798edadfa4dad24cb24ab6657c41</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2564, opened 2026-06-05
fix: guardrails on agentic/referral traffic read endpoints (SITES-46098)

akshaymagapu:
## Problem
Prod incident 2026-06-05 (SITES-46098). An automated client issued unbounded-date-range `agentic-traffic/by-url` queries (4-month span, `limit=500`, iterating `urlPathSearch`) that each ran ~30s, hit the `statement_timeout`, and saturated the single Aurora reader (92–99% CPU → 5xx/latency alerts). `parseAgenticTrafficParams` only format-validated dates — no max-range cap. Referral params weren't even format-validated.

## Changes 
- **Date range capped at 62 days (~2 months)**; wider spans rejected with **400**. Enforced in the shared auth wrappers (`withAgenticTrafficAuth`, `withReferralTrafficAuth`) so it covers **all** agentic-traffic and referral-traffic read endpoints. Also rejects malformed/non-real dates (e.g. `2026-02-30`) and reversed ranges.
- **Agentic by-url max page size 500 → 200.**
- **Referral by-url max page size 1000 → 200.**

New shared helper `src/controllers/llmo/traffic-date-range.js` (`checkDateRange`).

## Tests
- New `test/controllers/llmo/traffic-date-range.test.js` (boundary, reversed, malformed, non-real, incident range).
- Guardrail describe blocks added to both controller suites; updated the by-url cap assertions (500→200, 1000→200).
- Affected suites green (218 passing); lint clean.

## Out of scope (follow-ups)
- Per-caller/per-site rate limiting (429) — needs shared cross-Lambda state.
- `rpc_agentic_traffic_by_url` optimization + aligning the 30s function timeout with the Fastly/gateway budget.
- Reader auto-scaling.

### DIFF
--- a/docs/openapi/llmo-api.yaml
+++ b/docs/openapi/llmo-api.yaml
@@ -2784,13 +2784,25 @@ llmo-agentic-traffic-by-url:
     summary: Per-URL agentic traffic performance analysis (paginated)
     description: |
       Queries `rpc_agentic_traffic_by_url`. Powers the URL Performance table with
-      infinite-scroll pagination. Returns a page of `pageSize` rows (default 50, max 500)
+      infinite-scroll pagination. Returns a page of `pageSize` rows (default 50, max 200)
       plus `totalCount` for pagination context.
       Supports server-side substring search via `urlPathSearch`.
+
+      Date range guardrail (SITES-46098): `startDate` and `endDate` must be
+      supplied together — providing only one returns `400`. The requested span
+      may not exceed 62 days; a wider span returns `400`. Omitting both applies
+      the default 28-day window. This cap is enforced on every agentic-traffic
+      and referral-traffic read endpoint.
     operationId: getAgenticTrafficByUrl
     parameters:
-      - { name: startDate, in: query, schema: { type: string, format: date } }
-      - { name: endDate, in: query, schema: { type: string, format: date } }
+      - name: startDate
+        in: query
+        description: Range start (`YYYY-MM-DD`). Required if `endDate` is set; span with `endDate` must be <= 62 days.
+        schema: { type: string, format: date }
+      - name: endDate
+        in: query
+        description: Range end (`YYYY-MM-DD`). Required if `startDate` is set; span with `startDate` must be <= 62 days.
+        schema: { type: string, format: date }
       - { name: platform, in: query, schema: { type: string } }
       - { name: categoryName, in: query, schema: { type: string } }
       - { name: agentType, in: query, schema: { type: string } }
@@ -2799,11 +2811,11 @@ llmo-agentic-traffic-by-url:
       - $ref: './parameters.yaml#/agentTypes'
       - name: pageSize
         in: query
-        description: Number of rows per page (default 50, max 500)
+        description: Number of rows per page (default 50, max 200)
         schema:
           type: integer
           minimum: 1
-          maximum: 500
+          maximum: 200
           default: 50
       - name: pageOffset
         in: query
--- a/src/controllers/llmo/llmo-agentic-traffic.js
+++ b/src/controllers/llmo/llmo-agentic-traffic.js
@@ -395,6 +395,7 @@ async function withAgenticTrafficAuth(context, getSiteAndValidateAccess, handler
 
   const rangeError = checkDateRange(context.data);
   if (rangeError) {
+    log.info(`Agentic traffic ${handlerName} rejected (date range guardrail): ${rangeError}`);
     return badRequest(rangeError);
   }
 
--- a/src/controllers/llmo/llmo-agentic-traffic.js
+++ b/src/controllers/llmo/llmo-agentic-traffic.js
@@ -18,6 +18,7 @@ import { S3Client } from '@aws-sdk/client-s3';
 import crypto from 'crypto';
 import { generateIsoWeekRange, getWeekDateRange } from './llmo-brand-presence.js';
 import { parseAgentTypes } from './llmo-agent-types.js';
+import { checkDateRange } from './traffic-date-range.js';
 import { cachedOk } from '../../support/cached-response.js';
 
 // Site-scoped agentic traffic handlers. Queries mysticat-data-service via PostgREST.
@@ -43,7 +44,7 @@ const VALID_SORT_COLUMNS_BY_USER_AGENT = new Set([
   'page_type', 'agent_type', 'unique_agents', 'total_hits',
 ]);
 const DEFAULT_BY_URL_LIMIT = 50;
-const MAX_BY_URL_LIMIT = 500;
+const MAX_BY_URL_LIMIT = 200;
 
 // UI platform code → DB value. 'all' / unknown → null (no filter). Applied
 // in parseAgenticTrafficParams, so it affects every site-scoped endpoint.
@@ -390,6 +391,11 @@ async function withAgenticTrafficAuth(context, getSiteAndValidateAccess, handler
   if (!Site?.postgrestService) {
     log.error('Agentic traffic APIs require PostgREST (DATA_SERVICE_PROVIDER=postgres)');
     return badRequest('Agentic traffic data is not available. PostgreSQL data service is required.');
+  }
+
+  const rangeError = checkDateRange(context.data);
+  if (rangeError) {
+    return badRequest(rangeError);
   }
 
   const { siteId } = context.params;
--- a/src/controllers/llmo/llmo-referral-traffic.js
+++ b/src/controllers/llmo/llmo-referral-traffic.js
@@ -15,6 +15,7 @@ import {
 } from '@adobe/spacecat-shared-http-utils';
 import { cachedOk } from '../../support/cached-response.js';
 import { generateIsoWeekRange, getWeekDateRange } from './llmo-brand-presence.js';
+import { checkDateRange } from './traffic-date-range.js';
 
 /**
  * Site-scoped referral traffic handler factories.
@@ -40,7 +41,7 @@ const SOURCE_TO_TABLE = {
 };
 
 const DEFAULT_BY_URL_PAGE_SIZE = 50;
-const MAX_BY_URL_PAGE_SIZE = 1000;
+const MAX_BY_URL_PAGE_SIZE = 200;
 
 // Mirrors the CASE whitelist in rpc_referral_traffic_by_url for defence-in-depth.
 const VALID_BY_URL_SORT_COLUMNS = new Set([
@@ -132,6 +133,11 @@ async function withReferralTrafficAuth(
   if (!Site?.postgrestService) {
     log.error('Referral traffic APIs require PostgREST (DATA_SERVICE_PROVIDER=postgres)');
     return badRequest('Referral traffic data is not available. PostgreSQL data service is required.');
+  }
+
+  const rangeError = checkDateRange(context.data);
+  if (rangeError) {
+    return badRequest(rangeError);
   }
 
   const { siteId } = context.params;
--- a/src/controllers/llmo/llmo-referral-traffic.js
+++ b/src/controllers/llmo/llmo-referral-traffic.js
@@ -41,7 +41,11 @@ const SOURCE_TO_TABLE = {
 };
 
 const DEFAULT_BY_URL_PAGE_SIZE = 50;
-const MAX_BY_URL_PAGE_SIZE = 200;
+// 500 (not the agentic 200) so the elmo-ui referral "All URLs" export — which
+// still paginates by-url client-side at pageSize=500 — is not silently
+// truncated. Drop to 200 once that export migrates to the async urls/export
+// endpoint like the agentic one (SITES-46098 review).
+const MAX_BY_URL_PAGE_SIZE = 500;
 
 // Mirrors the CASE whitelist in rpc_referral_traffic_by_url for defence-in-depth.
 const VALID_BY_URL_SORT_COLUMNS = new Set([
@@ -137,6 +141,7 @@ async function withReferralTrafficAuth(
 
   const rangeError = checkDateRange(context.data);
   if (rangeError) {
+    log.info(`Referral traffic ${handlerName} rejected (date range guardrail): ${rangeError}`);
     return badRequest(rangeError);
   }
 
--- a/src/controllers/llmo/traffic-date-range.js
+++ b/src/controllers/llmo/traffic-date-range.js
@@ -45,8 +45,11 @@ function isRealDate(value) {
  */
 export function checkDateRange(data) {
   const q = data || {};
-  const startRaw = q.startDate || q.start_date;
-  const endRaw = q.endDate || q.end_date;
+  // Nullish coalescing: prefer the camelCase alias unless it is absent
+  // (null/undefined), so an empty-string value isn't silently swapped for the
+  // snake_case alias.
+  const startRaw = q.startDate ?? q.start_date;
+  const endRaw = q.endDate ?? q.end_date;
 
   // Both omitted → handler applies its default window; nothing to validate.
   if (startRaw == null && endRaw == null) {
--- a/src/controllers/llmo/traffic-date-range.js
+++ b/src/controllers/llmo/traffic-date-range.js
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+// Guardrail for the agentic-traffic / referral-traffic analytical endpoints
+// (SITES-46098). An unbounded date range lets a single query scan months of
+// partitions and hit the 30s statement timeout, saturating the Aurora reader.
+// Cap the requested span so over-wide queries are rejected (400) before they
+// reach Postgres.
+export const MAX_DATE_RANGE_DAYS = 62; // ~2 months
+
+const DATE_PATTERN = /^\d{4}-\d{2}-\d{2}$/;
+const MS_PER_DAY = 86_400_000;
+
+// Rejects malformed and non-real dates (e.g. 2026-02-30, 2026-13-01) by
+// round-tripping through Date and comparing back to the input.
+function isRealDate(value) {
+  if (typeof value !== 'string' || !DATE_PATTERN.test(value)) {
+    return false;
+  }
+  const d = new Date(`${value}T00:00:00Z`);
+  if (Number.isNaN(d.getTime())) {
+    return false;
+  }
+  return d.toISOString().slice(0, 10) === value;
+}
+
+/**
+ * Validate the requested date range on a traffic query.
+ *
+ * Reads camelCase and snake_case aliases from the request data. When both
+ * bounds are omitted the caller relies on the handler's default window, so
+ * there is nothing to validate.
+ *
+ * @param {object} [data] - context.data (parsed query params)
+ * @returns {string|null} an error message for a 400 response, or null if valid
+ */
+export function checkDateRange(data) {
+  const q = data || {};
+  const startRaw = q.startDate || q.start_date;
+  const endRaw = q.endDate || q.end_date;
+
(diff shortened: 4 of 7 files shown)


ekremney (approved):
referral is safe though.. not as high cardinality as agentic

rainer-friederich:
Reviewed against the incident analysis — this is the right primary fix (the date-range cap kills the exact 4-month pattern, and enforcing it in the shared `withAgenticTrafficAuth` / `withReferralTrafficAuth` wrappers covers every agentic + referral read endpoint). Three follow-up notes:

**1. elmo-ui regression from the by-url page-size cap (500 → 200)**
The URL Inspector "URL Details" dialog hardcodes `pageSize=500` (`useUrlInspectorAgenticByUrl`) to pull a path's full URL slice in one shot. Capping to 200 means a path with >200 URL rows will now under-show in that dialog. The paginated dashboard table is unaffected (true server-side paging), but that drilldown assumed 500. We've already posted this in the elmo-ui channel so the UI team can confirm/adjust the dialog.

**2. We should add per-caller rate limiting to api-service**
The measured load was ~1 req/s but **~30 by-url queries running in parallel**, each a ~30s scan — i.e. the driver was concurrency, not request rate. This PR lowers per-query cost but does not throttle the fan-out, and `checkDateRange` returns `null` when both bounds are omitted — so a client that simply **drops the explicit dates** (falling back to the default 28-day window) or chunks into ≤62-day windows, fired 30-wide, is still unthrottled. The date cap raises the bar; **per-caller / per-site rate limiting (429) is the durable fix** for "automated client hammering" and should be a fast-follow, not a someday-follow-up.

**3. 62-day cap may still be expensive — worth verifying**
May 2026 alone is ~124 GB / 147M rows, so a 2-month by-url scan can still approach the 30s `statement_timeout` (and the ~15s Fastly/gateway budget), 503-ing the client while still burning reader CPU. Recommend confirming a 62-day by-url query completes comfortably under timeout; if not, the cap may need tightening or the RPC optimizing (the deferred `rpc_agentic_traffic_by_url` item).

Net: 👍 to merge as the immediate mitigation, with per-caller rate limiting tracked as a near-term fast-follow.

MysticatBot (changes requested):
Hey @akshaymagapu,

## Strengths

- **Correct guardrail placement** (`llmo-agentic-traffic.js:393`, `llmo-referral-traffic.js:135`): `checkDateRange` runs in the shared auth wrappers before site validation or any DB call, so over-wide queries are rejected at near-zero cost. This gates every endpoint in both controller families without per-handler changes.
- **Robust date validation** (`traffic-date-range.js:17-28`): The `isRealDate` function round-trips through `Date` and compares back to the input, correctly rejecting nonsense like `2026-02-30`. UTC-pinned parsing (`T00:00:00Z`) avoids timezone-dependent ambiguity.
- **Incident scenario encoded as regression test** (`traffic-date-range.test.js:41`): Pinning the exact 113-day range that caused the production incident as a named test is excellent practice.
- **Clean extraction into shared module** (`traffic-date-range.js`): Single-purpose, well-commented with incident reference, exports only what callers need, and serves both controller families identically.
- **Actionable error messages** (`traffic-date-range.js:70`): "Date range too large: N days requested, max is 62 days" tells clients exactly what went wrong and the limit, which is critical for debugging automated integrations.
- **Page-size clamping confirmed by tests** (`llmo-agentic-traffic.test.js:1018`, `llmo-referral-traffic.test.js:624`): Tests verify that `limit: 99999` is clamped to 200, confirming the constants are enforced as caps (not just defaults).

## Issues

### Important (Should Fix)

**1. Partial date params now rejected - behavioral breaking change** - `traffic-date-range.js:26-28`

When a client sends only `startDate` (no `endDate`), `checkDateRange` returns "Both startDate and endDate are required when either is provided" (400). Previously, the handler applied a default for the missing bound via `sanitizeDateString(q.endDate || q.end_date, defaults.endDate)`. This is a tighter contract and intentional, but it is an undocumented breaking change for existing API consumers. If the automated client or UI currently sends only one date param, requests will start failing.

Confirm that no existing consumer relies on partial date params. If any do, consider only enforcing the range cap when both are present and skipping the "both required" check (letting the handler apply its default for the missing one).

**2. Integration tests missing per CLAUDE.md** - `test/it/` not modified

The repo's CLAUDE.md states: "New or modified endpoints must include integration tests in `test/it/`". This PR modifies endpoint behavior (adds validation, reduces page-size caps). Unit tests are thorough and cover the validation logic well, but the integration-level verification for the full request/response pipeline is absent. For a hotfix this is acceptable to ship, but integration tests should follow.

**3. OpenAPI specification not updated** - `docs/openapi/` not modified

The repo's CLAUDE.md states: "Specification Sync: Keep OpenAPI specs and implementation in sync." The PR introduces new validation constraints (62-day max range, required date pairing, reduced page sizes) that change the API contract for consumers. The OpenAPI spec should document these constraints so clients can discover the limits without trial-and-error. Acceptable as a follow-up given the hotfix context.

### Minor (Nice to Have)

**4. Same-day range boundary not tested** - `traffic-date-range.test.js`

A single-day query (`startDate === endDate`, span = 1 day) is a natural boundary that is not exercised. If the inclusive `+1` in `spanDays` were accidentally removed, this case would silently change semantics. One extra assertion would pin it.

## Recommendations

- **Log rejected requests at INFO level** before returning the 400. A structured log entry with the rejected params and handler name would help confirm the guardrail is working in production and identify clients that need updating.
- **Add a test asserting the default window <= MAX_DATE_RANGE_DAYS.** When both dates are omitted, `checkDateRange` passes through and the handler applies a 28-day default. If someone later changes the default to 90 days, the guardrail will not catch it. A one-line assertion in the controller test binds the invariant.
- **Add a happy-path wiring test** that passes a valid narrow range and verifies the RPC/data-access layer is reached. The current controller-level tests only exercise the rejection path.

## Assessment

Ready to merge? With fixes

The core guardrail logic is correct, well-placed, and thoroughly tested for rejection paths. The only substantive concern is confirming the partial-date behavioral change (finding 1) does not break existing consumers. Findings 2 and 3 (integration tests and OpenAPI sync) are acceptable as immediate follow-ups given this is a production incident hotfix. The CI failures appear unrelated to this PR's changes (the `build` job is pending, `it-postgres` likely depends on it).

## Next Steps

1. Confirm no existing consumer sends only one date param (finding 1) - check UI code and automated client behavior.
2. Add integration tests in `test/it/` as a follow-up PR.
3. Update OpenAPI specs to document the new constraints.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 31s | <b>Cost:</b> $3.39 | <b>Commit:</b> <code>97b6b2781994d6ff93dada5a92f0ad88b4f04d94</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

solaris007 (approved):
Hey @akshaymagapu - thanks for turning this around so fast. I reviewed it in the context of the SITES-46098 incident I just worked (I'm the on-call who filed it). Approving - this is the right origin-side fix - with a few non-blocking notes on what still needs to land at the Fastly edge.

**What's great**
- `checkDateRange` enforced in the shared `with*TrafficAuth` wrappers, *before* `getSiteAndValidateAccess` and the PostgREST/RPC call, so an over-wide query now 400s without ever touching the Aurora reader. That kills the incident's primary vector (the abuser sent a 113-day span).
- Page-size caps 500 -> 200 and 1000 -> 200, plus malformed/non-real/reversed-range handling. Well tested, including the exact incident range.

**Non-blocking, but please track these as required follow-ups (not optional) to actually prevent recurrence:**

1. **Per-caller rate limiting should live at Fastly, not in api-service.** The PR defers it as "needs shared cross-Lambda state" - that complexity is exactly the signal it belongs at the **edge**. Fastly Edge Rate Limiting (rate counters keyed on client IP / API token / the `*/agentic-traffic|referral-traffic/*` path) rejects floods with 429 *before* they reach Lambda or the reader - no distributed state to build, and it also protects Lambda concurrency. This is the one item I'd move off the api-service backlog onto the edge.

2. **The caps alone don't fully close the incident.** The abuser iterated `urlPathSearch` one call at a time. Post-PR each call is <=62 days / <=200 rows, but a high-frequency in-bounds scan still hammers the single Aurora reader, and a 62-day query on a large site (the incident site has ~147M rows/month) can still approach the 30s `statement_timeout`. So the edge rate-limit (1) plus the `rpc_agentic_traffic_by_url` optimization (already in your out-of-scope list) are what actually prevent a repeat - worth marking them required rather than nice-to-have.

3. **FYI / coordination:** as the emergency stopgap I hot-fixed a `recv` VCL snippet that 403s the abusive IP on both fronting Fastly services (`DlxppS2VoAizEqJ9bPasq6` v93 and the LLMO service `z1m7OynhCynDF77Ur7uGMJ` v8). That is drift from the checked-in `fastly/elmo` VCL and should be removed once this lands and the edge controls exist - ideally replaced by an ACL-backed blocklist so future blocks are a runtime ACL-entry add (no VCL version activation).

Net: api-service = per-request semantic correctness (this PR), Fastly = volumetric/abuse protection (rate-limit + blocklist) in front of the origin. They're complementary. Approving on the understanding that the edge rate-limit and the rpc optimization follow.

rainer-friederich:
https://github.com/adobe/project-elmo-ui/pull/1943 Elmo-UI changes

akshaymagapu:
Thanks for the thorough review — addressed the findings:

**#1 Partial date params (behavioral change)** — Verified the only consumer of these endpoints, the LLMO UI (`project-elmo-ui`), always sends `startDate` and `endDate` together (derived from a single `timeFrame`; the time-range picker is capped at 8 contiguous weeks = 56 days, so it never exceeds the 62-day cap either). Given that, we kept the **strict** contract rather than guessing a default for a missing bound: both omitted → handler default window; exactly one provided → `400`; both provided → validated. (commit `26c124f9`)

**#4 Same-day boundary** — Added a `startDate === endDate` (span = 1 day) test, plus a happy-path wiring test that a valid in-range request reaches the data layer. (commit `efb8d325`)

**Recommendation: log rejected requests** — Done. Both `withAgenticTrafficAuth` / `withReferralTrafficAuth` now emit an INFO log (`... rejected (date range guardrail): ...`) before the 400, so the guardrail firing and the handler are visible in prod. (commit `efb8d325`)

**#3 OpenAPI sync** — Documented on `llmo-agentic-traffic-by-url`: pageSize max `500 → 200`, the 62-day span cap, the both-dates-required rule, and the `400` behavior. (commit `9737d830`) Note: the **referral-traffic** endpoints are not present in the OpenAPI spec at all today, so their docs are tracked with that separate referral-spec gap.

**#2 Integration tests** — These endpoint families have no existing `test/it/` coverage and the path depends on PostgREST RPCs (`rpc_agentic_traffic_by_url`, etc.) that aren't in the IT harness, so this is tracked as a follow-up (as you noted, acceptable for the hotfix). Unit coverage of the guardrail is thorough.

**Two related decisions worth flagging:**
- **Referral by-url cap is 500, not 200** (agentic is 200). The elmo-ui referral "All URLs" export still paginates `by-url` client-side at `pageSize=500` and terminates on `batch.length < 500`, so a 200 cap would silently truncate it. Agentic's export already migrated to the async `urls/export` endpoint, so agentic is safe at 200. Referral drops to 200 once its export migrates too.
- **Per-caller rate limiting → Fastly edge** (per @solaris007's review): drafted a concrete plan (VCL `ratecounter`/`penaltybox` keyed on `x-api-key`/IP on the `*/agentic-traffic|referral-traffic/*` paths, plus an ACL-backed blocklist to replace the emergency inline-VCL IP block). Tracking as the durable fast-follow.

MysticatBot (commented):
Hey @akshaymagapu,

Re-reviewing after the 2 new commits (`efb8d32`, `26c124f`).

## Strengths

- **Previously flagged issues now addressed**: The strict both-dates-required decision is committed with clear rationale (`traffic-date-range.js:51-54`), and the same-day boundary test is now present (`traffic-date-range.test.js:60`).
- **Referral page-size cap adjusted with documented rationale** (`llmo-referral-traffic.js:43-56`): The comment explains why 500 (not 200), names the specific UI dependency (elmo-ui "All URLs" export), and includes the ticket reference for the follow-up. This is how you communicate a deliberate asymmetry to future readers.
- **Correct guardrail placement** (`llmo-agentic-traffic.js:393`, `llmo-referral-traffic.js:139`): `checkDateRange` runs in the shared auth wrappers before `getSiteAndValidateAccess` and before any RPC call, so over-wide queries are rejected at near-zero cost. Every endpoint in both controller families inherits the guard without per-handler changes.
- **Robust date validation** (`traffic-date-range.js:17-28`): The `isRealDate` function round-trips through `Date` and compares back to the input, correctly rejecting nonsense like `2026-02-30`. UTC-pinned parsing (`T00:00:00Z`) avoids timezone-dependent ambiguity. No ReDoS risk in the anchored `DATE_PATTERN`.
- **Incident scenario encoded as regression test** (`traffic-date-range.test.js:41`): Pinning the exact 113-day range that caused the production incident as a named test is excellent practice.
- **Order-of-operations verification** (`llmo-agentic-traffic.test.js:224`, `llmo-referral-traffic.test.js:157`): Tests assert `stubbedValidateAccess` was NOT called on rejected ranges, proving the guardrail fires before any DB or site-access work.
- **Actionable error messages** (`traffic-date-range.js:70`): "Date range too large: N days requested, max is 62 days" tells clients exactly what went wrong and the limit. Messages reveal no internal state or site data.

## Issues

### Minor (Nice to Have)

**1. `||` treats empty string as absent in param extraction** - `traffic-date-range.js:39-40`

```javascript
const startRaw = q.startDate || q.start_date;
const endRaw = q.endDate || q.end_date;
```

If a caller sends `startDate: ''` (empty string from query param parsing like `?startDate=&endDate=2026-01-28`), `||` falls through to `start_date`. If both are empty strings, the function returns `null` (both-omitted path). The current code is accidentally safe because `isRealDate('')` rejects empty strings via the regex check. Using `??` (nullish coalescing) would be more precise if the intent is "prefer camelCase unless absent":

```javascript
const startRaw = q.startDate ?? q.start_date;
const endRaw = q.endDate ?? q.end_date;
```

Not exploitable today, but clearer intent.

**2. Referral guardrail test suite thinner than agentic** - `test/controllers/llmo/llmo-referral-traffic.test.js:147-166`

The agentic test suite has 4 cases (over-range, order-of-ops, valid-passthrough, single-bound). The referral suite has only 2 (over-range, order-of-ops). Since both call the same `checkDateRange`, the shared logic is tested independently, so this is not a coverage gap. Adding a valid-passthrough case for referral would confirm the wiring end-to-end for the happy path.

## Recommendations

- **Track integration tests and OpenAPI sync as required follow-ups.** The repo CLAUDE.md states both are required ("New or modified endpoints must include integration tests in `test/it/`" and "Specification Sync: Keep OpenAPI specs and implementation in sync"). The prior review flagged these, and the team has explicitly accepted the trade-off for a hotfix. Ensure they land before the next release.
- **Consider making the 62-day cap configurable per-site.** The constant applies uniformly regardless of data volume. A site with 147M rows/month may still approach the 30s timeout at 62 days, while small sites could handle longer ranges. Accepting an override from `site.getConfig()` (e.g., `config.llmo.maxDateRangeDays`) falling back to the exported constant would avoid a code deploy if the cap needs tightening for specific sites.
- **Add an invariant assertion binding the default window to MAX_DATE_RANGE_DAYS.** When both dates are omitted, `checkDateRange` passes through and the handler applies a default window. If someone later changes the default to exceed 62 days, the guardrail will not catch it. A one-line assertion would pin this.

## Assessment

The core guardrail logic is correct, well-placed, and thoroughly tested. The two previously flagged follow-ups (integration tests, OpenAPI sync) are acknowledged by the team as fast-follows for a production incident hotfix - not blockers given two human approvals and active team discussion. The minor items above are improvement opportunities, not merge gates. Ship it.

## Next Steps

1. Track integration tests (`test/it/`) and OpenAPI spec updates as required follow-ups.
2. Consider the `??` vs `||` tweak and per-site configurability in follow-up work.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 7s | <b>Cost:</b> $3.12 | <b>Commit:</b> <code>26c124f900a4549b3ed32464de1a0e7c9766aa46</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

akshaymagapu:
Thanks @MysticatBot — addressed the two minor items and tracked the rest.

**Fixed in `4ba8a0ca`:**
- **1 `||` → `??`** in the date alias fallback (`traffic-date-range.js`) — empty-string camelCase value no longer silently swaps to the snake_case alias. Clearer intent; no behavior change (`isRealDate('')` already rejected it).
- **2 Referral guardrail test parity** — added the valid-passthrough (happy-path wiring) case to the referral suite to match agentic.

**Tracked as required follow-ups → SITES-46103** (under the Mysticat epic, linked to SITES-46098):
- Integration tests (`test/it/`) for both endpoint families
- OpenAPI sync — including the **referral-traffic spec gap** (not in OpenAPI at all today) and propagating the 62-day note across the other agentic endpoints
- **Per-site configurable cap** — `site.getConfig().llmo.maxDateRangeDays` overriding `MAX_DATE_RANGE_DAYS`
- `rpc_agentic_traffic_by_url` optimization + timeout alignment

The durable per-caller throttle (your #1/#2 + @solaris007's notes) is staged as the Fastly edge rate-limit + ACL blocklist in **adobe/spacecat-infrastructure#587**.

On the invariant-assertion recommendation: skipped a literal-vs-constant test (it'd be theater since `defaultDateRange()` isn't exported); the per-site-config follow-up will make the default/cap relationship explicit instead.

MysticatBot (approved):
Hey @akshaymagapu,

## Strengths

- **Previously flagged minor items addressed** (commit `4ba8a0ca`): The `||` to `??` fix in `traffic-date-range.js:43-44` correctly uses nullish coalescing for the alias fallback, and the referral guardrail test suite now has a valid-passthrough case matching the agentic suite's coverage shape.
- **Correct guardrail placement** (`llmo-agentic-traffic.js:393`, `llmo-referral-traffic.js:139`): `checkDateRange` runs in the shared auth wrappers before `getSiteAndValidateAccess` and any RPC call, so over-wide queries are rejected at near-zero cost. Tests explicitly verify this order-of-operations guarantee.
- **Clean single-purpose module** (`traffic-date-range.js`): Well-scoped, exports only what callers need, handles both param naming conventions. The `isRealDate` round-trip via UTC-pinned parsing is robust against non-real dates. No ReDoS risk in the anchored `DATE_PATTERN`.
- **Documented asymmetry** (`llmo-referral-traffic.js:43-56`): The comment explains why referral stays at 500 while agentic drops to 200, names the specific UI dependency (elmo-ui export), and includes the ticket for the follow-up.
- **No security concerns**: Input is strictly validated (anchored regex + round-trip) before it reaches error messages or database queries. No injection vectors, no information leakage, no bypass paths.
- **Strong assertion quality**: Each test verifies a specific behavior boundary. Controller-level tests confirm the data layer is never reached on rejection.
- **OpenAPI spec updated** (`llmo-api.yaml`): The 62-day cap, both-dates-required constraint, and reduced page size are documented for consumers.

## Issues

### Minor (Nice to Have)

**1. OpenAPI spec documents constraints only on `agentic-traffic-by-url`** - `docs/openapi/llmo-api.yaml`

The 62-day cap is enforced in the shared auth wrappers for every agentic-traffic and referral-traffic endpoint, but the OpenAPI update only annotates the `by-url` operation. Consumers hitting other operations (KPIs, by-user-agent) will discover the constraint via a 400 rather than the spec. Acceptable for the hotfix - worth including in the OpenAPI-sync follow-up (SITES-46103).

**2. Mixed alias precedence not tested** - `test/controllers/llmo/traffic-date-range.test.js`

The `??` fallback means camelCase takes precedence over snake_case when both are present. There is a test for pure snake_case, but no test for the case where both aliases are provided with conflicting values. In practice query-string parsers will not produce both, but a single assertion would document the precedence contract for future readers.

## Assessment

Ready to merge? Yes

Both prior Minor findings are addressed. The core guardrail logic is correct, well-placed, thoroughly tested, and secure. All substantive follow-ups (rate limiting, integration tests, OpenAPI propagation, RPC optimization) are tracked in SITES-46103. Two human approvals confirm the operational context. Ship it.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 37s | <b>Cost:</b> $2.42 | <b>Commit:</b> <code>4ba8a0ca3fc14626ef9cca982f8fabe47293ab60</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2565, opened 2026-06-05
feat(prompts): classify intent on write for human-added prompts

dzehnder:
## Stacked PR

Depends on and **must merge after #2562** (`feat/persist-prompt-intent-on-upsert`). Base branch is set to that branch; the diff here is only this change.

## The gap

PR #2562 makes the API **persist** a caller-supplied `prompts.intent` via `normalizeIntent()` (lowercase + legacy remap + validate against the 6 buckets, else NULL). That covers **pipeline** prompts — DRS forwards their intent.

**Human-added** prompts (added via the UI/API) have no generator, so they arrive **without** an intent and would stay `NULL` forever, leaving the `prompts/stats` intent breakdown incomplete.

## What this does

When a prompt is created/updated with **no intent** (absent/empty after normalization) **and** non-empty text, classify its text into one of the 6 canonical buckets and persist it. Prompts that **already carry an intent** (pipeline prompts) are **never** re-classified.

- `src/support/intent-classifier.js` — Node mirror of the DRS Python `UserIntentClassifier` (`llmo-data-retrieval-service/.../user_intent_classifier.py`): identical 6-bucket system prompt, `temperature: 0`, strict JSON `{intent, confidence, reasoning}` output, prompt text truncated to 2000 chars. Validation reuses `normalizeIntent` (single source of truth — no reimplementation).
- `src/support/intent.js` — extracts `normalizeIntent` + `INTENT_VALUES` out of `prompts-storage.js` so the classifier can reuse the helper **without an import cycle**; `prompts-storage.js` re-exports both for backward compatibility.
- `upsertPrompts` / `updatePromptById` take an optional `classifyIntent` callback; `BrandsController` builds it once per request and injects it into both call sites.

## LLM client reused

`AzureChatOpenAI` from `@langchain/openai`, the **same client already used by `OrgDetectorAgent`** (`src/agents/org-detector/agent.js`), reusing the existing `AZURE_OPEN_AI_API_KEY` / `_INSTANCE_NAME` / `_DEPLOYMENT_NAME` / `_API_VERSION` env vars. **No new secrets.** (The shared `@adobe/spacecat-shared-gpt-client` `AzureOpenAIClient` was considered but its `fetchChatCompletion` doesn't expose temperature and won't set `json_object` alongside a custom system prompt, so it can't reproduce the DRS contract exactly.)

## Sync vs async decision

**Synchronous, best-effort, batched** within the request:
- Only the subset of prompts **missing** an intent is classified; inputs are **deduplicated** so repeated text costs one LLM call.
- Bounded **concurrency** (cap 10) so a bulk create (arrays up to 3000) can't fan out unbounded in-flight requests.

Rationale: the storage layer already does sequential DB round-trips per prompt and there is **no fire-and-forget / enqueue pattern in this write path**; `gpt-4.1-nano`-class classification is cheap. Tradeoff: bulk creates with many un-intented prompts add some latency — bounded by the toggle (disable instantly) and the existing backfill/reconciliation path. An async/enqueue variant can be a follow-up if bulk latency becomes a concern.

## Non-fatal / best-effort guarantee

Classification **never** blocks or breaks a prompt create/update. If the LLM call fails, times out, returns malformed JSON or an invalid bucket, **or no client is configured**, the prompt persists `intent = NULL` and the request proceeds normally. The backfill path covers it later.

## Toggle

`ENABLE_PROMPT_INTENT_CLASSIFICATION` — **default-safe (off** unless explicitly truthy: `true` / `'true'` / `'1'`**)**. Lets classification be disabled without a redeploy. When off (or Azure creds absent), the controller injects no classifier and intent stays NULL.

## Test coverage

- `test/support/intent-classifier.test.js` (new): toggle default-safe + enable; null when disabled / Azure unconfigured; model constructed with temperature 0 + `json_object`; classifies into a bucket; legacy remap/lowercasing; code-fence stripping; invalid bucket → null; model reject → null (never throws); malformed JSON → null; empty/whitespace text skips the model; batch dedup, reject-to-null, concurrency cap.
- `test/support/prompts-storage.test.js` (extended): upsert classifies a prompt with no intent and persists it; a prompt **with** an intent is **not** re-classified; classification rejection persists NULL and does not fail the request; no classifier provided → NULL. Update path: classifies new text when no intent supplied; does **not** classify when intent supplied; rejection leaves intent unset without failing.

## Validation

- `npm run lint` — clean on all changed files.
- `npm test` — **11272 passing, 1 failing**; the single failure is the **known pre-existing** unrelated `test/controllers/fixes.test.js` issue (`Cannot stub non-existent property getAllFixesWithSuggestionsByOpportunityId`, stale node_modules). All new tests pass.

## Update

Follow-up after the base (#2562) received a critical fix:

- **Merged the updated base in** (`origin/feat/persist-prompt-intent-on-upsert`, not rebased) so this branch carries #2562's best-effort *missing `intent` column* fallback (`intentColumnSupported` WeakMap + `isMissingIntentColumnError` + strip/retry on insert/update and intent-omitting re-select on list/get). The fallback and the classifier integration **coexist** in `upsertPrompts`/`updatePromptById`: the classifier populates `intent` for null/human-added rows **before** insert, then the missing-column fallback strips/omits `intent` only when the column is unsupported. `normalizeIntent`/`INTENT_VALUES` live solely in `intent.js`; `isMissingIntentColumnError` solely in `prompts-storage.js` (no duplicates).
- **Bounded the synchronous LLM call** with a timeout (default 10s, overridable via `PROMPT_INTENT_CLASSIFICATION_TIMEOUT_MS`) so a hung Azure call can't stall prompt creation. On timeout the call is treated as a classification failure (intent stays null) and the write proceeds — same non-fatal contract.
- **Array `content` handling:** `AzureChatOpenAI.invoke()` may return `.content` as an array of content parts rather than a string; a new `contentToString` helper concatenates the text parts before JSON-parsing so array-shaped output no longer silently fails to null.
- **Enum-lock test:** asserts `INTENT_VALUES` (code) matches the `V2Prompt.intent` enum in `docs/openapi/schemas.yaml` (parsed from the yaml) so the two can't drift.
- Added focused unit tests for the new/uncovered defensive branches (array content, empty concatenation, invoke timeout, env-override fallback, model-construction failure) to close the `codecov/patch` gap. `npm test` post-merge: **11291 passing, 1 failing** (only the known pre-existing `fixes.test.js`).

### DIFF
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -15,6 +15,47 @@ import crypto from 'node:crypto';
 import { hasText, isValidUUID } from '@adobe/spacecat-shared-utils';
 
 /**
+ * The 6 canonical intent buckets persisted in `prompts.intent`. Mirrors the
+ * buckets DRS emits (see DRS `prompt_generation_agentic_traffic` generation
+ * prompts) and that the stats intent breakdown aggregates over.
+ */
+const INTENT_VALUES = ['informational', 'instructional', 'comparative', 'transactional', 'planning', 'delegation'];
+const CANONICAL_INTENTS = new Set(INTENT_VALUES);
+
+/**
+ * Legacy intent labels remapped onto the canonical buckets. Mirrors DRS
+ * `INTENT_REMAP` (src/providers/prompt_generation_agentic_traffic/utils/
+ * hard_validate.py) so values produced by older generations or external
+ * callers collapse onto the supported set instead of dropping to NULL.
+ */
+const INTENT_REMAP = {
+  statistical: 'informational',
+  navigational: 'informational',
+  commercial: 'transactional',
+};
+
+/**
+ * Normalizes a caller-supplied intent for persistence into `prompts.intent`.
+ *
+ * Lowercases the value, applies the legacy remap, then validates against the
+ * 6 canonical buckets. Absent, empty, or values that are still invalid after
+ * remapping yield `null` — gap-filling (e.g. LLM classification of
+ * human-added prompts) is handled elsewhere, so we never coerce to a default
+ * bucket here.
+ *
+ * @param {*} intent - Raw intent value from the request body
+ * @returns {string|null} Canonical lowercase intent, or null
+ */
+export function normalizeIntent(intent) {
+  if (!hasText(intent)) {
+    return null;
+  }
+  const lowered = intent.trim().toLowerCase();
+  const remapped = INTENT_REMAP[lowered] || lowered;
+  return CANONICAL_INTENTS.has(remapped) ? remapped : null;
+}
+
+/**
  * Resolves brandId (path param) to Postgres brands.id (uuid).
  * Tries: 1) valid uuid lookup, 2) case-insensitive name lookup.
  *
@@ -238,6 +279,7 @@ function mapRowToPrompt(row) {
     status: row.status || 'active',
     origin: row.origin || 'human',
     source: row.source || 'config',
+    intent: row.intent ?? null,
     createdAt: row.created_at,
     createdBy: row.created_by,
     updatedAt: row.updated_at,
@@ -335,6 +377,7 @@ export async function listPrompts({
     status,
     origin,
     source,
+    intent,
     category_id,
     topic_id,
     brand_id,
@@ -457,6 +500,7 @@ export async function getPromptById({
       status,
       origin,
       source,
+      intent,
       category_id,
       topic_id,
       brand_id,
@@ -574,6 +618,7 @@ export async function upsertPrompts({
       status: p.status || 'active',
       origin: p.origin || 'human',
       source: p.source || 'config',
+      intent: normalizeIntent(p.intent),
       updated_by: updatedBy,
     };
 
@@ -620,6 +665,7 @@ export async function upsertPrompts({
     status: r.status,
     origin: r.origin,
     source: r.source,
+    intent: r.intent,
     createdAt: r.created_at,
     createdBy: r.created_by,
     updatedBy: r.updated_by,
@@ -670,6 +716,9 @@ export async function updatePromptById({
   }
   if (updates.origin !== undefined) {
     patch.origin = updates.origin;
+  }
+  if (updates.intent !== undefined) {
+    patch.intent = normalizeIntent(updates.intent);
   }
   if (updates.categoryId !== undefined) {
     patch.category_id = hasText(updates.categoryId)
@@ -815,8 +864,6 @@ export async function checkPromptsExist({ brandUuid, prompts, postgrestClient })
 
   return data ?? [];
 }
-
-const INTENT_VALUES = ['informational', 'instructional', 'comparative', 'transactional', 'planning', 'delegation'];
 
 export async function getPromptStats({ organizationId, brandUuid, postgrestClient }) {
   if (!postgrestClient?.rpc) {
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -23,6 +23,41 @@ import { INTENT_VALUES, normalizeIntent } from './intent.js';
 export { INTENT_VALUES, normalizeIntent };
 
 /**
+ * Per-client cache of whether `prompts.intent` is selectable/writable. Keyed by
+ * the PostgREST client so unit tests (fresh mock clients) never bleed state and
+ * production detects once per client instance. Absent = unknown (try with
+ * intent); `false` = known-missing (skip intent up front).
+ *
+ * Why best-effort instead of bumping the IT image: the integration PostgREST
+ * image is pinned to a data-service version that predates the `intent`
+ * migration, and bumping it pulls in unrelated migrations + a constraint that
+ * rejects the IT seed brand. So the code self-defends: when `intent` is absent,
+ * prompts are still written/read WITHOUT intent rather than 500-ing.
+ */
+const intentColumnSupported = new WeakMap();
+
+/**
+ * Detects a PostgREST/Postgres error that indicates the `intent` column is
+ * absent. Covers the insert/upsert error (`PGRST204`, "Could not find the
+ * 'intent' column of 'prompts' in the schema cache") and the select error
+ * (`42703`, "column prompts.intent does not exist").
+ *
+ * @param {*} error - Error object from a PostgREST response (`{ message, details, hint, code }`)
+ * @returns {boolean} true when the error is specifically about a missing `intent` column
+ */
+function isMissingIntentColumnError(error) {
+  if (!error) {
+    return false;
+  }
+  const haystack = [error.message, error.details, error.hint, error.code]
+    .filter((v) => v != null)
+    .join(' ')
+    .toLowerCase();
+  return haystack.includes('intent')
+    && (haystack.includes('column') || haystack.includes('schema cache'));
+}
+
+/**
  * Resolves brandId (path param) to Postgres brands.id (uuid).
  * Tries: 1) valid uuid lookup, 2) case-insensitive name lookup.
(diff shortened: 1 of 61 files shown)


src/support/prompts-storage.js
```diff
@@ -728,6 +696,31 @@ export async function upsertPrompts({
...
+      const apply = (r) => {
+        if (r.intent === null && hasText(r.text) && intentByText.has(r.text)) {
+          // eslint-disable-next-line no-param-reassign
+          r.intent = intentByText.get(r.text);
+        }
+      };
+      toInsert.forEach(apply);
+      toUpdate.forEach(apply);
+      // Keep the returned payload consistent with what was persisted.
+      processed.forEach(apply);
```
MysticatBot, line 725: issue (blocking): **No aggregate timeout on the bulk classification batch.**

Each individual LLM call is bounded at 10s and concurrency is capped at 10, but for a bulk create with many unique prompts needing classification, the aggregate wall clock is unbounded: `ceil(uniqueCount / 10) * 10s`. For 300 unique prompts that is 300s worst-case, well beyond a Lambda timeout.

The per-call timeout guards against Azure going dark, but not against Azure responding slowly (e.g., 8-9s per call, just under the timeout). A slow Azure endpoint would stall the entire write request until Lambda kills it - and since DB writes happen *after* classification, the user gets a 5xx for a request that would have succeeded without classification.

**How to fix:** Wrap the entire `classifyIntents` call in a total-batch timeout (e.g., 30s default, overridable via `PROMPT_INTENT_CLASSIFICATION_BATCH_TIMEOUT_MS`). When it fires, use whatever results have completed so far and proceed to the DB writes. This preserves the best-effort contract while bounding worst-case latency to a predictable ceiling:

```js
const batchTimeoutMs = Number(env.PROMPT_INTENT_CLASSIFICATION_BATCH_TIMEOUT_MS) || 30000;
const intentByText = await withTimeout(
  classifyIntents(classifyIntent, rowsNeedingIntent.map((r) => r.text)),
  batchTimeoutMs,
).catch(() => new Map());
```

The existing `withTimeout` helper can be reused directly.

MysticatBot (changes requested):
Hey @dzehnder,

**Verdict: Request changes** - one production-safety concern on the bulk classification path.
_Changes: Adds LLM-based intent classification for human-added prompts on the write path, behind a feature toggle, with bounded concurrency and non-fatal error handling (6 files)._

## Must fix before merge

1. [Important] No aggregate timeout on bulk classification batch - `src/support/prompts-storage.js:696` (details inline)

<details>

- suggestion: Explicit null-check on the `intentByText.get()` result in the `apply` function for clarity (currently a no-op null-to-null assignment on failed classifications) - `src/support/prompts-storage.js:706`
- suggestion: Consider a lower per-call timeout for the single-update interactive path, or document that operators should tune `PROMPT_INTENT_CLASSIFICATION_TIMEOUT_MS` aggressively for UI-driven updates (10s is generous for a user-facing single-prompt edit) - `src/support/prompts-storage.js:823`
- suggestion: Add a test for `updatePromptById` when only non-text fields change (e.g., `updates: { name: 'X' }`) with `classifyIntent` provided, verifying the classifier is not called - `test/support/prompts-storage.test.js`
- nit: Greedy regex in `parseIntent` would mismatch on multi-object output (mitigated by try/catch and json_object response format, so low risk) - `src/support/intent-classifier.js:242`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 19s | <b>Cost:</b> $5.50 | <b>Commit:</b> <code>6ce2c71c91aed14ae2f04d93f3290e1aa6d7dfa4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

dzehnder:
Thanks for the review — addressed in `fa7f9014` (and merged latest `main`):

**Must-fix — no aggregate timeout on the bulk classification batch:** `classifyIntents` now takes a `timeoutMs` (default **30s**, overridable via `PROMPT_INTENT_CLASSIFICATION_BATCH_TIMEOUT_MS`, resolved in `brands.js` and threaded through `upsertPrompts`). It races the worker pool against the cap; on expiry it returns the classifications **completed so far** (workers populate the result map as they go) and the write proceeds — so a slow-but-not-dead Azure can no longer stall a bulk create past the Lambda timeout. The single-update path is unchanged (one per-call-bounded call). Bounded worst case is now `timeoutMs`, not `ceil(n/10)·10s`.

**Non-blocking #1 (apply null-check clarity):** `apply` now only overwrites when the classified value is non-null (no more null→null no-op).

**Non-blocking #3 (test for non-text update):** added a test asserting `updatePromptById` does **not** invoke the classifier when only non-text fields change (e.g. `{ name }`).

Also added tests for the batch timeout (returns partial/none without hanging, keeps completed results on timeout, disabled-timeout waits for all).

On the other two non-blocking notes: #2 (lower single-update timeout) — left the 10s default; it's env-tunable per the existing `PROMPT_INTENT_CLASSIFICATION_TIMEOUT_MS` and the single call is already bounded. #4 (greedy regex in `parseIntent`) — left as-is given the try/catch + `json_object` response format make it low-risk, happy to tighten if you'd prefer.

Coverage stays 100% (lines) on `prompts-storage.js` and `intent-classifier.js`.

MysticatBot (approved):
Hey @dzehnder,

**Verdict: Approve** - prior blocking finding addressed cleanly, no new concerns.
_Changes: Adds LLM-based intent classification for human-added prompts on the write path, behind a feature toggle, with bounded concurrency and batch timeout (6 files)._

<details>

- suggestion: Workers in `classifyIntents` continue running after the batch timeout resolves (the `Promise.race` does not abort in-flight calls). No correctness bug since Lambda freeze cancels them implicitly, but worth a one-line code comment documenting the known tradeoff - `src/support/intent-classifier.js:420`
- suggestion: Log when the batch timeout fires (e.g. `classified ${results.size}/${unique.length} prompts before batch timeout`) so slow-Azure scenarios are visible in production without tracing - could be added in the `upsertPrompts` caller after comparing `intentByText.size` vs `rowsNeedingIntent.length`
- suggestion: Add a focused unit test for `resolveBatchTimeoutMs` with invalid env values (empty string, negative, "nope") to match the existing coverage pattern for `resolveInvokeTimeoutMs` - `test/support/intent-classifier.test.js`
</details>

- Aggregate batch timeout added (30s default, env-tunable, returns partial results on expiry)
- Explicit null-check on classified value before overwriting intent
- Test added for `updatePromptById` when only non-text fields change

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 41s | <b>Cost:</b> $4.35 | <b>Commit:</b> <code>58957d36299a2a043629d8df30442f7f7e69f157</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

dzehnder:
Update (`688279c0`): **removed the `ENABLE_PROMPT_INTENT_CLASSIFICATION` opt-in flag** per product decision — once merged, intent is always classified for human-added prompts.

- Dropped the flag and `isIntentClassificationEnabled()`.
- `createIntentClassifier` now builds a classifier **whenever Azure OpenAI is configured**; missing Azure creds remains a hard dependency (returns null → intent stays null), not a toggle. The best-effort/non-fatal contract and the per-call + batch timeouts are unchanged.
- Tests updated; `intent-classifier.js` stays at 100% line coverage.

Note: this makes classification active in every environment where the Azure OpenAI creds are present (dev/stage/prod already have them, via the org-detector agent).

PR #2571, opened 2026-06-08
feat(s2s): allow S2S consumers with configuration:write to toggle site audits

ravverma:
## Summary

- `PATCH /configurations/sites/audits` was already gated by `CAP_CONFIGURATION_WRITE` at Layer 1 (`s2sAuthWrapper`) but `sites-audits-toggle.js` used a bare `hasAdminAccess()` check that blocked all S2S callers regardless of their capabilities
- Add `authorizeConfigWrite` helper (same dual-layer pattern as `configuration.js`): admin bypass first, then fresh DB fetch via `hasS2SCapability(CAP_CONFIGURATION_WRITE)` with structured `[acl]`/`[s2s]` log lines
- Update `setUpdatedBy` to emit `s2s:<clientId>` instead of the opaque `system` string for S2S callers, giving a traceable audit trail for mutations

## Test plan

- [ ] 5 new unit tests: grant, `updatedBy` audit trail, missing-capability denial, revoked consumer denial, `requestId=unknown` fallback
- [ ] All 31 tests in `sites-audits-toggle.test.js` pass (`npm test`)
- [ ] Capability-constants drift test still passes

### DIFF
--- a/src/controllers/sites-audits-toggle.js
+++ b/src/controllers/sites-audits-toggle.js
@@ -39,7 +39,7 @@ export default (ctx) => {
   const { Configuration, Site, Organization } = dataAccess;
   const accessControlUtil = AccessControlUtil.fromContext(ctx);
 
-  const authorizeConfigWrite = async (context, route) => {
+  const authorizeConfigWrite = async (context, route, message) => {
     const requestId = context?.invocation?.id || 'unknown';
     const isAdmin = accessControlUtil.hasAdminAccess();
     const s2sResult = isAdmin
@@ -47,7 +47,7 @@ export default (ctx) => {
       : await accessControlUtil.hasS2SCapability(CAP_CONFIGURATION_WRITE);
     if (!isAdmin && !s2sResult.allowed) {
       log.info(`[acl] Denied ${route} - reason=${s2sResult.reason} clientId=${s2sResult.clientId || 'n/a'} consumerId=${s2sResult.consumerId || 'n/a'} requestId=${requestId}`);
-      return forbidden('Only admins can change configuration settings.');
+      return forbidden(message);
     }
     if (s2sResult.allowed) {
       log.info(`[s2s] ${route} granted clientId=${s2sResult.clientId || 'n/a'} consumerId=${s2sResult.consumerId || 'n/a'} capability=${CAP_CONFIGURATION_WRITE} requestId=${requestId}`);
@@ -89,7 +89,7 @@ export default (ctx) => {
    * @returns {Promise<Response|*>}
    */
   const execute = async (context) => {
-    const denied = await authorizeConfigWrite(context, 'PATCH /configurations/sites/audits');
+    const denied = await authorizeConfigWrite(context, 'PATCH /configurations/sites/audits', 'Only admins can change configuration settings.');
     if (denied) {
       return denied;
     }
@@ -183,7 +183,7 @@ export default (ctx) => {
 
       if (hasUpdates === true) {
         const { authInfo: { profile } } = context.attributes;
-        const s2sClientId = ctx.s2sConsumer?.getClientId?.();
+        const s2sClientId = context.s2sConsumer?.getClientId?.();
         configuration.setUpdatedBy(
           profile.email || (s2sClientId ? `s2s:${s2sClientId}` : 'system'),
         );
@@ -192,7 +192,7 @@ export default (ctx) => {
 
       return createResponse(responses, 207);
     } catch (error) {
-      context.log.error(error.message);
+      log.error(error.message);
       return internalServerError('An error occurred while trying to enable or disable audits.');
     }
   };
--- a/src/controllers/sites-audits-toggle.js
+++ b/src/controllers/sites-audits-toggle.js
@@ -20,6 +20,7 @@ import {
   hasText, isValidUrl, isNonEmptyObject,
 } from '@adobe/spacecat-shared-utils';
 import AccessControlUtil from '../support/access-control-util.js';
+import { CAP_CONFIGURATION_WRITE } from '../routes/capability-constants.js';
 
 /**
  * @param {object} ctx - Context of the request.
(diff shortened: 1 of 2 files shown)


src/controllers/sites-audits-toggle.js
```diff
@@ -165,13 +183,16 @@ export default (ctx) => {
 
       if (hasUpdates === true) {
         const { authInfo: { profile } } = context.attributes;
```
MysticatBot, line 185: issue (blocking): `ctx.s2sConsumer` should be `context.s2sConsumer`.

The canonical `setUpdatedBy` helper in `configuration.js` (line 54) reads from `context.s2sConsumer` (the per-request parameter), not from `ctx` (the factory construction-time context). While in production these may be the same object (controllers are instantiated per request), reading from `ctx` diverges from the established pattern and creates fragility if the two ever differ.

The test passes because `contextMock.s2sConsumer` is set on the construction context, but this masks the inconsistency.

**Fix**: Change to `context.s2sConsumer?.getClientId?.()` to match the `configuration.js` pattern, where `context` is the execute-time parameter.

src/controllers/sites-audits-toggle.js
```diff
@@ -165,13 +183,16 @@ export default (ctx) => {
...
+        configuration.setUpdatedBy(
+          profile.email || (s2sClientId ? `s2s:${s2sClientId}` : 'system'),
+        );
         await configuration.save();
       }
 
       return createResponse(responses, 207);
     } catch (error) {
-      log.error(error.message);
+      context.log.error(error.message);
```
MysticatBot: issue (blocking): `context.log.error(error.message)` will throw `TypeError` if `context.log` is undefined.

In the old code, `log` was destructured from `context` earlier in `execute`, so its availability was explicit. Now that destructure was removed (line 97 only pulls `data`), but the catch block still accesses `context.log` without a guard.

If a caller passes a context without `log` and an exception fires inside the try block, the catch block itself throws, masking the real error and preventing the graceful 500 response.

**Fix**: Use the outer-scope `log` (from `ctx`) which is already available in the closure:
```js
log.error(error.message);
```
This is safe because `log` is destructured from `ctx` at the factory level and the middleware stack guarantees its presence.

MysticatBot (changes requested):
Hey @ravverma,

**Verdict: Request changes** - two issues with how `ctx` vs `context` is used in the new code paths.
_Changes: adds S2S capability-based authorization to the site audits toggle endpoint, replacing a bare admin check with the documented dual-layer pattern (2 files)._

## Must fix before merge

1. [Important] `setUpdatedBy` reads `s2sConsumer` from `ctx` instead of `context` - `src/controllers/sites-audits-toggle.js:185` (details inline)
2. [Important] `context.log.error(...)` in catch block will throw TypeError if `context` lacks `log` - `src/controllers/sites-audits-toggle.js:195` (details inline)

<details>

- nit: `authorizeConfigWrite` hardcodes the forbidden message instead of accepting it as a parameter like `configuration.js` does (reduces reusability if a second call site is added) - `src/controllers/sites-audits-toggle.js:48`
- nit: `isNonEmptyObject(context)` guard at line 93 is now unreachable since `authorizeConfigWrite` already accesses `context` properties before this check runs - `src/controllers/sites-audits-toggle.js:93`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 19s | <b>Cost:</b> $4.05 | <b>Commit:</b> <code>e7bd789483414bf2f2ba3a3292c41f179fecc7ad</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (commented):
Hey @ravverma,

**Verdict: Comment** - prior findings addressed cleanly, no blocking issues remain.
_Changes: adds S2S capability-based authorization to the site audits toggle endpoint with structured audit logging and traceable updatedBy attribution (2 files)._

- `setUpdatedBy` now correctly reads from `context.s2sConsumer` instead of `ctx`
- Catch block uses outer-scope `log` instead of potentially-undefined `context.log`
- `authorizeConfigWrite` accepts a `message` parameter instead of hardcoding the forbidden string

<details>

- nit: `isNonEmptyObject(context)` guard at line 96 is unreachable - `authorizeConfigWrite` already accesses `context` properties before this check runs (persists from prior review) - `src/controllers/sites-audits-toggle.js:96`
- nit: S2S deny tests omit `s2sConsumer` from the execute arg while the grant test includes it - pass it consistently to mirror the real runtime caller shape - `test/controllers/sites-audits-toggle.test.js:973`
- suggestion: consider extracting the duplicated `authorizeConfigWrite` into a shared module - the implementation is now identical to the one in `configuration.js`, creating drift risk when one copy is updated - `src/controllers/sites-audits-toggle.js:41`
- suggestion: add a test asserting `profile.email` takes precedence over `s2s:<clientId>` when both are present, to document the priority chain for future maintainers - `src/controllers/sites-audits-toggle.js:186`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 45s | <b>Cost:</b> $3.94 | <b>Commit:</b> <code>3f3014b6c34550629e099084bf7bb0361cd2865c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @ravverma,

**Verdict: Approve** - prior suggestions addressed cleanly, no issues remain.
_Changes: adds `s2sConsumer` to test execute() calls and a profile.email precedence test, addressing prior review suggestions (2 files)._

<details>

- nit: `isNonEmptyObject(context)` guard at line 96 is unreachable - `authorizeConfigWrite` already accesses `context` properties before this check runs (persists from prior review) - `src/controllers/sites-audits-toggle.js:96`
- suggestion: duplicated `authorizeConfigWrite` is now identical to the one in `configuration.js` - worth extracting into a shared utility in a follow-up PR to avoid drift - `src/controllers/sites-audits-toggle.js:41`
</details>

- S2S deny/fallback tests now pass `s2sConsumer` to execute args, matching the real runtime caller shape
- Email precedence test documents the `profile.email > s2s:<clientId> > system` priority chain

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 44s | <b>Cost:</b> $3.11 | <b>Commit:</b> <code>314ce3896f32ad772985e9bd9b4537b68a2fb8a2</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2572, opened 2026-06-08
feat(preflight): defer eligibility to Mysticat on site-scoped endpoints (SITES-46202)

GeezerPelletier:
## Summary

- Removes SpaceCat-side preflight eligibility check (`isAuditEnabledForSite` + `Configuration.handlers.preflight` + `TierClient.checkValidEntitlement`) from the new `POST /sites/:siteId/preflights` endpoint. Eligibility now lives **only** in Mysticat, which already owns the three-gate model (Gate 0 `tier.features.preflight`, Gate 1 `enabled_opportunity_types`, Gates 2/3 per-fact + per-site overrides) — see SITES-44736.
- Closes the source-of-truth drift that let a PLG Free site pass SpaceCat (`SiteEnrollment` exists for the productCode) but fail Mysticat's stricter Gate 0 silently as a `200` with `audits: []`.
- The legacy `/preflight/jobs` endpoints are unchanged — explicit out-of-scope per the [SITES-46202](https://jira.corp.adobe.com/browse/SITES-46202) brief.
- Aligns ADR-002 (response-codes table, "no phantom jobs" prose, consequences passage). ADR-003 verified — no eligibility prose, no change needed.

## What changed

**`src/controllers/preflight.js`** — removed:
- `isAuditEnabledForSite` function
- `Configuration.findLatest()` lookup that fed it
- `PREFLIGHT_NOT_ENABLED` error code path on the new endpoint
- `TierClient` import (no longer used)

**Kept unchanged**: `accessControlUtil.hasAccess(site)` tenancy boundary, URL hostname validation, page-auth resolution (`resolvePromiseToken` / `checkEnableAuthentication` / `retrievePageAuthentication`), `AsyncJob` + `Preflight` entity creation, **all** legacy `/preflight/jobs` endpoints.

**`test/controllers/preflight.test.js`** — removed 5 dead tests, added one positive test (`does not consult Configuration / TierClient (eligibility deferred to Mysticat per SITES-46202)`) that pins the new contract by stubbing `Configuration.findLatest` to throw — guards against a future regression.

**`docs/decisions/002-preflight-api-rest-redesign.md`** — error-codes table, error-body example, no-phantom-jobs section, and consequences passage updated. New paragraph explicitly documents "Eligibility is Mysticat's decision, not SpaceCat's."

## Test plan

- [x] `npx mocha test/controllers/preflight.test.js` → 76 passing
- [x] `npm run docs:lint` → clean (warnings unrelated to this PR)
- [x] Full suite: ~11,562 passing; small number of pre-existing flaky hook-timeouts in unrelated files (`ephemeral-run`, `Bot Blocker`, `Sites Controller` brand-profile, `approveSiteCandidate`, `edge-routing-auth`) — none touch preflight code
- [ ] Reviewer: please verify the ADR-002 prose changes read clearly to a fresh-eyes reader
- [ ] Reviewer: confirm no other consumer of `PREFLIGHT_NOT_ENABLED` (MFE, downstream tooling) needs updating in lockstep

## Notes

- The new `/sites/:siteId/preflights` endpoints aren't documented in OpenAPI yet (a separate gap from PR #2478 that didn't get backfilled). No OpenAPI changes needed here as a result; that's worth its own follow-up.
- An empty `audits: []` response from Mysticat is now the contract for "tier disables preflight" / "all opted out" — callers that polled a `CANCELLED` job for disabled-preflight detection must poll `Preflight.status` and inspect `result.audits` instead.

Related: [SITES-44686](https://jira.corp.adobe.com/browse/SITES-44686) (parent — REST redesign), [SITES-44736](https://jira.corp.adobe.com/browse/SITES-44736) (Mysticat-side gating).

### DIFF
--- a/docs/decisions/002-preflight-api-rest-redesign.md
+++ b/docs/decisions/002-preflight-api-rest-redesign.md
@@ -101,7 +101,6 @@ Body:
 |--------|-------------|-----------|
 | `400 Bad Request` | `PREFLIGHT_INVALID_REQUEST` | `url` is missing, not a valid URI, or does not belong to the site identified by `:siteId` |
 | `403 Forbidden` | `PREFLIGHT_ACCESS_DENIED` | Caller does not have access to the site |
-| `403 Forbidden` | `PREFLIGHT_NOT_ENABLED` | Preflight is not enabled for the site |
 | `404 Not Found` | `PREFLIGHT_SITE_NOT_FOUND` | `siteId` does not exist |
 | `502 Bad Gateway` | `PREFLIGHT_UPSTREAM_ERROR` | Mysticat returned a 5xx response |
 | `500 Internal Server Error` | `PREFLIGHT_INTERNAL_ERROR` | Unexpected error within this service |
@@ -109,16 +108,16 @@ Body:
 Error response body:
 ```json
 {
-  "errorCode": "PREFLIGHT_NOT_ENABLED",
-  "message": "Preflight is not enabled for this site"
+  "errorCode": "PREFLIGHT_ACCESS_DENIED",
+  "message": "Access denied"
 }
 ```
 
 `errorCode` gives consumers a stable machine-readable contract; `message` is a human-readable hint and should not be parsed by clients.
 
-No job record is created for `400`, `403`, or `404` responses. The current `/preflight/beta/jobs`
-behavior of creating a job and immediately setting it to `CANCELLED` when preflight is disabled
-is not carried forward — a `403` is returned immediately, keeping the job store clean.
+**Eligibility is Mysticat's decision, not SpaceCat's.** This endpoint does not consult `Configuration.handlers.preflight`, `Entitlement`, `SiteEnrollment`, or any product-code gating to decide whether the call should proceed. The only gate SpaceCat enforces is the tenancy boundary (`accessControlUtil.hasAccess(site)` — IMS org membership) plus the basic URL-belongs-to-site sanity check. Past those, the request proceeds to Mysticat unconditionally, and Mysticat's three-gate model (Gate 0 tier features, Gate 1 `enabled_opportunity_types`, Gates 2/3 per-fact + per-site overrides) is the sole source of truth for what runs. A site whose tier disables preflight will return `200` with `audits: []` — not a `403` from SpaceCat. See SITES-46202.
+
+No job record is created for `400`, `403`, or `404` responses **emitted by SpaceCat's own checks** (access, URL validation). Once those pass, the `AsyncJob` + `Preflight` rows are created unconditionally before dispatching to Mysticat; any subsequent Mysticat-side `5xx` flips the rows to `FAILED` with a `502 PREFLIGHT_UPSTREAM_ERROR` response.
 
 ---
 
@@ -230,10 +229,11 @@ Key changes:
   always a human-readable address for technical accounts). `displayName` is composed from
   `profile.first_name + ' ' + profile.last_name` (falling back to `profile.name`). No
   additional IMS lookup required — both fields are on the authenticated profile.
-- **No phantom jobs for rejected requests.** If the site is not found, the caller lacks access,
-  or preflight is not enabled for the site, the endpoint returns an error immediately without
-  creating a job record. The previous behavior of creating a `CANCELLED` job in these cases
-  is removed.
+- **No phantom jobs for SpaceCat-side rejections.** If the site is not found or the caller lacks
+  access, the endpoint returns an error immediately without creating a job record. The previous
+  behavior of creating a `CANCELLED` job in these cases is removed. Note that preflight-eligibility
+  is **not** a SpaceCat-side concern (see Eligibility below) — sites whose tier disables preflight
+  pass through to Mysticat and receive a `200` with `audits: []`, not a SpaceCat `403`.
 - **No `organizationId` in the path.** `siteId` is a globally unique UUID, consistent with
   all other site-scoped resources in this service.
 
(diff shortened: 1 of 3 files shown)


clotton (commented):
## Review — feat(preflight): defer eligibility to Mysticat (SITES-46202)

### What it does

Removes SpaceCat-side preflight eligibility checking (`isAuditEnabledForSite`, `Configuration.findLatest`, `TierClient`) from the new `POST /sites/:siteId/preflights` endpoint. Eligibility now lives exclusively in Mysticat. SpaceCat only enforces the tenancy boundary (`hasAccess`) and URL-belongs-to-site validation before passing through unconditionally.

---

### Correctness

**Good:**
- The removal is surgical — tenancy boundary, URL validation, auth resolution, `AsyncJob`/`Preflight` creation, and all legacy `/preflight/jobs` endpoints are untouched
- Closing source-of-truth drift (PLG Free site passing SpaceCat but silently failing Mysticat) is the right call
- Regression-guard test is clever: stubs `Configuration.findLatest` to throw, so any future re-introduction of the check causes an immediate test failure

**Worth flagging:**

1. **Missing test for the `audits: []` path.** Previously, an ineligible site got a 403 with no job record. Now it gets a 202, a job record is created, and Mysticat returns `200 audits: []`. There's no test that exercises this specific path and asserts the correct final state of the `AsyncJob`/`Preflight` rows after Mysticat returns empty. The behavior change is significant enough to warrant it.

2. **`PREFLIGHT_NOT_ENABLED` removal is a breaking API contract change.** Any MFE or polling client that branches on this error code gets a `202` + job record instead of a `403`. Is the MFE already updated / does it handle this gracefully?

3. **Job records now created for tier-disabled sites.** The previous contract was: ineligible → no job record. The new contract is: always create a record, inspect `result.audits` for emptiness. This changes write patterns and any monitoring that counts job creation rates could spike for tenants that are tier-disabled.

---

### Code quality

Clean. The removed `isAuditEnabledForSite` was already a near-duplicate of SAW's version, so removing it reduces drift.

One minor note: the inline comment block replacing the removed eligibility check is quite long (6 lines). The rationale is non-obvious enough to justify *a* comment, but most of that content already lives in the ADR and Jira. A one-liner pointing to SITES-46202 and the ADR would suffice — the current comment will rot as gate semantics change.

---

### ADR-002 updates

Read clearly. The "Eligibility is Mysticat's decision, not SpaceCat's" paragraph is well-placed and makes the architectural boundary explicit. The updated consequences passage correctly redirects callers from polling `CANCELLED` jobs to inspecting `result.audits`.

---

### Summary

The design is sound and the implementation is clean. Two things to confirm before merging:

1. MFE / any downstream consumer of `PREFLIGHT_NOT_ENABLED` is confirmed updated or tolerant of the contract change
2. Test covering the `200 audits: []` path from Mysticat (ineligible-site happy path under the new model)

GeezerPelletier:
Thanks for the careful read @clotton. Walking through each point.

## Worth flagging

### 1. Missing test for the `audits: []` path — addressed in `72f07a91`

Good catch. Added `returns 202 and does NOT roll back rows when Mysticat returns 200 with empty audits (tier-ineligible site)` which pins the new contract: empty audits is a successful dispatch, no FAILED rollback, no error. Stubs the Mysticat second call to return `{ok: true, status: 200, json: () => ({ audits: [], ... })}` and asserts (a) 202 response with the Preflight DTO, (b) `setStatus('FAILED')` was NOT called on either Preflight or AsyncJob, (c) `setError` was NOT called. 54 createPreflight tests pass.

### 2. `PREFLIGHT_NOT_ENABLED` removal as breaking contract — non-issue

The new `/sites/:siteId/preflights` endpoints are **unpublished** — no MFE or downstream consumer has shipped against them yet. PR #2478 (the original site-scoped redesign) landed only a few days ago and consumers are still on the legacy `/preflight/jobs` path. The MFE will adopt the new endpoints on its next sweep, at which point it'll see the post-46202 contract from day one. Verified with @GeezerPelletier when scoping the ticket — the explicit framing was "we are coding into an unpublished api workflow, we can revisit operational details as we test and augment."

The legacy `/preflight/jobs` is **unchanged** and never returned `PREFLIGHT_NOT_ENABLED` in the first place (no `isAuditEnabledForSite` call on that path), so existing callers see zero behavior change.

### 3. Job records now created for tier-disabled sites — intentional, monitoring posture changes

Acknowledged. This is the intended design under the SITES-46202 model: SpaceCat is a thin BFF, eligibility is Mysticat's decision, and an ineligible site gets a `COMPLETED` row with empty `result.audits` rather than a 4xx with no row. The monitoring posture shifts accordingly:

- **Old signal** for "this tenant attempted preflight but isn't entitled" → SpaceCat 403 rate or `CANCELLED` job rate
- **New signal** → `Preflight.status='COMPLETED' AND result.audits=[]` count per site / per tier

The ADR-002 update spells this out in the consequences section. Worth a separate observability ticket if anyone has dashboards keyed on the old signal; not aware of any today but flagging for the team.

## Code quality

> The inline comment block replacing the removed eligibility check is quite long (6 lines).

Trimmed in `72f07a91` to a one-liner: `// Eligibility is Mysticat's decision — see SITES-46202 + ADR-002.` Agree the full rationale belongs in the ticket / ADR where it can be updated as gate semantics evolve.

## Net

Both substantive items resolved; the contract-change one was a question that I'm confident is non-blocking given the unpublished-endpoint state. Ready for re-review when you have a moment.

clotton (approved):
All three items addressed — the  test pins the new contract cleanly, the one-liner comment is right-sized, and the contract-change / job-records points are well-reasoned. LGTM.

clotton (approved):
All three items addressed — the empty-audits test pins the new contract cleanly, the one-liner comment is right-sized, and the contract-change / job-records points are well-reasoned. LGTM.

PR #2573, opened 2026-06-08
fix(SITES-46200): reject projectId updates on PATCH /sites/:siteId

alinarublea:
## Problem

When a site is re-parented to a different Spacecat organization, its `projectId` can keep pointing at a project record that still lives in the **old** org. The site then shows up in `GET /organizations/{orgId}/sites` (enrollment-filtered) but is **missing from the ASO site picker**, which groups sites under projects from `GET /organizations/{orgId}/projects`. This is the `zepbound.lilly.com` symptom in SITES-46200.

`PATCH /sites/:siteId` currently *allows* `projectId` changes. Because a `projectId` references a project that is itself scoped to an `organizationId`, an ad-hoc `projectId` patch can point a site at a project in a different org and recreate exactly this mismatch.

## Change

Reject `projectId` changes on `PATCH /sites/:siteId` the same way `organizationId` changes are already rejected:

```js
if (hasText(requestBody.projectId)
  && requestBody.projectId !== site.getProjectId()) {
  return forbidden('Updating project ID is not allowed');
}
```

A matching/absent `projectId` is a no-op (does not error, does not trigger a save), mirroring the `organizationId` guard. Re-parenting must go through controlled flows (e.g. `PATCH /projects/:projectId`), not free-form site `projectId` updates.

This is one of several SITES-46200 follow-ups; the `set imsorg` project move/split and the ASO picker "unassigned" hardening are separate PRs.

### Tests
- Inverted the former `updates site with projectId` test -> now asserts **403** + no `setProjectId`/`save`.
- Added a test that a `projectId` equal to the current value is ignored (no error, save still proceeds for other fields).
- Full `test/controllers/sites.test.js` suite green (336 passing).

### Note on the API spec
The `SiteUpdate` schema lists `projectId` exactly as it lists `organizationId`. `organizationId` has had this same 403 behavior for a while with **no** schema note, so to keep the schema internally consistent I did **not** add a note for only `projectId`. Happy to document both in a follow-up if reviewers prefer.

## Related Issues
- SITES-46200

### DIFF
--- a/docs/openapi/site-api.yaml
+++ b/docs/openapi/site-api.yaml
@@ -61,6 +61,12 @@ site:
       This endpoint is useful for updating a site.
       Only the fields as per the request body schema will be updated.
       At least one field to update must be provided in the request body.
+
+      Re-parenting a site is not allowed through this endpoint: changing
+      `organizationId` or `projectId` returns 403. Both reference org-scoped
+      records, so re-parenting must go through controlled flows (e.g.
+      `PATCH /projects/{projectId}`) to keep a site and its project in the same
+      organization.
     operationId: patchSite
     security:
       - admin_key: [ ]
@@ -77,6 +83,18 @@ site:
           application/json:
             schema:
               $ref: './schemas.yaml#/Site'
+      '403':
+        description: >-
+          Attempted to change a field that is not allowed via this endpoint
+          (organizationId or projectId).
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                message:
+                  type: string
+                  example: "Updating project ID is not allowed"
       '400':
         $ref: './responses.yaml#/400-no-site-id-request-body-no-updates'
       '401':
--- a/src/controllers/sites.js
+++ b/src/controllers/sites.js
@@ -786,6 +786,16 @@ function SitesController(ctx, log, env) {
       return forbidden('Updating organization ID is not allowed');
     }
 
+    // A projectId references a project record that is itself scoped to an
+    // organizationId. Allowing ad-hoc projectId patches here lets a site point
+    // at a project in a different org, which hides the site from the org's site
+    // picker (SITES-46200). Re-parenting must go through controlled flows
+    // (e.g. PATCH /projects/:projectId), so reject it here like organizationId.
+    if (hasText(requestBody.projectId)
+      && requestBody.projectId !== site.getProjectId()) {
+      return forbidden('Updating project ID is not allowed');
+    }
+
     if (requestBody.name !== site.getName()) {
       site.setName(requestBody.name);
       updates = true;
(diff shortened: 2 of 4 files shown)


MysticatBot (commented):
Hey @alinarublea,

**Verdict: Comment** - clean fix, no blocking issues.
_Changes: Rejects projectId changes on PATCH /sites/:siteId with a 403 to prevent cross-org site misplacement, mirroring the existing organizationId guard (2 files)._

<details>

- suggestion: Add an integration test in `test/it/shared/tests/sites.js` exercising the new 403 rejection path. The CLAUDE.md rule ("modified endpoints must include integration tests") applies here, though the existing `organizationId` guard has the same gap and unit coverage is complete - `test/it/shared/tests/sites.js`
- suggestion: Document the 403 response in the OpenAPI spec for `patchSite` (applies equally to the existing `organizationId` guard - a follow-up for both would be cleanest) - `docs/openapi/site-api.yaml`
- suggestion: Consider edge case tests for when `site.getProjectId()` returns null (site never assigned to a project) and when `requestBody.projectId` is empty string - both are correctly handled by `hasText` but undocumented in tests - `test/controllers/sites.test.js`
- nit: The describe block `'isPrimaryLocale, language, and region updates'` now primarily tests projectId rejection - name is slightly misleading - `test/controllers/sites.test.js:4469`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 53s | <b>Cost:</b> $3.54 | <b>Commit:</b> <code>7f90394e9e0cf4c1b29e7c6d448f3453abb91ab1</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

alinarublea:
Thanks for the review! Addressed the non-blocking suggestions in `ef99830e`:

- **Integration test**: added `user: returns 403 when trying to change projectId` in `test/it/shared/tests/sites.js`, mirroring the existing `organizationId` IT (patches `SITE_1` → `PROJECT_2_ID`).
- **Edge cases**: added a unit test for the empty-string `projectId` case (skipped by `hasText`, no 403, other fields still save). The matching-`projectId` no-op case was already covered.
- **OpenAPI**: documented the `403` response on `patchSite` in `docs/openapi/site-api.yaml`, covering both `organizationId` and `projectId` re-parenting attempts.
- **Nit**: renamed the unit describe block to `projectId, isPrimaryLocale, language, and region updates`.

MysticatBot (approved):
Hey @alinarublea,

**Verdict: Approve** - clean fix, prior suggestions all addressed.
_Changes: Rejects projectId changes on PATCH /sites/:siteId with a 403 to prevent cross-org site misplacement, adds OpenAPI docs, unit tests (including edge cases), and an integration test (4 files)._

<details>

- nit: No unit test for the case where `site.getProjectId()` returns `null` (site never assigned to a project) - the guard correctly returns 403 in this case by construction (`'some-uuid' !== null` is `true`), but a test would document the intent - `test/controllers/sites.test.js`
</details>

- Integration test added for the 403 rejection path
- OpenAPI 403 response documented for patchSite (covers both organizationId and projectId)
- Edge-case unit test for empty-string projectId added
- Describe block renamed to reflect projectId coverage

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 38s | <b>Cost:</b> $2.36 | <b>Commit:</b> <code>ef99830e1378dd0d9a0f153943bab8271fcc2e46</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2575, opened 2026-06-08
feat(preflight): restore mystiqueUrl dev override on site-scoped endpoint (SITES-46216)

GeezerPelletier:
## Summary

Restores the `mystiqueUrl` body parameter override on `POST /sites/:siteId/preflights`. The override existed on the legacy `/preflight/beta/jobs` endpoint (PR #2140, hardened in 746138e4) but wasn't carried forward when SITES-44686 (PR #2478) replaced beta/jobs with the site-scoped endpoint. SITES-46216 puts it back.

## Why

The current dev env has `MYSTIQUE_API_BASE_URL` configured to an Ethos provisioner URL (`*.stage.cloud.adobe.io`) whose CNAME chain passes through `*.corp.ethos.adobe.net` — **not resolvable from public AWS Lambda DNS**. Result: every call to the new endpoint surfaces as `502 PREFLIGHT_UPSTREAM_ERROR` with `fetch failed`. Diagnosis trail in the SITES-46216 ticket.

The dev override is a parallel mechanism that lets per-request callers point at a reachable Mysticat host (e.g. `https://m-dev.adobe.io`) without waiting on a deploy-cycle secret update. Same approach the team had on the legacy endpoint; just porting it.

## Guards

Mirroring the original implementation with one allowlist widening:

1. **Environment gate** — only honored when `env.AWS_ENV !== 'prod'`. Dead code in production regardless of body content.
2. **Valid URL parse**.
3. **Hostname allowlist: `*.adobe.io`**. Broader than the original `*.stage.cloud.adobe.io` because corp-only Ethos hosts proved unreachable from public Lambda networking, and `m-dev.adobe.io` (the publicly-reachable canonical dev host) doesn't match the old pattern. `*.adobe.io` stays scoped to Adobe-owned hosts.
4. **Tenancy boundary unchanged** — caller must still pass `accessControlUtil.hasAccess(site)`.

When the override is present and valid, `callMysticatAnalyze` receives the override URL instead of `env.MYSTIQUE_API_BASE_URL`. The override also sidesteps the "Analyze service not configured" 500 — useful when the secret hasn't been wired yet for a new env.

## What changed

**`src/controllers/preflight.js`** — added override-validation block at the start of `createPreflight` (after URL/data validation, before the env-var check). Computed `mysticatBaseUrl` is threaded through to `callMysticatAnalyze`.

**`test/controllers/preflight.test.js`** — 6 new tests under the `createPreflight` describe block:
- honors override on non-prod with `*.adobe.io` host (202 + correct fetch target)
- falls back to env var when override absent on non-prod
- rejects invalid URL with 400 PREFLIGHT_INVALID_REQUEST
- rejects non-`*.adobe.io` host with 400 PREFLIGHT_INVALID_REQUEST
- ignores override in prod (dead code path)
- allows override even when MYSTIQUE_API_BASE_URL env is empty (non-prod)

## Test plan

- [x] `npx mocha test/controllers/preflight.test.js --grep createPreflight` → 60 passing (was 54, +6 for the new override paths)
- [x] No OpenAPI changes (the override stays undocumented per the fd027b5a precedent for the legacy endpoint)
- [ ] Reviewer: confirm the security guard set is sufficient (env-gate + valid-URL + hostname allowlist + access check)
- [ ] After merge: I'll update `docs/isolated-dev-environment.md` in `aem-sites-optimizer-preflight` to reflect the new endpoint shape

## Security review notes

Threat model:
- Override path is gated on **non-prod** + **org-membership** + **`*.adobe.io` hostname allowlist** — all three guards must clear.
- SSRF risk scoped to Adobe-owned hosts; even within `*.adobe.io`, no credentials are forwarded by the controller (the `Authorization` header is the page-auth bearer from `retrievePageAuthentication`, only set when `checkEnableAuthentication` returns true, and is only relevant to the customer URL — not to the Mysticat host).
- Same threat shape the team already approved on the legacy `/preflight/beta/jobs` endpoint (PR #2140, 746138e4). Just porting forward; not introducing a new pattern.

If anyone wants stricter guards (signed override, audit-log on use, separate scope check) — happy to add. The historical precedent felt sufficient here for the same risk level.

## Notes

- This PR is purely additive for non-prod. Prod behavior is unchanged.
- The proper long-term fix is still to update the dev-env secret to point at a publicly-reachable Mysticat host. That's a parallel ops task; this PR just removes the dev-blocking gap in the meantime.
- Doc update for `aem-sites-optimizer-preflight` will follow as a separate PR.

Related: [SITES-44686](https://jira.corp.adobe.com/browse/SITES-44686) (parent — the redesign that dropped the override), [SITES-46202](https://jira.corp.adobe.com/browse/SITES-46202) (the most-recent preflight controller change just landed in PR #2572).

### DIFF
--- a/src/controllers/preflight.js
+++ b/src/controllers/preflight.js
@@ -365,7 +365,35 @@ function PreflightController(ctx, log, env) {
       return preflightError('PREFLIGHT_INVALID_REQUEST', 'url is missing or not a valid URI', 400);
     }
 
-    if (!hasText(env.MYSTIQUE_API_BASE_URL)) {
+    // mystiqueUrl override (SITES-46216): in non-prod, allow the caller to
+    // point this request at a specific Mysticat host instead of the
+    // env-configured one. Same shape as the legacy /preflight/beta/jobs
+    // override (PR #2140, hardened in 746138e4), restored here after the
+    // SITES-44686 redesign dropped it. Guards:
+    //   1. AWS_ENV !== 'prod' — dead code in prod regardless of body content
+    //   2. Valid URL parse
+    //   3. Hostname suffix-match against *.adobe.io — broader than the
+    //      original *.stage.cloud.adobe.io because corp-only Ethos hosts
+    //      proved unreachable from public Lambda networking, and m-dev.adobe.io
+    //      is the current publicly-reachable canonical dev host
+    //   4. Tenancy boundary unchanged — caller must still pass hasAccess(site)
+    const isDevForOverride = env.AWS_ENV !== 'prod';
+    const useMystiqueUrlOverride = isDevForOverride && hasText(data.mystiqueUrl);
+    if (useMystiqueUrlOverride) {
+      if (!isValidUrl(data.mystiqueUrl)) {
+        return preflightError('PREFLIGHT_INVALID_REQUEST', 'mystiqueUrl must be a valid URL', 400);
+      }
+      const overrideHost = new URL(data.mystiqueUrl).hostname;
+      if (!/\.adobe\.io$/.test(overrideHost)) {
+        return preflightError('PREFLIGHT_INVALID_REQUEST', 'mystiqueUrl must point at an *.adobe.io host', 400);
+      }
+    }
+
+    const mysticatBaseUrl = useMystiqueUrlOverride
+      ? data.mystiqueUrl
+      : env.MYSTIQUE_API_BASE_URL;
+
+    if (!hasText(mysticatBaseUrl)) {
       return preflightError('PREFLIGHT_INTERNAL_ERROR', 'Analyze service not configured', 500);
     }
 
@@ -481,7 +509,7 @@ function PreflightController(ctx, log, env) {
 
     try {
       await callMysticatAnalyze(
-        env.MYSTIQUE_API_BASE_URL,
+        mysticatBaseUrl,
         asyncJob.getId(),
         siteId,
         url,
--- a/src/controllers/preflight.js
+++ b/src/controllers/preflight.js
@@ -383,10 +383,14 @@ function PreflightController(ctx, log, env) {
       if (!isValidUrl(data.mystiqueUrl)) {
         return preflightError('PREFLIGHT_INVALID_REQUEST', 'mystiqueUrl must be a valid URL', 400);
       }
-      const overrideHost = new URL(data.mystiqueUrl).hostname;
-      if (!/\.adobe\.io$/.test(overrideHost)) {
+      const parsedOverride = new URL(data.mystiqueUrl);
+      if (parsedOverride.protocol !== 'https:') {
+        return preflightError('PREFLIGHT_INVALID_REQUEST', 'mystiqueUrl must use https://', 400);
+      }
+      if (!/\.adobe\.io$/.test(parsedOverride.hostname)) {
         return preflightError('PREFLIGHT_INVALID_REQUEST', 'mystiqueUrl must point at an *.adobe.io host', 400);
       }
+      log.info(`Using caller-supplied mystiqueUrl override: ${data.mystiqueUrl}`);
     }
 
     const mysticatBaseUrl = useMystiqueUrlOverride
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @GeezerPelletier,

**Verdict: Approve**
_Changes: Restores the `mystiqueUrl` body parameter override on `POST /sites/:siteId/preflights`, gated on non-prod with URL validation and `*.adobe.io` hostname allowlist (2 files)._

<details>

- suggestion: Enforce HTTPS scheme on the override URL - the hostname allowlist validates `*.adobe.io` but accepts `http://`, which would send the request payload over plaintext. Adding `if (parsedOverride.protocol !== 'https:') return preflightError(...)` is cheap defense-in-depth. - `src/controllers/preflight.js:389`
- suggestion: Log when the override fires - a single `log.info({ mystiqueUrlOverride: data.mystiqueUrl }, 'using caller-supplied mystiqueUrl override')` would make non-prod debugging trivial when multiple callers hit the endpoint. - `src/controllers/preflight.js:393`
- nit: The "allows override even when env is empty" test asserts only `status === 202` but does not verify the fetch target URL. Adding `expect(fetchStub.secondCall.args[0]).to.equal('https://m-dev.adobe.io/v1/preflight/analyze')` would confirm the override was actually used, not just that the request avoided the 500. - `test/controllers/preflight.test.js:1342`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 37s | <b>Cost:</b> $2.67 | <b>Commit:</b> <code>f8868d6ec48042ccb867632f792285282389f509</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

GeezerPelletier:
Thanks @MysticatBot — all three applied in `2f85f0be`:

1. **HTTPS scheme enforced.** Parsed the override URL once, added a `parsedOverride.protocol !== 'https:'` check before the hostname allowlist. Returns `400 PREFLIGHT_INVALID_REQUEST` with `message: 'mystiqueUrl must use https://'`. New test pinning the rejection.

2. **Override use is logged.** Single `log.info(\`Using caller-supplied mystiqueUrl override: ${data.mystiqueUrl}\`)` after all validation passes. Should make non-prod debugging trivial when multiple callers hit the endpoint.

3. **Empty-env test now verifies the fetch target.** Added `expect(fetchStub.secondCall.args[0]).to.equal('https://m-dev.adobe.io/v1/preflight/analyze')` to the "allows override even when env is empty" test — confirms the override was actually used, not just that the 500 was avoided.

61 createPreflight tests passing (was 60, +1 for the http-scheme rejection).

MysticatBot (commented):
Hey @GeezerPelletier,

**Verdict: Comment** - clean re-review, all prior suggestions addressed.
_Changes: Restores the `mystiqueUrl` body parameter override on `POST /sites/:siteId/preflights`, gated on non-prod with URL validation, HTTPS enforcement, and `*.adobe.io` hostname allowlist (2 files)._

<details>

- nit: The log message uses string interpolation rather than structured key-value logging. `log.info('Using caller-supplied mystiqueUrl override', { mystiqueUrl: data.mystiqueUrl })` would be more consistent with Coralogix searchability. - `src/controllers/preflight.js:397`
- suggestion: Consider using `parsedOverride.origin` instead of raw `data.mystiqueUrl` when assigning `mysticatBaseUrl`. This strips any path/query/fragment the caller supplied, guaranteeing they control only which host is hit, not the request path. Low priority given the non-prod gate. - `src/controllers/preflight.js:399`
- nit: The "invalid URL" test omits `attributes: { authInfo: mockAuthInfo }` from its request fixture while all other override tests include it. Adding it would make the fixture consistent and isolate the validation behavior under test. - `test/controllers/preflight.test.js:1109`
</details>

- HTTPS scheme now enforced (protocol check added)
- Override use is now logged
- Empty-env test now verifies the fetch target URL

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 10s | <b>Cost:</b> $2.57 | <b>Commit:</b> <code>2f85f0be2ba19891225e9b2e13f5c97beaf4acbf</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2576, opened 2026-06-09
feat(llmo): auto-create missing value columns on sheet-row patch

dzehnder:
## What

Makes the LLMO sheet-write endpoint (`PATCH /sites/:siteId/llmo/data/.../row`) create a `values` column on the fly when it isn't already in the worksheet header, instead of rejecting the write with a 400.

## Why

GSC prompt dismissal needs to set a `status` column on the GSC workbook (`ai-prompts-gsc.xlsx`). The GSC generator in data-retrieval-service does not emit a `status` column, so the first dismissal had no column to write into and `patchSheetRows` returned `400 Unknown column`. Rather than pre-emitting the column from the generator, the dismiss-write path now lazily creates it — matching the request *"when a prompt is dismissed from GSC, the sheet should be updated; if the column does not exist yet, add it."*

This was discussed in the GSC prompt-dismissal thread: SharePoint stays the source of truth (no DB table), consistent with how Strategic Recommendations soft-delete already uses this same generic endpoint.

## Behavior

- **`values` columns** missing from the header row are appended after the last existing column and written. The cached header map is mutated in place so a batch creates the column **once** and reuses it across entries.
- **`match` columns** must still exist — an unknown match column returns `400`, since you can't identify a row by a column the sheet doesn't have.
- Existing data rows leave the new cell empty; only the matched row gets a value.
- Universal: applies to all callers of the endpoint (strategic-recommendations, brand-presence, GSC), not gated behind a flag.

## ⚠️ Known limitation (intentional, out of scope)

This is the **dismiss-write path only**. GSC regeneration overwrites `ai-prompts-gsc.xlsx` with a freshly built workbook that has no `status` column, so dismissals reset on the next GSC run. Making dismissals durable across regen requires a preserve-on-regen merge step in `data-retrieval-service` (the `mergeSemrushRows` equivalent), tracked separately. Accepted as ephemeral for now.

## Tests

- New unit tests: auto-create a value column (single + batch reuse), unknown *match* column still 400s.
- Updated the controller error-mapping test wording to match.
- `npx mocha test/controllers/llmo/llmo-sheet-write.test.js` → 44 passing
- `npx mocha test/controllers/llmo/llmo.test.js test/routes/index.test.js` → 401 passing
- `npm run docs:lint` → API description valid

### DIFF
--- a/docs/openapi/llmo-api.yaml
+++ b/docs/openapi/llmo-api.yaml
@@ -617,8 +617,10 @@ llmo-data-row-with-datasource:
       each other.
 
       Within each entry, the `match` object must uniquely identify exactly one row
-      (no match → 404, multiple matches → 409). Column names in `match` and `values`
-      must exist in the worksheet header row.
+      (no match → 404, multiple matches → 409). `match` column names must exist in the
+      worksheet header row (unknown → 400). `values` column names that do not exist yet
+      are created on the fly (appended after the last header column), so the first write
+      of e.g. a `status` column the generator never emitted succeeds rather than 400ing.
     operationId: patchLlmoDataRow
     requestBody:
       required: true
@@ -679,7 +681,9 @@ llmo-data-row-with-type:
 
       Used by the elmo UI to mark prompt(s) in the Strategic Recommendations
       workbook as deleted by writing `"deleted": "true"` on the matching row(s) of
-      the `Semrush`, `Synthetic Personas`, or `Citation Attempt` worksheet.
+      the `Semrush`, `Synthetic Personas`, or `Citation Attempt` worksheet, and to
+      dismiss GSC prompts by writing `"status": "dismissed"` on the matching row of
+      the GSC workbook — the `status` column is created automatically on first write.
     operationId: patchLlmoDataRowWithType
     requestBody:
       required: true
--- a/src/controllers/llmo/llmo-sheet-write.js
+++ b/src/controllers/llmo/llmo-sheet-write.js
@@ -220,13 +220,33 @@ export const validateSheetRowPatch = (data) => {
 // identical to the pre-batch contract.
 const updateWorksheet = (worksheet, entryRef, headerMap, match, values) => {
   const prefix = entryRef ? `${entryRef}: ` : '';
+
+  // `match` columns must already exist — you cannot identify a row by a column the
+  // worksheet doesn't have, so an unknown match column is a caller error (400).
   const unknownMatchCols = Object.keys(match).filter((c) => !headerMap.has(c));
-  const unknownValueCols = Object.keys(values).filter((c) => !headerMap.has(c));
-  if (unknownMatchCols.length > 0 || unknownValueCols.length > 0) {
+  if (unknownMatchCols.length > 0) {
     const headers = [...headerMap.keys()].join(', ');
-    const err = new Error(`${prefix}Unknown column(s) ${[...unknownMatchCols, ...unknownValueCols].join(', ')}. Available: ${headers}`);
+    const err = new Error(`${prefix}Unknown match column(s) ${unknownMatchCols.join(', ')}. Available: ${headers}`);
     err.statusCode = 400;
     throw err;
+  }
+
+  // `values` columns are created on demand: a value column that isn't in the header
+  // row yet is appended after the last existing column, so e.g. the first-ever GSC
+  // prompt dismissal can introduce a `status` column the generator never emitted.
+  // `headerMap` is mutated in place so subsequent entries in the same batch (which
+  // share the cached map) see the new column. Existing data rows leave the new cell
+  // empty — only the matched row below gets a value.
+  const missingValueCols = Object.keys(values).filter((c) => !headerMap.has(c));
+  if (missingValueCols.length > 0) {
+    const headerRow = worksheet.getRow(1);
+    let nextCol = headerMap.size > 0 ? Math.max(...headerMap.values()) : 0;
+    missingValueCols.forEach((column) => {
+      nextCol += 1;
+      headerRow.getCell(nextCol).value = column;
+      headerMap.set(column, nextCol);
+    });
+    headerRow.commit();
   }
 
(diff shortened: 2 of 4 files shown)


MysticatBot (commented):
Hey @dzehnder,

**Verdict: Comment** - clean, well-scoped change with solid test coverage; only minor hardening observations.
_Changes: Auto-creates missing value columns on the LLMO sheet-row patch endpoint instead of rejecting with 400, while keeping strict validation on match columns (4 files)._

<details>

- nit: No validation on auto-created column name count, length, or character set - an authenticated caller could gradually pollute the worksheet header with arbitrary keys. Consider validating that `values` keys match `SAFE_PATH_SEGMENT_RE` already used for path segments. - `src/controllers/llmo/llmo-sheet-write.js:230`
- nit: Both new tests create exactly one new column (`status`). The `forEach` loop assigning sequential `nextCol` indices is untested for the multi-column case (e.g. `values: { status: 'dismissed', priority: 'low' }` where both are new). - `test/controllers/llmo/llmo-sheet-write.test.js`
- suggestion: A test where a batch entry creates a new column and a later entry fails (e.g. no matching row) would explicitly prove column-creation is discarded with the rest of the in-memory workbook on failure. The guarantee holds today (upload only after all entries succeed), but the test would document it. - `test/controllers/llmo/llmo-sheet-write.test.js`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 55s | <b>Cost:</b> $4.09 | <b>Commit:</b> <code>8020fabff0befd62ecd95e43eae41220ff1872e1</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

dzehnder:
@jjenscodee — here's what's needed on the UI side (project-elmo-ui) to dismiss GSC prompts into the sheet once this merges.

**Backend recap:** this PR makes the existing sheet-write endpoint auto-create a `values` column if it's missing, so no `status` column needs to pre-exist. No new endpoint — the route `PATCH /sites/{siteId}/llmo/data/gsc/ai-prompts-gsc/row` already works.

### 1. Generalize the patch call
`patchLlmoDataRow` in `src/api/spacecat.ts` (~line 1670) is hardcoded to the strategic-recommendations path. Add `sheetType`/`dataSource` params (or a `patchGSCDataRow` variant) so it can hit:
```
PATCH /sites/{siteId}/llmo/data/gsc/ai-prompts-gsc/row
```

### 2. Wire the GSC dismiss handler
Send this body:
```json
{
  "updates": [{
    "sheet": "shared-Source GSC+Keywords",
    "match": { "prompt": "<prompt text>", "region": "<region>" },
    "values": { "status": "dismissed" }
  }]
}
```
Today the GSC "ignore" button writes to customer-config `ignoredPromptsData` via `useCustomerConfiguration.ignorePrompt()` — **not** the sheet. Per the thread (keep everything in the sheet), re-point the GSC ignore handler at this sheet-write call instead.

### 3. ⚠️ Sheet-name gotcha (this one bites)
You **read** GSC data under the JSON key `Source GSC+Keywords`, but the `sheet` field in the PATCH must be the real worksheet name `shared-Source GSC+Keywords` (Helix strips the `shared-` prefix in the JSON projection only). Wrong name → `404 Worksheet not found`. This mirrors how SR does it (`sheet: 'shared-Semrush'`).

### 4. Match key
GSC rows have no id column — the key is the prompt text. Use `match: { prompt, region }` (mirrors the existing `createPromptRegionKey`). Send raw values straight from the GSC JSON you rendered — matching is **exact string equality** against the cell, so don't transform/uppercase. If the same prompt+region appears in two topics you'll get a `409 ambiguous` — add `topic` (and/or `category`) to `match` to pin one row.

### 5. Read dismissed state back
After dismissal the GSC JSON rows carry a `status` field — add it to the `SuggestedPrompt` type + `normalizeSuggestedPrompt`, and filter/mark rows where `status === "dismissed"`. Update **optimistically** on click (like SR's `setApprovalMap`) — the endpoint republishes to Helix preview+live, so the CDN has propagation lag and an immediate refetch may not reflect the change yet.

### 6. Contract to confirm
The backend auto-creates whatever `values` column you send, so we just need to agree on the exact name+value. I put `status: "dismissed"` in the OpenAPI — shout if you'd rather match SR's `deleted: "ignored"` and I'll align the docs. (New column names are validated: letters/digits/space/`. _ + -`, max 100 chars.)

### ⚠️ Known limitation
GSC regeneration overwrites `ai-prompts-gsc.xlsx` without the status column, so dismissals reset on the next GSC run. Making them durable across regen needs a preserve-on-regen step in data-retrieval-service (tracked separately). For now, dismissed prompts can reappear after a GSC refresh.

MysticatBot (approved):
Hey @dzehnder,

**Verdict: Approve** - clean, well-scoped re-iteration that addresses all prior feedback.
_Changes: Auto-creates missing value columns on the LLMO sheet-row patch endpoint with input validation, instead of rejecting with 400 (4 files)._

<details>

- nit: `Math.max(...headerMap.values())` will throw a `RangeError` if spread into function args on a very large map. In practice Excel's 16,384-column limit is within V8's argument ceiling, so this is theoretical - but `[...headerMap.values()].reduce((a, b) => Math.max(a, b), 0)` would be unconditionally safe if you want to harden it. - `src/controllers/llmo/llmo-sheet-write.js:253`

</details>

- Column name validation added (`isCreatableColumnName` with regex + length cap)
- Multi-column creation test case added
- All-or-nothing failure test (column created then discarded on later entry failure) added

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 51s | <b>Cost:</b> $2.78 | <b>Commit:</b> <code>eddd743ed0c72321e221629b900f0e987f8246a0</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2577, opened 2026-06-09
feat: ai visibility brand data

dmitriigulkin-semrush:
Closes https://jira.corp.adobe.com/browse/LLMO-5464


calvarezg (approved):
Approving — clean, additive read-only endpoints that follow the existing v1-handler patterns. A few non-blocking items:

**1. `isWorldwideCountry` may misfire depending on proto enum string form.** With `enumAsInteger: false`, `country.country` serializes to the enum *name* string, so the `country === COUNTRY_ENUM.WORLDWIDE` integer comparison is effectively dead — only the string-normalization branch (`=== 'WORLDWIDE' || === 'WW'`) can match. If the serialized enum name carries a prefix (e.g. `COUNTRY_WORLDWIDE`), the check fails and a duplicate `WORLDWIDE` entry gets appended even when Semrush already returns one. Worth confirming against the actual proto enum naming.

**2. Doc typo — stray Cyrillic character.** `docs/openapi/ai-visibility-api.yaml:884` reads `Еhe response appends…` — the leading character is a Cyrillic `Е` (U+0415), not `Th`. Should be `The`. This also propagated into the generated `docs/index.html`, so fix the source and regenerate.

**3. `dateFrom`/`dateTo` pairing not enforced (optional).** The description notes Semrush requires both dates omitted or both present, but `stats-by-llm` forwards them independently and relies on Semrush to reject a half-specified range. Consistent with the pass-through proxy approach — a cheap pre-check would just yield a cleaner 400 than a downstream gRPC error.

calvarezg:
**Code review summary**

Plumbing is solid and consistent with existing v1 handler patterns. A few issues:

**Correctness**
- `meta.js:119` — `computeWorldwideMeta` is always appended unconditionally. If Semrush starts returning a `WORLDWIDE` country, the response will contain duplicates. Guard with `countries.some(c => c.country === WORLDWIDE_COUNTRY)` before appending.

**Consistency**
- `stats-by-country.js:28`, `stats-by-llm.js:28` — `domain` is used without validation. Existing v1 brand handlers return a 400 when `domain` is missing; these should match.


**Minor**
- `/v1/meta/meta` double-segment path looks like a typo; a comment explaining it mirrors the gRPC `Meta/Meta` naming would help.
- `meta.js` uses `messageToJson` from `proto-json.js`; brand handlers use `toJson` from `@bufbuild/protobuf` directly — the intentional difference (`useProtoFieldName: true`) isn't obvious from the imports.

dmitriigulkin-semrush:
> **Code review summary**
> 
> Plumbing is solid and consistent with existing v1 handler patterns. A few issues:
> 
> **Correctness**
> 
> * `meta.js:119` — `computeWorldwideMeta` is always appended unconditionally. If Semrush starts returning a `WORLDWIDE` country, the response will contain duplicates. Guard with `countries.some(c => c.country === WORLDWIDE_COUNTRY)` before appending.
> 
> **Consistency**
> 
> * `stats-by-country.js:28`, `stats-by-llm.js:28` — `domain` is used without validation. Existing v1 brand handlers return a 400 when `domain` is missing; these should match.
> 
> **Minor**
> 
> * `/v1/meta/meta` double-segment path looks like a typo; a comment explaining it mirrors the gRPC `Meta/Meta` naming would help.
> * `meta.js` uses `messageToJson` from `proto-json.js`; brand handlers use `toJson` from `@bufbuild/protobuf` directly — the intentional difference (`useProtoFieldName: true`) isn't obvious from the imports.

hey @calvarezg, thanks!!

`computeWorldwideMeta` is always appended unconditionally
that's expected, SR API never returns WORLDWIDE and we have no plans on adding that

`stats-by-country.js:28`, `stats-by-llm.js:28` — `domain` is used without validation. 
I couldn't find any v1/ handlers that would have this behaviour, and it's not needed anyways since we rely on validation from grpc message build or from the upstream 

/v1/meta/meta` double-segment path looks like a typo
I myself think we should stick with notation from original proto files, but we ofc can rename it to smth like /meta/snapshots/, lmk what you think

meta.js uses messageToJson from proto-json.js;
fixed

calvarezg:
Thanks for the update. The proto-serialization inconsistency in `meta.js` is resolved. The other three items from the previous review are still open:

1. **WORLDWIDE deduplication** (`meta.js:119`) — still appends unconditionally. If Semrush ever returns a `WORLDWIDE` country the response will have duplicates. A one-liner guard fixes it:
   ```js
   const alreadyHasWorldwide = countries.some((c) => c.country === WORLDWIDE_COUNTRY);
   countries: [...countries, ...(alreadyHasWorldwide ? [] : [computeWorldwideMeta(countries)])],
   ```

2. **Missing `domain` validation** (`stats-by-country.js:28`, `stats-by-llm.js:28`) — `domain` is passed directly to gRPC without checking for null. Other v1 brand handlers return a 400 for missing `domain`; these should match.

3. **`/* c8 ignore */` wraps all of `meta.js`** — the pure helper functions (`computeWorldwideMeta`, date intersection/union logic) have no gRPC dependencies and are worth testing separately. The gRPC call itself is fine to ignore, but the transformation logic has real edge cases.

calvarezg (approved):
Thanks for the explanations — all makes sense. LGTM.

PR #2578, opened 2026-06-09
fix: generate fallback slug for all-multibyte category names | LLMO-5473

igor-grubic:
**Ticket:** [LLMO-5473](https://jira.corp.adobe.com/browse/LLMO-5473)

## Context

Creating a category whose name has **no ASCII alphanumerics** (e.g. CJK `カテゴリ`, `中文名稱`) returns **HTTP 500**:

```
API Error: 500 - {"message":"Category name produces an empty slug after normalization; supply an explicit `id`"}
```

Reported by Jiang Long for paying customer Daiichi Life (`#llm-optimizer-engineering`). This is **not** the same bug as [LLMO-4250](https://jira.corp.adobe.com/browse/LLMO-4250) — that fix ([project-elmo-ui#1508](https://github.com/adobe/project-elmo-ui/pull/1508) + #2199) addressed the *prompt-upload* path. The **category-create** endpoint (`POST /v2/orgs/:id/categories`) still slugified the name and threw on an empty result. Tracked as LLMO-5473.

## Root cause

`createCategory()` in `src/support/categories-storage.js` derives `category_id` via `name.toLowerCase().replace(/[^a-z0-9]+/g, '-').replace(/^-+|-+$/g, '')`. For an all-multibyte name this yields an empty string, and the code `throw`s a plain `Error` → controller maps it to a 500.

## Fix

When the derived slug is empty, fall back to a **synthetic UUID slug** (`crypto.randomUUID()`) instead of throwing. This mirrors the DB column default `category_id text DEFAULT (gen_random_uuid())::text NOT NULL`. It is safe because:

- Idempotency keys on **name** (`findCategoryByName` runs first), so the same name re-resolves to the existing row and never regenerates a slug.
- Every real FK (`category_prompts`, `prompts`, `brand_presence_executions`, …) targets `categories.id` (the UUID v7 PK), **not** the `category_id` text slug — the slug is never an FK target.

An explicit client-supplied `id` still wins. No controller/route/DTO/OpenAPI change needed (categories aren't in `docs/openapi/paths/`; `id` is already optional).

## Tests

- **Unit** (`test/support/categories-storage.test.js`): flipped the degenerate-slug test from "rejects" to "generates a synthetic slug"; added a punctuation-only case; kept the explicit-id test. Added a `createCaptureInsertClient` helper.
- **Integration** (`test/it/shared/tests/categories-prompts.js`): POST `{ name: 'カテゴリ' }` (no id) → 201 with a UUID-shaped id and preserved name; re-POST → 200 idempotent, same id.

## Verification

- `npm test` — 11,608 passing locally.
- `npm run lint` — clean.
- Postgres IT suite needs Docker + ECR (not available locally) — relying on CI.

## Out of scope

- [LLMO-4298](https://jira.corp.adobe.com/browse/LLMO-4298) (Created By / Uploaded By showing "Adobe Internal") — separate ticket.

### DIFF
--- a/docs/openapi/categories-v2-api.yaml
+++ b/docs/openapi/categories-v2-api.yaml
@@ -118,7 +118,7 @@ v2-categories-for-org:
              deleted) between our lookup and our update — message
              `Category was concurrently modified; please retry`.
           2. A unique-constraint other than `uq_category_name_per_org` was
-             tripped on insert (e.g. `uq_category_id_per_org` when the
+             tripped on insert (e.g. `uq_category_per_org` when the
              client-supplied slug collides with an unrelated existing
              row). Message echoes the actual constraint name.
           Treated as an idempotent retry signal by DRS-class clients;
--- a/src/support/categories-storage.js
+++ b/src/support/categories-storage.js
@@ -10,6 +10,7 @@
  * governing permissions and limitations under the License.
  */
 
+import crypto from 'node:crypto';
 import { hasText } from '@adobe/spacecat-shared-utils';
 
 // Zero-width / BOM characters that pass `trim()` but render invisible. Strip
@@ -274,7 +275,7 @@ export async function createCategory({
   // Derive a slug from the canonical name when none supplied, trimming
   // leading/trailing dashes so "   !!  " and Unicode-only names don't
   // produce a degenerate bare '-'. The client-supplied slug is trusted
-  // verbatim (FK-stability).
+  // verbatim.
   let derivedSlug;
   if (category.id) {
     derivedSlug = category.id;
@@ -283,10 +284,16 @@ export async function createCategory({
       .toLowerCase()
       .replace(/[^a-z0-9]+/g, '-')
       .replace(/^-+|-+$/g, '');
+    // Names with no ASCII alphanumerics (e.g. CJK "カテゴリ") and
+    // punctuation-only names slugify to empty. Fall back to a synthetic
+    // UUID slug rather than failing — mirrors the DB column default
+    // (gen_random_uuid()::text). Idempotency keys on `name` (lookup above)
+    // and every FK targets categories.id (the UUID PK), not this text slug,
+    // so a synthetic category_id is safe and stable across re-posts. The
+    // same name re-resolves to the existing row and never regenerates.
+    // LLMO-5473.
     if (!derivedSlug) {
-      throw new Error(
-        'Category name produces an empty slug after normalization; supply an explicit `id`',
-      );
+      derivedSlug = crypto.randomUUID();
     }
   }
 
--- a/src/support/categories-storage.js
+++ b/src/support/categories-storage.js
@@ -294,6 +294,15 @@ export async function createCategory({
     // LLMO-5473.
     if (!derivedSlug) {
       derivedSlug = crypto.randomUUID();
+      // `category_id` is now polymorphic (human-readable slug OR synthetic
+      // UUID). Emit a structured signal so operators can quantify the
+      // fallback rate and find affected rows in Coralogix — otherwise the
+      // UUID slugs are indistinguishable from client-supplied ones.
+      log?.warn?.('Category name slugifies to empty; generated synthetic UUID slug', {
+        organization_id: organizationId,
+        category_id: derivedSlug,
+        name: canonicalName,
+      });
     }
   }
 
@@ -338,7 +347,7 @@ export async function createCategory({
         }
       } else {
         // Any other unique-constraint violation (e.g. slug collision via
-        // uq_category_id_per_org when the client ships a drifted `id` that
+        // uq_category_per_org when the client ships a drifted `id` that
         // maps to a different name already occupying that slug) surfaces as
         // a typed 409 echoing the actual constraint — mirrors the topics
         // pattern so callers don't have to mine 500 bodies. LLMO-4370 #5/#6.
--- a/test/it/shared/tests/categories-prompts.js
+++ b/test/it/shared/tests/categories-prompts.js
@@ -147,5 +147,47 @@ export default function categoriesPromptsTests(getHttpClient, resetData) {
         expect(matches[0].name).to.equal('Test');
       });
     });
+
+    // ── Multibyte / non-ASCII-only names (LLMO-5473) ──
+
+    describe('Category creation with an all-multibyte name (no explicit id)', () => {
+      const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+
+      before(() => resetData());
+
+      it('creates a CJK-named category, generating a synthetic slug, and is idempotent by name', async () => {
+        const http = getHttpClient();
+
+        // Name slugifies to empty (no ASCII alphanumerics). Previously this
+        // threw "empty slug after normalization" → HTTP 500 (LLMO-5473). Now
+        // the API falls back to a synthetic UUID slug and the create succeeds.
+        const res1 = await http.admin.post(
+          `/v2/orgs/${ORG_1_ID}/categories`,
+          { name: 'カテゴリ' },
+        );
+        expect(res1.status).to.equal(201);
+        expect(res1.body.name).to.equal('カテゴリ');
+        expect(res1.body.id).to.match(UUID_RE);
+
+        const generatedId = res1.body.id;
+
+        // Idempotent by name: re-posting the same name resolves the existing
+        // row (no new synthetic slug) and returns 200.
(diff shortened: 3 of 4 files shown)


MysticatBot (commented):
Hey @igor-grubic,

**Verdict: Comment** - clean fix, no blocking concerns.
_Changes: Replaces the throw on degenerate category slugs (all-multibyte/punctuation-only names) with a `crypto.randomUUID()` fallback, fixing HTTP 500 for CJK category names (3 files)._

<details>

- nit: Comment block (7 lines) is verbose for a 1-line fallback; could trim to ~2 lines while preserving the safety rationale (e.g. "All-multibyte/punctuation-only names slugify to empty. Fall back to synthetic UUID - safe because FKs target .id, not this slug.") - `src/support/categories-storage.js:286`
- nit: Unit test asserts on `captured.row.category_id` (inserted row) but not on `result.category.category_id` (returned DTO); the integration test covers end-to-end shape, but this unit test has a gap if the row-to-result mapping ever diverges - `test/support/categories-storage.test.js:233`
- nit: Punctuation-only test omits `updatedBy` while the CJK test includes it; inconsistency makes it unclear whether the omission is intentional or accidental - `test/support/categories-storage.test.js:248`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 26s | <b>Cost:</b> $2.29 | <b>Commit:</b> <code>0e501feaaee4234fc4dd4677baa020034c8e50bf</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

JayKid:
## Multi-Persona PR Review

**Reviewed by:** Architect, DBA, Security, Tester, Junior Dev agents
**Verdict:** Request changes — fix logic is sound, 2 items to address before merge.

---

### Major: Add a unit test for the idempotency invariant

The safety argument ("same name never regenerates the slug") is load-bearing but has no unit test. `createCaptureInsertClient` hardcodes `maybeSingle` to `{ data: null }` in every test — the "lookup finds existing row → return early without UUID" branch is never exercised at the unit level.

Add a test where `maybeSingle` returns an existing row with a UUID `category_id` and assert `result.created === false` and no new UUID was generated.

---

### Major: Log a warning when the fallback fires

`category_id` is now silently polymorphic (slug or UUID) with no distinguishing flag and no log line. Production operators cannot surface the fallback frequency or find affected rows.

```js
if (!derivedSlug) {
  log.warn('Category name produces empty slug; generating synthetic UUID', { name: category.name, organizationId });
  derivedSlug = crypto.randomUUID();
}
```

---

### Major: Verify DB constraint name (medium confidence)

DBA agent found references to `uq_category_id_per_org` in unit test stubs but the actual migration may name it `uq_category_per_org`. If mismatched, the 409-conflict unit test validates a phantom constraint. Quick `\d categories` in the DB will confirm.

---

### Minor items (at author's discretion)

- Both new unit tests omit `expect(result.category.id).to.equal(captured.row.category_id)` — round-trip identity unasserted
- Mixed-script name `"AI カテゴリ"` not tested — verify UUID fallback does NOT fire when name has some ASCII
- `createCaptureInsertClient` missing `neq`/`order` stubs vs. real PostgREST client chain
- `id` param in `createCaptureInsertClient` should be renamed `rowId` to avoid collision with `body.id` naming
- UUID bypasses the slug's trailing-dash sanitization pipeline silently — either run it through or make the exemption explicit

---

### What's working well

Fix is correct and minimal. FK safety verified. Integration test covers the golden path + idempotency. Codecov shows 100% coverage on modified lines. No security or data integrity regressions.

igor-grubic:
Thanks for the thorough multi-persona pass — all three Majors were fair. Addressed in `dbf74cb6`:

**Major 1 — idempotency unit test.** Added `does NOT regenerate a synthetic slug when a non-ASCII name already exists (idempotent by name)`: an existing CJK row → `created:false`, `noop`, the existing slug is returned unchanged, a single round-trip (`from` called once), and no fallback WARN. One nuance: the lookup-finds-existing → early-return branch *was* already exercised (e.g. `short-circuits (no write) when an existing row already matches`, `does NOT downgrade origin from human to ai on an idempotent re-POST`) — the name-lookup runs *before* slug derivation, so it's name-agnostic. The new test makes the load-bearing "same name never regenerates the slug" invariant explicit at the unit level rather than leaning on the integration test alone.

**Major 2 — log on fallback.** Done. Emits a structured WARN matching the existing `resolveExistingCategory` downgrade-warn style (`organization_id`, `category_id`, `name`) so the now-polymorphic `category_id` (slug vs synthetic UUID) is quantifiable and affected rows are findable in Coralogix. Covered by a unit assertion in the CJK test.

**Major 3 — constraint name.** Good catch, and confirmed: the authoritative schema (`mysticat-data-service/db/schema.sql`) names it `uq_category_per_org` — `uq_category_id_per_org` was a phantom (pre-existing from LLMO-4370, no logic depended on it; the only equality check is against `uq_category_name_per_org`). Corrected the source comment, the 409 test fixture + assertion, and the OpenAPI description.

**Minor items:**
- ✅ Round-trip identity (`result.category.id === captured.row.category_id`) — added to both fallback tests.
- ✅ Mixed-script `"AI カテゴリ"` — added a test proving the fallback does **not** fire (slug stays `"ai"`, no WARN), so the UUID path is narrowly scoped to genuinely-empty slugs.
- ✅ Renamed the capture-helper param `id` → `rowId`.
- ⏭️ `neq`/`order` stubs — the insert path's lookup only calls `.eq().ilike().maybeSingle()`, so they'd be dead stubs here; skipped.
- ⏭️ UUID-through-sanitization — `crypto.randomUUID()` is well-formed (no leading/trailing dash), so the trailing-dash strip is a no-op on it; left explicit in the comment instead.

**Note on docs:** fixed the constraint name in the OpenAPI *source* (`categories-v2-api.yaml`) but deliberately did **not** commit a regenerated `docs/index.html` — rebuilding it pulls in unrelated drift (the committed artifact lags main by ~8 releases: embedded spec version `1.554.2` → `1.562.0`, plus redoc build-hash/tab-id churn). Bundling that into this PR would obscure the diff; the HTML is regenerated out-of-band.

Full unit suite green, lint clean.

JayKid:
## Delta Review — commit `dbf74cb6`

**Previous verdict:** Request changes (3 Majors, 7 Minors)
**This review:** Delta only — changes since `0e501fea`

---

### Resolution status

| Finding | Status |
|---------|--------|
| Major 1: Idempotency unit test | ✅ Resolved |
| Major 2: Log warn when fallback fires | ✅ Resolved |
| Major 3: Phantom constraint name `uq_category_id_per_org` | ✅ Resolved — fixed in docs, source comment, test fixture, test assertion |
| Minor: Round-trip identity assertion | ✅ Resolved — added to CJK, punctuation, and idempotency tests |
| Minor: Mixed-script `"AI カテゴリ"` not tested | ✅ Resolved — new test asserts slug stays `"ai"`, no WARN emitted |
| Minor: `id` → `rowId` rename | ✅ Resolved |
| Minor: `neq`/`order` stubs | ⏭️ Skipped — justification accepted |
| Minor: UUID through sanitization pipeline | ⏭️ Skipped — justification accepted |
| MysticatBot nit 1: verbose comment | ❌ Still open |
| MysticatBot nit 2: result DTO assertion gap | ✅ Resolved |
| MysticatBot nit 3: `updatedBy` inconsistency | ✅ Resolved |

---

### One remaining nit (not a blocker)

The punctuation-only test doesn't pass a `log` stub and doesn't assert the WARN fires — unlike the CJK test, which does both. Since `log?.warn?.()` skips silently when no logger is injected, the warn is untested for the punctuation path. Same code branch, so no correctness gap, but worth making consistent:

```js
const log = { warn: sinon.stub() };
const result = await createCategory({ ..., log });
expect(log.warn).to.have.been.calledOnce;
```

---

### What's working well

All three Majors are cleanly resolved. The idempotency test correctly simulates the lookup-finds-existing branch and asserts `created: false`, `outcome: 'noop'`, single `from` call, and no WARN — exactly the invariant needed. The structured `log?.warn?.()` with `organization_id`, `category_id`, and `name` gives operators everything needed to find affected rows. Constraint name corrected in all four locations.

**Verdict: Approve** — all blocking items resolved.

JayKid:
Ottimo lavoro, Igor! 🎉 Hai affrontato tutti i punti critici con precisione e chiarezza — il test sull'idempotenza, il logging strutturato e la correzione del nome del constraint erano esattamente quello di cui aveva bisogno la PR. Grazie per la risposta rapida e dettagliata. Mergi con fiducia!

JayKid (approved):
All blocking items resolved. Approving.

PR #2580, opened 2026-06-10
feat: auto-detect unreachable base URL at onboarding and persist working variant as overrideBaseURL

tkotthakota-adobe:
https://jira.corp.adobe.com/browse/SITES-46284

Problem and Fix:                                                                                                            
                                                                                                                       
When a site like dupont.com is onboarded with an unreachable base URL, the platform had no way to detect this at            registration time. Every subsequent audit would silently fail trying to reach an unreachable domain. The fix adds a reachability probe at site creation — if the registered base URL fails but the www-toggled variant responds, it             automatically saves the working URL as overrideBaseURL in the site config so all future audits use the correct domain from
the start.

### DIFF
--- a/src/controllers/sites.js
+++ b/src/controllers/sites.js
@@ -47,7 +47,8 @@ import {
 import AccessControlUtil from '../support/access-control-util.js';
 import { CAP_SITE_READ_ALL, CAP_SITE_CREATE } from '../routes/capability-constants.js';
 import { auditTargetURLsPatchGuard } from '../support/audit-target-urls-validation.js';
-import { updateRumConfig, autoDetectOverrideBaseURL } from '../support/rum-config-service.js';
+import { updateRumConfig } from '../support/rum-config-service.js';
+import { autoDetectOverrideBaseURL } from '../support/base-url-detection.js';
 import { triggerBrandProfileAgent } from '../support/brand-profile-trigger.js';
 import {
   ensureSiteEntitlementAndEnrollment,
@@ -401,12 +402,10 @@ function SitesController(ctx, log, env) {
           ...context.data,
           baseURL, // override with normalized value
         });
-        updateRumConfig(site, context).catch((e) => {
-          log.warn(`[sites] RUM config update failed for ${site.getBaseURL()}: ${e.message}`);
-        });
-        autoDetectOverrideBaseURL(site, context).catch((e) => {
-          log.warn(`[sites] base URL auto-detect failed for ${site.getBaseURL()}: ${e.message}`);
-        });
+        updateRumConfig(site, context)
+          .catch((e) => log.warn(`[sites] RUM config update failed for ${site.getBaseURL()}: ${e.message}`))
+          .then(() => autoDetectOverrideBaseURL(site, context))
+          .catch((e) => log.warn(`[sites] base URL auto-detect failed for ${site.getBaseURL()}: ${e.message}`));
         status = 201;
       }
     } catch (error) {
--- a/src/controllers/sites.js
+++ b/src/controllers/sites.js
@@ -47,7 +47,7 @@ import {
 import AccessControlUtil from '../support/access-control-util.js';
 import { CAP_SITE_READ_ALL, CAP_SITE_CREATE } from '../routes/capability-constants.js';
 import { auditTargetURLsPatchGuard } from '../support/audit-target-urls-validation.js';
-import { updateRumConfig } from '../support/rum-config-service.js';
+import { updateRumConfig, autoDetectOverrideBaseURL } from '../support/rum-config-service.js';
 import { triggerBrandProfileAgent } from '../support/brand-profile-trigger.js';
 import {
   ensureSiteEntitlementAndEnrollment,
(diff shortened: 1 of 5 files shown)


src/controllers/sites.js
```diff
@@ -404,6 +404,9 @@ function SitesController(ctx, log, env) {
         updateRumConfig(site, context).catch((e) => {
           log.warn(`[sites] RUM config update failed for ${site.getBaseURL()}: ${e.message}`);
         });
+        autoDetectOverrideBaseURL(site, context).catch((e) => {
+          log.warn(`[sites] base URL auto-detect failed for ${site.getBaseURL()}: ${e.message}`);
+        });
```
MysticatBot: issue (blocking): **Race condition between concurrent fire-and-forget saves.**

`updateRumConfig` and `autoDetectOverrideBaseURL` both run concurrently as detached promises on the same `site` instance. Both perform a read-modify-write cycle: `site.getConfig()` -> mutate -> `site.setConfig(Config.toDynamoItem(...))` -> `site.save()`. If both read config before either writes, the second save clobbers the first - one config mutation is silently lost (either `rumConfig` or `fetchConfig.overrideBaseURL`).

In practice `autoDetectOverrideBaseURL` usually takes longer (network probes), but the window exists and the failure is silent.

**Fix** - chain the calls so the second runs after the first completes:
```javascript
updateRumConfig(site, context)
  .catch((e) => log.warn(`[sites] RUM config update failed for ${site.getBaseURL()}: ${e.message}`))
  .then(() => autoDetectOverrideBaseURL(site, context))
  .catch((e) => log.warn(`[sites] base URL auto-detect failed for ${site.getBaseURL()}: ${e.message}`));
```
This preserves the fire-and-forget pattern (site creation still returns 201 immediately) while eliminating the concurrent-save race.

MysticatBot (changes requested):
Hey @tkotthakota-adobe,

**Verdict: Request changes** - concurrent fire-and-forget saves create a race condition that can silently drop config mutations.
_Changes: adds a reachability probe at site creation to auto-detect and persist a working www-toggled overrideBaseURL when the registered base URL is unreachable (3 files)._

## Must fix before merge
1. [Important] Race condition between concurrent fire-and-forget saves - `src/controllers/sites.js:406` (details inline)

<details>

- nit: `isUrlReachable` returns `true` for any HTTP response including 4xx/5xx - the intent is to detect "no listener", so this is arguably correct, but checking `response.ok` would avoid setting overrideBaseURL on domains that respond with parking pages or blanket 403s - `src/support/rum-config-service.js:82`
- suggestion: Extract protocol from original `baseURL` instead of hardcoding `https://` when constructing `toggledURL`, so the probe works correctly for sites registered with `http://` - `src/support/rum-config-service.js:118`
- suggestion: `autoDetectOverrideBaseURL` has no relationship to RUM configuration - consider moving to its own module (e.g., `src/support/base-url-detection.js`) for discoverability - `src/support/rum-config-service.js:92`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 24s | <b>Cost:</b> $3.88 | <b>Commit:</b> <code>fb8f043a3ee3af29f38c6d9982f43a21be1b28fa</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (commented):
Hey @tkotthakota-adobe,

**Verdict: Comment** - all prior findings addressed, no blocking issues remain.
_Changes: adds a reachability probe at site creation to auto-detect and persist a working www-toggled overrideBaseURL when the registered base URL is unreachable (4 files)._

<details>

- nit: Missing test for malformed baseURL path (the 1 line codecov flags) - a test passing e.g. `'not-a-url'` as baseURL would cover the `catch` at line 61 trivially - `src/support/base-url-detection.js:61`
- nit: Test "no-ops when baseURL returns non-ok response" asserts `calledTwice` but does not assert `site.save` was NOT called - worth adding for parity with the other no-op tests - `test/support/base-url-detection.test.js:223`
- suggestion: `toggledURL` is constructed as `${protocol}//${toggledHostname}`, dropping any path from the original baseURL - if a site is registered with a path (e.g. `https://www.example.com/site`), the probe tests bare domain instead - likely intentional given the domain-level nature of the check, but worth a one-line comment confirming the design choice - `src/support/base-url-detection.js:76`
</details>

- Race condition between concurrent fire-and-forget saves - now chained sequentially via `.then()`
- `isUrlReachable` accepted any HTTP response as reachable - now gates on `response.ok`
- Hardcoded `https://` when constructing toggled URL - now extracts protocol from original baseURL
- Auto-detect logic co-located in rum-config-service - extracted to dedicated `base-url-detection.js` module

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 53s | <b>Cost:</b> $2.63 | <b>Commit:</b> <code>6ee2c66b3ec0e7b6e7d3ba62a49fbc5b69a7cc4c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/base-url-detection.js
```diff
@@ -0,0 +1,90 @@
...
+    return;
+  }
+
+  const uri = new URI(baseURL);
+  const subdomain = uri.subdomain();
+  if (hasText(subdomain) && subdomain !== 'www') {
+    return;
+  }
+
+  if (await isUrlReachable(baseURL)) {
```
rpapani, line 70: @tkotthakota-adobe is [resolveCanonicalUrl](https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-utils/src/url-helpers.js#L161C16-L161C35) not sufficient, which is also used to set the fetchConfig? This check we do in onboarding and update the fetchConfig, is there reason to do this from site controller? how is this controller called, REST?

PR #2587, opened 2026-06-10
fix(llmo): retire categories.category_id business key; resolve prompt category filter by UUID (LLMO-5515)

igor-grubic:
## Problem

A brand-new category (e.g. a Japanese-named one) showed a **phantom "Prompts" count** equal to *every* prompt for the brand.

Root cause: the legacy `categories.category_id` TEXT business key (DB default `gen_random_uuid()::text`) was exposed as the category `id`. The prompt-list `categoryId` filter resolved that value against `categories.id` (the real UUIDv7 PK), found **no row**, and `listPrompts` then **silently dropped the filter (fail-open)** — returning the brand's entire prompt set.

`categories.category_id` was always redundant: `name` is already unique per org (`uq_category_name_per_org`) and `id` is the UUIDv7 PK. The duplicate business key was pure surface area for bugs.

## Change (app layer now, DB column drop deferred)

- **`mapDbCategoryToV2`** returns `id = uuid = categories.id` (the UUID PK). No more business key.
- **`createCategory`** no longer sets `category_id` (the DB default fills it); no slug derivation.
- **`resolveCategoryUuid`** is **UUID-only** — non-UUID input resolves to `null`.
- **`listPrompts`** **fails closed**: when a `categoryId`/`topicId` filter is requested but doesn't resolve, it returns an **empty page** instead of the full set.
- **Category CRUD** (`update`/`delete`) addresses rows by UUID `id`; controllers validate the path param is a UUID (`400` otherwise).
- OpenAPI specs (`categoryId` → `format: uuid`, fail-closed note, `V2Category.id` → uuid, removed client-supplied id), unit tests, and the PostgreSQL IT factory updated.

The physical column drop (`categories.category_id`, `uq_category_per_org`, `idx_categories_category_id`) is filed as a **follow-up** (screen for non-spacecat consumers first).

## Testing

- `npm run lint` ✓ · `npm test` (436 in touched suites, full build job green) ✓ · `npm run docs:lint` / `docs:build` ✓
- `it-postgres` ✓ (after fixing one pre-existing IT that soft-deleted a category by its legacy slug instead of UUID)

## Paired change

Requires the matching UI change in **project-elmo-ui** (consumes `categories.id` as the single identifier).

Relates to LLMO-5476 · Fixes LLMO-5515

### DIFF
--- a/docs/openapi/categories-v2-api.yaml
+++ b/docs/openapi/categories-v2-api.yaml
@@ -67,7 +67,7 @@ v2-categories-for-org:
         downgraded from `human` to `ai`), returns `200 OK` with the
         updated row.
       - **Soft-deleted row matched** — resurrected: status flipped to
-        `active`, stable `id` (slug) preserved, returns `201 Created`.
+        `active`, stable `id` (UUID) preserved, returns `201 Created`.
 
       Clients discriminating creation from idempotent update should inspect
       the HTTP status, not the response body (identical in both cases).
@@ -88,10 +88,9 @@ v2-categories-for-org:
     responses:
       '200':
         description: |
-          Existing category returned (idempotent update or no-op).
-          The `id` (slug) of the returned row is authoritative — it may
-          differ from a client-supplied `id` if the stored row was created
-          with a different slug.
+          Existing category returned (idempotent update or no-op). The `id`
+          (UUID) of the returned row is the server-assigned primary key and
+          is authoritative.
         content:
           application/json:
             schema:
@@ -118,9 +117,7 @@ v2-categories-for-org:
              deleted) between our lookup and our update — message
              `Category was concurrently modified; please retry`.
           2. A unique-constraint other than `uq_category_name_per_org` was
-             tripped on insert (e.g. `uq_category_id_per_org` when the
-             client-supplied slug collides with an unrelated existing
-             row). Message echoes the actual constraint name.
+             tripped on insert. Message echoes the actual constraint name.
           Treated as an idempotent retry signal by DRS-class clients;
           logged at WARN (not ERROR) server-side.
       '500':
@@ -141,10 +138,11 @@ v2-category-for-org:
     - name: categoryId
       in: path
       required: true
-      description: Category business key (category_id slug)
+      description: Category UUID (categories.id)
       schema:
         type: string
-        example: 'baseurl-discovery-research'
+        format: uuid
+        example: 'a1b2c3d4-5678-90ab-cdef-1234567890ab'
   patch:
     tags:
       - customer-config
--- a/docs/openapi/prompts-v2-api.yaml
+++ b/docs/openapi/prompts-v2-api.yaml
@@ -37,7 +37,11 @@ v2-prompts-by-brand:
       required: false
       schema:
         type: string
-      description: Filter by category business key
+        format: uuid
+      description: |
+        Filter by category UUID (categories.id). A categoryId that does not
+        resolve to a category in this organization returns an empty page
+        (the filter fails closed; it is never silently dropped).
     - name: topicId
       in: query
       required: false
--- a/docs/openapi/schemas.yaml
+++ b/docs/openapi/schemas.yaml
@@ -6246,7 +6246,8 @@ V2Prompt:
       description: Regions where this prompt is applicable
     categoryId:
       type: string
-      description: Category business key
+      format: uuid
+      description: Category UUID (categories.id)
     topicId:
       type: string
       description: Topic business key
@@ -6293,14 +6294,17 @@ V2Prompt:
       properties:
         id:
           type: string
+          format: uuid
           description: |
-            Currently the UUID PK (categories.id) for backward compat —
-            aligning this to the category_id business key is a deferred
-            breaking change. Use `uuid` for FK linkage today.
+            UUID primary key (categories.id). The sole category identifier
+            (the legacy category_id business key was retired — LLMO-5515).
+            `id` and `uuid` carry the same value; either works for FK linkage.
         uuid:
           type: string
           format: uuid
-          description: UUID primary key (categories.id)
+          description: |
+            UUID primary key (categories.id). Alias of `id`, kept for
+            backward compatibility.
         name:
           type: string
         origin:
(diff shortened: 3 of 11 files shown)


MysticatBot (approved):
Hey @igor-grubic,

**Verdict: Approve** - Clean security fix, well-tested, properly scoped.
_Changes: Retires the `categories.category_id` TEXT business key in favor of the UUID primary key and makes the prompt-list category filter fail-closed to prevent data leakage (12 files)._

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 35s | <b>Cost:</b> $9.22 | <b>Commit:</b> <code>a624c7e6055ff0186a63ba2ec126a65c08c3055c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jjenscodee:
## Code Review

### What it does
Fixes a real production bug (LLMO-5515): brand-new categories showed a phantom prompt count equal to *all* prompts for the brand. Root cause: `mapDbCategoryToV2` was returning `category_id` (a TEXT business key) as `id`. When a UUID-shaped business key (e.g. generated for a Japanese-named category) failed to resolve against `categories.id`, the `listPrompts` filter silently dropped the constraint and returned every prompt in the brand.

The fix is correct and complete: expose the UUID PK as the sole `id`, make both category and topic filters **fail closed** (return empty page instead of unfiltered set), and add UUID validation at the controller boundary.

---

### Strengths
- **Fail-closed is the right call.** The new guard in `listPrompts` — resolve first, return `emptyPage` on null — directly eliminates the bug class. The analogous guard on `topicId` is a good bonus.
- **Controller-level UUID validation** (`isValidUUID` → 400) is good defense-in-depth; clients learn immediately when they're sending the wrong thing rather than silently getting an empty list.
- **`createCategory` simplification** is clean — removing slug derivation eliminates the root source of the polymorphic `category_id` problem, not just the symptom.
- **IT test rewrite** is excellent. The new `categories-prompts.js` suite directly exercises the phantom-count scenario end-to-end and has a clear regression guard.
- **Backward compat**: keeping `uuid` as an alias of `id` in the API response is sensible while other consumers catch up.

---

### Issues / questions

**1. `uuid` alias vs. UI removing the field (coordination risk)**

`mapDbCategoryToV2` still emits both `id` and `uuid` (both now equal to `row.id`). The UI PR removes `uuid` from its types entirely. Before merging both, confirm there are no other consumers (DRS, brand-presence client, any Lambda that reads `categories[n].uuid` directly) that need a deprecation window. The alias is harmless to keep but the question is whether anything still relies on it that isn't part of this paired change.

**2. `ensureLookupEntries` upsert conflict key change**

```diff
- { onConflict: 'organization_id,category_id' }
+ { onConflict: 'organization_id,name' }
```

This is correct — `uq_category_name_per_org` is the real constraint — but it's a behavior change for bulk prompt imports (e.g. DRS posting the same category name twice in one batch). Previously, two rows with the same `category_id` deduped; now they dedup on `name`. For correctly-formed data this is identical, but worth flagging if DRS ever sends rows where name doesn't normalize identically across calls.

**3. Deferred DB column drop is technical debt to track**

`categories.category_id`, `uq_category_per_org`, and `idx_categories_category_id` are all still in the DB with the app layer now simply ignoring them (the column defaults to `gen_random_uuid()::text`). The PR correctly defers this, but there's a subtle risk: `uq_category_per_org` on the `category_id` column is still enforced by the DB. A UUID collision on the default would be astronomically unlikely, but it would surface as an unexpected 409 hitting the fallback branch in `createCategory`. Not a blocker, just make sure the follow-up ticket is tracked.

**4. Missing files** (12 changed files reported, 9 visible including docs)

`gh` reports 12 changed files; 9 are visible (8 source + generated docs). The OpenAPI spec changes mentioned in the PR description may be in JSDoc on route handlers or separate YAML files. If there are spec files, confirm that the `V2Category.id` format is updated to `uuid` and the fail-closed behavior is documented.

---

### Minor
- Log field rename `category_id` → `category_uuid` in structured log events is correct and will help Coralogix queries.
- Removed `createCaptureInsertClient` helper in tests is clean — it was only needed to assert on the written `category_id` which no longer exists.

---

### Verdict: **Approve** ✅

Pre-merge checklist:
- [ ] Confirm no non-spacecat consumers read `categories[n].uuid` directly
- [ ] Confirm API deploys before/with UI (the UI's `getCategoryFilterId` change requires the API to return UUID `id`)
- [ ] File/assign the follow-up to drop `categories.category_id` column and its constraints

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

jjenscodee (approved):

igor-grubic:
Thanks for the thorough review @jjenscodee — I went through all four points against the diff and the DRS codebase (`llmo-data-retrieval-service`). Summary: **#2–#4 are non-issues / already handled; #1 is a real DRS dependency but we're choosing not to address it (rationale below).**

### #1 — `uuid` alias & DRS dependency — **won't fix (accepted)**

You're right that DRS reads category responses and assumes `id` (slug) ≠ `uuid` (PK):

- `get_org_categories_map()` / `get_org_categories_name_map()` build slug-keyed maps (`{cat["id"] → cat["uuid"]}`). After this change both are the same UUID, so the slug-keyed lookup in the v1 scheduler backfill (`_backfill_prompt_entity_fields`) always misses.

**Why we're not fixing it:**

- **Degrades gracefully, doesn't break.** The backfill resolution order is slug→UUID first, then **name→UUID fallback** (`get_org_categories_by_name_map`, which is name-keyed and unaffected by this change). When the slug path misses, resolution falls through to the name map. Only a prompt that carries a slug but *no* name goes unresolved (blank category in the xlsx).
- **Scoped to v1 orgs only.** That backfill path exists solely for orgs "not yet fully migrated to the brandalf entity model." We've moved most customers to v2, so the residual v1 compatibility surface is small and shrinking — not worth a coordinated cross-service deploy for a graceful fallback.
- The brand-presence FK path (`map_v2_prompts_to_standard` → `category.uuid`) is **unaffected** — `uuid` still returns the UUID PK, which is exactly the FK value it consumes.

We'll let the slug-keyed maps relabel/retire naturally on the DRS side when it next touches that path.

### #2 — `onConflict: category_id` → `name` — **correct, required**

Targets the real `uq_category_name_per_org` constraint. DRS posts categories/prompts by **name** (`ensure_categories_v2` dedups on `(name, origin)`), and the upsert dedups missing names within a single bulk request. The old `category_id` key (now a random DB default) would never have deduped same-named categories anyway, so this is strictly more correct.

### #3 — `uq_category_per_org` collision on the random default — **non-issue, tracked**

`category_id` defaults to `gen_random_uuid()::text`; a v4 collision is astronomically improbable, and `createCategory`'s `23505` fallback already converts any non-name-constraint violation into a clean typed 409 (caller retries → new default). The column + constraint drop is tracked as a follow-up.

### #4 — OpenAPI spec / "missing files" — **done**

Verified in the diff: `V2Category.id` is now `format: uuid` with the LLMO-5515 retirement note; the `categoryId` path/query params are `format: uuid`; fail-closed behavior is documented. The "12 vs 9 files" delta was just generated docs.

PR #2590, opened 2026-06-11
fix: enforce site-scope validation for staging domains on subpath prod sites

dhavkuma:
## Summary

When a production site is registered under a subpath (e.g. `example.com/docs`), staging domains must fall within that same path scope. Previously, both the LLMO API and the Slack `add-oae-stage-domain` command accepted staging domains with no matching path component (e.g. `staging.example.com`), causing the stage edge config to intercept at the host level rather than scoping to `/docs`.

- **`llmo.js`**: Added a pathname-scope guard in `createOrUpdateStageEdgeConfig` — rejects with `400 Bad Request` if any staging domain does not fall within the production site's path scope
- **`add-oae-stage-domain.js`**: Added the same per-domain guard in the Slack command — reports an error for each domain that violates the scope before writing the Tokowaka config
- **`url-utils.js`**: Path-scope utilities (`hasSamePathname` / `allHaveSamePathname`) extracted here, pending promotion to `@adobe/spacecat-shared-utils`
- **Tests**: Updated to cover rejection when a staging domain lacks the prod site's subpath, and acceptance when the staging URL is within the prod path scope

## Test plan

- [ ] Staging domain with no path rejected for a subpath prod site (HTTP 400 from API, error message in Slack)
- [ ] Staging domain within the prod path scope accepted
- [ ] Root-path prod sites (`example.com`) unaffected — no path restriction applied
- [ ] Existing base-domain validation still enforced before path check


MysticatBot (dismissed):
Hey @dhavkuma,

**Verdict: Request changes** - one robustness gap in the new `getPathname` helper needs addressing.
_Changes: Adds subpath matching to staging domain validation and a CDN routing guard that rejects subpath sites from host-level auto-routing (2 files)._

## Must fix before merge

1. [Important] `getPathname` returns `'/'` on unparseable input, making parse failure indistinguishable from a legitimate root path - `src/controllers/llmo/llmo.js:1865` (details inline)

<details>

- nit: Add a test exercising a malformed `baseURL` reaching the CDN guard catch block - fixes the codecov/patch failure and documents the intended fall-through behavior - `src/controllers/llmo/llmo.js:1495`
- nit: The trailing-slash test asserts `status !== 400` but the guard returns 200 when it rejects - assert the response body does NOT include "subpath sites" instead, which directly proves the guard did not fire - `test/controllers/llmo/llmo.test.js:6281`
- suggestion: The CDN guard duplicates the "prepend https:// if missing then new URL()" pattern that `getPathname` already encapsulates - consider calling `getPathname(baseURL) !== '/'` instead of re-implementing inline - `src/controllers/llmo/llmo.js:1486`

</details>

src/controllers/llmo/llmo.js
```diff
@@ -1838,9 +1855,24 @@ function LlmoController(ctx) {
...
    * @returns {boolean} true if all URLs share the same domain as prodBaseURL
    */
+  function getPathname(urlString) {
+    try {
+      return new URL(urlString).pathname.replace(/\/$/, '') || '/';
+    } catch {
+      try {
+        return new URL(`https://${urlString}`).pathname.replace(/\/$/, '') || '/';
+      } catch {
+        return '/';
```
MysticatBot: issue (blocking): `getPathname` returns `"/"` on unparseable input, making the return value ambiguous.

**What:** The inner `catch { return '/'; }` means any garbage string (empty, whitespace, protocol-only like `"https://"`) normalizes to `"/"`. When both prod and stage URLs are malformed, `getPathname` returns `"/"` for both and the pathname comparison passes vacuously. The only remaining guard is `getDomain`, which may or may not reject the same garbage.

**Why it matters:** The function has an unclear contract - callers cannot distinguish "this is a root path" from "I could not parse this at all." If `getDomain` ever gains a similarly lenient fallback, validation would pass for two broken URLs.

**Fix:** Return a sentinel (e.g. `null`) from the final catch and treat it as a non-match in `areDomainsSameAsBase`:

```js
function getPathname(urlString) {
  try {
    return new URL(urlString).pathname.replace(/\/$/, '') || '/';
  } catch {
    try {
      return new URL(`https://${urlString}`).pathname.replace(/\/$/, '') || '/';
    } catch {
      return null;
    }
  }
}

// In areDomainsSameAsBase:
return urlList.every(
  (stageURL) => {
    const stagePath = getPathname(stageURL);
    return stagePath !== null && getDomain(stageURL) === prodDomain && stagePath === prodPath;
  },
);
```

This also makes the CDN guard clearer if you reuse it there.

dhavkuma: Fixed in b145ff1d — `getPathname` now returns `null` in the final catch. `areDomainsSameAsBase` guards against `null` before comparing paths, so two malformed URLs can no longer match vacuously. The new malformed-URL test covers the CDN guard catch block (fall-through behavior) and the trailing-slash test now asserts the response body excludes "subpath sites" rather than checking the status code.

src/controllers/llmo/llmo.js
```diff
@@ -1855,9 +1855,27 @@ function LlmoController(ctx) {
    * @param {string} prodBaseURL the production base URL to match against.
    * @returns {boolean} true if all URLs share the same domain as prodBaseURL
    */
+  function getPathname(urlString) {
```
dhavkuma: try to look for reusable function

src/controllers/llmo/llmo.js
```diff
@@ -1855,9 +1855,27 @@ function LlmoController(ctx) {
...
+  }
+
   function areDomainsSameAsBase(urlList, prodBaseURL) {
     const prodDomain = getDomain(prodBaseURL);
-    return urlList.every((stageBaseURL) => getDomain(stageBaseURL) === prodDomain);
+    const prodPath = getPathname(prodBaseURL);
+    return urlList.every(
+      (stageURL) => {
+        const stagePath = getPathname(stageURL);
+        return stagePath !== null && getDomain(stageURL) === prodDomain && stagePath === prodPath;
```
dhavkuma: consider the cases
1. override urls

dhavkuma: Also onboard stagin domain command should not enforce this check yet

dhavkuma:
@MysticatBot review

src/controllers/llmo/llmo.js
dhavkuma: > consider the cases
> 
> 1. override urls

staging domains checks on baseUrl - so no need to check on overrideBaseUrl

dhavkuma: > Also onboard stagin domain command should not enforce this check yet

This no longer is applicable.
onboarding staging should relax on domain name
but should be restrictive as per pathname scoping of the prod site

dhavkuma:
@MysticatBot review

src/utils/url-utils.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+}
+
+// TODO: move to spacecat-shared-util
+export function hasSamePathname(url, referenceUrl) {
+  return toPathname(url) === toPathname(referenceUrl);
+}
+
+// TODO: move to spacecat-shared-util
+export function allHaveSamePathname(urls, referenceUrl) {
```
jindaliiita: allMatchPathname

jindaliiita: Before merging this PR, will you move these changes to `spacecat-shared-utils`?

src/utils/url-utils.js
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+/**
+ * Extracts the pathname from a URL string, stripping trailing slashes on non-root paths.
+ * Falls back to the raw string when the URL is not parseable (e.g. invalid or relative).
+ *
+ * @param {string} url
+ * @returns {string} pathname, or the original string on parse failure
+ */
+export function toPathname(url) {
```
jindaliiita: This is already present in `shared-utils`, and I don't think you need lowercase here
https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-utils/src/url-helpers.js#L254

jindaliiita:
Approving these changes, but I have added a couple of comments for you to review.

jindaliiita (approved):

src/utils/url-utils.js
dhavkuma: Done. Will refactor PR now

src/utils/url-utils.js
dhavkuma: will add the refactor on spacecat shared function

src/utils/url-utils.js
dhavkuma: done. Now importing from spacecat

src/utils/url-utils.js
dhavkuma: https://github.com/adobe/spacecat-shared/pull/1729

MysticatBot (approved):
Hey @dhavkuma,

**Verdict: Approve** - clean implementation using shared-utils, closes the subpath scope gap correctly.
**Complexity: MEDIUM** - small diff; dependency bump signal.
_Changes: Adds pathname-scope validation for staging domains on subpath production sites, using shared-utils functions instead of local implementation (6 files)._

<details>

- nit: Error messages say "within the site pathname scope" but the underlying `allHaveSamePathname`/`hasSamePathname` performs exact pathname equality, not prefix/containment matching. If `/docs/v2` against prod `/docs` should be rejected (which it will be), "must match the pathname" is clearer than "within ... scope" - `src/controllers/llmo/llmo.js:2043`
- nit: Rejection test uses bare domain (`staging.lovesac.com`) while acceptance test uses full URL (`https://staging.lovesac.com/docs`) - the pair works correctly but the asymmetric format makes it less obvious that the rejection is due to pathname mismatch rather than parse failure - `test/controllers/llmo/llmo.test.js:7282`
- suggestion: Add a multi-domain test where one domain fails the pathname check (e.g. `['https://staging.lovesac.com/docs', 'staging2.lovesac.com']`) to confirm and document the all-or-nothing rejection behavior of `allHaveSamePathname` - `test/controllers/llmo/llmo.test.js:7276`

</details>

- `getPathname` ambiguity on unparseable input - entire local implementation replaced with shared-utils imports

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 48s | <b>Cost:</b> $4.57 | <b>Commit:</b> <code>0255830f54d0716d36fd9f6ceb920d24b98f2b01</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2593, opened 2026-06-11
fix(brands): make baseSiteId immutable in upsertBrand (LLMO-5556)

byteclimber:
## Problem (LLMO-5556)

`upsertBrand` unconditionally wrote `site_id` from `baseSiteId` and upserted on `(organization_id, name)`. Re-onboarding a **second site under an existing brand name** resolved onto the existing row and **silently re-pointed the brand's primary site** — and, because `upsertBrand` calls the full-replace `syncBrandUrls`/`syncBrandSites`/alias syncs when those fields are passed, could also drop the brand's existing URLs/aliases.

Confirmed in production on two customers (both remediated manually):
- **MongoDB** — onboarding `learn.mongodb.com` overwrote primary `mongodb.com`.
- **Merck/MSD** — "Merck" brand primary became `keytruda.com` instead of `merck.com`.

`updateBrand` already had the immutability guard; `upsertBrand` did not.

## Fix

**Layer 1 — `src/support/brands-storage.js`:** guard `site_id` in `upsertBrand` — set it only when the brand has no persisted `site_id` (mirrors `updateBrand`); `log.warn` and ignore on an attempted change. Adds an optional `log` param, threaded from the controller.

**Layer 2 — `src/controllers/llmo/llmo-onboarding.js`:** before the initial-brand write, look up the existing brand by `(org, name)`; if it already points at a **different** site, skip the write and warn (surfacing the collision instead of overwriting / clobbering).

## Tests
- `test/support/brands-storage.test.js` — capturing-mock unit tests asserting the upsert payload **preserves** `site_id` on the update path, **sets** it on insert, **warns** on an attempted change, and does not warn for a same-site re-upsert.
- `test/controllers/llmo/llmo-onboarding.test.js` — onboarding collision test: existing brand on a different site ⇒ `upsertBrand` not called, warn logged, onboarding still completes.
- Full suite + 100% coverage gate pass; lint clean.

## Out of scope (follow-up)
Admin-gated "reassign primary site" path so genuine corrections don't require a direct DB write (the gap that blocked the original Support request). Tracked separately.

### DIFF
--- a/src/controllers/llmo/llmo-onboarding.js
+++ b/src/controllers/llmo/llmo-onboarding.js
@@ -1381,14 +1381,21 @@ export async function performLlmoOnboarding(params, context, say = () => {}) {
       // initial-brand write — the brand already exists so DRS sync still resolves
       // it; a human then decides whether the new site is a sub-brand or a URL.
       try {
-        const { data: existingBrand } = await postgrestClient
+        const { data: existingBrand, error: lookupError } = await postgrestClient
           .from('brands')
           .select('id, site_id')
           .eq('organization_id', organization.getId())
           .eq('name', brandName.trim())
           .maybeSingle();
 
-        if (existingBrand?.site_id && existingBrand.site_id !== site.getId()) {
+        if (lookupError) {
+          // Fail closed (LLMO-5556): PostgREST returns { data: null, error } on a
+          // query failure rather than throwing, so without this check a transient
+          // failure would fall through to upsertBrand and could re-point an
+          // existing brand's primary site. Skip the write as a precaution.
+          log.warn(`Skipping initial brand write: failed to look up existing brand "${brandName.trim()}" `
+            + `(org ${organization.getId()}): ${lookupError.message}`);
+        } else if (existingBrand?.site_id && existingBrand.site_id !== site.getId()) {
           log.warn(`Skipping initial brand write: brand "${brandName.trim()}" `
             + `(org ${organization.getId()}) already exists with a different primary site `
             + `(existing=${existingBrand.site_id}, onboarding=${site.getId()}). `
--- a/src/controllers/llmo/llmo-onboarding.js
+++ b/src/controllers/llmo/llmo-onboarding.js
@@ -1401,6 +1401,9 @@ export async function performLlmoOnboarding(params, context, say = () => {}) {
             + `(existing=${existingBrand.site_id}, onboarding=${site.getId()}). `
             + `Add ${baseURL} as a brand URL or onboard under a distinct brand name.`);
         } else {
+          // No collision: proceed with upsert (new brand, existing brand with a
+          // null primary site, or a re-onboard of the same site). upsertBrand's
+          // own guard keeps an already-set site_id immutable on the last case.
           await upsertBrand({
             organizationId: organization.getId(),
             brand: {
--- a/src/controllers/llmo/llmo-onboarding.js
+++ b/src/controllers/llmo/llmo-onboarding.js
@@ -1375,20 +1375,40 @@ export async function performLlmoOnboarding(params, context, say = () => {}) {
       // Use baseURL (matches sites.base_url) so syncBrandSites can link the brand
       // to the site. overrideBaseURL may differ (e.g., www.blick.ch vs blick.ch)
       // and would fail the exact-match lookup against the sites table.
+      // LLMO-5556: a second site onboarded under an existing brand name must not
+      // re-point that brand's primary site, nor clobber its URLs/aliases via the
+      // full-replace syncs inside upsertBrand. Detect the collision and skip the
+      // initial-brand write — the brand already exists so DRS sync still resolves
+      // it; a human then decides whether the new site is a sub-brand or a URL.
       try {
-        await upsertBrand({
-          organizationId: organization.getId(),
-          brand: {
-            name: brandName.trim(),
-            status: 'active',
-            baseSiteId: site.getId(),
-            urls: [{ value: baseURL, type: 'base' }],
-            brandAliases: [{ name: brandName.trim(), regions: ['gl'] }],
-          },
-          postgrestClient,
-          updatedBy: 'llmo-onboarding',
-        });
-        log.info(`Created initial brand "${brandName}" in normalized table for site ${site.getId()}`);
+        const { data: existingBrand } = await postgrestClient
+          .from('brands')
+          .select('id, site_id')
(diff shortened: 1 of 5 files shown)


MysticatBot (changes requested):
Hey @byteclimber,

**Verdict: Request changes** - one gap in the error handling of the new collision-check query.
_Changes: Adds baseSiteId immutability guard to `upsertBrand` and a pre-check in the onboarding controller to prevent re-pointing a brand's primary site on re-onboarding (5 files)._

## Must fix before merge
1. [Important] Unchecked PostgREST error on brand lookup silently bypasses the immutability guard under transient failures - `src/controllers/llmo/llmo-onboarding.js:1380` (details inline)

<details>

- suggestion: Add an onboarding-level test for the "existing brand with site_id: null" path (brand exists but was created without a primary site, so upsert should proceed and set it) - `test/controllers/llmo/llmo-onboarding.test.js`
</details>

src/controllers/llmo/llmo-onboarding.js
```diff
@@ -1375,20 +1375,40 @@ export async function performLlmoOnboarding(params, context, say = () => {}) {
...
-            status: 'active',
-            baseSiteId: site.getId(),
-            urls: [{ value: baseURL, type: 'base' }],
-            brandAliases: [{ name: brandName.trim(), regions: ['gl'] }],
-          },
-          postgrestClient,
-          updatedBy: 'llmo-onboarding',
-        });
-        log.info(`Created initial brand "${brandName}" in normalized table for site ${site.getId()}`);
+        const { data: existingBrand } = await postgrestClient
```
MysticatBot: issue (blocking): The destructuring `{ data: existingBrand }` discards the `error` field. PostgREST returns `{ data: null, error: {...} }` on query failures rather than throwing, so a transient DB error makes `existingBrand` null and the code falls through to the `else` branch - calling `upsertBrand` as if no brand exists. The outer `catch (brandError)` will not fire because PostgREST does not throw on query errors.

The storage-layer guard has the same vulnerability: if its own internal lookup also returns `{ data: null, error }`, then `existing?.site_id` is falsy and the immutability condition `!hasText(existing?.site_id)` evaluates to true, allowing the overwrite.

Under a partial PostgREST failure, both defense layers are bypassed silently - the exact scenario the PR is designed to prevent.

**Fix:** Destructure and check `error` before using `data`. Fail closed (skip the write) rather than open. If `lookupError` is truthy, log a warning and skip the brand write as a precaution.

byteclimber:
Thanks @MysticatBot — good catch on the PostgREST error handling. Addressed in 8ceeadb7.

**Must-fix (blocking) — fixed.** Both layers now check the lookup `error` and **fail closed** instead of falling through:
- `upsertBrand` (`brands-storage.js`): destructures `error` from the existing-brand lookup and **throws** when it's set, so a transient PostgREST failure surfaces (and SQS retries) rather than being treated as "no brand exists" and overwriting the primary.
- `llmo-onboarding.js`: the collision pre-check now branches on `lookupError` and **skips** the initial-brand write with a warning, instead of proceeding to `upsertBrand`.

**Non-blocking suggestion — added.** New onboarding test for the "existing same-name brand with `site_id: null`" path (write proceeds and the upsert sets the primary).

New tests: storage throw-on-lookup-error; onboarding fail-closed-on-lookup-error; onboarding site_id-null-proceeds. Full suite green (11,746 passing) with the 100% coverage gate held.

MysticatBot (commented):
Hey @byteclimber,

**Verdict: Comment** - clean fix, both prior findings addressed.
_Changes: Adds baseSiteId immutability guard to `upsertBrand` and a pre-check in the onboarding controller to prevent re-pointing a brand's primary site on re-onboarding (5 files)._

<details>

- suggestion: The `else` branch at `src/controllers/llmo/llmo-onboarding.js:1399` folds three distinct semantic cases (new brand, null site_id, same-site re-onboard) into one path. A short inline comment like `// No collision: proceed with upsert (new brand, null site_id, or same site)` would help a future reader parse the branching without inverting the conditions mentally.
- suggestion: Add a controller-level test where `existingBrand.site_id === site.getId()` (same-site re-onboard) and assert `upsertBrand` is called without a warning. The storage layer covers this path, but the controller decision point is unexercised - `test/controllers/llmo/llmo-onboarding.test.js`.
</details>

- PostgREST error on brand lookup now checked and fails closed in both layers.
- Onboarding test for "existing brand with site_id: null" path added.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 35s | <b>Cost:</b> $2.50 | <b>Commit:</b> <code>8ceeadb72e7696489d4313de7b0a9350ec54ffa5</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

byteclimber:
Thanks @MysticatBot — both non-blocking suggestions addressed in 0b5b6b87:

- Added an inline comment on the no-collision `else` branch noting it covers the three cases (new brand / existing brand with null `site_id` / same-site re-onboard), with the reminder that `upsertBrand`'s own guard keeps an already-set `site_id` immutable on the last case.
- Added a controller-level test for the same-site re-onboard path (`existingBrand.site_id === site.getId()`): asserts `upsertBrand` is called and no collision warning is logged.

Affected-file tests green (`llmo-onboarding` 119, `brands-storage` 92). The one failure in the full run was an unrelated flaky 10s timeout in `plg-onboarding.test.js` (passes in isolation at ~8.2s) — not touched by this PR.

MysticatBot (approved):
Hey @byteclimber,

**Verdict: Approve** - clean fix, all prior findings addressed, both non-blocking suggestions landed.
_Changes: Adds baseSiteId immutability guard to `upsertBrand` and a pre-check in the onboarding controller to prevent re-pointing a brand's primary site on re-onboarding (5 files)._

No blocking or non-blocking findings on this revision.

- Inline comment on the no-collision `else` branch added (three-case summary with upsertBrand guard reminder).
- Controller-level test for same-site re-onboard path added.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 28s | <b>Cost:</b> $3.44 | <b>Commit:</b> <code>0b5b6b878b0d7eaf429dee62aebfe15763066e4f</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2594, opened 2026-06-11
docs: update S2S consumer guide example to a non-paginated endpoint | SITES-41367

martinst06:
## What
Step 3 of the S2S Consumer Integration Guide showed a `getSites` example that fetched all sites in one call. `GET /sites` is now cursor-paginated, so that example no longer matches the real response shape and is hitting the legacy.

## Change
Swapped it for a single-site fetch (`GET /sites/{siteId}`) - a clean tenant-scoped read that pairs naturally with the opportunities example below it.

Docs-only change

## Issues
SITES-41367


ravverma (approved):

martinst06:
SITES-41367

PR #2602, opened 2026-06-12
docs: add money-pages to OpenAPI AuditType enum and audit examples

nitinja:
## What
Document `money-pages` as a valid audit type in the OpenAPI spec.

- **`schemas.yaml`** — add `money-pages` to the `AuditType` enum, and a money-pages variant to the `AuditResult` `oneOf` (`{ siteUrl, fullAuditRef }`, matching the audit-worker handler output).
- **`examples.yaml`** — add a `money-pages-audit` example (mirrors `broken-internal-links-audit`).
- **`audit-api.yaml`** — reference the new example in both audit response blocks (`audit-for-site`, `latest-audit-for-site`).
- **`docs/index.html`** — regenerated via `npm run docs:build` (repo convention for spec changes).

## Why
`money-pages` is a real, queryable audit type (e.g. `GET /sites/{siteId}/audits/money-pages`) but was missing from the `AuditType` enum. That enum is `$ref`'d by the `:auditType` path/query parameters (`parameters.yaml`), so the documented contract omitted a valid type. Pure doc/contract-accuracy fix — no runtime behavior change.

## Scope notes
- **Not** added to the `OpportunityType` enum — money-pages produces no opportunities (the audit only emits an SQS message to Mystique); listing it there would document a type that never exists.
- The vendored OpenAPI snapshot in `llmo-data-retrieval-service` is intentionally untouched (reference-only; nothing consumes/enforces it).

## Testing
- `npm run docs:lint` → exits 0 (`API description is valid 🎉`). The one `no-invalid-media-type-examples` warning on the new example is structural to `AuditResult`'s `oneOf` (its first variant is `additionalProperties: true`, so every per-type example matches 2 variants) — this same warning already occurs 46× for the existing examples (broken-backlinks, costs, meta-tags, …), so the addition is consistent.
- `npm run docs:build` → bundles successfully.

### DIFF
--- a/docs/openapi/audit-api.yaml
+++ b/docs/openapi/audit-api.yaml
@@ -38,6 +38,8 @@ audit-for-site:
                 $ref: './examples.yaml#/related-urls-audit'
               broken-internal-links-audit:
                 $ref: './examples.yaml#/broken-internal-links-audit'
+              money-pages-audit:
+                $ref: './examples.yaml#/money-pages-audit'
       '400':
         $ref: './responses.yaml#/400-audit-for-site'
       '401':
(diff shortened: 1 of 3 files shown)


tkotthakota-adobe (approved):
LGTM

PR #2605, opened 2026-06-14
fix: edge deploy api - support domain wide deployment

dipratap:
Since the current 10 limit is too low to gauge impact from domain wide deployment, we are extending it to 100 urls.
Also, updated the filter to identify high impact urls for domain wide deployment - it is in sync with the UI High Impact URLs Filter. 100 is a safe limit since DRS supports it.


src/controllers/suggestions.js
```diff
@@ -1907,24 +1907,31 @@ function SuggestionsController(ctx, sqs, env) {
...
+            .slice(0, 100);
         } else {
           promptSources = validSuggestions;
         }
-        const domainWideSuggestionIds = new Set(
-          domainWideSuggestions.map(({ suggestion }) => suggestion.getId()),
-        );
         urls = promptSources
           .filter((s) => !domainWideSuggestionIds.has(s.getId()))
           .map((s) => s.getData()?.url)
```
MysticatBot, line 1937: issue (blocking): The old code capped at 10 prompt sources (top 15 by contentGainRatio, then top 10 by agenticTraffic). The new code takes the top 100. These sources feed into an S3 upload (`-prompts.json`) consumed by the edge layer.

A 10x increase in payload size changes the operational characteristics of the downstream consumer (storage, parse time, memory at the edge). The PR title says "support domain wide deployment" but does not explain why 100 is the right number.

**Suggestion:** Either document in the PR body why 100 is safe for all downstream consumers, or extract the limit to a named constant with a rationale comment. If a downstream constraint exists (LLM context window, edge worker memory), validate against it.

src/controllers/suggestions.js
```diff
@@ -1907,24 +1907,31 @@ function SuggestionsController(ctx, sqs, env) {
...
-          promptSources = top15ByContentGainRatio
-            .sort((a, b) => (b.getData()?.agenticTraffic || 0) - (a.getData()?.agenticTraffic || 0))
-            .slice(0, 10);
+          promptSources = [...allSuggestions]
+            .filter((s) => {
+              const data = s.getData() || {};
+              return !domainWideSuggestionIds.has(s.getId())
+                && s.getStatus() === SuggestionModel.STATUSES.NEW
+                && !data.edgeDeployed
+                && data.aiSummary
```
MysticatBot, line 1921: issue (blocking): The new filter adds four conditions (`!domainWideSuggestionIds.has`, `s.getStatus() === NEW`, `!data.edgeDeployed`, `data.aiSummary`, `data.valuable === true`), but the test only exercises the happy path where items pass the filter and verifies sort order. Codecov confirms 5 lines of missing coverage here.

There is no test covering:
- A suggestion with `edgeDeployed: true` being excluded
- A suggestion without `aiSummary` being excluded
- A suggestion with `valuable !== true` being excluded

These exclusion paths are the core behavioral change. Add a test that includes suggestions which should be filtered OUT and assert they do not appear in the uploaded prompts.

MysticatBot (changes requested):
Hey @dipratap,

**Verdict: Request changes** - the core logic and sorting are correct, but the 10x limit increase needs documented rationale and the new filter conditions lack test coverage.
_Changes: Refactors domain-wide edge deploy prompt-source selection to use a composite score (agenticTraffic x contentGainRatio) with stricter filtering criteria (2 files)._

## Must fix before merge
1. [Important] 10x increase in prompt source limit (10 to 100) without documented downstream safety - `src/controllers/suggestions.js:1936` (details inline)
2. [Important] New filter conditions (edgeDeployed, aiSummary, valuable) lack exclusion-path test coverage - `src/controllers/suggestions.js:1916-1921` (details inline)

<details>

- nit: Redundant `domainWideSuggestionIds` filter - promptSources already excludes domain-wide IDs at line 1916, but the `urls =` assignment re-filters the same set - `src/controllers/suggestions.js:1937`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 44s | <b>Cost:</b> $4.60 | <b>Commit:</b> <code>5ad0a41d98aeabe730a79f537f6174c2cda12890</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @dipratap,

**Verdict: Approve** - filter logic and tests are solid; two minor nits below.
_Changes: Refactors domain-wide edge deploy prompt-source selection to use a composite score (agenticTraffic x contentGainRatio) with stricter filtering criteria and a 100-item limit (2 files)._

<details>

- nit: Redundant `domainWideSuggestionIds` filter - `promptSources` already excludes domain-wide IDs at line 1916, but the `urls =` assignment re-filters the same set. Dead code that may confuse future readers. - `src/controllers/suggestions.js:1937`
- suggestion: Sort test should assert `expect(uploadedPrompts).to.have.lengthOf(2)` before the positional checks to explicitly verify the exclusion of `approvedSugg` (currently the exclusion test covers this, but the sort test's assertions would pass even if an extra item appeared after position 1). - `test/controllers/suggestions.test.js`
</details>

- Filter exclusion path test coverage added (edgeDeployed, aiSummary, valuable conditions now exercised)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 47s | <b>Cost:</b> $2.59 | <b>Commit:</b> <code>ee5995e3e3c987744737dd347c49f91d8631ff4b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2606, opened 2026-06-15
fix: consume launchdarkly-client log-noise fix and dedupe nested copy

solaris007:
## What

Bump two shared deps (exact pins, no caret):
- `@adobe/spacecat-shared-http-utils` `1.27.2` -> `1.29.1`
- `@adobe/spacecat-shared-launchdarkly-client` `^1.1.0` -> `1.3.0`

## Why

This is the consuming end of a 3-phase fix that eliminates LaunchDarkly SDK log noise, which was **~71% of SpaceCat prod `error`-level logs** (api-service was the dominant source).

Root cause: the IMS auth path loaded a **nested** `@adobe/spacecat-shared-launchdarkly-client@1.0.4` copy under `spacecat-shared-http-utils` (which hard-pinned `1.0.4`). That old client passed no `logger` to `ld.init()`, so the SDK `basicLogger` wrote every level to stderr -> Lambda tagged it `error`. It also created a streaming client per authenticated request with no init timeout, producing reconnect/`waitForInitialization` storms across Lambda freeze/thaw.

- Phase 1 (`launchdarkly-client@1.3.0`): cache (one client per warm container) + `stream:false`/`pollInterval` + `waitForInitialization` timeout + logger mapping.
- Phase 2 (`http-utils@1.29.1`): re-pin the nested dep to `1.3.0`.
- **Phase 3 (this PR):** bump both deps in api-service so the top-level range and the http-utils pin resolve to a **single hoisted** `1.3.0` copy. The nested `1.0.4` is gone, so the fixed client actually loads on the auth path in prod.

## Verification

- `npm install` -> exactly one `launchdarkly-client` in the tree, hoisted at `1.3.0`; no nested copy under `http-utils`; `@launchdarkly/node-server-sdk` at `9.11.2`.
- Lockfile churn is npm hoisting duplicate nested copies (`launchdarkly-client`, `util-arn-parser`, `helix-universal`) to shared copies - same versions, no behavior change.
- `npm test` -> 11742 passing.

## Post-merge

Redeploy api-service, then validate in Coralogix (dev -> prod): LD error templates drop to ~0 and api-service error volume drops ~70%.

Exact pins per repo convention (no `^`/`~`).


MysticatBot (approved):
Hey @solaris007,

**Verdict: Approve** - clean dependency bump completing the 3-phase LaunchDarkly log-noise fix.
_Changes: Bumps `@adobe/spacecat-shared-http-utils` to 1.29.1 and `@adobe/spacecat-shared-launchdarkly-client` to 1.3.0 (exact pin), eliminating the nested 1.0.4 copy that was the root cause of error-level log noise (2 files)._

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 27s | <b>Cost:</b> $1.89 | <b>Commit:</b> <code>c1ea3137bd2b2fdd4cc26faf82d2d28c210e920b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2611, opened 2026-06-15
feat(preflight): attach IMS service token on x-ims-authorization for Mysticat call (SITES-43236)

GeezerPelletier:
## Summary

- Mints a spacecat-api-service IMS v3 service token (via `context.imsClient.getServiceAccessTokenV3()`) and attaches it on a dedicated `x-ims-authorization` header on the `POST /v1/preflight/analyze` call to Mysticat.
- Companion to mystique-deploy [PR #463](https://git.corp.adobe.com/experience-platform/mystique-deploy/pull/463), which adds `/v1/preflight/analyze` as an IMS-validated CGW-Flex endpoint at the `m-dev.adobe.io` edge.
- Required for the deployed spacecat-dev Lambda to call mystique-dev once the public ingress flip in PR #463 lands.

## Why a dedicated header

The CGW-Flex edge gate is keyed on `x-ims-authorization` — **not** `Authorization` — because `callMysticatAnalyze` already uses `Authorization` to forward the customer site's page-auth token (mystique upstreams it to DRS for authenticated page-HTML fetch). Reusing the same header for the IMS service token would clobber that pass-through, and the CGW edge would reject the page-auth token anyway (not an IMS token, not in `allowedClientIDs`).

Both headers ride independently. Three tokens, three distinct purposes:

| Header | Value | Purpose |
|---|---|---|
| `Authorization` (inbound to spacecat) | Customer's IMS user/service token | Authenticates the customer to spacecat (`hasAccess(site)`). Stays at the spacecat edge. |
| `Authorization` (outbound from `callMysticatAnalyze`) | Customer site's page-auth token | Forwarded by mystique to DRS. Unchanged by this PR. |
| `x-ims-authorization` (NEW, outbound) | spacecat IMS v3 service token | Validates at the CGW edge against `allowedClientIDs`. |

## Mint placement (before DB writes)

The mint runs in `createPreflight` **before** the AsyncJob + Preflight DB records are created. A transient IMS failure short-circuits to `500 PREFLIGHT_INTERNAL_ERROR` with no orphaned IN_PROGRESS rows to reconcile.

## Phasing

- **Phase 1 (this PR + companion deploy PR #463).** Mystique-deploy edge gate is `optional: true` — dev e2e works regardless of whether spacecat attaches the header. This PR's spacecat-side attachment lights up the gate.
- **Phase 2 (deploy-side, separate cycle).** Mystique-deploy flips `optional: false` and adds spacecat's IMS client id to `allowedClientIDs`. No further change needed on the spacecat side; the attachment is already in place.

The `tokenType: user` vs `service` question Alina flagged inline on PR #463 is downstream; with `optional: true` the edge doesn't enforce tokenType, so this attachment works either way.

## Reuse — no new env vars

- IMS client wired into the request context via `imsClientWrapper` ([src/index.js:456](https://github.com/adobe/spacecat-api-service/blob/main/src/index.js#L456)).
- Reads existing `IMS_HOST` / `IMS_CLIENT_ID` / `IMS_CLIENT_SECRET` / `IMS_SCOPE` already provisioned in Vault (`dx_mysticat/dev/api-service`).
- `ImsClient.getServiceAccessTokenV3()` already caches the minted token on the instance — warm Lambda invocations short-circuit the IMS round-trip.

## Test plan

- [x] `attaches the IMS service token on x-ims-authorization (Bearer prefix)`
- [x] `keeps Authorization (page-auth) and x-ims-authorization (IMS) on separate headers when both are present`
- [x] `returns 500 PREFLIGHT_INTERNAL_ERROR when IMS service-token mint fails`
- [x] `does not create AsyncJob or Preflight records when IMS mint fails`
- [x] `does not call Mysticat when IMS mint fails`
- [x] All 89 existing + new `createPreflight` tests pass locally
- [ ] Empirical: once PR #463 + the mystique-side `PREFLIGHT_ENABLED` flip both land, a dev-Lambda call lands at mystique-dev with both headers populated (verify via CGW edge logs / pod-side request inspection)

## Fixed Issues

- [SITES-43236](https://jira.corp.adobe.com/browse/SITES-43236) — Phase 1 spacecat-side attachment


src/controllers/preflight.js
```diff
@@ -48,7 +48,7 @@ function PreflightController(ctx, log, env) {
   if (!isNonEmptyObject(ctx)) {
     throw new Error('Context required');
   }
-  const { dataAccess, sqs } = ctx;
+  const { dataAccess, sqs, imsClient } = ctx;
```
MysticatBot, line 51: issue (blocking): `imsClient` is destructured from `ctx` but has no presence guard, unlike `dataAccess` and `sqs` which both throw if missing. If the runtime context ever fails to provide `imsClient` (misconfigured environment, new deployment mode), the error surfaces much later as a cryptic `Cannot read properties of undefined (reading 'getServiceAccessTokenV3')` deep inside the analyze flow.

The established contract in this constructor is fail-fast validation for required dependencies.

**Fix:** Add a guard after the existing `sqs` check:

```js
if (!isNonEmptyObject(imsClient)) {
  throw new Error('IMS client required');
}
```

src/controllers/preflight.js
```diff
@@ -471,6 +483,24 @@ function PreflightController(ctx, log, env) {
...
+    // edge gate (SITES-43236 / mystique-deploy PR #463). Sent on the dedicated
+    // x-ims-authorization header — separate from `Authorization` above, which
+    // carries the customer site's page-auth token for DRS upstream. The
+    // ImsClient caches the token in-memory on the wrapper instance, so warm
+    // Lambda invocations short-circuit the IMS round-trip.
+    //
+    // Mint before the AsyncJob / Preflight DB writes so a transient IMS
+    // failure doesn't leave orphaned IN_PROGRESS records to clean up.
+    let imsServiceToken;
+    try {
```
MysticatBot, line 507: issue (blocking): If `getServiceAccessTokenV3()` resolves successfully but returns an unexpected shape (e.g., `{ accessToken: ... }` after an SDK version bump, or `{}`), `imsServiceToken` becomes `undefined`. Since `hasText(undefined)` returns `false`, the downstream `callMysticatAnalyze` proceeds without the `x-ims-authorization` header silently. When Phase 2 flips CGW to `optional: false`, requests will be rejected at the edge with a 401/403, surfacing as a generic upstream error with no indication the root cause was a missing token.

**Fix:** Add a post-condition check inside the try block:

```js
const tokenPayload = await imsClient.getServiceAccessTokenV3();
imsServiceToken = tokenPayload?.access_token;
if (!hasText(imsServiceToken)) {
  throw new Error('IMS token payload missing access_token');
}
```

This collapses the failure into the existing 500 path with structured logging. Also add a test case where the stub resolves `{}` and assert 500.

MysticatBot (changes requested):
Hey @GeezerPelletier,

**Verdict: Request changes** - two small defensive-coding gaps that will matter once Phase 2 makes the IMS token mandatory.
_Changes: Mints an IMS v3 service token and attaches it on a dedicated `x-ims-authorization` header for CGW-Flex edge validation on the Mysticat preflight analyze call (2 files)._

## Must fix before merge

1. [Important] Missing constructor guard on `imsClient` - `src/controllers/preflight.js:51` (details inline)
2. [Important] `access_token` used without defensive check, enabling silent degradation - `src/controllers/preflight.js:~493` (details inline)

<details>

- nit: Error log interpolates only `e.message`, losing the stack trace for diagnosis of network vs config failures. Consider `log.error('Failed to acquire IMS service token for preflight analyze', { error: e })` or passing the error object directly. - `src/controllers/preflight.js:500`
- suggestion: Test "does not call Mysticat when IMS mint fails" asserts `fetchStub.callCount === 1` which couples to the HEAD probe being exactly one call. Filtering by Mysticat URL would be more resilient to future fetch additions. - `test/controllers/preflight.test.js:~1680`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 32s | <b>Cost:</b> $3.90 | <b>Commit:</b> <code>771a0692cc1fb1a5128fac286622798574e4daee</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

clotton (approved):
Solid, focused change: clean header separation, fail-before-DB-write placement, no new env vars, and genuinely thorough tests (mint-fail → no DB rows → no Mysticat call). The PR description is excellent. A few things worth raising before this graduates to Phase 2.

## 🔴 1. The cached service token is **not expiry-aware** (most important)

The PR cites instance caching as a benefit, but `getServiceAccessTokenV3()` in `shared-ims-client` caches on the instance and returns the token **as long as `access_token` is non-empty — it never checks `expires_in`**:

```js
async getServiceAccessTokenV3() {
  if (hasText(this.serviceAccessTokenV3?.access_token)) {
    return this.serviceAccessTokenV3;   // returned forever; no TTL check
  }
  ...
}
```

Lambda execution environments routinely live well past the 3600s `expires_in`, so a warm instance keeps sending the **same token after it expires**.
- **Phase 1 (`optional: true`)** — harmless, the edge doesn't enforce.
- **Phase 2 (`optional: false`)** — a long-lived warm Lambda starts presenting an expired token → edge 401 → surfaces here as `PREFLIGHT_UPSTREAM_ERROR` (502), intermittently and hard to repro.

Not introduced by this PR (the bug is in `shared-ims-client`), but this PR is the first to depend on that cache. Worth either confirming the CGW edge's TTL tolerance or filing a follow-up to make the cache refresh-aware **before** flipping Phase 2.

## 🟡 2. Mint is now a hard dependency for *all* preflight creation — before it's required

Mint failure returns `500` for every `createPreflight`, but in Phase 1 the edge gate is `optional: true`, so the header isn't needed yet. Net effect: this PR introduces a new failure mode (any IMS blip → every preflight 500) ahead of the benefit. Reasonable alternative: in Phase 1 make a mint failure non-fatal (log + skip the header), and make it fatal in the same Phase 2 change that flips `optional: false`. Fail-closed is defensible too (simpler, warm cache makes it rare) — but it's a deliberate availability tradeoff worth stating explicitly in the PR.

## 🟡 3. `imsClient` isn't validated like its siblings

The constructor destructures `imsClient` but, unlike `dataAccess`/`sqs`, has no `isNonEmptyObject` guard. If the wrapper were ever missing, you'd get `TypeError: Cannot read properties of undefined` caught and reported as *"Failed to acquire IMS service token"* — misleading. Low priority (`imsClientWrapper` is always in the stack), but the matching guard keeps the pattern consistent.

## 🟢 Minor / nits
- **Mint placement** runs *after* the page-auth flow (HEAD probe + `retrievePageAuthentication`, possibly a promise-token round-trip). If you want to fail fast and skip that work on IMS issues, mint earlier. Trivial — "before DB writes" is the rationale that matters and it holds.
- Tests are good. The Bearer-prefix assertion and the both-headers-coexist case (via esmock) are exactly right.

## Verdict
**Approving.** Nothing blocks Phase 1. Before the Phase 2 `optional: false` flip, please get a clear answer on **#1 (token expiry vs. edge TTL)**, and consider a one-line note in the PR about the **#2** availability tradeoff.

GeezerPelletier:
Thanks @clotton and @MysticatBot — pushed `6d5a87b8` addressing the actionable items in one commit:

**Both blocking items (MysticatBot) + the matching constructor-guard call-out (@clotton):**
- Added `isNonEmptyObject(imsClient)` guard in the constructor, matching the existing `dataAccess` / `sqs` validation pattern. New unit test covers the guard. Legacy-controller test `beforeEach` blocks (`createPreflightJob`, `getAllPreflights`, `getPreflightById`) get an `imsClient: { unused: true }` sentinel so they pass the new guard without pulling in the real stub.
- Added a post-condition `hasText(access_token)` check on the mint result, with an explanatory comment + a new test that resolves `{}` and asserts the explicit 500. Closes the silent-degradation path on SDK shape drift.

**Nits:**
- `log.error` now passes the error object alongside the message — stack-aware loggers get the stack; visible message stays consistent with sibling logs in this file (which also use `${e.message}`).
- `fetchStub.callCount === 1` swapped for `fetchStub.getCalls().filter(c => c.args[0]?.includes('/v1/preflight/analyze'))` — resilient to future fetch additions.

**Documented in code (@clotton point #2 — fail-closed availability tradeoff):**
- Added an inline comment on the mint block explaining the **deliberate fail-closed posture in both Phase 1 and Phase 2**. Net: yes this introduces a new failure mode (any IMS blip → 500) ahead of the edge actually enforcing, but the symmetry with Phase 2 (no behavior change at the `optional: false` flip → no new failure modes to surprise callers) is the load-bearing rationale, and IMS reliability makes this a narrow cost in practice. Happy to revisit if anyone disagrees, but I'd prefer not to ship one behavior in Phase 1 and a different one at the gate flip.

**Out of scope (@clotton point #1 — cache-expiry):**

This is the most substantive concern, and you're right it lands on this PR by association rather than ownership. To confirm: yes, `getServiceAccessTokenV3()` in `spacecat-shared-ims-client` ([ims-client.js:197](https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-ims-client/src/clients/ims-client.js#L197)) caches purely on `hasText(access_token)` with no `expires_in` check. Warm Lambda containers commonly outlive 3600s, so a long-warm instance will keep presenting the same token past expiry.

Phase 1 (`optional: true`) is harmless. Phase 2 (`optional: false`) is the risk — intermittent edge 401s surfacing as `PREFLIGHT_UPSTREAM_ERROR` (502), with the symptom decoupled from any caller-side change, which is exactly the kind of flake that takes days to bisect.

My plan, in order:

1. **Empirical CGW edge clock-skew tolerance** — before Phase 2 flips, instrument one warm container's mint timestamp + token-issued-at and confirm the edge actually rejects a stale token at expiry (vs. some grace window). If there's a grace window large enough to cover Lambda container lifetime, we may not need a code change.

2. **If the edge does enforce strictly**, file a follow-up against `spacecat-shared-ims-client` to make the cache refresh-aware (re-mint when `Date.now() >= cachedAt + (expires_in - safetyMargin) * 1000`). One-package upgrade across all spacecat services that touch IMS — modest blast radius, easy to verify.

3. **Sequencing constraint**: 1 (and 2 if needed) must land before Alina's `optional: false` flip on the mystique-deploy side. I'll thread that into SITES-43236's Phase 2 acceptance criteria so it doesn't fall through.

Sound reasonable? Happy to file the follow-up ticket now if you'd rather we work from a concrete tracked item.

MysticatBot (approved):
Hey @GeezerPelletier,

_Changes: Mints an IMS v3 service token and attaches it on a dedicated `x-ims-authorization` header for CGW-Flex edge validation on the Mysticat preflight analyze call (2 files)._

- Constructor guard on `imsClient` now matches the `dataAccess`/`sqs` validation pattern
- Post-condition `hasText(access_token)` check prevents silent header omission on SDK shape drift
- Error log now passes the error object for stack-trace visibility
- Mysticat-call assertion filters by URL rather than raw call count

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 31s | <b>Cost:</b> $3.65 | <b>Commit:</b> <code>6d5a87b876c90824fd29b07420a76db813642bc9</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/controllers/preflight.js
```diff
@@ -308,18 +312,29 @@ function PreflightController(ctx, log, env) {
...
+   *    Kept on a dedicated header — not `Authorization` — so the IMS edge gate
+   *    does not collide with the customer's page-auth token above.
+   *
    * @param {string} mysticatBaseUrl - The base URL of the Mystique service (MYSTIQUE_API_BASE_URL).
    * @param {string} scanId - The AsyncJob ID used as the scan identifier for write-back.
    * @param {string} siteId - The site ID.
    * @param {string} url - The page URL to analyze.
-   * @param {string} [authorizationHeader] - Optional auth header forwarded to Mysticat.
+   * @param {string} [authorizationHeader] - Optional customer-site page-auth header.
+   * @param {string} [imsServiceToken] - Optional spacecat IMS v3 service token (raw access_token).
```
johobot, line 329: The `[imsServiceToken]` JSDoc marks this as optional, but in practice `createPreflight` always provides it — the fail-closed mint guard either yields a non-empty token or returns 500 before this function is reached.

The "optional" framing could mislead a future caller into passing `undefined` expecting a graceful no-op; with Phase 2's `optional: false` edge gate that would become a silent 401. Consider:
```
 * @param {string} imsServiceToken - spacecat IMS v3 service token (raw access_token). Always provided; mint failure short-circuits to 500 before this is called.
```

test/controllers/preflight.test.js
```diff
@@ -150,6 +150,11 @@ describe('Preflight Controller', () => {
       {
         dataAccess: mockDataAccess,
         sqs: mockSqs,
+        // Sentinel — only `createPreflight` actually calls the IMS client;
+        // legacy `createPreflightJob` / `getPreflight*` tests never reach it.
+        // The `createPreflight` describe block injects a real stub via its
+        // own beforeEach.
+        imsClient: { unused: true },
```
johobot, line 157: `{ unused: true }` satisfies `isNonEmptyObject` for suites that never reach `createPreflight`. Works today, but if any code path outside `createPreflight` starts calling `imsClient.getServiceAccessTokenV3()`, these tests will throw a cryptic "not a function" rather than a clear stub failure. A `{ getServiceAccessTokenV3: sinon.stub() }` would be equally terse and surface that more cleanly. Fine to leave as-is if it's considered low risk.

test/controllers/preflight.test.js
```diff
@@ -1567,6 +1607,139 @@ describe('Preflight Controller', () => {
...
+
+    it('attaches the IMS service token on x-ims-authorization (Bearer prefix)', async () => {
+      const response = await preflightController.createPreflight({
+        params: { siteId: 'test-site-123' },
+        data: { url: 'https://main--example-site.aem.page/test.html' },
+        attributes: { authInfo: mockAuthInfo },
+      });
+      expect(response.status).to.equal(202);
+      expect(mockImsClient.getServiceAccessTokenV3).to.have.been.calledOnce;
+      const [, calledOptions] = fetchStub.secondCall.args;
```
johobot, line 1620: `fetchStub.secondCall` assumes the second fetch is always the Mysticat analyze call (the first being the HEAD probe). If a future code path adds another fetch before the analyze call, this assertion silently tests the wrong call.

The `'does not call Mysticat when IMS mint fails'` test (line 1718) already uses a URL-filter pattern that's more resilient — using the same approach here would be consistent:
```js
const mysticatCall = fetchStub.getCalls().find(
  (c) => typeof c.args[0] === 'string' && c.args[0].includes('/v1/preflight/analyze'),
);
const [, calledOptions] = mysticatCall.args;
```

src/controllers/preflight.js
GeezerPelletier: Keeping the `[imsServiceToken]` optional framing deliberately, paired with the runtime `...(hasText(imsServiceToken) && ...)` defense.

The JSDoc describes the **function's parameter contract**, not the current caller's invariant. The function safely handles undefined/empty — that's what the conditional spread is for. If the JSDoc said *"always provided"* but the code still defended against absence, the doc would lag the implementation.

The fail-closed mint guard in `createPreflight` guarantees a non-empty token reaches *that* call site — that's a caller-side invariant, not a parameter constraint. A future caller (different surface that wants to call mystique) could legitimately pass undefined; the defensive spread keeps that safe and Phase 2's `optional: false` edge gate would surface the 401 to *that* caller, not to this PR's `createPreflight`.

Net: the JSDoc accurately describes what the function accepts.

test/controllers/preflight.test.js
GeezerPelletier: Going to leave `{ unused: true }` — your own caveat (*"Fine to leave as-is if it's considered low risk"*) captures the call, but worth spelling out why the suggested alternative isn't strictly better for future readers:

`{ getServiceAccessTokenV3: sinon.stub() }` covers exactly **one** method on the ImsClient surface. If a future code path adds a call to `getServicePrincipalAccessToken` (or any other ImsClient method) and lands in a controller covered by one of these legacy beforeEach blocks, you'd hit the same opaque *"not a function"* failure mode — just with a partial stub silently succeeding on one method and breaking on others. A truly defensive stub would be a Proxy that throws on any property access, but that's overkill for tests that don't exercise the IMS path at all.

The sentinel's job is narrowly to pass `isNonEmptyObject(imsClient)` so the new constructor guard doesn't block legacy controller construction. The "not a function" failure mode you flag is actually informative when it happens — it tells the next person *"you're calling an IMS method from a test that doesn't stub it"* rather than hiding behind a partial stub. Keeping the sentinel minimal is intentional.

test/controllers/preflight.test.js
GeezerPelletier: Pushing back on this one — `fetchStub.secondCall.args` is the **established file convention** for `createPreflight` tests; see lines 1236, 1248, 1321, 1355, 1572 and many others. The convention works because every test in this `describe` block sets up `fetchStub.onFirstCall().resolves(HEAD)` + `fetchStub.onSecondCall().resolves(Mysticat)` in the shared `beforeEach`, so the second call IS the Mysticat call by construction.

The URL-filter pattern was applied to *"does not call Mysticat when IMS mint fails"* because there it served a specific purpose — counting calls and asserting the **absence** of the Mysticat call. For tests like this one that just want the Mysticat headers and expect exactly one Mysticat call, `secondCall` is the right shape and matches the surrounding tests.

Switching only the new test to URL-filtering would create asymmetry that future readers would (rightly) question — *"why does this one test use a different pattern?"*. If the file's convention needs to migrate to URL-filtering wholesale because a future code path adds a third pre-Mysticat fetch, that's a separate sweep across the whole file, not a piecemeal change here.

PR #2616, opened 2026-06-16
fix: honor date filter on gap-topics endpoints | LLMO-5625

calvarezg:
## Summary
The Topic Opportunities table (and its CSV export) on the Visibility Overview dashboard ignored the date-range filter. The gap-topics handlers set `target_date` (a string) on `GapTopicsRequest` / `GapTopicsTotalsRequest`, but those protos have no `target_date` field — only `date` (a `common.v1.Date` message). With `ignoreUnknownFields`, the value was silently dropped and never reached the upstream gRPC. This fixes the list, totals, and export handlers to send the proto `date` message instead.

## Changes
- `gap-topics.js`, `gap-topics-totals.js`, `gap-topics-export.js`: parse the incoming `date` query param (`YYYY-MM-DD`) into a `{year, month, day}` `date` message and drop the silently-dropped `target_date` string — mirroring the working `source-opportunities` handler.
- The `date` is omitted when absent/invalid, so the upstream falls back to the latest snapshot rather than failing on a day-less date.
- `brand-topics` is intentionally untouched: its proto has a real `target_date` string field, so it was already correct.

## Test plan
- [x] Verified live (local stack): `gap-topics-totals` ALL count now varies by snapshot day — 1,013,763 (`2026-06-06`) and 1,006,366 (`2026-06-10`) vs the previously-constant 1,000,437.
- [x] `eslint` passes on all three handlers.
- [ ] Confirm Topic Opportunities table + CSV export reflect the selected date on stage (export not yet live-verified).

## Notes
No unit tests added — these handlers are excluded from coverage (`/* c8 ignore */`) per the existing convention; behavior verified manually against the live stack.

## Related Issues
https://jira.corp.adobe.com/browse/LLMO-5625

---
- [x] make sure to link the related issues in this description.
- API specification is unchanged (the `date` query param was already accepted; this only fixes how it maps to the proto), so the API-spec checklist items below are N/A.

### DIFF
--- a/src/support/ai-visibility/handlers/v1/topic/gap-topics-export.js
+++ b/src/support/ai-visibility/handlers/v1/topic/gap-topics-export.js
@@ -49,7 +49,15 @@ export async function handleGapTopicsExport(sp, clients) {
   const country = resolveCountry(sp) || COUNTRY_ENUM.WORLDWIDE;
   const sortBy = sp.get('sortBy') || TOPICS_REQUEST_ORDER_BY_ENUM.TOTAL_COMPETITOR_MENTIONS;
   const sortDirection = sp.get('sortDirection') || ORDER_DIRECTION_ENUM.DESC;
-  const date = sp.get('date');
+  // `GapTopicsRequest.date` is a `common.v1.Date` message, not the string
+  // `target_date`. Parse an exact `YYYY-MM-DD` snapshot day so the export honors
+  // the date filter like the list/totals handlers; omit otherwise for latest. See
+  // gap-topics.js.
+  const dateRaw = sp.get('date')?.trim();
+  const dm = dateRaw && /^(\d{4})-(\d{2})-(\d{1,2})$/.exec(dateRaw);
+  const snapshotDate = dm
+    ? { year: Number(dm[1]), month: Number(dm[2]), day: Number(dm[3]) }
+    : undefined;
   const { limit, offset } = parseLimitOffset(sp);
 
   const gapKindsRaw = sp.get('gapKinds');
@@ -78,7 +86,7 @@ export async function handleGapTopicsExport(sp, clients) {
         range: { limit, offset },
         dimension_filter_ql: dimensionFilterQl,
         metric_filter_ql: metricFilterResult.metricFilterQl,
-        target_date: date,
+        ...(snapshotDate ? { date: snapshotDate } : {}),
       },
       PROTO_FROM_JSON,
     );
--- a/src/support/ai-visibility/handlers/v1/topic/gap-topics-totals.js
+++ b/src/support/ai-visibility/handlers/v1/topic/gap-topics-totals.js
@@ -39,7 +39,14 @@ export async function handleGapTopicsTotals(sp, clients) {
   const competitorDomains = parseCompetitorDomainsList(sp);
   const engine = engineToLlm(sp.get('engine')) || LLM_ENUM.ALL;
   const country = resolveCountry(sp) || COUNTRY_ENUM.WORLDWIDE;
-  const date = sp.get('date');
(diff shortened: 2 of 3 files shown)


MysticatBot (approved):
Hey @calvarezg,

**Verdict: Approve** - clean, well-scoped bug fix that follows the established pattern.
_Changes: Fixes gap-topics list, totals, and export handlers to send a proto common.v1.Date message instead of the silently-dropped target_date string, so the date-range filter actually reaches the upstream gRPC service (3 files)._

<details>

- nit: Regex inconsistency - month requires exactly 2 digits but day allows 1-2, so 2024-6-05 silently falls back to latest while 2024-06-5 works. Mirrors the existing competitors.js pattern and the UI sends ISO 8601, so unlikely to bite in practice. - `gap-topics.js:79`
- suggestion: Extract a parseSnapshotDate(sp) helper into grpc-utils.js - same 5-line block is now in 3 files here plus competitors.js. One place to fix the regex or add validation later. - `gap-topics.js:78-82`
- nit: No semantic validation on parsed integers - 2025-13-99 produces month:13, day:99 and flows to the upstream gRPC. The upstream proto likely rejects it, but a bounds check (month 1-12, day 1-31) would fail fast at the API boundary. - `gap-topics.js:79`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 4s | <b>Cost:</b> $2.96 | <b>Commit:</b> <code>953575aa13ec967e3a416f3b95f308ffebcc21dd</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

calvarezg:
## Triage of MysticatBot review

Thanks @MysticatBot — [review](https://github.com/adobe/spacecat-api-service/pull/2616#pullrequestreview-4506385249). All three are non-blocking; here's the disposition.

**Deferred to a follow-up (out of scope for this focused bug fix):**
- Extract a shared `parseSnapshotDate(sp)` helper into `grpc-utils.js`, with month/day bounds validation, and adopt it across `gap-topics`, `source-opportunities`, and `competitors`. This folds together the DRY suggestion (the 5-line parse block now lives in ~5 sites) and the `2025-13-99` validation nit — both are best solved once in a shared helper. Kept out of this PR because the real value needs the pre-existing copies refactored too, which is broader than this fix.

**Dismissed:**
- Regex `month \d{2}` vs `day \d{1,2}` inconsistency — intentional. This regex is identical to the established `source-opportunities` (`brands.js:661`) and `competitors` handlers, and the UI only ever sends zero-padded ISO dates (`srDateToISO` pads), so the 1-digit case isn't reachable. Tightening only `gap-topics` would diverge from the family.

calvarezg:
Closing — superseded by #2620 (LLMO-5652, "bump ai-seo-ts").

This PR switched the gap-topics handlers to send the proto `date` (a `common.v1.Date` message) because, in the SDK we had pinned (`@quazar/ai-seo-ts v0.97.3`), `GapTopicsRequest` had no `target_date` field — so the existing `target_date: date` was silently dropped by `fromJson` and the date filter never reached the backend.

#2620 bumps the SDK so `GapTopicsRequest` / `GapTopicsTotalsRequest` now expose `target_date` (string, accepts `YYYY-MM` or `YYYY-MM-DD`) and mark `date` deprecated. With that, the **pre-existing** `target_date: date` code works as-is — no handler change needed — and `target_date` is the non-deprecated field. So this change is both redundant and would use the deprecated field.

Verified end-to-end after #2620: Topic Opportunities now varies by month (e.g. 801,607 for Feb vs 927,155 for Apr, vs the old constant 1,000,437), with the UI sending a month-granular `date` (project-elmo-ui#2091).

PR #2617, opened 2026-06-16
fix: remove dormant data-access-v2 fallback (drops electrodb/jsonschema, fixes DEP0169 log noise)

solaris007:
## What

Removes the dormant `@adobe/spacecat-shared-data-access-v2` (legacy ElectroDB/DynamoDB) fallback from `src/support/data-access.js`. The data-access middleware is now V3-only (PostgREST): it requires `POSTGREST_URL` and delegates to `@adobe/spacecat-shared-data-access`.

## Why

After the Postgres migration, every environment runs V3 (`DATA_SERVICE_PROVIDER=postgres` in Vault across dev/stage/prod - verified). The v2 alias (`npm:@adobe/spacecat-shared-data-access@2.109.0`) was a never-taken switch-back fallback, and its static import was the **only** path pulling `electrodb@3.5.0 -> jsonschema@1.2.7` (the org-wide **DEP0169** `url.parse()` source) into the production tree. api-service has no `cloud-manager-client`, so removing v2 clears DEP0169 here.

## Changes

- `src/support/data-access.js`: V3-only (require `POSTGREST_URL`, delegate to V3; drop the v2 import, the `DATA_SERVICE_PROVIDER === 'postgres'` branch, the V2 fallback, and the `c8 ignore` markers).
- `package.json`: remove the `@adobe/spacecat-shared-data-access-v2` alias.
- `test/support/data-access.test.js`: new (esmock, no real I/O).
- `test/index.test.js`: add `POSTGREST_URL` to the middleware-chain test's `context.env` (it exercises the real wrapper, which now prechecks `POSTGREST_URL`).
- `README.md`: drop the stale DynamoDB integration-test suite references (that suite/dir no longer exists); PostgreSQL-only now.
- `package-lock.json`: regenerated.

## Note: electrodb devDependency (intentional, not a regression)

`test/controllers/fixes.test.js` imports `electrodb` directly (`new electrodb.Service(...)`). Since the v2 alias was electrodb's sole provider, removing it broke that test-only import. V3 declares electrodb only as a devDependency, so I added `electrodb: "3.9.0"` (exact, matching V3's dev pin) to **devDependencies**. This keeps the test green with **zero production impact**: electrodb is dev-only (not bundled into the Lambda), and 3.9.0 ships `jsonschema@1.5.0` (not the DEP0169 `1.2.7`).

## The 8 DATA_SERVICE_PROVIDER references

Only the 1 provider-selection branch (in `data-access.js`) was removed. The other 8 occurrences in `src/controllers/**` (feature-flags, brands, llmo/*) are operator-facing error/log message strings gated on runtime PostgREST capability (`Site?.postgrestService` / `dataAccess?.services?.postgrestClient`), not provider selection - left untouched.

## Verification

- `npm ls electrodb --omit=dev` -> **empty** (production tree clean) | `jsonschema@1.2.7` gone (only dev `1.5.0` via electrodb@3.9.0 dev pin remains)
- `grep spacecat-shared-data-access-v2` -> empty
- `npm test` -> 11780 passing; coverage 99.9% (gate is 90%); lint clean
- behavioral no-op (all envs already V3)

Completes the org-wide v2-removal / DEP0169 sweep (13th repo).


MysticatBot (approved):
Hey @solaris007,

**Verdict: Approve** - clean dead-code removal with correct test coverage.
_Changes: removes the dormant V2 (DynamoDB/ElectroDB) data-access fallback, making the middleware V3-only (PostgREST), and eliminates the electrodb/jsonschema DEP0169 source from the production dependency tree (6 files)._

<details>

- nit: `const { env = {} } = context` - JS default assignment does not activate for `null`, only `undefined`. If upstream code ever sets `context.env = null`, `null.POSTGREST_URL` throws a generic `TypeError` instead of the friendly "POSTGREST_URL is not configured" message. Consider `const env = context.env || {};` if you want to guard both cases. - `src/support/data-access.js:15`
- suggestion: a whitespace-only `POSTGREST_URL` (e.g. `' '`) passes the `!env.POSTGREST_URL` guard since non-empty strings are truthy; `!env.POSTGREST_URL?.trim()` would catch that edge case at this layer rather than letting it surface deeper in the PostgREST client. - `src/support/data-access.js:16`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 51s | <b>Cost:</b> $3.86 | <b>Commit:</b> <code>f004956c039d0a6146aef6b2ecbb54aaf2d1a6c3</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2619, opened 2026-06-16
fix(llmo): plumb operator market into V2 onboarding + cascade region to prompts (LLMO-5645)

rainer-friederich:
## Problem
The operator-selected market only reached the legacy V1 path. V2/Brandalf onboardings (all since the GA cutoff 2026-04-01) dropped it, so brands/prompts ended up region `WW` (worldwide), which DRS schedulers cannot map. Jira: LLMO-5645.

## TL;DR:
- Passes the market into V2 onboarding — the operator-selected region is now plumbed through Brandalf onboarding and into the DRS job parameters (it was only wired to the legacy V1 path before).
- Seeds the brand region correctly — new brands start with the selected region instead of a placeholder.
- Guards region consistency on brand save — rejects (400) removing a region that prompts still use, telling the operator to reassign or delete those prompts first.
- Clear error message — e.g. Cannot remove region(s) still used by prompts: DE (3 prompts).
- Backward-compatible — works with the current DRS; no deploy-ordering dependency with the DRS PR.

## Changes
- **onboarding** (`llmo-onboarding.js`): forward the operator `region` into the V2 Brandalf DRS job parameters, and seed it into the initial stub brand (in place of the hardcoded `['gl']`) when supplied.
- **brand-region consistency guard** (`prompts-storage.js` + `brands.js`): on a brand region change, `updateBrandForOrg` rejects the update (400) if any non-deleted prompt still uses a region being **removed** (`Cannot remove region(s) still used by prompts: …`), and asks the operator to relocate those prompts first. DRS schedules off each prompt's `regions`, so dropping a brand region would otherwise orphan prompts on a market the brand no longer covers. Enforced before the brand is mutated; `WW` is treated like any region (strict — legacy WW cleanup goes through the recovery script, which rewrites brand + prompt regions together).

  _(Earlier revisions of this PR used a brand→prompt cascade; it was replaced by this guard because wholesale-rewriting prompt regions has no sound semantics when a region is removed from a multi-market brand.)_

## Companion PR
Sends the `region` job parameter to llmo-data-retrieval-service: https://github.com/adobe-rnd/llmo-data-retrieval-service/pull/2205
Either merge order is safe — DRS falls back to US when the param is absent.

## Notes
- Elmo UI needs **no** change (region picker is from the static catalog; the propagation belongs server-side).
- The guard enforces `prompt.regions ⊆ brand.regions` only at the point of **removing** a brand region (multi-market prompts remain fully supported); it never mutates prompts.

## Testing
`test/controllers/brands.test.js`, `test/support/prompts-storage.test.js`, `test/controllers/llmo/llmo-onboarding.test.js` — 529 passing (incl. the >200-row chunk boundary), ESLint clean.


src/controllers/brands.js
```diff
@@ -1368,6 +1369,15 @@ function BrandsController(ctx, log, env) {
       // baseUrl is read-only (resolved from baseSiteId) — strip from updates.
       delete updates.baseUrl;
 
+      // LLMO-5645: capture the brand region BEFORE the update so we can cascade
+      // a region change to the brand's prompt regions (the field DRS schedules
+      // off). Only read when a region change was actually requested.
+      let oldRegions = null;
+      if (updates.region !== undefined) {
```
MysticatBot, line 1383: issue (blocking): There is a time-of-check-to-time-of-use window between `getBrandById` (reading old region) and `updateBrand` (writing new region). A concurrent region update to the same brand could cause the cascade to operate on stale `oldRegions`, potentially matching the wrong prompts.

The non-fatal design and self-healing property (next edit re-runs the cascade) mitigate this, but the limitation should be documented so future maintainers understand the trade-off.

Suggested fix: add a code comment noting the race, e.g.:
```js
// Known TOCTOU: concurrent region edits may cause a partial cascade;
// the next edit retries because the cascade is non-fatal and idempotent.
```

src/support/prompts-storage.js
```diff
@@ -897,6 +897,122 @@ export async function updatePromptById({
...
+  let updated = 0;
+  for (let i = 0; i < matchedIds.length; i += CHUNK) {
+    const chunk = matchedIds.slice(i, i + CHUNK);
+    // eslint-disable-next-line no-await-in-loop
+    const { data: rows, error: updErr } = await postgrestClient
+      .from('prompts')
+      .update({ regions: newRegions, updated_by: updatedBy })
+      .eq('organization_id', organizationId)
+      .eq('brand_id', brandUuid)
+      .in('id', chunk)
```
MysticatBot: issue (blocking): If the PostgREST update succeeds for chunk 0 (200 prompts) but fails on chunk 1, the function throws. The caller catches this and logs the error, but neither the log message nor the thrown error includes how many prompts were already updated before the failure. An operator seeing "prompt cascade failed" has no visibility into partial state.

Suggested fix - include the partial count in the error:
```js
throw new Error(`Failed to cascade region to prompts (${updated}/${matchedIds.length} already updated): ${updErr.message}`);
```

src/controllers/llmo/llmo-onboarding.js
```diff
@@ -215,18 +232,19 @@ export async function ensureInitialCustomerConfigV2({
       brandId = `${brandId}-${siteId.slice(0, 8)}`;
     }
```
MysticatBot, line 233: issue (blocking): The `region` parameter was added to `ensureInitialCustomerConfigV2` and is used in the existing-config branch (where a new brand is pushed onto an existing config). This path has no test coverage - the new tests only exercise `buildInitialCustomerConfigV2` and `triggerBrandalfOnboardingJob`.

This is the path Codecov flags as under-covered. If it regresses, brands added to multi-site organizations would silently revert to `["gl"]`.

Suggested fix: add a test case for `ensureInitialCustomerConfigV2` where existing config already exists and a new site/brand is appended with a specific `region`, verifying both `regions` and `brandAliases.regions` on the pushed brand.

MysticatBot (changes requested):
Hey @rainer-friederich,

**Verdict: Request changes** - three addressable issues around operational visibility and test coverage.
_Changes: Plumbs operator-selected market into V2/Brandalf onboarding and adds a cascading region update from brands to prompts (6 files)._

## Must fix before merge
1. [Important] TOCTOU race between read-old-region and update-brand should be documented - `src/controllers/brands.js:1376` (details inline)
2. [Important] Partial cascade failure hides how many prompts were already updated - `src/support/prompts-storage.js:1005` (details inline)
3. [Important] Missing test coverage for `ensureInitialCustomerConfigV2` existing-config path with `region` - `src/controllers/llmo/llmo-onboarding.js:233` (details inline)

<details>

- nit: `onboardingStubRegions(region)` called twice with identical input at lines 1433/1435 - hoist to a `const` like the pattern at line 232 - `src/controllers/llmo/llmo-onboarding.js:1433`
- nit: `normalizeRegionsForCompare` silently coerces `null`/`undefined` to `"null"`/`"undefined"` via `String(r)` - filter `r != null` before coercion - `src/support/prompts-storage.js:940`
- suggestion: test for `buildInitialCustomerConfigV2` asserts `brand.brandAliases` but not top-level `brand.regions` - add `expect(brand.regions).to.deep.equal(['US'])` - `test/controllers/llmo/llmo-onboarding.test.js:3935`
- suggestion: 5000-row READ_CAP is adequate for current scale but consider cursor-based pagination if prompt counts grow - `src/support/prompts-storage.js:955`
- nit: cascade error log uses `.message` only - pass full error object as second arg for stack visibility - `src/controllers/brands.js:1407`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 18s | <b>Cost:</b> $6.56 | <b>Commit:</b> <code>2ad7a5eeaac01fb0ace52052d6fb0ca7c7572d03</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the review in 6fa0a3b5.

**Must-fix (blocking) — done:**
- Documented the read-old-region/update-brand TOCTOU in `updateBrandForOrg` with a comment explaining why it's tolerated (cascade is non-fatal and idempotent — the next region edit re-runs it and converges).
- The cascade chunk-failure error now includes partial progress: `Failed to cascade region to prompts (N/total already updated): ...`, so an operator sees how many prompts were already rewritten before the failure.
- Added test coverage for the `ensureInitialCustomerConfigV2` **existing-config** branch with an operator `region` — verifies both `regions` and `brandAliases.regions` on the appended brand (this was the Codecov-flagged under-covered path).

**Non-blocking — done:**
- `normalizeRegionsForCompare` now filters `null`/`undefined` before coercion (no more `"null"` entries).
- Hoisted the duplicated `onboardingStubRegions(region)` call to a `const stubRegions`.
- Cascade-failure `log.error` now passes the full error object as the second arg for stack visibility.

**Not applied (with reason):**
- The suggestion to also assert top-level `brand.regions` in the `buildInitialCustomerConfigV2` test: that function builds the brand via `convertV1ToV2`, which carries regions only in `brandAliases` — there is no top-level `brand.regions` on that path (it exists only on the existing-config append path, which the new test covers). Asserting it would test a field the code doesn't produce.
- Cursor-based pagination for the 5000-row `READ_CAP`: noted as a future consideration; a single brand carries tens-to-hundreds of prompts today, and the cap already warns rather than silently truncating.

`prompts-storage.js` patch coverage is back to 100%. All affected suites pass, ESLint clean.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all prior blocking findings addressed cleanly, no new concerns.
_Changes: Plumbs operator-selected market into V2/Brandalf onboarding and adds a cascading region update from brands to prompts (6 files)._

Note: CI checks are currently failing (deploy/release jobs) - confirm these are pre-existing or environment-specific before merge.

- TOCTOU race between read-old-region and update-brand is now documented with a clear tolerance rationale (non-fatal, idempotent, self-healing on next edit)
- Partial cascade failure error now includes progress count (`N/M already updated`) and passes full error object for stack visibility
- Test coverage for `ensureInitialCustomerConfigV2` existing-config path with `region` added - verifies both `regions` and `brandAliases` on the appended brand
- `normalizeRegionsForCompare` null/undefined guard added before coercion
- Duplicated `onboardingStubRegions(region)` call hoisted to `const stubRegions`

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 30s | <b>Cost:</b> $3.08 | <b>Commit:</b> <code>6fa0a3b570570fcf488a911bbc4eb5c5572836ac</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Design change (commit e923fc9b): **replaced the brand→prompt region cascade with a consistency guard.**

Why — reviewing the cascade against the multi-region case surfaced real problems:
- It only rewrote prompts whose **entire** region set exactly equaled the brand's **old** set, then replaced it wholesale. The only multi-region prompt in the tests was a *spared* case — we never actually proved a multi-region prompt gets rewritten, and wholesale-replacing a multi-market prompt's regions is semantically wrong anyway.
- Removing a region from a multi-market brand (`['US','DE'] → ['US']`) only touched prompts that were *exactly* `['US','DE']`; a `['DE']`-only or `['DE','FR']` prompt was left orphaned on a market the brand no longer covered.
- There's no sound way to auto-distribute prompts when a region is removed (reassign? delete? keep?) — it's a business decision, not something code should guess.

New behavior — `updateBrandForOrg` now **rejects** (400) a region change that would remove a region still used by any non-deleted prompt:
`Cannot remove region(s) still used by prompts: DE (3 prompts), FR (1 prompt). Reassign or delete those prompts first…`
This enforces `prompt.regions ⊆ brand.regions` at the point of removal and keeps multi-market prompts fully supported. `WW` is treated like any region (strict); legacy WW cleanup goes through the recovery script, which rewrites brand + prompt regions together in one transaction (DRS runbook updated in the companion PR).

- `prompts-storage.js`: removed `cascadeBrandRegionToPrompts`; added `findPromptsBlockingRegionRemoval` (per-removed-region prompt counts, read-capped + warns).
- `brands.js`: guard runs **before** the brand is mutated.
- Tests: guard unit tests (no-removal/superset, blocking counts, multi-region strip, case-insensitive, WW-strict, read-cap warn, client/read errors) + controller allow/reject(400, brand not mutated)/no-op. `prompts-storage.js` and `brands.js` at 100% line coverage.

Re-requesting review since this is a behavior change, not the earlier nits.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - guard design is sound, no blocking concerns.
_Changes: Replaces the brand-region cascade with a consistency guard that rejects (400) region removal when prompts still reference the removed region, and plumbs the operator-selected market into V2/Brandalf onboarding (6 files)._

Note: CI checks are currently failing (deploy/release jobs) - confirm these are pre-existing or environment-specific before merge.

<details>

- nit: The guard has a TOCTOU window (read brand regions, check prompts, then update brand) with no inline documentation. The prior cascade revision had a TOCTOU comment; the new guard design dropped it. A one-liner noting the accepted race (e.g. "Best-effort guard, not transactional; concurrent prompt creation could bypass - acceptable given low-frequency brand-region edits") would prevent future maintainers from assuming the guard is airtight. - `src/controllers/brands.js:1377`
- suggestion: No controller-level test for `updates.region = []` (remove all regions from a brand). The logic is correct (all old regions become "removed", guard fires if any prompt uses them) and the unit tests on `findPromptsBlockingRegionRemoval` exercise this, but a controller test would document the edge case explicitly. - `test/controllers/brands.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 34s | <b>Cost:</b> $4.60 | <b>Commit:</b> <code>e923fc9b736a99e3dc606a5abfd46d6ebea14048</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the 2 non-blocking nits from the approving review in 55e29b8c (rebased onto the latest main merge):

- **TOCTOU comment** — documented the guard as best-effort/non-transactional in `updateBrandForOrg`: a prompt created in a removed region between the check and the brand update could slip past; acceptable given how infrequent brand-region edits are, it's re-checked on the next edit, and a later prompt still can't be scheduled for a region the brand lacks.
- **`region: []` controller test** — added a test for clearing *all* regions: every old region becomes "removed" and the guard fires (400, brand not mutated) if any prompt still uses one.

`brands.js` stays at 100% line coverage; 411 tests pass, ESLint clean. Leaving the approval in place since these are non-blocking polish.

Re CI: the red checks are the pre-existing `it-postgres` ECR-image-missing infra failure (`mysticat-data-service:v5.13.0` not in the registry; the pin was last touched in May, not by this PR) plus the expected PR-context `deploy-stage`/`branch-deploy`/`semantic-release` jobs. `codecov/patch` and `build` pass.

PR #2620, opened 2026-06-16
feat: LLMO-5652 bump ai-seo-ts

dmitriigulkin-semrush:
Bump ai-seo-ts version to support latest API changes


calvarezg (approved):
LGTM — local testing confirms no callsite breaks from the non-optional  field. Clean vendor bump.

calvarezg (approved):
LGTM - local testing confirms no callsite breaks from the non-optional searchType field. Clean vendor bump.

PR #2626, opened 2026-06-17
feat: add v2 brand claims guidance fields

davidaurelio:
## Summary
- expose flat v2 brand guidance fields brandContext and mentionSentimentGuidance on existing brand APIs
- validate wrong types and over-4000-character guidance with 400 responses
- normalize persistence so omitted fields preserve, blank/null clears, and non-empty strings are trimmed
- map legacy v1 claims guidance into the v2 brand shape and document fields in OpenAPI
- add unit and Postgres integration coverage for create/get/list/update/clear behavior

## Testing
- npm test -- --spec=test/support/brands-storage.test.js --spec=test/controllers/brands.test.js --spec=test/support/customer-config-mapper.test.js
- npx mocha --timeout 10000 test/support/brands-storage.test.js test/controllers/brands.test.js test/support/customer-config-mapper.test.js
- IT_SERVER_PORT=3005 npx mocha --require test/it/postgres/harness.js --timeout 30000 test/it/postgres/brands.test.js
- IT_SERVER_PORT=3005 npx mocha --require test/it/postgres/harness.js --timeout 30000 test/it/postgres/**/*.test.js using local mysticat-data-service checkout image/migrations
- npm run docs:lint
- npm run lint

### DIFF
--- a/src/support/brands-storage.js
+++ b/src/support/brands-storage.js
@@ -34,7 +34,11 @@ function normalizeNullableText(value, fieldName) {
     return null;
   }
   if (typeof value !== 'string') {
-    throw new Error(`${fieldName} must be a string or null`);
+    // Tag with a 400 status so callers that reach storage without going through
+    // the controller's validation surface a client error rather than a 500.
+    const error = new Error(`${fieldName} must be a string or null`);
+    error.status = 400;
+    throw error;
   }
   const trimmed = value.trim();
   return hasText(trimmed) ? trimmed : null;
@@ -134,8 +138,8 @@ function mapDbBrandToV2(row) {
     status: row.status || 'active',
     origin: row.origin || 'human',
     description: row.description || null,
-    brandContext: row.brand_context || null,
-    mentionSentimentGuidance: row.mention_sentiment_guidance || null,
+    brandContext: row.brand_context ?? null,
+    mentionSentimentGuidance: row.mention_sentiment_guidance ?? null,
     vertical: row.vertical || null,
     region: row.regions || [],
     urls,
--- a/src/support/brands-storage.js
+++ b/src/support/brands-storage.js
@@ -26,6 +26,20 @@ const BRAND_SELECT = [
   'brand_urls(url)',
 ].join(', ');
 
+function normalizeNullableText(value, fieldName) {
+  if (value === undefined) {
+    return undefined;
+  }
+  if (value === null) {
+    return null;
+  }
+  if (typeof value !== 'string') {
+    throw new Error(`${fieldName} must be a string or null`);
+  }
+  const trimmed = value.trim();
+  return hasText(trimmed) ? trimmed : null;
+}
+
 /**
  * Splits a full URL string into its base URL and path.
  * e.g. "https://example.com/products" -> { base: "https://example.com", path: "/products" }
@@ -120,6 +134,8 @@ function mapDbBrandToV2(row) {
     status: row.status || 'active',
     origin: row.origin || 'human',
     description: row.description || null,
+    brandContext: row.brand_context || null,
+    mentionSentimentGuidance: row.mention_sentiment_guidance || null,
     vertical: row.vertical || null,
     region: row.regions || [],
     urls,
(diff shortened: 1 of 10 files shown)


src/support/brands-storage.js
```diff
@@ -120,6 +134,8 @@ function mapDbBrandToV2(row) {
     status: row.status || 'active',
     origin: row.origin || 'human',
     description: row.description || null,
+    brandContext: row.brand_context || null,
```
MysticatBot: issue (blocking): The `||` operator coerces empty string to `null`, collapsing a semantic state the write path (`normalizeNullableText`) explicitly distinguishes. If any other writer (migration, direct SQL, another service) stores `""` in these columns, the API silently returns `null` instead of surfacing the inconsistency.

The write path carefully handles undefined vs null vs blank vs non-empty. The read path should match.

**Fix:** Replace `row.brand_context || null` with `row.brand_context ?? null` (same for `mention_sentiment_guidance`). Nullish coalescing maps only `null`/`undefined` to `null`, preserving the contract.

src/support/customer-config-mapper.js
```diff
@@ -166,6 +166,8 @@ export function convertV1ToV2(llmoConfig, brandName, imsOrgId) {
     origin: 'system',
     region: brandRegions,
     description: '',
+    brandContext: llmoConfig.claims?.brandContext,
```
MysticatBot: issue (blocking): Values from `llmoConfig.claims` bypass the 4000-character length validation (which lives only in the controller). If legacy V1 config data exceeds 4000 characters, it persists successfully and the API serves a response that violates the OpenAPI `maxLength: 4000` contract. Clients using strict schema validation will reject the response.

**Fix:** Either (a) apply the same max-length check here (truncate or skip oversized values with a log warning - truncation is likely correct for a legacy compat path), or (b) move the length check into `normalizeNullableText` so it is enforced regardless of entry point.

MysticatBot (changes requested):
Hey @davidaurelio,

**Verdict: Request changes** - two semantic issues in the persistence and legacy mapping layers.
_Changes: Adds two nullable text fields (brandContext, mentionSentimentGuidance) to the v2 brand API surface with input validation, normalization, legacy v1 mapping, and OpenAPI documentation (9 files)._
Note: CI check `validate-pr-title` is currently failing - resolve before merge.

## Must fix before merge
1. [Important] Falsy coercion (`||`) in read mapping loses semantic distinction between empty-string and null - `src/support/brands-storage.js:137` (details inline)
2. [Important] V1-to-V2 mapper bypasses length validation; legacy data >4000 chars would violate OpenAPI maxLength contract on responses - `src/support/customer-config-mapper.js:169` (details inline)

<details>

- nit: `normalizeNullableText` throws a bare `Error` (surfaces as 500 if reached) rather than returning null with a log warning for defense-in-depth consistency - `src/support/brands-storage.js:35`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 43s | <b>Cost:</b> $7.86 | <b>Commit:</b> <code>33bb3e02a10adefb634938ebc277977c7c794739</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @davidaurelio,

**Verdict: Approve** - all prior findings addressed cleanly, no new blocking issues.
_Changes: Adds two nullable text fields (brandContext, mentionSentimentGuidance) to the v2 brand API surface with input validation, normalization, legacy v1 mapping, and OpenAPI documentation (9 files)._

<details>

- suggestion: OpenAPI `maxLength: 4000` on input schemas counts raw string length, but the controller validates `value.trim().length` - a 4004-char padded string passes the controller but violates the schema. If any API gateway enforces the schema pre-controller, valid requests would be rejected. Consider removing `maxLength` from input schemas (keep on response) or adding a description note about trim-then-validate semantics - `docs/openapi/schemas.yaml`
- suggestion: `BRAND_GUIDANCE_MAX_LENGTH = 4000` is defined independently in `src/controllers/brands.js:71` and `src/support/customer-config-mapper.js:45`. If this limit ever changes, both must be updated in lockstep. A single shared constant would eliminate that coordination risk (low urgency with only two usages today).
</details>

- Falsy coercion (`||`) in read mapping now uses `??` - preserves empty-string semantics from DB
- V1-to-V2 mapper now normalizes and truncates at 4000 chars via `normalizeBrandGuidance`
- `normalizeNullableText` now tags errors with `.status = 400` for defense-in-depth
- Controller now validates trimmed length (mirrors what storage persists), with test for the boundary case

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 44s | <b>Cost:</b> $5.67 | <b>Commit:</b> <code>82c6ee022ec80fb493969e121718372b2af73330</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2627, opened 2026-06-17
fix(preflight): construct custom-env ImsClient with IMS_SCOPE=system at the call site (SITES-43236 follow-up)

GeezerPelletier:
## Summary

Follow-up to [#2611](https://github.com/adobe/spacecat-api-service/pull/2611). Empirical test on the deployed spacecat-api-service dev Lambda after #2611 landed surfaced an IMS 400 on `getServiceAccessTokenV3` — caught by the fail-closed mint guard exactly as designed. Splunk traced it to the spacecat default IMS client (`aem-project-collab-service` per Vault `dx_mysticat`) lacking `IMS_SCOPE` provisioning for the v3 `client_credentials` grant.

Per ASO team Slack confirmation (2026-06-17): `aem-project-collab-service` is the correct client for spacecat-api-service S2S calls, and Adobe IMS mandates the `system` scope for v3 service-client tokens. Rather than wait for Vault provisioning ceremony, this PR hardcodes `IMS_SCOPE='system'` at the mint call site by constructing a custom-env `ImsClient` — mirrors the existing pattern in [`email-service.js`](https://github.com/adobe/spacecat-api-service/blob/main/src/support/email-service.js) and [`trial-users.js`](https://github.com/adobe/spacecat-api-service/blob/main/src/controllers/trial-users.js), both of which override IMS env at construction for their dedicated LLMO email IMS client.

## Source changes (`src/controllers/preflight.js`)

- Import `ImsClient` alongside the existing `retrievePageAuthentication`.
- Drop the `imsClient` destructure + constructor guard from `PreflightController` (no longer using the wrapper-wired `context.imsClient`).
- Replace `context.imsClient.getServiceAccessTokenV3()` with a custom-env `ImsClient.createFrom({ ...context, env: { ...context.env, IMS_SCOPE: 'system' } })` at the mint call site.
- Preserve all defensive guards from #2611:
  - Post-condition `hasText(access_token)` for SDK shape drift.
  - Structured error logging that passes the error object for stack-trace preservation.
  - Fail-closed posture in Phase 1 and Phase 2 (mint failure → `500 PREFLIGHT_INTERNAL_ERROR`).
  - Mint runs before `AsyncJob` / `Preflight` DB writes so a transient IMS failure doesn't orphan IN_PROGRESS records.

## Test changes (`test/controllers/preflight.test.js`)

- Remove the now-unused `imsClient` constructor guard test + ctx injections (the controller no longer requires `context.imsClient`).
- Update the `createPreflight` `beforeEach` to esmock `ImsClient.createFrom` returning the shared `mockImsClient` stub, so per-test assertions on `getServiceAccessTokenV3` keep working without change.
- Per-test esmocks that override `retrievePageAuthentication` now also stub `ImsClient.createFrom` — partial-mocking a module replaces unspecified exports with the real implementation, which would otherwise call out to real IMS endpoints during the mint.
- Remove `imsClient` sentinel injections from `createPreflightJob` / `getAllPreflights` / `getPreflightById` `beforeEach` blocks (no longer needed with the guard gone).

## Phase 2 follow-up (mystique-deploy side)

For the eventual hard-enforce flip on mystique-deploy ([PR #463's TODO comment](https://git.corp.adobe.com/experience-platform/mystique-deploy/pull/463)), Alina will need to:
- Add the `aem-project-collab-service` IMS client id to `allowedClientIDs` on the `/v1/preflight/analyze` CGW endpoint
- Flip `tokenType: user → service`
- Flip `optional: true → false`

Phase 1 works as-is once this PR deploys.

## Test plan

- [x] All 90 preflight controller tests pass locally
- [x] Constructor guard removal verified by passing tests that previously asserted on it being present (one removed test case + ctx-injection cleanup)
- [x] Custom-env mint path verified by all 5 IMS-related tests added in #2611 (header attach, both-headers, mint-fail 500, no-DB-records on mint fail, no-Mysticat-call on mint fail, plus the SDK-shape-drift `{}` payload test)
- [ ] Empirical: once this lands and the dev Lambda deploys, retry the original test call and confirm no more IMS 400

## Fixed Issues

- [SITES-43236](https://jira.corp.adobe.com/browse/SITES-43236) — reopened after the dev runtime IMS 400; this PR is the in-team unblock


src/controllers/preflight.js
```diff
@@ -505,7 +520,11 @@ function PreflightController(ctx, log, env) {
     // this a narrow availability cost in practice.
     let imsServiceToken;
     try {
-      const tokenPayload = await imsClient.getServiceAccessTokenV3();
+      const scopedImsClient = ImsClient.createFrom({
+        ...context,
+        env: { ...context.env, IMS_SCOPE: 'system' },
+      });
```
MysticatBot, line 530: issue (blocking): **Per-invocation `ImsClient` construction loses warm-Lambda token caching.**

The previous code used `context.imsClient` from the `imsClientWrapper` middleware, which maintained a singleton across warm Lambda invocations. The old comment (now removed) explicitly documented this: "The ImsClient caches the token in-memory on the wrapper instance, so warm Lambda invocations short-circuit the IMS round-trip."

The new code constructs a fresh `scopedImsClient` on every `createPreflight` call. Since the controller is instantiated per-request, every invocation now makes a fresh IMS round-trip even on warm Lambdas. For a transitional workaround this is tolerable, but it should be explicitly acknowledged.

**Fix (pick one):**
- (a) Hoist `scopedImsClient` to the `PreflightController` closure scope (constructed once when the controller is instantiated, since it only depends on `env` which is a constructor parameter). This preserves caching across requests within a warm Lambda.
- (b) Add a FIXME comment acknowledging the caching regression and link it to the Vault provisioning follow-up so the revert restores caching semantics.

Option (b) is sufficient if the team accepts the latency cost as part of the transitional state and the follow-up is tracked.

test/controllers/preflight.test.js
```diff
@@ -1167,9 +1147,13 @@ describe('Preflight Controller', () => {
 
       // SITES-46202: createPreflight no longer consults Configuration / TierClient
       // for eligibility — Mysticat owns that decision. The controller is mocked
-      // here only against AccessControlUtil (tenancy boundary) and the standard
-      // dataAccess stubs.
+      // here only against AccessControlUtil (tenancy boundary), the IMS client
```
MysticatBot, line 1156: issue (blocking): **No assertion that `ImsClient.createFrom` receives `IMS_SCOPE: 'system'` in the env.**

The entire purpose of this PR is to pass `IMS_SCOPE='system'` to the IMS client factory. The test mocks `createFrom` to return the stub unconditionally but never asserts the override was passed. If a future refactor accidentally drops the scope override, the tests still pass.

**Fix:** Make `createFrom` a proper stub and add an assertion in the happy-path IMS header test:
```js
const createFromStub = sandbox.stub().returns(mockImsClient);
// ... in esmock:
ImsClient: { createFrom: createFromStub },
// ... in assertion:
expect(createFromStub).to.have.been.calledWith(
  sinon.match({ env: sinon.match({ IMS_SCOPE: 'system' }) })
);
```

MysticatBot (changes requested):
Hey @GeezerPelletier,

**Verdict: Request changes** - two material concerns around caching regression and test coverage of the core fix.
_Changes: Replaces the wrapper-wired `context.imsClient` with a locally constructed `ImsClient` that hardcodes `IMS_SCOPE='system'` to unblock the v3 client_credentials grant (2 files)._

## Must fix before merge
1. [Important] Per-invocation `ImsClient` construction loses warm-Lambda token caching - `src/controllers/preflight.js:523` (details inline)
2. [Important] No test assertion verifying `IMS_SCOPE: 'system'` reaches `ImsClient.createFrom` - `test/controllers/preflight.test.js` (details inline)

<details>

- nit: 19-line comment block is disproportionate to the 4-line code change; a 3-5 line comment with a `SITES-43236` reference and FIXME tag would be more maintainable - `src/controllers/preflight.js:489`
- suggestion: Pass only what `ImsClient.createFrom` consumes (`{ log, env: { ...env, IMS_SCOPE: 'system' } }`) instead of spreading the full request context, to document the contract explicitly - `src/controllers/preflight.js:523`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 56s | <b>Cost:</b> $8.52 | <b>Commit:</b> <code>7f4faff06def1cb1ec70bc8477744df7b44f6a6b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

GeezerPelletier:
Thanks @MysticatBot — pushed `87cf0447` addressing both blocking items.

## Blocking 1 — Caching regression (per-invocation `ImsClient`)

Acknowledged in the existing Reversibility paragraph: revert restores singleton token caching; the per-invocation mint is the deliberate transitional cost.

Worth flagging the opposing dynamic — @clotton's review on #2611 raised the *opposite* concern: `getServiceAccessTokenV3()`'s cache isn't expiry-aware ([ims-client.js:197](https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-ims-client/src/clients/ims-client.js#L197) — it returns the cached token forever as long as `access_token` is non-empty, with no `expires_in` check). The singleton-caching path that the wrapper-wired client gave us would have meant warm Lambdas serving expired tokens past 3600s. The per-invocation mint here happens to sidestep that. Net trade: small per-request IMS round-trip for eliminated cache-staleness risk — for the transitional state, that's the right side to be on.

Option (a) — hoisting `scopedImsClient` to controller closure scope — was checked but doesn't help: [src/index.js:244](https://github.com/adobe/spacecat-api-service/blob/main/src/index.js#L244) instantiates `PreflightController` per request via `PreflightController(context, log, context.env)`, so closure-scope state has the same lifetime as request-scope state.

## Blocking 2 — `IMS_SCOPE: 'system'` assertion

Stub-ified `ImsClient.createFrom` and added the assertion in the happy-path test exactly as suggested. Now if a refactor drops the scope override, this test fails.

## Nits — pushing back on both

**"19-line comment is disproportionate to 4-line code change."** Disagreeing here. The code is 4 lines but the *why* spans four genuinely non-obvious facets: (a) why custom-env at all (missing Vault config — empirically discovered, not assumed), (b) why `'system'` specifically (Adobe IMS mandate per Slack thread), (c) contrast with `email-service.js` precedent so future readers don't mistake this for the canonical pattern when in fact those callers have a structurally different reason to do the same thing, (d) the revert path. Compressing this loses load-bearing context. Also worth noting: the suggested "FIXME tag" framing was [explicitly flagged by the reviewer](https://github.com/adobe/spacecat-api-service/pull/2627/commits/7f4faff0) as the anti-pattern this PR is moving *away* from (a TODO/FIXME implies tracked action that doesn't exist; the conditional "Reversibility" framing is the deliberate replacement).

**"Pass only `{ log, env }` instead of spreading full context."** Tighter explicit contract, agreed — but it would diverge from the established precedent. [`email-service.js`](https://github.com/adobe/spacecat-api-service/blob/main/src/support/email-service.js) does `ImsClient.createFrom({ ...context, env: emailEnv })` with the same full-spread + env-override shape. A future reader comparing our call site to the precedent expects them to look identical; tightening one creates an unmotivated asymmetry. Worth refactoring at the precedent level if at all (cross-call-site consistency), not just here.

Re-requesting review.

MysticatBot (approved):
Hey @GeezerPelletier,

**Verdict: Approve** - both prior blocking items resolved; pushback on caching and nits holds up on re-evaluation.
_Changes: Replaces the wrapper-wired `context.imsClient` with a locally constructed `ImsClient` that hardcodes `IMS_SCOPE='system'` to unblock the v3 client_credentials grant (2 files)._

- Token-caching regression acknowledged in comment; author's counter-argument (non-expiry-aware cache, per-request controller instantiation) holds
- `IMS_SCOPE: 'system'` contract assertion now present via `createFromStub` + `sinon.match`
- Comment length justified by 4 non-obvious facets and revert path documentation
- Full-spread `createFrom` call consistent with `email-service.js` precedent

<details>

- suggestion: Test context omits `env`, so the `{ ...context.env, IMS_SCOPE: 'system' }` spread produces just `{ IMS_SCOPE: 'system' }` in test. Adding `env: { IMS_SCOPE: 'something_else', OTHER: 'x' }` to the test context would verify the override wins over a pre-existing value, not just that it's present - `test/controllers/preflight.test.js:1168`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 49s | <b>Cost:</b> $6.84 | <b>Commit:</b> <code>87cf0447f1301e31b2f4569fca47fcca3b906ac9</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @GeezerPelletier,

**Verdict: Approve** - prior blocking findings resolved; no new issues on re-review.
_Changes: Replaces wrapper-wired `context.imsClient` with a locally constructed `ImsClient` that hardcodes `IMS_SCOPE='system'` to unblock the v3 client_credentials grant (2 files)._

<details>
<summary>Non-blocking (1): suggestions</summary>

- suggestion: Track the Vault provisioning (`IMS_SCOPE=system` for `aem-project-collab-service`) as a concrete Jira subtask under SITES-43236 so this transitional bypass does not become permanent by inertia - the in-code reversibility instructions are thorough but a backlog item ensures visibility
</details>

- IMS_SCOPE assertion now verifies `createFromStub` receives `{ env: { IMS_SCOPE: 'system' } }` via `sinon.match`
- Caching regression concern dropped: author demonstrated the cache is not expiry-aware (`ims-client.js:197` never checks `expires_in`) and the controller is instantiated per-request regardless

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 21s | <b>Cost:</b> $7.18 | <b>Commit:</b> <code>6756ec5b519d52cff25555c766bc1d97e5ebd453</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2631, opened 2026-06-18
fix(topics): guard topic_categories link against non-UUID categoryId and log SQLSTATE

solaris007:
## What

`createTopic` (`src/support/topics-storage.js`) upserts a `topic_categories` junction row from the request payload's `categoryId`, which is a `uuid` FK to `categories.id`. Two problems:

1. A non-UUID `categoryId` (a category slug / business-key sent by mistake) can never satisfy the FK and produced a **guaranteed PostgREST 400** — observed in prod as a burst of ~41 identical `POST /topic_categories ... 400` lines during a single org onboarding flow.
2. The junction error was swallowed with only `.message` logged, so the failures were **invisible** in `spacecat-api-service` (the topic create still returns 201). They only surfaced now that data-service ships its PostgREST access log to Splunk.

## Changes

- **Skip the junction upsert when `categoryId` is not a valid UUID** — warn instead of firing a doomed write. Uses `isValidUUID` from `@adobe/spacecat-shared-utils` (version-agnostic regex, so v7 category IDs validate correctly).
- **On a real junction failure, log the SQLSTATE (`.code`) and `.details`** so a recurring cause is diagnosable from logs — e.g. `23503` when the category row is not committed yet or belongs to another org (the onboarding race that's the most likely driver of the observed burst).

Topic creation remains **non-fatal** on junction errors (unchanged behavior).

## Not in scope

This does not change the onboarding flow that creates categories + topics in the same ~0.5s burst (the suspected race). The richer logging is deliberately the enabling step: it lets us confirm `23503` (race) vs `22P02` (slug) on the next occurrence before touching that flow.

## Testing

TDD. Two new failing tests written first (skip-on-invalid-UUID; SQLSTATE+details in the warn), then implementation. Existing tests that used non-UUID category fixtures (`'cat-uuid-123'`, `'bad-category-uuid'`, `'cat-id'`) were corrected to real UUIDs — production category IDs are always UUIDs.

- `test/support/topics-storage.test.js`: 32 passing (incl. `code=n/a` fallback when an error carries no SQLSTATE).
- `test/controllers/brands.test.js`: 274 passing (caller unaffected).
- `npx eslint` clean.

## Context

Surfaced by the data-service → Splunk (ECS FireLens) cutover analysis (SITES-46451 / SITES-45430). The `topic_categories` 400 flood was the only error-level access log in the new stream.

### DIFF
--- a/src/support/topics-storage.js
+++ b/src/support/topics-storage.js
@@ -145,7 +145,9 @@ export async function createTopic({
       // A non-UUID categoryId (e.g. a category slug / business-key) can never
       // satisfy the uuid FK to categories.id and would only produce a
       // guaranteed PostgREST 400. Skip the doomed write rather than emit it.
-      log?.warn(`Skipping topic_categories link for topic ${data.id}: categoryId "${categoryId}" is not a valid UUID`);
+      // The value is caller-controlled and unbounded — truncate it so a
+      // malformed payload cannot produce multi-KB log lines.
+      log?.warn(`Skipping topic_categories link for topic ${data.id}: categoryId "${String(categoryId).slice(0, 64)}" is not a valid UUID`);
     } else {
       const { error: junctionError } = await postgrestClient
         .from('topic_categories')
--- a/src/support/topics-storage.js
+++ b/src/support/topics-storage.js
@@ -10,7 +10,7 @@
  * governing permissions and limitations under the License.
  */
 
-import { hasText } from '@adobe/spacecat-shared-utils';
+import { hasText, isValidUUID } from '@adobe/spacecat-shared-utils';
 
 import { throwOnPgConstraintViolation } from './errors.js';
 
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @solaris007,

**Verdict: Approve** - clean, well-scoped fix with good test coverage.
_Changes: Guards the topic_categories junction upsert against non-UUID categoryId values and enriches junction-failure logging with SQLSTATE and details (2 files)._

<details>

- nit: The warn message at line ~148 interpolates the raw `categoryId` value without length-bounding it. A malformed categoryId from a caller could be an arbitrarily long string (URL, base64 blob) producing multi-KB log lines. Consider `String(categoryId).slice(0, 64)` or switching to structured log fields. - `src/support/topics-storage.js:148`
- nit: The new "skips when categoryId is not a valid UUID" test uses `id: 'uuid-bad-cat'` and `topic_id: 'bad-cat-topic'` in its fixture, which are not valid UUIDs. Cosmetic inconsistency with the other new test that uses proper UUID-formatted fixture values. - `test/support/topics-storage.test.js:337`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 7s | <b>Cost:</b> $5.01 | <b>Commit:</b> <code>c72559cff7ce9c5ed3904f50e140cee3adda8e90</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

solaris007:
Thanks @MysticatBot. On the two non-blocking nits:

1. **nit (unbounded categoryId in log line)** - good catch, fixed in 852f9f0c: the skip-path warn now truncates to `String(categoryId).slice(0, 64)`. Doubly relevant given this PR is about log volume. The failure-path warn is already bounded (the value passed `isValidUUID`, so ≤36 chars). Added a test asserting a 500-char value doesn't reach the log line.

2. **nit (fixture ids not UUIDs)** - declining. The `id`/`topic_id` fixture values are mock DB return values, and the whole file uses placeholder ids (`uuid-new`, `uuid-tc`, `uuid-code`, ...) - including the other new test. Only `categoryId` needs UUID format because that's what the new guard validates, and those values are valid UUIDs. Making just this fixture's id "real" would be more inconsistent with the file, not less.

PR #2643, opened 2026-06-19
feat(serenity): mirror each Semrush market as a SpaceCat Site

rainer-friederich:
## 1. Abstract
Every Semrush market now gets a resolvable SpaceCat Site mirroring its primary domain, linked to the owning brand via a `brand_sites` row tagged `type='serenity'`.

## 2. Reasoning
A Serenity brand is a domain-less shell (like its Semrush sub-workspace); the real domains live on its markets. Integrations and lookups that resolve a site by its base URL had nothing to find for these brands. Mirroring each market's domain as a Site restores base-URL resolvability and backwards compatibility without changing the brand model.

- Before: a Serenity brand had no Site, so a base-URL lookup of a market's domain resolved to nothing.
- After: `ensureMarketSite` idempotently ensures one Site per market domain (deliveryType `other`) and links it to the brand with a `brand_sites` row marked `type='serenity'`. Global `sites.base_url` uniqueness means at most one Site per domain; a brand whose markets span distinct domains owns several Sites.
- Wired into the sub-workspace-mode hooks only: brand creation (initial market), activation (activated markets), and market creation (that market). Best-effort and same-org guarded, so it never fails a live market. Sites/links are never auto-deleted; deleting a market leaves them in place.
- `syncBrandSites` now preserves `type='serenity'` (and NULL-type) rows so the delete-then-reinsert brand-edit rebuild no longer wipes these links, and a re-upsert cannot downgrade the marker.
- `mapDbBrandToV2` excludes `type='serenity'` rows from `urls[]` and `siteIds`, so they stay a pure backend linkage and the Brand V2 response is unchanged.
- Per-market `domain` is exposed on the markets API + OpenAPI; `docs/serenity.md` gains a "Site mirroring" section.

## 4. Required information
- Spec: docs/serenity.md "Site mirroring" section (design-of-record, updated in this PR)

- mysticat-data-service — depends-on: requires the migration allowing `brand_sites.type='serenity'`; the link insert is silently rejected by the CHECK constraint until that migration is deployed.
- project-elmo-ui — consumed: surfaces the new per-market `SerenityMarket.domain` field.

## 7. Test plan
(a) Local: exercise the Serenity through-API local e2e (activate brand -> create market -> deactivate) against the local stack and confirm each market domain resolves to a Site and a `brand_sites` row with `type='serenity'` exists for the brand. Run only after the data-service migration is applied to the local DB.
(b) Per-env: after the data-service migration deploys to a given env, create or activate a Serenity market and verify (1) a Site exists for the market's domain, (2) a `brand_sites` row tagged `type='serenity'` links it to the brand, (3) the brand V2 response still omits that domain from `urls[]`/`siteIds`, and (4) the markets API returns the market `domain`.

- mysticat-data-service (branch feat/semrush-project-site-entities, migration `brand_sites.type='serenity'`) — depends-on: MUST merge and deploy first; until then the link insert is silently rejected (migration-first rollout).
- project-elmo-ui PR 2078 — related: the UI consuming the new per-market domain.
- Order: deploy the data-service migration first, then merge this PR, then the elmo UI change.


src/support/brands-storage.js
```diff
@@ -304,7 +329,11 @@ async function syncBrandSites(organizationId, brandId, urls, postgrestClient, up
     brand_id: brandId,
     site_id: s.id,
     paths: pathsByBase.get(s.base_url) || [],
-    type: typeByBase.get(s.base_url) || null,
+    // A brand URL may resolve to the same site as a preserved market row. Keep
+    // that row tagged 'serenity' rather than downgrading it to the URL's type.
+    type: protectedSiteIds.has(s.id)
```
MysticatBot, line 396: issue (blocking): When a brand URL resolves to the same `site_id` as an existing serenity market row, `protectedSiteIds.has(s.id)` forces `type = SERENITY_BRAND_SITE_TYPE` on the re-inserted row. `mapDbBrandToV2` then excludes rows with this type from `urls[]` and `siteIds`.

The effect: a brand URL that previously appeared in the V2 response silently vanishes the moment a market is created for the same domain.

The `(brand_id, site_id)` uniqueness constraint means only one row can exist for the pair, so the type must serve both purposes. Consider:
- Skip inserting the brand-URL row entirely when the site is already protected (the serenity linkage already covers it, and the brand has no domain of its own per the domain model).
- Or accept the overlap explicitly: if `protectedSiteIds.has(s.id)` AND the brand URL list also references this site, keep type=serenity but also include the site in `mapDbBrandToV2` output (e.g. exclude only rows where `brand_urls` has no matching base_url).

The right choice depends on whether a brand URL that coincides with a market domain should appear in the brand response. The current code silently hides it without documenting this as intentional.

src/support/brands-storage.js
```diff
@@ -251,10 +260,26 @@ async function replaceChildRows(table, brandId, rows, onConflict, postgrestClien
...
+  // brand.urls, so the delete-all-then-reinsert below would wipe these links on
+  // every brand edit. Preserve them: collect the protected site ids first, exclude
+  // them from the delete, and keep their type from being downgraded on re-upsert
+  // (when a brand URL happens to resolve to the same site as a market).
+  const { data: protectedRows } = await postgrestClient
+    .from('brand_sites')
+    .select('site_id')
+    .eq('brand_id', brandId)
+    .eq('type', SERENITY_BRAND_SITE_TYPE);
+  const protectedSiteIds = new Set((protectedRows || []).map((r) => r.site_id));
```
MysticatBot, line 333: issue (blocking): The destructured `{ data: protectedRows }` does not check for an `error` return from PostgREST. If the SELECT fails (transient DB error, network issue), `protectedRows` is `null`, `(protectedRows || [])` produces an empty array, and `protectedSiteIds` is empty. The subsequent DELETE correctly skips serenity rows (the `.or` filter is right), but the re-INSERT uses the stale/empty `protectedSiteIds` to decide the type. If a brand URL resolves to a serenity-protected site, it gets `type=null` instead of `type='serenity'`, unprotecting that row for the next sync.

Fix: check for the error and throw (consistent with the existing delete-error pattern):
```js
const { data: protectedRows, error: selectError } = await postgrestClient
  .from('brand_sites').select('site_id')
  .eq('brand_id', brandId)
  .eq('type', SERENITY_BRAND_SITE_TYPE);
if (selectError) {
  throw new Error(`Failed to sync brand_sites: cannot determine protected rows: ${selectError.message}`);
}
```

MysticatBot (changes requested):
Hey @rainer-friederich,

**Verdict: Request changes** - one blocking concern about brand URL visibility when a brand URL and market domain resolve to the same site.
_Changes: mirrors each Semrush market as a SpaceCat Site entity with a `brand_sites` row tagged `type='serenity'`, preserving it through brand edits and excluding it from the brand V2 response (12 files)._

## Must fix before merge
1. [Important] Brand URL silently vanishes from V2 response when it resolves to the same site as a market domain - `src/support/brands-storage.js:329` (details inline)
2. [Important] `syncBrandSites` does not check for error on the protected-rows SELECT - a transient DB failure silently empties `protectedSiteIds`, causing the re-INSERT to downgrade serenity rows - `src/support/brands-storage.js:269` (details inline)

<details>

- nit: race window between `syncBrandSites` (SELECT then DELETE then INSERT) and concurrent `ensureMarketSite` could temporarily downgrade a serenity row's type; self-healing on next market write but worth a comment acknowledging the eventual-consistency guarantee - `src/support/brands-storage.js:275`
- suggestion: validate `brandDomain` format before passing to `ensureMarketSite` (hostname pattern in OpenAPI or runtime check) to prevent junk/internal hostnames from creating Site records - `src/support/serenity/site-linkage.js:68`
- nit: `ensureMarketSite` returns a non-null siteId when the existing site belongs to another org (no link created), indistinguishable from full success for callers - `src/support/serenity/site-linkage.js:85`
- nit: `brandDomainFromPayload(brandData)` re-derives the domain instead of capturing the earlier `brandDomain` in a shared scope - `src/controllers/brands.js:1486`
- nit: the `.or()` filter `type.is.null,type.neq.serenity` relies on PostgREST NULL semantics that are non-obvious; a one-line comment explaining why `is.null` is needed alongside `neq` would help future readers - `src/support/brands-storage.js:282`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 7s | <b>Cost:</b> $8.81 | <b>Commit:</b> <code>828dfaaf01e79dd36000a2b3d4baf34be2ac50f7</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Thanks for the review. Addressed in `fc2943dd` (the review ran on `828dfaa`; several non-blocking items were already fixed in the intervening `f6d2c170`/`a97155ce`). Also: the CI failure you noted is fixed in `a97155ce` (an activate mirror hook evaluated `brand.getOrganizationId()` outside `ensureMarketSite`'s try/catch → 500; now optional-chained).

Must fix before merge:
- [Important] Brand URL vanishes when it shares a site with a market domain (`brands-storage.js`) — fixed. `mapDbBrandToV2` now surfaces a `type='serenity'` row when the brand ALSO lists that exact domain as a brand URL, so it keeps `onboarded:true`/`siteId`. Pure market domains (not in `brand_urls`) stay excluded. New test: "surfaces a Serenity row when its domain is ALSO an explicit brand URL".
- [Important] `syncBrandSites` ignores the protected-rows SELECT error — fixed. Now throws (fails closed), consistent with the delete/upsert error pattern; a swallowed error would empty `protectedSiteIds` and downgrade serenity rows. Replaced the earlier "silently proceeds" test with a fail-closed one.

Non-blocking:
- nit: race-window comment (`brands-storage.js`) — done (added in `f6d2c170`): the SELECT→DELETE→INSERT vs concurrent `ensureMarketSite` downgrade window is now documented as doubly-rare + self-healing on next market write.
- suggestion: validate `brandDomain` before `ensureMarketSite` — partially done / rest declined. `ensureMarketSite` now normalizes via `hostnameFromUrlString` (added in `f6d2c170`), so unparseable/junk input is skipped. A stage/internal-hostname blocklist is declined as out of scope here: the market domain is provisioned to Semrush upstream, which is its source of truth.
- nit: cross-org `ensureMarketSite` returns a non-null siteId — done (`f6d2c170`): returns `null` now when the site belongs to another org (no link created).
- nit: `brandDomainFromPayload` re-derives the domain — done (`f6d2c170`): lifted to a shared `provisionedBrandDomain`.
- nit: `.or()` filter needs a comment on the `is.null`+`neq` semantics — done: the comment ("a bare `.neq` would skip NULLs; type IS DISTINCT FROM 'serenity'") is in place.

src/support/brands-storage.js
rainer-friederich: Fixed in `fc2943dd`: `mapDbBrandToV2` now surfaces a serenity row when its domain is also an explicit brand URL (keeps onboarded/siteId); pure market domains stay excluded. Test added.

src/support/brands-storage.js
rainer-friederich: Fixed in `fc2943dd`: now throws on the protected-rows SELECT error (fails closed), consistent with the delete/upsert handling. Test updated to assert the throw.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - both prior blocking findings are addressed with tests; remaining observations are non-blocking.
_Changes: mirrors each Semrush market as a SpaceCat Site entity with a `brand_sites` row tagged `type='serenity'`, preserving it through brand edits and excluding it from the brand V2 response (13 files)._

<details>

- nit: `ensureMarketSite` returns `siteId` on brand_sites upsert failure, which contradicts the JSDoc ("null when skipped/failed"); no caller depends on this today, but updating the return or the doc would keep the contract honest - `src/support/serenity/site-linkage.js:120`
- suggestion: the `ensureMarketSite` call in `createBrandForOrg` sits inside the try/catch that triggers workspace-release compensation; the never-throws contract holds today (outer catch-all in site-linkage.js), but a one-line comment at the call site like "MUST NOT throw - catch below would release a live workspace" would make the invariant explicit for future editors - `src/controllers/brands.js:1483`
</details>

- Brand URL vanishing when it shares a site with a market domain - `mapDbBrandToV2` now surfaces a serenity row when the domain is also an explicit brand URL
- `syncBrandSites` protected-rows SELECT error handling - now throws (fails closed), consistent with delete/upsert pattern
- Race-window comment on SELECT/DELETE/INSERT vs concurrent `ensureMarketSite` - documented
- `brandDomainFromPayload` re-derivation - lifted to shared `provisionedBrandDomain`
- Cross-org `ensureMarketSite` return value - now returns null when site belongs to another org
- `.or()` filter NULL semantics comment - added
- Domain normalization via `hostnameFromUrlString` - added

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 4s | <b>Cost:</b> $9.79 | <b>Commit:</b> <code>fc2943dd387c7cf53eb27785efb06a7410bdc443</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Both non-blocking nits addressed in `27b02fef` (leaving your approval standing — doc tweak + return-value honesty fix, no logic rework):
- `ensureMarketSite` now returns the site id only when the link is established; null otherwise (no client / failed link write). Tests updated.
- Added the explicit MUST-NOT-throw invariant comment at the `brands.js` call site.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - both post-approval nits addressed cleanly; no new concerns.
_Changes: mirrors each Semrush market as a SpaceCat Site entity with a `brand_sites` row tagged `type='serenity'`, preserving it through brand edits and excluding it from the brand V2 response (20 files)._

- `ensureMarketSite` now returns null when the link is not established (upsert failure or missing postgrest client), honoring the JSDoc contract that non-null means "linked"
- MUST-NOT-throw invariant documented at the `brands.js` call site
- IT seed data and end-to-end assertions wired for the serenity market-mirror exclusion (brand GET single + list, delivery-type filter)
- Data-service image bumped to v5.44.0 (required for `brand_sites` CHECK constraint)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 27s | <b>Cost:</b> $4.53 | <b>Commit:</b> <code>a4436387bd59bdd94d0208ddd0327824849751ab</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Re-requesting MysticatBot. The prior approval (`a4436387`) was clean, but four substantive commits have landed since and are not covered by it:

- `20a0b4a5` — optional prompt generation + all-or-nothing activation hardening
- `b7e726d3` — site-URL immutability guard + self-referential competitor filtering
- `643f59b6` — back the Semrush project transport with the typed client
- `dc6f7e15` — consume the published `@adobe/spacecat-shared-project-engine-client` instead of vendoring

These deserve a fresh review pass on current HEAD (`dc6f7e15`).

Note: the "Run Tests" check stays red until `@adobe/spacecat-shared-project-engine-client@1.1.1` publishes from https://github.com/adobe/spacecat-shared/pull/1706 ; the lockfile gets regenerated once that lands. That CI gate is orthogonal to this review.

rainer-friederich:
Ran a local multi-agent review of this PR (the hosted MysticatBot re-review crashed on `error_max_budget_usd` given the diff size). It surfaced no regressions on top of the prior approval, but a few quality/consistency items on this PR's own new surface. All of them — issues and recommendations — are now addressed in `5e28efc6`.

**Fixes**

- **Site-URL immutability guard now covers Semrush market mirrors.** `getBrandBySite` resolves a brand only via `brands.site_id`, but a serenity brand shell has no `site_id` — its market mirrors are linked solely through `brand_sites` (`type='serenity'`). The guard in `updateSite` therefore left every market-mirror site's `baseURL` editable, free to desync from its Semrush project. Added `isSemrushMarketMirrorSite()` and the guard now checks it alongside `getBrandBySite`.
- **`updateSite` no longer throws unmapped.** The brand-attachment lookup is wrapped so a transient PostgREST failure returns a mapped 500 instead of escaping the catch-less handler as an opaque error.
- **Stale activation docs corrected** (`docs/serenity.md`). The doc claimed empty `markets` and a missing `brandDomain` both return 400; the code now falls back to a single US/en project for empty markets and activates **sub-workspace-only** (200, empty `markets[]`) when no domain resolves. Documented the sub-workspace-only shape and the single-domain site-link assumption.

**Hardening / recommendations**

- **SSRF guard on Site mirroring.** `ensureMarketSite` now refuses non-public hostnames (localhost, loopback, link-local, RFC1918, `*.internal`, bare IPs) before `Site.create`, via a new `isPublicHostname()` chokepoint shared by the brand-create / market-create / activate callers (`Site.create` itself only validates the scheme).
- **`listProjects` fan-out collapsed.** A brand edit now lists the sub-workspace's projects once and shares the result across the URL / competitor / alias syncs (optional `prefetchedProjects` arg, internal fallback preserved), removing up to 3 redundant upstream round-trips per edit.
- **`getImsUserTokenStrict` fails closed** — a missing/non-standard `authInfo` is treated as not-IMS and refused, rather than falling through to proxy an unverified bearer upstream.
- Deterministic `brand_name` fallback in the alias sync (never `undefined`).

**Tests** added for: the market-mirror guard block and the lookup-error 500; `isSemrushMarketMirrorSite`; `isPublicHostname`; `ensureMarketSite` SSRF refusal; benchmark-only alias drift and the `updateBenchmark`-throws path; a full re-activate of an already-active brand (200); and pre-fetched-listing reuse across all three across-markets syncs. Removed one tautological constant assert.

`npm run type-check`, `lint`, `docs:lint`, and the full set of affected unit tests (1211) are green locally.

src/support/brands-storage.js
rainer-friederich: Resolved in `3c7fa668`. The protected-rows read now destructures `error` and throws `Failed to sync brand_sites: cannot read protected rows: …` before `protectedSiteIds` is built — fail-closed, matching the other reads in this function.

src/support/brands-storage.js
rainer-friederich: Confirmed handled: a brand URL colliding with a preserved market site keeps the `serenity` tag via `protectedSiteIds.has(s.id)` on the upsert, so it is not downgraded to the URL's type. (The narrow concurrent-write downgrade is documented and self-healing.)

rainer-friederich:
Ran a local multi-agent re-review of the commits landed since the last MysticatBot approval (the hosted re-review crashed on `error_max_budget_usd` again given the diff size). Team: staff engineer, distinguished engineer, security researcher, QA, project-conventions. Scope was the delta — the `address local review` fix commit plus the coverage-restore tests — not the already-approved base.

**Prior blocking threads: both verified resolved** (replied inline). The protected-rows PostgREST read now fails closed, and the site_id-collision case keeps its `serenity` tag.

**New findings — all fixed in `3c7fa668`:**

Important
- `syncBrandSites` had a second fail-open read: the `sites` lookup didn't check the PostgREST `error`, so a transient failure silently no-op'd `brand_sites` linkage. Now throws, matching the protected-rows read 50 lines above (the same class the prior thread flagged).
- The new `isPublicHostname` SSRF guard lived in `src/support/url-utils.js`, which carried no `// @ts-check` and wasn't in the tsconfig `include` — so security-relevant code was invisible to the type-check gate. Added both; documented that it assumes a `URL`-normalized hostname.
- The shared `listProjects` prefetch had moved outside the post-commit diagnostic try/catch in `brands.js`, dropping the workspace breadcrumb on the hardest-to-debug path. Moved back inside.
- `getImsUserTokenStrict`'s new fail-closed branch (missing `getType` to 401) had only transitive coverage. Added direct unit tests.

Minor / cleanup
- Completed the `@param` tags on the three `*AcrossMarkets` sync functions and extracted the duplicated prefetch-resolve into `resolve-projects.js`.
- `beforeEach` reset for the market-mirror stub (was an order-fragile end-of-test reset); un-nested the brand-provisioning defensive describe; extra reserved-TLD cases.
- Documented in `docs/serenity.md` that a market-mirror `baseURL` is immutable at the API.

On the SSRF guard, the security pass empirically confirmed the upstream WHATWG `URL` normalization neutralizes the decimal/hex/octal/unicode IP-literal evasions for every current call path — no exploitable bypass. Consolidating `isPublicHostname` with the existing hardened `isSafeDomain` (in plg-onboarding) is left as a tracked follow-up since it touches a different feature.

Coverage gate (90% c8) restored — the `ci/build` failure (exit 15, from an untested provisioning error path) is fixed; full suite green locally (12,641 passing, type-check clean).

rainer-friederich (commented):
Hey @rainer-friederich,

**Verdict: Comment** - no blocking findings remain; the serenity delta since the last approval is correct, type-checked, and well-tested, and the one fail-open read this pass surfaced has been fixed in `98e3ca54`. (Posted as a comment rather than an approval because GitHub blocks self-approval on your own PR.)
_Changes: market-mirror URL guard + SSRF host check, cross-market project fan-out, and coverage-restoring tests (re-review of the serenity commits since the 2026-06-22 approval)._

This was a re-review focused on the post-approval serenity delta (`5e28efc6`, `97330fc0`, `3c7fa668`, `69bb53de`), not the already-approved base. A 5-specialist pass (staff, architect, conventions, security, QA) verified the prior blocking threads and the recent fixes, and surfaced one new data-integrity issue, now resolved.

<details>
<summary>Non-blocking (3): tracked follow-ups</summary>

- suggestion: `isPublicHostname` could normalize its input via `new URL()` internally so it is self-defending rather than relying on every caller passing `hostnameFromUrlString` output - `src/support/url-utils.js` (documented as an assumption for now)
- out of scope, worth tracking: `isPublicHostname` overlaps the hardened `isSafeDomain` in plg-onboarding - two SSRF host-classifiers that could drift; consolidation deferred as cross-feature
- out of scope, worth tracking: a competitor edit lists Semrush projects twice (`resolveReservedDomains` pre-write, then the post-commit sync) - `src/controllers/brands.js`; pre-existing, not introduced by this delta
</details>

- `updateBrand` baseSiteId-immutability read now fails closed - it was fail-open, silently re-pointing the immutable `site_id` on a transient PostgREST error (the LLMO-5556 regression the guard exists to prevent) - `98e3ca54`
- `syncBrandSites` sites read now fails closed (was a second fail-open of the same class)
- Protected-rows read fails closed, and a brand-URL/site_id collision keeps the `serenity` tag - the two prior MysticatBot blocking threads, verified resolved
- SSRF guard `isPublicHostname` is now inside the type-check gate (`// @ts-check` + tsconfig include); empirically verified the upstream `URL` normalization neutralizes decimal/hex/octal/unicode IP-literal evasions on all current call paths
- `@param` gaps closed on the serenity sync functions; duplicated prefetch-resolve extracted into `resolve-projects.js` with a standalone unit test; coverage gate (90% c8) restored

rainer-friederich:
Addressed two of the three non-blocking follow-ups from the re-review (commit 10cf644f4):

- **`isPublicHostname` is now self-defending.** It canonicalizes its own input through the WHATWG `URL` normalization (via `hostnameFromUrlString`) rather than relying on every caller to pre-normalize. Decimal/hex/octal IP-literal evasions (e.g. `2130706433` → `127.0.0.1`) and full-URL inputs are folded to their bare-host form and caught by the existing IPv4/IPv6/single-label/reserved-TLD guards; unparseable input fails closed. Added tests for the evasion forms, URL input, bracketed IPv6, and the fail-closed path. Still a syntactic guard (the `nip.io`-class wildcard-DNS caveat is unchanged and documented).

- **Single project listing on a competitor edit.** `brands.js` previously listed the workspace's projects twice on a competitor edit — once in the pre-write self-reference guard and again in the post-commit re-sync. The pre-write guard now lists once via `resolveProjects` and stashes the result; the post-commit sync reuses it (the workspace is never re-pointed across the row write, confirmed against `updateBrand`), re-listing only when the guard didn't run. The now-unused `resolveReservedDomains` helper and its tests were removed.

The third item — consolidating `isPublicHostname` with the hardened `isSafeDomain` in plg-onboarding — is left as the documented cross-feature follow-up; it spans a different module and isn't in this PR's scope.

Coverage: `url-utils` 100% lines / 96% branch, `competitor-benchmarks` 100%; both new `brands.js` branches exercised. type-check + lint clean.

rainer-friederich (commented):
Hey @rainer-friederich,

**Verdict: Comment** - clean re-review of the latest commits; zero blocking findings, and the two non-blocking polish nits this pass surfaced are already fixed in 710d5bb9d. (Posted as a comment rather than an approval because GitHub blocks self-approval on your own PR; the computed state is approve.)
_Changes: re-review of the post-approval serenity follow-ups - self-defending SSRF host check, single Semrush project listing on a competitor edit, and this pass's two polish fixes (3 files in the latest commit)._

A 5-specialist pass (staff, architect, conventions, security, QA) reviewed the delta since the last review (commits 10cf644f4 and 710d5bb9d). All five returned ready-to-merge with no Critical or Important findings.

The security pass empirically tested `isPublicHostname` against 47 strings across 8 evasion families with a direct old-vs-new behavioral diff: the self-defending normalization is a strict narrowing - zero widening, fail-closed intact. `@`-confusion (`evil.com@127.0.0.1`), decimal/hex/octal IP literals, bracketed/bare IPv6, and whitespace/control chars all reject correctly, and no legitimate domain is over-rejected. The `prefetchedProjects` reuse was independently verified safe by three reviewers: `updateBrand` builds its patch from a strict allow-list and never re-points `semrush_workspace_id`, so the cached listing stays keyed to the same immutable workspace. QA confirmed 100% coverage on the changed source, both new branches exercised, and the `calledOnceWith` assertion genuinely proves single-listing.

## Polish applied this pass (710d5bb9d)
- The post-commit sync now threads the pre-write competitor-guard listing through `resolveProjects`' purpose-built `prefetchedProjects` param instead of a hand-rolled `!== null` ternary, keeping the reuse-vs-list decision in one place (staff + architect).
- `isPublicHostname`'s parameter renamed `hostname` to `hostnameOrUrl` to match that it now accepts and canonicalizes a full URL or bare host (conventions).

Two further nits were considered and intentionally not actioned: the double-normalization on the `ensureMarketSite` path is intentional and idempotent (leave as-is), and a bracketed-IPv6-with-port test case would be a trivial superset of an already-covered path.

Note: `ci / build` is pending on the latest commit (`type-check` passed); the equivalent gate passes locally.

PR #2646, opened 2026-06-19
fix(prompts): deterministic 500 on POST /prompts — dedup + error classification + collation fix

aliciadriani:
## Summary

POST `/prompts` was returning HTTP 500 for 7 paid-site brandalf-secondaries restores, blocking them indefinitely. Root cause was a chain of three bugs in `upsertPrompts` that together converted a DB constraint violation into an untyped error → always 500.

- [A `b8f5375d`] Classify `23505` INSERT constraint violation as 409 not 500
- [B `650f4567`] Intra-batch dedup of `toInsert` by `(lower(text), sorted_regions)` before bulk INSERT
- [`38ecc0a8`] Test: deterministic distinct-topic tie-break (input-order invariance proof)
- [C `8150ff81`] Lowercase `getKey` text component to match DB `lower(text)` collation

## Root cause

Migration `20260530000003` added `uq_prompt_text_region_per_brand` — a partial unique index on `(brand_id, lower(text), sorted_regions(regions)) WHERE status = 'active'`. For new brands (`existingByKey` empty), V1 configs with cross-topic duplicate prompt text generate two identical `(text, regions)` rows in `toInsert`. The bulk INSERT violates the constraint. PostgREST returns a plain `{ code: '23505' }` error object which had no `.status` set, so `createErrorResponse` always fell through to `internalServerError` → **HTTP 500**.

Three compounding issues:

| # | Gap | Effect |
|---|-----|--------|
| 1 | `upsertPrompts` didn't dedup within `toInsert` | Both rows hit INSERT → constraint fires |
| 2 | `23505` not classified as typed error | `error.status` absent → always 500 |
| 3 | `getKey` used case-sensitive text | Case-variant re-runs would miss `existingByKey` and re-trigger constraint |

## What each commit does

**A `b8f5375d` — classify `23505` INSERT constraint violation as 409 not 500**

Calls `throwOnPgConstraintViolation` on the INSERT error path; maps `23505` → `{ status: 409 }`. Updates `createPromptsByBrand` catch to log at `warn` (not `error`) for 409. Means callers can distinguish a duplicate-input problem from a server fault.

**B `650f4567` — intra-batch dedup of `toInsert` by `(lower(text), sorted_regions)`**

After the `for` loop fully populates `toInsert`, deduplicate before INSERT. Deterministic tie-break: sort `(topic_id, prompt_id)` asc, keep first. Each dropped row is logged at `console.warn` with `brand_id`, `text_hash` (SHA-256 first 12 hex), `winning_prompt_id`/`topic_id`, `dropped_prompt_id`/`topic_id`. Dropped entries removed from `processed` so counts are honest.

The dropped topic association is **by design** (DROP+LOG, not MERGE): brandalf BP executes by brand not topic filter, so the association is latent. MERGE would require direct writes to `topic_prompts`, racing the `sync_prompt_junction_tables` trigger.

**`38ecc0a8` — test: deterministic distinct-topic tie-break**

Two prompts, same `(lower(text), sorted_regions)`, distinct `topic_id` UUIDs (`T_ALPHA < T_BETA`). Asserts: (a) one row to INSERT, (b) `topic_id = T_ALPHA`, (c) drop-log names `winning_topic_id=T_ALPHA` and `dropped_topic_id=T_BETA` (correct assignment), (d) `[beta,alpha]` input yields same winner.

**C `8150ff81` — lowercase `getKey` text to match `lower(text)` constraint**

`getKey` used case-sensitive text — `"Hello World"` ≠ `"hello world"` — so case-variant re-posts missed `existingByKey`, fell through to `toInsert`, and re-triggered the constraint. **Blast radius: changes routing for any case-variant text re-post repo-wide** (not only the 7 dark sites).

## Scope and limits

- Does **not** re-run `restore.py` — the 7 dark sites stay dark until merge + deploy + gated restore.py re-run.
- Does **not** undo existing DB state from previous failed restores.

## Testing

Full suite: **12,392 passing, 0 failing**.

| Test | File | Validates |
|------|------|-----------|
| throws typed 409 on 23505 | `prompts-storage.test.js` | Commit A — error classification |
| createPromptsByBrand returns 409 + warn not error | `brands.test.js` | Commit A — controller catch |
| deduplicates duplicate text+regions in toInsert | `prompts-storage.test.js` | Commit B — dedup fires |
| picks winner by (topic_id, promptId) asc regardless of input order | `prompts-storage.test.js` | Tie-break determinism + drop-log fidelity |
| routes case-variant text to update not insert | `prompts-storage.test.js` | Commit C — collation fix |

## Related

- Epic: SITES-39521 (Mysticat)
- DRS restore.py re-run for the 7 dark sites gated on merge + deploy.

## Known limitation / follow-up

`getKey` lowercases each region value before sorting (`['US'] → ['us']`), but the DB `sorted_regions()` function only sorts without lowercasing. For canonical-lowercase regions — the controlled vocabulary used throughout — this has no effect. If any row or source config ever carries a non-lowercase region string, `getKey` would false-positive match two rows the DB treats as distinct, silently routing an INSERT to UPDATE instead. This is pre-existing, out of scope for this PR, and in the conservative (non-500) direction. Tracked in [LLMO-XXXX](https://jira.corp.adobe.com/browse/LLMO-XXXX). **Reviewer: please confirm whether region values are ever non-canonical case in the data or in V1 source configs.**

### DIFF
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -492,7 +492,7 @@ function BrandsController(ctx, log, env) {
       return createResponse({ created, updated, prompts: outPrompts }, 201);
     } catch (error) {
       if (error?.status === 409) {
-        log.warn(`Prompt unique-constraint conflict for brand ${brandId}: ${error.message}`);
+        log.warn(`Prompt unique-constraint conflict for brand ${brandId} (org ${spaceCatId}): ${error.message}`);
         return createErrorResponse(error);
       }
       log.error(`Error creating prompts for brand ${brandId}:`, error);
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -491,6 +491,10 @@ function BrandsController(ctx, log, env) {
 
       return createResponse({ created, updated, prompts: outPrompts }, 201);
     } catch (error) {
+      if (error?.status === 409) {
+        log.warn(`Prompt unique-constraint conflict for brand ${brandId}: ${error.message}`);
+        return createErrorResponse(error);
+      }
       log.error(`Error creating prompts for brand ${brandId}:`, error);
       return createErrorResponse(error);
     }
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -15,6 +15,7 @@ import crypto from 'node:crypto';
 import { hasText, isValidUUID } from '@adobe/spacecat-shared-utils';
 
 import { classifyIntents } from './intent-classifier.js';
+import { throwOnPgConstraintViolation } from './errors.js';
 import { INTENT_VALUES, normalizeIntent } from './intent.js';
 
 // Re-exported for backward compatibility — `normalizeIntent`/`INTENT_VALUES` now
@@ -762,6 +763,9 @@ export async function upsertPrompts({
         .select(),
     );
     if (error) {
+      throwOnPgConstraintViolation(error, {
+        23505: { status: 409, message: 'A prompt with the same text and region already exists for this brand.' },
+      });
       throw new Error(`Failed to insert prompts: ${error.message}`);
     }
     created = inserted?.length ?? toInsert.length;
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -719,6 +719,45 @@ export async function upsertPrompts({
     }
   }
 
+  // Guard against uq_prompt_text_region_per_brand: deduplicate toInsert by
+  // (lower(text), sorted_regions) before the bulk INSERT. For a new brand
+  // existingByKey is empty, so cross-topic text collisions all land here.
+  // Deterministic tie-break: sort by (topic_id, prompt_id) asc, keep first.
+  // Each drop is logged (warn) with a text hash — auditable without echoing
+  // customer data. Dropped entries are removed from processed so counts stay
+  // honest.
+  if (toInsert.length > 1) {
+    const dedupKey = (row) => `${String(row.text || '').toLowerCase()}:${[...(row.regions || [])].sort().join(',')}`;
+    const sortedForDedup = [...toInsert].sort((a, b) => {
+      const tCmp = String(a.topic_id ?? '').localeCompare(String(b.topic_id ?? ''));
+      return tCmp !== 0 ? tCmp : String(a.prompt_id).localeCompare(String(b.prompt_id));
+    });
+    const winnerByKey = new Map();
+    const droppedIds = new Set();
+    for (const row of sortedForDedup) {
+      const key = dedupKey(row);
+      if (winnerByKey.has(key)) {
+        droppedIds.add(row.prompt_id);
+        // eslint-disable-next-line no-console
+        console.warn('[upsertPrompts] dedup-drop', {
+          brand_id: row.brand_id,
+          text_hash: crypto.createHash('sha256').update(row.text || '').digest('hex').slice(0, 12),
+          dropped_prompt_id: row.prompt_id,
+          dropped_topic_id: row.topic_id ?? null,
+          winning_prompt_id: winnerByKey.get(key).prompt_id,
+          winning_topic_id: winnerByKey.get(key).topic_id ?? null,
+        });
+      } else {
+        winnerByKey.set(key, row);
+      }
+    }
+    if (droppedIds.size > 0) {
+      const keep = (r) => !droppedIds.has(r.prompt_id);
+      toInsert.splice(0, toInsert.length, ...toInsert.filter(keep));
+      processed.splice(0, processed.length, ...processed.filter(keep));
+    }
+  }
+
   // Best-effort intent classification for prompts that arrived WITHOUT an
   // intent (typically human-added). Pipeline prompts already carry a
   // normalized intent and are skipped. Failures leave intent null; they MUST
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -660,7 +660,7 @@ export async function upsertPrompts({
 
   const getKey = (p) => {
     const norm = (p.regions || []).map((r) => String(r).toLowerCase()).sort();
-    return `${String(p.prompt || p.text || '').trim()}:${norm.join(',')}`;
+    return `${String(p.prompt || p.text || '').trim().toLowerCase()}:${norm.join(',')}`;
   };
 
   const existingById = new Map((existing || []).map((p) => [p.prompt_id, p]));
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -725,9 +725,13 @@ export async function upsertPrompts({
   // Deterministic tie-break: sort by (topic_id, prompt_id) asc, keep first.
   // Each drop is logged (warn) with a text hash — auditable without echoing
   // customer data. Dropped entries are removed from processed so counts stay
-  // honest.
+  // honest. Guard is > 1: a single-row batch cannot collide with itself.
   if (toInsert.length > 1) {
-    const dedupKey = (row) => `${String(row.text || '').toLowerCase()}:${[...(row.regions || [])].sort().join(',')}`;
+    const dedupKey = (row) => {
+      const t = String(row.text || '').trim().toLowerCase();
+      const r = [...(row.regions || [])].map((x) => String(x).toLowerCase()).sort().join(',');
+      return `${t}:${r}`;
+    };
     const sortedForDedup = [...toInsert].sort((a, b) => {
       const tCmp = String(a.topic_id ?? '').localeCompare(String(b.topic_id ?? ''));
       return tCmp !== 0 ? tCmp : String(a.prompt_id).localeCompare(String(b.prompt_id));
--- a/test/controllers/brands.test.js
+++ b/test/controllers/brands.test.js
@@ -919,6 +919,57 @@ describe('Brands Controller', () => {
       expect(response.status).to.equal(500);
     });
 
+    it('createPromptsByBrand returns 409 and logs at warn (not error) when INSERT fails with 23505', async () => {
+      const th = (v) => ({ then: (resolve) => resolve(v), catch: () => th(v) });
+      mockDataAccess.services.postgrestClient.from = sandbox.stub().callsFake((table) => {
+        if (table === 'prompts') {
+          return {
+            select: () => ({
+              eq: () => ({
+                eq: () => ({
+                  ...th({ data: [], error: null }),
+                  in: () => th({ data: [], error: null }),
+                }),
+              }),
+            }),
+            insert: () => ({
+              select: () => th({
+                data: null,
+                error: {
+                  code: '23505',
+                  message: 'duplicate key value violates unique constraint "uq_prompt_text_region_per_brand"',
+                },
+              }),
+            }),
+            update: () => ({ eq: () => th({ error: null }) }),
+          };
+        }
+        const chain = {
+          select: sandbox.stub().returnsThis(),
+          eq: sandbox.stub().returnsThis(),
+          maybeSingle: sandbox.stub().resolves({ data: { id: BRAND_UUID }, error: null }),
+        };
+        if (table === 'llmo_customer_config') {
+          chain.maybeSingle = sandbox.stub()
+            .resolves({ data: { config: { customer: { brands: [] } } }, error: null });
+        }
+        return chain;
+      });
+      loggerStub.warn.resetHistory();
+      loggerStub.error.resetHistory();
+
+      const response = await brandsController.createPromptsByBrand({
+        ...context,
+        params: { spaceCatId: ORGANIZATION_ID, brandId: BRAND_UUID },
+        data: [{ id: 'p-1', prompt: 'Synthetic test prompt', regions: ['us'] }],
+        dataAccess: mockDataAccess,
+      });
+
+      expect(response.status).to.equal(409);
+      expect(loggerStub.warn).to.have.been.called;
+      expect(loggerStub.error).to.not.have.been.called;
+    });
+
     it('updatePromptByBrandAndId returns 200 when prompt updated', async () => {
       const response = await brandsController.updatePromptByBrandAndId({
         ...context,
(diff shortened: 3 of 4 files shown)


src/support/prompts-storage.js
```diff
@@ -718,6 +719,45 @@ export async function upsertPrompts({
     }
   }
```
MysticatBot, line 720: issue (blocking): `dedupKey` does not apply the same normalization as `getKey` - it is missing both `.trim()` and region lowercasing.

`getKey` (line 661) does `.trim().toLowerCase()` on text and `.map(r => String(r).toLowerCase())` on regions. `dedupKey` does only `.toLowerCase()` on text (no trim) and `.sort().join(',')` on regions (no lowercasing).

This means a batch containing prompts with whitespace-variant text or mixed-case regions would not be deduped, and the INSERT would fire the DB constraint. The 409 fallback prevents a 500, but the dedup was meant to prevent the constraint fire entirely.

Fix:
```javascript
const dedupKey = (row) => `${String(row.text || '').trim().toLowerCase()}:${[...(row.regions || [])].map((r) => String(r).toLowerCase()).sort().join(',')}`;
```

This aligns `dedupKey` with `getKey` so both fences use the same normalization.

MysticatBot (changes requested):
Hey @aliciadriani,

**Verdict: Request changes** - one normalization gap in the dedup logic.
_Changes: fixes deterministic HTTP 500 on POST /prompts by adding intra-batch dedup, classifying 23505 as 409, and aligning getKey collation with the DB constraint (4 files)._

## Must fix before merge

1. [Important] `dedupKey` normalization inconsistent with `getKey` - `src/support/prompts-storage.js:720` (details inline)

<details>

- nit: `console.warn` used for dedup-drop logging instead of a structured logger - `src/support/prompts-storage.js:742`. Follows existing pattern in the file (lines 271, 297) but this is a new operational signal worth routing through structured logging. Consider threading an optional `log` parameter (defaulting to console) as `findPromptsBlockingRegionRemoval` already does at line 1016.
- suggestion: add a test with 3+ duplicates in a single batch to prove the Map-based dedup handles N>2 correctly - `test/support/prompts-storage.test.js`.

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 4s | <b>Cost:</b> $7.65 | <b>Commit:</b> <code>8150ff81c4938f3a7729d35434d083398b2b9fb5</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/prompts-storage.js
```diff
@@ -718,6 +719,45 @@ export async function upsertPrompts({
...
 
+  // Guard against uq_prompt_text_region_per_brand: deduplicate toInsert by
+  // (lower(text), sorted_regions) before the bulk INSERT. For a new brand
+  // existingByKey is empty, so cross-topic text collisions all land here.
+  // Deterministic tie-break: sort by (topic_id, prompt_id) asc, keep first.
+  // Each drop is logged (warn) with a text hash — auditable without echoing
+  // customer data. Dropped entries are removed from processed so counts stay
+  // honest.
+  if (toInsert.length > 1) {
```
aliciadriani, line 729: **Nit:** Worth a one-liner: `// single-row batch can't be a duplicate of itself`. Also note: this guard only covers intra-batch duplicates. A single-row batch that conflicts with an *existing DB row* (e.g. a race condition) is still handled by `throwOnPgConstraintViolation` below — the two mechanisms are complementary.

aliciadriani (commented):
## Review: fix/prompts-upsert-dedup-500

### Summary

This PR fixes a deterministic HTTP 500 on `POST /prompts` for new brands whose V1 config contains cross-topic prompts with duplicate text. Three tightly scoped commits address it at separate layers: A surfaces any residual 23505 as a typed 409 instead of an opaque 500; B prevents 23505 from firing in the first place via intra-batch dedup (DROP+LOG, deterministic tie-break); C aligns `getKey`'s text comparison to `lower(text)` so case-variant re-posts route to UPDATE not INSERT. Logic is correct, commit separation is clean, and the DROP+LOG design correctly avoids racing the `sync_prompt_junction_tables` trigger. One coverage gap stands out and should be addressed before merge.

---

### Must Fix

_None blocking merge._

---

### Should Fix

**1. Dedup drop path has zero test coverage (lines 740–749, 755–757 in `prompts-storage.js`)**

Coverage report shows those exact lines as uncovered. The `if (winnerByKey.has(key))` branch — `droppedIds.add`, the `console.warn` drop-log, and both `splice` mutations — never fires in any test. The tests assert on return values and INSERT call counts, but the actual mechanism that removes the duplicate row is untested. This means a subtle bug (e.g. wrong field name in `dedupKey`, a Map key mismatch) could silently pass all tests while still sending duplicates to INSERT in production.

Fix: add a dedicated unit test where `toInsert` has two rows with the same `dedupKey` result (same `row.text` after lowercasing, same `regions`), verify `droppedIds.size === 1`, and assert the `console.warn` spy was called with the correct `dropped_prompt_id` and `winning_prompt_id`.

**2. `dedupKey` and `getKey` apply different normalization to regions**

See inline comment on line 730.

**3. `console.warn` bypasses the structured logger**

See inline comment on line 742.

---

### Nits

- The `if (toInsert.length > 1)` threshold is self-evident to the author but worth a one-liner — a reader skimming the block might not immediately see why it isn't `> 0`. See inline on line 729.
- The 409 message (`"A prompt with the same text and region already exists for this brand."`) doesn't hint at cross-topic attribution, which is the common trigger. Minor: `"...already exists for this brand (same text across multiple topics)"` would help callers self-diagnose.
- `log.warn` in `brands.js` logs `brandId` but drops `organizationId`. Adding it makes Coralogix correlation faster.

---

### What's Good

- **DROP+LOG design is exactly right.** `upsertPrompts` makes zero direct junction writes; using MERGE + direct `topic_prompts` writes would have raced the `sync_prompt_junction_tables` trigger. DROP+LOG sidesteps that entirely.
- **Tie-break is deterministic and input-order invariant.** `(topic_id, prompt_id)` asc sort guarantees the same winner regardless of batch order — the tie-break test correctly verifies this with two feed orderings.
- **SHA-256 hash for the drop log** (`text_hash`) is privacy-conscious: the dedup reason is auditable without exposing customer prompt content.
- **`throwOnPgConstraintViolation` correctly handles the residual single-row case.** Even after dedup, a race condition (concurrent upsert inserting the same row between SELECT and INSERT) now produces a 409 not a 500.
- **Commit A/B/C are surgically scoped.** No refactoring, no unrelated changes. Each commit is independently reviewable and individually revertable.
- **12,392 tests passing, 0 failing.**

src/support/prompts-storage.js
```diff
@@ -718,6 +719,45 @@ export async function upsertPrompts({
...
+    const dedupKey = (row) => `${String(row.text || '').toLowerCase()}:${[...(row.regions || [])].sort().join(',')}`;
```
aliciadriani: **Should Fix:** `dedupKey` sorts regions but does **not** lowercase them (`[...(row.regions || [])].sort()`), while `getKey` above lowercases regions (`(p.regions || []).map(r => String(r).toLowerCase()).sort()`). The two key functions apply different normalization to the same field:

- `getKey` → `sort+lowercase` (over-normalizes vs DB)
- `dedupKey` → `sort-only` (matches DB `sorted_regions()`)

If a batch has one row with `regions: ['US']` and another with `regions: ['us']`, `dedupKey` treats them as distinct keys (no dedup) while `getKey` would treat them as the same row. In practice regions arrive pre-lowercased, but the inconsistency is a latent footgun. Both functions should agree on whether they lowercase regions — and `dedupKey`'s sort-only approach is actually more correct (it matches the DB constraint). Consider aligning `getKey`'s region normalization to sort-only in a follow-up (tracked as LLMO-XXXX).

src/support/prompts-storage.js
```diff
@@ -718,6 +719,45 @@ export async function upsertPrompts({
...
+      return tCmp !== 0 ? tCmp : String(a.prompt_id).localeCompare(String(b.prompt_id));
+    });
+    const winnerByKey = new Map();
+    const droppedIds = new Set();
+    for (const row of sortedForDedup) {
+      const key = dedupKey(row);
+      if (winnerByKey.has(key)) {
+        droppedIds.add(row.prompt_id);
+        // eslint-disable-next-line no-console
+        console.warn('[upsertPrompts] dedup-drop', {
```
aliciadriani, line 746: **Should Fix:** `console.warn` here is forced because `upsertPrompts` has no `log` parameter, but this dedup-drop event is an operational audit trail — it should flow through the same structured logger (with request correlation ID, brand context, etc.) as the rest of the call stack. Consider adding an optional `log` parameter to `upsertPrompts` in a follow-up PR so these drops show up in Coralogix alongside the request that triggered them.

src/support/prompts-storage.js
```diff
@@ -718,6 +719,45 @@ export async function upsertPrompts({
...
+          brand_id: row.brand_id,
+          text_hash: crypto.createHash('sha256').update(row.text || '').digest('hex').slice(0, 12),
+          dropped_prompt_id: row.prompt_id,
+          dropped_topic_id: row.topic_id ?? null,
+          winning_prompt_id: winnerByKey.get(key).prompt_id,
+          winning_topic_id: winnerByKey.get(key).topic_id ?? null,
```
aliciadriani, line 752: **Should Fix (coverage):** The `if (winnerByKey.has(key))` branch that contains this block (lines 740–749) is **not covered by any test** — the coverage report shows lines 740–749 and 755–757 as uncovered. The dedup block is entered (`toInsert.length > 1` fires), but `dedupKey` never produces a collision in the existing tests, so the drop path and both `splice` mutations are dead code from the test suite's perspective.

Add a test where two rows have the same `row.text` (after `.toLowerCase()`) and the same `regions` — the INSERT mock returns success (no 23505 needed) and the assertion verifies `console.warn` was called once with the correct `dropped_prompt_id`/`winning_prompt_id`, and that `result.created === 1` not 2.

src/controllers/brands.js
```diff
@@ -491,6 +491,10 @@ function BrandsController(ctx, log, env) {
 
       return createResponse({ created, updated, prompts: outPrompts }, 201);
     } catch (error) {
+      if (error?.status === 409) {
+        log.warn(`Prompt unique-constraint conflict for brand ${brandId}: ${error.message}`);
```
aliciadriani: **Nit:** The warn log includes `brandId` but not `organizationId`. The request handler has access to both; adding it makes Coralogix incident correlation faster (brands are looked up by org in most runbooks).

src/support/prompts-storage.js
aliciadriani: Fixed in commit 8e61cac6. `dedupKey` now applies `.trim().toLowerCase()` on text and `.map(x => String(x).toLowerCase()).sort()` on regions, matching `getKey`. Expanded to multi-line to stay within the 100-char limit:

```js
const dedupKey = (row) => {
  const t = String(row.text || '').trim().toLowerCase();
  const r = [...(row.regions || [])].map((x) => String(x).toLowerCase()).sort().join(',');
  return `${t}:${r}`;
};
```

src/support/prompts-storage.js
aliciadriani: Done in commit 8e61cac6. Added to the existing comment block: _'Guard is > 1: a single-row batch cannot collide with itself.'_

src/support/prompts-storage.js
aliciadriani: Fixed in commit 8e61cac6 along with the MysticatBot blocking issue. Both `trim()` and region lowercasing are now applied in `dedupKey`, making it consistent with `getKey`.

src/support/prompts-storage.js
aliciadriani: Acknowledged. `upsertPrompts` has no `log` parameter and adding one is a larger change that would touch all call sites. Leaving `console.warn` as-is for now; the follow-up to thread a `log` param through (as `findPromptsBlockingRegionRemoval` already does at line 1016) is tracked in LLMO-XXXX.

src/support/prompts-storage.js
aliciadriani: Fixed in commit 67d03704. Added a dedicated test _'dedup-drop fires once per duplicate and splice reduces toInsert to exactly one row'_ with a 3-prompt batch (all same synthetic text+regions). The drop path fires twice, `console.warn` is spied and asserted, and the splice reduces `toInsert` to exactly one row. `prompts-storage.js` is now at 100% line coverage.

src/controllers/brands.js
aliciadriani: Done in commit 8e61cac6: `log.warn(`Prompt unique-constraint conflict for brand ${brandId} (org ${spaceCatId}): ${error.message}`)`

MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - prior findings addressed, fix commits are clean.
_Changes: fixes deterministic HTTP 500 on POST /prompts by adding intra-batch dedup, classifying 23505 as 409, and aligning getKey collation with the DB constraint (4 files)._

- `dedupKey` normalization now matches `getKey` (`.trim().toLowerCase()` on text, `.map(x => String(x).toLowerCase()).sort()` on regions) - fixed in 8e61cac6.
- Dedup drop path has full test coverage (3-duplicate batch test fires the branch twice, verifies splice reduces to 1 row) - added in 67d03704.
- Controller `log.warn` now includes `organizationId` for Coralogix correlation - fixed in 8e61cac6.

<details>

- nit: `text_hash` in the dedup-drop log hashes raw `row.text` rather than the normalized form used for the key - an operator trying to reproduce a collision from the hash alone cannot, since normalization (trim + lowercase) is not reflected in the hash. Consider hashing the already-computed normalized `t` instead - `src/support/prompts-storage.js:745`.

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 27s | <b>Cost:</b> $4.36 | <b>Commit:</b> <code>67d037041f6b9abb5d3727303613707c6ae2c5f7</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2650, opened 2026-06-19
feat: server-side text search for prompt research | LLMO-5216

calvarezg:
## Summary
Adds an optional `textFilter` query param to the four Prompt Research endpoints (`/topics/research`, `/prompts`, `/brands`, `/source-domains`) and their CSV exports. It filters the **full result set server-side** via a `dimension_filter_ql` `CONTAINS` clause — replacing the current client-side, current-page-only table search — so both the table and the export reflect all matching rows, and `total` reflects the filtered count.

## Param + columns
- New param: `textFilter` (distinct from `searchQuery`, which is the topic FTS query).
- `dimension_filter_ql` column per dataset: prompts→`prompt`, topics→`topic`, brands→`name`, source-domains→`domain`.
- Empty/absent `textFilter` → field omitted (request unchanged).

## Changes
- `grpc-utils.js`: new `buildTextFilterQl(textFilter, column)` (reuses `escapeQlString`).
- `topics.js`: threaded onto every gRPC call — list, `*Totals`, each `FTS_LLMS` per-engine call, the topic count helpers, and the prompts `topicId` path — so filtered totals stay consistent.
- Export: `runFtsResearchExport` + 4 wrappers (`textFilterColumn`) set it on the export request.
- OpenAPI: shared `textFilter` param on all 8 ops; rendered docs regenerated.
- No new routes (capability/route/index tests untouched).

## Test plan
- [x] `npx mocha` on touched files: 358 passing (new `buildTextFilterQl` unit tests + per-dataset threading assertions + "omits when absent").
- [x] `npm test`: 12424 passing, coverage 99.95%.
- [x] `npm run docs:lint` valid; `npm run docs:build` regenerated; `eslint` clean.
- Note: brands→`name` is the one column not previously proven against Semrush — validated in manual testing.

## Related Issues
https://jira.corp.adobe.com/browse/LLMO-5216

- [x] make sure to link the related issues in this description.
- [x] when merging / squashing, make sure the fixed issue references are visible in the commits.

### DIFF
--- a/docs/openapi/ai-visibility-api.yaml
+++ b/docs/openapi/ai-visibility-api.yaml
@@ -118,6 +118,19 @@ x-ai-vis-search-query: &searchQueryParam
     type: string
     example: coffee
 
+x-ai-vis-text-filter: &textFilterParam
+  name: textFilter
+  in: query
+  required: false
+  description: |
+    Optional free-text filter applied server-side across the full result set (not just the
+    current page), via a `dimension_filter_ql` `CONTAINS` clause on the dataset's primary
+    column (prompts→`prompt`, topics→`topic`, brands→`name`, source-domains→`domain`). When
+    set, `total` reflects the filtered count. Distinct from `searchQuery` (the topic FTS query).
+  schema:
+    type: string
+    example: pdf
+
 x-ai-vis-topic-id: &topicIdParam
   name: topicId
   in: query
@@ -1078,6 +1091,7 @@ ai-visibility-topics-research:
       - *regionParam
       - *engineParam
       - *ftsTopicsSortByParam
+      - *textFilterParam
       - *ftsSortDirectionParam
       - *limitParam
       - *offsetParam
@@ -1172,6 +1186,7 @@ ai-visibility-topics-research-prompts:
       - *regionParam
       - *engineParam
       - *ftsPromptsSortByParam
+      - *textFilterParam
       - *ftsSortDirectionParam
       - *limitParam
       - *offsetParam
@@ -1206,6 +1221,7 @@ ai-visibility-topics-research-brands:
       - *regionParam
       - *engineParam
       - *ftsBrandsSortByParam
+      - *textFilterParam
       - *ftsSortDirectionParam
       - *limitParam
       - *offsetParam
@@ -1242,6 +1258,7 @@ ai-visibility-topics-research-source-domains:
       - *regionParam
       - *engineParam
       - *ftsSourceDomainsSortByParam
+      - *textFilterParam
       - *ftsSortDirectionParam
       - *limitParam
       - *offsetParam
@@ -1276,6 +1293,7 @@ ai-visibility-v1-prompt-research-prompts-export:
       - *regionParam
       - *engineParam
       - *ftsPromptsSortByParam
+      - *textFilterParam
       - *ftsSortDirectionParam
       - *limitParam
       - *offsetParam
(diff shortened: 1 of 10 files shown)


calvarezg:
## Code Review

### What's well done

- `buildTextFilterQl` is a clean, properly isolated utility. `escapeQlString` is reused correctly — no injection risk.
- The YAML anchor `*textFilterParam` is the right approach: one definition, 8 references, zero drift.
- The "omit field when empty" pattern (`...(dimensionFilterQl ? { dimensionFilterQl } : {})`) is applied consistently across all call sites, so absent `textFilter` leaves gRPC request shapes completely unchanged — safe for callers that don't know about the field.
- Unit tests for `buildTextFilterQl` cover the full edge set: trim, escape, empty/whitespace/null/undefined.

---

### Issues

**Missing test: multi-LLM (ALL engines) path in `handleTopicsResearch`**

The new test only exercises `engine=chatgpt` (single-LLM path). The ALL-engines branch calls `countDistinctTopicIdsAcrossFtsLlms` and maps over `FTS_LLMS` — both of which were also changed. A test with no `engine` param would cover this.

**Missing test: `topicId` path in `handleTopicsResearchPrompts`**

`promptsByTopicIdsPage` was updated to accept and forward `dimensionFilterQl`, but the test suite only exercises the `searchQuery`-based FTS path. Adding `topicId=123` to the test params would hit the `promptsByTopicIdsPage` code path.

**Silent footgun in `runFtsResearchExport` config**

If a future export is added without `textFilterColumn`, `buildTextFilterQl(sp.get('textFilter'), undefined)` produces `"undefined CONTAINS \"...\""` — syntactically valid enough to reach Semrush but semantically wrong. The code is in a `c8 ignore` block so this won't surface in coverage. Cheapest fix: a guard at the top of `runFtsResearchExport`:

```js
if (textFilterColumn === undefined) {
  throw new Error('runFtsResearchExport: config.textFilterColumn is required');
}
```

Or just document it as required in the JSDoc `@param`.

**Minor: inconsistent `filterField` usage**

`handleTopicsResearchPrompts`, `handleTopicsResearchBrands`, and `handleTopicsResearchSourceDomains` extract a `filterField` variable, but `handleTopicsResearch` inlines the spread directly. Not a correctness issue, just slightly inconsistent within the same file.

---

### Verdict

Implementation is correct and the happy-path coverage is solid. The two missing test cases (multi-LLM path, `topicId` path) are the main gap — they're the paths most likely to regress silently if someone restructures the parallel call patterns later. The `textFilterColumn` footgun is low-priority given the internal config shape, but a one-liner guard would be cheap insurance.

calvarezg:
## Triage

Thanks for the review — addressed one item and intentionally skipping the rest, rationale below. (From [the code review comment](https://github.com/adobe/spacecat-api-service/pull/2650#issuecomment-4752177414).)

**Addressed**
- **Robustness — `textFilterColumn` footgun.** Added a fail-fast guard in `runFtsResearchExport` so a future export wrapper missing `textFilterColumn` can't emit `undefined CONTAINS "..."`. (commit `52b97f8`)

**Dismissed**
- **Missing multi-LLM (all-engines) test** and **missing `topicId` path test** — **Dismissed.** These FTS research handlers are deliberately `c8 ignore`d and follow the same no-added-unit-test pattern as the sibling v1 export handlers (`brand-prompts-export` et al.); the `textFilter` threading is covered by the `buildTextFilterQl` unit tests plus the happy-path handler assertions, and the conditional-spread keeps unfiltered request shapes byte-identical.
- **`filterField` vs inline spread inconsistency** — **Dismissed.** Cosmetic; `handleTopicsResearch` also passes `{ dimensionFilterQl }` unconditionally to the count helpers, so a single shared variable wouldn't fully unify the pattern anyway.

calvarezg:
✅ **Approve** — footgun guard is in, triage rationale on the test gaps is defensible. Good to merge.

aniham (approved):

PR #2652, opened 2026-06-19
feat: add backoffice review capture endpoint + reviews composition (SITES-43974)

tathagat2241:
What

POST /sites/:siteId/opportunities/:opportunityId/suggestions/:suggestionId/backoffice-reviews — captures an ESE verdict. Mandatory client-supplied event_id (idempotent via PG 23505→200), IMS auth + org check, verdict→signal, server-derived reviewer, tier from entitlement, secret-scrub + markdown sanitise. Raw patches not echoed.
Extends the suggestion fetch with ?include=reviews (read-time composition; ,patches opt-in).
New feedback-redaction util + OpenAPI (SuggestionReview schema) + tests.
⚠️ Merge checklist (required — do before merge)

not done
Bump @adobe/spacecat-shared-data-access to the version published by PR 2 (it currently pins 3.75.1, which lacks the new exports). The capture handler statically imports them; without the bump the controller fails to load.
Testing

Syntax + redaction logic + OpenAPI YAML validated. Controller + util unit tests added (mocha/coverage run in CI).

- [X] make sure you add a "Not implemented yet" note the endpoint description, if the implementation is not ready 

## Related Issues
https://jira.corp.adobe.com/browse/SITES-39001

Thanks for contributing!


src/support/feedback-redaction.js
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * Strip dangerous HTML from a markdown string (NFR-05). Removes
+ * script/style/iframe blocks, inline event-handler attributes, and
+ * javascript:/data: URI schemes. Conservative — leaves ordinary markdown intact.
+ *
+ * @param {string} markdown
+ * @returns {string} sanitised markdown
+ */
+export function sanitizeMarkdown(markdown) {
```
absarasw, line 64: **Markdown sanitisation is denylist-shaped, not allowlist (spec called for nh3 / bleach-equivalent).**

The file's own JSDoc names it correctly: *"Allowlist-ish HTML sanitisation."* The implementation is regex-based denylist (`<script>`/`<style>`/`<iframe>` blocks, `on*=` event handlers, `javascript:`/`data:text/html` schemes). The spec is explicit:

> *"Library: an allowlist-based HTML sanitiser — implementation should use `nh3` (Rust-based, Mozilla's `ammonia` Python binding — the maintained successor that the `bleach` project itself points at since `bleach`'s January 2023 deprecation). `bleach` is acceptable as a short-term fallback if `nh3` isn't yet wired into the spacecat-api-service dependency tree, but the spec is 'allowlist-based HTML sanitiser,' not a specific library name."*

For a JS controller, the equivalent allowlist sanitiser is **DOMPurify** (the Node-side `isomorphic-dompurify` or `jsdom`-backed). Allowlist sanitisers reject anything not on the safelist by construction — they don't have the bypass surface that denylists do.

Known bypass classes the current regex doesn't catch:

- **Mutation XSS** via `<noscript>`, `<template>`, or comment-context smuggling. `<noscript></noscript>` renders in some clients.
- **HTML-entity-encoded payloads** — `&lt;script&gt;` becomes `<script>` after one decode pass, slipping past the regex.
- **SVG-internal scripts** — `<svg><script>alert(1)</script></svg>` is NOT stripped by the script-regex (the script tag IS inside SVG which is not in the strip list; see the `data:image/svg+xml` finding below for the URI variant).
- **Unicode-escaped event handlers** or alternate-encoding variants.

For an internal Backoffice with trusted ESEs, the risk is modest — but the rationale text round-trips into the API response when other ESEs read the review history (`?include=reviews`), so a malicious ESE could in principle XSS a colleague's Backoffice UI. The spec was explicit about "allowlist-based" specifically to close this class of risk.

**Suggestion**: replace `sanitizeMarkdown` with a DOMPurify call (or equivalent allowlist sanitiser already in spacecat-api-service deps). Allowed tags: `p, br, em, strong, code, pre, ul, ol, li, blockquote, a` (with `href` constrained to `http:`/`https:`/`mailto:`); allowed attrs: `href` on `a`, `class` on `code`/`pre`. Everything else stripped by construction. If DOMPurify isn't yet a dep, document the gap explicitly and track a follow-up — but the regex denylist as-is isn't what the spec calls for.

src/support/feedback-redaction.js
```diff
@@ -0,0 +1,139 @@
...
+  }
+  return markdown
+    // drop whole script/style/iframe elements (with or without a closing tag)
+    .replace(/<\s*(script|style|iframe)\b[^>]*>[\s\S]*?<\s*\/\s*\1\s*>/gi, '')
+    .replace(/<\s*(script|style|iframe)\b[^>]*>/gi, '')
+    // strip inline event-handler attributes: onclick=, onerror=, ...
+    .replace(/\son\w+\s*=\s*("[^"]*"|'[^']*'|[^\s>]+)/gi, '')
+    // neutralise dangerous URI schemes
+    .replace(/javascript:/gi, '')
+    .replace(/data:(?=text\/html|application)/gi, '');
```
absarasw: **`data:` URI strip misses `data:image/svg+xml` — known XSS vector.**

`data:(?=text\/html|application)` only strips `data:text/html,...` and `data:application/*`. The spec said *"stripped explicitly: ... `javascript:` and `data:` URLs"* (all `data:` URLs).

What's currently missed:

- **`data:image/svg+xml;base64,...`** — SVG is a script-execution surface. A `data:image/svg+xml,<svg><script>alert(1)</script></svg>` in a markdown link would survive sanitisation.
- **`data:video/*`** — less common XSS vector but on the same class.
- **`data:image/*` with embedded XML** — generally less risky but worth considering.

Not stripping `data:image/png;base64,...` etc. is a reasonable carve-out (common harmless markdown pattern). But `data:image/svg+xml` IS dangerous because SVG executes scripts.

**Suggestion**: either tighten the regex to `data:(?=text\/html|application|image\/svg)` to also strip SVG, OR (if you're moving to an allowlist sanitiser per the previous finding) this concern goes away because allowlist sanitisers reject SVG by default. Note: the more common pattern is to strip ALL `data:` URIs in `href` / `src` contexts and trust that the inline-image use case goes through proper markdown image syntax with an http(s) URL.

src/controllers/suggestions.js
```diff
@@ -2701,7 +2797,150 @@ function SuggestionsController(ctx, sqs, env) {
...
+    }
+
+    const postgrestClient = context.dataAccess?.services?.postgrestClient;
+    if (!postgrestClient?.from) {
+      return createResponse({ message: 'Feedback store unavailable' }, 503);
+    }
+
+    // reviewer_id is server-derived from the authenticated principal — never the body.
+    const { profile } = context.attributes?.authInfo ?? {};
+    const reviewerId = profile?.email ?? null;
```
absarasw, line 2934: **Per-`(reviewer_id, site_id)` rate-limit is missing.**

The v4-design `§11.1` specifies:

> *"Rate-limiting (per v3 review minor): the capture endpoint enforces a per-`(reviewer_id, site_id)` rate limit (default: 60 captures per hour) to prevent a malicious-but-authenticated client from flooding `reviews[]` with unique `event_uuid`s. The rate-limit floor is implemented at the spacecat-api-service layer (Phase 6). Excess requests return HTTP 429."*

OpenAPI documents the 429 response for this endpoint (line 88 of `site-opportunities.yaml`), but the controller never emits 429 — there's no rate-limit check anywhere in `createBackofficeReview`.

Without it, an authenticated reviewer (or a compromised reviewer credential) can flood the `feedback_event` table with up-to-`event_id`-uniqueness rows. The DB has no `MAX_REVIEWS_PER_PAGE`-equivalent cap (that lives in the design only because the v2-blackboard-native version was page-scoped; this PR is the SpaceCat-table version which has no such cap).

**Suggestion**: add a rate-limit check after `reviewerId` is derived. Two reasonable shapes:

1. **At the spacecat-api-service middleware layer** — if there's an existing rate-limit middleware in this service, configure it for this route at `60/hr per (reviewer_id, site_id)` and remove the 429 from OpenAPI if no rate-limit lands. (Cleaner.)
2. **In the controller** — query `feedback_event` for the count of rows matching `(reviewer_id, site_id, event_time > now() - 1 hour)` before insert; if ≥ 60, return 429. Adds one read per write (acceptable at expected ESE volume).

If this is being tracked as follow-up (acceptable — the absolute volume is small at ESE scale), please update the OpenAPI to remove the 429 response or add a TODO/comment naming the gap.

src/controllers/suggestions.js
```diff
@@ -2701,7 +2797,150 @@ function SuggestionsController(ctx, sqs, env) {
...
+    } = redactFeedbackContent({ detailMarkdown, previousFix, editedFix });
+
+    const scrubbed = Object.entries(scrubHits);
+    if (scrubbed.length > 0) {
+      context.log?.info?.(`feedback_capture.scrub_hit_total ${JSON.stringify(scrubHits)} suggestion=${suggestionId}`);
+    }
+
+    const row = {
+      event_id: eventId,
+      organization_id: site.getOrganizationId?.() ?? opportunity.getSiteId?.(),
```
absarasw: **`organization_id` fallback is a real bug — substitutes a `site_id` for an `org_id` on the unlikely path the fallback fires.**

```js
organization_id: site.getOrganizationId?.() ?? opportunity.getSiteId?.(),
```

If `site.getOrganizationId()` returns `null`/`undefined` (or is missing), this falls back to `opportunity.getSiteId()` — which returns a **SITE id**, not an **ORG id**. The two are unrelated UUID columns.

If this fallback ever fires in production:
- The `feedback_event.organization_id` column gets a site UUID stored as an org UUID.
- DMAI's training corpus + any org-scoped reporting based on `organization_id` would attribute the event to the wrong (and nonexistent) org.
- The DB has no FK on `organization_id` (durable record per #718's design), so this corruption goes silent.

The fallback was probably written defensively ("what if `site.getOrganizationId()` is missing?"), but the fallback value is incorrect — better paths:

1. **Just trust the site model**: `organization_id: site.getOrganizationId(),` — if `site` was loaded successfully by `Site.findById(siteId)`, `getOrganizationId()` should always return a valid org id; if it doesn't, that's a data-integrity issue worth surfacing as a 500.
2. **Throw on missing org id**: `const orgId = site.getOrganizationId(); if (!orgId) return createResponse({ message: 'site has no organization' }, 500);` — fail loud, not silent.
3. **Drop `?.()`**: `site.getOrganizationId()` (no optional chaining) — if site is somehow null/missing the method, we'd want a TypeError, not silent corruption.

**Suggestion**: pick (1) or (2). Option (3) might break tests that mock `site` without `getOrganizationId`, but those tests probably want fixing anyway.

src/controllers/suggestions.js
```diff
@@ -2701,7 +2797,150 @@ function SuggestionsController(ctx, sqs, env) {
...
+      site_id: siteId,
+      suggestion_id: suggestionId,
+      opportunity_type: opportunity.getType?.() ?? null,
+      source: REVIEW_SOURCES.BACKOFFICE,
+      signal,
+      reviewer_id: reviewerId,
+      detail_markdown: cleanMarkdown ?? null,
+      previous_fix: cleanPreviousFix ?? null,
+      edited_fix: cleanEditedFix ?? null,
+      state_transition: hasText(stateTransition) ? stateTransition : null,
```
absarasw, line 2963: **`state_transition` is passed through without enum validation — LA corpus could receive arbitrary strings.**

```js
state_transition: hasText(stateTransition) ? stateTransition : null,
```

The spec enumerates the valid `state_transition` values:

> *`'PENDING_VALIDATION->NEW' | 'PENDING_VALIDATION->REJECTED' | 'EDIT' | 'ASO_UI_VOTE' | null`*

The DB migration (#718) deliberately doesn't constrain `state_transition` (free text, by spec — to allow future values without a schema change). But the controller is the natural enforcement point for the v1 enum.

Without enum validation here, a Backoffice UI bug or a malicious client could send arbitrary strings (`"hello"`, `"<script>"`, a 10 KB JSON blob, etc.) — they all land in the JSONL export and into DMAI's training corpus as legitimate-looking labels.

**Suggestion**: add a validation similar to the `rejection_category` check a few lines above:

```js
const VALID_STATE_TRANSITIONS = ['PENDING_VALIDATION->NEW', 'PENDING_VALIDATION->REJECTED', 'EDIT', 'ASO_UI_VOTE'];
if (stateTransition != null && !VALID_STATE_TRANSITIONS.includes(stateTransition)) {
  return badRequest('invalid state_transition');
}
```

Ideally this constant lives in `spacecat-shared` (PR #1696) alongside `REVIEW_VERDICTS` et al. so it doesn't drift across api-service / jobs-dispatcher. (Already noted in my #1696 review — `state_transition` was one of the minor items there too.)

absarasw (changes requested):
Thanks Tathagat — substantial PR. A lot lands cleanly against the spec:

- **FR-09 (mandatory `event_id`)** — no server fallback; rejects missing + non-UUID with 400; PG 23505 collapses to 200 with the existing row. Tests cover all three paths.
- **FR-10 (source bound to route)** — controller stamps `REVIEW_SOURCES.BACKOFFICE` from the route; rejects body-supplied source mismatches with 400. Route variant matches the per-source design.
- **Cross-tenant write guard** — `accessControlUtil.hasAccess(site)` rejects with 403 when caller lacks org access; suggestion's opportunity site-id is verified against the URL site-id.
- **Server-derived reviewer_id**, **8 KB cap with 413**, **comprehensive secret-pattern scrub** (PEM, AWS, GH/GL PATs, Slack, JWT, Bearer, basic-auth-URL, Adobe hosts/emails — covers each pattern in a dedicated test), **idempotency tests**, **OpenAPI updated with `SuggestionReview` schema + per-status responses**, **`?include=reviews,patches` opt-in for raw patches**, **503 when feedback store unavailable**, **graceful empty-reviews on query error**. Test coverage is genuinely strong (20+ cases on the controller + a per-pattern coverage test on the scrubber).

## Findings

Five `Important` issues inline. They cluster on two themes:

- **Security tightenings the spec was explicit about** — markdown sanitisation is denylist-shaped (the file's own JSDoc says *"Allowlist-ish"*) where the spec called for an allowlist sanitiser (nh3 / bleach-equivalent); the `data:` URI strip misses `data:image/svg+xml` (known XSS vector); per-`(reviewer_id, site_id)` rate-limit is missing (OpenAPI documents 429 but the controller can't return it).
- **Correctness gaps** — `organization_id` fallback would substitute a `site_id` into the org column on the unlikely path the fallback fires; `state_transition` is free-text passed straight through (no enum validation), so the LA corpus can receive arbitrary strings.

A few `Minor` observations not worth PR comments — happy to mention on Slack if useful: scrub-hits emitted as a log line not a metric counter (depends on whether SpaceCat's metrics derive from logs), `reviewer_id` stored as IMS email vs the v4-design's `ims_sub` wording (spec is internally inconsistent; impl matches the DB-migration comment), `Buffer.byteLength` size check happens before `redactFeedbackContent` (correct ordering — protects against giant payloads regardless of HTML content), `scrubbed = Object.entries(scrubHits)` could be `Object.keys(...).length` (style).

## Assessment

Ready to merge? With fixes.

The security findings are real — the spec was explicit about an allowlist sanitiser, and the missing rate-limit + incomplete `data:` strip combined open a small but actual XSS-via-rationale surface that ESEs are trusted-but-not-infallible carriers of. The `organization_id` fallback bug is unlikely to fire in practice but would silently corrupt the corpus if it did. The `state_transition` enum gap is data-quality.

None of these block the overall design; all are addressable in this PR without restructuring.

## Merge prereq from the PR description (carried forward)

The PR body already notes that PR #1696 (spacecat-shared) must publish before this PR can merge — the controller statically imports `REVIEW_SOURCES`, `REVIEW_VERDICTS`, `REJECTION_CATEGORIES`, `FEEDBACK_TIERS`, `verdictToSignal`, `toReviewView`. Acknowledged. ✓

ramboz (changes requested):
**Verdict: REQUEST CHANGES**

Well-structured, well-tested PR — the controller follows the repo's established auth/validation patterns, idempotency is handled correctly, and test coverage is thorough (24 controller cases + redaction unit tests). But two hard blockers and one cross-repo contract item must be resolved before merge.

> **Cross-repo note (added during synthesis):** two of this agent's flags were *cleared* by cross-checking the other PRs:
> - "should be a `wrpc_` RPC" — DB PR 718 intentionally exposes no write RPC and grants `postgrest_writer` INSERT; direct insert is the design. **Not a divergence.**
> - "is the writer JWT wired?" — confirmed: shared `service/index.js:71` sets `Authorization: Bearer` from `POSTGREST_API_KEY`; existing write controllers (e.g. `brands.js`) rely on it. **Cleared** (operational caveat: the `POSTGREST_API_KEY` secret must be present in the Lambda env — already true for current postgres writes).

## Blockers

**B1 — Dependency pinned to an untrusted personal gist tarball.** `package.json:88` pins `@adobe/spacecat-shared-data-access` to `https://gist.github.com/tathagat2241/.../adobe-spacecat-shared-data-access-3.79.0.tgz`. Supply-chain risk (unversioned, unsigned, single-author, mutable URL); will not pass review/release. Replace with the registry version published by PR 1696. Those six imported symbols (`REVIEW_SOURCES`, `REVIEW_VERDICTS`, `REJECTION_CATEGORIES`, `FEEDBACK_TIERS`, `verdictToSignal`, `toReviewView`) don't exist in any published data-access version today, and are statically imported at module top-level, so a wrong/missing version fails the entire suggestions controller load. Do not merge until 1696 has landed + published.

**B2 — `reviewer_id` stores the IMS user GUID but the contract says "email".** `src/controllers/suggestions.js` (~line 2871): `const reviewerId = profile?.email ?? null;`. For IMS callers, `profile.email` is the IMS user_id GUID, not a mailbox — documented elsewhere in this same file. Yet `schemas.yaml` and the PR both say "IMS email." This mislabels identity for the entire corpus/export/training pipeline. Either use the real address (`profile.trial_email`) or rename/redocument the field as an opaque IMS user id. The unit test uses a fake `ese@adobe.com` so it never catches the GUID behavior.

## Concerns

**C1 — Direct table INSERT vs RPC.** *(Resolved during synthesis — see cross-repo note above; DB uses writer INSERT grant, no write RPC.)*

**C2 — Writer auth / JWT to PostgREST.** *(Resolved — wired via shared, used by existing controllers.)*

**C3 — The markdown "sanitizer" is regex-based and bypassable.** `src/support/feedback-redaction.js:sanitizeMarkdown`. Concrete gaps: event-handler stripper requires whitespace (`/\son\w+/`) so `<svg/onload=...>` is not stripped; `.replace(/javascript:/gi,'')` is text-deletion not neutralization (`[x](javascript:evil())` → `[x](evil())`); only `script|style|iframe` removed (`<object>`/`<embed>`/`<a href>` pass). Acceptable only as documented defense-in-depth *if* `detail_markdown` is never rendered as trusted HTML downstream. Since it feeds an S3 corpus + possibly a UI via `?include=reviews`, confirm the render path escapes; if anything renders it as HTML this is stored XSS → blocker. Recommend `sanitize-html`/DOMPurify or plain text, and document the "escape at render" assumption.

**C4 — `organization_id` fallback is wrong.** `~line 2901`: `organization_id: site.getOrganizationId?.() ?? opportunity.getSiteId?.()`. The fallback is a **site** id, not an org id — silently corrupts the column if `getOrganizationId` is ever nullish. Use `null` (or fail).

**C5 — Secret-scrub can corrupt legitimate patch data and isn't reversible.** `scrubDeep` rewrites `previous_fix`/`edited_fix`. For a corpus that captures code patches, false positives (a base64 blob matching `jwt`, any `sk-...` string) silently mangle the stored fix, degrading training data with only a count for audit. Consider scrub metadata or gating scrub to `detail_markdown` (free text) only.

**C6 — `?include=reviews` has no pagination/limit.** `select('*').order('event_time')` — a suggestion with many reviews returns all rows, and `?include=...,patches` pulls full patch JSONB. Add a `.limit(N)` and confirm a `(suggestion_id, event_time)` index DB-side (718 provides `idx_feedback_event_suggestion`). Query parsing itself is correct (GET reads `include` from query string, no injection); no N+1 for single-suggestion GET.

## Nits
- N1 — Read path returns `[]` on missing postgrest client while write path returns 503. Fine but worth a comment that it's intentional.
- N2 — `detailMarkdown` 8 KB cap uses `Buffer.byteLength > 8192` (bytes) but OpenAPI says `maxLength: 8192` (chars). Multi-byte UTF-8 skew.
- N3 — `state_transition` stored free-form with no whitelist.
- N4 — `opportunity_type: opportunity.getType?.() ?? null` — confirm DB tolerates arbitrary type strings (it does; unconstrained).
- N5 — Test `reviewer_id: 'ese@adobe.com'` reinforces the B2 mislabel; after the fix, add a GUID-shaped test.

## Strengths
- Authorization matches the repo baseline exactly (UUID validation → `Site.findById` → `accessControlUtil.hasAccess` → 403); route added to `routeRequiredCapabilities` with `CAP_SUGGESTION_WRITE`.
- `source` route-bound (rejects body spoofing, FR-10); `event_id` idempotency via PG `23505` → 200-with-existing-row, tested (FR-09).
- Enum validation uses the **shared constants** + `verdictToSignal` — no hand-rolled lists. Exactly the cross-repo discipline needed (once the dep is fixed).
- Raw patches not echoed by default; `?include=...,patches` is explicit opt-in.
- `deriveFeedbackTier` and reviews fetch fail soft (never throw).
- Genuinely thorough tests (happy path, all validation failures, 403/404 ownership, idempotency, 500/503, tier derivation, scrub metric, all five `?include` permutations). OpenAPI complete.

src/support/feedback-redaction.js
tathagat2241: Addressed: (d) organization_id now uses site.getOrganizationId() directly (no site-id fallback); (e) state_transition is validated against an allowlist → 400; (a)/(b) hardened the sanitiser (svg/noscript/template + data:image/svg+xml) and corrected the "Allowlist-ish" wording — full allowlist sanitiser (sanitize-html) tracked as a follow-up since no sanitiser dep exists and the consumer renders via react-markdown without rehype-raw; (c) removed the 429 from OpenAPI to match reality, with the rate limiter tracked as a follow-up. Added tests for the new validation + sanitiser cases.

tathagat2241:
B2 — kept profile.email (it's the stable IMS user id, good for reviewer-continuity) but relabeled it everywhere as an opaque IMS user identifier, not a mailbox (controller comment, schemas.yaml, feedback_event.reviewer_id column comment) + added a GUID-shaped test.
C3 — added <object>/<embed> to the strip list; kept the whitespace-anchored event-handler match (broadening to / mangles legit markdown URLs); documented the allowlist-sanitiser follow-up and the render-escape assumption (Backoffice uses react-markdown without rehype-raw).
C6 — added .limit(100) to ?include=reviews (backed by idx_feedback_event_suggestion).
C5 — kept scrubbing all three fields per spec §11.3; feedback_capture.scrub_hit_total is the audit trail.
C4 / Comment 3 (org_id), state_transition enum, data:svg, 429 were addressed earlier in the PR.
B1 — acknowledged; the gist pin is temporary and will be swapped for the registry version once #1696 publishes (documented merge prerequisite).
N2 (byte vs char cap) — minor; can align the OpenAPI note if you'd like.

src/controllers/suggestions.js
tathagat2241: Fixed.

src/controllers/suggestions.js
tathagat2241: Fixed.

src/controllers/suggestions.js
tathagat2241: fixed.

src/support/feedback-redaction.js
tathagat2241: Fixed.

ramboz (approved):

absarasw (approved):
Thanks for the turnaround — 4 of the 5 Important findings cleanly addressed in `712b3c01`, plus the cross-repo `reviewer_id` mislabel from the follow-up review.

**Cleanly fixed:**
- **`data:image/svg+xml` XSS gap** — regex now strips it (`data:(?=text\/html|application|image\/svg)`); new test asserts svg-data-uri is neutralised but `data:image/png` is left alone.
- **`organization_id` fallback** — dropped the `?? opportunity.getSiteId?.()` fallback; now just `site.getOrganizationId()`. New test asserts `inserted.organization_id === ORG_ID`.
- **`state_transition` enum unvalidated** — new `FEEDBACK_STATE_TRANSITIONS = ['PENDING_VALIDATION->NEW', 'PENDING_VALIDATION->REJECTED', 'EDIT']`; controller returns 400 on unknown values; tests cover both `HACK->ARBITRARY` (rejected) and missing (accepted). Minor note (not blocking): `ASO_UI_VOTE` from the v4 spec is intentionally not in this route's list — it'll need adding when the ASO UI capture route lands for SITES-39002.
- **`reviewer_id` GUID-vs-email label (cross-repo X5)** — OpenAPI schema updated to `"Server-derived, opaque IMS user identifier (e.g. <id>@<authSrc>) — NOT a mailbox. Stable per user (reviewer-continuity signal)."`; controller has an explicit NOTE comment; new test asserts a GUID-shaped `82521DEADBEEF0123@AdobeOrg` value is stored verbatim.

**Sanitiser hardening (finding #1) — accepted with the follow-up:**
Still a denylist, but significantly closed: `svg`/`noscript`/`template`/`object`/`embed` added to the strip list, `data:image/svg+xml` neutralised, JSDoc explicitly acknowledges denylist and documents the defence-in-depth (Backoffice renders via `react-markdown` **without `rehype-raw`**, so embedded HTML doesn't execute anyway). Allowlist migration tracked as SITES-43974 / OQ-5. Tests added per new strip class. Reasonable interim posture given the render-time non-execution guarantee.

**Bonus cleanup:** new `FEEDBACK_REVIEW_READ_LIMIT = 100` on the `?include=reviews` read path (backed by `idx_feedback_event_suggestion`) — bounds payload size esp. with `?include=patches`.

**Two things to flag before merge (neither new to this commit):**

1. **Rate-limit disposition (finding #3).** OpenAPI's 429 was removed to match the (still-absent) controller implementation, so spec and impl now agree — good. The reviewer-flood surface itself remains open though. Confirm the intent: is per-`(reviewer_id, site_id)` rate-limiting deferred to a gateway/WAF layer, or is the risk accepted for v1 (mitigated by IMS-authenticated `reviewer_id` + the 8 KB per-row cap)? Either is fine — just want it explicit so it doesn't fall through the crack.

2. **Gist-tarball pin on `@adobe/spacecat-shared-data-access` (Mihir's cross-repo X4).** The pin was updated to a newer gist URL but is still a `https://gist.github.com/tathagat2241/.../adobe-spacecat-shared-data-access-3.79.0.tgz` tarball rather than the published registry version. Already acknowledged as a merge prereq in the PR body (needs #1696 to merge + publish first) — flagging here so it's not forgotten in the merge order.

LGTM, approving. Ramboz already approved at the current HEAD — this seconds it.

PR #2657, opened 2026-06-22
feat: paid-gated self-serve site-only onboarding endpoint | LLMO-5606

igor-grubic:
## Summary

Adds **`POST /v2/orgs/:spaceCatId/llmo/onboard-site`** — Piece 1 of self-serve domain onboarding ([LLMO-3749]). It stands up the **site entity only** ("nothing DRS"): no brand entity, no prompt generation, no brand-presence schedule, no `llmo-customer-analysis`. Activating a brand + generating prompts is Piece 2 ([LLMO-5605]).

Today onboarding is admin-only (`/onboard-llmo`, `POST /llmo/onboard`). This adds a paid-customer self-serve entry point that reuses the canonical `performLlmoOnboarding` via a new `siteOnly` flag.

## What changed

- **`llmo-onboarding.js`** — `performLlmoOnboarding` gains a `siteOnly` flag. The brand/prompt-gen block (v2: brand collision + `upsertBrand` + Brandalf; v1: `submitPromptGenerationJob`) is **extracted into a new `activateBrandAndGeneratePrompts`** (the Piece-2 seam) and gated behind `!siteOnly`. `ensureInitialCustomerConfigV2` + the `brandalf` feature flag stay in the always-run path. `siteOnly` drops `llmo-customer-analysis` from the enable list and triggers `[...BASIC_AUDITS, 'llm-error-pages', 'wikipedia-analysis']` directly (the DRS→SNS path that normally fires those is skipped). **Non-`siteOnly` behavior is unchanged.**
- **`llmo.js`** — new `onboardSiteOnly` handler.
- **`routes/index.js`** — the route (`:spaceCatId` UUID validation already exists in `index.js`).
- **`docs/openapi/*`** + regenerated `docs/index.html`.

## Auth — PAID + admin gate

In order: org **membership** (`hasAccess(org)`) → explicit **PAID** LLMO entitlement via `TierClient...checkValidEntitlement()` (`getTier() === PAID`; a FREE_TRIAL org 403s — no `getStatus()` check, it's an unbacked stub) → **brand-management (admin)** access (`hasAdminAccess() || isLLMOAdministrator()`).

## Execution model — ⚠️ deviation from the ticket

The ticket describes a sync-crucial-bits + deferred-background split. spacecat-api-service is a **synchronous Lambda** with no reliable in-process fire-and-forget (the env freezes after the response), and the only codebase deferral path is SQS→audit-worker. By decision, this endpoint runs **fully synchronously** like the existing onboarding routine; `status` is always `processing` because the triggered audits run asynchronously (the UI polls site/audit readiness for completion). No new audit-worker handler, no AsyncJob.

Customer-facing failures are always **generic**; the real reason is logged and posted to the ops alerts channel (both success and failure). No customer Slack (`say` defaults to no-op).

## Responses
- **201** `{ siteId, organizationId, baseURL, dataFolder, status: 'processing' }`
- **403** not a member / not PAID / lacks admin · **404** org not found · **400** generic (validation / already onboarded)

## Testing
- **New** `test/controllers/llmo/onboard-site-only.test.js` (14 tests): happy path, all 403/404/400 branches, generic errors, ops alerts, `siteOnly: true` threaded, brand-profile agent **not** called.
- **New** `siteOnly` orchestration tests in `llmo-onboarding.test.js` (v2 + v1): assert **no brand, no DRS job, `llmo-customer-analysis` neither enabled nor triggered**, while v2 config + brandalf flag still run.
- `npm run lint`, the full `llmo`/`llmo-onboarding` suites, and `docs:lint`/`docs:build` all pass.
- **Integration tests (`test/it/`) are deferred** to a follow-up.

## Non-goals (Piece 2 / [LLMO-5605])
No prompt generation / Brandalf / DRS job · no brand entity · no brand-presence schedule · no `llmo-customer-analysis` · no `triggerBrandProfileAgent`.

[LLMO-3749]: https://jira.corp.adobe.com/browse/LLMO-3749
[LLMO-5605]: https://jira.corp.adobe.com/browse/LLMO-5605

### DIFF
--- a/docs/openapi/llmo-api.yaml
+++ b/docs/openapi/llmo-api.yaml
@@ -1248,36 +1248,23 @@ llmo-onboard-site:
       response `status` is always `processing` because the triggered audits run
       asynchronously — the UI reflects completion by polling site/audit readiness.
 
-      Access requires all of: organization membership, an explicit **PAID** LLMO
-      entitlement (a FREE_TRIAL org is rejected), and brand-management (admin)
-      access. Customer-facing failures are generic; the internal reason is logged
-      and posted to the ops alerts channel.
+      Access requires organization membership and an explicit **PAID** LLMO
+      entitlement (a FREE_TRIAL org is rejected) — mirroring the v2 brand-management
+      routes; no platform-admin claim is required. Customer-facing failures are
+      generic; the internal reason is logged and posted to the ops alerts channel.
+
+      **Note — domain ownership is not yet verified.** This endpoint does not prove
+      the caller controls the submitted domain (a brand-new domain is accepted; only
+      a domain already claimed by a different org is rejected). DNS/file-based
+      ownership verification (planned via Google Search Console) is a future,
+      cross-repo item and out of scope for Piece 1.
     operationId: onboardSiteOnly
     requestBody:
       required: true
       content:
         application/json:
           schema:
-            type: object
-            properties:
-              domain:
-                type: string
-                format: hostname
-                description: The domain name to onboard (without protocol).
-                example: "example.com"
-              brandName:
-                type: string
-                description: |
-                  The brand label for the site's LLMO config (set via
-                  `siteConfig.updateLlmoBrand`). This is NOT a brand entity.
-                example: "Example Brand"
-              deliveryType:
-                type: string
-                description: Optional delivery type forwarded to site creation.
-                example: "aem_edge"
-            required:
-              - domain
-              - brandName
+            $ref: './schemas.yaml#/LlmoOnboardSiteRequest'
     responses:
       '201':
         description: |
@@ -1286,38 +1273,7 @@ llmo-onboard-site:
         content:
           application/json:
             schema:
-              type: object
-              properties:
-                siteId:
-                  type: string
-                  format: uuid
-                  description: The created or existing site ID.
-                  example: "987fcdeb-51a2-43d1-9f12-345678901234"
-                organizationId:
-                  type: string
-                  format: uuid
-                  description: The organization ID.
-                  example: "123e4567-e89b-12d3-a456-426614174000"
-                baseURL:
-                  type: string
-                  format: uri
-                  description: The normalized base URL of the onboarded site.
-                  example: "https://example.com"
-                dataFolder:
-                  type: string
-                  description: The LLMO data folder for the site.
-                  example: "example-com"
-                status:
-                  type: string
-                  enum: ["processing"]
-                  description: Always `processing` — triggered audits run asynchronously.
-                  example: "processing"
-              required:
-                - siteId
-                - organizationId
-                - baseURL
-                - dataFolder
-                - status
+              $ref: './schemas.yaml#/LlmoOnboardSiteResponse'
       '400':
         $ref: './responses.yaml#/400'
       '403':
--- a/docs/openapi/llmo-api.yaml
+++ b/docs/openapi/llmo-api.yaml
@@ -1221,6 +1221,115 @@ llmo-onboard:
         $ref: './responses.yaml#/500'
     security: []
 
+llmo-onboard-site:
+  parameters:
+    - name: spaceCatId
+      in: path
+      required: true
+      description: SpaceCat Organization ID (UUID)
+      schema:
+        type: string
+        format: uuid
+        example: 'a1b2c3d4-5678-90ab-cdef-1234567890ab'
+  post:
+    tags:
+      - llmo
+    summary: Paid self-serve, site-only onboarding (LLMO-5606)
+    description: |
+      Paid-gated, self-serve onboarding that stands up the **site entity only** —
+      "nothing DRS": no brand entity, no prompt generation, no brand-presence
+      schedule, and no `llmo-customer-analysis`. Activating a brand and generating
+      prompts is a separate step (LLMO-5605).
+
+      Synchronously creates the organization/site, a `SiteEnrollment` under the
+      org's existing PAID LLMO entitlement, the base LLMO config (brand label +
+      data folder), SharePoint folder, `helix-query.yaml`, CDN/override detection,
+      and enables + triggers the site-analysis audits and top-pages import. The
+      response `status` is always `processing` because the triggered audits run
+      asynchronously — the UI reflects completion by polling site/audit readiness.
+
+      Access requires all of: organization membership, an explicit **PAID** LLMO
+      entitlement (a FREE_TRIAL org is rejected), and brand-management (admin)
+      access. Customer-facing failures are generic; the internal reason is logged
+      and posted to the ops alerts channel.
+    operationId: onboardSiteOnly
+    requestBody:
+      required: true
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
+              domain:
+                type: string
+                format: hostname
+                description: The domain name to onboard (without protocol).
+                example: "example.com"
+              brandName:
+                type: string
+                description: |
+                  The brand label for the site's LLMO config (set via
+                  `siteConfig.updateLlmoBrand`). This is NOT a brand entity.
+                example: "Example Brand"
+              deliveryType:
+                type: string
+                description: Optional delivery type forwarded to site creation.
+                example: "aem_edge"
+            required:
+              - domain
+              - brandName
+    responses:
+      '201':
+        description: |
+          Site onboarded. The site, entitlement/enrollment, and base config exist;
+          site-analysis audits run asynchronously (`status: processing`).
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                siteId:
+                  type: string
+                  format: uuid
+                  description: The created or existing site ID.
+                  example: "987fcdeb-51a2-43d1-9f12-345678901234"
+                organizationId:
+                  type: string
+                  format: uuid
+                  description: The organization ID.
+                  example: "123e4567-e89b-12d3-a456-426614174000"
+                baseURL:
+                  type: string
+                  format: uri
+                  description: The normalized base URL of the onboarded site.
+                  example: "https://example.com"
+                dataFolder:
+                  type: string
+                  description: The LLMO data folder for the site.
+                  example: "example-com"
+                status:
+                  type: string
+                  enum: ["processing"]
+                  description: Always `processing` — triggered audits run asynchronously.
+                  example: "processing"
+              required:
+                - siteId
+                - organizationId
+                - baseURL
+                - dataFolder
+                - status
+      '400':
+        $ref: './responses.yaml#/400'
+      '403':
+        $ref: './responses.yaml#/403'
+      '404':
+        $ref: './responses.yaml#/404-organization-not-found-with-id'
+      '500':
+        $ref: './responses.yaml#/500'
+    security:
+      - ims_key: []
+      - api_key: []
+
 llmo-offboard:
   parameters:
     - $ref: './parameters.yaml#/siteId'
--- a/src/controllers/llmo/llmo-onboarding.js
+++ b/src/controllers/llmo/llmo-onboarding.js
@@ -1245,6 +1245,178 @@ export async function enqueueLlmoOnboardingPublish(context, site, dataFolder) {
 }
 
 /**
+ * Activates the brand and kicks off prompt generation for an onboarded site.
+ *
+ * This is the brand/prompt-generation half of onboarding, owned by Piece 2
+ * (LLMO-5605). For v2 it writes the initial brand to the normalized brands table
+ * and triggers the Brandalf DRS job; for v1 it submits the legacy DRS
+ * prompt-generation job directly. Site-only onboarding (LLMO-5606) does NOT call
+ * this — it stands up the site with no brand entity, no Brandalf job, and no
+ * prompt-generation job.
+ *
+ * @param {object} params
+ * @param {string} params.onboardingMode - Resolved LLMO onboarding mode (v1/v2).
+ * @param {object} params.organization - The organization model.
+ * @param {object} params.site - The site model.
+ * @param {object} params.siteConfig - The (already-saved) site config object.
+ * @param {string} params.brandName - Brand name (label).
+ * @param {string} params.imsOrgId - IMS org ID.
+ * @param {string} params.baseURL - Site base URL.
+ * @param {string} [params.region] - Optional ISO 3166-1 alpha-2 region.
+ * @param {object} params.context - The request context.
+ * @param {Function} [params.say] - Optional Slack say callback.
+ * @returns {Promise<void>}
+ */
+export async function activateBrandAndGeneratePrompts({
+  onboardingMode,
+  organization,
+  site,
+  siteConfig,
+  brandName,
+  imsOrgId,
+  baseURL,
+  region,
+  context,
+  say = () => {},
+}) {
+  const { log } = context;
+
+  if (onboardingMode === LLMO_ONBOARDING_MODE_V2) {
+    const postgrestClient = context.dataAccess?.services?.postgrestClient;
+
+    // Write initial brand to normalized brands table so DRS prompt sync can
+    // find it via GET /v2/orgs/{orgId}/brands before the Brandalf job finishes.
+    // Brandalf will upsert over this with LLM-identified sub-brands later.
+    // Use baseURL (matches sites.base_url) so syncBrandSites can link the brand
+    // to the site. overrideBaseURL may differ (e.g., www.blick.ch vs blick.ch)
+    // and would fail the exact-match lookup against the sites table.
+    // LLMO-5556: a second site onboarded under an existing brand name must not
+    // re-point that brand's primary site, nor clobber its URLs/aliases via the
+    // full-replace syncs inside upsertBrand. Detect the collision and skip the
+    // initial-brand write — the brand already exists so DRS sync still resolves
+    // it; a human then decides whether the new site is a sub-brand or a URL.
+    try {
+      const { data: existingBrand, error: lookupError } = await postgrestClient
+        .from('brands')
+        .select('id, site_id')
+        .eq('organization_id', organization.getId())
+        .eq('name', brandName.trim())
+        .maybeSingle();
+
+      if (lookupError) {
+        // Fail closed (LLMO-5556): PostgREST returns { data: null, error } on a
+        // query failure rather than throwing, so without this check a transient
+        // failure would fall through to upsertBrand and could re-point an
+        // existing brand's primary site. Skip the write as a precaution.
+        log.warn(`Skipping initial brand write: failed to look up existing brand "${brandName.trim()}" `
+          + `(org ${organization.getId()}): ${lookupError.message}`);
+      } else if (existingBrand?.site_id && existingBrand.site_id !== site.getId()) {
+        log.warn(`Skipping initial brand write: brand "${brandName.trim()}" `
+          + `(org ${organization.getId()}) already exists with a different primary site `
+          + `(existing=${existingBrand.site_id}, onboarding=${site.getId()}). `
+          + `Add ${baseURL} as a brand URL or onboard under a distinct brand name.`);
+      } else {
+        // No collision: proceed with upsert (new brand, existing brand with a
+        // null primary site, or a re-onboard of the same site). upsertBrand's
+        // own guard keeps an already-set site_id immutable on the last case.
+        // LLMO-5645: seed operator market when supplied; else the 'gl'
+        // placeholder (consistent with the V2 config + brandAliases, both
+        // overwritten by Brandalf's async result).
+        const stubRegions = onboardingStubRegions(region);
+        await upsertBrand({
+          organizationId: organization.getId(),
+          brand: {
+            name: brandName.trim(),
+            status: 'active',
+            baseSiteId: site.getId(),
+            region: stubRegions,
+            urls: [{ value: baseURL, type: 'base' }],
+            brandAliases: [{ name: brandName.trim(), regions: stubRegions }],
+          },
+          postgrestClient,
+          updatedBy: 'llmo-onboarding',
+          log,
+        });
+        log.info(`Created initial brand "${brandName}" in normalized table for site ${site.getId()}`);
+      }
+    } catch (brandError) {
+      log.warn(`Failed to create initial brand in normalized table: ${brandError.message}`);
+    }
+
+    // Trigger Brandalf immediately after the v2 config exists so downstream
+    // brand sync can attach results to the newly created organization.
+    try {
+      const drsClient = DrsClient.createFrom(context);
+      if (drsClient.isConfigured()) {
+        const companyWebsite = siteConfig.getFetchConfig?.()?.overrideBaseURL || baseURL;
+        await triggerBrandalfOnboardingJob({
+          drsClient,
+          organizationId: organization.getId(),
+          siteId: site.getId(),
+          imsOrgId,
+          brandName: brandName.trim(),
+          companyWebsite,
+          onboardingMode,
+          region,
+          log,
+          say,
+        });
+      } else {
+        log.debug('DRS client not configured, skipping Brandalf flow');
+      }
+    } catch (drsError) {
+      log.error(`Failed to start DRS Brandalf flow: ${drsError.message}`);
+      say(':warning: Failed to start DRS Brandalf flow (will need manual trigger)');
+    }
+  } else {
+    // V1 has no Brandalf trigger, so DRS will not submit prompt generation
+    // automatically. Submit it directly here so v1 onboardings still get
+    // prompts written to the legacy LLMO config (LLMO-4534).
+    try {
+      const drsClient = DrsClient.createFrom(context);
+      if (drsClient.isConfigured()) {
+        const trimmedBrand = brandName.trim();
+        const brandProfile = siteConfig.getBrandProfile?.();
+        // LLMO-4534: v1 fallback audience is English-only. v2 onboardings get a
+        // locale-aware audience from brand_profile.main_profile.target_audience.
+        const audience = brandProfile?.main_profile?.target_audience
+          || `General consumers interested in ${trimmedBrand} products and services`;
+
+        // The audit-worker `drs-prompt-generation` handler takes the legacy LLMO
+        // config write path when `onboarding_mode` is absent from the DRS job
+        // metadata. Do NOT pass `onboarding_mode` here — adding it would route v1
+        // prompts to the v2 customer-config storage and break v1 onboardings.
+        //
+        // LLMO-4683: forward operator-supplied `region` so the GPT prompt-generation
+        // job conditions on the brand's market. Omitted → DRS client default ('US')
+        // applies, preserving prior behavior.
+        if (region) {
+          log.info(`Using operator-supplied region "${region}" for v1 DRS prompt generation`);
+        }
+        const drsJob = await drsClient.submitPromptGenerationJob({
+          baseUrl: siteConfig.getFetchConfig?.()?.overrideBaseURL || baseURL,
+          brandName: trimmedBrand,
+          audience,
+          siteId: site.getId(),
+          imsOrgId,
+          ...(region ? { region } : {}),
+        });
+        if (!drsJob?.job_id) {
+          throw new Error('DRS submitPromptGenerationJob returned no job_id');
+        }
+        log.info(`Started DRS prompt generation: job=${drsJob.job_id}`);
+        say(`:robot_face: Started DRS prompt generation job: ${drsJob.job_id}`);
+      } else {
+        log.debug('DRS client not configured, skipping prompt generation');
+      }
+    } catch (drsError) {
+      log.error(`Failed to start DRS prompt generation: ${drsError.message}`);
+      say(`:warning: Failed to start DRS prompt generation for site ${site.getId()} (will need manual trigger)`);
+    }
+  }
+}
+
+/**
  * Complete LLMO onboarding process.
  * @param {object} params - Onboarding parameters
  * @param {string} [params.domain] - The domain name (alternative to baseURL)
@@ -1256,6 +1428,10 @@ export async function enqueueLlmoOnboardingPublish(context, site, dataFolder) {
  *   HTTP clients set this via the `temp-onboarding` body field.
  * @param {string} [params.region] - Optional ISO 3166-1 alpha-2 region code forwarded to V1 DRS
  *   prompt generation. Omitted → DRS client default ('US') applies.
+ * @param {boolean} [params.siteOnly=false] - Site-only onboarding (LLMO-5606). When true, stands
+ *   up the site, entitlement/enrollment, config, and site-analysis audits, but skips the entire
+ *   brand activation + prompt-generation block (no brand entity, no Brandalf job, no v1
+ *   prompt-generation job) and never enables/triggers llmo-customer-analysis.
  * @param {object} context - The request context
  * @param {Function} [say] - Optional function to send progress messages
  * @returns {Promise<object>} Onboarding result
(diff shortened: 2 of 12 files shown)


MysticatBot (changes requested):
Hey @igor-grubic,

**Verdict: Request changes** - three fixable issues in the new handler's error paths and input validation.
_Changes: adds a paid-gated self-serve site-only onboarding endpoint (POST /v2/orgs/:spaceCatId/llmo/onboard-site) that reuses performLlmoOnboarding with a new siteOnly flag, extracting brand activation into a separate seam for Piece 2 (10 files)._

## Must fix before merge

1. [Important] `postLlmoAlert` can throw in catch block, preventing the 500 response from returning - `src/controllers/llmo/llmo.js:~1175`
   If the Slack/alert service is down, `postLlmoAlert` throws an unhandled rejection and the customer never receives the generic 500 response. Wrap in a defensive try/catch so `internalServerError` always returns.

2. [Important] No format/type/length validation on `domain` and `brandName` for this customer-facing endpoint - `src/controllers/llmo/llmo.js:~1132`
   Only truthiness is checked. Non-string types (arrays, objects), arbitrarily long strings, whitespace-only strings, and domains with protocol prefixes all pass. The OpenAPI spec declares `format: hostname` but the server does not enforce it. Add `typeof` checks, `.trim()`, length caps (253 chars per RFC 1035 for domain), and reject protocol prefixes.

3. [Important] `checkValidEntitlement()` exceptions become generic 500 instead of appropriate 403 - `src/controllers/llmo/llmo.js:~1115`
   `TierClient.checkValidEntitlement()` can throw (e.g., `UnauthorizedProductError` or transient DB failure). Because it is inside the outer try/catch, any throw becomes a generic 500 rather than the correct 403. Catch entitlement-lookup failures specifically and return the appropriate status code.

4. [Important] Missing integration tests per CLAUDE.md rule: "New or modified endpoints must include integration tests in test/it/" - acknowledged as deferred in PR body, but the documented convention is explicit.

<details>

- nit: OpenAPI response schema is inlined in llmo-api.yaml rather than defined in schemas.yaml per the project's OpenAPI guidelines - `docs/openapi/llmo-api.yaml:~1296`
- suggestion: validate `deliveryType` against a known allowlist or document accepted values in the OpenAPI enum - `src/controllers/llmo/llmo.js:~1135`
- nit: confirm that src/index.js UUID validation covers the `:spaceCatId` param name (the existing validation targets `:siteId` and `:organizationId` by name) - `src/routes/index.js`
- suggestion: extract `[...BASIC_AUDITS, 'llm-error-pages', 'wikipedia-analysis']` into a named constant to avoid drift between `auditsToEnable` and `auditsToTrigger` in the siteOnly path - `src/controllers/llmo/llmo-onboarding.js:~1494`

</details>

igor-grubic:
Thanks @MysticatBot — triaged each point. Pushed `c7cdc0fe`.

**Fixed**
- **#2 (input validation):** `onboardSiteOnly` now validates `domain`/`brandName` as non-empty strings (`hasText`), caps lengths (253 for domain per RFC 1035, 256 for brandName), and rejects domains that don't normalize to a valid URL. Added unit tests for each branch.
- **Suggestion (audit-list drift):** extracted `SITE_ONLY_AUDITS` so the siteOnly enable and trigger lists share one source.

**Respectfully declining — verified false positives**
- **#1 (`postLlmoAlert` throws in catch → no 500):** `postLlmoAlert` already wraps `postSlackMessage` in its own try/catch and only logs (`llmo-onboarding.js`), so it can't throw — the generic 500 always returns. No change needed.
- **#3 (`checkValidEntitlement` throw → 500 not 403):** for an unentitled / FREE_TRIAL org it returns `{}` / `{ entitlement }` (no throw), so the handler already returns **403** via the `!entitlement || getTier() !== PAID` check. It only re-throws on a genuine lookup/DB error, where a **500 is the correct** response — no "not entitled" path produces a 500.

**Other**
- **#4 (integration tests):** deferred by decision for this PR; tracked as the follow-up. The fully-synchronous execution model is also a deliberate deviation from the spec (called out in the PR description).
- nit (`:spaceCatId` UUID validation): already enforced in `src/index.js` (validates `params.spaceCatId` via `isValidUUID`).
- nit (inline OpenAPI schema): kept consistent with the sibling `llmo-onboard` block, which also inlines its response schema.
- suggestion (`deliveryType` allowlist): validated downstream by the Site data model on create.

MysticatBot (approved):
Hey @igor-grubic,

**Verdict: Approve** - clean, well-gated self-serve endpoint with thorough follow-up addressing prior findings.
_Changes: adds a paid-gated self-serve site-only onboarding endpoint (POST /v2/orgs/:spaceCatId/llmo/onboard-site) that reuses performLlmoOnboarding with a new siteOnly flag, extracting brand activation into a separate seam for Piece 2 (12 files)._

Note: CI integration tests (it-postgres) are still pending - verify they pass before merge.

<details>

- suggestion: express the full-onboarding auditsToEnable list relative to SITE_ONLY_AUDITS (e.g. `[...SITE_ONLY_AUDITS, 'llmo-customer-analysis']`) to make the difference self-documenting and prevent future drift - `src/controllers/llmo/llmo-onboarding.js:~1501`
- nit: `ensureInitialCustomerConfigV2` + brandalf flag run unconditionally in siteOnly mode; if Piece 2 is never activated, the flag is set for an org that never uses it - consider a code comment clarifying the contract ("Piece 2 assumes this was already set by Piece 1") - `src/controllers/llmo/llmo-onboarding.js:~1567`
- nit: `!data || typeof data !== 'object'` body check is defensive given the framework already parses JSON bodies, but harmless - `src/controllers/llmo/llmo.js:~1126`

</details>

- Input validation (domain/brandName): hasText + length caps + URL format added (c7cdc0fe)
- Integration tests: auth gate tests added against real PostgREST (b8ef5910)
- SITE_ONLY_AUDITS constant extracted to prevent enable/trigger list drift
- Prior finding about `postLlmoAlert` throwing in catch: verified - already wraps in try/catch, cannot propagate
- Prior finding about `checkValidEntitlement` exceptions: verified - returns data for expected paths, only throws on infra failure where 500 is correct

rainer-friederich:
Reviewed against the Piece-1 spec (project-elmo-ui issue 2092). Implementation is faithful and the "nothing DRS / nothing brand" invariants are well tested. Two things to flag before the self-serve UI ships against this — neither blocks correctness, but both deserve a conscious sign-off.

## PAID gate — admin requirement may lock out actual self-serve customers

The third auth gate requires `hasAdminAccess() || isLLMOAdministrator()`. For an IMS customer token these resolve to platform-level claims:

- `hasAdminAccess()` → for IMS returns `authInfo.isAdmin()` — the platform `is_admin` flag (`access-control-util.js:87-94`).
- `isLLMOAdministrator()` → `authInfo.isLLMOAdministrator() && !_lastAccessWasDelegated` (`access-control-util.js:192`).

Consequences worth confirming:

1. A paying customer who is an ordinary org member but is **not** a platform admin and is **not** flagged `is_llmo_administrator` gets a **403 despite being PAID**.
2. The delegation clause means a customer whose org access comes via a **delegation grant** has `_lastAccessWasDelegated === true`, so `isLLMOAdministrator()` returns `false` even if they hold the claim — and `isAdmin()` is also false — so they are locked out.

This is faithful to the spec's explicit decision (inherit brand-management's gate, enforced server-side), and it's correct *if* every intended self-serve user really carries `is_llmo_administrator`. But that equivalence to the UI's `canManageConfiguration` is asserted, not proven, and the unit tests stub these predicates to `true`, so the gate isn't exercised against a realistic token.

**Ask:** validate end-to-end with a real non-admin paying-customer token (ideally a delegated one) that the gate actually opens, before the Piece-1/Piece-2 UI depends on it. Otherwise this risks being "approved, green, and unusable by the target user."

## Note — brandalf feature flag is set org-wide on a site-only onboard

In v2 mode, the site-only path still runs `ensureInitialCustomerConfigV2` + flips the org-level `brandalf` flag to `true` (deliberately kept in the always-run path per Design Decision 2). For a brand-new paid org this is harmless scaffolding. For an **existing** org it is not strictly a no-op — it is the same write full onboarding already performs, so there's no regression here, but flagging so it's known:

- If a customer onboards a site-only domain and never proceeds to Piece 2, the org carries `brandalf = true` with no v2 brand/prompts/schedule behind it.
- Worth a one-line confirmation that flipping `brandalf` on an org that has pre-existing v1 brands does not change those brands' prompt-resolution / scheduling behavior.

Just recording it for visibility — no change requested.

src/routes/required-capabilities.js
```diff
@@ -123,6 +123,7 @@ export const INTERNAL_ROUTES = [
   'GET /sites/:siteId/llmo/strategy/demo/brand-presence',
   'GET /sites/:siteId/llmo/strategy/demo/recommendations',
   'POST /llmo/onboard',
+  'POST /v2/orgs/:spaceCatId/llmo/onboard-site',
```
rainer-friederich, line 126: This is about self-onboarding. So this could use an org scoped IMS s2s token.

src/controllers/llmo/llmo-onboarding.js
```diff
@@ -1244,6 +1251,178 @@ export async function enqueueLlmoOnboardingPublish(context, site, dataFolder) {
...
+        });
+      } else {
+        log.debug('DRS client not configured, skipping Brandalf flow');
+      }
+    } catch (drsError) {
+      log.error(`Failed to start DRS Brandalf flow: ${drsError.message}`);
+      say(':warning: Failed to start DRS Brandalf flow (will need manual trigger)');
+    }
+  } else {
+    // V1 has no Brandalf trigger, so DRS will not submit prompt generation
```
rainer-friederich, line 1379: Do we still need to support this?

rainer-friederich (changes requested):
Otherwise See /pr-review review

src/controllers/llmo/llmo.js
```diff
@@ -1056,6 +1057,142 @@ function LlmoController(ctx) {
...
+        EntitlementModel.PRODUCT_CODES.LLMO,
+      );
+      const { entitlement } = await tierClient.checkValidEntitlement();
+      if (!entitlement || entitlement.getTier() !== EntitlementModel.TIERS.PAID) {
+        return forbidden('A paid LLMO entitlement is required to onboard a site');
+      }
+
+      // Brand-management (admin) access — the backend equivalent of the UI's
+      // canManageConfiguration. Self-serve onboarding inherits that access model.
+      if (!accessControlUtil.hasAdminAccess() && !accessControlUtil.isLLMOAdministrator()) {
```
rainer-friederich: issue (blocking): [Critical] This admin gate admits no real customer IMS token. `isLLMOAdministrator()` reads `profile.is_llmo_administrator` and `hasAdminAccess()` -> `isAdmin()` reads `profile.is_admin`; both are JWT-only claims. The IMS handler builds the profile from the raw token (`transformProfile`, spacecat-shared `handlers/ims.js:108`) and never sets them, so both are falsy for a customer IMS token -> every real paying customer 403s here. The endpoint is reachable only by Adobe platform/LLMO admins.

The PR's stated rationale (backend equivalent of the UI's `canManageConfiguration`) is also contradicted by this repo: every v2 brand-management write route - `createBrandForOrg`/`updateBrandForOrg`/`deleteBrandForOrg` (brands.js:1372/1519/1684) and all category/topic/prompt CRUD - gates on `hasAccess(organization)` ALONE, with no admin predicate.

Decide before merge: either (a) drop the admin predicate and gate on `hasAccess(organization)` + the explicit PAID check (matching brands.js), or (b) if admin-operated is the real intent, correct the 'paid self-serve, customer-facing' description in the OpenAPI summary and PR. Either way, add an end-to-end test with a real customer IMS token asserting 201 - the current suite is green only because it stubs these predicates to true.

src/controllers/llmo/llmo.js
```diff
@@ -1056,6 +1057,142 @@ function LlmoController(ctx) {
...
+      // 253 chars; brandName is a label, so cap it defensively too.
+      if (domain.trim().length > 253) {
+        return badRequest('domain is too long');
+      }
+      if (brandName.trim().length > 256) {
+        return badRequest('brandName is too long');
+      }
+
+      const baseURL = composeBaseURL(domain.trim());
+      if (!isValidUrl(baseURL)) {
```
rainer-friederich, line 1143: issue (blocking): [Important] SSRF surface. `domain` is validated only by `hasText`, length <= 253, and `isValidUrl(composeBaseURL(domain))`. `composeBaseURL` keeps the host, and `isValidUrl` accepts any http(s) URL, so `169.254.169.254`, `localhost`, and internal hostnames pass. That host then flows to `detectCdnForDomain(new URL(baseURL).hostname, ...)` (llmo-onboarding.js:1543), which performs DNS resolution and an outbound `fetch` from the api-service Lambda's network position; timing/error differences plus the ops-Slack alert form a reachability oracle. This path was previously admin-only via `onboardCustomer`; this endpoint is the seam that exposes it to paid customers once the gate (finding 1) is opened.

Fix: before any side effect, reject IP-literal hosts and hosts resolving to private/loopback/link-local/metadata ranges. Minimal in-repo guard: assert `getDomain(host)` (already imported from `tldts`) is non-null and the host is not an IP literal / not `localhost`. Add a negative test.

src/controllers/llmo/llmo.js
```diff
@@ -1056,6 +1057,142 @@ function LlmoController(ctx) {
...
+        return badRequest('domain is invalid');
+      }
+      const dataFolder = generateDataFolder(baseURL, env.ENV);
+      const imsOrgId = organization.getImsOrgId();
+
+      log.info(`Starting site-only LLMO onboarding for org ${spaceCatId} (IMS ${imsOrgId}), domain ${domain}`);
+
+      // validateSiteNotOnboarded returns { isValid, error } (never throws) and
+      // already ops-alerts the conflict cases. Surface only a generic 400.
+      const validation = await validateSiteNotOnboarded(baseURL, imsOrgId, dataFolder, context);
```
rainer-friederich, line 1165: issue (blocking): [Important] No proof-of-ownership. `validateSiteNotOnboarded` only blocks a domain already claimed by a *different* org; a brand-new domain returns `{ isValid: true }`. So a qualifying caller can onboard arbitrary domains it has no relationship with - each call creates a SpaceCat site + entitlement enrollment + SharePoint folder + a `helix-query.yaml` commit to `adobe/project-elmo-ui-data` + audit/Ahrefs enqueues + an outbound probe to that domain. It also enables a first-claim race: once org A onboards `victim.com`, the real owner is blocked ('assigned to a different organization').

Full DNS/file ownership verification is a cross-repo design item (track separately). In-repo, add at least a per-org/per-caller rate limit and gate the external writes behind it.

src/controllers/llmo/llmo-onboarding.js
```diff
@@ -1388,143 +1579,40 @@ export async function performLlmoOnboarding(params, context, say = () => {}) {
...
-          log.debug('DRS client not configured, skipping prompt generation');
-        }
-      } catch (drsError) {
-        log.error(`Failed to start DRS prompt generation: ${drsError.message}`);
-        say(`:warning: Failed to start DRS prompt generation for site ${site.getId()} (will need manual trigger)`);
-      }
+    // Brand activation + prompt generation. This block is owned by Piece 2
+    // (LLMO-5605). Site-only onboarding (LLMO-5606) skips it entirely — no brand
+    // entity (upsertBrand), no Brandalf job, and no v1 prompt-generation job.
+    // ensureInitialCustomerConfigV2 + the brandalf feature flag deliberately stay
```
rainer-friederich, line 1589: issue (blocking): [Important] Org-wide side effect on a site-only onboard. `ensureInitialCustomerConfigV2` + `upsertFeatureFlag(brandalf=true)` sit in the always-run (pre-`siteOnly`) block and are keyed by `organization.getId()` (org-scoped, not site-scoped). For a brand-new org this is benign, but for an org that ALREADY has v1 brands actively scheduled in DRS, a site-only onboard flips that org v1->v2 prompt selection in the DRS scheduler - an org-level behavior change for every existing brand, triggered by adding one site, in tension with Piece 1's 'zero DRS process' contract.

Fix: only flip the flag when the org has no pre-existing v2 config/brands, or, if the always-run placement is deliberate, ops-alert when the flag transitions false->true on an org with existing enrollments and document the blast radius. (Also flagged by MysticatBot and a prior comment; still unresolved in code.)

test/controllers/llmo/onboard-site-only.test.js
```diff
@@ -0,0 +1,346 @@
...
+      const body = await res.json();
+      expect(body.message).to.equal(GENERIC_ERROR);
+      expect(body.message).to.not.contain('already onboarded');
+      expect(performLlmoOnboardingStub).to.not.have.been.called;
+    });
+  });
+
+  describe('orchestration failure', () => {
+    it('returns a generic 500 and posts the real reason to ops', async () => {
+      performLlmoOnboardingStub.rejects(new Error('SharePoint folder creation exploded'));
```
rainer-friederich, line 337: issue (blocking): [Important] Test coverage gaps on a paid customer-facing path.

1) The 'nothing DRS / nothing brand' guarantee - this PR's entire reason for existing - is asserted only via esmock 'not called' spies in `llmo-onboarding.test.js`. That is the weakest regression guard: a refactor of the audit-list construction could pass the spies silently. If the IT harness can stub only the external HTTP boundaries while keeping the real DB, add one IT asserting the post-onboard Configuration row has `llm-error-pages` enabled and `llmo-customer-analysis` disabled.

2) The siteOnly partial-failure rollback is untested: `performLlmoOnboarding`'s catch runs `deleteSharePointFolder` + `revokeEnrollment` then rethrows (llmo-onboarding.js:1626-1637), but here `performLlmoOnboarding` is stubbed to reject, so the real rollback never runs on the siteOnly branch. A half-provisioned paid customer (entitlement + SharePoint folder, then a later throw) must roll back - add one test driving a mid-flight failure with `siteOnly:true`.

3) `enqueueLlmoOnboardingPublish`'s swallow-on-failure (llmo-onboarding.js:1483-1487) is untested - assert the happy path enqueues exactly one publish message and that a failed enqueue does not abort onboarding.

docs/openapi/llmo-api.yaml
```diff
@@ -1221,6 +1221,115 @@ llmo-onboard:
         $ref: './responses.yaml#/500'
     security: []
 
+llmo-onboard-site:
```
rainer-friederich, line 1224: issue (blocking): [Important] The request body and 201 response schemas are inlined here. CLAUDE.md:348 states 'All schemas defined in `docs/openapi/schemas.yaml` (reuse, don't inline)' and the 'Adding a New Endpoint' checklist (CLAUDE.md:542) repeats it; the rest of this same file follows the rule (`$ref: './schemas.yaml#/...'`). 'Matches the sibling `llmo-onboard` block' is not a defense - the sibling also violates it. Extract both schema objects to `schemas.yaml` and reference via `$ref`.

rainer-friederich (changes requested):
Hey @igor-grubic,

**Verdict: Request changes** - a careful, well-tested PR whose "nothing DRS" safety guarantee is sound, but the auth gate admits none of its stated audience and it newly fronts a provisioning path with an unguarded SSRF surface.

_Changes: adds a paid-gated, self-serve site-only LLMO onboarding endpoint (`POST /v2/orgs/:spaceCatId/llmo/onboard-site`) reusing `performLlmoOnboarding` via a new `siteOnly` flag, extracting brand activation into a Piece-2 seam (12 files)._

_Note: the only failing check is the advisory `Mysticat / pr-review` bot (it crashed on a cost budget, not a code failure); `ci / build`, `it-postgres`, and `codecov/patch` are green._

## Must fix before merge
1. [Critical] Admin gate admits no real customer IMS token, and contradicts the PR's own canManageConfiguration-parity rationale - `src/controllers/llmo/llmo.js:1121` (details inline)
2. [Important] SSRF: customer-controlled `domain` reaches DNS + outbound HTTP with no internal-host / IP-literal guard - `src/controllers/llmo/llmo.js:1143` (details inline)
3. [Important] No domain-ownership proof: a qualifying caller can onboard domains it does not control (external write amplification + first-claim race) - `src/controllers/llmo/llmo.js:1153` (details inline)
4. [Important] Org-wide `brandalf` flag flip on a site-only onboard changes v1 to v2 prompt resolution for every existing brand in the org - `src/controllers/llmo/llmo-onboarding.js:1589` (details inline)
5. [Important] Test gaps on a paid path: the no-side-effect guarantee is only mock-pinned, plus untested siteOnly rollback and publish-enqueue swallow - `test/controllers/llmo/onboard-site-only.test.js:331` (details inline)
6. [Important] OpenAPI schemas inlined instead of `$ref` into `schemas.yaml` (explicit CLAUDE.md rule) - `docs/openapi/llmo-api.yaml:1224` (details inline)

<details>
<summary>Non-blocking (9): minor issues and suggestions</summary>

- nit: PR description says integration tests are deferred, but the auth-gate IT is present - only the 201 happy path is deferred. Correct the description.
- nit: reject `domain` containing `/ ? # @` - a path-bearing domain passes `isValidUrl` and yields a multi-segment `dataFolder` (no traversal, but contradicts OpenAPI `format: hostname`) - `src/controllers/llmo/llmo.js:1143`
- nit: guard `result` before reading `result.baseURL` in the success path - a partial return would post a "failed" ops alert for a partial success - `src/controllers/llmo/llmo.js`
- suggestion: add `maxLength` (253 domain, 256 brandName) to the OpenAPI requestBody to match the handler caps - `docs/openapi/llmo-api.yaml`
- suggestion: extract `MAX_DOMAIN_LENGTH` / `MAX_BRAND_NAME_LENGTH` constants instead of magic numbers - `src/controllers/llmo/llmo.js`
- suggestion: surface "already onboarded" specifically (safe for self-serve) rather than the generic 400 - `src/controllers/llmo/llmo.js:1156`
- nit: the v1 skip log dropped the "and Brandalf flow" wording from the original line; operators grepping the old string will miss it - `src/controllers/llmo/llmo-onboarding.js:1583`
- nit: `region` is intentionally omitted for Piece 1 but the OpenAPI body does not say so vs Piece 2 - `docs/openapi/llmo-api.yaml`
- nit: trim the delegation doc-comment at `src/controllers/llmo/llmo.js:1099-1100` - for an Organization entity with no productCode the delegation flag is never set, so the ordering rationale does not apply here.

Out of scope, worth tracking: a customer-obtainable "manage LLMO config" claim is a spacecat-auth-service / IMS role decision; full domain-ownership verification (DNS TXT / file token) is cross-repo (project-elmo-ui issue 2092). One candidate finding (a "double `validateSiteNotOnboarded`" concern) was dropped after verifying `performLlmoOnboarding` does not call it internally.
</details>

igor-grubic:
Thanks @rainer-friederich — both flags make sense. Findings on each:

## 1. brandalf flag on existing v1 brands — confirmed safe (no change to their behavior)

The `brandalf` flag is the DRS scheduler's per-org v1/v2 switch, so the concern is legitimate — but this endpoint can't flip it for an org that already has v1 brands:

- We only set `brandalf=true` inside the **v2 branch** of `performLlmoOnboarding`.
- `resolveLlmoOnboardingMode` keeps **any org with a pre-Brandalf-cutoff site on v1** (the LLMO-4176 legacy-customer guard, `src/support/llmo-onboarding-mode.js`). An org with pre-existing v1 brands has pre-cutoff sites → resolves **v1** → we never enter the v2 branch → **`brandalf` is never flipped** for it.
- The only orgs we set `brandalf=true` for are already-v2 orgs (an idempotent re-write) or brand-new orgs with no pre-cutoff sites (hence no v1 brands to affect).

For the "onboard site-only and never proceed to Piece 2" case (a brand-new v2 org): the scheduler resolves v2, finds no brand prompts, falls back to the v1 config path, finds none either, and **skips the site** — nothing schedules until Piece 2 adds prompts. That's the intended "nothing DRS" state, not a regression.

→ Net: no change to any existing v1 brand's prompt-resolution or scheduling.

## 2. PAID + admin gate — agree it deserves a conscious sign-off; flagging to product, not changing in this PR

You're right: for an IMS customer token, `hasAdminAccess()` → `is_admin` and `isLLMOAdministrator()` → the `is_llmo_administrator` claim **and** `!_lastAccessWasDelegated` — so a PAID ordinary member (or a delegated customer, where `_lastAccessWasDelegated` flips it to `false`) gets a 403 despite paying.

The gate is faithful to the spec's explicit decision (inherit brand-management's `canManageConfiguration`, enforced server-side), and the unit tests stub these predicates — so the equivalence to the UI's gate is asserted, not proven against a realistic token. Before the self-serve UI depends on this, we'll:

- confirm with product/auth that the self-serve customer token actually carries `is_llmo_administrator`, and decide the delegated-customer case;
- run an end-to-end check with a real non-admin (ideally delegated) paying-customer token.

Leaving the gate as-is in this PR since it matches the agreed Piece-1 spec; tracking the validation as a follow-up.

igor-grubic:
Thanks @rainer-friederich — thorough pass, much appreciated. Pushed `c1ffefba` addressing the findings.

## [Critical] Admin gate — agreed and fixed
You're right, and I verified it: `transformProfile` (ims.js:108) never sets `is_admin`/`is_llmo_administrator`, and the IMS handler grants an admin *scope* only for `@adobe.com` platform admins (ims.js:267) — so the gate admitted **no** real customer token, and every v2 brand-management write route gates on `hasAccess(organization)` alone. **Dropped the admin predicate**; the gate is now `hasAccess(organization)` + the explicit PAID check, matching `brands.js`. Added a unit test asserting a non-admin member (no `is_admin`/`is_llmo_administrator`) gets 201. The real-customer-token **201 e2e** still can't run in the IT harness (the success path needs SharePoint/GitHub/Ahrefs) — tracking that as the e2e follow-up.

## [Important] SSRF — fixed
Added a guard before any side effect: reject IP-literal hosts (incl. `169.254.169.254`), `localhost`, and single-label/internal hosts via `tldts` (`isIp || !domain`), with negative tests. Resolve-time private-range checks remain a deeper follow-up (noted).

## [Important] OpenAPI schemas — fixed
Extracted the request + 201 response schemas to `schemas.yaml` and `$ref` them. You're right that "matches the sibling" wasn't a defense.

## [Important] brandalf org-wide flip — correcting my earlier reply
My earlier "confirmed safe" was over-stated; here's the precise version. The flip only runs in the **v2 branch**, and `resolveLlmoOnboardingMode` keeps any org with a **pre-cutoff site on v1** — so the realistic population (v1 brands ⇒ pre-cutoff sites) is never flipped. The **one** edge where it bites: an org onboarded v1 **during a kill-switch window** on *post-cutoff* sites (`brandalf=false`) — a later site-only onboard there resolves v2 and flips it. That's narrow, and it's **mooted by the planned v1 retirement next month**, so we're consciously not guarding it. (No regression vs. full onboarding either — same shared write.)

## [Important] No proof-of-ownership — deferring (documented)
We can't do real ownership verification in-repo right now; it's planned via **Google Search Console** (cross-repo) and is out of scope for Piece 1. Documented in the OpenAPI description and tracking as a follow-up. Not adding an in-repo rate limit for this PR.

## [Important] Test gaps — addressed
Added (2) a siteOnly **mid-flight rollback** test (asserts `deleteSharePointFolder` + `revokeEnrollment` run, then rethrow) and (3) an `enqueueLlmoOnboardingPublish` **swallow + exactly-once** test. For (1), the nothing-DRS guarantee keeps positive+negative assertions on the real `Configuration` mock (enables `llm-error-pages`, never `llmo-customer-analysis`; triggers neither it nor a DRS job); a full real-DB IT that stubs only the HTTP boundaries is a harness enhancement I've noted as follow-up.

## Questions
- **`required-capabilities.js` / org-scoped S2S token:** with the admin gate gone, the route is reachable by customer IMS tokens; `INTERNAL_ROUTES` only governs S2S-consumer exposure, so it doesn't block that. If a service needs S2S access later we can map it to a capability in `routeRequiredCapabilities`.
- **`llmo-onboarding.js` "do we still need this?":** that's the existing **v1 prompt-generation** path, shared with full onboarding and untouched by Piece 1 — it'll be removed with the v1 retirement, not in this PR.

rainer-friederich (approved):
Hey @igor-grubic,

**Verdict: Approve.** Commit c1ffefba resolves the blocking findings from the prior review, and the remaining items below are knowingly accepted (documented here so the trade-offs are on the record, not lost).

## Addressed
- [Critical] Admin gate dropped — now `hasAccess(org)` + explicit PAID, matching the v2 brand-management routes; OpenAPI/description corrected to stop claiming an admin requirement. The endpoint is now reachable by the paying customers it targets.
- [Important] SSRF — non-public hosts (IP literals incl. the metadata address, `localhost`, single-label/internal hosts) are rejected via `tldts` before any outbound probe, with negative tests.
- [Important] OpenAPI request/201 schemas extracted to `schemas.yaml` and referenced via `$ref`.
- [Important] Test gaps — siteOnly mid-flight rollback (`deleteSharePointFolder` + `revokeEnrollment`) and `enqueueLlmoOnboardingPublish` exactly-once/swallow tests added.

## Knowingly accepted (not done — and that's OK for this PR)
- **Org-wide `brandalf` flag flip (finding 4):** left as-is. Accepted on the stated rationale — benign for the realistic case, and the only adverse edge (an org with existing v1 brands flipped v1→v2 by adding one site) is bounded by the planned v1 retirement. If v1 retirement slips, this is the thing to revisit.
- **Domain-ownership verification (finding 3):** documented as an out-of-scope, GSC-based follow-up. Accepted. We are also deferring the lighter in-repo mitigation (per-org rate limit) — noting explicitly that, until ownership or a rate limit lands, a qualifying caller can onboard domains it does not control and first-claim-race a real owner.
- **SSRF resolve-time hardening:** the guard is hostname-shape only; DNS-resolution-time private-range/rebind checks are deferred. Accepted as a follow-up.
- **No-side-effect guarantee in IT:** still asserted only via unit-level mock spies (the 201 path needs SharePoint/GitHub/Ahrefs). Accepted; a real-DB Configuration assertion remains a nice-to-have.
- **Minor nits:** OpenAPI `maxLength`, `MAX_DOMAIN_LENGTH`/`MAX_BRAND_NAME_LENGTH` constants, path-bearing-domain rejection, a specific "already onboarded" message, and the `region` OpenAPI note are not taken. Non-blocking, fine to leave or fold into a later pass.

## Two open questions (non-blocking, answer at your leisure)
Still unanswered on the threads — recording that we are not blocking on them: whether this self-onboarding route should support an org-scoped IMS S2S token (i.e. revisit the `INTERNAL_ROUTES` classification), and the "do we still need to support this?" note on the extracted brand/prompt-gen helper. A one-line reply to close them out would be appreciated, but they don't gate the merge.

Thanks for the thorough turnaround.

src/routes/required-capabilities.js
igor-grubic: Closing out (non-blocking): for Piece 1 the only consumer is the customer-facing UI on a customer IMS (JWT) token, which works regardless of `INTERNAL_ROUTES` — that classification only governs S2S-consumer exposure. No backend service needs to call this today, so it stays in `INTERNAL_ROUTES`; if an org-scoped S2S consumer ever needs it, we map it to a capability in `routeRequiredCapabilities` then.

src/controllers/llmo/llmo-onboarding.js
igor-grubic: Closing out (non-blocking): yes for now — this is the existing brand-activation + v1 prompt-generation path, kept for full (admin/v1) onboarding and extracted here as the seam Piece 2 (LLMO-5605) plugs into. Site-only never calls it. The v1 sub-branch is removed with the planned v1 retirement (~next month).

igor-grubic:
Both closed out in-thread now — the `INTERNAL_ROUTES`/org-scoped-S2S classification and the "do we still need this?" note on the extracted brand/prompt-gen helper. Thanks again for the careful review — genuinely useful.

PR #2638, opened 2026-06-18
feat(redirects): serve ASO redirect overlay via api-service (Lite-E)

alinarublea:
Implements **Lite-E** from the ASO dispatcher-overlay ADR ([mysticat-architecture#159](https://github.com/adobe/mysticat-architecture/pull/159)): move the redirect-overlay read path behind api-service instead of having Fastly sign SigV4 to S3 directly — **with the per-request authorization that makes it safe** (closes the cross-tenant enumeration hole rather than relocating it).

## Endpoint

`GET /config/:service/redirects.txt` (e.g. `/config/cm-p154709-e1629980/redirects.txt`)

1. **AuthN** — inbound `X-ASO-API-Key` validated by `AsoOverlayKeyHandler` in the `AUTH_HANDLERS` chain (constant-time, length-safe HMAC compare; path-scoped early-bail like the GitHub webhook handler).
2. **AuthZ (per request)** — parse `cm-p<program>-e<env>` → resolve the site via the indexed `Site.findByExternalOwnerIdAndExternalSiteId('p<program>','e<env>')`; require the site's org to hold an **ASO entitlement** AND the site to be **enrolled** in it.
3. Reads `config/<service>/redirects.txt` from **`S3_ASO_OVERLAYS_BUCKET`** using the Lambda's **own execution role**.
4. Returns `text/plain` (200). **Every authz miss → an indistinguishable `404`** (no signal of which programs exist/are entitled). `400` for a malformed service id; `500` only for unexpected failures.

## Why this is the *safe* Lite-E

The shared `X-ASO-API-Key` alone would let any key-holder enumerate every program's overlay by walking `cm-pX-eY`. The per-request entitlement check resolves only **provisioned, ASO-entitled** tenants — enumeration is gone, and authz is server-side in code we own and test. (Isolation ceiling is still Option A per the ADR; Full-E = swap the static key for a consumer TA, separate follow-up.)

## Notable details

- **`:env` dropped** from the route — the bucket is already env-scoped via the Lambda's own deployment, so the segment was dead. Edge/dispatcher URL change to be coordinated separately with the dispatcher side.
[...]

PR #2663, opened 2026-06-22
feat(aso-overlay): dual-key overlap for API key rotation (ADR I13)

miakobchuk:
## Summary

Adds dual-key overlap support to the ASO overlay auth handler so API key rotation doesn't 401 dispatcher pods during the EVO propagation window.

Ports the dual-key pattern from the Fastly VCL side (spacecat-infrastructure #611) to the api-service Lite-E path where the key check now lives (#2638).

## Changes

**`src/support/aso-overlay-key-handler.js`** — check `ASO_OVERLAY_API_KEY` (current) first; if no match, check `ASO_OVERLAY_API_KEY_PREVIOUS` (rotation overlap). Both use the existing constant-time `safeEqual` compare. When `ASO_OVERLAY_API_KEY_PREVIOUS` is empty or unset (steady state), behaviour is identical to today.

## Rotation flow

```
T+0  Copy current ASO_OVERLAY_API_KEY → ASO_OVERLAY_API_KEY_PREVIOUS in Vault
     Write new value into ASO_OVERLAY_API_KEY
     Redeploy api-service → overlap window starts (both keys accepted)

T+N  Wait for all dispatcher pods to roll with new key (EVO propagation)

T+M  Clear ASO_OVERLAY_API_KEY_PREVIOUS in Vault
     Redeploy api-service → only new key accepted
```

## Vault change needed

Add `ASO_OVERLAY_API_KEY_PREVIOUS` (initially empty) to `dx_mysticat/<env>/api-service` for each env.

## Test plan

- [x] 5 new tests covering: previous key accepted during overlap, current key still works, wrong key rejected, empty previous key (steady state), unset previous key (steady state)
- [x] All existing key handler tests unchanged

### DIFF
--- a/src/support/aso-overlay-key-handler.js
+++ b/src/support/aso-overlay-key-handler.js
@@ -35,7 +35,10 @@ function safeEqual(a, b) {
 /**
  * Authenticates the ASO dispatcher-overlay read path
  * (`GET /config/:service/redirects.txt`) via the inbound `X-ASO-API-Key`,
- * validated against `ASO_OVERLAY_API_KEY`.
+ * validated against `ASO_OVERLAY_API_KEY` (current) and optionally
+ * `ASO_OVERLAY_API_KEY_PREVIOUS` (rotation overlap). During a key rotation both
+ * values are populated; in steady state the previous key is empty and only the
+ * current key is checked. See ADR I13 for the rotation playbook.
  *
  * This must live in the auth-handler chain (not the controller): `authWrapper`
  * runs `authenticate()` before any controller, so a route that is neither in
@@ -76,7 +79,9 @@ class AsoOverlayKeyHandler extends AbstractHandler {
       return null;
     }
 
-    if (!safeEqual(providedKey, expectedKey)) {
+    const previousKey = context.env?.ASO_OVERLAY_API_KEY_PREVIOUS;
+    if (!safeEqual(providedKey, expectedKey)
+        && !(previousKey && safeEqual(providedKey, previousKey))) {
       this.log('invalid X-ASO-API-Key', 'warn');
       return null;
     }
--- a/test/support/aso-overlay-key-handler.test.js
+++ b/test/support/aso-overlay-key-handler.test.js
@@ -125,4 +125,52 @@ describe('AsoOverlayKeyHandler', () => {
     expect(authInfo).to.not.be.null;
     expect(authInfo.getType()).to.equal('aso_overlay_key');
   });
+
+  describe('dual-key rotation overlap', () => {
+    const PREVIOUS_KEY = 'old-aso-key-being-rotated';
+
+    it('accepts the previous key during rotation overlap', async () => {
+      const authInfo = await handler.checkAuth(
+        makeRequest({ 'x-aso-api-key': PREVIOUS_KEY }),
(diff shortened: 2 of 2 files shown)


alinarublea (commented):
Reviewed — the dual-key logic itself is correct and well-tested, but there's an architectural gap that undermines the stated goal (avoiding rotation 401s).

## The code ✓
- `accept current OR (previous-set AND matches previous); else reject` is right.
- `previousKey && …` short-circuits when empty/unset → steady-state behaviour is identical to today.
- Reuses the constant-time `safeEqual`; `providedKey` is already guaranteed non-empty upstream, so no empty-match risk.
- 5 tests cover overlap / current / wrong / empty / unset — good coverage.

## ⚠️ Gap: the Fastly edge gate runs first and is single-key
After the Lite-E re-point, `/config/*` auth happens at **two** layers, and the **edge runs before api-service**:

1. **Fastly `recv 40`** validates `X-ASO-API-Key` against the `aso_keys` edge dictionary → `error 401` on mismatch, *before the request reaches api-service*. (This edge gate is required — it protects the cached `/config` responses from being served to unauthenticated requests.)
2. **api-service handler** (this PR) only sees requests that already passed the edge.

The edge currently has **only `aso_keys` (single key)** — `aso_keys_previous` was never actually deployed to the service (it throws "Undefined table" if referenced). So during rotation a dispatcher pod presenting the **old** key gets `401` **at the edge** and never reaches this handler's dual-key logic.

So this PR alone does **not** prevent the rotation 401s. The description says it ports #611's edge dual-key, but that edge dual-key isn't live (single-key only). For rotation to work end-to-end the Fastly edge needs matching dual-key: create the `aso_keys_previous` edge dictionary + add the previous-key check to `recv 40` (next to the existing gate).

## Minor
- Add `ASO_OVERLAY_API_KEY_PREVIOUS=` (empty) to `.env.example` for discoverability / local parity.
- Worth noting the Fastly-edge dependency in the PR body so nobody assumes rotation is functional after merging just this half.

## Recommendation
Fine to merge — it's the correct api-service half and a no-op in steady state. **But** it must be paired with the Fastly edge dual-key (`aso_keys_previous` dict + `recv 40` check) before any actual key rotation; otherwise the edge 401s the old key first. I'd hold off calling rotation "supported" until both halves are in.

alinarublea (approved):
Approving — and the cross-layer dependency I raised earlier is now resolved.

**The Fastly edge dual-key half is live.** `aso-prod` is on **v106** (active), which adds the `aso_keys_previous` edge dictionary (empty) + the dual-key check in `recv 40`, mirroring this handler. So during a rotation the old key is accepted at **both** layers (edge gate → api-service), not just here. Verified on v106: additive-diff vs v105 (only `recv 40` changed), dict present + empty, and steady-state intact (no-key/wrong-key still `401`, valid key `200`). Edge change is in spacecat-infrastructure#627.

The code itself is correct and well-tested (current-or-previous with empty-guards, constant-time compare, 5 good tests). 👍

Non-blocking nit (still open): add `ASO_OVERLAY_API_KEY_PREVIOUS=` (empty) to `.env.example` for discoverability/local parity.

Rotation reminder for whoever runs it: don't start a rotation until this is merged + deployed, then populate **both** `aso_keys_previous` (edge dict) **and** `ASO_OVERLAY_API_KEY_PREVIOUS` (Vault) with the old value in lockstep — populating only one layer 401s the old key on the other. Flow is documented in #627's `MIGRATION-lite-e.md`.

PR #2666, opened 2026-06-23
feat: introduce semrush elements apis wrapper for llmo ui

vivesing:
# LLMO Semrush Elements API — Summary

One SpaceCat wrapper endpoint over Semrush Elements APIs, powering the URL Inspector filter dimensions. (The former brand-, market-, and tag-selector endpoints were removed — that data is now served by the existing Serenity APIs.)

## Endpoint

| # | Endpoint | Purpose | Elements |
|---|----------|---------|----------|
| 1 | `GET .../serenity/all/brand-presence/url-inspector/filter-dimensions` | All 6 dimensions in one call (3 parallel element fetches) | `BRANDS`, `MARKETS`, `TOPICS` |

*(Prefixed with `/v2/orgs/:spaceCatId`)*

## Parameters

- **`spaceCatId`** (path) — required.
- **`model`** (query, optional) — must be one of 11 valid models, else falls back to `search-gpt` (default).

## Response

Returns a single object with six dimension arrays: `brands`, `regions`, `topics`, `categories`, `page_intents`, `origins`.

- **`brands`** — `id` always `null`; enriched with `spacecat_brand_id` via case-insensitive name match against SpaceCat brands (`null` if unmatched).
- **`regions`** — `id` extracted from label (`"AU-en"` → `"AU"`); enriched from `brand_to_semrush_projects` with `semrush_project_id`, `spacecat_brand_id`, `geoTargetId`, `languageCode`.
- **`topics`** — `topic:`-prefixed tags → `{ id: null, label }`.
- **`categories`** — `category:`-prefixed tags → `{ id: null, label }`.
- **`page_intents`** — `intent:`-prefixed tags; `id` uppercased (`"commercial"` → `"COMMERCIAL"`), `label` kept raw.
- **`origins`** — `source:`-prefixed tags → `{ id: value, label: value }`.

## Source of Truth

Valid models live in `src/support/elements/constants.js` (`ELEMENT_MODELS` / `DEFAULT_ELEMENT_MODEL`).

## Wiki

https://wiki.corp.adobe.com/spaces/AEMSites/pages/3928196548/Project+Serenity+LLMO+x+Semrush+API+for+Brand+Presence+Data


calvarezg:
## Multi-Persona Review — Architect · DBA · Security · Tester/Junior Dev

Purely additive (24 files, 3860+/0-). No prior human reviews. Verified against `brands-storage.js` and `route-utils.js` from `main` (not in the diff).

### 🟠 Major (1)

**Region enrichment is last-write-wins if a Semrush project maps to >1 brand** — `src/support/elements/definitions/markets.js:80-98`
`transformMarketsToFilterDimensions` keys a Map by `semrushProjectId`; duplicate keys keep only the last row, silently attributing a region to one brand and hiding the others. `listAllMarkets`/URL-inspector aggregate across all org brands, so a shared project would surface here. **Q: is `brand_to_semrush_projects.semrush_project_id` unique per brand?** If 1:1, waive this. If a project can belong to multiple brands, emit one region per match (flatMap) or collect `spacecat_brand_id[]`. No log today either way.

### 🔵 Minor (7)

1. **Unbounded fan-out** — `elements.js:219-220,318-319,268-275`. One DB query per brand and one upstream POST per project via uncapped `Promise.all`. Batch the query / cap with `p-limit`.
2. **Null `semrushProjectId` collides as a Map key** — `markets.js:81-86`. Guard against null keys and falsy `item.value`.
3. **Dead transformers** — `transformBrandsResponse`/`transformMarketsResponse` are exported + tested but unused in production (`definitions/index.js:13-14`). Remove or mark as Phase-2.
4. **Wrong env var in comments** — `elements-transport.js:20,117` say `SEMRUSH_ELEMENTS_BASE_URL`; code reads `SEMRUSH_PROJECTS_BASE_URL`.
5. **Stale JSDoc** — `elements.js:166` says `:brandName`; code/route use `:brandId`.
6. **Route precedence untested** — `/serenity/all/markets` vs `/serenity/:brandId/markets` are both dynamic + equal length; resolution relies on insertion order (correct today). Add a dispatch test so a reorder can't misroute `all`.
7. **No explicit 504→502 test** — covered by equivalence via the default branch; optional to assert directly.

### ✅ Working well

Security: no findings. SSRF contained (https-only, origin-only base, `encodeURIComponent` on path segments, params only in body). IMS token never logged/leaked; upstream error bodies never reach clients. `authorizeOrg` runs first on all 6 handlers; `getBrandById` is org-scoped (cross-tenant read not possible). Clean transport/service/definitions layering; thorough branch-level tests.

**Recommendation:** Changes recommended — answer the cardinality question (waives or promotes the Major); the rest is polish.

vivesing:
Major(1):

Good catch to flag — checked the data model and this is safe to waive. semrush_project_id is globally unique across brand_to_semrush_projects, not just per-brand:

CONSTRAINT uq_brand_to_semrush_project UNIQUE (semrush_project_id)
Per the column comment: "UNIQUE across the table — one Semrush project binds to exactly one (brand, location, language) slice." So a project can never map to >1 brand — the DB rejects the second INSERT. The last-write-wins Map can't collide.

Added a comment at the Map construction documenting the dependency on that constraint, so if anyone ever relaxes it the assumption is right there in the code. flatMap / spacecat_brand_id[] would just be dead code guarding an impossible state.

vivesing:
| # | Comment | Fix |
|---|---------|-----|
| 1 | Unbounded fan-out | Added `mapWithConcurrency` + `fetchBrandSemrushProjects` helpers (cap 8) in `elements.js`; applied to `listAllMarkets`, `listUrlInspectorFilterDimensions`, and `listBrandTags`. Also deduped the repeated brand→project mapping into `toPlainProject`. |
| 2 | Null Map key collision | `markets.js` now filters out rows without `semrushProjectId` and only looks up when `item.value` is truthy; added a regression test. |
| 3 | Dead transformers | Removed `transformBrandsResponse` / `transformMarketsResponse` (functions, exports, and their test blocks) — they were only referenced by their own tests. |
| 4 | Wrong env var in comments | Fixed `SEMRUSH_ELEMENTS_BASE_URL` → `SEMRUSH_PROJECTS_BASE_URL` in transport JSDoc. |
| 5 | Stale JSDoc `:brandName` | Updated `listMarkets` JSDoc to `:brandId`. |
| 6 | Route precedence untested | Added a dispatch test in `index.test.js` asserting `/serenity/all/markets` → `listAllMarkets` and `/serenity/:brandId/markets` → `listMarkets`, guarding against a reorder. |
| 7 | No explicit 504→502 test | Added an explicit `504 → 502` assertion in the controller test. |

@calvarezg addressed review comments, please review

rainer-friederich:
So we implement a lot of endpoints for data via elements api from Semrush which we already have via project-engine and user-manager api for the prompts management. We should discuss this as this will be strange in the future and hard to understand.

**1.List brands**
Why do we need that? We can just use our brands?

**2. List all markets**
Wouldn't `/v2/orgs/:spaceCatId/brands/all/serenity/market` be more consistent regarding routing with the already existing endpoints?
I think you need to iterate first on all sub-workspaces and then over all projects in t hat sub-workspace. I also think we should not do this at all if possible - For Adobe this will be hundreds of requests.
We have this for brand level, perhaps the code can be reused.
https://github.com/adobe/spacecat-api-service/blob/main/src/routes/index.js#L227

**3. List Markets for a Brand**
This exists already https://github.com/adobe/spacecat-api-service/blob/main/src/routes/index.js#L227

**4. List Tags (workspace-wide)**
Same as for 2 applies. We have this for brand level, perhaps the code can be reused.
https://github.com/adobe/spacecat-api-service/blob/main/src/routes/index.js#L231

**5. List Tags for a Brand**
See 4

**6. List URL Inspector Filter Dimensions**
There was some exploration work regarding url inspector from Claudia. Unsure if it is worth to reuse: https://github.com/adobe/spacecat-api-service/pull/2588

**7. Supported Models**
This exists already: https://github.com/adobe/spacecat-api-service/blob/main/src/routes/index.js#L237 But the source is semrush, not a static list.
This also exists for languages.

There is also typed clients to work with these apis:
https://github.com/adobe/spacecat-shared/tree/main/packages/spacecat-shared-user-manager-client
https://github.com/adobe/spacecat-shared/tree/main/packages/spacecat-shared-project-engine-client

calvarezg:
## Re-review (delta) — all prior findings cleared

Re-ran the multi-persona pass on the updated diff (now 2760+/0-, scope narrowed to the single `listUrlInspectorFilterDimensions` endpoint). Verified each item against the new code:

| # | Prior finding | Status |
|---|---------------|--------|
| Major | Region last-write-wins on shared project | ✅ **Waived (valid)** — `uq_brand_to_semrush_project UNIQUE (semrush_project_id)` makes project↔brand strictly 1:1; assumption now documented at the Map construction (`markets.js:100-108`) |
| 1 | Unbounded fan-out | ✅ `mapWithConcurrency` + `fetchBrandSemrushProjects` (cap 8); worker-pool is correct (order-preserving, safe on empty input) |
| 2 | Null `semrushProjectId` Map-key collision | ✅ filters falsy `semrushProjectId`, guards the lookup on truthy `item.value` + regression test |
| 3 | Dead transformers | ✅ `transformBrandsResponse`/`transformMarketsResponse` fully removed (0 refs) |
| 4 | Wrong env var in comments | ✅ `SEMRUSH_PROJECTS_BASE_URL` throughout, incl. JSDoc |
| 5 | Stale `:brandName` JSDoc | ✅ moot — `listMarkets` removed |
| 6 | Route precedence untested | ✅ moot — the conflicting `all/markets` vs `:brandId/markets` routes were removed; only the 8-segment filter-dimensions route remains, no ambiguity |
| 7 | No 504→502 test | ✅ added (`elements.test.js`) |

**No new correctness issues** in the delta.

### Two notes (non-blocking)

- **DB fan-out is now bounded in *concurrency* but not in *count*.** `fetchBrandSemrushProjects` still issues one `allByBrandId` per brand (≤8 in flight). For a very large org that's still O(brands) round-trips — a batched `allByBrandIds([...])`/org-level query would scale better if the data-access layer grows one. This overlaps with @rainer-friederich's scale concern.
- **Architecture direction (@rainer-friederich's review) is a team call, not a code-correctness item.** Removing the 5 duplicate endpoints already addresses most of it; whether the remaining URL-inspector aggregate should build on the existing project-engine/user-manager typed clients (and reuse the PR #2588 exploration) is worth resolving before merge, but it's outside this automated pass.

**Recommendation (code-correctness lens):** Approve — all flagged findings resolved/waived. Merge gate is the architecture discussion, not the code.

calvarezg:
## Multi-Persona Review — Final Summary & Verdict

Two review rounds (Architect · DBA · Security · Tester/Junior Dev). This is the consolidated outcome after the update that narrowed scope to the single `listUrlInspectorFilterDimensions` endpoint.

### Findings & resolution

| Sev | Finding | Resolution |
|-----|---------|------------|
| 🟠 Major | Region enrichment last-write-wins if a Semrush project maps to >1 brand (`markets.js`) | ✅ **Waived (valid)** — `uq_brand_to_semrush_project UNIQUE (semrush_project_id)` enforces project↔brand 1:1; assumption now documented at the Map construction |
| 🔵 Minor | Unbounded parallel fan-out (DB per-brand, upstream per-project) | ✅ `mapWithConcurrency` + `fetchBrandSemrushProjects`, concurrency cap 8; worker-pool verified correct |
| 🔵 Minor | Null `semrushProjectId` collides as Map key | ✅ filters falsy keys, guards lookup on truthy `item.value`, + regression test |
| 🔵 Minor | Dead transformers (`transformBrandsResponse`/`transformMarketsResponse`) | ✅ removed entirely (0 refs) |
| 🔵 Minor | Wrong env var in comments (`SEMRUSH_ELEMENTS_BASE_URL`) | ✅ corrected to `SEMRUSH_PROJECTS_BASE_URL`, incl. JSDoc |
| 🔵 Minor | Stale `:brandName` JSDoc | ✅ moot — `listMarkets` removed |
| 🔵 Minor | Route precedence (`all/markets` vs `:brandId/markets`) untested | ✅ moot — conflicting routes removed; only the unambiguous 8-segment route remains |
| 🔵 Minor | No explicit 504→502 test | ✅ added |

### Security

No findings. SSRF contained (https-only, origin-only base, `encodeURIComponent` on path segments, params only in the POST body). IMS token never logged or leaked; upstream error bodies never reach clients. `authorizeOrg` runs first on the handler; `getBrandById` is org-scoped (no cross-tenant read).

### Open (non-blocking, team decision)

- **DB fan-out** is now concurrency-bounded (≤8) but still O(brands) total round-trips; a batched org-level query would scale better for very large orgs.
- **Architecture direction** (@rainer-friederich): removing the 5 duplicate endpoints addressed most of it; whether the remaining URL-inspector aggregate should build on the existing project-engine/user-manager typed clients (and reuse PR #2588) is worth settling before merge.

---

## Verdict: ✅ Approve (code-correctness)

0 critical, 0 major outstanding — every flagged finding is resolved, validly waived, or made moot by the scope reduction. No new correctness issues in the delta. The only remaining merge gate is the architecture/duplication discussion, which is a team call rather than a code defect.

calvarezg (approved):
Approving from a code-correctness standpoint. All multi-persona review findings (1 Major + 7 Minor) are resolved, validly waived, or made moot by the scope reduction; security is clean. Remaining architecture/duplication discussion is a non-blocking team decision.

PR #2667, opened 2026-06-23
fix: match prompts region filter case-insensitively | LLMO-5755

igor-grubic:
## Problem

The brand-scoped prompts list endpoint (`GET /v2/orgs/{orgId}/brands/{brandId}/prompts?region=…`) filtered by region using a **case-sensitive** Postgres array containment:

```js
baseQuery = baseQuery.contains('regions', [region]); // regions @> ARRAY['us']
```

Region codes are persisted with **inconsistent casing** depending on the onboarding path:
- `onboard-llmo` (Slack) and Serenity brand-provisioning store them **UPPERCASE** (`['US','NO',…]`).
- CSV upload and the customer-config-mapper store them **lowercase** (`['us',…]`).

The UI always sends the filter value lowercased (`code.toLowerCase()`). So for uppercase-onboarded brands — e.g. **Creditsafe**, which was onboarded via `onboard-llmo` — `ARRAY['US'] @> ARRAY['us']` is `false`, every prompt is filtered out, and the market filter appears "completely bugged out." Lowercase-onboarded brands were unaffected, which is why this looked Creditsafe-specific.

Reported in the LLMO support thread → ticket **LLMO-5755**.

## Fix

Match an array **overlap** against both case variants instead of a single case-sensitive containment:

```js
const regionVariants = [...new Set([region.toLowerCase(), region.toUpperCase()])];
baseQuery = baseQuery.overlaps('regions', regionVariants); // regions && ARRAY['us','US']
```

The filter is now case-insensitive and works regardless of how regions were stored. **No stored data is rewritten** — this is a read-path-only change, so no migration/backfill is required. The frontend needs no change (it already sends the lowercased code).

## Tests / docs

- **Unit** (`test/support/prompts-storage.test.js`): asserts the region filter calls `.overlaps('regions', ['us','US'])`. Suite: 148 passing.
- **Integration** (`test/it/shared/tests/categories-prompts.js`): stores prompts with UPPERCASE regions, filters with a lowercase `?region=us`, and expects both case-variant rows back (and that filtering still narrows the set).
- **OpenAPI** (`docs/openapi/prompts-v2-api.yaml`): documented the case-insensitive matching; `npm run docs:lint` passes.

## Verification still to do

End-to-end Playwright verification of the Prompts Management market filter against a deployed env with the Creditsafe brand (pending deploy of this branch).

### DIFF
--- a/docs/openapi/prompts-v2-api.yaml
+++ b/docs/openapi/prompts-v2-api.yaml
@@ -66,7 +66,11 @@ v2-prompts-by-brand:
       required: false
       schema:
         type: string
-      description: Filter by region (matches prompts whose regions array contains this value)
+      description: >-
+        Filter by region code (e.g. `us`). Matches prompts whose `regions`
+        array contains the value, case-insensitively — region codes are stored
+        with inconsistent casing depending on the onboarding path, so both
+        lower- and upper-case variants are matched.
     - name: origin
       in: query
       required: false
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -509,7 +509,16 @@ export async function listPrompts({
     }
 
     if (hasText(region)) {
-      baseQuery = baseQuery.contains('regions', [region]);
+      // Region codes are persisted with inconsistent casing depending on the
+      // onboarding path: `onboard-llmo` and Serenity provisioning store them
+      // UPPERCASE (e.g. ['US','NO']), while CSV / config-mapper imports store
+      // them lowercase. The UI sends the code lowercased, so a case-sensitive
+      // `.contains` (regions @> ARRAY['us']) silently matches nothing for the
+      // uppercase brands. Match an array overlap against both case variants so
+      // the filter is case-insensitive without rewriting stored values.
+      // (LLMO-5755)
+      const regionVariants = [...new Set([region.toLowerCase(), region.toUpperCase()])];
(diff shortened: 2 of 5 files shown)


MysticatBot (approved):
Hey @igor-grubic,

**Verdict: Approve** - clean, well-scoped bug fix with solid test coverage at both the unit and integration level.
_Changes: Fixes case-sensitive Postgres array containment filter for region codes by switching from `.contains()` to `.overlaps()` with both lower- and upper-case variants (4 files)._

<details>

- nit: `[...new Set([region.toLowerCase(), region.toUpperCase()])]` - the Set dedup is a no-op for alphabetic region codes (lower and upper are always distinct); `.overlaps()` handles duplicate array elements fine, so the Set is strictly optional. Harmless and defensive, but a reader might wonder what edge case it guards against. - `src/support/prompts-storage.js:512`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 13s | <b>Cost:</b> $3.76 | <b>Commit:</b> <code>bbcba0e72d2a4a0a73557c2380fb9f6654104874</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jjenscodee (approved):

Issue #2668, opened 2026-06-23
Opt-in TypeScript type-checking for /serenity routes (checkJs + JSDoc)


rainer-friederich:
## Summary

Stand up opt-in TypeScript type-checking (via `checkJs` + JSDoc, no `.ts` source) for the `src/support/serenity/**` routes, and enforce it in CI. This is a follow-up to adopting the typed Semrush Project Engine client — see https://github.com/adobe/spacecat-api-service/pull/2643 and the publishing/type-fix in https://github.com/adobe/spacecat-shared/pull/1706 .

## Why now

`src/support/serenity` now talks to Semrush through the typed `@adobe/spacecat-shared-project-engine-client` (`openapi-fetch` over a generated `paths` contract). That contract is only useful if something actually checks it. Today api-service has **no type-checking at all** — no `tsconfig`, no `typescript` dep, no `tsc`/CI step (it is a pure ESM + JSDoc repo, the Helix/Franklin convention shared across the spacecat fleet; there is no ADR recording that choice).

A scoped `tsc --checkJs` probe over `/serenity` already paid for itself:

- It surfaced a real **Semrush swagger defect** — the workspace `id` was declared `in: query` on 11 operations where `{id}` is a path-template variable. That is now fixed upstream via overlay correction CR4 (https://github.com/adobe/spacecat-shared/pull/1706), so the published types are correct.
- It surfaced a genuine **type-modeling gap in our own code**: `ErrorWithStatusCode` only declares `.status`, but handlers assign `err.code = …` dynamically (6 call sites). The class should model `code`.

Principle to carry forward (learned here): a type error can mean the **swagger/spec is wrong**, not our code. Discrepancies must be checked from two sides — static trace (generated types ↔ vendored swagger ↔ overlay) and, when ambiguous, the live API with an IMS token.

## Measured effort (scoped to `src/support/serenity/**`)

| Mode | Errors in serenity files |
|------|--------------------------|
| `strict: true` | ~502 (≈372 are just implicit-`any` params — missing `@param` types) |
| `noImplicitAny: false` (pragmatic) | **15** — and most are legitimate fixes |

Recommended start: **pragmatic** (`noImplicitAny:false`), then ratchet stricter. The 15:

- 6× `ErrorWithStatusCode.code` — declare `code` on the class (one fix).
- `getProject` passes `draft` as boolean where the query type is string → `String(draft)`.
- `competitor-benchmarks` — optional param before a required one (JSDoc signature).
[...]

PR #2669, opened 2026-06-23
feat: opt-in TypeScript type-checking (checkJs + JSDoc) for /serenity routes

rainer-friederich:
## 1. Abstract
Stands up opt-in TypeScript type-checking (`checkJs` + JSDoc, no `.ts` source, no emit) over `src/support/serenity/**`, and enforces it in a new blocking CI job and the pre-commit hook.

## 2. Reasoning
`src/support/serenity` now talks to Semrush through the typed `@adobe/spacecat-shared-project-engine-client` (`openapi-fetch` over a generated `paths` contract). That contract is only worth shipping if something actually checks it — and api-service had no type-checking at all (no `tsconfig`, no `typescript` dep, no `tsc`/CI step). A scoped `tsc --checkJs` probe over `/serenity` also surfaced genuine type-modeling gaps in our own code, so the change both enforces the Semrush contract and catches latent shape/null bugs.

Introduces an opt-in seam so files are type-checked one at a time, with no runtime impact (`noEmit`, JSDoc-only).

- Adds `tsconfig.json`: `checkJs: false` plus a per-file `// @ts-check` pragma is the opt-in seam (only annotated files are error-reported; the rest of the import graph is parsed for types only). Pragmatic floor — `noImplicitAny: false` and `useUnknownInCatchVariables: false` — to be ratcheted later. `types: ["node"]` so `node:*` builtins resolve. `include` is scoped to `src/support/serenity/**`.
- Adds `typescript` + `@types/node` devDeps and a `type-check` npm script.
- Adds `// @ts-check` to all 15 serenity files and fixes the errors that surfaced. The behaviour-relevant ones: `ErrorWithStatusCode` now models the `code` property that six serenity sites already assign dynamically; completing JSDoc `@param` tags that had left destructured options inferred as `never[]`; local null-narrowing after `hasText` guards; spreading the `readonly` tag constants; and a few JSDoc signature corrections. None of these change runtime behaviour.
- Adds a least-privilege, blocking `type-check` job to `.github/workflows/ci.yaml` — the reusable `adobe/mysticat-ci` service-ci workflow has no type-check seam yet, so it runs as a self-contained local job — and adds `npm run type-check` to the husky pre-commit hook.
- Records the decision (and the ratchet path) in ADR-005.

## 4. Required information
- Jira / issue: https://github.com/adobe/spacecat-api-service/issues/2668
- ADR(s): https://github.com/adobe/spacecat-api-service/blob/feat/typescript/docs/decisions/005-opt-in-type-checking.md
- Other: follow-up to the typed project-engine-client adoption — https://github.com/adobe/spacecat-api-service/pull/2643 and the published type-fix https://github.com/adobe/spacecat-shared/pull/1706

Verified the new CI gate is not a no-op: injected a deliberate `@ts-check` type error into a serenity file, confirmed `npm run type-check` reported it (`TS2322`), then reverted. Also confirmed the modeling fixes preserve behaviour by running the affected serenity + utils unit suites locally with no regressions.

## 7. Test plan
- Local: ran the `test/support/serenity/**` and `test/support/utils.test.js` suites to confirm the type-modeling fixes (notably `ErrorWithStatusCode.code`) introduce no behavioural change; ran `npm run type-check` and exercised the pre-commit hook.
- Per-environment: this is a dev-tooling / CI change with no runtime or deployment impact (`noEmit`, JSDoc-only — the Lambda bundle is unchanged). Verification is on PR CI, where the new blocking `type-check` job must pass; no dev/stage/prod runtime verification applies.


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-structured opt-in type-checking with no runtime impact; a few documentation and ergonomic nits below.
_Changes: introduces opt-in TypeScript type-checking (checkJs + JSDoc, noEmit) over the serenity routes, enforced by a new blocking CI job and pre-commit hook, with type-modeling fixes that preserve existing behavior (22 files)._

<details>

- nit: CLAUDE.md Commands section does not list the new `npm run type-check` script - agents and developers using the doc as a reference won't know about this blocking gate - `CLAUDE.md`
- nit: pre-commit hook runs full `tsc` on every commit regardless of staged files; consider gating with a path check (`git diff --cached --name-only | grep -q '^src/support/serenity/'`) as coverage expands - `.husky/pre-commit:4`
- suggestion: `@types/node@24.13.2` is exact-pinned; a caret range (`^24.13.2`) would absorb non-breaking type-definition patches without Dependabot noise - also verify the project actually targets Node 24 (if it runs Node 20/22, the stubs advertise APIs that don't exist at runtime) - `package.json:149`
- nit: `noImplicitAny: false` under `strict: true` is non-obvious to future readers; a brief inline comment in tsconfig.json noting it is an intentional relaxation (per the ADR ratchet) would help - `tsconfig.json:10`
- suggestion: add `timeout-minutes: 5` to the `type-check` CI job to prevent billing surprises if a future tsconfig triggers infinite type recursion - `.github/workflows/ci.yaml:27`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 29s | <b>Cost:</b> $5.10 | <b>Commit:</b> <code>11c60bacb89716d842d9f0afbc9cb140ac1d8408</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Thanks @MysticatBot — addressed the non-blocking items in `03d336bc`.

MysticatBot (all 5 non-blocking):
- nit: CLAUDE.md missing `npm run type-check` — done. Added to the Commands section.
- nit: tsconfig `noImplicitAny:false` under `strict` is non-obvious — done. Added inline JSONC comments noting the opt-in `checkJs:false` seam and the pragmatic relaxations (pointing at ADR-005's ratchet).
- suggestion: `timeout-minutes` on the type-check CI job — done. Added `timeout-minutes: 5`.
- nit: gate the pre-commit `tsc` on a staged `src/support/serenity/` path check — declined. The checked program spans imported files *outside* that path (e.g. `src/support/utils.js` `ErrorWithStatusCode`, consumed by serenity), so a serenity-path-only gate would skip commits that can still break serenity types. The run is also fast today (scope is serenity-only via `include`). If coverage expands and commit latency bites, the better lever is CI-only (it's already a blocking job).
- suggestion: caret-range `@types/node` + verify Node 24 target — Node 24 confirmed (`engines.node` `>=24.0.0 <25.0.0`, `.nvmrc` 24.15.0, `hlx.nodeVersion: 24`), so the v24 stubs match runtime. Kept the exact pin (declined caret): it matches this repo's exact-pin devDep convention and keeps the type-check gate reproducible — a `@types/node` patch can newly surface or mask diagnostics, which we don't want a caret to pull in silently.

PR #2670, opened 2026-06-23
feat(sites): scope latest-metrics RUM to locale path prefix

alinarublea:
> **Draft — blocked on [adobe/spacecat-shared#1707](https://github.com/adobe/spacecat-shared/pull/1707).** CI is red with `ETARGET` until that PR merges and publishes the new `@adobe/spacecat-shared-utils` + `@adobe/spacecat-shared-rum-api-client`. **The version pins here are predictions** — reconcile them against the actually-published versions before un-drafting (see "Version pins" below).

## Why

`GET /sites/{siteId}/latest-metrics` resolves the **bare hostname** for the RUM domain key (keys exist only for the main domain), then `totalMetrics` aggregates RUM for the **entire** parent domain. Path-based locale sites (e.g. `https://example.com/de`) therefore report domain-wide numbers instead of their own subtree.

## What

- Derive the locale path prefix from `site.getBaseURL()` via the new `getBaseURLPathPrefix` helper.
- Pass it as `pathPrefix` to both `totalMetrics` queries (current + previous window). `spacecat-shared-rum-api-client` filters bundles to that subtree before aggregating (boundary-safe: `/de` ≠ `/design`).
- `null` for whole-domain sites, subdomain locales, and file-pointing baseURLs → behavior unchanged.

`domain` stays the bare hostname (still required for the bundler URL and main-domain key resolution); scoping happens purely at the result level.

## Verified against live RUM data

`www.adobe.com`: 242,339,200 PV (30d) → scoped to `/jp` = **9,494,200 PV (3.92%)**. Subdomain locales (`business.adobe.com`) have their own domain key + bundles and already scope correctly — `pathPrefix` is `null` for them, no change. See the companion PR for full evidence and the design rationale (including why a file-pointing baseURL like `/us/en.html` falls back to whole-domain rather than over-filtering).

## Version pins

This PR bumps two shared deps to **predicted** next releases:

- `@adobe/spacecat-shared-utils` → `1.122.0` — note `1.121.0` is already published by an unrelated change and does **not** contain `getBaseURLPathPrefix`; the helper ships in the next release after #1707 merges.
- `@adobe/spacecat-shared-rum-api-client` → `2.44.0` (latest published is `2.43.1`).

**Before un-drafting:** once #1707 merges and semantic-release publishes, read the actual published versions, update these two pins to match, confirm CI goes green, then mark ready for review.

## Out of scope

Subdomain-based locales are handled by their own RUM domain key (no path filtering needed). Keyless-subdomain configs (subdomain reporting under the apex key only) were not observed in practice and are not addressed here.

## Tests

- New unit test asserts `pathPrefix` is derived and passed to **both** RUM queries for a locale site.
- Existing `getLatestSiteMetrics` tests unchanged and passing (non-locale sites → `pathPrefix = null`). Verified locally against a pre-release build of the helper (19 tests green); lint clean.


MysticatBot (approved):
Hey @alinarublea,

**Verdict: Approve** - clean, well-scoped locale-scoping fix with appropriate tests.
**Complexity: MEDIUM** - small diff; dependency bump.
_Changes: Derives a locale path prefix from site baseURL and passes it to both RUM total-metrics queries so locale sites report their own subtree instead of domain-wide numbers (4 files)._

Note: CI checks are currently pending - verify green before merge.

<details>

- nit: Test does not assert the argument passed to `getBaseURLPathPrefix` - `test/controllers/sites.test.js:1762`. Adding `expect(getBaseURLPathPrefix).to.have.been.calledWith(site.getBaseURL())` would catch wiring drift if a refactor accidentally passes the wrong value.
- suggestion: Consider documenting the locale-scoping behavior in `docs/openapi/site-metrics-api.yaml` endpoint description - the response shape is unchanged but the narrowing behavior is worth noting for API consumers.
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 37s | <b>Cost:</b> $3.87 | <b>Commit:</b> <code>79d0133648dcd75a4961ef58d9ae3752392df038</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2672, opened 2026-06-23
fix: reactivate soft-deleted prompts on bulk re-upload | LLMO-5669

jjenscodee:
### Bug fix — reactivate soft-deleted prompts on bulk re-upload
- **Root cause**: \`upsertPrompts()\` had a \`continue\` that silently skipped any prompt whose DB row had a non-active status, so re-uploading a previously deleted prompt returned \`created: 0, updated: 0\` with no DB change
- **Fix**: When a matched row has \`status='deleted'\`, flip it to \`active\` and count it as \`updated\`. All other non-active statuses (e.g. \`pending\`) remain skipped — the guard is scoped to \`deleted\` only
- **Intent preservation**: The existing-prompts SELECT now fetches \`intent\` (guarded with \`withMissingIntentFallback\` for column-absent environments); reactivated rows use the stored DB intent as fallback when the incoming payload omits it, preventing silent data loss
- **Spec**: Updated OpenAPI description for \`POST /v2/orgs/:orgId/brands/:brandId/prompts\` to document reactivation and skip semantics

### Performance — concurrent bulk update pool
- Replaced the serial \`for\` loop in \`upsertPrompts\` with a 20-worker shared-index pool (\`UPDATE_CONCURRENCY = 20\`), reducing 1000-row update time from ~15 s to ~750 ms and keeping responses within Fastly's 15 s first-byte timeout
- Error reporting now includes the full failure count and all error messages: \`Failed to update N prompt(s): msg1; msg2\`

### Performance — intent classification with 8 s best-effort cap
- Re-added \`classifyIntents\` to the \`upsertPrompts\` path with a configurable wall-clock cap (\`classifyIntentBatchTimeoutMs\`, default 8000 ms). \`classifyIntents\` resolves to a partial map on timeout (never rejects), so whatever classifies in 8 s gets an intent value; remaining prompts stay \`null\` and are handled by the backfill path. Combined with the concurrent update pool, total request time stays well under 15 s
- **Timeout correctness fix**: When the batch cap fires, the shared index is drained so in-flight workers exit after their current call instead of picking up the next text. Without this, workers continued making new LLM calls after the cap fired, rendering the timeout ineffective

## Code review findings fixed before PR
Two bugs caught during self-review and fixed:
1. Original condition \`status !== 'active'\` was too broad — would have force-reactivated \`pending\` prompts. Narrowed to \`status === 'deleted'\`
2. Reactivation update would write \`intent: null\` over a previously stored intent. Fixed by fetching \`intent\` in the existing SELECT and using it as fallback

## Code review findings fixed after PR opened
Four issues caught during review and fixed:
1. \`classifyIntentBatchTimeoutMs\` was accidentally dropped as a param and hardcoded to 8000 — restored as a param with 8000 as the default
2. Concurrent update pool threw only the first error — now reports all failures with count: \`Failed to update N prompt(s): …\`
3. Existing-prompts SELECT included \`intent\` without the \`withMissingIntentFallback\` guard — wrapped with the same pattern the rest of the file uses
4. Batch timeout was ineffective — workers continued starting new LLM calls after the cap fired; fixed by draining the shared index in the timer callback

## Test recording

https://github.com/user-attachments/assets/a4c9a1af-ac0e-4839-9559-11d6d5480954

## Related issues
https://jira.corp.adobe.com/browse/LLMO-5669


MysticatBot (approved):
Hey @jjenscodee,

**Verdict: Approve** - clean, narrowly-scoped fix with solid test coverage.
**Complexity: LOW** - small diff, single-service, single code path.
_Changes: Reactivates soft-deleted prompts on bulk re-upload instead of silently skipping them, preserving the stored DB intent as fallback (3 files)._

<details>

- suggestion: Add a brief inline comment explaining why the intent fallback (`row.intent ?? match.intent`) exists only on the reactivation path and not on the normal update path - the asymmetry is intentional but non-obvious to future readers - `src/support/prompts-storage.js:715`
- nit: `processed.push({ ...reactivated, prompt_id: promptId })` redundantly overrides `prompt_id` which is already present via the `...row` spread in `reactivated` - harmless but noisy - `src/support/prompts-storage.js:716`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 28s | <b>Cost:</b> $3.13 | <b>Commit:</b> <code>dc1ff150974d59f2219e700a469e304191a6edb0</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/prompts-storage.js
```diff
@@ -766,30 +777,25 @@ export async function upsertPrompts({
...
+  // intent. Failures leave intent null; the backfill path covers them later.
   if (typeof classifyIntent === 'function') {
     const rowsNeedingIntent = [...toInsert, ...toUpdate]
       .filter((r) => r.intent === null && hasText(r.text));
     if (rowsNeedingIntent.length > 0) {
       const intentByText = await classifyIntents(
         classifyIntent,
         rowsNeedingIntent.map((r) => r.text),
-        { timeoutMs: classifyIntentBatchTimeoutMs },
+        { timeoutMs: 8000 },
```
MysticatBot: issue (blocking): The batch timeout for intent classification was reduced from a configurable 30s default (via `PROMPT_INTENT_CLASSIFICATION_BATCH_TIMEOUT_MS` env var, resolved by `resolveBatchTimeoutMs`) to a hardcoded 8000ms. This is a 3.75x reduction in classification budget bundled into a bug fix without stated rationale.

**Why it matters:** For brands with many unclassified prompts in a single bulk upload, the batch will now time out far more aggressively, leaving more prompts with `intent: null` for the backfill path. More importantly, removing the env-var escape hatch means operators cannot recover during Azure latency spikes without a code deploy.

**How to fix:** Either (a) keep the env-var override with a lower default: `{ timeoutMs: resolveBatchTimeoutMs(env) }` where the default is 8000, or (b) if 8s is a deliberate choice, add a brief comment noting the rationale (e.g. calibrated against Lambda timeout minus write budget) so future maintainers understand the constraint.

src/support/prompts-storage.js
```diff
@@ -817,20 +823,38 @@ export async function upsertPrompts({
...
+        }
+      }
+    };
+    await Promise.all(
+      Array.from({ length: Math.min(UPDATE_CONCURRENCY, toUpdate.length) }, () => worker()),
     );
-    if (error) {
-      throw new Error(`Failed to update prompt: ${error.message}`);
+    if (errors.length > 0) {
+      throw new Error(`Failed to update prompt: ${errors[0].message}`);
```
MysticatBot: issue (blocking): The error contract changed from fail-fast (throw on first error, stop processing) to collect-all-then-throw. When one update fails and others succeed:

1. The function continues processing all remaining rows concurrently
2. Only `errors[0].message` is reported - subsequent failures are silently discarded
3. The caller receives a throw and cannot distinguish "1 of 25 failed" from "all failed"

**Why it matters:** The caller treats a throw as a total failure. Partial writes have already been committed to the DB, but the API response will be an error. Debugging is harder because subsequent distinct error causes (permission vs constraint violation) are lost.

**How to fix:** Either (a) abort remaining workers on first error (add a shared `aborted` flag workers check before picking the next item) to preserve the old fail-fast contract, or (b) if partial-success is intentional, include `{ succeeded: updated, failed: errors.length }` in the error message so operators can reason about state. Also consider logging all collected errors at warn level before throwing.

MysticatBot (changes requested):
Hey @jjenscodee,

**Verdict: Request changes** - two behavioral regressions bundled into the bug fix need attention.
**Complexity: HIGH** - medium diff; API surface risk flag (controllers/).
_Changes: Reactivates soft-deleted prompts on bulk re-upload instead of silently skipping them, adds bounded-concurrency update pool, and hardcodes intent-classification timeout (4 files)._
Note: Recommend a human read before merge - this change modifies a shared contract (OpenAPI spec `docs/openapi/prompts-v2-api.yaml`). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Important] Batch timeout reduced from configurable 30s to hardcoded 8s without stated rationale - `src/support/prompts-storage.js:788` (details inline)
2. [Important] Error semantics changed from fail-fast to partial-success-then-throw, reporting only first error - `src/support/prompts-storage.js:856` (details inline)

<details>

- nit: Shared `cursor` read-then-increment relies on no `await` between those two synchronous lines; a one-line comment noting the single-threaded atomicity assumption would protect against future refactors - `src/support/prompts-storage.js:833`
- suggestion: Concurrency test uses synchronous stubs that resolve immediately, so it validates row count but does not actually prove bounded concurrency (peak in-flight never exceeds 20) - `test/support/prompts-storage.test.js:1145`
- suggestion: Reactivation tests assert `updateStub.callCount` and `result.updated` but do not inspect the patch to confirm `status: 'active'` was set; the intent-preservation test already demonstrates the pattern - `test/support/prompts-storage.test.js:1212`
- suggestion: If `deleted` status can ever be set by moderation (not just user-initiated removal), the reactivation path would bypass that enforcement; worth a brief comment clarifying the contract - `src/support/prompts-storage.js:711`
</details>

- Prior non-blocking suggestions about intent-fallback comment and redundant prompt_id override remain unaddressed (author's prerogative on non-blocking items).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 59s | <b>Cost:</b> $5.62 | <b>Commit:</b> <code>64de8babf83a7d96725ba2ec325aa0e9d6d1cf52</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (changes requested):
Hey @jjenscodee,

**Verdict: Request changes** - the existing-prompts SELECT adds `intent` without the schema-fallback guard the rest of the file uses.
**Complexity: HIGH** - medium diff; API surface risk flag (controllers/).
_Changes: Reactivates soft-deleted prompts on bulk re-upload, adds bounded-concurrency update pool, and restores configurable intent-classification timeout (4 files)._
Note: Recommend a human read before merge - this change modifies a shared contract (OpenAPI spec `docs/openapi/prompts-v2-api.yaml`). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Important] Unguarded `intent` column in existing-prompts SELECT will break in column-absent environments - `src/support/prompts-storage.js:648` (details inline)

<details>

- nit: OpenAPI description says "counted as skipped" but no `skipped` field exists in the 201 response schema; consider "neither counted as created/updated nor returned" - `docs/openapi/prompts-v2-api.yaml:128`
- suggestion: Concurrency test uses synchronous stubs and would pass even with an incorrect cursor implementation; consider asserting peak in-flight calls stays at `Math.min(UPDATE_CONCURRENCY, toUpdate.length)` - `test/support/prompts-storage.test.js:1145`
- suggestion: Reactivation-by-prompt_id test asserts `updated=1` but does not inspect the update payload for `status: 'active'`; the intent-preservation test already demonstrates the pattern - `test/support/prompts-storage.test.js:1212`
</details>

- Batch timeout parameter restored with 8000 default (configurable via caller)
- Error message now reports all failed updates with count and joined messages

src/support/prompts-storage.js
```diff
@@ -643,7 +645,7 @@ export async function upsertPrompts({
 
   let existingQuery = postgrestClient
     .from('prompts')
-    .select('id,prompt_id,text,regions,status')
+    .select('id,prompt_id,text,regions,status,intent')
```
MysticatBot: issue (blocking): The existing-prompts SELECT adds intent directly in the column string without the withMissingIntentFallback guard that every other intent-aware query in this file uses (e.g., listPrompts builds its select via buildSelect(includeIntent) and runs through the helper). In environments where the prompts.intent column does not yet exist (documented in the TODO referencing the pinned IT PostgREST image), PostgREST returns error code 42703. The result is data: null, error: ..., and since the error is not checked after the select, existing becomes null. Downstream, (existing || []) means no matches are found, so every prompt is treated as a new insert - causing duplicate constraint violations on re-upload. Fix: Wrap the select in withMissingIntentFallback following the listPrompts pattern, or conditionally include intent in the column list based on the cached column-support flag.

MysticatBot (approved):
Hey @jjenscodee,

**Verdict: Approve** - prior blocking finding (unguarded intent column) is addressed; no new blocking issues.
**Complexity: HIGH** - medium diff; API surface risk flag (controllers/).
_Changes: Fixes the existing-prompts SELECT to use `withMissingIntentFallback`, completing the schema-guard pattern across all intent-aware queries (4 files)._
Note: Recommend a human read before merge - this change modifies a shared contract (OpenAPI spec `docs/openapi/prompts-v2-api.yaml`). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: OpenAPI description says "counted as skipped" but no `skipped` field exists in the 201 response schema; consider rewording to "are left unchanged and excluded from the response" - `docs/openapi/prompts-v2-api.yaml:128`
- suggestion: Concurrency test uses synchronous stubs that resolve immediately, so it validates row count but does not prove bounded concurrency (peak in-flight never exceeds 20); a stub with a 1ms delay would exercise real interleaving - `test/support/prompts-storage.test.js:1145`
- suggestion: Reactivation tests assert `updateStub.callCount` and `result.updated` but do not inspect the patch payload for `status: 'active'`; the intent-preservation test demonstrates the pattern already - `test/support/prompts-storage.test.js:1212`
</details>

- Unguarded `intent` column in existing-prompts SELECT now wrapped in `withMissingIntentFallback()`, consistent with all other intent-aware queries in this file.
- Batch timeout parameter restored as configurable function parameter (8000ms default).
- Error reporting now aggregates all failed update messages with count.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 1s | <b>Cost:</b> $4.18 | <b>Commit:</b> <code>6a1a28b6c5ce83dd99e8b3abb2c856ef2c0d45d7</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

igor-grubic (approved):

MysticatBot (approved):
Hey @jjenscodee,

**Verdict: Approve** - prior approval stands; the new cursor-drain fix is correct and well-tested.
**Complexity: HIGH** - medium diff; API surface risk flag (controllers/).
_Changes: Fixes batch timeout effectiveness by draining the shared cursor on cap expiry so workers exit their loop instead of starting new LLM calls (6 files)._
Note: Recommend a human read before merge - this change modifies a shared contract (OpenAPI spec `docs/openapi/prompts-v2-api.yaml`). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: OpenAPI description says "counted as skipped" but no `skipped` field exists in the 201 response schema; consider "are left unchanged and excluded from the response" - `docs/openapi/prompts-v2-api.yaml:128`
- suggestion: Concurrency test uses synchronous stubs that resolve immediately; a stub with a short delay would exercise real interleaving and prove peak in-flight stays at the pool size - `test/support/prompts-storage.test.js:1145`
- suggestion: Reactivation tests assert `updateStub.callCount` and `result.updated` but do not inspect the patch payload for `status: 'active'`; the intent-preservation test demonstrates the pattern - `test/support/prompts-storage.test.js:1212`
</details>

- Batch timeout was ineffective after cap fired - workers continued starting new LLM calls. Fixed by draining the shared cursor in the timer callback; regression test confirms callCount stays at 1.
- Batch timeout parameter restored as configurable function parameter (8000ms default).
- Error reporting now aggregates all failed update messages with count.
- Unguarded `intent` column in existing-prompts SELECT now wrapped in `withMissingIntentFallback()`.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 29s | <b>Cost:</b> $3.76 | <b>Commit:</b> <code>69d1d26d553c6618439208aaf1a1700a92888224</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2640, opened 2026-06-18
feat(s2s): allow readAll consumers to use by-base-url and by-ims-org-id

alinarublea:
## What

Open two cross-tenant lookup endpoints to S2S consumers holding a `readAll` capability, while keeping tenant-scoped `read` consumers working for their own org (non-breaking):

- `GET /sites/by-base-url/:baseURL` → `['site:read', 'site:readAll']`
- `GET /organizations/by-ims-org-id/:imsOrgId` → `['organization:read', 'organization:readAll']`

The controllers (`getByBaseURL`, `getByImsOrgID`) bypass the org-scoping `hasAccess` check for admins or `readAll` S2S consumers; everyone else must still belong to the entity's org. Mirrors the existing `getAll` readAll pattern, including `[s2s-readall]` grant and `[acl]` deny logging.

## Why

A platform analytics consumer (DaaS) needs to resolve `domain → siteId → imsOrgId` across all tenants. It holds `site:readAll` / `organization:readAll` (already used for `GET /sites` and `GET /organizations`), but these two by-key endpoints required tenant-scoped `read` and enforced org-scoping, returning 403. Today the consumer must bulk-fetch all sites/orgs and join locally; this enables direct single-call resolution.

## ⚠️ Release ordering — do not merge before the dependency

Depends on `@adobe/spacecat-shared-http-utils` with multi-capability (array) route support: **adobe/spacecat-shared#1693**.

1. Merge + release spacecat-shared#1693.
2. Bump `@adobe/spacecat-shared-http-utils` here to that version.
3. Then merge this PR.

If the array route values ship against the current string-only wrapper, those routes would 403 **all** S2S consumers (including existing `site:read` callers). Holding until the bump lands.

## Behavior matrix (getByBaseURL / getByImsOrgID)

| Caller | Before | After |
|---|---|---|
| Full / read-only admin | 200 | 200 (no S2S DB fetch) |
| IMS user, in-org | 200 | 200 |
| IMS user, cross-org | 403 | 403 |
| S2S `read`, own org | 200 | 200 (preserved) |
| S2S `read`, cross-org | 403 | 403 (+ `[acl]` deny log) |
| S2S `readAll`, cross-org | 403 | **200** (+ `[s2s-readall]` grant log) |

## Tests

- New controller tests for both endpoints: readAll cross-org grant, tenant-scoped read preserved, cross-org without readAll denied (asserts deny log), admin bypass without DB fetch.
- Route-map + drift tests updated to handle array values (`.flat()`).
[...]

PR #2674, opened 2026-06-23
fix(auth): restore read-only admin access on by-key + agentic-traffic reads

ravverma:
## Summary

Read-only admins regressed to **403** on three LLMO GET endpoints:
- `GET /organizations/by-ims-org-id/:imsOrgId`
- `GET /sites/by-base-url/:baseURL`
- `GET /sites/:siteId/agentic-traffic/kpis-trend`

## Root cause

`abb0775` (#2640) changed the first two routes' required capability from a single string to an **array** (`['organization:read', CAP_ORG_READ_ALL]` etc.) so a `readAll` S2S consumer could reach them. But `readOnlyAdminWrapper` classifies read-vs-write with `capability.split(':')` — string-only. An array has no `.split`, so the wrapper throws and its catch-all **fail-closes to 403** for read-only admins. (`s2sAuthWrapper` got array support in `spacecat-shared#1693`; the RO-admin wrapper never did.)

The `kpis-trend` endpoint is a separate but related gap: it lived in `INTERNAL_ROUTES` (deliberately unmapped), so for RO admins it was **ownership-gated** rather than taking the read fast-path → 403.

## Approach

Per team policy, we **don't widen a route's capability** (and don't teach the RO-admin wrapper to accept arrays) to satisfy an S2S consumer. A consumer missing a capability requests it from the **S2S admins**. So:

- **Revert `abb0775`'s array approach** — `by-ims-org-id` → `'organization:read'`, `by-base-url` → `'site:read'`; controllers drop the `readAll`/admin org-scoping bypass; tests restored to single-capability expectations. (http-utils left at main's version — no dep churn.)
- **Map `kpis-trend` to `'site:read'`** (mirrors the existing `agentic-traffic/has-data` entry) so RO admins and the UI hit the read fast-path. Removed from `INTERNAL_ROUTES` to preserve the disjointness invariant.

## Follow-up (flagged, not in this PR)

Sibling agentic-traffic GET reads still in `INTERNAL_ROUTES` (`kpis`, `by-region`, `by-category`, `by-page-type`, `by-status`, `by-user-agent`, `by-url`, `filter-dimensions`, `weeks`, `movers`) will show the same RO-admin behavior if the UI uses them — should likely also move to `site:read`. Confirm UI usage and batch.

## Test plan

- [ ] CI (unit + bundle). Local run not possible here (needs node 24 + `npm install`).
- [x] Syntax-checked `required-capabilities.js`; verified `INTERNAL_ROUTES` ∩ `routeRequiredCapabilities` disjointness for `kpis-trend`.


src/controllers/organizations.js
```diff
@@ -186,24 +186,9 @@ function OrganizationsController(ctx, env) {
...
-    // See READALL_CAPABILITY_DESIGN.md.
-    const { log } = ctx;
-    const requestId = context?.invocation?.id || 'unknown';
-    const isAdmin = accessControlUtil.hasAdminReadAccess();
-    const s2sResult = isAdmin
-      ? { allowed: false, reason: 'admin-bypass' }
-      : await accessControlUtil.hasS2SCapability(CAP_ORG_READ_ALL);
-    if (!isAdmin && !s2sResult.allowed && !(await accessControlUtil.hasAccess(organization))) {
-      log.info(`[acl] Denied GET /organizations/by-ims-org-id - reason=${s2sResult.reason} clientId=${s2sResult.clientId || 'n/a'} consumerId=${s2sResult.consumerId || 'n/a'} requestId=${requestId}`);
+    if (!await accessControlUtil.hasAccess(organization)) {
```
MysticatBot, line 189: question: Was any S2S consumer provisioned with `organization:readAll` specifically for `by-ims-org-id` since PR 2640 merged? The revert deliberately removes cross-tenant access per team policy. If a consumer was provisioned in the interim, they will need `organization:read` + org membership (or to request the capability via the S2S admin process). Worth a quick check of production logs for `[s2s-readall] GET /organizations/by-ims-org-id` entries.

MysticatBot (approved):
Hey @ravverma,

**Verdict: Approve** - well-scoped regression fix with appropriate test updates.
**Complexity: HIGH** - medium diff; API surface change (route capabilities).
_Changes: Restores read-only admin access on by-key and agentic-traffic GET endpoints by reverting the array-capability approach and simplifying controller access checks (7 files)._

<details>

- question: Was any S2S consumer provisioned with `organization:readAll` / `site:readAll` specifically for `by-ims-org-id` or `by-base-url` since PR 2640 merged? The revert removes cross-tenant access per team policy, which is correct, but confirming no active consumer depends on it avoids a second regression. - `src/controllers/organizations.js:189`
- nit: Removing `.flat()` from the drift-coverage tests creates a silent dependency on the format-validation test (which rejects arrays). Keeping `.flat()` (idempotent on strings) is a zero-cost belt-and-suspenders approach. - `test/routes/capability-constants.test.js:49`
- nit: The `kpis-trend` promotion from `INTERNAL_ROUTES` to `routeRequiredCapabilities` widens the S2S access surface (any consumer with `site:read` can now call it). This is intentional per the code comment but the PR title says "restore" rather than "expand" - worth noting in the description. - `src/routes/required-capabilities.js:384`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 25s | <b>Cost:</b> $8.63 | <b>Commit:</b> <code>f77be98aebfb312f13e923df8450f1dddd926ae4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2675, opened 2026-06-24
fix: timebox overrideBaseURL detection so onboarding beats the CDN timeout | LLMO-5606

igor-grubic:
## Problem

The first real call to the paid site-only onboarding endpoint ([LLMO-5606], `POST /v2/orgs/:spaceCatId/llmo/onboard-site`) returned a **503** to the client. CloudWatch shows the request actually **succeeded server-side in ~15.2s** — but Fastly's **first-byte timeout (~15s)** fired first, so the client got a `503 first byte timeout` even though the site was created.

Per-step timing of the invocation:

| Step | Elapsed |
|---|---|
| cold start + auth + org + site + entitlement | ~1.1s |
| SharePoint copy | ~2.5s |
| GitHub `helix-query.yaml` | ~1.0s |
| **`determineOverrideBaseURL` (Ahrefs/SEO) + CDN detect** | **~10s ⟵ dominant** |
| brandalf flag + triggerAudits + return | ~0.4s |

The `determineOverrideBaseURL` step fired **dozens of slow Semrush/SEO API calls** (~10s) — that's the budget-blower.

## Fix

`overrideBaseURL` detection is a **best-effort refinement** (it only tunes which host audits scrape — www vs apex), so it shouldn't block the response. This timeboxes it via a small `settleWithin(promise, timeoutMs, fallback)` helper: the detection is capped at **5s** and resolves to `null` on timeout **or** error, so it's skipped (non-fatal) rather than blocking or failing onboarding. The synchronous path drops to ~6–7s, comfortably under Fastly's window; the override is still applied in the common fast case.

It's in the shared `performLlmoOnboarding`, so the admin/Slack onboarding paths get the same protection.

## Notes
- Not a deferral/architecture change — stays fully synchronous (the LLMO-5606 execution model), just bounded.
- Secondary follow-up worth filing: the **dozens of SEO calls** for one detection look inefficient and could be reduced independently.

## Testing
- New unit tests for `settleWithin` (resolves fast → value; rejects → fallback; exceeds timeout → fallback).
- Existing `performLlmoOnboarding` override tests still pass (the "override is set" test no longer mocks `setTimeout` to fire immediately, which would trip the new timebox).
- Full suite green (12,467 passing); `llmo-onboarding.js` at 100% line/branch coverage.

[LLMO-5606]: https://jira.corp.adobe.com/browse/LLMO-5606

### DIFF
--- a/src/controllers/llmo/llmo-onboarding.js
+++ b/src/controllers/llmo/llmo-onboarding.js
@@ -63,6 +63,36 @@ export const ASO_DEMO_ORG = '66331367-70e6-4a49-8445-4f6d9c265af9';
 export const ASO_CRITICAL_SITES = [];
 const LLMO_ONBOARDING_PUBLISH_TRIGGER = 'trigger:llmo-onboarding-publish';
 
+// Cap for the best-effort Ahrefs/SEO overrideBaseURL detection during onboarding.
+// It can fire many slow SEO API calls (~10s observed) and, on the synchronous
+// onboarding path, push the response past the CDN first-byte timeout (~15s) — the
+// client gets a 503 even though onboarding succeeded. (LLMO-5606 follow-up.)
+const OVERRIDE_DETECT_TIMEOUT_MS = 5000;
+
+/**
+ * Awaits `promise`, but resolves to `fallback` if it rejects or doesn't settle
+ * within `timeoutMs`. The underlying promise is left running and any late
+ * rejection is swallowed, so a slow/flaky best-effort step can neither block past
+ * the cap nor surface as an unhandled rejection.
+ * @param {Promise<*>} promise - The in-flight work.
+ * @param {number} timeoutMs - Max time to wait before giving up.
+ * @param {*} fallback - Value to resolve to on timeout or rejection.
+ * @returns {Promise<*>} The promise's value, or `fallback`.
+ */
+export async function settleWithin(promise, timeoutMs, fallback) {
+  let timer;
+  try {
+    return await Promise.race([
+      Promise.resolve(promise).catch(() => fallback),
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @igor-grubic,

**Verdict: Approve** - clean, well-scoped incident fix with solid test coverage.
**Complexity: MEDIUM** - small diff; API surface (controllers/).
_Changes: Timeboxes the best-effort SEO/Ahrefs override detection at 5s via a new `settleWithin` helper so onboarding responds within the CDN first-byte timeout (2 files)._

<details>

- suggestion: Add a log line at the call site when `overrideBaseURL` resolves to `null` to distinguish "detection found nothing" from "detection timed out or threw" - without it, future CDN-adjacent debugging requires inference from the absence of logs rather than an explicit signal - `src/controllers/llmo/llmo-onboarding.js:1559`
- suggestion: Consider extracting `settleWithin` to a shared utility (e.g. `src/utils/promise-utils.js`) - the timebox-a-best-effort-call pattern will recur given the 15s Fastly first-byte window, and burying a reusable primitive in a 1600-line controller makes it hard to discover - `src/controllers/llmo/llmo-onboarding.js:82`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 22s | <b>Cost:</b> $3.50 | <b>Commit:</b> <code>1bccb3a4d420811e6afc193182301b23be720fc7</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jjenscodee (approved):

PR #2676, opened 2026-06-24
feat(state-mappings): add state-layer access-mapping endpoints (dev-gated)

ravverma:
## What

Adds the hybrid-permission-model **state-layer management + capability introspection** endpoints to api-service:

| Method | Route |
|--------|-------|
| GET    | `/state/access-mappings` |
| GET    | `/state/access-mappings/history` |
| POST   | `/state/access-mappings` |
| PATCH  | `/state/access-mappings/:id` |
| GET    | `/product/capabilities` |
| GET    | `/user/capabilities/:resourceId` |
| GET    | `/organizations/:organizationId/permission/audit-logs` |

## Authorization

The controller **self-gates** on FACS JWT permissions — `<product>/can_manage_users` for CRUD, `<product>/can_view` for capability introspection — with **per-resource scoping** for state-layer managers (hybrid-model flow 8.3: a state manager acts only on resources where it holds state `can_manage_users`; FACS managers and admins remain org-wide). It emits append-only audit events (a failed audit write only warns; it never fails the mapping op) and exposes `updatedBy`/`updatedAt`. PATCH accepts an empty `grantedCapabilities` array; there is **no bulk DELETE**.

## ⚠️ Dev-only blocker (temporary)

`facsWrapper` is **not yet attached** in api-service (that's the follow-up step). Until it is, the controller restricts every one of these endpoints to **`AWS_ENV === 'dev'`** — they return **404** in all other environments, fully hiding the staged feature in stage/prod. This is a single `devOnly` wrapper at the controller's return block; it is removed once `facsWrapper` fronts the routes upstream (the controller's `can_manage_users` / `can_view` gating is the permanent authorization layer).

## Notes for review

- `facsWrapper` is intentionally **not** wired into the `.with()` chain here — only the routes + controller land. Redirects/ASO-overlay handlers are untouched.
- The routes are classified as `INTERNAL_ROUTES` (never reachable by S2S consumers).
- `routes/facs-capabilities.js` is included because the controller consumes `PRODUCTS_CAPABILITIES` / `PRODUCTS_ROUTES` / resource-param aliases from it. The previously-suspended `routeFacsCapabilities` route-coverage suite is **re-enabled** in this PR, with the route classification reconciled against current `main` (stale entries dropped; new ai-visibility/serenity/agentic routes classified; redirects.txt + onboard-site marked internal).
- `support/postgrest-availability.js` extracts the PostgREST guard; the `requirePostgrestForV2Config` alias is ready for `brands.js` to adopt in a follow-up.

## Testing

- `npm test` → **12656 passing**, 0 failing. Coverage gate met (100% lines).
- `npm run docs:lint` → 0 errors (204 pre-existing baseline warnings).
- Postgres IT suite added: brandManager persona + flow-8.3 resource-scoping, audit-event seeding (writer JWT). IT server boots with `AWS_ENV=dev`.


src/controllers/state-access-mappings.js
```diff
@@ -0,0 +1,1061 @@
...
+    if (scopeErr) {
+      return scopeErr;
+    }
+
+    const queryParams = getQueryParams(ctx);
+    const decoded = decodeCursor(queryParams.cursor);
+    const limit = clampLimit(filters.limit);
+    // Helper applies a single page-size limit; for cursor pagination we
+    // over-fetch one row to detect whether a next page exists.
+    filters.limit = limit + 1 + (decoded?.offset ?? 0);
```
MysticatBot: **issue (blocking):** The pagination pattern `filters.limit = limit + 1 + (decoded?.offset ?? 0)` fetches rows 0…(offset+pageSize+1) from PostgREST, then slices to `[offset, offset+limit]`. Page N fetches N×pageSize rows from the database every time.

At page 10 with limit=200, this requests 2001 rows from PostgREST and discards the first 2000. The same pattern repeats in `listHistory` and `getAuditLogs`. Additionally, the utility’s `MAX_LIST_LIMIT=500` cap means pagination silently breaks once offset+limit+1 exceeds 500 (the slice returns fewer rows than expected and `hasMore` evaluates to false).

**Fix:** Pass the offset directly to PostgREST’s `.range(offset, offset+limit)` header, or switch to keyset pagination (e.g. `created_at < last_seen` cursor). The cursor already encodes the offset — forward it to the DB layer rather than applying it client-side.

src/controllers/state-access-mappings.js
```diff
@@ -0,0 +1,1061 @@
...
+      return { orgWide: true, managedResourceIds: null };
+    }
+    const manageCap = `${product.toLowerCase()}/can_manage_users`;
+    const userIdent = resolveCallerUserIdent(ctx);
+    const { postgrestClient } = ctx.dataAccess.services;
+    const queries = [
+      listFacsAccessMappings(postgrestClient, {
+        imsOrgId, product, subjectType: 'org', subjectId: imsOrgId, limit: MAX_PAGE_SIZE,
+      }),
+    ];
```
MysticatBot, line 395: **issue (blocking):** `resolveManageAuthority` queries `listFacsAccessMappings` with `limit: MAX_PAGE_SIZE` (200). A state-layer manager who holds `can_manage_users` on more than 200 resources will have their `managedResourceIds` set silently truncated — they will be denied access to resources beyond the first 200.

This is a correctness bug in permission evaluation: the failure mode is a false 403 (deny when it should allow).

**Fix options:**
1. Paginate through all bindings here (loop until no more rows).
2. Use a targeted query for the specific resource being acted upon (`WHERE resource_id = ? AND capability = 'can_manage_users'`) — both correct and more efficient than building the full set.
3. Document a supported maximum and enforce it at grant time.

MysticatBot (changes requested):
Hey @ravverma,

**Verdict: Request changes** — two correctness issues in pagination and authority resolution must be addressed before the dev gate is removed.
**Complexity: HIGH** — large diff; auth/permission model + API surface + security signal.
_Changes: Adds hybrid-permission-model state-layer management and capability introspection endpoints (7 routes), dev-gated until facsWrapper is attached (26 files, ~7k lines)._
Note: Recommend a human read before merge — this change modifies shared OpenAPI contracts (`docs/openapi/schemas.yaml`, `docs/openapi/state-access-mappings-api.yaml`).
Note: CI codecov/patch check is failing (99.3% coverage, 21 lines uncovered in controller).

## Must fix before merge

**1. [Important] Offset pagination over-fetches all preceding rows and slices client-side — O(N²) total DB cost and breaks past the 500-row utility cap**
`src/controllers/state-access-mappings.js` (listMappings, listHistory, getAuditLogs)

The pattern `filters.limit = limit + 1 + (decoded?.offset ?? 0)` asks PostgREST for all rows up to the offset, then discards everything before it. Page 10 at limit=200 requests 2001 rows but uses only 200. Worse: the utility clamps to `MAX_LIST_LIMIT = 500`, so once offset + limit + 1 > 500, the controller slices an array shorter than the offset and returns an empty page with `cursor: null` — pagination silently stops even if more rows exist.

The controller also uses `MAX_PAGE_SIZE = 200` while the utility uses `MAX_LIST_LIMIT = 500` — a dual-clamping contract that is confusing and produces surprising behavior at boundary values.

**Fix:** Pass the offset directly to PostgREST’s `.range(offset, offset + limit)` header, or switch to keyset pagination (`WHERE created_at < last_seen_value`). The cursor already encodes the offset — forward it to the DB layer instead of applying it client-side. Unify the limit constant to one place.

**2. [Important] `resolveManageAuthority` silently capped at 200 rows — produces false 403s for high-cardinality managers**
`src/controllers/state-access-mappings.js` (~line 872)

The function queries the caller’s bindings with `limit: MAX_PAGE_SIZE` (200). A state-layer manager holding `can_manage_users` on >200 resources gets a truncated `managedResourceIds` set. Authority checks against resources beyond the first 200 return a false 403 — a silent authorization failure with no error, no log.

**Fix options:**
1. Paginate exhaustively when building the managed set.
2. Replace the full-set approach with a targeted single-resource query (`WHERE resource_id = ? AND capability = 'can_manage_users'`) for create/patch operations where the target resource is known.
3. Document a supported maximum and enforce it at grant time (reject creating >N bindings per user).

**3. [Important] Access control pattern diverges from documented `AccessControlUtil` without CLAUDE.md acknowledgment**
`src/controllers/state-access-mappings.js` (~line 841)

The codebase CLAUDE.md documents `AccessControlUtil` as the standard access control pattern. This controller implements a parallel FACS-specific authorization model (resolveManageAuthority + callerHasFacsManageUsers) without updating documentation to acknowledge the divergence. Future contributors will not know which pattern governs.

**Fix:** Add a section to the repo CLAUDE.md (or a doc) explaining that state-access-mapping endpoints use FACS-native authorization (distinct from AccessControlUtil) and why.

<details>

- **nit:** `normalizeImsOrgId` inlined with a TODO but no tracking issue — will linger silently. Consider filing a follow-up. (`src/controllers/state-access-mappings.js:39`)
- **suggestion:** `getProductCapabilities` skips `preamble()` (no PostgREST guard, no IMS org validation). Intentional since it reads in-memory config, but inconsistent with the defense-in-depth posture of sibling handlers. Document the intent. (`src/controllers/state-access-mappings.js:~1330`)
- **suggestion:** `grantedCapabilities` array is not deduplicated before storage. `["llmo/can_view", "llmo/can_view"]` passes validation and is stored verbatim. A one-liner `[...new Set(grantedCapabilities)]` would prevent dirty rows.
- **suggestion:** Cursor offset has no upper-bound validation in `decodeCursor`. A crafted cursor with a very large offset produces a misleading empty-page response with `cursor: null` rather than a clear error.
- **suggestion:** Consider a delivery guarantee (SQS/structured log) for audit events before removing the dev gate — fire-and-forget with `log.warn` on failure means permanently lost audit records on transient PostgREST failures. (Out of scope for this PR, worth tracking.)

</details>

---
_Reviewed by: staff-engineer, architect, conventions-reviewer, security-researcher, qa-engineer (5 specialists)._

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 4m 28s | <b>Cost:</b> $9.20 | <b>Commit:</b> <code>afefb811f66bf3213de128996ed95e68dc51e88b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ravverma:
Thanks for the thorough review — addressed in `7b06775f`.

## Must-fix

**1. Offset pagination over-fetch / breaks past the 500-row cap** — fixed. The cursor offset is now forwarded to PostgREST's `.range(offset, offset + limit - 1)` in the three list helpers (`listFacsAccessMappings`, `listFacsAccessMappingHistory`, `listFacsAccessMappingAuditEvents`); the controller no longer over-fetches `offset + limit + 1` and slices client-side. It fetches exactly `limit + 1` rows from `offset` (the +1 only for `hasMore` detection), so the util's 500-row cap is never the binding constraint for a page and pagination can't silently stop. Applied across `listMappings`, `listHistory`, `getAuditLogs`.

**2. `resolveManageAuthority` truncated at 200 → false 403s** — fixed. It now pages the caller's bindings exhaustively (bounded by `MAX_AUTHORITY_PAGES = 50`, i.e. 10k bindings) when building `managedResourceIds`, so a manager with >200 managed resources is no longer truncated. If the bound is ever hit it emits a `log.warn` — never a silent truncation. Covered by a multi-page test (binding on page 2 is found) and a page-cap test (warns, stays deterministic).

**3. Divergence from `AccessControlUtil` undocumented** — fixed. Added a "FACS-native authorization" subsection to `CLAUDE.md` under Access Control, explaining that these endpoints evaluate JWT `facs_permissions` ∪ state-layer `granted_capabilities` directly (org-wide FACS manager vs resource-scoped state-layer manager) and why, plus the future `facsWrapper` attachment.

## Non-blocking

- **Dedup `grantedCapabilities`** — `[...new Set(...)]` before storage in both create and patch.
- **Cursor offset upper bound** — `decodeCursor` now rejects offsets above `MAX_CURSOR_OFFSET` (1,000,000), treating them as offset 0.
- **`getProductCapabilities` skips preamble** — documented the intent inline (reads only in-memory catalog; no DB / org access to guard).
- **`normalizeImsOrgId` TODO** — reworded to tie the swap to the follow-up PR that bumps `@adobe/spacecat-shared-http-utils` and attaches `facsWrapper` (spacecat-shared#1717), rather than a dangling TODO.
- **Audit-event delivery guarantee** — agree it's worth a durable channel; tracking as a follow-up since it's out of scope here (the dev-only gate keeps these endpoints non-prod until `facsWrapper` is attached anyway).

## CI / coverage

`npm test` → **12664 passing**, `npm run docs:lint` → 0 errors. The new logic is covered; the few remaining uncovered lines in the controller are pre-existing defensive branches (DTO null-guards, an unreachable multi-resource-type path — no product currently declares >1 resource type) outside this change's diff.

MysticatBot (approved):
Hey @ravverma,

**Verdict: Approve** - all prior blocking findings are addressed; remaining observations are non-blocking performance notes.
**Complexity: HIGH** - large diff; auth/permission model + API surface + security signal.
_Changes: Addresses all prior review findings by switching to DB-side pagination, exhaustive authority resolution with bounded paging, cursor offset validation, capability deduplication, and CLAUDE.md documentation of the FACS-native auth pattern (26 files)._
Note: Recommend a human read before merge - this change modifies shared OpenAPI contracts (`docs/openapi/schemas.yaml`, `docs/openapi/state-access-mappings-api.yaml`) and root `CLAUDE.md`.
Note: CI build checks are pending - confirm they pass before merge.

<details>

- suggestion: `resolveManageAuthority` fetches all bindings (including `can_view`) when only `can_manage_users` rows are needed for the authority check. A PostgREST `.contains('granted_capabilities', ...)` filter would eliminate most paging for realistic managers. Not a correctness issue (bounded by MAX_AUTHORITY_PAGES) but worth a follow-up when production load patterns are observed. - `src/controllers/state-access-mappings.js:~370`
- nit: `getUserCapabilities` hard-caps state-layer queries at MAX_PAGE_SIZE (200) without paging, unlike `resolveManageAuthority`. Unlikely to matter in practice since a single (subject, resource) pair should not accumulate >200 active binding rows, but the inconsistency is worth noting. - `src/controllers/state-access-mappings.js:~958`
- nit: Two independent `clampLimit` functions exist with different caps (controller: 200, utils: 500). The naming collision is fine (module-scoped) but a name like `clampApiLimit` in the controller would signal the intentional divergence to future readers. - `src/controllers/state-access-mappings.js:~166`

</details>

- Offset pagination over-fetch (O(N^2) DB cost) - now uses PostgREST `.range()` for DB-side pagination
- `resolveManageAuthority` silently capped at 200 rows causing false 403s - now exhaustively pages with bounded loop and warning log
- Access control pattern divergence undocumented - CLAUDE.md now documents FACS-native authorization
- `normalizeImsOrgId` TODO with no tracking - rewritten with clear rationale and linked follow-up
- `getProductCapabilities` skips preamble without explanation - comment added
- `grantedCapabilities` not deduplicated - `[...new Set()]` applied to both create and patch
- Cursor offset unbounded - MAX_CURSOR_OFFSET validation added to `decodeCursor`

---
_Reviewed by: staff-engineer, architect, conventions-reviewer, security-researcher, qa-engineer (5 specialists)._

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 3s | <b>Cost:</b> $7.57 | <b>Commit:</b> <code>ec1d1b556a69a2d8f367677b8b1d3115f8ffc90c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2654, opened 2026-06-20
feat(llmo): add CloudFront wizard read endpoints

ABHA61:
## What

Adds the read-only backend endpoints that power the CloudFront BYOCDN **"Deploy routing in CloudFront"** wizard in LLMO UI. The api-service assumes the customer's cross-account **connector role server-side** (no AWS credentials in the browser) and performs read-only CloudFront calls via that role.

## Endpoints (all POST, all in `INTERNAL_ROUTES` — admin/IMS-only, not exposed to S2S)

- `POST /sites/:siteId/llmo/edge-optimize/connect` — verify the connector role is assumable; returns `{ connected }` so the wizard can poll while the customer is still creating the role.
- `POST /sites/:siteId/llmo/edge-optimize/distributions` — list the account's CloudFront distributions.
- `POST /sites/:siteId/llmo/edge-optimize/prerequisites` — confirm connector-role assumability + CloudFront read access.
- `POST /sites/:siteId/llmo/edge-optimize/origins` — read a distribution's origins (and detect whether an Edge Optimize origin already exists).
- `POST /sites/:siteId/llmo/edge-optimize/behaviors` — read a distribution's default + ordered cache behaviors.

New support module `src/support/edge-optimize.js`: `assumeConnectorRole` (STS AssumeRole with the per-session external ID), `listCloudFrontDistributions`, `getDistributionConfig`. Adds `@aws-sdk/client-sts` and `@aws-sdk/client-cloudfront`.

## Stacked on #2642

This branch was cut from the bootstrap-url branch (#2642), so its diff also includes `getEdgeOptimizeBootstrapUrl`. Merge #2642 first (then this rebases down to just the read endpoints), or this PR can supersede it.

## ⚠️ TEMP testing defaults — MUST be removed before prod

Carried from #2642 (marked with `TODO` in `getEdgeOptimizeBootstrapUrl`):
- `EDGE_OPTIMIZE_TEMPLATE_BUCKET` defaults to `llmo-edgeoptimize-cf-template`
- `EDGE_OPTIMIZE_TRUSTED_PRINCIPAL_ARN` defaults to `arn:aws:iam::682033462621:root` (dev) — **security-sensitive**; must be set to Adobe's backend role ARN in prod.

## Deferred (next phases)

Mutation steps are not in this PR: cache-policy detect/apply, Lambda@Edge, distribution associations, verify, revoke (these create/modify CloudFront resources and the long-running provisioning will be async).

## Tests

[...]

PR #2680, opened 2026-06-25
feat(llmo): CloudFront CDN log delivery with assume-role setup (LLMO-5566)

claudiaboldis:
## Summary

- Adds server-side STS `AssumeRole` (connector role) so the LLMO UI CloudFront wizard never needs AWS credentials in the browser — all CloudFront and CloudWatch Logs calls are performed by the Lambda
- Adds 5 read-only wizard endpoints (`connect`, `distributions`, `prerequisites`, `origins`, `behaviors`) that power the step-by-step CloudFront BYOCDN setup flow
- Adds `POST /sites/:siteId/llmo/cdn-log-delivery` (diagram step 8) — idempotently creates a CloudWatch Logs delivery source in the customer's account and links it to Adobe's cross-account delivery destination, enabling CloudFront standard log push to the cdn-logs S3 bucket

## New files

| File | Purpose |
|---|---|
| `src/support/edge-optimize.js` | `assumeConnectorRole` (STS), `listCloudFrontDistributions`, `getDistributionConfig`, idempotent step-on-poll deploy orchestrator |
| `src/support/cdn-log-delivery.js` | Provider-agnostic log-delivery registry; CloudFront path uses `logs:PutDeliverySource` + `logs:CreateDelivery` via the assumed role |

## Endpoints (all `POST`, all `INTERNAL_ROUTES` — admin/IMS-only)

- `POST /sites/:siteId/llmo/edge-optimize/connect`
- `POST /sites/:siteId/llmo/edge-optimize/distributions`
- `POST /sites/:siteId/llmo/edge-optimize/prerequisites`
- `POST /sites/:siteId/llmo/edge-optimize/origins`
- `POST /sites/:siteId/llmo/edge-optimize/behaviors`
- `POST /sites/:siteId/llmo/cdn-log-delivery`

## ⚠️ Cross-service dependencies (not in this PR — must land before prod)

1. **Connector-role IAM policy** (`customer-bootstrap-role.yaml` in `EDGE_OPTIMIZE_TEMPLATE_BUCKET`) must add `logs:PutDeliverySource`, `logs:CreateDelivery`, `logs:GetDeliverySource`, `logs:DescribeDeliveries` — today the role only grants CloudFront perms
2. **spacecat-auth-service** must provision the `cdn-logs-<org>` delivery destination + cross-account policy for the BYOCDN CloudFront path (the handler surfaces a clear "destination not provisioned" 400 if it's missing)

## Test plan

```bash
# verify connector role assumable
curl -s -X POST "https://spacecat.experiencecloud.live/sites/$SITE_ID/llmo/edge-optimize/connect" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"accountId":"123456789012","externalId":"<externalId>"}'
# Expected: HTTP 200 { "connected": true }
```

```bash
# list distributions via assumed role
curl -s -X POST "https://spacecat.experiencecloud.live/sites/$SITE_ID/llmo/edge-optimize/distributions" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"accountId":"123456789012","externalId":"<externalId>"}'
# Expected: HTTP 200 { "distributions": [...] }
```

```bash
# enable log delivery (idempotent)
curl -s -X POST "https://spacecat.experiencecloud.live/sites/$SITE_ID/llmo/cdn-log-delivery" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"accountId":"123456789012","externalId":"<externalId>","distributionId":"EDFDVBD6EXAMPLE"}'
# Expected: HTTP 200 { "alreadyExisted": false, "deliveryId": "..." }
# Re-run: HTTP 200 { "alreadyExisted": true }
```

## Related

- Closes LLMO-5566
- Based on PR #2654 (wizard read endpoints) and PR #2671 (log-forwarding endpoint) by @akbhardwaj

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Akash Bhardwaj", "Dominique Jäggi"]
```


MysticatBot (changes requested):
Hey @claudiaboldis,

**Verdict: Request changes** - two issues to address before merge (concurrency cap and HTTP status correctness).
**Complexity: HIGH** - large diff; API surface + new dependency.
_Changes: Adds CDN log delivery endpoints (single-distribution and bulk rescan) using STS AssumeRole to create CloudWatch Logs delivery sources in customer accounts (13 files)._
Note: Recommend a human read before merge - this change modifies the OpenAPI contract (`docs/openapi/`). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Important] Unbounded concurrency in `rescanCdnLogDelivery` will throttle on large accounts - `src/controllers/llmo/llmo-cloudfront.js:1228`. `Promise.allSettled` fires `createCdnLogDelivery` for ALL distributions in parallel with no concurrency cap. Each call issues 1-3 CloudWatch Logs API calls. A customer with 50+ distributions will hit CloudWatch Logs per-account throttling (10-25 TPS), causing most calls to fail with `ThrottlingException`. Fix: add a concurrency limiter (e.g. `p-limit(5)`) or batch distributions in small groups.
2. [Important] Missing server config returns 400 (client error) instead of 500 - `src/controllers/llmo/llmo-cloudfront.js:1164`. When `CDN_LOG_DELIVERY_DEST_ACCOUNT_ID` is not set, `badRequest(...)` tells the caller their input is wrong when the server is misconfigured. Fix: return `internalServerError(...)` instead (same applies in `rescanCdnLogDelivery` at line ~1218).

<details>

- nit: `docs/LLMO-5566/plan.md` references `src/controllers/llmo/llmo.js` and 16 endpoints, but the actual controller is `llmo-cloudfront.js` and this PR delivers 2 endpoints - the plan is misleading from day one
- suggestion: Both new handlers diverge from the existing `mutationErrorResponse` pattern used by sibling endpoints in this controller (which surfaces actionable AWS errors like `AccessDeniedException` and `ThrottlingException` as 400s instead of generic 500) - consider using `mutationErrorResponse` in the outer catch for consistency - `src/controllers/llmo/llmo-cloudfront.js:1270`
- suggestion: `rescanCdnLogDelivery` response includes raw `outcome.reason?.message` per distribution, which can expose internal AWS error details (ARNs, role names) to the caller - consider sanitizing to error category only - `src/controllers/llmo/llmo-cloudfront.js:1244`
</details>

src/support/cdn-log-delivery.js
```diff
@@ -0,0 +1,188 @@
...
+        return {
+          created: false,
+          alreadyExisted: true,
+          deliverySourceName,
+          deliveryId: existing.id,
+        };
+      }
+      nextToken = page.nextToken;
+    } while (nextToken);
+  }
```
MysticatBot, line 199: issue (blocking): The idempotency check has a TOCTOU window. If `enableCdnLogDelivery` and `rescanCdnLogDelivery` run concurrently for the same distribution, both can pass the `GetDeliverySource` existence check, both succeed at `PutDeliverySource` (idempotent PUT), and then the second `CreateDelivery` call hits a `ConflictException` that is unhandled - it bubbles to the outer catch and returns a 500 to the caller.

Fix: catch `ConflictException` (or `ResourceAlreadyExistsException`) from `CreateDelivery` and treat it as `{ created: false, alreadyExisted: true }`, mirroring the existing idempotency contract. This is the minimal change - the check-then-act pattern stays, but the race outcome is no longer an error.

Alternatively, if concurrent use is not expected for this admin tooling, document the mutual-exclusion constraint in the plan or a code comment.

test/controllers/llmo/llmo-cloudfront.test.js
```diff
@@ -2287,4 +2307,321 @@ describe('LlmoCloudFrontController', () => {
       expect(result.status).to.equal(403);
```
MysticatBot, line 2283: issue (blocking): CLAUDE.md rule: "New or modified endpoints must include integration tests in `test/it/` - add shared test logic in `shared/tests/`, seed data in `postgres/seed-data/`, and a wiring file in `postgres/`."

Both `enableCdnLogDelivery` and `rescanCdnLogDelivery` are new endpoints with unit tests only. Integration tests that exercise the full request lifecycle (auth gating, route resolution, response shape) through the PostgreSQL harness are required by the documented convention.

Note: these endpoints call external AWS APIs via assumed roles, so the IT scope would focus on the auth/routing/validation layers (mocking the AWS calls at the boundary), not the CloudWatch Logs interactions themselves.

MysticatBot (changes requested):
Hey @claudiaboldis,

**Verdict: Request changes** - two issues to address (concurrency race and missing integration tests).
**Complexity: HIGH** - large diff; API surface + new dependency + OpenAPI contract change.
_Changes: Adds CDN log delivery endpoints (single-distribution and bulk rescan) using STS AssumeRole to create CloudWatch Logs delivery sources in customer accounts, with bounded concurrency and proper error categorization (13 files)._
Note: Recommend a human read before merge - this change modifies the OpenAPI contract (`docs/openapi/`). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Important] Unhandled `ConflictException` on concurrent `CreateDelivery` calls - `src/support/cdn-log-delivery.js:167` (details inline)
2. [Important] Missing integration tests for both new endpoints - CLAUDE.md rule: "New or modified endpoints must include integration tests in `test/it/`" (details inline)

<details>

- suggestion: Delivery source name (`llmo-cf-{safeOrgId}-{distributionId}`) may exceed the CloudWatch Logs 60-character delivery-source name limit for long IMS org IDs - consider a defensive length check or hash-based truncation in `buildDeliverySourceName` - `src/support/cdn-log-delivery.js:89`
- suggestion: `createCdnLogDelivery` creates a new `CloudWatchLogsClient` per invocation - during a rescan with 50+ distributions this adds unnecessary object churn; consider accepting an optional pre-built client or lifting client creation to the caller - `src/support/cdn-log-delivery.js:135`
</details>

- Unbounded concurrency in `rescanCdnLogDelivery` now capped at `CDN_LOG_RESCAN_CONCURRENCY = 5` with a batched for-loop
- Missing server config (`CDN_LOG_DELIVERY_DEST_ACCOUNT_ID`) now correctly returns `internalServerError()` instead of `badRequest()`

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 27s | <b>Cost:</b> $8.49 | <b>Commit:</b> <code>a2e7638ffd078cd5730fc1711ba46b55252fbfb0</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @claudiaboldis,

**Verdict: Approve** - prior blocking findings are addressed; no new blocking issues.
**Complexity: HIGH** - large diff; API surface + new dependency.
_Changes: Adds CDN log delivery endpoints (single-distribution and bulk rescan) using STS AssumeRole with ConflictException race handling and bounded concurrency (15 files)._
Note: Recommend a human read before merge - this change modifies the OpenAPI contract (`docs/openapi/`). The bot review is a complement to, not a replacement for, a human read here.

<details>

- suggestion: `buildDeliverySourceName` can produce names exceeding the CloudWatch Logs 60-character delivery-source name limit for long IMS org IDs - consider a defensive length check or hash-based truncation - `src/support/cdn-log-delivery.js:89`
- suggestion: After a `ConflictException` race in `createCdnLogDelivery`, `findExistingDelivery()` could return undefined (eventual consistency), resulting in `deliveryId: null` in the response - consider a single retry or an explicit note in the response when the ID cannot be resolved - `src/support/cdn-log-delivery.js:199`
</details>

- Unhandled ConflictException on concurrent CreateDelivery calls now caught with idempotent fallback
- Missing integration tests now added (access-control and validation layers covered in test/it/)
- Unbounded concurrency in rescanCdnLogDelivery now capped at CDN_LOG_RESCAN_CONCURRENCY = 5
- Missing server config now correctly returns internalServerError() instead of badRequest()

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 11s | <b>Cost:</b> $6.05 | <b>Commit:</b> <code>6d39666c21acedd9e4d9ab324cd107ebe3afb640</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ABHA61:
**Reviewed the two new CDN log-delivery endpoints (`enableCdnLogDelivery`, `rescanCdnLogDelivery`), the `cdn-log-delivery.js` support module, wiring, tests, and OpenAPI — and cross-checked against the public CloudFront log-forwarding docs.** Structure, routing, capability gating, client return shapes, OpenAPI (no drift), and the core idempotency logic all verified clean; temp AWS credentials / externalId / role ARNs are never logged or returned. Items worth addressing, grouped by severity:

| # | Sev · type | Finding | Loc |
|---|---|---|---|
| 1 | **MED** · correctness | Idempotency matches on delivery-**source** name only, ignoring the destination ARN — an existing delivery to a *different/older* destination is reported `alreadyExisted` and never re-pointed, so if Adobe's destination ARN ever changes, logs silently keep flowing to the old place. | `cdn-log-delivery.js:157` |
| 2 | **MED** · correctness | `PutDeliverySource` throwing `ConflictException` (distribution already bound to another delivery source — e.g. a customer who set it up manually first) is not handled: enable returns 500, rescan marks it failed. The public CLI doc explicitly warns this exact error can occur, so it's a realistic idempotency break. | `cdn-log-delivery.js:191` |
| 3 | **MED** · correctness/UX | A valid-format but nonexistent/wrong-account `distributionId` makes `PutDeliverySource` throw `ResourceNotFoundException`, which the handler maps to "Adobe log destination is not provisioned — run cdn-logs provisioning first," sending ops to chase a nonexistent provisioning gap. | `llmo-cloudfront.js:1191` |
| 4 | **MED** · security | `enableCdnLogDelivery` enables forwarding for any caller-supplied `distributionId` in the account with no check that the distribution serves this site. In an AWS account shared across two IMS orgs, org X's admin can route org Y's distribution access logs into org X's cdn-logs bucket (confused-deputy / cross-org capture). The existing `assertDistributionServesSite` helper (with its `allowDomainMismatch` override) closes this — and also fixes #3. | `llmo-cloudfront.js:1153` |
| 5 | **MED** · correctness/UX | The AWS-error categorization set is CloudFront-only; CloudWatch delivery errors — `ServiceQuotaExceededException` (deliveries-per-account cap, likely on rescan), `ValidationException`, `ConflictException`, `ExpiredTokenException` — fall through to a generic 500 instead of an actionable 4xx. | `llmo-cloudfront.js:237` |
| 6 | **MED** · operational | `CDN_LOG_DELIVERY_DEST_ACCOUNT_ID` is a new required env var with no in-repo reference (absent from `.env.example`/README/config), and it isn't listed as a provisioning follow-up. If unset on deploy, both endpoints hard-500 for every call until it's added out-of-band. | `llmo-cloudfront.js:1173` |
| 7 | **MED** · test-fidelity | Tests assert only the *result* of `createCdnLogDelivery`, never the AWS command inputs (resourceArn, `logType`, recordFields, suffixPath, region, credentials) — a typo in any of these ships green and only surfaces at live onboarding. | `test/support/cdn-log-delivery.test.js:115` |
| 8 | **MED** · test-fidelity | Controller tests stub the helpers and never assert their args — passing the customer accountId as `adobeAccountId`, swapping `imsOrgId`, or handing over wrong credentials would misroute logs with all tests green. | `test/controllers/llmo/llmo-cloudfront.test.js:2346` |
| 9 | **LOW-MED** · efficiency | `DescribeDeliveriesCommand` is passed a `deliverySourceName` filter the API doesn't accept (SDK input is only `nextToken`/`limit`); it's silently dropped, so every call paginates all account deliveries and relies on a client-side `.find`. Correct today, but O(N×total) during rescan, and the comment misrepresents it as server-filtered. | `cdn-log-delivery.js:154` |
| 10 | **LOW-MED** · security | On categorized errors, enable returns the raw AWS message in the 400 body — for `AccessDenied` that includes Adobe's cross-account destination ARN (the dest account id) and the assumed-role ARN. The per-distribution rescan summary already avoids this by returning the error *name* only; enable should match. | `llmo-cloudfront.js:1205` |
| 11 | **LOW-MED** · test-fidelity | The "AccessDenied" test constructs `new Error('AccessDeniedException: …')` whose `.name` is `Error`, so it hits the 500 fallback, not the 4xx categorization branch — the test asserts 500 and passes, leaving the categorization path unexercised. | `test/…/llmo-cloudfront.test.js:2424` |
| 12 | **LOW-MED** · UX | Rescan has no `ResourceNotFoundException` short-circuit: if the destination isn't provisioned, every distribution fails and it returns 200 with all-failed and opaque per-item codes — no actionable message, unlike enable. | `llmo-cloudfront.js:1267` |
| 13 | **LOW-MED** · scalability | Rescan runs the whole account in one synchronous request reusing one set of assumed credentials (STS session is 15 min). A large account can exceed the ~60s gateway timeout and/or the 15-min credential lifetime, failing late batches with no summary returned. | `llmo-cloudfront.js:1243` |
| 14 | **LOW-MED** · cross-service | `toSafeAwsName` must match spacecat-auth-service byte-for-byte and only strips `@AdobeOrg`/`@`; a non-standard `imsOrgId` (dots/spaces) would produce an invalid source name or a divergent destination ARN (surfacing as the misleading #3 message). Standard IMS org ids are fine. | `cdn-log-delivery.js:58` |

**Nits (optional, non-blocking):**
- `accountId` is derived via `replace(/\D/g,'')` + length-12 rather than a strict `/^\d{12}$/`, so embedded junk silently coerces to a 12-digit id (bounded by STS trust). `llmo-cloudfront.js:183`
- `findExistingDelivery`'s `do…while(nextToken)` has no max-page / seen-token guard (infinite-loop risk if the API echoes a stable token). `cdn-log-delivery.js:151`
- Untested reachable branches: partial-state fall-through (source exists, no delivery → `created:true`) and empty-distributions rescan. `cdn-log-delivery.test.js:134`
- Defense-in-depth: nothing server-side binds the client-supplied `externalId`/account to the site's org (STS bounds it; hardening = persist externalId at bootstrap and validate the account against the org). `llmo-cloudfront.js:1179`
- `ResourceAlreadyExistsException` in the CreateDelivery catch is dead code (CreateDelivery signals a dup via `ConflictException`, which is handled); the >60-char source-name hash-truncation is effectively unreachable for real ids; the two methods duplicate ~20 lines of preamble (a small shared helper would prevent drift); enable's `ResourceNotFoundException` branch writes no audit line; a new `CloudWatchLogsClient` is built per `createCdnLogDelivery` call.

**Please confirm:**
- **IAM permissions** — the code uses `logs:PutDeliverySource`, `logs:CreateDelivery`, `logs:GetDeliverySource`, and `logs:DescribeDeliveries`. Please confirm the bootstrap connector-role template grants all four (the public console doc lists only the first two), and that the template PR lands with or before this — otherwise the idempotency calls fail with AccessDenied.
- **Shared-client placement** — the new CloudWatch log-delivery control-plane (`cdn-log-delivery.js`) sits in api-service, while the connector-role AWS operations (`assumeConnectorRole`, `CloudFrontEdgeClient`) live in `@adobe/spacecat-shared-tokowaka-client`. It's the same class of connector-role operation and other consumers may need to (re)enable delivery, so worth deciding whether it belongs in the shared client (that also keeps the `toSafeAwsName` byte-for-byte contract in one place). Related direction: the CloudFront control-plane currently lives in `spacecat-shared-tokowaka-client`, but longer-term it should move into a dedicated `cloudfront-client` package — mirroring the existing `cloudflare-client` — and this log-delivery code should follow to that same home. Fine as a fast follow-up.

claudiaboldis:
**IAM permissions — confirmed.** Both `customer-bootstrap-role.yaml` and `customer-bootstrap-role-log-only.yaml` already grant all four permissions: `logs:PutDeliverySource`, `logs:CreateDelivery`, `logs:GetDeliverySource`, and `logs:DescribeDeliveries`. The templates are included in this PR (added in the same commit), so they land together — no sequencing risk.

**Shared-client placement — deferred, tracked.** Agreed that `cdn-log-delivery.js` belongs in a shared package long-term. Moving it now would block this PR on a shared-client release cycle and add unrelated churn. The plan is to follow the CloudFront control-plane when it migrates to a dedicated `cloudfront-client` package (mirroring `cloudflare-client`), and `cdn-log-delivery.js` + `toSafeAwsName` move there at that point — keeping the byte-for-byte contract in one place. Will track as a follow-up ticket.

MysticatBot (commented):
## AI Code Review — PR #2680

**Complexity:** HIGH (large diff; API surface + Infrastructure + Dependencies)
**Review type:** Re-review (branch rewritten since last review — full re-review)
**Grounding:** Implementation plan only (docs/LLMO-5566/plan.md); no formal spec
**Verdict:** REQUEST CHANGES — one critical production bug

---

### Summary

This PR adds CloudWatch Logs delivery for CDN access logs via cross-account STS AssumeRole. Two new internal endpoints (`log-delivery`, `log-rescan`) are added to the existing CloudFront onboarding controller. The implementation is well-structured with good security posture: strict input validation, server-derived delivery destination ARNs, cross-org guards, error message sanitization, bounded concurrency, and idempotent delivery creation with TOCTOU race handling.

Prior review findings from @ABHA61 have been addressed (idempotency, ConflictException handling, cross-org guard, error sanitization, .env documentation, bounded concurrency).

---

### Findings

#### Critical (Must Fix)

| # | Category | Location | Finding |
|---|----------|----------|---------|
| 1 | Bug | `src/controllers/llmo/llmo-cloudfront.js` (enableCdnLogDelivery, rescanCdnLogDelivery) | **`externalId` is always `undefined` — both endpoints will fail with STS AccessDenied in production.** The value is destructured from `validateCloudfrontCredentials()`, which only returns `{ accountId, distributionId }`. Every sibling endpoint in this controller gets `externalId` from `gateEdgeOptimizeWizard()` (server-derived from the site's IMS org). The new endpoints discard it by destructuring only `{ error, site }`. Unit tests don't catch this because `assumeConnectorRoleStub` accepts any args. **Fix:** `const { error, site, externalId } = await gateEdgeOptimizeWizard(...)` and remove `externalId` from the `validateCloudfrontCredentials` destructuring. Add a test assertion that the stub receives the expected externalId value. |

#### Important (Should Fix)

| # | Category | Location | Finding |
|---|----------|----------|---------|
| 2 | Operational | `src/controllers/llmo/llmo-cloudfront.js` (rescanCdnLogDelivery) | **No hardcoded upper bound on distributions when `CDN_LOG_RESCAN_MAX_DISTRIBUTIONS` is unset.** A customer account with hundreds of distributions could exhaust Lambda timeout or generate oversized responses. Consider a sensible hardcoded default (e.g., 200) rather than relying solely on an env var that may not be configured initially. |
| 3 | Performance | `src/controllers/llmo/llmo-cloudfront.js` (enableCdnLogDelivery, rescanCdnLogDelivery) | **Redundant `Organization.findById` lookup.** Both endpoints call `gateEdgeOptimizeWizard` (which internally resolves the IMS org) then immediately perform a second DB query for the same org. After fixing #1, expose `imsOrgId` from the gate or use the already-fetched org. |
| 4 | Validation | `src/controllers/llmo/llmo-cloudfront.js` (validateCloudfrontCredentials) | **`distributionId` format not validated beyond non-empty.** CloudFront IDs follow a strict pattern (`E` + 13 alphanumeric). A format regex would reject obviously invalid input earlier, reducing unnecessary AWS API calls. |
| 5 | Documentation | `src/controllers/llmo/llmo-cloudfront.js` (enableCdnLogDelivery comment) | **JSDoc comment is misleading.** States "externalId is the per-session UUID from bootstrap (client-supplied)" but the existing architecture uses a server-derived externalId (IMS org ID). Should be corrected to match the actual contract. |

#### Minor (Nice to Have)

| # | Category | Location | Finding |
|---|----------|----------|---------|
| 6 | Test coverage | `test/support/cdn-log-delivery.test.js` | **`recordFields` array only partially asserted.** Test checks one field out of 12. Consider `deep.equal` or checking array length. |
| 7 | Test coverage | `src/support/cdn-log-delivery.js` | **MAX_DESCRIBE_PAGES (50) pagination cap untested.** Tests verify 2-page pagination but not the safety cap. |
| 8 | Operational | `src/controllers/llmo/llmo-cloudfront.js` (rescanCdnLogDelivery) | **No inter-batch delay.** Bounded concurrency (5) is good, but rapid sequential batches could still trip throttling on larger accounts. Consider a configurable inter-batch delay. |
| 9 | Documentation | `templates/customer-bootstrap-role-log-only.yaml` | **`logs:DescribeDeliveries` on `Resource: '*'`** is a CloudWatch Logs API limitation, not a misconfiguration. Consider a comment for customer security teams reviewing the template. |

---

### Prior Findings Status

All findings from the previous human review (@ABHA61) have been addressed in this revision:
- Idempotency matching and ConflictException handling ✓
- Cross-org log capture guard (`assertDistributionServesSite`) ✓
- Error message sanitization (only `error.name` returned) ✓
- `.env.example` documentation ✓
- Bounded concurrency (`CDN_LOG_RESCAN_CONCURRENCY = 5`) ✓

---

### Security Assessment

No critical or high-severity security issues. The security posture is solid:
- Strict 12-digit accountId validation (`/^\d{12}$/`)
- `deliveryDestinationArn` entirely server-derived (not client-controllable)
- `assertDistributionServesSite` prevents cross-org log capture
- `mutationErrorResponse` returns only `error.name` (no ARN/credential leakage)
- Routes registered in `INTERNAL_ROUTES` (not FACS-exposed)
- CloudFormation templates enforce `sts:ExternalId` condition (confused-deputy prevention)
- Only temporary STS credentials used

---

### Verdict

**REQUEST CHANGES** — Finding #1 is a blocking production bug. Both new endpoints pass `undefined` as the STS ExternalId because they destructure it from `validateCloudfrontCredentials` (which doesn't return it) instead of from `gateEdgeOptimizeWizard` (which does). Every assume-role call will fail with AccessDenied in production. The fix is straightforward. Once resolved, this is ready to ship.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 34s | <b>Cost:</b> $7.13 | <b>Commit:</b> <code>6ea0de8949c361435ed72a4a7d4886057aa7f2c6</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @claudiaboldis,

**Verdict: Approve** - prior blocking findings are all addressed; no new blocking issues found.
**Complexity: HIGH** - large diff; API surface + new dependency + OpenAPI contract change.
_Changes: Adds CDN log delivery endpoints (single-distribution and bulk rescan) using STS AssumeRole to create CloudWatch Logs delivery sources in customer accounts, with bounded concurrency, TOCTOU race handling, and cross-org guards (19 files)._
Note: Recommend a human read before merge - this change modifies the OpenAPI contract (`docs/openapi/`). The bot review is a complement to, not a replacement for, a human read here.

<details>

- suggestion: `rescanCdnLogDelivery` has no gateway timeout protection - with concurrency=5 and 1-3 API calls per distribution, accounts with 100+ distributions will exceed the 29s API Gateway timeout; the Lambda finishes but the caller gets a 504 with no summary. Consider a hard cap default for `CDN_LOG_RESCAN_MAX_DISTRIBUTIONS` (e.g. 50) rather than unlimited - `src/controllers/llmo/llmo-cloudfront.js:1310`
- suggestion: Both endpoints call `Organization.findById(site.getOrganizationId())` after `gateEdgeOptimizeWizard` already resolves the org internally - this is a redundant DB round-trip per request; consider exposing `imsOrgId` from the gate return value - `src/controllers/llmo/llmo-cloudfront.js:1183`
- suggestion: `findExistingDelivery` paginates the entire account delivery list per distribution during a rescan (O(N * pages)). A single-fetch local cache of the delivery list per rescan invocation would reduce this to O(pages) with a Map lookup - `src/support/cdn-log-delivery.js:154`
- nit: OpenAPI schema for `log-delivery` declares `deliveryId` as `type: string` but the implementation can return undefined (ConflictException race + eventual consistency); add `nullable: true` for spec-generated client correctness - `docs/openapi/llmo-api.yaml`
- nit: `CDN_LOG_RESCAN_CONCURRENCY` is a hardcoded const (5) while `CDN_LOG_RESCAN_MAX_DISTRIBUTIONS` is env-configurable; making concurrency also env-configurable would allow operational tuning without a code change - `src/controllers/llmo/llmo-cloudfront.js:49`
</details>

- Critical: externalId was always undefined (both endpoints now correctly destructure from gateEdgeOptimizeWizard, verified by test assertions)
- Unbounded concurrency in rescanCdnLogDelivery now capped at CDN_LOG_RESCAN_CONCURRENCY = 5
- Missing server config now correctly returns internalServerError() instead of badRequest()
- Unhandled ConflictException now caught in both PutDeliverySource (DeliverySourceConflict) and CreateDelivery (idempotent fallback)
- Missing integration tests now added (access-control and validation layers covered in test/it/)
- Cross-org log capture guard (assertDistributionServesSite) now applied in enableCdnLogDelivery
- Error message sanitization - mutationErrorResponse now returns only error.name, not the full message

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 8m 30s | <b>Cost:</b> $9.37 | <b>Commit:</b> <code>386aaae9c31e5345c1e4da6786f4c956f35483f5</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2682, opened 2026-06-25
feat(edge-optimize): source CloudFront control-plane from tokowaka-client

ABHA61:
## What
  
  Moves the CloudFront "Optimize at Edge" control-plane out of `spacecat-api-service` and consumes it from the shared `@adobe/spacecat-shared-tokowaka-client` instead. The LLMO CloudFront controller now
  orchestrates onboarding through the client rather than a local copy of the AWS automation.

  ## Why

  The control-plane (assume-role + CloudFront/Lambda@Edge provisioning) is reusable infrastructure that belongs in the shared client alongside the existing CloudFront layer. Keeping it here duplicated AWS-SDK
  logic in the controller and made it harder to reuse. Sourcing it from the client keeps the controller thin and gives a single home for the automation.

  ## What it does

  The `LlmoCloudFront` controller drives the "Deploy routing in CloudFront" wizard end to end, using the shared client:

  - **Connect** — assume the customer's cross-account connector role and confirm it's usable.
  - **Discover** — list the customer's distributions and read a distribution's origins / cache behaviors.
  - **Provision** — add the Edge Optimize origin, create the viewer-request CloudFront Function, update/clone the cache policy, and create the Lambda@Edge origin handler (+ execution role).
  - **Associate & verify** — wire the function and Lambda@Edge onto the chosen behavior, then probe the live domain to confirm routing.
  - **Orchestrate** — a step-on-poll deploy endpoint (`runDeployStep`) and a non-mutating dry-run (`planDeploy`) advance the full sequence and report per-step status to the wizard.

  The Edge Optimize origin headers (`x-forwarded-host`, API key) and the verify-probe domain are resolved server-side per target environment — the customer's production site, or their staging domain when
  onboarding stage.

  ## Changes
  
  - New controller `src/controllers/llmo/llmo-cloudfront.js` + routes under `/sites/:siteId/llmo/cdn-onboard/cloudfront/*` (connect, distributions, origins, behaviors, create-origin / -function / -lambda,
  apply-cache, associate, deploy, plan, verify, bootstrap-url, permissions).
  - Control-plane sourced from `@adobe/spacecat-shared-tokowaka-client` (`CloudFrontEdgeClient`, `assumeConnectorRole`, `verifyRouting`); no local AWS-SDK copy in the controller.
  - Per-environment config via Vault: `SPACECAT_CDN_CLOUDFRONT_TEMPLATE_BUCKET`, `SPACECAT_CDN_CLOUDFRONT_TRUSTED_PRINCIPAL_ARN`, `EDGE_OPTIMIZE_EDGE_DOMAIN`.

  ## Tests

  Controller unit tests for the CloudFront wizard (`llmo-cloudfront`) and LLMO controller (`llmo`); lint, type-check, and build green.


docs/openapi/api.yaml
```diff
@@ -630,6 +630,36 @@ paths:
     $ref: './llmo-api.yaml#/site-llmo-edge-optimize-config'
   /sites/{siteId}/llmo/edge-optimize-config/stage:
     $ref: './llmo-api.yaml#/site-llmo-edge-optimize-config-stage'
+  /sites/{siteId}/llmo/edge-optimize-bootstrap-url:
```
nit23uec: /sites/{siteId}/llmo/onboarding/cloudfront/*
- /bootstrap-url
- /connect
- /distributions, etc

MysticatBot (changes requested):
Hey @ABHA61,

**Verdict: Request changes** - supply-chain risk from the gist dependency and error-handling semantics need addressing before merge.
**Complexity: HIGH** - large diff; API surface + dependency change + cross-account AWS operations.
_Changes: Sources the CloudFront "Optimize at Edge" control-plane from the shared tokowaka client, removes ~1700 LOC of local implementation, and adds 16 new wizard API endpoints (10 files)._

Note: CI checks are passing. The PR is explicitly marked DO NOT MERGE pending adobe/spacecat-shared#1722.

## Must fix before merge

1. [Critical] Unpublished gist dependency is a supply-chain risk - `package.json:100` (details inline)
2. [Important] Controller file grows to ~3200 lines with no separation - `src/controllers/llmo/llmo.js:34` (details inline)
3. [Important] Broad catch-all returns 400 for internal errors that should be 500 - `src/controllers/llmo/llmo.js:2240` (details inline)
4. [Important] No format validation on `lambdaVersionArn` - a caller could associate an arbitrary Lambda ARN with the customer's distribution - `src/controllers/llmo/llmo.js` (applyEdgeOptimizeAssociationsHandler)
5. [Important] `externalId` is not bound to site or session server-side - any LLMO admin who intercepts a valid externalId+accountId pair can operate on that customer's AWS account through any site they control - `src/controllers/llmo/llmo.js` (all wizard endpoints)

<details>

- nit: `js-yaml` is imported at module scope but only used by the permissions endpoint - consider lazy import or isolating it - `src/controllers/llmo/llmo.js:34`
- suggestion: The 16 endpoints share identical validate/gate/assumeRole boilerplate (~400 lines of repetition); a thin middleware or higher-order function would reduce duplication - `src/controllers/llmo/llmo.js:2260`
- nit: Read-only operations (distributions, origins, behaviors) are POST - works due to credentials in body, but document the rationale in the OpenAPI description for API consumers - `docs/openapi/llmo-api.yaml`
- suggestion: `resolveEoTarget` returns `{ error }` (a Response object) on failure - this mixes error-as-value with try/catch in the same flow; consider throwing and letting the outer catch handle uniformly - `src/controllers/llmo/llmo.js:2476`
- nit: AWS error messages passed through `cleanupHeaderValue` can still contain ARNs, account IDs, and resource names - consider a generic message for non-validation errors with details logged server-side only
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 4s | <b>Cost:</b> $8.51 | <b>Commit:</b> <code>0c0e9f973f1fd6661249afb4958753d4e2eeacd5</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/routes/index.js
```diff
@@ -504,6 +504,22 @@ export default function getRouteHandlers(
     'POST /sites/:siteId/llmo/edge-optimize-config': llmoController.createOrUpdateEdgeConfig,
     'GET /sites/:siteId/llmo/edge-optimize-config': llmoController.getEdgeConfig,
     'POST /sites/:siteId/llmo/edge-optimize-config/stage': llmoController.createOrUpdateStageEdgeConfig,
+    'POST /sites/:siteId/llmo/edge-optimize-bootstrap-url': llmoController.getEdgeOptimizeBootstrapUrl,
```
jindaliiita: Shouldn't this be `edge-optimize/bootstrap-url`

src/controllers/llmo/llmo.js
```diff
@@ -2213,7 +2242,961 @@ function LlmoController(ctx) {
...
+  // per-check status rather than failing the whole step on a single problem.
+  const checkEdgeOptimizePrerequisites = async (context) => {
+    const { log, dataAccess, env } = context;
+    const { siteId } = context.params;
+    const { Site } = dataAccess;
+    const accountId = String(context.data?.accountId || '').replace(/\D/g, '');
+    const externalId = String(context.data?.externalId || '').trim();
+    const roleName = env?.EDGE_OPTIMIZE_ROLE_NAME || undefined;
+
+    if (accountId.length !== 12) {
```
jindaliiita: can we make this common ?
as this check is in all the functions

jindaliiita:
I just had a quick look at it from a high level. I still need to review it in detail.

docs/openapi/api.yaml
ABHA61: Done — all 16 wizard routes are now under `/sites/{siteId}/llmo/onboarding/cloudfront/*` (`bootstrap-url`, `connect`, `distributions`, `prerequisites`, `origins`, `behaviors`, `create-origin`, `create-function`, `apply-cache`, `create-lambda`, `lambda-status`, `apply-associations`, `verify`, `deploy`, `plan`, `permissions`). See `src/routes/index.js`.

src/routes/index.js
ABHA61: Reworked — the wizard endpoints were moved under the `/llmo/onboarding/cloudfront/*` namespace (per @nit23uec's suggestion), so this is now `POST /sites/:siteId/llmo/onboarding/cloudfront/bootstrap-url`.

src/controllers/llmo/llmo.js
ABHA61: Done — extracted a `parseEoCredentials(context, { requireDistribution })` helper that centralises the `accountId`/`externalId`/`distributionId` validation; all wizard handlers now call it instead of repeating the checks.

ABHA61:
Status update on the review feedback (code at `5c398f46`):

**Fixed**
- **Route namespace** — all 16 wizard endpoints moved under `/sites/{siteId}/llmo/onboarding/cloudfront/*`.
- **Internal errors → 500** — wizard handlers' catch blocks now return `internalServerError` (generic client message, full error logged server-side); up-front validation still returns `400`.
- **`lambdaVersionArn` validation** — enforced with `^arn:aws:lambda:us-east-1:\d{12}:function:[A-Za-z0-9_-]+:\d+$` and the ARN account segment must equal the request `accountId`.
- **Duplicated credential checks** — centralised into `parseEoCredentials(...)`.
- **POST-for-reads rationale** — documented in the OpenAPI descriptions ("read-only").
- **Error sanitisation** — unexpected errors return a generic message; raw AWS details (ARNs/account ids) are logged server-side only.

**Not required (by design)**
- **`externalId` not bound to site/session** — confused-deputy protection is provided by a per-session, randomly generated `externalId` created at connector-role setup and reused for the assume; a new session generates a new id, so there is no static secret to bind. Can add an explicit server-side site↔externalId record if you'd prefer defence-in-depth.
- **`resolveEoTarget` error-as-value** — kept intentionally; the `{ error }` Response is returned immediately by the caller, so it doesn't actually interleave with the try/catch flow.

**Pending before merge (intentional)**
- **Gist dependency** — temporary dev bridge while adobe/spacecat-shared#1722 is in review (this PR is marked DO-NOT-MERGE). Will be re-pinned to the published `@adobe/spacecat-shared-tokowaka-client` version once #1722 merges.

**Deferred (optional follow-up)**
- **Split the ~3200-line `llmo.js`** — planned as a follow-up to extract the CloudFront wizard into `src/controllers/llmo/llmo-cloudfront.js` (parallel to `llmo-cloudflare.js`); kept out of this PR to limit churn.
- **Full HOF/middleware** for the gate/validate/assume boilerplate — partially done via `parseEoCredentials`; a thin wrapper can follow.
- **`js-yaml` lazy import** — minor; can switch the permissions endpoint to a dynamic import.

src/controllers/llmo/llmo-cloudfront.js
```diff
@@ -0,0 +1,968 @@
...
+        return ok({ connected: true, accountId, roleArn });
+      } catch (assumeError) {
+        // The role may not exist yet (customer still creating it) or the external ID may not
+        // match — surface as not-connected so the wizard can keep polling rather than erroring.
+        log.info(`[edge-optimize-connect] Role not yet assumable for site ${siteId}: ${assumeError.message}`);
+        return ok({ connected: false, reason: cleanupHeaderValue(assumeError.message) });
+      }
+    } catch (error) {
+      log.error(`Failed to connect edge optimize role for site ${siteId}:`, error);
+      return badRequest(cleanupHeaderValue(error.message));
```
MysticatBot: issue (blocking): The outer catch in `connect` returns `badRequest(cleanupHeaderValue(error.message))` while every other handler returns `internalServerError(...)` with a generic message. This leaks internal error details (DynamoDB table names, connection info) and returns 400 (implying client fault) for server-side failures.

This was supposedly addressed from the prior review (finding: broad catch-all returns 400) but this one handler was missed.

**Fix:**
```javascript
} catch (error) {
  log.error(`Failed to connect edge optimize role for site ${siteId}:`, error);
  return internalServerError('Failed to connect the edge optimize role, please try again');
}
```

src/controllers/llmo/llmo-cloudfront.js
```diff
@@ -0,0 +1,968 @@
...
+
+  // Create/update + publish the `edgeoptimize-routing` CloudFront Function (mutation, idempotent).
+  // Needs the default-behavior target origin id so the function's failover origin group is correct.
+  const createRoutingFunction = async (context) => {
+    const { log, dataAccess, env } = context;
+    const { siteId } = context.params;
+    const { Site } = dataAccess;
+    const targetedPaths = Array.isArray(context.data?.targetedPaths)
+      ? context.data.targetedPaths
+      : null;
```
MysticatBot, line 570: issue (blocking): `targetedPaths` is validated only as "is it an array or null" and then forwarded directly to `createEdgeOptimizeRoutingFunction`. If the downstream function interpolates these values into CloudFront Function code (JavaScript executed at the CDN edge), a caller could inject arbitrary code.

Even if the shared client validates, defense in depth is warranted for inputs that flow into code generation.

**Fix:** Add boundary validation before passing downstream:
```javascript
if (targetedPaths !== null) {
  if (targetedPaths.length > 50) {
    return badRequest('targetedPaths must not exceed 50 entries');
  }
  if (!targetedPaths.every((p) => typeof p === 'string' && p.length > 0 && p.length <= 256
    && /^[a-zA-Z0-9/_.*-]+$/.test(p))) {
    return badRequest('Each targetedPaths entry must be a non-empty path pattern (max 256 chars, alphanumeric/slashes/wildcards only)');
  }
}
```

MysticatBot (changes requested):
Hey @ABHA61,

**Verdict: Request changes** - error handling gap missed from last round, plus unsanitized code-generation input needs defense-in-depth.
**Complexity: HIGH** - large diff; API surface + dependency change + cross-account AWS operations.
_Changes: Extracts the CloudFront "Optimize at Edge" control-plane into a dedicated controller sourced from the shared tokowaka client, adding 16 wizard endpoints (14 files)._
Note: Recommend a human read before merge - this change adds a new API surface (docs/openapi/llmo-api.yaml) with cross-account AWS operations. The bot review is a complement to, not a replacement for, a human read here.

Note: CI build is pending. The PR is explicitly marked DO NOT MERGE pending adobe/spacecat-shared#1722 and the gist dependency replacement.

## Must fix before merge

1. [Important] `connect` outer catch returns 400 with raw error message (inconsistent with all other handlers) - `src/controllers/llmo/llmo-cloudfront.js:226` (details inline)
2. [Important] `targetedPaths` passed unsanitized to CloudFront Function code-generation - `src/controllers/llmo/llmo-cloudfront.js:497` (details inline)

<details>

- nit: `getPermissions` inner catch returns `badRequest` for an S3 read failure (server-side problem, should be `internalServerError`) - `src/controllers/llmo/llmo-cloudfront.js:770`
- nit: No format validation on `distributionId` - CloudFront IDs match `^[A-Z0-9]{12,14}$`; adding a regex in `parseEoCredentials` would reject garbage early - `src/controllers/llmo/llmo-cloudfront.js:181`
- suggestion: Add a test case that passes an actual `targetedPaths` array to `createRoutingFunction` and verifies it is forwarded to the downstream function (currently only the null case is tested) - `test/controllers/llmo/llmo-cloudfront.test.js:1001`
</details>

- Controller extracted into dedicated `llmo-cloudfront.js` (968 lines vs prior 3200-line monolith)
- Catch blocks return `internalServerError` with generic messages (except `connect`, see finding 1)
- `lambdaVersionArn` validated with regex + account-ID cross-check
- Routes moved under `/sites/{siteId}/llmo/cdn-onboard/cloudfront/*` namespace
- Credential checks centralized in `parseEoCredentials`

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 12s | <b>Cost:</b> $7.34 | <b>Commit:</b> <code>e29e782e3a56b26674e3452d68164da785e11024</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/controllers/llmo/llmo-cloudfront.js
ABHA61: Agree this is useful defense-in-depth, but we are not taking it as a current release blocker. The FE wizard does not expose arbitrary targeted paths today, and the shared code serializes the array into generated Function code rather than concatenating raw path text. I added this to the CloudFront punch list as a P1 follow-up: add controller-level targetedPaths limits before passing downstream (max entries, max length, accepted path characters).

src/routes/index.js
ABHA61: Follow-up: this was aligned with the Cloudflare route convention. The final endpoint is now `POST /sites/:siteId/llmo/cdn-onboard/cloudfront/bootstrap-url`, with the rest of the CloudFront wizard under the same `/llmo/cdn-onboard/cloudfront/*` namespace.

ABHA61:
**Review feedback update (latest pushed code `9b974316`)**

**Fixed**
- Route namespace: all CloudFront wizard routes and OpenAPI paths are now under `/sites/{siteId}/llmo/cdn-onboard/cloudfront/*`, aligned with the existing Cloudflare convention.
- `connect` outer catch: now logs the real error server-side and returns `internalServerError("Failed to connect the edge optimize role, please try again")` instead of returning `400` with the raw error message.
- Shared-client refresh: api-service now consumes the rebuilt shared tarball and imports the AWS-style shared aliases.

**Deferred / asking for next release**
- rest pending items seems p1 to me,  Can we take this in the next release .

Focused verification after rebase: `llmo-cloudfront.test.js` = 165 passing; eslint on the CloudFront controller/test files passed.

src/controllers/llmo/llmo-cloudfront.js
```diff
@@ -0,0 +1,984 @@
...
+
+    try {
+      const { error } = await gateEdgeOptimizeWizard(siteId, Site, 'apply edge optimize cache headers');
+      if (error) {
+        return error;
+      }
+
+      const { cloudFrontClient } = await assumeCloudFrontClient({
+        accountId, externalId, roleName,
+      });
```
nit23uec, line 663: The distribution id that is getting updated must have a domain that matches the site's base url

src/controllers/llmo/llmo-cloudfront.js
```diff
@@ -0,0 +1,984 @@
...
+  //
+  // - 'production' / absent → today's behavior: prod site baseURL → first metaconfig apiKey +
+  //   calculateForwardedHost(prod baseURL).
+  // - 'stage' → resolve the single stage domain persisted on the prod site's edgeOptimizeConfig
+  //   (stagingDomains[0]); compose its baseURL; look up the (already-onboarded) stage site; use
+  //   that stage site's first metaconfig apiKey + calculateForwardedHost(stage baseURL).
+  //
+  // Returns `{ target: { baseURL, apiKey, forwardedHost }, error }`. On any resolution failure
+  // `error` is a badRequest Response the caller returns directly; otherwise `error` is undefined.
+  const resolveEoTarget = async (context, site, environment, log) => {
```
nit23uec: FE has to make sure that it makes the call to the correct siteId -> in stage case it would be the siteid of the stage domain. - environment distinction is not required on server - it will behave similar for stage or prod onboarding.

src/controllers/llmo/llmo-cloudfront.js
```diff
@@ -0,0 +1,984 @@
...
+    try {
+      const { error } = await gateEdgeOptimizeWizard(siteId, Site, 'list CloudFront distributions');
+      if (error) {
+        return error;
+      }
+
+      const { cloudFrontClient } = await assumeCloudFrontClient({
+        accountId, externalId, roleName,
+      });
+      const distributions = await cloudFrontClient.listDistributions();
```
nit23uec, line 343: can we list only those distributions whose domain matches the host of the llmo site?

src/controllers/llmo/llmo-cloudfront.js
```diff
@@ -0,0 +1,984 @@
...
+    return { site };
+  };
+
+  // Shared input validation for the CloudFront wizard endpoints that act through the
+  // cross-account connector role. Parses + validates the caller-supplied AWS account id and
+  // per-session external id (and optionally the CloudFront distribution id). Returns
+  // `{ accountId, externalId, distributionId, error }` where `error` is a badRequest Response
+  // when validation fails (undefined otherwise) — keeping the messages/status identical to the
+  // previously inlined checks.
+  const parseEoCredentials = (context, { requireDistribution = false } = {}) => {
```
nit23uec: validateCloudfrontCredentials
```suggestion
  const validateCloudfrontCredentials = (context, { requireDistribution = false } = {}) => {
```

src/controllers/llmo/llmo-cloudfront.js
ABHA61: @nit23uec  Currently, there is no separate wizard or button on the frontend for Stage environment domain onboarding. If a customer has a Stage domain, it is displayed when they are selecting the distribution.
So, at the moment, there is no dedicated Stage button. If we want to support this change, I think we would need to introduce a separate option/button specifically for Stage environment. Should i make this change as per your suggestion or should we ask Domain of stage vs prod at the start.

image

We also show to add stage domain account id as info here 
image

nit23uec:
# Review — prod-safety, attribution & auditability

This control plane mutates a **customer's production CloudFront** cross-account (assume-role → `UpdateDistribution`, CloudFront Function publish, Lambda@Edge + IAM role). The idempotency, ETag concurrency, and "merge-don't-clobber" association handling in `spacecat-shared-tokowaka-client` are solid and I'm not re-litigating them. My asks are about **making sure a mutation lands on the right object** and **being able to prove who did what afterwards** — the things that matter when the blast radius is someone else's prod CDN. None of these are "unauthorized caller" issues (the LLMO authz gate + the customer's role trust policy cover that); they're authorized-operator-error and forensic gaps, which authz can't address by construction.

## Already good (so we don't re-review it)
- Idempotency via ETag + existence gates across all mutators; `runDeployStep` gates each step so re-polls don't re-deploy.
- `applyAssociations` **merges** and **refuses** to overwrite a customer's existing viewer-request function / origin-request Lambda — the most important "don't break customer config" property, correctly done.
- `planDeploy` discloses cache/MinTTL changes before commit; `deploy` returns partial step status on failure.
- EO API key and STS credentials are never logged; STS creds are correctly ephemeral.

---

## 1. MUST FIX — no guardrail that the target distribution belongs to the site

`distributionId` is taken from the request body and passed straight to every mutating call (`createOrigin`, `createRoutingFunction`, `applyCache`, `applyAssociations`, `deploy`, `plan`) with **no check that the distribution actually serves this site's domain**. `site.getBaseURL()` is used only to derive injected origin headers (`resolveEoTarget`, `llmo-cloudfront.js:431,440`) and the post-deploy probe host (`verifyRouting:730`) — neither blocks a mutation.

**Failure mode (realistic — operators manage many customers):** an authorized admin picks the wrong distribution `D′` (a *different* customer's). origin → function → cache → lambda → associate all apply to `D′`'s prod config. `verify` then probes *this* site's host (served by the correct, untouched distribution), so it sits at `in_progress` forever. The wrong customer's prod is mutated; verify never blocks it and never rolls it back.

**Fix (cheap, no persistence, no new AWS calls):** `planDeploy` already reads the full distribution config — compare `config.Aliases.Items` (and/or the distribution `DomainName`) against the `site.getBaseURL()` host there and set `canProceed:false` / `blocker` on mismatch. Mirror the assertion at the top of `runDeployStep` so the granular and orchestrated paths can't diverge. Permit an explicit, logged override only if a real onboarding case needs it.

---

## 2. MUST FIX — mutations are not attributable in the customer's CloudTrail

`assumeConnectorRole` hard-codes the session name (`tokowaka-client/cloudfront/index.js:57,138`):

```js
const SESSION_NAME = 'llmo-edge-optimize';
// ...
RoleSessionName: SESSION_NAME,
```

Every change Adobe makes in the **customer's** account appears in **their** CloudTrail as the same `assumed-role/AdobeLLMOptimizerCloudFrontConnectorRole/llmo-edge-optimize` — no operator, no site. We already hold the acting identity (`context.authInfo.getProfile().email`, an IMS GUID whose characters are all valid in an STS `RoleSessionName`); it just isn't propagated. Carrying it as the session name puts it inline on **every** event with no trust-policy change and no session-tag permission. Three coupled changes:

**2a. Client — `assumeConnectorRole` accepts an optional `roleSessionName`** (default-preserving; sanitize/clamp inside the client so a bad value can't break the assume):

```js
// STS RoleSessionName: 2–64 chars, charset [\w+=,.@-]. Sanitize + clamp so a
// caller-supplied actor id can never break AssumeRole; fall back to the constant.
function toRoleSessionName(raw) {
  const cleaned = String(raw || '').replace(/[^\w+=,.@-]/g, '_').slice(0, 64);
  return cleaned.length >= 2 ? cleaned : SESSION_NAME;
}

export async function assumeConnectorRole({
  accountId, externalId,
  roleName = EDGE_OPTIMIZE_DEFAULT_ROLE_NAME,
  region = EDGE_OPTIMIZE_REGION,
  roleSessionName,                                    // NEW — optional
}) {
  // ...existing validation...
  const response = await sts.send(new AssumeRoleCommand({
    RoleArn: roleArn,
    RoleSessionName: toRoleSessionName(roleSessionName),  // was SESSION_NAME
    ExternalId: externalId,
    DurationSeconds: SESSION_DURATION_SECONDS,
  }));
  // ...unchanged...
}
```

**2b. Controller — derive the actor once and thread it through both assume sites** (`llmo-cloudfront.js`):

```js
function LlmoCloudFrontController(ctx) {
  const accessControlUtil = AccessControlUtil.fromContext(ctx);

  // Per-request acting identity → CloudFront session name so the customer's CloudTrail can
  // attribute every mutation to the operator. Client sanitizes/clamps to STS rules; use the id
  // portion (before '@') to guarantee headroom under the 64-char limit.
  const actor = (() => {
    try {
      const profile = ctx.authInfo?.getProfile?.();
      return profile?.email || profile?.userId || 'unknown';
    } catch { return 'unknown'; }
  })();
  const roleSessionName = `eo-${String(actor).split('@')[0]}`;
```

Shared helper (covers `listDistributions`, `checkPrerequisites`, `fetchOrigins`, `fetchBehaviors`, `createOrigin`, `createRoutingFunction`, `applyCache`, `createLambda`, `fetchLambdaStatus`, `applyAssociations`, `verifyRouting`, `deploy`, `plan` in one place):

```js
  const assumeCloudFrontClient = async ({ accountId, externalId, roleName }) => {
    const assumed = await assumeConnectorRole({
      accountId, externalId, roleName, roleSessionName,        // NEW
    });
    return {
      ...assumed,
      cloudFrontClient: new CloudFrontEdgeClient({ credentials: assumed.credentials }),
    };
  };
```

...and the direct call in `connect` (`:220`):

```js
  const { roleArn } = await assumeConnectorRole({
    accountId, externalId, roleName, roleSessionName,          // NEW
  });
```

**2c. Adobe-side log — stamp the same `actor`** into every mutation log line (see §3), so we can answer "who/when/what" from our side without reading the customer's account.

> **Why session name, not session tags:** for a single identifier this is strictly better — it shows on **every** downstream event (session tags appear only on the `AssumeRole` event, forcing an `accessKeyId` join) and needs **no** `sts:TagSession` grant in the customer's trust policy (no CloudFormation redeploy). Session tags would only be warranted later for multiple structured attributes or trust-policy enforcement.

---

## 3. MUST FIX — Adobe-side logs omit the actor and aren't a durable audit trail

The mutation logs (`[edge-optimize-origin] … site ${siteId}, distribution ${distributionId}` `:495`; `[edge-optimize-associate]` `:694`; `[edge-optimize-deploy]` `:820`) have three gaps:

- **No acting identity** — available via `authInfo.getProfile().email`; never logged.
- **Inconsistent object fields** — `connect` logs `accountId` (`:221`); the mutators log `siteId`+`distributionId` but **not** `accountId`; `deploy` logs neither.
- **Success-only, ephemeral** — emitted after the await, to CloudWatch only. No "attempting" line (a mid-call crash leaves no record of what was tried), and no queryable record of what changed (which behavior, which managed cache policy was cloned, before/after).

**Fix:**
1. **Standardized structured line** on every mutating endpoint, emitted *before* the AWS call (intent) and *after* (result): `{ actor, siteId, accountId, distributionId, behavior, action, result, requestId, ts }`.
2. **Persist an audit record** per mutation (DB row or appended to the site's edge-optimize config). For a tool changing customer production infra, an Adobe-side queryable "who/when/what/account/distribution/outcome" trail shouldn't depend on CloudWatch retention.
3. **Categorize granular-endpoint errors.** `deploy` surfaces per-step `detail` (good), but the granular endpoints collapse all AWS errors to `internalServerError('… please try again')` (e.g. `:498-499`), hiding `AccessDenied` / `PreconditionFailed` / `Throttling` and inviting a blind retry of a partially-applied prod mutation. Surface a cleaned, categorized error like `connect`/`checkPrerequisites` already do.

---

## Tests to add
- **Client:** `assumeConnectorRole` forwards a provided `roleSessionName` to `AssumeRoleCommand`; defaults to `SESSION_NAME` when omitted; sanitizes illegal chars; clamps `>64`; floors `<2`-char to default.
- **Controller:** the assume call receives `roleSessionName` derived from `authInfo`; resolves to `eo-unknown` when the profile is missing (no throw). Mutation handlers reject when the distribution domain doesn't match the site (§1).

## Summary

| # | Severity | Theme | Ask |
|---|----------|-------|-----|
| 1 | Must | Prod safety | Block deploy when distribution `Aliases`/domain ≠ site base URL (in `planDeploy` + `runDeployStep`) |
| 2 | Must | Attribution | `roleSessionName` param (2a) + thread actor through call sites (2b) + Adobe-side actor log (2c) |
| 3 | Must | Audit/logging | Structured intent+result line, durable audit record, categorized granular-endpoint errors |

#1 and the client half of #2 (2a) require changes in `spacecat-shared-tokowaka-client`, not just this repo. Caveats: the session-name value is an opaque IMS GUID (low-PII) landing in the customer's CloudTrail — call it out as a conscious decision; use the id portion before `@` to stay under the 64-char limit.

nit23uec:
I'd suggest to fix atleast 1 and 3 above before merging this and then do 2 in next iteration after AMS validation (pls log a ticket for that).
Also, the stage flow (this code should be environment agnostic) needs to be fixed before merge.

ABHA61:
> 1. MUST FIX — no guardrail that the target distribution belongs to the site
1. We have guardrail of external id , that A customer cannot access customer B account if they have account id and Distribution id. 
2. If wrong Distribution be choosen from FE , we show warning for now that it is not attached to same domain as per site id.
3. Customer can still corrupt there own distribution by giving wrong distribution if they will directly call our API.

nit23uec:
Thanks for the quick turnaround — but I think points 1–3 are addressing a different threat than the one I raised, so let me re-anchor.

My (1) is **not** an unauthorized-caller / cross-account concern (I called that out as already covered by the role trust policy + LLMO authz). It's **authorized-operator error**: a legitimate caller pointing a real mutation at the wrong distribution.

**1. `externalId` doesn't cover this.** `externalId` is an account-granular confused-deputy control — it stops us assuming into a *different customer's account*. It does nothing once we're correctly inside the right account and the operator passes the wrong `distributionId`. That happens two realistic ways: (a) the same customer owns several distributions and the wrong one is selected, or (b) a multi-tenant onboarding console sends a mismatched `accountId`/`externalId`/`distributionId` tuple. In both cases the assume succeeds and we mutate the wrong prod distribution. So this guardrail is orthogonal to the finding.

**2. An FE warning isn't a server-side guardrail.** This is exactly the gap, not the mitigation. The API mutates a customer's **production** CloudFront; a UI warning is ignorable, and it's bypassed entirely by any direct API call. The check has to live where the mutation is decided.

**3. Point 3 actually confirms the hole** — "a wrong distribution via direct API" is the failure mode I'm describing. And it's worse than "their own distribution": `verify` then probes the *correct* host (served by the untouched distribution), so it sits at `in_progress` forever while the wrong prod config was already changed — no block, no rollback signal.

The fix is essentially free and server-side: `planDeploy` already reads the full distribution config, so compare `config.Aliases.Items` (and/or `DomainName`) against the `site.getBaseURL()` host right there and set `canProceed:false` / `blocker` on mismatch. Mirror the same assertion at the top of `runDeployStep` so the granular and orchestrated paths can't diverge. No new AWS call, no persistence. If a genuine onboarding case ever needs to override, allow it explicitly and log it.

ABHA61:
**Update — addressed @nit23uec's review (pushed in `b85707f1`):**

**Distribution ↔ site guardrail (must-fix #1 / `:634` / `:326`)** — now a **hard block in the BE**: `assertDistributionServesSite` requires the site host to be among the chosen distribution's `Aliases` (CNAMEs) before any mutation — `deploy` hard-blocks, `plan` returns `canProceed:false`/`blocker`, and the granular mutators block too. It blocks on a CNAME mismatch **and** when the distribution has no CNAME, with a logged `allowDomainMismatch` override for edge cases. On the **FE** we also show a non-blocking **warning** plus an **ⓘ tooltip** on each distribution showing the domain(s) it serves, so the customer picks the correct distribution up front. (Server-side *filtering* of the list — vs. block-at-mutation — is tracked as a follow-up.)

**Environment-agnostic / stage flow (`:392`)** — removed the server-side `environment` flag; `resolveEoTarget` resolves against whatever site it's given. The wizard is environment-agnostic: a stage domain is its own onboarded LLMO site, and the FE runs the wizard against that site's `siteId` (the domain is chosen up front on **Allow access**, so all calls are scoped to the right site). OpenAPI + tests updated.

**Actor logging + error handling (must-fix #3)** — every mutation now emits a structured before/after audit line `{ actor, siteId, accountId, distributionId, behavior, action, result, requestId, ts }` (Splunk-queryable), and granular AWS errors are categorized (`AccessDenied`/`PreconditionFailed`/`Throttling` → 400) instead of a blanket 500 + "try again". The `connect` 400→500 fix is in too.

**Naming (`:169` + convention)** — `parseEoCredentials` → `validateCloudfrontCredentials` (+ `distributionId` format check); routes under `/sites/{siteId}/llmo/cdn-onboard/cloudfront/*`. Aligned to the Cloudflare sibling: "Edge Optimize" only on the EO origin, "CloudFront" for the connector plumbing, "LLMO API key" for the credential, single `[cdn-onboard-cloudfront]` log tag.

**Parked → [LLMO-5909](https://jira.corp.adobe.com/browse/LLMO-5909):** must-fix #2 (CloudTrail `RoleSessionName` = operator identity, after AMS validation) and a separately *persisted* audit record (the structured logs land in Splunk today).

package.json
```diff
@@ -98,14 +98,18 @@
     "@adobe/spacecat-shared-scrape-client": "2.6.3",
     "@adobe/spacecat-shared-slack-client": "1.6.7",
     "@adobe/spacecat-shared-tier-client": "1.5.1",
-    "@adobe/spacecat-shared-tokowaka-client": "1.19.0",
+    "@adobe/spacecat-shared-tokowaka-client": "^1.20.0",
```
nit23uec: better to pin the version

nit23uec (approved):
pls pin the version of tokowaka-client. otherwise, looks good.

MysticatBot (changes requested):
Hey @ABHA61,

**Verdict: Request changes** - input validation gap on targetedPaths needs defense-in-depth before this hits production edge infrastructure.
**Complexity: HIGH** - large diff; API surface + dependency change + cross-account AWS operations.
_Changes: Extracts the CloudFront "Optimize at Edge" control-plane into a dedicated controller sourced from the shared tokowaka client, adding 16 wizard endpoints (14 files)._
Note: Recommend a human read before merge - this change adds a new API surface (docs/openapi/llmo-api.yaml) with cross-account AWS operations. The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge

1. [Important] `targetedPaths` forwarded to CF Function code-generation without element-level validation - `src/controllers/llmo/llmo-cloudfront.js:363`

   The array passes only `Array.isArray()` before being forwarded to `cloudFrontClient.createCloudFrontFunction()` which interpolates the values into CloudFront Function source code (confirmed in the shared client: `JSON.stringify(targetedPaths)` into a `var TARGETED_PATHS = ...` assignment). While `JSON.stringify` prevents classic injection, defense-in-depth at the API boundary should validate each element is a well-formed URL path:
   ```javascript
   if (targetedPaths) {
     if (targetedPaths.length > 50) return badRequest("targetedPaths exceeds maximum of 50 entries");
     const invalid = targetedPaths.find((p) => typeof p !== "string" || !/^\/[a-zA-Z0-9\/_\-.*]+$/.test(p) || p.length > 500);
     if (invalid !== undefined) return badRequest("Each targetedPaths entry must be a valid URL path");
   }
   ```
   This rejects non-string elements, excessively long paths, and garbage input that would waste CloudFront Function size budget (10KB max).

2. [Important] `getPermissions` inner catch returns 400 for a server-side S3 failure - `src/controllers/llmo/llmo-cloudfront.js:905`

   The inner `catch (s3Error)` returns `badRequest('CloudFront connector permissions are not available')` but S3 read failure is a server-side problem. This misleads client retry logic (clients do NOT retry 400s), suppresses it from server-error alerting, and violates HTTP semantics. Every other catch in this file correctly uses `internalServerError` for server failures. Fix: `return internalServerError('CloudFront connector permissions are not available');`

<details>

- suggestion: AWS error messages surfaced via `mutationErrorResponse` can contain ARNs, account IDs, and resource names - return only the `error.name` to the caller and log the full message server-side - `src/controllers/llmo/llmo-cloudfront.js:229`
- nit: `verifyRouting` builds a URL from user-controlled `domain` without hostname format validation (low risk given LLMO admin gate, but a regex rejecting path segments and RFC-1918 ranges would close the surface cheaply) - `src/controllers/llmo/llmo-cloudfront.js:843`
- suggestion: Add a test case that passes an actual `targetedPaths` array to `createRoutingFunction` and asserts it is forwarded to the stub (currently only the null case is tested) - `test/controllers/llmo/llmo-cloudfront.test.js`
- nit: Presigned URL TTL is env-configurable with no upper bound - consider `Math.min(presignTtlSeconds, 3600)` as a guardrail against misconfiguration - `src/controllers/llmo/llmo-cloudfront.js:100`

</details>

- Controller extracted into dedicated llmo-cloudfront.js (well-structured at 1138 lines)
- Catch blocks return internalServerError with generic messages (except getPermissions, see finding 2)
- lambdaVersionArn validated with regex + account-ID cross-check
- Routes namespaced under /sites/{siteId}/llmo/cdn-onboard/cloudfront/*
- Credential checks centralized in validateCloudfrontCredentials with distributionId regex
- Gist dependency replaced with proper npm package


---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 54s | <b>Cost:</b> $8.40 | <b>Commit:</b> <code>ff73936a1d15804250346637fd608b13e4182973</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

package.json
ABHA61: Done — pinned to `1.20.0` (dropped the caret) in `e21bb091`, matching the exact-pin convention of the other `@adobe/spacecat-shared-*` deps. Lock updated to match.

src/controllers/llmo/llmo-cloudfront.js
ABHA61: Addressed in `098a6789` — `createRoutingFunction` now validates `targetedPaths` before passing it downstream: a count cap (max 20 entries) plus a per-entry check (non-empty string, ≤256 chars, path charset `^[a-zA-Z0-9/_.*-]+$`).

On the injection concern specifically: the shared client does not interpolate these into the Function source — `buildCloudfrontFunctionCode` embeds them with `JSON.stringify` (`var TARGETED_PATHS = ${JSON.stringify(targetedPaths)}`), so any string is emitted as escaped JSON data, not executable code. So this validation is sanity + size guard (defense-in-depth), not the injection fix.

On the count: `targetedPaths` is an optional allowlist used mainly for testing (route only specific paths instead of all HTML), so it is inherently a short list. A count cap keeps the generated Function well under the CloudFront 10 KB code limit with headroom for the base function to grow — I avoided a fixed byte budget for exactly that reason. Tests cover the valid-array, malformed-entry, and too-many cases.

ABHA61:
Re: must-fix #1 (`targetedPaths` element validation) — this is already addressed on the latest branch. This review ran on `ff73936a`, but the validation landed in `098a6789` (pushed after): `createRoutingFunction` now caps the list (max 20 entries) and validates each entry (non-empty string, ≤256 chars, path charset `^[a-zA-Z0-9/_.*-]+$`) before forwarding downstream, with unit tests for the valid-array, malformed-entry, and too-many cases.

On the count: I used 20 rather than 50 because `targetedPaths` is an optional allowlist used mainly for testing (route specific paths instead of all HTML pages), so it's inherently a short list — and a small count keeps the generated CloudFront Function well under the 10 KB code limit *with headroom for the base function to grow* (I avoided a fixed byte budget for that reason). As you noted, `JSON.stringify` already prevents injection, so this is defense-in-depth + size guard. Full rationale in the thread above (r3486501597).

MysticatBot (approved):
Hey @ABHA61,

**Verdict: Approve** - all prior findings addressed cleanly; incremental changes are well-tested and correctly scoped.
**Complexity: HIGH** - large diff; API surface + dependency change + cross-account AWS operations.
_Changes: Pins tokowaka-client to exact 1.20.0, adds element-level validation on targetedPaths (count cap, length cap, charset allowlist), and fixes getPermissions to return 500 for server-side S3 failures (14 files)._

<details>

- nit: `getPermissions` now returns 500 for both "S3 broke" and "template exists but has no permissions metadata" - if callers ever need to distinguish template-is-malformed from infrastructure-failure, a separate status or error code would help. Low priority given both were previously 400. - `src/controllers/llmo/llmo-cloudfront.js:1104`
- nit: The charset regex allows `*` which has semantic meaning in CF Function routing. A `targetedPaths` entry like `/*` would match all traffic rather than a targeted subset. Acceptable given the LLMO admin gate, but consider documenting or rejecting pure-wildcard entries if the intent is truly "targeted" paths. - `src/controllers/llmo/llmo-cloudfront.js:583`
</details>

- targetedPaths validated with count cap (20), length cap (256), and charset allowlist before forwarding to CF Function code-generation
- getPermissions returns internalServerError (500) for server-side S3/YAML failures instead of badRequest (400)
- tokowaka-client pinned to exact 1.20.0 (matching the convention of other @adobe/spacecat-shared-* deps)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 32s | <b>Cost:</b> $5.81 | <b>Commit:</b> <code>5ec370587242137224ba9542b20ab67739af35d8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2700, opened 2026-06-26
feat(brands): allow pending brands to unset/repoint primary URL (LLMO-5870)

igor-grubic:
## 1. Abstract
Makes `updateBrand`'s `baseSiteId` (primary URL) handling status-aware: a **pending** brand may now clear or re-point its primary URL, while **active** brands keep the strict immutable-once-set guard. Backend half of LLMO-5870.

## 2. Reasoning
Split out from LLMO-5607. `updateBrand` treated `baseSiteId` as immutable-once-set for **all** brands — active or pending — silently dropping both an unset (`baseSiteId: null`) and a value→value re-point. That over-applied the active-brand safety guard (the LLMO-5556 / express.adobe.com regression guard) to pending brands, so a pending draft could never free or change its primary URL and the site stayed reserved forever — the endpoint returns `409 "This site is already the primary URL for another brand"`. Customers had no way to free a reserved URL.

- **`updateBrand` (`src/support/brands-storage.js`) is now status-aware:**
  - **Pending** brand: may **clear** its primary site (`baseSiteId: null` → `site_id = NULL`, freeing the site for reuse) and **re-point** to a different site.
  - **Active** brand: unchanged — strict immutable-once-set; a routine field save echoing a stale `baseSiteId` can never strip or re-point a live anchor (LLMO-5556 / express.adobe.com regression guard stays intact).
- Broadened the existing-row fetch so `existing.status` is loaded even when `baseSiteId === null`.
- Promote-to-active guard now treats a same-PATCH clear as site-less, returning the typed 400 instead of leaning on the stale `site_id`.
- **`docs/openapi/schemas.yaml`** — `Brand.baseSiteId` widened to `type: [string, 'null']` with updated semantics; generated docs regenerated.
- **DB-level safety:** the `brands_base_site_unique` partial unique index skips NULLs and `chk_active_brand_has_site_id` only constrains active brands, so clearing a pending brand's `site_id` is safe; the unique index still rejects a re-point that collides with another brand's primary URL (409 preserved).

## 4. Required information
- Jira / issue: LLMO-5870 — https://jira.corp.adobe.com/browse/LLMO-5870
- Other: parent story LLMO-5607 — https://jira.corp.adobe.com/browse/LLMO-5607

- **project-elmo-ui** — *coordinated / consumed*: the UI half (https://github.com/adobe/project-elmo-ui/pull/2180, same `LLMO-5870` branch) sends `baseSiteId: null` to unset and re-points pending brands through this endpoint. Must ship after this change.

## 7. Test plan
(a) **Local** — behaviour exercised by unit, controller, and integration suites:
- Unit (`test/support/brands-storage.test.js`): pending clear → `site_id` NULL; active clear → no `site_id` in patch; pending re-point → updates; pending re-point onto a taken site → 409; existing "immutable when already set" case preserved.
- Controller (`test/controllers/brands.test.js`): `baseSiteId: null` is forwarded to storage, not stripped.
- Integration (`test/it/shared/tests/brands.js`): set a pending brand's primary URL, confirm a second brand is blocked, clear the first, confirm the second can reuse the freed site.
- The Postgres IT suite and the `--test-bundle` build run in CI rather than locally (ECR Docker auth / deploy-env not available locally).

(b) **Per-env (dev / stage):** set a pending brand's primary URL → confirm a second brand gets `409`; PATCH the first with `baseSiteId: null` (unset) → confirm the site frees and the second brand can claim it; PATCH a pending brand to re-point to a different site → confirm it updates; confirm an **active** brand still cannot strip or re-point its primary URL.

Coordinated cross-repo pair (same `LLMO-5870` branch). The backend must lead so the UI's unset/re-point is honored.
- **This PR** (spacecat-api-service) — *blocks*: enables clearing / re-pointing a pending brand's `baseSiteId`. Merge + deploy first.
- **project-elmo-ui** (UI half) — *depends-on*: sends `baseSiteId: null` and re-points via this endpoint; if it ships first, the unset is silently dropped by the old backend. https://github.com/adobe/project-elmo-ui/pull/2180

Ordered sequence: merge + deploy this PR → merge/ship the project-elmo-ui UI PR.


MysticatBot (approved):
Hey @igor-grubic,

**Verdict: Approve** - well-scoped, correctly gated, DB constraints provide defense-in-depth.
**Complexity: MEDIUM** - medium diff; API surface change with status-aware mutation rules.
_Changes: Makes `updateBrand` status-aware so pending brands can clear or re-point their primary URL while active brands retain the immutable-once-set guard (6 files)._

<details>

- suggestion: Consider returning 409 when an active brand sends `baseSiteId: null` rather than silently ignoring it. The demotion guard and activation guard both throw typed errors; the immutability guard silently drops. A 409 with a message like "Cannot clear baseSiteId on an active brand" would make the API self-documenting and help consumers distinguish "nothing to change" from "change rejected." - `src/support/brands-storage.js:1075`
- nit: `type: [string, 'null']` in the OpenAPI schema is valid 3.1 syntax but will not be understood by older 3.0 codegen tools (which expect `nullable: true`). The schema already uses this syntax for `semrushWorkspaceId`, so this is consistent within the project. No action needed unless consumer tooling is a concern. - `docs/openapi/schemas.yaml:3814`
- suggestion: Add a unit test for the "clear + activate in same PATCH returns 400" path. The production code explicitly handles `wantsClearBaseSite` in the promote-to-active guard (line ~1101), and a targeted test would document the invariant. - `test/support/brands-storage.test.js`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 8s | <b>Cost:</b> $7.57 | <b>Commit:</b> <code>243209fc153d255121bb95cbae0447c74cf1afba</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jjenscodee (approved):

PR #2697, opened 2026-06-26
feat(llmo): pass Cloudflare account/zone IDs in body; scope zone listing by account

nit23uec:
Follow-up to #2681 (merged). Cloudflare identifiers are not SpaceCat ReBAC entities, so they should not appear as path segments.

## Changes

- **Route-create endpoint** is now `POST /sites/{siteId}/llmo/cdn-onboard/cloudflare/routes` with `{ zoneId, pattern }` in the **request body** (previously `.../zones/{zoneId}/routes`). This drops the `:zoneId` path param and its `FACS_NON_RESOURCE_PARAMS` classification.
- **`GET /sites/{siteId}/llmo/cdn-onboard/cloudflare/zones`** now requires an **`accountId` query param** and returns only the zones belonging to that Cloudflare account (filtered server-side on the zone's `account.id`).

`accountId` was already supplied in the deploy request body, so no change was needed there.

## Notes

- The published `@adobe/spacecat-shared-cloudflare-client@1.0.0` `listZones()` can't push `account.id` to the Cloudflare API, so the account scoping is applied in the controller. Pushing the filter down into the client (so the CF API does it) is a reasonable follow-up if/when that client gains the capability — same bucket as the deferred worker-existence pre-check.

## Testing

- Updated routes, capability classification (`required-capabilities.js` + `facs-capabilities.js`), OpenAPI specs, and unit + integration tests.
- `src/controllers/llmo/llmo-cloudflare.js` stays at **100% coverage**; full unit suite green; `docs:lint` / `docs:build` clean.

## Related Issues

_None linked._

PR #2701, opened 2026-06-26
feat(llmo): temporary cloudflareToken body/query fallback for CF token

nit23uec:
## Temporary workaround — revert before/after header allowlisting

The Cloudflare onboarding endpoints expect the Cloudflare token in the `x-cloudflare-token` header, but that custom header is **not yet allowlisted in the CDN/network config**, so it's stripped before reaching the service — blocking end-to-end verification.

This change adds a **temporary fallback**: `getCfToken` accepts the token from a `cloudflareToken` field in the request **body** (POST: deploy / routes) or **query string** (GET: accounts / zones) when the header is absent. The header still takes precedence.

### Notes
- **Intentionally NOT documented in the OpenAPI spec** — it's a temporary workaround, not part of the contract.
- ⚠️ **Revert this once `x-cloudflare-token` is allowlisted** in the network config. A token in a URL query is visible in access logs, so the query path in particular should not outlive the workaround.
- Controller stays at **100% coverage** (unit tests cover both the GET/query and POST/body fallback paths); full suite green.

Follow-up to the merged Cloudflare onboarding work (#2681, #2697).

## Related Issues

_None linked._

### DIFF
--- a/src/controllers/llmo/llmo-cloudflare.js
+++ b/src/controllers/llmo/llmo-cloudflare.js
@@ -20,6 +20,10 @@ import AccessControlUtil from '../../support/access-control-util.js';
 import { deriveWorkerName, hostInSiteDomain, routePatternHost } from './llmo-cloudflare-utils.js';
 
 const CF_TOKEN_HEADER = 'x-cloudflare-token';
+// TEMPORARY: body/query fallback field for the CF token, used only until the
+// x-cloudflare-token header is allowlisted in the CDN/network config (for e2e verification).
+// Intentionally undocumented in the OpenAPI spec. Remove once the header passes through.
+const CF_TOKEN_BODY_FIELD = 'cloudflareToken';
 const CF_TOKEN_MISSING = 'Missing x-cloudflare-token header';
 const EDGE_OPTIMIZE_API_KEY_SECRET = 'EDGE_OPTIMIZE_API_KEY';
 const EDGE_OPTIMIZE_TARGET_HOST_BINDING = 'EDGE_OPTIMIZE_TARGET_HOST';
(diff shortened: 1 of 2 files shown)


ssilare-adobe (approved):

src/controllers/llmo/llmo-cloudflare.js
```diff
@@ -61,8 +65,14 @@ function LlmoCloudflareController(ctx) {
...
-    const token = context.pathInfo?.headers?.[CF_TOKEN_HEADER];
-    return hasText(token) ? token : null;
+    const headerToken = context.pathInfo?.headers?.[CF_TOKEN_HEADER];
+    if (hasText(headerToken)) {
+      return headerToken;
+    }
+    // TEMPORARY fallback (see CF_TOKEN_BODY_FIELD): accept the token from the request
+    // body/query while the header is not yet allowlisted in the network config.
+    const fallbackToken = context.data?.[CF_TOKEN_BODY_FIELD];
+    return hasText(fallbackToken) ? fallbackToken : null;
```
MysticatBot, line 75: issue (blocking): The fallback path has no observable signal. Without a log entry when this branch is taken, the team cannot measure adoption or confirm zero fallback usage before removal. The plan is to remove once the header passes through, but without telemetry you cannot verify that no caller still depends on this path.

**Why it matters:** Temporary workarounds without removal signals become permanent fixtures. This one-liner gives the team grep-able evidence in logs to confirm it is safe to delete.

**Suggested fix:**
```js
const fallbackToken = context.data?.[CF_TOKEN_BODY_FIELD];
if (hasText(fallbackToken)) {
  log.warn('getCfToken: using body/query fallback - header not present');
  return fallbackToken;
}
return null;
```

This also creates an alert-able condition if the workaround outlives its intent.

MysticatBot (dismissed):
Hey @nit23uec,

**Verdict: Request changes** - missing observability on the fallback path leaves the removal plan unverifiable.
**Complexity: LOW** - trivial diff; API surface signal.
_Changes: Adds a temporary body/query fallback for the Cloudflare token when the x-cloudflare-token header is stripped by CDN (2 files)._

## Must fix before merge
1. [Important] No log signal when fallback path is used - removal plan has no observable trigger - `src/controllers/llmo/llmo-cloudflare.js:74` (details inline)

<details>

- suggestion: Consider restricting the fallback to POST-only requests so the token never travels in a URL query string. The body path has no log-exposure vector; the GET/query path does. This would narrow the acknowledged risk while still unblocking e2e verification for deploy/routes endpoints - `src/controllers/llmo/llmo-cloudflare.js:74`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 43s | <b>Cost:</b> $2.46 | <b>Commit:</b> <code>0c36a1e4c22386099041a93b449303b21c06bdd8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/controllers/llmo/llmo-cloudflare.js
nit23uec: this is not required now as this is only a temp workaround for ETE POC. no alerting is required.

Issue #2678, opened 2026-06-24
Extend opt-in type-checking to the Serenity controllers (ADR-005 ratchet)


rainer-friederich:
## Context

spacecat-api-service has opt-in TypeScript type-checking (`checkJs: false` + per-file `// @ts-check`, blocking `npm run type-check` gate) scoped to `src/support/serenity/**` — see `docs/decisions/005-opt-in-type-checking.md` (ADR-005).

In PR https://github.com/adobe/spacecat-api-service/pull/2643 we:
- added `src/support/serenity/CLAUDE.md` codifying the convention (every Serenity JS file carries `// @ts-check`, stays in the tsconfig `include`, uses the typed clients/models, and follows the documented idioms), and
- closed the one support-file gap (`site-linkage.js` was missing the pragma).

The **Serenity controllers were deliberately left out** of that PR — turning on `@ts-check` there surfaces ~49 real type errors, and #2643 was already over the auto-review tooling's size budget. This issue tracks the documented ADR-005 ratchet step so we don't forget.

## Scope

1. Add `src/controllers/serenity.js` and `src/controllers/brands.js` to the tsconfig `include` (currently `src/support/serenity/**/*.js` only).
2. Add `// @ts-check` to both files. Note: `brands.js`'s license header is `*/` immediately followed by `import` with **no blank line** — insert the pragma + a blank line after `*/`.
3. Fix the surfaced errors (measured against the current branch):
   - **`serenity.js`: 26 errors**
   - **`brands.js`: 23 errors** (17× `TS2345` argument-type, 4× `TS2322` assignment, 2× `TS2353` excess-property)
4. Update `docs/decisions/005-opt-in-type-checking.md` to record the scope expansion.

## Guidance

- Use the same idioms as the support files: narrow at boundaries (`!x || !hasText(x)` — `hasText` is not a type guard), reach data-access model values via a namespace cast (the package re-exports its models as type-only), and complete JSDoc `@param` tags. Do **not** widen shared helper signatures or blanket-cast to `any`.
- `brands.js` is the **general** brands controller (not purely Serenity), so type-checking it benefits the whole controller surface, not just Serenity.

## Done when

- [ ] Both controllers carry `// @ts-check` and are in the tsconfig `include`.
- [ ] `npm run type-check` is green with zero errors.
- [ ] ADR-005 updated to reflect the expanded scope.

rainer-friederich:
[...]

PR #2702, opened 2026-06-26
feat(serenity): extend opt-in type-checking to the serenity controllers (ADR-005 ratchet)

rainer-friederich:
## 1. Abstract
Turns on the repo's opt-in TypeScript type-checking (`// @ts-check` + JSDoc) for the two Serenity controllers — `src/controllers/serenity.js` and `src/controllers/brands.js` — and fixes every error the checker surfaces.

## 2. Reasoning
The opt-in type-check seam (ADR-005) was scoped to `src/support/serenity/**`. The Serenity controllers that drive those support files were deliberately left out of the earlier PR (https://github.com/adobe/spacecat-api-service/pull/2643) because turning the pragma on there surfaces ~50 real errors and that PR was already over the auto-review tooling's size budget. This is the documented ADR-005 ratchet step (step 1: expand opt-in scope) so the controller surface is held to the same typed contract as the support layer. `brands.js` is the general brands controller — not purely Serenity — so type-checking it benefits the whole controller surface.

- Adds `// @ts-check` to both controllers and adds them to the tsconfig `include`. The blocking `type-check` CI job and the pre-commit hook now cover both files.
- The dominant surfaced error class was an over-narrow **published** type, not our code: `@adobe/spacecat-shared-http-utils` types `ok()`'s body parameter as `string`, while at runtime `ok` delegates to `createResponse` and JSON-stringifies an object (the package's own `createResponse` JSDoc says `object|string|Buffer`). Every idiomatic `ok({ ... })` call was therefore flagged. Rather than patch the dependency's types locally, the flagged object-bodied `ok(x)` calls are switched to the sibling `createResponse(x, 200)` — which the shipped types already accept for objects, and which is what `ok` calls internally, so it is runtime-identical. Both controllers already use `createResponse` elsewhere. The five deliberate empty-body `createResponse(null, 204)` responses become `noContent()` (also 204), which drops a stray `"null"` body (a 204 must carry no body). The proper long-term fix is upstream in `spacecat-shared-http-utils`; if that lands, these sites can revert to `ok(x)`.
- The remaining errors use the idioms already established for the support files, with no shared-helper signature widened and no blanket `any` cast: boundary null/undefined narrowing (`?? ''` where a callee already treats empty like null, `?? undefined` where the param is optional, `!x || !hasText(x)` / `x && hasText(x)`), a `Record<…>` annotation for a dynamically-keyed object that inferred as `{}`, readonly-tuple spreads, a single local non-null assertion guaranteed by `authorize()`, and completing previously-incomplete JSDoc `@param` tags on `upsertPrompts` / `upsertBrand` (those helpers already accept the params — doc-only).
- Updates ADR-005 to record the scope expansion and the per-class fixes; it now references `tsconfig.json` rather than pasting a copy that would drift.

## 4. Required information
- Jira / issue: https://github.com/adobe/spacecat-api-service/issues/2678
- ADR(s): docs/decisions/005-opt-in-type-checking.md (updated in this PR)

Because a "type fix" must not silently change runtime behaviour, the diff was reviewed specifically for behaviour-preservation: each narrowing rewrite, each `?? ''` / `?? undefined`, the readonly-tuple spreads, the `auth.brandUuid` assertion, and the `classifyIntent ?? undefined` change were checked against the actual callee code (e.g. `ensureSubworkspace` / `handleCreateMarketSubworkspace` throw a 404 on a falsy parent workspace, so `null` and `''` are equivalent there; the prompt helpers gate the classifier on `typeof === 'function'`, so `null` vs `undefined` is a no-op). The `ok(x)` -> `createResponse(x, 200)` swap is runtime-identical (`ok` is literally `createResponse(x, 200, {})`). The only behavioural delta is the 204 responses no longer carrying a `"null"` body.

## 7. Test plan
(a) Apart from the 204 empty-body normalization, this is a JSDoc/types-only change (`noEmit`, no `.ts`), so there is no new runtime path to exercise. The existing Serenity and brands controller unit suites were run locally and remain green, confirming the narrowings and the `ok`->`createResponse` swap are behaviour-preserving; the delete/204 endpoints still return 204.
(b) No environment-specific verification is required — there is no deployable behaviour change beyond the empty 204 bodies. The only operational surface is the blocking `type-check` gate (CI + pre-commit), which now covers the two controllers.


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-executed type-safety ratchet step with one correct behavioral improvement (204 body cleanup).
**Complexity: MEDIUM** - medium diff, API surface signal.
_Changes: extends opt-in TypeScript type-checking to the two Serenity controllers per ADR-005 ratchet step 1, fixing ~50 type errors and removing stray `"null"` bodies from 204 responses (6 files)._
Note: Recommend a human read before merge - this change modifies an architectural decision record (`docs/decisions/005-opt-in-type-checking.md`). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: CLAUDE.md inline scope hint `(serenity)` in the `type-check` command comment is now incomplete since scope includes `brands.js` (the general brands controller) - `CLAUDE.md:23`
- nit: `auth.parentWorkspaceId ?? ''` (lines 534, 834, 928, 955) vs `?? undefined` (line 1165) in `src/controllers/serenity.js` use different coalescing strategies - both are correct for their respective callees but a one-line comment at each site explaining why would help future readers
- suggestion: the `/** @type {string} */ (auth.brandUuid)` assertion at `src/controllers/serenity.js:784` is safe given authorize()'s contract, but a runtime guard (`assert(auth.brandUuid)`) would be more resilient to future refactors of authorize()
- suggestion: the delete-market handler discards the result and hardcodes `noContent()` - the comment in the diff documents the contract, but an `assert(result.status === 204)` would make it fail loudly if either handler ever evolves beyond 204 - `src/controllers/serenity.js:594`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 58s | <b>Cost:</b> $6.49 | <b>Commit:</b> <code>3f040b22045ea4068ebecf9a155d1371948bca68</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2637, opened 2026-06-18
feat(brands): guard against active->pending demotion in updateBrand/upsertBrand (LLMO-5587)

aliciadriani:
## Why
Active brands can be silently demoted to `pending` by the per-brand write primitives (`updateBrand` / `upsertBrand`) — no error to the editor, no alert. A demoted brand stops binding in the scheduler (the resolver requires `status == active`) and its site goes dark. This took **express.adobe.com** offline on 2026-06-05 (brand `40f16625`); same class as the MongoDB and Merck/MSD incidents. Incident: [LLMO-5587](https://jira.corp.adobe.com/browse/LLMO-5587).

## Deploy gate
This PR's 409 demotion guard must **NOT** deploy until elmo-ui #2139 is live in prod. Otherwise elmo's still-generic "Move to pending" hits the guard and 409s (loud, not silent).

**Order:** spacecat #2621 → elmo #2137 → elmo #2139 (prod) → this PR (#2637).

## What
The generic per-brand write paths now **refuse** to silently demote an active brand to pending; intentful demotions go through the dedicated status-transition endpoint added in #2621.

- **`updateBrand`** — when a PATCH would take a persisted-`active` brand to `pending` → **409** `brand_status_demotion_not_allowed`.
- **`upsertBrand`** — when a by-name (`org, name`) create/upsert would resolve onto an existing **active** brand and demote it → same **409**. (Closes the M1 collision vector.)
- The single existing-row fetch is reused for all three brand guards (immutability, demotion, active-without-site).

## Credit — re-landed from #2504 (@igor-grubic)
#2504 was closed in favor of this staged work; its complementary guards are re-landed here and credited in the code:
- the **`23514` / `chk_active_brand_has_site_id` → 400** mapping (covers the SELECT→write race), and
- the **active-without-site promotion guard** (`status:'active'` with no `site_id` → 400).
- Its **existing-fetch pattern** is the basis for the fetch here, broadened from `site_id`-only to also read `status` for the demotion comparison.

## Status-code semantics
- **409** — demotion of an active brand via the generic path (a routing/state conflict; the caller should use the `/status` endpoint).
- **400** — row-invariant violations (activate-without-site, `23514`).

## Alerting — `BrandDemotionBlocked` metric
A best-effort CloudWatch **EMF** metric is emitted from the reject branch (never affects the response), modeled on the LLMO-5150 P1 pattern:
- **Namespace:** `Mysticat/Brands` · **Metric:** `BrandDemotionBlocked` (Count)
[...]

PR #2621, opened 2026-06-16
feat(brands): add explicit brand status-transition endpoint (LLMO-5587)

aliciadriani:
## Why
Active brands can be silently demoted to `pending` by the per-brand write primitives (`updateBrand` / `upsertBrand`) — no error to the editor, no alert. A demoted brand stops binding in the scheduler (the resolver requires `status == active`) and its site goes dark. This took **express.adobe.com** offline on 2026-06-05 (brand `40f16625`, org `5d4e5082`); same class as the MongoDB and Merck/MSD incidents documented in #2593. Incident: [LLMO-5587](https://jira.corp.adobe.com/browse/LLMO-5587).

## What — PR1 of a staged rollout (additive)
Expand → migrate → contract. This is the **additive expand step**: a dedicated, intentful status-transition endpoint so legitimate demotions have a sanctioned path *before* the generic `PATCH` is locked down.

- `PATCH /v2/orgs/{spaceCatId}/brands/{brandId}/status` → `transitionBrandStatusForOrg` (capability `organization:write`)
- New `setBrandStatus` storage fn — minimal (status + `updated_by` only, no child-table sync), deliberately **separate** from `updateBrand` so it carries no demotion guard and the generic path can.
- Body `{ "status": "active" | "pending" }`; `400` invalid status / activate-without-site, `403/404/503` mirror the sibling brand handlers.
- **No DB migration.** The row invariant (`active ⇒ site_id`) stays at the data layer via `chk_active_brand_has_site_id`.

## Credit — lifted from #2504 (@igor-grubic)
The `23514` / `chk_active_brand_has_site_id` → `400` mapping is **lifted verbatim from @igor-grubic's #2504** (closed in favor of this staged work). PR3 will additionally **adapt #2504's `needsExistingFetch` existing-row fetch** (broadened to `select('status')`) for the demotion comparison — credit carried forward there too. Pattern parallel: #2593 (@byteclimber) anchors immutability in the same `brands-storage.js` block.

## Staged rollout / dependency chain
1. **PR1 (this PR, spacecat #2621)** — additive transition endpoint. Safe to merge/deploy independently.
2. **PR2 (elmo-ui #2137)** — migrate `moveToPending` (and `approve`) onto this endpoint. **Must deploy before PR3.**
3. **PR2 prod (elmo-ui #2139)** — elmo-ui migration live in prod. **Must be live before PR3 deploys.**
[...]

PR #2706, opened 2026-06-26
feat(brands): self-serve activate-brand endpoint | LLMO-5605

igor-grubic:
## 1. Abstract
Adds `POST /v2/orgs/{spaceCatId}/brands/{brandId}/activate` — the explicit, customer-triggered "make this brand live" action: resolves the onboarded primary site, anchors it and promotes the brand to `active` via the **sanctioned `setBrandStatus` transition path**, optionally generates AI prompts, and ensures the recurring brand-presence schedule. LLMO-5605, Piece 2.

## 2. Reasoning
Piece 2 of self-serve domain onboarding (LLMO-3749 / LLMO-5605). After Piece 1 stands up the site entity, the customer needs an explicit action to promote a pending brand whose primary URL is already onboarded — optionally kicking off AI-prompt generation and the recurring brand-presence schedule.

New endpoint `POST /v2/orgs/{spaceCatId}/brands/{brandId}/activate`. For an existing pending brand whose primary URL is already onboarded:

1. **State guard (promote-only, idempotent):** a soft-`deleted` brand → `404` (no resurrection); an already-`active` brand → `200` no-op (retry-safe, no duplicate work); any other non-`pending` state → `400`.
2. **Resolve the onboarded primary `Site`** — from the brand's `baseSiteId`, else its stashed Semrush `primaryUrl`; it deliberately does **not** guess from `urls[]`. `400` if no onboarded site resolves.
3. **Anchor + promote as two sanctioned steps (LLMO-5587):**
   - (a) If the brand isn't already anchored, set `baseSiteId` via `updateBrand` — a **field write, not a status change** (brand stays pending; `409` on `brands_base_site_unique`).
   - (b) Promote `pending → active` via **`setBrandStatus`** — the sanctioned status-transition path added in #2621. `updateBrand` deliberately refuses status demotions (#2637), so every brand status transition routes through the one auditable primitive.
4. **Optional prompt generation:** if `generatePrompts`, submit a **direct** `prompt_generation_base_url` DRS job with `source: 'brand-activation'` (never Brandalf, never the `llmo-customer-analysis` cascade). Best-effort idempotent via an in-flight `listJobs` check.
5. **Schedule:** ensure the recurring brand-presence schedule **only when prompts exist / are generated**, via the shared `createBrandPresenceSchedule` helper (POST + tolerate `409`).
6. **Best-effort side-effects:** prompt-gen + schedule run in a guarded block **after** the brand is committed active — a DRS failure returns **`200`** (the brand *is* active) with a `:warning:` Slack alert to ops carrying the real error; it never surfaces a `5xx` or leaks upstream detail to the client.
7. Returns **`200`** quickly (`{ brandId, status:'active', baseSiteId, promptGenerationJobId?, scheduleId? }`); prompts generate async.

- **Auth:** membership (`hasAccess`) + explicit **PAID** entitlement (`TierClient`), mirroring Piece 1 — no admin requirement (decision: don't `403` paying non-admin members).
- OpenAPI specs (`api.yaml` / `brands-v2-api.yaml` / `schemas.yaml`) updated; bumps `@adobe/spacecat-shared-drs-client` → 1.13.0.

## 4. Required information
- Jira / issue: LLMO-5605 — https://jira.corp.adobe.com/browse/LLMO-5605
- Spec: Piece 2 design spec — https://github.com/adobe/project-elmo-ui/issues/2095
- Other: Piece 1 spec — https://github.com/adobe/project-elmo-ui/issues/2092

- **spacecat-shared** (drs-client) — *consumed*: uses `listJobs` (dedup) + `createBrandPresenceSchedule` from 1.13.0 — https://github.com/adobe/spacecat-shared/pull/1721 (merged, released).
- **llmo-data-retrieval-service** (DRS) — *depends-on (runtime)*: the `(site, provider_id, source)` dedup needs the `source` filter from https://github.com/adobe-rnd/llmo-data-retrieval-service/pull/2370 deployed.
- **spacecat-audit-worker** — *contract / parallel*: shares the `createBrandPresenceSchedule` helper with the `llmo-customer-analysis` cascade (https://github.com/adobe/spacecat-audit-worker/pull/2737); both route through the same method to preserve DRS provider-set dedup.

## 7. Test plan
(a) **Local unit** — coverage for: the two-step anchor+promote (`200`); the promote-only state guard (`deleted → 404`, `active → 200` no-op, non-pending → `400`); `generatePrompts: true` → one prompt job + one schedule; `false` / no-prompts → neither; double-click / QUEUED+RUNNING idempotency; `baseSiteId` re-lookup; DRS side-effect failure → `200` + `:warning:` alert (never `5xx`/leak); alert-throw safety; and the `400` / `403` / `404` / `409` (anchor conflict) / `503` paths. `brands.js` at 100% line/stmt/func.

(b) **Integration (`test/it/`)** — **done**: `test/it/shared/tests/activate-brand-for-org.js` + PostgREST wiring exercise the auth gate (org lookup → membership → PAID) + request validation + org-scoped brand lookup against a real DB (modeled on the `onboard-site-only` IT). Passing in CI (`it-postgres`).

(c) **Per-env (dev / stage):** activate a pending brand whose primary URL is onboarded → confirm `status:'active'` with `baseSiteId` set; with `generatePrompts:true` confirm a single `source:'brand-activation'` prompt job and one brand-presence schedule (re-activate is idempotent); with `generatePrompts:false` confirm neither; confirm a non-PAID or non-member caller is rejected.

Final consumer in the LLMO-5605 Piece 2 chain.
- drs-client 1.13.0 (spacecat-shared) — *released* — https://github.com/adobe/spacecat-shared/pull/1721.
- llmo-data-retrieval-service `source` filter — *depends-on (deploy)*: deploy to DRS before this endpoint's prompt-gen dedup discriminates by source — https://github.com/adobe-rnd/llmo-data-retrieval-service/pull/2370.
- spacecat-audit-worker anti-drift migration — *related (parallel)*: keep both on the shared schedule helper — https://github.com/adobe/spacecat-audit-worker/pull/2737.
- **The in-repo `/status` (#2621) + demotion-guard (#2637) work has merged**; this branch has been **re-synced with `main`**, and activate now delegates its promotion to the `setBrandStatus` primitive those PRs introduced. No remaining pre-merge blockers on this PR (approved · CI green · mergeable).


MysticatBot (changes requested):
Hey @igor-grubic,

**Verdict: Request changes** - post-activation alert failure leaks a 500 to the client despite successful brand mutation.
**Complexity: MEDIUM** - medium diff; API surface + dependency update.
_Changes: Adds explicit self-serve brand activation endpoint with paid entitlement gate, DRS prompt generation, and brand-presence schedule creation (14 files)._

## Must fix before merge

1. [Critical] `postLlmoAlert` failure after successful activation returns error response to client despite brand already being active in DB - `src/controllers/brands.js:1168` (details inline)
2. [Important] No guard against re-activating an already-active brand; retry after alert-failure 500 may mutate `baseSiteId` - `src/controllers/brands.js:1013` (details inline)
3. [Important] Missing test coverage for QUEUED job idempotency path (code checks both QUEUED and RUNNING but only RUNNING is tested) - `test/controllers/brands.test.js:1590` (details inline)

<details>

- nit: `brandId` path parameter lacks UUID validation unlike `spaceCatId` which gets `isValidUUID()` - `src/controllers/brands.js:1027`
- nit: 409 description in OpenAPI exposes internal constraint name `brands_base_site_unique` - `docs/openapi/brands-v2-api.yaml:288`
- suggestion: differentiate "no primary URL configured" from "primary site not onboarded" in the 400 response for better client debugging - `src/controllers/brands.js:1078`
- suggestion: add test for string-form entries in `brand.urls[]` since the code defensively handles `typeof u === 'string'` - `src/controllers/brands.js:1072`
- suggestion: DRS client instantiation could be lazily deferred since it's unused when `generatePrompts:false` and no existing prompts - `src/controllers/brands.js:1100`
</details>

src/controllers/brands.js
```diff
@@ -2007,6 +2011,194 @@ function BrandsController(ctx, log, env) {
...
+      if (!generatePrompts) {
+        const stats = await getPromptStats({
+          organizationId: spaceCatId,
+          brandUuid,
+          postgrestClient,
+        });
+        hasExistingPrompts = (stats.branded + stats.unbranded) > 0;
+      }
+      if ((generatePrompts || hasExistingPrompts) && drsConfigured) {
+        const schedule = await drsClient.createBrandPresenceSchedule({
```
MysticatBot: issue (blocking): **Post-activation alert failure returns error to client despite brand already being active.**

If this line throws after updateBrand succeeds (line 2087), the catch block returns createErrorResponse(error). The client sees a 500, but the brand is already active in the database.

**Why it matters:** This violates the API contract (200 means activated, 500 means it did not). Clients that retry will find the brand already active with potentially different promptGenerationJobId/scheduleId values.

**Fix:** Wrap all post-activation steps (DRS job submission, schedule creation, alerting) in their own try/catch that logs failures but always returns 200. The activation itself succeeded; these are best-effort post-activation actions.

src/controllers/brands.js
```diff
@@ -2007,6 +2011,194 @@ function BrandsController(ctx, log, env) {
...
+      // column (getStatus() is an unbacked stub; revocation = row delete), so a PAID
+      // row is the "paying" signal. No separate admin requirement — mirrors Piece 1,
+      // which omits it so paying non-admin members aren't 403'd.
+      if (!await accessControlUtil.hasAccess(organization)) {
+        return forbidden('User does not have access to this organization');
+      }
+      if (!await hasPaidLlmoEntitlement(context, organization)) {
+        return forbidden('A paid LLMO entitlement is required to activate a brand');
+      }
+
```
MysticatBot, line 2114: issue (blocking): **No idempotency guard for re-activating an already-active brand.**

The endpoint does not check brand.status before proceeding. If a client retries (e.g., after receiving a 500 from the alert-failure bug), updateBrand runs again with status: active and whatever baseSiteId resolves now. If the URL mapping changed between calls, baseSiteId could silently mutate.

**Why it matters:** Idempotency is critical for retry safety on state-mutating endpoints. The OpenAPI spec says this promotes a pending brand but does not reject an already-active one.

**Fix:** After fetching the brand, add a short-circuit for already-active brands:

```javascript
if (brand.status === 'active') {
  return createResponse({
    brandId: brandUuid,
    status: 'active',
    baseSiteId: brand.baseSiteId,
  }, 200);
}
```

This makes retries safe and documents the intended one-shot semantics.

test/controllers/brands.test.js
```diff
@@ -6375,6 +6375,606 @@ describe('Brands Controller', () => {
...
+        ...mockDataAccess,
+        services: { postgrestClient: {} }, // no `from`
+      };
+      const response = await controller.activateBrandForOrg(
+        buildActivateRequest({ dataAccessOverride: brokenDataAccess }),
+      );
+      expect(response.status).to.equal(503);
+    });
+
+    // -------------------------------------------------------------------------
```
MysticatBot, line 6855: issue (blocking): **Missing test for QUEUED job idempotency.**

The production code checks `job.status === 'QUEUED' || job.status === 'RUNNING'` but only the RUNNING case is tested here. Since QUEUED is an explicit code branch in the find() predicate, it should have its own test verifying submitPromptGenerationJob is not called and the existing job_id is returned.

**Fix:** Add a companion test:
```javascript
it('reuses an in-flight QUEUED job instead of submitting a new one', async () => {
  const listJobsStub = sinon.stub().resolves([{ job_id: 'queued-1', status: 'QUEUED' }]);
  // ... same pattern as this RUNNING test
  expect(body.promptGenerationJobId).to.equal('queued-1');
  expect(submitJobStub).to.not.have.been.called;
});
```

igor-grubic:
Thanks for the thorough review — pushed `e17a859c` addressing the should-fix and medium items.

**Should-fix**
- **#1 soft-delete resurrection** — `activateBrandForOrg` now rejects `status:'deleted'` with a 404 (treat-as-gone), parity with the `/status` sibling's `.neq('status','deleted')` guard. Added a test asserting `updateBrand` is never called.
- **#4 partial-failure semantics** — prompt-gen + schedule are now treated as best-effort async side-effects. They run in a guarded block **after** the brand is committed active; on a DRS failure the client gets a **200** with the brand active (and whatever ids succeeded), instead of a misleading 5xx. The `:x: activation failed` alert now only fires for true activation failures (validation / `updateBrand`).
- **#5 raw DRS detail leaking to the client** — resolved as a side effect of #4: DRS errors are caught inside the guarded block and never reach `createErrorResponse`, so they can't echo into the body / `x-error`. The real (offending) error is **not lost** — it's sent to ops via a `:warning:` Slack alert that includes `error.message`. The only errors that still surface to the client are the clean validation / `409 \"This site is already the primary URL…\"` messages.

**Medium / nits**
- **#6** — dropped the redundant `resolveBrandUuid` round-trip; `getBrandById(spaceCatId, brandId)` is fetched directly (brandId is UUID-validated upstream in `index.js`).
- **#7** — when `baseSiteId` is set but the `base_site` embed carried no `baseUrl`, we now re-look up the site by id so prompt-gen never submits an empty `base_url`.
- **#9 / urls[] fallback** — went a step further than suggested: per maintainer preference we **removed the `urls[]` fallback entirely**. `urls[]` is the brand's listed URLs, not a declared activation anchor, so guessing one is wrong. Resolution is now `baseSiteId` → `pendingSemrushProvisioning.primaryUrl` only; otherwise 400. Added a test asserting we don't touch `urls[]`.
- **#8** — added a comment documenting that the prompt-gen dedup is best-effort (TOCTOU) and depends on the DRS `source` filter deploy; the weekly schedule retains real 409 dedup.
- **#10** — left as-is intentionally: `postLlmoAlert` swallows its own errors and this matches the other LLMO controllers. Flagging the latency tradeoff is fair; happy to move it off the hot path in a follow-up if you'd prefer.

`brands.js` stays at 100% line/stmt/func coverage; 382 brand tests pass, lint + type-check clean.

**Still open / need your call**
- **#2 integration tests** — acknowledged as the gating item; IT in `test/it/` is tracked as a separate task (needs seed data + wiring).
- **#3 docs/index.html** — currently *intentionally* not regenerated (a maintainer preference in this workspace to avoid committing the rendered bundle). You're right that leaving it tracked-but-stale is the worst of both, and that my earlier commit message mischaracterized it. How would you like to reconcile: (a) rebuild + commit per the CLAUDE.md convention, or (b) a deliberate `git rm --cached docs/index.html` to untrack it? I'll do whichever you prefer.

igor-grubic:
Addressed the MysticatBot review in `ec2c79a5`:

- **① Post-activation 500** — already handled by the prior commit (`e17a859c`): DRS job-submission + schedule now run in a guarded block that always returns 200, and `postLlmoAlert` wraps its Slack call in its own try/catch (never re-throws). Hardened further here by moving `DrsClient.createFrom()`/`isConfigured()` *inside* that guard, so **nothing** after the `updateBrand` commit can produce a 5xx. The outer catch now only covers pre-commit failures (validation / auth / `updateBrand`).
- **② Already-active idempotency** — added a short-circuit: if `brand.status === 'active'`, return 200 with the current state without re-running the write, prompt-gen, or schedule. This makes retries safe (avoids a duplicate prompt-gen job when the prior one has COMPLETED — the `listJobs` dedup only catches QUEUED/RUNNING) and documents the one-shot promote semantics. (Note: the comment's `baseSiteId` mutation risk can't actually occur — active brands always have `baseSiteId` set, so the resolution branch is skipped and it's immutable-from-null in storage; the real win is retry-safety.)
- **③ QUEUED idempotency test** — added the companion test asserting a `QUEUED` in-flight job is reused and `submitPromptGenerationJob` is not called.

`brands.js` stays at 100% line/stmt/func coverage (30 activate tests, 379 total); lint + type-check + docs:lint clean.

src/controllers/brands.js
```diff
@@ -2007,6 +2011,251 @@ function BrandsController(ctx, log, env) {
...
+                ...(hasText(region) ? { region } : {}),
+                source: 'brand-activation',
+              });
+              promptGenerationJobId = job?.job_id;
+            }
+          } else {
+            log.warn('DRS client not configured; skipping prompt generation');
+          }
+        }
+
```
MysticatBot, line 2255: issue (blocking): **Missing integration tests for new endpoint.**

CLAUDE.md "Adding a New Endpoint" step 10 requires: "Write integration tests in `test/it/`." The PR body (section 7) acknowledges this as outstanding pre-merge work. This finding confirms the convention requirement blocks merge.

**Fix:** Add integration tests following the documented pattern: seed IDs in `test/it/shared/seed-ids.js`, seed data in `postgres/seed-data/`, registration in `postgres/seed.js`, shared test factory in `test/it/shared/tests/`, wiring file in `postgres/`.

MysticatBot (changes requested):
Hey @igor-grubic,

**Verdict: Request changes** - integration tests required by CLAUDE.md are missing; secondary hardening items around error propagation.
**Complexity: MEDIUM** - medium diff; API surface + dependency update.
_Changes: Adds explicit self-serve brand activation endpoint with paid entitlement gate, DRS prompt generation, and brand-presence schedule creation (14 files)._

## Must fix before merge

1. [Important] Missing integration tests in `test/it/` for the new endpoint per CLAUDE.md "Adding a New Endpoint" step 10 - author acknowledges as pre-merge item - `test/it/` (details inline)
2. [Important] `postLlmoAlert` in the success/warning path is awaited outside the inner guarded block; if it throws (despite its internal guard), the outer catch returns an error after the brand is already active - wrap the post-commit alert calls in their own try/catch - `src/controllers/brands.js:2217`
3. [Important] No status precondition on activation beyond active/deleted; code accepts any state (e.g. hypothetical future `suspended`/`draft`) though PR says "promotes a pending brand" - add explicit `brand.status === 'pending'` check or document the any-state semantics intentionally - `src/controllers/brands.js:2075`

<details>

- suggestion: guard `getPromptStats` call with `drsConfigured` check to skip a wasted DB roundtrip when DRS is unavailable - `src/controllers/brands.js:2153`
- nit: comment says "brandId is UUID-validated upstream (index.js)" but the route framework does not auto-validate path params; the guard is `getBrandById` returning null - `src/controllers/brands.js:2057`
- nit: OpenAPI 409 description exposes internal constraint name `brands_base_site_unique`; rephrase to user-facing language - `docs/openapi/brands-v2-api.yaml:288`
- suggestion: add a test verifying `hasPaidLlmoEntitlement` throwing results in a safe error without leaking upstream detail - `test/controllers/brands.test.js`
</details>

- Post-activation alert failure no longer returns error to client; all post-commit DRS steps run in a guarded block that always returns 200.
- Already-active brand now short-circuits to 200 no-op without re-running writes or side effects.
- QUEUED job idempotency path now has dedicated test coverage.

igor-grubic:
Round 2 addressed in `3879db7a`:

- **[must-fix] Explicit promote-only precondition** — `activateBrandForOrg` now rejects any non-pending state with a 400 (`deleted → 404` and `active → 200 no-op` are handled just above), so a hypothetical future `suspended`/`draft` brand can't be silently activated.
- **[must-fix] Alert wrapped** — the post-commit `postLlmoAlert` calls now run in their own try/catch, so even if the alert threw (it already swallows its own Slack errors) it can't escape to the outer catch and 5xx an already-committed activation. Combined with the earlier guard, **nothing** after the `updateBrand` commit can return an error to the client.
- **[suggestion] getPromptStats** — now skipped when DRS is unavailable (no schedule would be created anyway), avoiding a wasted DB roundtrip.
- **[nit] OpenAPI 409** — reworded to user-facing language (no longer names the internal `brands_base_site_unique` constraint).
- **[nit] comment** — clarified that `brandId` is UUID-validated by the shared param guard in `index.js` before the controller runs (with `getBrandById` returning null as the safety net).

`brands.js` stays at 100% line/stmt/func (33 activate tests, 382 total); lint + type-check + docs:lint clean.

**Remaining:** the integration tests (#1) — still the gating item, tracked separately. I'll follow up on those.

MysticatBot (approved):
Hey @igor-grubic,

**Verdict: Approve** - prior findings all addressed; no remaining blocking issues.
**Complexity: HIGH** - large diff; API surface + dependency update.
_Changes: Adds self-serve brand activation endpoint with paid entitlement gate, DRS prompt generation, brand-presence schedule, and full test coverage (16 files)._

<details>

- nit: `postLlmoAlert` in the outer catch block is not wrapped in try/catch, unlike the inner success/warning path which is defensively wrapped. Author previously noted this is intentional (postLlmoAlert swallows its own errors internally, matches other LLMO controllers). Cosmetic inconsistency only - `src/controllers/brands.js:2253`
- nit: `brandId` path parameter is not inline-UUID-validated in the controller body (relies on upstream route param guard in index.js). The comment documents this, but adding a defensive `isValidUUID(brandId)` check (same as `spaceCatId`) would make the controller self-contained - `src/controllers/brands.js:2057`
- suggestion: add a test verifying `hasPaidLlmoEntitlement` propagates (does not swallow) when `TierClient.checkValidEntitlement()` rejects - `test/support/llmo-paid-gate.test.js`
- suggestion: OpenAPI 400 description says "Brand has no onboarded primary site" but 400 also fires for missing/invalid params, non-boolean generatePrompts, and non-pending status. Broaden to "Invalid request (see response body for details)" or list the cases - `docs/openapi/brands-v2-api.yaml:288`

</details>

- Post-activation alert failure no longer returns error to client; all post-commit DRS steps run in a guarded block that always returns 200.
- Already-active brand short-circuits to 200 no-op without re-running writes or side-effects.
- QUEUED job idempotency path now has dedicated test coverage.
- Status precondition added (pending-only; deleted returns 404, active returns 200 no-op).
- Integration tests now present in test/it/ covering auth gate and brand lookup against real PostgREST.
- getPromptStats call now skipped when DRS is unavailable.
- OpenAPI 409 description reworded to user-facing language.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 23s | <b>Cost:</b> $6.76 | <b>Commit:</b> <code>090d656bacf3ca7582677734c8b4a36dacc29bc3</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich (changes requested):
Hey @igor-grubic,

**Verdict: Request changes** - the endpoint logic is correct and the whole LLMO-5605 Piece 2 chain verifies end-to-end per the spec, but there are 4 items worth addressing (2 blocking, 1 blocking-ish, 1 nit). This was a local multi-repo review across #2706 / audit-worker 2737 / DRS 2370; the cross-repo seams (dedup-key parity, `listJobs(source)` filter contract, "no shebang", and the LLMO-5604 resolver already being on DRS `main`) all check out.

## Must fix before merge

1. **[Important] Outer-catch `postLlmoAlert` is not defensively wrapped - it can turn a 4xx into a 500.** In `activateBrandForOrg`, the inner best-effort side-effect block correctly hardens its alert so an alert failure can never escape. The **outer** catch does not: `await postLlmoAlert(...)` runs before `createErrorResponse(error)`, so if that alert call rejects, the rejection escapes the handler and the intended status (e.g. the `brands_base_site_unique` 409, or a 400) is replaced by an unhandled-rejection 500. This is the exact failure mode the inner block was hardened against, left unguarded on the error path. Fix: wrap the outer alert in its own try/catch (log `alertError`), identical to the inner block, then `return createErrorResponse(error)` unconditionally. `src/controllers/brands.js`.

2. **[Important] Integration tests only cover the auth gate, not the behaviour the spec's test plan requires.** The spec (project-elmo-ui#2095, Testing / line 176) explicitly requires `test/it/` coverage of: real promotion -> brand `active` + `base` `brand_sites` row, exactly one prompt-gen job + one schedule when `generatePrompts:true`, `generatePrompts:false` -> no schedule, and double-click idempotency (no duplicate schedule, no duplicate prompt-gen job). The new IT file (`test/it/shared/tests/activate-brand-for-org.js`) only exercises missing-body/404/403 auth cases - it never activates a real pending brand. The "DRS unconfigured in the harness" note only excuses the prompt-gen/schedule side-effects; the **promotion path** (`updateBrand`/`setBrandStatus` writing `baseSiteId`/`status`, the `ensure_primary_brand_site` trigger creating the `base` row, and the `brands_base_site_unique` -> 409) is pure Postgres and is testable now without DRS. Recommend seeding one pending-brand + onboarded-site fixture and asserting the full promote path + the 409. (Unit coverage in `test/controllers/brands.test.js` is strong and non-tautological - this is specifically the integration layer being thinner than the spec asks.)

3. **[Important] The two-step-promote code comment is factually incorrect.** The anchor-then-promote sequence (`updateBrand({ baseSiteId })` while pending, then `setBrandStatus({ status:'active' })`) is functionally safe and recoverable on retry via the `alreadyAnchored` path - no objection to the behaviour. But the comment justifying it states that `updateBrand` "deliberately refuses status changes (#2637), so every status transition routes through setBrandStatus." As merged, `updateBrand` (`brands-storage.js`) refuses only **demotions** (active->pending); a combined `updateBrand({ baseSiteId, status:'active' })` on a pending, unanchored brand satisfies the active-has-site guard and works in one atomic write - which is exactly what the spec (Backend section 2, lines 122-128) asked for. So the two-step introduces an unnecessary non-atomic window (anchor commits, promote fails -> pending-but-anchored) that the atomic single-write would not have, and it is defended by a non-existent constraint. Either collapse to the spec's single `updateBrand({ baseSiteId, status:'active' })`, or keep the two-step but rewrite the comment to state the real reason (a design preference to route status transitions through `setBrandStatus`).

## Non-blocking

4. **nit: the `brands_base_site_unique` 409 leaks the DB constraint name to the client.** On the anchor write, a 409 rejection flows to `createErrorResponse(error)`, which puts the raw `error.message` (the DB/internal constraint detail) into both the response body and the `x-error` header, while the OpenAPI 409 promises a "generic message." Consider remapping this path to a generic conflict message. `src/controllers/brands.js`.

Verified fixed (prior bot-blocking items, all genuinely present in code): already-active -> 200 no-op short-circuit, deleted -> 404, best-effort side-effects returning 200 on DRS failure with a `:warning:` ops alert, QUEUED (not just RUNNING) idempotency test, and an integration-test file present. The auth gate (membership + PAID via tier-equality, no admin claim) is correct and matches shipped Piece 1 - note the spec/test-plan is now stale on the admin requirement (it would 403 every paying customer), so update project-elmo-ui#2095 line 99 / 175 rather than the code.

igor-grubic:
Thanks @rainer-friederich — really thorough cross-repo read. Pushed `d6603368`:

**1. Outer-catch alert (blocking) — ✅ fixed.** Wrapped the failure-path `postLlmoAlert` in its own try/catch (identical to the inner best-effort block), so an alert rejection can no longer escape and replace the intended 4xx (e.g. the 409) with an unhandled-rejection 500. Added a test that forces the failure-path alert to throw and asserts the 409 survives.

**3. Two-step promote comment (blocking) — ✅ you're right, reverted.** Collapsed back to the spec's single atomic write `updateBrand({ status:'active', baseSiteId })`. `updateBrand` supports promote-to-active — it guards only active→pending demotions (#2637) — so the two-step was unnecessary, introduced the non-atomic anchor-committed-but-promote-failed window you flagged, and its justifying comment cited a constraint that doesn't exist. Good catch; gone.

**4. 409 message (nit) — no change needed.** The `updateBrand` 409 is already remapped to a clean, user-facing message (`"This site is already the primary URL for another brand"`), not the raw `brands_base_site_unique` constraint name, and the `/activate` OpenAPI 409 already describes exactly that. So code and spec already agree on this path — no constraint-name leak.

**Auth / spec staleness — agreed.** The code (membership + PAID, no admin) is correct and matches shipped Piece 1; I'll flag project-elmo-ui#2095 (lines 99 / 175) for correction rather than touching the code.

**2. Integration tests (blocking) — in progress.** You're right that the promotion path is pure Postgres and testable now without DRS. I'm seeding a pending-brand + onboarded-site fixture under a PAID org and asserting the real promote (→ `active` + `baseSiteId` + base `brand_sites` row) plus the `brands_base_site_unique` → 409. Pushing in a follow-up commit.

rainer-friederich (approved):
LGTM

PR #2711, opened 2026-06-26
feat(llmo): stage-domain Cloudflare deploy + bump cloudflare-client 1.1.1

ssilare-adobe:
## Summary
- Allow Cloudflare deploy/route `targetHost` validation for stage domains that share the production site's registrable domain (sibling subdomains, e.g. `tokowaka.aem-screens.net` vs `frescopa.aem-screens.net`), matching stage-domain onboarding rules in the UI.
- Resolve worker script name and Edge Optimize API key from the deploy target host via `deployWorkerBaseURL` so stage hosts get their own worker name and staging metaconfig key.
- Bump `@adobe/spacecat-shared-cloudflare-client` to `1.1.1`.

Pairs with [project-elmo-ui#2173](https://github.com/adobe/project-elmo-ui/pull/2173) (stage-domain Cloudflare wizard).

## Test plan
- [x] `npm test -- --grep llmo-cloudflare` (utils + controller tests pass)
- [ ] Deploy worker from stage-domain wizard with a sibling subdomain host — expect 200 (not `targetHost must belong to the site's domain`)
- [ ] Deploy worker for production host — unchanged behavior
- [ ] Add route for stage host pattern — worker name matches stage-specific script

Made with [Cursor](https://cursor.com)

PR #2704, opened 2026-06-26
chore(deps): bump @adobe/spacecat-shared-cloudflare-client to 1.1.1

ssilare-adobe:
## Summary
- Bumps `@adobe/spacecat-shared-cloudflare-client` from `1.1.0` to `1.1.1` in `package.json` and `package-lock.json`.
- Picks up the fix from [spacecat-shared#1736](https://github.com/adobe/spacecat-shared/pull/1736): worker existence check uses **GET** instead of **HEAD** (Cloudflare returns **405** for HEAD on `/workers/scripts/{name}`), which was blocking Cloudflare BYOCDN worker deploy.

## Test plan
- [ ] CI passes
- [ ] After deploy to dev/ci, retry Cloudflare worker deploy from the onboarding wizard and confirm it no longer fails with 405

Made with [Cursor](https://cursor.com)

PR #2712, opened 2026-06-26
feat(llmo): support stage-domain Cloudflare deploy and routing

ssilare-adobe:
## Summary
- Allow Cloudflare deploy/route `targetHost` validation for stage domains that share the production site's registrable domain (sibling subdomains, e.g. `tokowaka.aem-screens.net` vs `frescopa.aem-screens.net`), matching stage-domain onboarding in the UI.
- Resolve worker script name and Edge Optimize API key from the deploy target host via `deployWorkerBaseURL` so stage hosts get their own worker name and staging metaconfig key.

Replaces #2711 (cloudflare-client 1.1.1 bump is already on main via #2704).

Pairs with [project-elmo-ui#2173](https://github.com/adobe/project-elmo-ui/pull/2173).

## Test plan
- [x] `llmo-cloudflare` utils + controller unit tests pass
- [ ] Deploy worker from stage-domain wizard with a sibling subdomain host — expect 200 (not `targetHost must belong to the site's domain`)
- [ ] Deploy worker for production host — unchanged behavior
- [ ] Add route for stage host pattern — worker name matches stage-specific script

### DIFF
--- a/src/controllers/llmo/llmo-cloudflare-utils.js
+++ b/src/controllers/llmo/llmo-cloudflare-utils.js
@@ -16,6 +16,8 @@
  * unit-tested in isolation.
  */
 
+import { getDomain } from 'tldts';
+
 const WORKER_NAME_PREFIX = 'edge-optimize-router';
 const CF_MAX_SCRIPT_NAME_LEN = 63; // Cloudflare worker script name max length
 
@@ -35,6 +37,23 @@ export const deriveWorkerName = (baseURL) => {
   return `${WORKER_NAME_PREFIX}-${slug}`.slice(0, CF_MAX_SCRIPT_NAME_LEN).replace(/-+$/g, '');
 };
 
+/** Canonical hostname (no leading www, lowercased). */
+export const canonicalHostname = (host) => host.replace(/^www\./i, '').toLowerCase();
+
+/**
+ * Base URL used to derive the worker name and resolve the Edge Optimize API key for deploy.
+ * Production target hosts reuse the site base URL; staging / other subdomains use https://{targetHost}/.
+ */
+export const deployWorkerBaseURL = (siteBaseURL, targetHost) => {
+  const siteHost = canonicalHostname(new URL(siteBaseURL).hostname);
+  const targetHostOnly = targetHost.trim().replace(/^https?:\/\//i, '').split('/')[0];
+  const targetCanon = canonicalHostname(targetHostOnly);
+  if (targetCanon === siteHost) {
+    return siteBaseURL;
+  }
+  return `https://${targetHostOnly}/`;
+};
+
 /**
  * Whether `host` belongs to the onboarded site's domain: it must equal the site's canonical
  * host (base URL host minus leading "www.") or be a subdomain of it. Prevents pointing the
@@ -45,6 +64,26 @@ export const hostInSiteDomain = (host, baseURL) => {
   const h = host.toLowerCase();
   return h === siteHost || h.endsWith(`.${siteHost}`);
 };
+
+/**
+ * Whether `host` shares the same registrable domain as the production site (matches stage
+ * domain onboarding rules — sibling subdomains on the same zone).
+ */
+export const hostSharesSiteRegistrableDomain = (host, baseURL) => {
+  try {
+    const prodDomain = getDomain(new URL(baseURL).hostname);
+    const hostOnly = host.replace(/^https?:\/\//i, '').split('/')[0];
+    const targetDomain = getDomain(hostOnly);
+    return Boolean(prodDomain && targetDomain && prodDomain === targetDomain);
+  } catch {
+    return false;
+  }
+};
+
+/** Accept production subdomains or registrable-domain siblings (stage hosts). */
+export const isCloudflareTargetHostAllowed = (host, baseURL) => (
+  hostInSiteDomain(host, baseURL) || hostSharesSiteRegistrableDomain(host, baseURL)
+);
 
 /**
  * Extracts the host from a Cloudflare route pattern (e.g. "*.example.com/path*" -> "example.com",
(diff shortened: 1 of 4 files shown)


src/controllers/llmo/llmo-cloudflare-utils.js
```diff
@@ -35,6 +37,23 @@ export const deriveWorkerName = (baseURL) => {
...
+ * Production target hosts reuse the site base URL; staging / other subdomains use https://{targetHost}/.
+ */
+export const deployWorkerBaseURL = (siteBaseURL, targetHost) => {
+  const siteHost = canonicalHostname(new URL(siteBaseURL).hostname);
+  const targetHostOnly = targetHost.trim().replace(/^https?:\/\//i, '').split('/')[0];
+  const targetCanon = canonicalHostname(targetHostOnly);
+  if (targetCanon === siteHost) {
+    return siteBaseURL;
+  }
+  return `https://${targetHostOnly}/`;
```
MysticatBot, line 54: issue (blocking): `targetHostOnly` preserves the original casing from the caller's input. If a client sends `targetHost: "Staging.Example.Com"`, this returns `https://Staging.Example.Com/` which becomes the `metaconfigBaseURL` passed to `fetchMetaconfig` and the input to `deriveWorkerName`. If either downstream call is case-sensitive on the URL, the lookup will fail for mixed-case input that is otherwise valid.

Fix: use `targetCanon` (already lowercased) in the constructed URL, or explicitly lowercase `targetHostOnly` before building the URL:
```js
return `https://${targetHostOnly.toLowerCase()}/`;
```

test/controllers/llmo/llmo-cloudflare-utils.test.js
```diff
@@ -75,4 +97,78 @@ describe('llmo-cloudflare-utils', () => {
...
+  describe('deployWorkerBaseURL', () => {
+    it('reuses the site base URL for the production host (with www)', () => {
+      expect(deployWorkerBaseURL('https://www.example.com', 'www.example.com')).to.equal('https://www.example.com');
+    });
+
+    it('reuses the site base URL for the canonical production host (no www)', () => {
+      expect(deployWorkerBaseURL('https://www.example.com', 'example.com')).to.equal('https://www.example.com');
+    });
+
+    it('uses https://{targetHost}/ for a subdomain that is not the production host', () => {
```
MysticatBot, line 110: issue (blocking): This suite has only positive assertions (both tests assert `true`). Add at least one negative case proving unrelated domains are rejected:
```js
it('rejects hosts on a different registrable domain', () => {
  expect(isCloudflareTargetHostAllowed('evil.com', 'https://www.example.com')).to.equal(false);
});
```

Also: the controller-level test covers `deployWorker` with a stage host but there is no matching test for `addRoute` with a stage-domain route pattern. Add one to confirm the route-creation path also accepts and correctly resolves stage domains.

src/controllers/llmo/llmo-cloudflare-utils.js
```diff
@@ -46,6 +65,26 @@ export const hostInSiteDomain = (host, baseURL) => {
...
+    const hostOnly = host.replace(/^https?:\/\//i, '').split('/')[0];
+    const targetDomain = getDomain(hostOnly);
+    return Boolean(prodDomain && targetDomain && prodDomain === targetDomain);
+  } catch {
+    return false;
+  }
+};
+
+/** Accept production subdomains or registrable-domain siblings (stage hosts). */
+export const isCloudflareTargetHostAllowed = (host, baseURL) => (
```
MysticatBot, line 84: question: This now accepts any host sharing the same registrable domain via `tldts` PSL data. The trust assumption is that registrable-domain siblings are under the same ownership. For shared-hosting domains missing from the PSL, two unrelated tenants could share a registrable domain. Is there an upstream check (Cloudflare zone membership, site-model allowlist, or UI-level constraint from the companion project-elmo-ui PR) that provides defense-in-depth here?

MysticatBot (changes requested):
Hey @ssilare-adobe,

**Verdict: Request changes** - two findings need attention before merge.
**Complexity: HIGH** - medium diff touching API controller layer.
_Changes: Widens Cloudflare deploy/route targetHost validation to accept stage-domain siblings sharing the same registrable domain, and resolves worker name + API key from the deploy target host (4 files)._

## Must fix before merge
1. [Important] Mixed-case `targetHost` produces a non-normalized metaconfig lookup URL - `src/controllers/llmo/llmo-cloudflare-utils.js:54` (details inline)
2. [Important] No negative test for `isCloudflareTargetHostAllowed`; no `addRoute` controller test for stage domains - `test/controllers/llmo/llmo-cloudflare-utils.test.js:110` (details inline)

<details>

- nit: `hostSharesSiteRegistrableDomain` bare `catch` returns `false` without any logging, making stage-domain rejections hard to debug in production - `src/controllers/llmo/llmo-cloudflare-utils.js:78`
- suggestion: Extract `routePatternHost(pattern)` to a local const before the validation check to avoid calling it twice (divergence risk if parsing ever changes) - `src/controllers/llmo/llmo-cloudflare.js:365`
- nit: Inconsistent trailing slash: production path returns `siteBaseURL` as-is (may lack trailing `/`), stage path always appends `/` - if `fetchMetaconfig` or `deriveWorkerName` are slash-sensitive this would produce different behavior per branch - `src/controllers/llmo/llmo-cloudflare-utils.js:52-54`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 4m 2s | <b>Cost:</b> $6.66 | <b>Commit:</b> <code>66ddce7c0798e6b73d917633a266ccd2f6664f44</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ssilare-adobe:
Not required

PR #2713, opened 2026-06-26
feat(preflight): shape new DTOs and join AsyncJob on detail (SITES-47254)

GeezerPelletier:
## Summary

First PR for [SITES-47254](https://jira.corp.adobe.com/browse/SITES-47254) — isolated to spacecat-api-service. Shapes the response model for the site-scoped Preflight endpoints (`POST/GET /sites/{siteId}/preflights` and `GET /sites/{siteId}/preflights/{preflightId}`) and shifts the source of truth for `result`/`error` off the (soon-to-be-removed) duplicated columns on `preflights` onto the joined `async_jobs` row.

The storage migration that drops `result` / `error` / `started_at` from the `preflights` table is a follow-up PR in `mysticat-data-service` — once it lands, this controller code keeps working unchanged because `result`/`error` already read from `async_jobs` and `startedAt` is no longer surfaced on the wire at all.

### `PreflightCreated` — POST 202 body (just-created identity + state)

`updatedAt` and `endedAt` are deliberately omitted — at creation time `updatedAt === createdAt` and `endedAt` is always `null`. Subsequent GETs return the full shapes below.

```
{
  preflightId, siteId, status, url, createdAt,
  createdBy: { email, displayName? }
}
```

### `PreflightListItem` — GET list body — fields *added* vs today: `siteId`, `updatedAt`, `endedAt`

```
{
  preflightId, siteId, status, url,
  createdAt, updatedAt, endedAt,
  createdBy: { email, displayName? }
}
```

All fields read from the `Preflight` entity directly — list path stays index-only, no join.

### `PreflightDetail` — GET single body — fields *added* vs today: `siteId`; fields *removed* vs today: `startedAt`

```
{
  preflightId, siteId, status, url,
  createdAt, updatedAt, endedAt,
  createdBy: { email, displayName? },
  result, error
}
```

Source split:
- `status`, `endedAt` ← `Preflight` (denormalized cache, kept in sync by the projector)
- `result`, `error` ← joined `AsyncJob` (truth)

`asyncJobId`, `scanId`, and `startedAt` are deliberately *not* surfaced — `asyncJobId`/`scanId` stay internal correlation; `startedAt` is an AsyncJob concern, not a Preflight attribute (consumers that need scan timing can read it out of `result`). Content typing of `result` is left passthrough (`object | null`) and will be tackled separately under [SITES-47180](https://jira.corp.adobe.com/browse/SITES-47180).

## Changes

- **`src/dto/preflight.js`** — `// @ts-check` + JSDoc `@typedef PreflightCreated | PreflightListItem | PreflightDetail | PreflightError` pinning the three wire shapes. New `toCreatedJSON(preflight)` for the POST 202 body. `toJSON` (list) extended with `siteId` / `updatedAt` / `endedAt`. `toDetailJSON(preflight, asyncJob)` adds `siteId`, removes `startedAt`, sources `result` / `error` from the AsyncJob (falling back to `null` when missing).
- **`src/controllers/preflight.js`** — `createPreflight` switched to `toCreatedJSON`; `Location` header unchanged. `getPreflightById` fetches the joined AsyncJob via `preflight.getAsyncJob()`. On a thrown fetch (replica lag, transient DB blip) it logs a warning and passes `null` to the DTO; the response still returns 200 with the Preflight-sourced fields populated and `result`/`error` surfaced as `null`. The MFE polls `Preflight.status`, which is sourced from the `Preflight` row itself — degraded join paths never block termination signals.
- **`test/dto/preflight.test.js`** — new file: list / detail / created happy paths, FAILED with structured error, degraded null AsyncJob, exact-key pin tests (catch field-leakage drift), confirms `asyncJobId` / `scanId` / `startedAt` are never on the wire.
- **`test/controllers/preflight.test.js`** — `mockPreflight.getAsyncJob` added and reset every test; the existing createPreflight happy-path test now asserts the new `PreflightCreated` shape (`siteId`/`createdAt`/`createdBy` present; `updatedAt`/`endedAt` absent); the existing detail test asserts `siteId` is present and `startedAt`/`asyncJobId`/`scanId` are absent; two new tests cover the AsyncJob source split for `result`/`error` and the degraded-fetch path; the list test locks in `siteId`/`updatedAt`/`endedAt` presence and internal-field absence.

## Test plan

- [x] `npm run lint` on touched files — clean
- [x] `npx tsc -p tsconfig.json` (`// @ts-check` opt-in gate) — clean
- [x] `npx mocha test/dto/preflight.test.js test/controllers/preflight.test.js` — **112 passing, 0 failing**
- [ ] Manual: hit `POST /v1/sites/{siteId}/preflights` and verify the 202 body matches `PreflightCreated` and the `Location` header points at the detail URL.
- [ ] Manual: hit `GET /v1/sites/{siteId}/preflights/{preflightId}` on dev after deploy and verify `siteId` is present, `startedAt` is gone, and `result`/`error` reflect the AsyncJob row.
- [ ] Manual: hit `GET /v1/sites/{siteId}/preflights` and verify `siteId`, `updatedAt`, `endedAt` populate per row.

## OpenAPI

Deliberately deferred — the site-scoped preflight endpoints aren't in `docs/openapi/` yet, and we don't want to surface the contract publicly until the upstream chain (SITES-47253 + the planned storage migration) is fully verified end-to-end. OpenAPI alignment will land alongside the deeper response-shape work in SITES-47180.

## Related

- [SITES-47254](https://jira.corp.adobe.com/browse/SITES-47254) — this ticket; remaining piece is the data-service migration that drops `result`/`error`/`started_at` from `preflights` (follow-up PR in `mysticat-data-service`).
- [SITES-47253](https://jira.corp.adobe.com/browse/SITES-47253) — projector PATCH fix that unblocks `Preflight.status` flipping to terminal state ([mysticat-projector-service#216](https://github.com/adobe/mysticat-projector-service/pull/216)).
- [SITES-47180](https://jira.corp.adobe.com/browse/SITES-47180) — deeper exercise on response-shape typing (especially `result` content) + OpenAPI alignment.
- [SITES-45776](https://jira.corp.adobe.com/browse/SITES-45776) — projector completion path that originally duplicated lifecycle fields onto `preflights`.

### DIFF
--- a/src/controllers/preflight.js
+++ b/src/controllers/preflight.js
@@ -750,16 +750,24 @@ function PreflightController(ctx, log, env) {
         return preflightError('PREFLIGHT_NOT_FOUND', `Preflight with ID ${preflightId} not found`, 404);
       }
 
-      // Lifecycle truth (startedAt, result, error) lives on async_jobs.
-      // Fetch the joined row; on a missing/fetch-failed join, surface the
-      // three fields as null rather than 404'ing — the rest of the preflight
-      // is still a valid response, and the MFE polls Preflight.status which
-      // is sourced from the preflight row itself.
-      let asyncJob = null;
+      // Lifecycle truth (status, endedAt, result, error) lives on async_jobs;
+      // toDetailJSON sources those from the joined row so a polling client
+      // sees terminal status alongside terminal result (no split-brain across
+      // the projector's two-write window).
+      //
+      // On a transient fetch failure (PostgREST 5xx / network blip) return
+      // 503 rather than 200-with-nulls: a 200 + `result: null` is
+      // indistinguishable on the wire from a legitimately empty completed
+      // scan, and a polling client would cache the wrong terminal answer.
+      // A null AsyncJob (no row yet — transitional/legacy flow) is NOT an
+      // error; lifecycle fields fall back to the Preflight cache.
+      let asyncJob;
       try {
         asyncJob = await preflight.getAsyncJob();
       } catch (e) {
-        log.warn(`Failed to fetch AsyncJob for preflight ${preflightId}: ${e.message}`);
+        const msg = e?.message ?? String(e);
+        log.warn(`Failed to fetch AsyncJob for preflight ${preflightId}: ${msg}`);
+        return preflightError('PREFLIGHT_LIFECYCLE_UNAVAILABLE', 'Failed to load preflight lifecycle data', 503);
       }
 
       return ok(PreflightDto.toDetailJSON(preflight, asyncJob));
--- a/src/controllers/preflight.js
+++ b/src/controllers/preflight.js
@@ -674,7 +674,7 @@ function PreflightController(ctx, log, env) {
     const locationUrl = `https://spacecat.experiencecloud.live/api/${isDev ? 'ci' : 'v1'}`
       + `/sites/${siteId}/preflights/${preflight.getId()}`;
 
-    return createResponse(PreflightDto.toJSON(preflight), 202, { Location: locationUrl });
+    return createResponse(PreflightDto.toCreatedJSON(preflight), 202, { Location: locationUrl });
   };
 
   /**
--- a/src/controllers/preflight.js
+++ b/src/controllers/preflight.js
@@ -638,7 +638,8 @@ function PreflightController(ctx, log, env) {
         url,
         status: AsyncJob.Status.IN_PROGRESS,
         createdBy,
-        startedAt: new Date().toISOString(),
+        // SITES-47254: startedAt is an AsyncJob concern (set automatically on
+        // AsyncJob.create); the Preflight row no longer carries it.
       });
     } catch (e) {
       log.error(`Failed to create Preflight: ${e.message}`);
@@ -657,16 +658,20 @@ function PreflightController(ctx, log, env) {
       );
     } catch (mysticatError) {
       log.error(`Mysticat analyze failed for preflight ${preflight.getId()}: ${mysticatError.message}`);
-      preflight.setStatus(AsyncJob.Status.FAILED);
-      // Stored error message mirrors the external 502 response — the raw
-      // upstream body could carry internal hostnames / stack traces and is
-      // exposed via GET /sites/:siteId/preflights/:preflightId. Full detail
-      // is in the log.error above for ops visibility.
-      preflight.setError({ code: 'MYSTICAT_ERROR', message: 'Upstream analyze service failed' });
-      preflight.setEndedAt(new Date().toISOString());
-      await preflight.save().catch((e) => log.warn(`Failed to persist FAILED state on preflight ${preflight.getId()}: ${e.message}`));
+      const endedAt = new Date().toISOString();
+      // SITES-47254: error lives on AsyncJob only (source of truth); the
+      // Preflight row holds status + ended_at as a cache. Stored message
+      // mirrors the external 502 response — the raw upstream body could
+      // carry internal hostnames / stack traces and is exposed via
+      // GET /sites/:siteId/preflights/:preflightId. Full detail is in the
+      // log.error above for ops visibility.
       asyncJob.setStatus(AsyncJob.Status.FAILED);
+      asyncJob.setError({ code: 'MYSTICAT_ERROR', message: 'Upstream analyze service failed' });
+      asyncJob.setEndedAt(endedAt);
       await asyncJob.save().catch((e) => log.warn(`Failed to persist FAILED state on AsyncJob ${asyncJob.getId()}: ${e.message}`));
+      preflight.setStatus(AsyncJob.Status.FAILED);
+      preflight.setEndedAt(endedAt);
+      await preflight.save().catch((e) => log.warn(`Failed to persist FAILED state on preflight ${preflight.getId()}: ${e.message}`));
       return preflightError('PREFLIGHT_UPSTREAM_ERROR', 'Upstream analyze service failed', 502);
     }
 
--- a/src/controllers/preflight.js
+++ b/src/controllers/preflight.js
@@ -745,7 +745,19 @@ function PreflightController(ctx, log, env) {
         return preflightError('PREFLIGHT_NOT_FOUND', `Preflight with ID ${preflightId} not found`, 404);
       }
 
-      return ok(PreflightDto.toDetailJSON(preflight));
+      // Lifecycle truth (startedAt, result, error) lives on async_jobs.
+      // Fetch the joined row; on a missing/fetch-failed join, surface the
+      // three fields as null rather than 404'ing — the rest of the preflight
+      // is still a valid response, and the MFE polls Preflight.status which
+      // is sourced from the preflight row itself.
+      let asyncJob = null;
+      try {
+        asyncJob = await preflight.getAsyncJob();
+      } catch (e) {
+        log.warn(`Failed to fetch AsyncJob for preflight ${preflightId}: ${e.message}`);
+      }
+
+      return ok(PreflightDto.toDetailJSON(preflight, asyncJob));
     } catch (e) {
       log.error(`Failed to fetch preflight ${preflightId}: ${e.message}`);
       return preflightError('PREFLIGHT_INTERNAL_ERROR', 'Failed to fetch preflight', 500);
--- a/src/dto/preflight.js
+++ b/src/dto/preflight.js
@@ -9,26 +9,85 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
+// @ts-check
+
+/**
+ * Wire-contract typedefs for the site-scoped Preflight GET endpoints. The
+ * MFE polls `status` and consumes `result`/`error`. Field-content typing
+ * (esp. the shape of `result`) is the deeper exercise tracked in
+ * SITES-47180; passthrough `object` is intentional for now.
+ *
+ * @typedef {'IN_PROGRESS' | 'COMPLETED' | 'FAILED' | 'CANCELLED'} PreflightStatus
+ *
+ * @typedef {{ email: string, displayName?: string }} PreflightActor
+ *
+ * @typedef {Object} PreflightListItem
+ * @property {string} preflightId
+ * @property {string} siteId
+ * @property {PreflightStatus} status
+ * @property {string} url
+ * @property {string} createdAt              ISO 8601
+ * @property {string} updatedAt              ISO 8601
+ * @property {string | null} endedAt         ISO 8601, null while not terminal
+ * @property {PreflightActor} createdBy
+ *
+ * @typedef {Object} PreflightDetail
+ * @property {string} preflightId
+ * @property {string} siteId
+ * @property {PreflightStatus} status
+ * @property {string} url
+ * @property {string} createdAt              ISO 8601
+ * @property {string} updatedAt              ISO 8601
+ * @property {string | null} startedAt       ISO 8601, sourced from AsyncJob
+ * @property {string | null} endedAt         ISO 8601
+ * @property {PreflightActor} createdBy
+ * @property {object | null} result          Sourced from AsyncJob; null while not terminal
+ * @property {PreflightError | null} error   Sourced from AsyncJob
+ *
+ * @typedef {{ code: string, message: string, details?: object }} PreflightError
+ */
 
 export const PreflightDto = {
+  /**
+   * List-view DTO. Sources entirely from the Preflight entity — `status` and
+   * `endedAt` are denormalized on the row (kept in sync by the projector) so
+   * the list path stays index-only without joining `async_jobs`.
+   *
+   * @param {import('@adobe/spacecat-shared-data-access').Preflight} preflight
+   * @returns {PreflightListItem}
+   */
   toJSON: (preflight) => ({
     preflightId: preflight.getId(),
-    status: preflight.getStatus(),
+    siteId: preflight.getSiteId(),
+    status: /** @type {PreflightStatus} */ (preflight.getStatus()),
     url: preflight.getUrl(),
     createdAt: preflight.getCreatedAt(),
+    updatedAt: preflight.getUpdatedAt(),
+    endedAt: preflight.getEndedAt(),
     createdBy: preflight.getCreatedBy(),
   }),
 
-  toDetailJSON: (preflight) => ({
+  /**
+   * Detail-view DTO. Lifecycle truth (`startedAt`, `result`, `error`) lives
+   * on `async_jobs`; the caller fetches the joined AsyncJob and passes it in.
+   * When `asyncJob` is null (defensive degrade — caller logs the gap), those
+   * three fields surface as null rather than 404'ing the whole response.
+   *
+   * @param {import('@adobe/spacecat-shared-data-access').Preflight} preflight
+   * @param {import('@adobe/spacecat-shared-data-access').AsyncJob | null} asyncJob
+   * @returns {PreflightDetail}
+   */
+  toDetailJSON: (preflight, asyncJob) => ({
     preflightId: preflight.getId(),
-    status: preflight.getStatus(),
+    siteId: preflight.getSiteId(),
+    status: /** @type {PreflightStatus} */ (preflight.getStatus()),
     url: preflight.getUrl(),
     createdAt: preflight.getCreatedAt(),
-    createdBy: preflight.getCreatedBy(),
     updatedAt: preflight.getUpdatedAt(),
-    startedAt: preflight.getStartedAt(),
+    startedAt: asyncJob ? asyncJob.getStartedAt() : null,
     endedAt: preflight.getEndedAt(),
-    result: preflight.getResult(),
-    error: preflight.getError(),
+    createdBy: preflight.getCreatedBy(),
+    result: asyncJob ? asyncJob.getResult() : null,
+    error: asyncJob ? asyncJob.getError() : null,
   }),
 };
--- a/src/dto/preflight.js
+++ b/src/dto/preflight.js
@@ -38,7 +38,6 @@
  * @property {string} url
  * @property {string} createdAt              ISO 8601
  * @property {string} updatedAt              ISO 8601
- * @property {string | null} startedAt       ISO 8601, sourced from AsyncJob
  * @property {string | null} endedAt         ISO 8601
  * @property {PreflightActor} createdBy
  * @property {object | null} result          Sourced from AsyncJob; null while not terminal
@@ -68,10 +67,15 @@ export const PreflightDto = {
   }),
 
   /**
-   * Detail-view DTO. Lifecycle truth (`startedAt`, `result`, `error`) lives
-   * on `async_jobs`; the caller fetches the joined AsyncJob and passes it in.
-   * When `asyncJob` is null (defensive degrade — caller logs the gap), those
-   * three fields surface as null rather than 404'ing the whole response.
+   * Detail-view DTO. `result` and `error` are AsyncJob-owned (the projector
+   * writes them there, not on the Preflight row), so the caller fetches the
+   * joined AsyncJob and passes it in. When `asyncJob` is null (defensive
+   * degrade — caller logs the gap), the two fields surface as null rather
+   * than 404'ing the whole response.
+   *
+   * `startedAt` is intentionally not surfaced — it's an AsyncJob concern,
+   * not a Preflight attribute. Consumers that need timing internals can
+   * read them out of `result`.
    *
    * @param {import('@adobe/spacecat-shared-data-access').Preflight} preflight
    * @param {import('@adobe/spacecat-shared-data-access').AsyncJob | null} asyncJob
@@ -84,7 +88,6 @@ export const PreflightDto = {
     url: preflight.getUrl(),
     createdAt: preflight.getCreatedAt(),
     updatedAt: preflight.getUpdatedAt(),
-    startedAt: asyncJob ? asyncJob.getStartedAt() : null,
     endedAt: preflight.getEndedAt(),
     createdBy: preflight.getCreatedBy(),
     result: asyncJob ? asyncJob.getResult() : null,
--- a/src/dto/preflight.js
+++ b/src/dto/preflight.js
@@ -21,6 +21,14 @@
  *
  * @typedef {{ email: string, displayName?: string }} PreflightActor
  *
+ * @typedef {Object} PreflightCreated
+ * @property {string} preflightId
+ * @property {string} siteId
+ * @property {PreflightStatus} status
+ * @property {string} url
+ * @property {string} createdAt              ISO 8601
+ * @property {PreflightActor} createdBy
+ *
  * @typedef {Object} PreflightListItem
  * @property {string} preflightId
  * @property {string} siteId
@@ -47,6 +55,24 @@
  */
 
 export const PreflightDto = {
+  /**
+   * Just-created DTO for the 202 response on POST. Omits `updatedAt` and
+   * `endedAt` because they carry no information at creation time
+   * (updatedAt === createdAt; endedAt is always null for IN_PROGRESS).
+   * The full shape returns on subsequent GETs.
+   *
+   * @param {import('@adobe/spacecat-shared-data-access').Preflight} preflight
+   * @returns {PreflightCreated}
+   */
+  toCreatedJSON: (preflight) => ({
+    preflightId: preflight.getId(),
+    siteId: preflight.getSiteId(),
+    status: /** @type {PreflightStatus} */ (preflight.getStatus()),
+    url: preflight.getUrl(),
+    createdAt: preflight.getCreatedAt(),
+    createdBy: preflight.getCreatedBy(),
+  }),
+
   /**
    * List-view DTO. Sources entirely from the Preflight entity — `status` and
    * `endedAt` are denormalized on the row (kept in sync by the projector) so
--- a/src/dto/preflight.js
+++ b/src/dto/preflight.js
@@ -34,6 +34,13 @@
  *
  * @typedef {{ code: string, message: string, details?: object }} PreflightError
  *
+ * Entity getter return types use `| undefined` (not `| null`) for nullable
+ * columns because `spacecat-shared`'s `normalizeModelValue` maps DB NULL →
+ * `undefined`. Encoding that here lets tsc flag missing `?? null`
+ * coercions at the wire boundary — without it, `JSON.stringify` would
+ * silently drop those keys and violate the `string | null` /
+ * `object | null` wire contract.
+ *
  * @typedef {Object} PreflightEntity
  * @property {() => string} getId
  * @property {() => string} getSiteId
@@ -41,14 +48,14 @@
  * @property {() => string} getUrl
  * @property {() => string} getCreatedAt
  * @property {() => string} getUpdatedAt
- * @property {() => (string | null)} getEndedAt
+ * @property {() => (string | undefined)} getEndedAt
  * @property {() => PreflightActor} getCreatedBy
  *
  * @typedef {Object} AsyncJobEntity
  * @property {() => string} getStatus
- * @property {() => (string | null)} getEndedAt
- * @property {() => (object | null)} getResult
- * @property {() => (PreflightError | null)} getError
+ * @property {() => (string | undefined)} getEndedAt
+ * @property {() => (object | undefined)} getResult
+ * @property {() => (PreflightError | undefined)} getError
  *
  * @typedef {Object} PreflightCreated
  * @property {string} preflightId
@@ -102,7 +109,10 @@ export const PreflightDto = {
   toJSON: (preflight) => ({
     ...PreflightDto.toCreatedJSON(preflight),
     updatedAt: preflight.getUpdatedAt(),
-    endedAt: preflight.getEndedAt(),
+    // Coerce DB NULL (returned as `undefined` by normalizeModelValue) to
+    // `null` so JSON.stringify keeps the key — `endedAt: undefined` would
+    // be silently dropped, violating the `string | null` wire contract.
+    endedAt: preflight.getEndedAt() ?? null,
   }),
 
   /**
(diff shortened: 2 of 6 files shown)


MysticatBot (changes requested):
Hey @GeezerPelletier,

**Verdict: Request changes** - one data-regression bug in the error path, plus two documented-convention gaps.
_Changes: Reshapes Preflight DTOs into three distinct wire contracts (Created/List/Detail) and shifts result/error source of truth to the joined AsyncJob row (4 files)._

## Must fix before merge

1. [Important] **Error payload lost for preflights that fail during creation** - `src/controllers/preflight.js:665`

   The `createPreflight` catch block calls `preflight.setError({ code: 'MYSTICAT_ERROR', message: 'Upstream analyze service failed' })` but only calls `asyncJob.setStatus(AsyncJob.Status.FAILED)` without a corresponding `asyncJob.setError(...)`. After this PR, `toDetailJSON` reads `error` from `asyncJob.getError()`, which will be `null`.

   Impact: if a caller POSTs a preflight, gets a 502, then later GETs the detail, the response will show `status: "FAILED"` but `error: null` - losing the structured error context.

   **Fix:** Add `asyncJob.setError({ code: 'MYSTICAT_ERROR', message: 'Upstream analyze service failed' })` in the catch block alongside the existing `asyncJob.setStatus(AsyncJob.Status.FAILED)` call (~line 668).

2. [Important] **Missing integration tests per CLAUDE.md rule** - `test/it/`

   CLAUDE.md states: "New or modified endpoints must include integration tests in `test/it/`." This PR modifies the response shapes of three site-scoped preflight endpoints but adds no integration tests. The unit tests are solid, but the IT suite validates the full request lifecycle against a real PostgreSQL database - the layer where the AsyncJob join behavior matters most.

   **Fix:** Add IT coverage per the CLAUDE.md checklist (seed IDs, seed data, shared test factory, wiring file). If deliberately deferred, document the gap explicitly with a follow-up ticket.

3. [Important] **OpenAPI specs not updated** - `docs/openapi/`

   CLAUDE.md rule: "OpenAPI First: Define API contract in docs/openapi/ before implementation."

   The PR description acknowledges this deferral with rationale (waiting for upstream chain verification, tracked as SITES-47180). The wire contract is test-locked via exact-key pin tests, which reduces drift risk. Flagging for traceability - the deferral is pragmatic given the upstream dependencies, but should be temporary.

<details>

- nit: `toJSON` rename changes the semantics of an exported symbol - consumers outside this file would silently get the new list shape instead of the old created shape. Verified only one import site exists, so safe here - `src/dto/preflight.js`
- suggestion: `asyncJob.getResult()` and `asyncJob.getError()` are passed through unfiltered - consider allowlisting fields when SITES-47180 lands to prevent accidental info disclosure - `src/dto/preflight.js:107`
- suggestion: Add a controller-level test where `mockPreflight.getAsyncJob.resolves(null)` (not rejects) to close the wiring gap for the "no linked job yet" case - `test/controllers/preflight.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 6m 8s | <b>Cost:</b> $5.25 | <b>Commit:</b> <code>b96be58a2cc987ae72e118579ced71e54e5e4558</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/controllers/preflight.js
```diff
@@ -745,7 +745,19 @@ function PreflightController(ctx, log, env) {
...
+      // Lifecycle truth (startedAt, result, error) lives on async_jobs.
+      // Fetch the joined row; on a missing/fetch-failed join, surface the
+      // three fields as null rather than 404'ing — the rest of the preflight
+      // is still a valid response, and the MFE polls Preflight.status which
+      // is sourced from the preflight row itself.
+      let asyncJob = null;
+      try {
+        asyncJob = await preflight.getAsyncJob();
+      } catch (e) {
+        log.warn(`Failed to fetch AsyncJob for preflight ${preflightId}: ${e.message}`);
```
clotton: **Blocker — silent 200 on infra failure**

When `getAsyncJob()` throws a real PostgREST/transport fault, this catch swallows the error and `toDetailJSON` returns `result: null, error: null` inside HTTP 200. A COMPLETED preflight then looks identical to a finished-but-empty scan on the wire — clients can't tell the difference and won't retry.

Suggested options:
- Add a `degraded: true` boolean to the response shape so callers know to retry
- Return 503 (retryable) instead of 200 when the join fails on a terminal preflight
- Source both `status` and `result`/`error` from AsyncJob (where the truth lands first) to avoid this split-brain entirely

Also: `e.message` is `undefined` for non-Error rejections — use `e?.message ?? String(e)` to ensure the warning always captures the actual reason.

src/controllers/preflight.js
```diff
@@ -745,7 +745,19 @@ function PreflightController(ctx, log, env) {
...
+      }
+
+      return ok(PreflightDto.toDetailJSON(preflight, asyncJob));
```
clotton, line 773: **Blocker — status/result split-brain during projector's two-write window**

`toDetailJSON` reads `status` from the Preflight row and `result`/`error` from the AsyncJob row. The projector (`writer.ts`) writes AsyncJob first, then Preflight, and tolerates the second write failing (relies on SQS redrive). In the gap between those two writes a GET returns `status: IN_PROGRESS` + a fully-populated `result`. Polling clients that key off `status` will spin indefinitely on a scan that's actually done.

The clean fix is to read `status` from AsyncJob too (where the terminal state lands first), so the two halves of the response always agree.

src/dto/preflight.js
```diff
@@ -9,26 +9,114 @@
...
+   *
+   * `startedAt` is intentionally not surfaced — it's an AsyncJob concern,
+   * not a Preflight attribute. Consumers that need timing internals can
+   * read them out of `result`.
+   *
+   * @param {import('@adobe/spacecat-shared-data-access').Preflight} preflight
+   * @param {import('@adobe/spacecat-shared-data-access').AsyncJob | null} asyncJob
+   * @returns {PreflightDetail}
+   */
+  toDetailJSON: (preflight, asyncJob) => ({
```
clotton, line 141: **Medium — `toDetailJSON` re-copies all 8 fields from `toJSON` verbatim**

The test constants already encode the right inheritance (`DETAIL_FIELDS = [...LIST_FIELDS, 'result', 'error']`) but the DTO doesn't follow it. A future field addition to the list shape requires two edits. The established pattern in this codebase (`plg-onboarding.js:51`) is to spread a sibling method:

```js
toDetailJSON: (preflight, asyncJob) => ({
  ...PreflightDto.toJSON(preflight),
  result: asyncJob?.getResult() ?? null,
  error: asyncJob?.getError() ?? null,
}),
```

Same opportunity exists in `toJSON` (could spread `toCreatedJSON` and add `updatedAt`/`endedAt`).

src/dto/preflight.js
```diff
@@ -9,26 +9,114 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
+// @ts-check
```
clotton, line 12: **Low — `// @ts-check` has no precedent in `src/dto/`**

No other file under `src/` uses `// @ts-check`. Every peer DTO (`audit.js`, `suggestion.js`, `opportunity.js`) opens with only the licence header. Without a `tsconfig` that explicitly includes this file, the checker runs in isolation and the type annotations aren't validated against the actual model methods. If this is an intentional pilot, document it; otherwise align with the existing DTO convention.

clotton (changes requested):
## Review: feat(preflight) shape new DTOs and join AsyncJob on detail (SITES-47254)

The DTO split and test coverage are well-structured. The field-key pin tests are exactly right for catching future drift, and the `toCreatedJSON`/`toJSON`/`toDetailJSON` naming cleanly separates the three wire contracts. Two design-level issues need resolution before merge, plus a few smaller concerns.

---

### CONFIRMED BLOCKERS

**1. Silent 200 on infra failure masks a completed scan as "no results" (controller, line 757)**

When `getAsyncJob()` throws due to a real PostgREST/network fault, the catch swallows the error, `asyncJob` stays `null`, and `toDetailJSON` returns `result: null, error: null` inside an HTTP 200. A COMPLETED preflight then looks identical to an empty scan on the wire — the MFE cannot distinguish "scan finished, 0 issues" from "scan finished but API couldn't load results." A transient DB blip during the result-fetch step will produce a silently wrong response that a polling client caches as successful.

The projector code (`writer.ts:76-96`) confirms this is real: it upserts `async_jobs` first, then `preflights`, and tolerates partial failures via SQS redrive. During the redrive window `getAsyncJob()` will succeed but `preflight.getStatus()` may still show `IN_PROGRESS` — or the inverse, where the Preflight row is COMPLETED but the AsyncJob fetch fails. Either way the two halves of the terminal state can legitimately disagree, and returning 200 with `result: null` is the wrong signal.

The degrade path is the right idea but needs a wire-distinguishable signal: either a `degraded: true` flag on the response, a 503 (retryable), or sourcing both `status` and `result`/`error` from the same row to eliminate the split-brain window.

**2. `status` (Preflight row) and `result`/`error` (AsyncJob row) can race during the projector's two-write commit (DTO, lines 113+119-120)**

The projector writes AsyncJob first (with COMPLETED status + result) and then Preflight (with COMPLETED status). It tolerates the second write failing and relies on SQS redrive. Between those two writes — or across a failed redrive — `GET /preflights/:id` returns `status: IN_PROGRESS` (stale Preflight row) alongside a populated `result` from the already-finished AsyncJob. Clients that poll `status` to decide when to render results will spin indefinitely on a scan that is actually done. Sourcing both fields from AsyncJob (where the terminal truth lands first) would close this race.

---

### PLAUSIBLE MEDIUM

**3. DTO copy-paste: `toDetailJSON` re-lists all 8 fields from `toJSON` verbatim (DTO, line 110)**

`toDetailJSON` copies all eight fields from `toJSON` instead of spreading it. The test constants already model the correct inheritance (`DETAIL_FIELDS = [...LIST_FIELDS, 'result', 'error']`), but the DTO doesn't reflect it. A future field addition to the list shape requires two edits to stay in sync. The pattern in `plg-onboarding.js:51` (`{ ...PlgOnboardingDto.toJSON(onboarding), extraField }`) is the established codebase idiom:

```js
toDetailJSON: (preflight, asyncJob) => ({
  ...PreflightDto.toJSON(preflight),
  result: asyncJob?.getResult() ?? null,
  error: asyncJob?.getError() ?? null,
}),
```

Same applies to `toJSON` spreading `toCreatedJSON`.

**4. `// @ts-check` and `@typedef` blocks have no precedent in `src/dto/` (DTO, line 12)**

`// @ts-check` appears on no other file under `src/` in this repo. Every peer DTO (`audit.js`, `suggestion.js`, `opportunity.js`) opens with only the licence header. Without a `tsconfig` that includes this file, the checker runs in isolation and the JSDoc types are not validated against the actual model methods. If this is intentional as a local opt-in, it should be noted in the PR; otherwise bring it to the existing DTO convention (per-function `@param`/`@returns` only, no module-level `// @ts-check`).

**5. `e.message` unsafe on non-Error rejections (controller, line 757)**

```js
catch (e) {
  log.warn(`Failed to fetch AsyncJob for preflight ${preflightId}: ${e.message}`);
}
```

If `getAsyncJob()` rejects with a non-Error value (string, plain object, or `null`), `e.message` is `undefined` and the warning logs `"...preflight <id>: undefined"` — the actual reason is silently lost on the one code path designed to explain why `result` came back null. `e?.message ?? String(e)` covers non-Error throws and costs nothing.

GeezerPelletier:
Thanks for the review. Addressed two of three blockers in `fab8a6b3`; pushing back on the third + the degraded-review-confidence preamble — context below.

---

### Blocker #1 — `createPreflight` catch path loses the error payload

Great catch — real data regression. Fixed by moving the error write to `asyncJob` (source of truth post-this-PR) and dropping the dead `preflight.setError(...)` call:

```js
// catch (mysticatError):
asyncJob.setStatus(AsyncJob.Status.FAILED);
asyncJob.setError({ code: 'MYSTICAT_ERROR', message: 'Upstream analyze service failed' });
asyncJob.setEndedAt(endedAt);
await asyncJob.save();

preflight.setStatus(AsyncJob.Status.FAILED);
preflight.setEndedAt(endedAt);                // cache mirrors async_jobs
await preflight.save();                       // error NOT written to preflights
```

Also took the opportunity to drop `startedAt: new Date().toISOString()` from the `Preflight.create` payload — forward-compat with [spacecat-shared#1740](https://github.com/adobe/spacecat-shared/pull/1740) which removes the attribute. `async_jobs.started_at` is set automatically on `AsyncJob.create`, so the data isn't lost.

Existing controller test for this path updated: asserts `mockJob.setError` is called with the structured payload and `mockJob.setEndedAt` fires, while `mockPreflight.setError` is asserted **not** to have been called.

### Blocker #2 — Missing integration tests in `test/it/`

This one I'd like to defer, with rationale:

- The existing `test/it/postgres/preflight.test.js` + `test/it/shared/tests/preflight.js` cover the legacy `/preflight/jobs/*` endpoint family. They have no site-scoped-endpoint cases today — meaning this would be **net-new IT scaffolding** (new shared test factory, new seed data for the `preflights` table, wiring file), not just an update to existing assertions.
- The unit tests are response-shape-pinned (every key locked via deep-equal + exact-key-set assertions) so wire drift fails fast at PR time.
- The endpoints in question have no production consumers other than us (confirmed up-thread), and the cross-service correctness will be re-verified end-to-end on dev once [mysticat-projector-service#216](https://github.com/adobe/mysticat-projector-service/pull/216) deploys.

Happy to file a follow-up ticket for IT coverage of the new site-scoped endpoints, scoped under the broader SITES-47180 typed-models work. That ticket already owns response-shape pinning + OpenAPI alignment, and adding IT to its plate makes the new contract end-to-end-verified rather than scattering coverage across two tickets. Let me know if you'd rather I write the IT now under this PR.

### Blocker #3 — OpenAPI deferral

Acknowledged and called out in the PR description. SITES-47180 is the umbrella for `docs/openapi/` alignment + `result` content typing. The wire contract is locked via response-shape pin tests in the meantime; adding the OpenAPI now would mean shipping a contract surface we know will churn under 47180.

### Degraded-review-confidence preamble

Spec doc is now real: amended **ADR-003** ([docs/decisions/003-preflight-entity-design.md](docs/decisions/003-preflight-entity-design.md)) with a SITES-47254 section capturing the source-of-truth shift, three-DTO surface, read/write paths, storage migrations, and deploy ordering. The ADR is now linked from this PR and serves as the spec-context anchor for the rest of the chain (mysticat-data-service#746, spacecat-shared#1740, mysticat-projector-service#216) — bot reviews on those PRs should be able to validate against it.

### CI

`ci / build`, `ci / it-postgres`, `type-check`, `codecov/patch`, `Mysticat / pr-review` were all green at `b96be58a`. The bot's "CI checks are currently failing" preamble was stale — review was submitted at 15:52 UTC, before all checks completed.

### Non-blocking nits

- `toJSON` rename — verified one import site, agreed safe; called out in the ADR amendment now.
- `result`/`error` passthrough — agreed; allowlisting belongs with the typed-models work under SITES-47180.
- `getAsyncJob.resolves(null)` test — added (in addition to the `.rejects()` one) to close the no-linked-job-yet wiring gap.

Tests: 113 passing locally (was 112, +1 for the null-AsyncJob case). Re-requesting review.

GeezerPelletier:
Thanks @clotton — both blockers were sharp catches. Both are real bugs introduced by my refactor, not pre-existing issues. Addressed all five findings in `bf598733` + `a2c8ebed`.

---

### Blocker #1 — Silent 200 on `getAsyncJob` failure

You're right that a 200 with `result: null` is wire-indistinguishable from a legitimately empty completed scan; a polling client would cache the wrong terminal answer.

Fixed: the controller now returns **`503 PREFLIGHT_LIFECYCLE_UNAVAILABLE`** when `getAsyncJob()` throws. Standard "transient infra failure, retry" signal — client doesn't cache, polling resumes. A null AsyncJob (legitimate transitional flow with no row yet) is still treated as a 200 with lifecycle falling back to the Preflight cache; that's distinct from a fetch failure.

Test coverage: `returns 503 when getAsyncJob throws (do not silently 200 with null lifecycle)` replaces the old "degrade-to-nulls" test, with an explicit comment about why 503 over 200.

### Blocker #2 — Status/result split-brain across the projector's two-write window

Also right. The projector writes `async_jobs` (terminal) first, then `preflights` (cache), and tolerates the second write failing via SQS redrive. With `status` from Preflight and `result` from AsyncJob, the gap lets `IN_PROGRESS` ship alongside a populated `result` — polling clients spin.

Fixed by sourcing **both `status` and `endedAt` from AsyncJob in the detail DTO**, so the terminal answer lands as a unit. List view still reads from the Preflight cache (no join, hot path). The source-of-truth split is now:

| Field | List | Detail |
|---|---|---|
| `status` | `Preflight` (cache) | `AsyncJob` (truth) |
| `endedAt` | `Preflight` (cache) | `AsyncJob` (truth) |
| `result`, `error` | — | `AsyncJob` |

Regression-guarded by a new DTO test (`AsyncJob.status wins over Preflight.status when they disagree (projector two-write window)`) that seeds the two rows with disagreeing states and asserts terminal wins.

### Medium #3 — DTO copy-paste

Adopted the `plg-onboarding.js` spread idiom:

```js
toJSON: (preflight) => ({
  ...PreflightDto.toCreatedJSON(preflight),
  updatedAt: preflight.getUpdatedAt(),
  endedAt: preflight.getEndedAt(),
}),

toDetailJSON: (preflight, asyncJob) => ({
  ...PreflightDto.toJSON(preflight),
  ...(asyncJob && {
    status: asyncJob.getStatus(),
    endedAt: asyncJob.getEndedAt(),
  }),
  result: asyncJob ? asyncJob.getResult() : null,
  error: asyncJob ? asyncJob.getError() : null,
}),
```

Future list-field additions automatically propagate to detail; AsyncJob-sourced overrides for `status`/`endedAt` ride on top.

### Medium #4 — `// @ts-check` was editor-only

Sharper catch than the comment suggested: this file wasn't in `tsconfig.json` `include`, so the pragma was IDE-only and the `type-check` blocking gate wasn't validating anything. Fixed by:

1. Adding `src/dto/preflight.js` to `tsconfig.json` `include` (per the ADR-005 opt-in pattern).
2. Replacing `import('@adobe/spacecat-shared-data-access').Preflight/.AsyncJob` (which doesn't actually resolve — spacecat-shared v3.79.1 doesn't re-export those interfaces from its top-level type entry) with local `PreflightEntity` / `AsyncJobEntity` typedefs that capture only the methods we consume.

`npx tsc -p tsconfig.json` now actually validates the DTO and is clean. Doc-comment on the file explains the opt-in.

### Medium #5 — `e?.message ?? String(e)`

Added. Plus a regression-guard test (`hardens warn log against non-Error throws`) that rejects with a plain `{statusCode: 503}` (via `callsFake` because `.rejects(x)` wraps non-Error values into Error instances) and asserts the warn does NOT contain `'undefined'`.

---

**ADR amendment updated** (`a2c8ebed`) to record the list-vs-detail source split and the 503 path, so future bot reviews on the chain see the round-2 design.

Final state: 115 passing locally, lint + type-check clean. Re-requesting review.

MysticatBot (approved):
Hey @GeezerPelletier,

**Verdict: Approve** - no new blocking issues found; prior error-payload bug is fixed and the ADR grounds the design.
_Changes: Splits Preflight DTO into three wire contracts (Created/List/Detail) and shifts result/error source of truth to the joined AsyncJob row (5 files)._
Note: Recommend a human read before merge - this change amends an architectural decision record (`docs/decisions/003-preflight-entity-design.md`). The bot review is a complement to, not a replacement for, a human read here.

<details>

- suggestion: `toJSON` is a misleading name for the list-specific DTO - JavaScript convention and this codebase's peer DTOs use `toJSON` for the canonical full representation. Consider `toListJSON` to parallel `toCreatedJSON` / `toDetailJSON` and prevent a future maintainer from expecting `result`/`error` on the return value - `src/dto/preflight.js:87`
- suggestion: `asyncJob.getResult()` is passed through unfiltered to the wire. The legacy endpoint does the same, so the attack surface is unchanged, but add a one-line comment noting the trust assumption (result content is owned by the projector and assumed wire-safe; SITES-47180 will add schema validation) - `src/dto/preflight.js:113`
- nit: `startedAt` removal from the detail response is a field subtraction from a live 200 shape. The ADR documents it and you confirmed no external consumers - just ensure the MFE team is aware before deploy so no snapshot tests break on their side - `src/dto/preflight.js`
- nit: `makeAsyncJob` test fixture defines `getStartedAt` and `getEndedAt` that `toDetailJSON` never reads from AsyncJob - dead accessors that could mislead a future reader - `test/dto/preflight.test.js:57-58`

</details>

- Error payload lost on creation failure path - now correctly written to AsyncJob with `setError(...)` and Preflight row no longer carries error data.
- Missing null-AsyncJob controller test - added (both `.resolves(null)` and `.rejects(...)` paths covered).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 5m 31s | <b>Cost:</b> $6.89 | <b>Commit:</b> <code>fab8a6b39b4436cb7d614040d88ca52cf5cb542c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/dto/preflight.js
```diff
@@ -9,26 +9,132 @@
...
+   *
+   * @param {PreflightEntity} preflight
+   * @param {AsyncJobEntity | null} asyncJob
+   * @returns {PreflightDetail}
+   */
+  toDetailJSON: (preflight, asyncJob) => ({
+    ...PreflightDto.toJSON(preflight),
+    ...(asyncJob && {
+      status: /** @type {PreflightStatus} */ (asyncJob.getStatus()),
+      endedAt: asyncJob.getEndedAt(),
```
clotton: `asyncJob.getEndedAt()` returns `undefined` (not `null`) when `async_jobs.ended_at IS NULL` — the spacecat-shared `normalizeModelValue` maps DB NULL → `undefined`. This spread then **overrides** the `endedAt: preflight.getEndedAt()` already set by `...PreflightDto.toJSON(preflight)`, replacing a correct `null` with `undefined`. `JSON.stringify` drops the key entirely, violating the `string | null` wire contract for in-progress scans.

Fix: `endedAt: asyncJob.getEndedAt() ?? null`

src/dto/preflight.js
```diff
@@ -9,26 +9,132 @@
...
+    }),
+    result: asyncJob ? asyncJob.getResult() : null,
```
clotton: When `async_jobs.result IS NULL`, `asyncJob.getResult()` returns `undefined` (via `normalizeModelValue`). `result: undefined` is silently dropped by `JSON.stringify` — the client gets no `result` key, not `null` as the `object | null` contract promises. The existing test mock returns JS `null` directly and bypasses this path entirely.

Fix: `asyncJob ? (asyncJob.getResult() ?? null) : null`

Add a test case where the mock getter returns `undefined` and assert the serialised key is `null`.

src/dto/preflight.js
```diff
@@ -9,26 +9,132 @@
...
+    error: asyncJob ? asyncJob.getError() : null,
```
clotton: Same `undefined`-vs-`null` issue as `result` above: `asyncJob.getError()` returns `undefined` for `async_jobs.error IS NULL`. JSON drops the key; wire contract violated.

Fix: `asyncJob ? (asyncJob.getError() ?? null) : null`

src/dto/preflight.js
```diff
@@ -9,26 +9,132 @@
...
+ * @property {() => string} getCreatedAt
+ * @property {() => string} getUpdatedAt
+ * @property {() => (string | null)} getEndedAt
+ * @property {() => PreflightActor} getCreatedBy
+ *
+ * @typedef {Object} AsyncJobEntity
+ * @property {() => string} getStatus
+ * @property {() => (string | null)} getEndedAt
+ * @property {() => (object | null)} getResult
+ * @property {() => (PreflightError | null)} getError
```
clotton: This typedef claims `getError: () => (PreflightError | null)` but the runtime returns `PreflightError | undefined` (DB NULL → undefined via `normalizeModelValue`). Same for `getResult` (line 50) and `getEndedAt` (line 49). The wrong types make `// @ts-check` accept `asyncJob.getResult()` without the `?? null` coercion — the compile gate passes but the runtime drops the key. Change `| null` → `| undefined` on all three so `tsc` flags missing null coercions as type errors.

clotton (changes requested):
## Overview

Solid redesign — three distinct DTO shapes are a clear improvement over the single overloaded `toDetailJSON`, the spread composition avoids field-duplication drift, the 503 on fetch failure correctly distinguishes infra error from legitimate null result, and `e?.message ?? String(e)` handles non-Error rejections. One confirmed wire-contract violation needs fixing before merge.

---

### [BLOCKER] DB-null fields silently drop from the JSON response — `?? null` missing on three getters

`spacecat-shared`'s data model maps DB `NULL` → `undefined` (not `null`) via `normalizeModelValue`. When `async_jobs.result IS NULL`, `asyncJob.getResult()` returns `undefined`. An object-literal property set to `undefined` is silently omitted by `JSON.stringify`. The client receives **no `result` key at all** — not `null` as the wire contract promises.

The same applies to:
- `result: asyncJob ? asyncJob.getResult() : null` → when asyncJob exists but result IS NULL → key dropped
- `error: asyncJob ? asyncJob.getError() : null` → same
- `endedAt: asyncJob.getEndedAt()` inside the spread → when asyncJob exists but ended_at IS NULL (any in-progress scan), `endedAt: undefined` **overrides** the `endedAt: preflight.getEndedAt()` that `toJSON(preflight)` already placed correctly, then JSON drops the key entirely

The existing tests don't catch this because the mock getters return JS `null` directly, bypassing the `normalizeModelValue` DB-null-to-undefined path.

Fix:
```js
...(asyncJob && {
  status: /** @type {PreflightStatus} */ (asyncJob.getStatus()),
  endedAt: asyncJob.getEndedAt() ?? null,     // ← prevents undefined override
}),
result: asyncJob ? (asyncJob.getResult() ?? null) : null,
error: asyncJob ? (asyncJob.getError() ?? null) : null,
```

Add a test that seeds a mock AsyncJob where `getResult()` / `getError()` / `getEndedAt()` return `undefined` (not `null`) and asserts the serialized response keys are present with value `null`.

---

### [MEDIUM] `AsyncJobEntity` typedef declares wrong return types — `// @ts-check` gate can't catch the mismatch

```js
@property {() => (object | null)} getResult
@property {() => (PreflightError | null)} getError
@property {() => (string | null)} getEndedAt
```

The runtime returns `object | undefined`, `PreflightError | undefined`, `string | undefined` respectively (DB NULL → undefined via `normalizeModelValue`). The JSDoc promises `| null`, so `// @ts-check` considers `asyncJob.getResult() ?? null` a no-op (`null` already handled) and would accept `asyncJob.getResult()` directly without complaint — exactly the pattern that produces the wire bug above. The typedef should match reality:

```js
@property {() => (object | undefined)} getResult
@property {() => (PreflightError | undefined)} getError
@property {() => (string | undefined)} getEndedAt
```

Once fixed, `tsc` will flag `result: asyncJob.getResult()` (missing null coercion) as a type error at the `// @ts-check` gate — making the compile step a reliable guard against regressions of finding #1.

GeezerPelletier:
Sharp catch @clotton — both findings were real and the tests genuinely missed them. Addressed in `066dc6b0`.

### Blocker — DB-NULL silently dropping from the JSON response

Confirmed via the spacecat-shared `normalizeModelValue` path: DB NULL → JS `undefined`, and `JSON.stringify({ k: undefined })` drops the key. So an IN_PROGRESS preflight's detail response was missing `result`, `error`, and `endedAt` entirely — and the AsyncJob `undefined` was clobbering the correct `null` from the `toJSON(preflight)` spread on `endedAt` specifically. Wire contract `string | null` / `object | null` violated for every in-progress scan.

Fixed by adding `?? null` at every wire-boundary getter:

```js
toJSON: (preflight) => ({
  ...PreflightDto.toCreatedJSON(preflight),
  updatedAt: preflight.getUpdatedAt(),
  endedAt: preflight.getEndedAt() ?? null,
}),

toDetailJSON: (preflight, asyncJob) => ({
  ...PreflightDto.toJSON(preflight),
  ...(asyncJob && {
    status: asyncJob.getStatus(),
    endedAt: asyncJob.getEndedAt() ?? null,
  }),
  result: asyncJob ? (asyncJob.getResult() ?? null) : null,
  error: asyncJob ? (asyncJob.getError() ?? null) : null,
}),
```

Inline comments on both DTO methods explain why the coercion is load-bearing so a future hand can't strip them as "redundant."

### Regression coverage

Two new DTO tests that mirror the real entity behaviour (mock returns `undefined`, not `null`):

1. `coerces DB-NULL (returned as undefined) to JSON null on result/error/endedAt` — asserts both the object-literal value AND that `JSON.parse(JSON.stringify(out))` round-trips with the keys present. That's the actual failure mode the existing tests missed: they checked the in-memory value, never the serialised wire output.
2. `list: coerces DB-NULL endedAt to JSON null (in-progress preflight)` — same on the list shape, since `Preflight.getEndedAt()` returns `undefined` for any in-progress row.

Both pin the wire shape against the exact bug.

### Medium — `AsyncJobEntity` / `PreflightEntity` typedefs claimed `| null` but the runtime returns `| undefined`

You're right that this is what let the bug ship through `// @ts-check` — the typedef told tsc that `asyncJob.getResult()` already matched `object | null`, so a missing coercion looked fine. Updated all three nullable getters on both typedefs to `() => (T | undefined)`:

```js
@property {() => (string | undefined)} getEndedAt
@property {() => (object | undefined)} getResult
@property {() => (PreflightError | undefined)} getError
```

Added a brief explanation block at the top of the typedef section: "Entity getter return types use `| undefined` (not `| null`) for nullable columns because spacecat-shared's `normalizeModelValue` maps DB NULL → undefined. Encoding that here lets tsc flag missing `?? null` coercions at the wire boundary."

Verified locally: if I remove any one of the `?? null` coercions, `npx tsc -p tsconfig.json` now fails with `Type 'X | undefined' is not assignable to type 'X | null'`. The `// @ts-check` gate is finally doing the job it was supposed to.

### State

- Lint clean.
- `npx tsc -p tsconfig.json` clean.
- 117 passing locally (was 115, +2 for the DB-NULL regression tests).

Re-requesting review.

clotton (approved):
All findings resolved in this push:

- **`?? null` on all three AsyncJob getters** — `getEndedAt()`, `getResult()`, `getError()` now coerce `undefined` (DB NULL via `normalizeModelValue`) to `null` so JSON.stringify keeps the keys. Comments on each explain why it's load-bearing, not redundant.
- **`AsyncJobEntity` typedef updated to `| undefined`** — `// @ts-check` now correctly flags bare getter calls without `?? null` as type errors, making the compile gate a reliable guard against regressions.
- **`preflight.getEndedAt() ?? null`** also fixed in `toJSON` (list DTO) — consistent treatment across all three shapes.

LGTM.

PR #2716, opened 2026-06-27
feat(llmo): tag Cloudflare workers for idempotent deploys and add audit logging

nit23uec:
## Summary

Improves auditability/logging of the LLMO Cloudflare onboarding endpoints, adopts the worker **tags** feature from `@adobe/spacecat-shared-cloudflare-client@1.2.1`, and hardens the route-add step so onboarding can never disturb a customer's existing routing.

### Worker tagging → idempotent deploys
- Every deployed worker is tagged with the stable `adobe-llmo` ownership tag **plus** the caller's IMS identity.
- `adobe-llmo` is always present and always first, so the client's idempotency match (`tags.some((t) => settings.tags?.includes(t))`) recognizes a worker we previously deployed **regardless of which IMS user originally created it**. The per-user identity tag is an audit breadcrumb only and never participates in cross-user matching.
- On idempotent skip (client returns `null` for an already-tagged worker): return **200 with `alreadyDeployed: true`**, without re-setting the secret.
- A same-named worker **without** our tag remains a foreign collision → **409**.

### Tag-safety
- The caller-derived tag is sanitized to a Cloudflare-safe charset (`[A-Za-z0-9._-]`, capped at 80 chars). IMS GUID emails contain `@`/`.`, and Cloudflare's tag validation is undocumented beyond disallowing `,` and `&` — an invalid tag would fail the entire deploy. Audit **logs** keep the raw identity; only the **tag** is sanitized.

### Route protection (don't affect existing customer routing)
- Before adding a route, reject (**409**) if any existing route in the zone bound to **a different worker** shares a host with the one being added — protecting the customer's current routing for **any path, with any other worker**.
- Host comparison is **wildcard-aware** (`routeHostsOverlap`) and honors Cloudflare semantics: a bare host matches only itself; `*.base` matches strict subdomains of `base` (not the apex). So a customer `*.example.com/*` worker route correctly blocks an onboarding attempt on `a.example.com/*`, and scheme/path/string variants on the same host are caught — not just byte-identical duplicate patterns.
- A route already bound to **our own** derived worker is treated as idempotent (**200**, `alreadyRouted: true`) rather than a conflict.
- The 409 body returns the conflicting routes (`existingRoute` + `conflictingRoutes`) so the UI can surface exactly which routes/workers blocked the add.

### Auditability / structured logging
- Added greppable structured audit lines (`[llmo-cf] action=… outcome=… caller=… requestId=… <fields>`) across `deploy-worker` and `add-route`, including all error paths, for end-to-end correlation in Splunk.

## Test plan
- [x] `npm test` — full suite green (13026 passing)
- [x] Unit tests (controller): ownership + sanitized caller tag, `adobe-llmo`-always-first cross-user matching, char sanitization, idempotent null-skip; route host-overlap conflicts (incl. `*.example.com/*` vs `a.example.com/*`, scheme/path variants, sibling-subdomain allowed, own-worker idempotent)
- [x] Unit tests (utils): `routePatternHostGlob`, `routeHostsOverlap` wildcard semantics
- [x] `npm run lint` clean
- [ ] Deploy to a Cloudflare account where the derived worker already exists (tagged by another IMS user) → confirm idempotent 200; a same-named untagged worker → 409
- [ ] Add a route on a zone with an existing `*.example.com/*` worker route → confirm 409 with `conflictingRoutes`

### DIFF
--- a/src/controllers/llmo/llmo-cloudflare.js
+++ b/src/controllers/llmo/llmo-cloudflare.js
@@ -17,7 +17,9 @@ import { hasText, tracingFetch as fetch } from '@adobe/spacecat-shared-utils';
 import TokowakaClient from '@adobe/spacecat-shared-tokowaka-client';
 import CloudflareClient from '@adobe/spacecat-shared-cloudflare-client';
 import AccessControlUtil from '../../support/access-control-util.js';
-import { deriveWorkerName, hostInSiteDomain, routePatternHost } from './llmo-cloudflare-utils.js';
+import {
+  deriveWorkerName, hostInSiteDomain, routePatternHost, routeHostsOverlap,
+} from './llmo-cloudflare-utils.js';
 
 const CF_TOKEN_HEADER = 'x-cloudflare-token';
 // TEMPORARY: body/query fallback field for the CF token, used only until the
@@ -447,9 +449,10 @@ function LlmoCloudflareController(ctx) {
   /**
    * POST /sites/:siteId/llmo/cdn-onboard/cloudflare/routes
    * Body: { zoneId, pattern }
-   * Verifies server-side that the pattern targets the site's own domain and does not collide
-   * with an existing route in the zone before creating it, so a deploy cannot silently override
-   * a route the customer already has. `zoneId` is a Cloudflare identifier (not a SpaceCat
+   * Verifies server-side that the pattern targets the site's own domain and that no existing
+   * route in the zone already targets the same host (compared by resolved host, not raw pattern
+   * string) before creating it, so onboarding cannot silently add a second/overlapping route on
+   * a host the customer already routes. `zoneId` is a Cloudflare identifier (not a SpaceCat
    * entity), so it is supplied in the body rather than the path.
    */
   const addRoute = async (context) => {
@@ -490,8 +493,12 @@ function LlmoCloudflareController(ctx) {
       siteId, zoneId, scriptName, pattern,
     }));
 
-    // Guard against overriding an existing route: fetch the zone's current routes and reject
-    // if the requested pattern already exists.
+    // Protect the customer's existing routing: fetch the zone's routes and reject if any route
+    // bound to a DIFFERENT worker shares a host with the one we are about to add. We compare host
+    // globs (wildcard-aware, path-ignored) rather than raw pattern strings, so a customer's
+    // "*.example.com/*" worker route blocks an onboarding attempt on "a.example.com/*", and
+    // scheme/path/string variants on the same host are caught too — not just exact duplicates.
+    // Routes already bound to our own derived worker are safe and treated as idempotent.
     let existingRoutes;
     try {
       existingRoutes = await cfClient.listRoutes(zoneId);
@@ -499,15 +506,37 @@ function LlmoCloudflareController(ctx) {
       return cfErrorResponse(e, 'route lookup', context, { siteId, zoneId });
     }
 
-    const conflict = (existingRoutes || []).find((route) => route?.pattern === pattern);
-    if (conflict) {
-      log.info(auditLine(context, 'add-route', 'conflict', {
-        siteId, zoneId, scriptName, pattern,
+    const targetRouteHost = routePatternHost(pattern).toLowerCase();
+    const overlapping = (existingRoutes || []).filter(
+      (route) => hasText(route?.pattern) && routeHostsOverlap(route.pattern, pattern),
+    );
+
+    const foreignConflicts = overlapping.filter((route) => route.script !== scriptName);
+    if (foreignConflicts.length > 0) {
+      log.info(auditLine(context, 'add-route', 'conflict-foreign', {
+        siteId,
+        zoneId,
+        scriptName,
+        pattern,
+        host: targetRouteHost,
+        conflicts: foreignConflicts.map((r) => `${r.pattern}->${r.script || 'none'}`).join(','),
       }));
       return createResponse({
-        message: `A route for pattern '${pattern}' already exists in this zone`,
-        existingRoute: conflict,
+        message: `Existing route(s) in this zone already route host '${targetRouteHost}' to `
+          + 'another worker; refusing to add a route that could affect the customer\'s current '
+          + 'routing',
+        existingRoute: foreignConflicts[0],
+        conflictingRoutes: foreignConflicts,
       }, 409);
+    }
+
+    // An overlapping route already bound to our worker means this host is already onboarded.
+    const ownExisting = overlapping.find((route) => route.script === scriptName);
+    if (ownExisting) {
+      log.info(auditLine(context, 'add-route', 'already-routed', {
+        siteId, zoneId, scriptName, pattern, host: targetRouteHost,
+      }));
+      return ok({ ...ownExisting, alreadyRouted: true });
     }
 
     try {
--- a/src/controllers/llmo/llmo-cloudflare.js
+++ b/src/controllers/llmo/llmo-cloudflare.js
@@ -18,8 +18,12 @@ import TokowakaClient from '@adobe/spacecat-shared-tokowaka-client';
 import CloudflareClient from '@adobe/spacecat-shared-cloudflare-client';
 import AccessControlUtil from '../../support/access-control-util.js';
 import {
-  deriveWorkerName, hostInSiteDomain, routePatternHost, routeHostsOverlap,
+  deriveWorkerName, hostInSiteDomain, routePatternHost, routePatternHostGlob, routePatternsOverlap,
 } from './llmo-cloudflare-utils.js';
+
+// Cap the conflicting-routes list returned in a 409 so a zone with many overlapping routes can't
+// produce an unbounded response payload.
+const MAX_CONFLICTING_ROUTES = 10;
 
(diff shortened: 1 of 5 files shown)


src/controllers/llmo/llmo-cloudflare.js
```diff
@@ -39,6 +53,48 @@ const WORKER_SCRIPT_FETCH_TIMEOUT_MS = 10_000;
...
+const auditLine = (context, action, outcome, fields = {}) => {
+  const entries = {
+    action,
+    outcome,
+    caller: getCallerId(context),
+    requestId: context?.invocation?.id || 'unknown',
+    ...fields,
+  };
+  const kv = Object.entries(entries)
+    .filter(([, v]) => v !== undefined && v !== null && v !== '')
```
MysticatBot, line 103: issue (blocking): **Audit log values containing spaces break key=value parsing**

The `auditLine` helper joins `key=value` pairs with spaces. When a value itself contains a space (e.g. an error message like `Worker 'foo' does not exist`), downstream log parsers (Splunk, grep) will misparse the line - the space inside the value looks like a new key-value boundary.

The manual single-quote wrapping on some fields only covers fields you remember to quote and produces inconsistent quoting across the format.

**Why it matters:** The first time an error message with spaces lands in Splunk and someone tries to filter on `outcome=error`, they will get false matches from split fields.

**Fix options (pick one):**
1. Quote all values unconditionally in the map function
2. Replace spaces in values with underscores
3. Switch to JSON structured logging (`log.info({ msg: '[llmo-cf]', action, outcome, ...fields })`) - integrates better with Splunk/Coralogix and sidesteps escaping entirely

src/controllers/llmo/llmo-cloudflare.js
```diff
@@ -28,6 +30,18 @@ const CF_TOKEN_MISSING = 'Missing x-cloudflare-token header';
...
+
+// Cloudflare allows up to 8 tags per script and uses them in comma/colon-delimited filter
+// expressions, so caller-derived tag values must be sanitized. Cloudflare explicitly disallows
+// ',' and '&'; we conservatively restrict to [A-Za-z0-9._-] (dropping '@', ':', etc.) and cap the
+// length, since the full validation rules are undocumented and a rejected tag fails the deploy.
+const CF_TAG_MAX_LEN = 80;
+
 // Pin the worker script to an immutable commit SHA rather than a mutable branch HEAD so a
 // push to llmo-code-samples can never silently change what is deployed to customer accounts.
 // Overridable via env for forward-compat, but the default is always a pinned SHA.
```
MysticatBot, line 51: issue (blocking): **OpenAPI spec not updated for new response fields**

The CLAUDE.md rule states: "Specification Sync: Keep OpenAPI specs and implementation in sync - Run `npm run docs:lint` after modifying specs"

This PR adds three new fields to endpoint responses:
- `alreadyDeployed: true` on the deploy endpoint 200 response
- `alreadyRouted: true` on the routes endpoint 200 response
- `conflictingRoutes` array on the routes endpoint 409 response

None of these are reflected in `docs/openapi/llmo-api.yaml`. Clients consuming the OpenAPI spec will not see the new fields.

**Fix:** Update the response schemas in `docs/openapi/llmo-api.yaml` for both endpoints, then run `npm run docs:lint` to validate. Also update the route endpoint description which still says "rejects if a route for the same pattern already exists" (now uses wildcard-aware host overlap detection).

MysticatBot (changes requested):
Hey @nit23uec,

**Verdict: Request changes** - OpenAPI spec is out of sync with the new response fields, and an audit-logging issue warrants attention.
**Complexity: HIGH** - API surface change, 600+ changed lines across 4 files.
_Changes: Adds worker tagging for idempotent Cloudflare deploys, wildcard-aware route overlap detection, and structured audit logging to the LLMO onboarding endpoints (4 files)._

## Must fix before merge
1. [Critical] OpenAPI spec missing new `alreadyDeployed` field on deploy endpoint 200 response - `docs/openapi/llmo-api.yaml` (details inline)
2. [Critical] OpenAPI spec missing new `alreadyRouted` and `conflictingRoutes` fields on routes endpoint - `docs/openapi/llmo-api.yaml` (details inline)
3. [Important] Audit log values containing spaces will break key=value parsing in Splunk - `src/controllers/llmo/llmo-cloudflare.js:92` (details inline)
4. [Important] OpenAPI description for routes endpoint is stale (still says "same pattern already exists" but logic now uses wildcard-aware host overlap) - `docs/openapi/llmo-api.yaml` (details inline)

<details>

- nit: `targetRouteHost` in the 409 message uses `routePatternHost` (strips wildcards) while overlap logic uses `routePatternHostGlob` (preserves them) - a wildcard pattern shows the apex in the error message, which may confuse API consumers - `src/controllers/llmo/llmo-cloudflare.js:501`
- nit: The golden-path deploy test does not assert the 5th argument (`{ tags }`) - coverage exists in dedicated tests but the canonical assertion is incomplete - `test/controllers/llmo/llmo-cloudflare.test.js:305`
- suggestion: Add a unit test exercising `toWorkerTag` with an identity string longer than 80 characters to verify `CF_TAG_MAX_LEN` truncation
- suggestion: Cap `conflictingRoutes` array in the 409 response (e.g. `.slice(0, 10)`) to bound payload size for zones with many overlapping routes - `src/controllers/llmo/llmo-cloudflare.js:509`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 5m 12s | <b>Cost:</b> $6.88 | <b>Commit:</b> <code>99c9707522de27733974d60c7b0f59214065fec5</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

nit23uec:
Thanks for the review — all four must-fix items and the non-blocking nits are addressed in `0ecf2029`.

**Must fix**
1. ✅ **OpenAPI `alreadyDeployed`** — added to the deploy `200` schema; description now documents the idempotent-skip behavior.
2. ✅ **OpenAPI `alreadyRouted` / `conflictingRoutes`** — added to the routes `200` and `409` schemas (with `maxItems: 10`).
3. ✅ **Audit-log space parsing** — `auditLine` now quotes any value containing whitespace (`key="value with spaces"`) and the inconsistent manual single-quoting at call sites was removed. Locked in by a test asserting `error="..."`.
4. ✅ **Stale routes description** — rewritten to describe wildcard-aware host-overlap detection.

**Non-blocking**
- ✅ 409 message host now uses the wildcard-preserving glob (`routePatternHostGlob`), so a wildcard pattern shows `*.example.com` rather than the stripped apex.
- ✅ Golden-path deploy test now asserts the 5th `{ tags }` argument.
- ✅ Added a `toWorkerTag` truncation test (>80 chars → `CF_TAG_MAX_LEN`).
- ✅ `conflictingRoutes` capped at 10 (`MAX_CONFLICTING_ROUTES`), with a test.

**Extra (from reviewer/author follow-up):** generalized the route-conflict check from a `*.`-prefix special case to a generic **host-glob intersection** (`globsIntersect` / `routePatternsOverlap`). This correctly handles the broader `*example.com` wildcard (which, per Cloudflare's `*` = zero-or-more-chars operator, matches the apex and look-alikes), while `*.example.com` still never matches the apex. Only the host is compared; path is ignored by design.

Coverage remains 100% on both files; full suite green (13039 passing).

MysticatBot (approved):
Hey @nit23uec,

**Verdict: Approve** - all prior blocking findings are resolved; remaining observations are non-blocking.
**Complexity: HIGH** - API surface change, 600+ lines across 6 files.
_Changes: Adds worker tagging for idempotent Cloudflare deploys, wildcard-aware route overlap detection, and structured audit logging to the LLMO onboarding endpoints (6 files)._

- OpenAPI spec now includes `alreadyDeployed` on deploy 200 response
- OpenAPI spec now includes `alreadyRouted` and `conflictingRoutes` on routes 200/409 responses
- `auditLine` now quotes values containing whitespace, fixing Splunk key=value parsing
- Routes endpoint description updated to reflect wildcard-aware host-overlap detection

<details>

- nit: `auditLine` `fmt()` replaces internal double-quotes with single-quotes (lossy) - if an error message contains `"`, the logged value is altered. Consider backslash-escaping (`s.replace(/"/g, '\\"')`) for data fidelity - `src/controllers/llmo/llmo-cloudflare.js:423`
- suggestion: Add a combined-length guard to `globsIntersect` (e.g. `if (a.length + b.length > 500) return true`) so a malformed Cloudflare route pattern cannot exhaust the call stack via unbounded recursion - `src/controllers/llmo/llmo-cloudflare-utils.js:76`
- nit: The 409 `conflictingRoutes` response returns raw Cloudflare route objects including the foreign worker's `script` name. In shared-zone scenarios this leaks another customer's worker names. Consider returning only `{ pattern }` per entry - `src/controllers/llmo/llmo-cloudflare.js:660`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 25s | <b>Cost:</b> $4.89 | <b>Commit:</b> <code>0ecf2029ffcc8803952cfffc9eccae204f413ab9</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2717, opened 2026-06-29
fix(edge-optimize): resolve CloudFront host via overrideBaseURL + www-insensitive guardrail

ABHA61:
## What
The CloudFront onboarding controller derived hosts from the raw site `baseURL`, ignoring `fetchConfig.overrideBaseURL` and treating apex/www as different hosts. This made the distribution↔site guardrail **falsely block a www-only distribution for an apex site** (and vice versa), and could send the wrong `x-forwarded-host`.

## Changes
- Add `effectiveBaseURL(site)` = `overrideBaseURL` (when valid) `|| baseURL` — mirrors the AEM-CS-Fastly edge-routing path (`llmo.js`) and the shared client's `getEffectiveBaseURL`, so EO routing resolves the same host as the prerender pipeline.
- Use it for the forwarded host (`resolveEoTarget`, `verifyRouting`). **apiKey lookup stays on `baseURL`** (unchanged) to avoid affecting key resolution.
- Guardrail (`assertDistributionServesSite`) now matches **www-insensitively** via `getHostnameWithoutWww` against both `baseURL` and `overrideBaseURL`, so apex-only, www-only, and both-CNAME distributions all resolve correctly.

## Safety
No behavior change for sites without `overrideBaseURL` (e.g. apex-only) — `effectiveBaseURL` falls back to `baseURL` and `calculateForwardedHost` behaves exactly as before. Matching is exact-root equality (not substring), so unrelated domains never match.

## Tests
176 passing, incl. a new "allows an apex CNAME for a www site (www-insensitive match)" case.

- `verifyRouting` to reuse `probeSiteAndResolveDomain` (301-follow) — separate change.
- Human-UA in the shared client's `verifyRouting` (`Mozilla/5.0` → real browser UA) — separate client repo + republish.


test/controllers/llmo/llmo-cloudfront.test.js
```diff
@@ -1846,6 +1846,16 @@ describe('LlmoCloudFrontController', () => {
...
     });
 
+    it('allows an apex CNAME for a www site (www-insensitive match)', async () => {
+      // site baseURL is https://www.example.com; the distribution serves the apex example.com.
+      listDistributionsStub.resolves([{
+        id: 'E2EXAMPLE123', domainName: 'd1.cloudfront.net', aliases: ['example.com'], status: 'Deployed', enabled: true,
+      }]);
+      const result = await controller.deploy(deployContext);
+      expect(result.status).to.equal(200);
+      expect(runDeployStepStub.calledOnce).to.equal(true);
```
MysticatBot, line 1869: issue (blocking): The primary behavioral change in this PR - `effectiveBaseURL` returning `overrideBaseURL` when valid - has no end-to-end test with a configured override. The www-insensitive test here uses the default `mockSite` whose config presumably does not set `overrideBaseURL`, so it only proves the `getHostnameWithoutWww` normalization, not the override preference.

The codecov/patch failure (4 lines uncovered) likely includes this path. Without a test proving the override is actually used for `forwardedHost` derivation and guardrail matching, a regression to the old `baseURL`-only behavior would go undetected.

**Suggested fix:** Add a test where `mockSite.getConfig().getFetchConfig().overrideBaseURL` is set to a valid URL (e.g. `https://override.example.com`), then assert:
- `resolveEoTarget` returns `forwardedHost` derived from the override (not `baseURL`)
- `assertDistributionServesSite` passes when the distribution alias matches the override host

This closes the coverage gap and proves the fix works for the case it was written to address.

MysticatBot (changes requested):
Hey @ABHA61,

**Verdict: Request changes** - missing test coverage for the primary behavioral change (overrideBaseURL preference).
**Complexity: LOW** - small diff, single service.
_Changes: Fixes CloudFront host derivation to honor `overrideBaseURL` and match www-insensitively in the distribution-site guardrail (2 files)._

## Must fix before merge

1. [Important] No test exercises `effectiveBaseURL` when `overrideBaseURL` is configured - `test/controllers/llmo/llmo-cloudfront.test.js:1856` (details inline)

<details>

- nit: `verifyRouting` catch conflates "routing not ready" with infrastructure failures (network timeout, DNS) under a single `{ passed: false }` response - consider distinguishing transient errors so the wizard can surface actionable feedback instead of polling indefinitely - `src/controllers/llmo/llmo-cloudfront.js:889`
- nit: `siteHost` fallback initializes to `baseURL` (a full URL like `https://example.com`) rather than extracting `.host` first, so the error message shows the scheme when parsing fails - `src/controllers/llmo/llmo-cloudfront.js:289`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 6s | <b>Cost:</b> $4.68 | <b>Commit:</b> <code>5f130d4f14fdd999d3676b3883e37a748b0d70bb</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

test/controllers/llmo/llmo-cloudfront.test.js
ABHA61: Added — `verifyRouting` test *"probes the overrideBaseURL (not baseURL) when one is configured"*: it stubs `getFetchConfig().overrideBaseURL` and asserts `probeSiteAndResolveDomain` is called with the override URL, proving `effectiveBaseURL` honors it (the same helper drives the distribution guardrail and the forwarded-host). Also added an empty-probe → 400 guard test. Controller line + statement coverage is now 100%, so codecov/patch is back to ~100%.

ABHA61:
On the two non-blocking nits — leaving both as-is, reasoning below:

**`verifyRouting` conflates "routing not ready" with transient infra errors.**
Intentional. `verifyRouting` is driven by the wizard's deploy **poll loop**, so returning `{ passed: false, reason }` for any not-yet-routed state (including a transient DNS/network blip) lets the next poll retry, with `reason` surfaced for display. It never polls indefinitely — there's a `MAX_POLLS` cap that ends in a timeout. Distinguishing transient infra errors from "not ready" would need a new `retryable` signal **plus** corresponding FE handling; tracked as a follow-up enhancement rather than a change in this PR.

**`siteHost` fallback initializes to `baseURL` (full URL, shows scheme).**
That fallback only applies inside a catch that's unreachable in practice: `new URL(effectiveBaseURL(site)).host` throws only for an unparseable URL, but `effectiveBaseURL` is always the site's valid `baseURL` or an `isValidUrl`-gated override. So `siteHost` is always the extracted host and the scheme is never shown — no change needed.

(Both `assertDistributionServesSite` catches are `c8 ignore`d for the same reason: `getHostnameWithoutWww` only throws on unparseable input, which can't occur for an onboarded site's baseURL/overrideBaseURL or a CloudFront alias — and a malformed baseURL would throw earlier in `resolveEoTarget`.)

MysticatBot (approved):
Hey @ABHA61,

**Verdict: Approve** - prior blocking finding addressed, clean re-review.
**Complexity: LOW** - small diff, single service.
_Changes: Fixes CloudFront host derivation to honor `overrideBaseURL` and match www-insensitively in the distribution-site guardrail (2 files)._

<details>

- nit: `assertDistributionServesSite` iterates `[baseURL, effectiveBaseURL(site)]` through `getHostnameWithoutWww` - when no override is configured (majority case), both resolve to the same hostname and the second iteration is a wasted parse. A `new Set([baseURL, effectiveBaseURL(site)])` on the input would self-document the dedup intent - `src/controllers/llmo/llmo-cloudfront.js:262`
- suggestion: `validateCloudfrontCredentials(context, { requireDistribution: true })` still gates entry on `distributionId` presence, but `verifyRouting` no longer uses the distribution ID. If intentional (wizard-step sequencing guard), a one-line comment would help future maintainers; if not, `requireDistribution` can be dropped - `src/controllers/llmo/llmo-cloudfront.js:874`
- suggestion: Consider extracting `effectiveBaseURL` to `edge-routing-utils.js` (where `getHostnameWithoutWww` and `probeSiteAndResolveDomain` already live) so the inline copies in `llmo.js` and `llmo-cloudfront.js` collapse to one shared source of truth.

</details>

- Test for `effectiveBaseURL` when `overrideBaseURL` is configured - now covered by "probes the overrideBaseURL (not baseURL) when one is configured" test; codecov confirms 100% coverage on modified lines.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 5s | <b>Cost:</b> $6.78 | <b>Commit:</b> <code>bcfd7142e5dc217e16989cd4d7741d1dfc959516</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

package.json
```diff
@@ -98,7 +98,7 @@
     "@adobe/spacecat-shared-scrape-client": "2.6.3",
     "@adobe/spacecat-shared-slack-client": "1.6.7",
     "@adobe/spacecat-shared-tier-client": "1.5.1",
-    "@adobe/spacecat-shared-tokowaka-client": "1.20.0",
+    "@adobe/spacecat-shared-tokowaka-client": "https://gist.githubusercontent.com/ABHA61/6c10e4a793e58dd6d0e544b24edcf2fa/raw/69334c164e390a9ae54cf42235dbe3aabd759ed8/adobe-spacecat-shared-tokowaka-client-1.20.1-eo-bridge.0.tgz",
```
MysticatBot, line 101: issue (blocking): `@adobe/spacecat-shared-tokowaka-client` is resolved from `https://gist.githubusercontent.com/ABHA61/...` - a tarball hosted on a personal GitHub Gist - instead of the npm registry.

**Why this blocks:**
- The Gist is mutable: the owner (or anyone who gains access to the account) can replace the tarball content at any time without a version bump.
- `npm audit`, Dependabot, and Renovate cannot track this dependency for CVEs.
- No peer review of dependency content: unlike a proper npm publish (which goes through CI and team review on the shared-client repo), this tarball's contents are unauditable by other maintainers.
- CI passes, creating false confidence - it only proves tests work with the current tarball, not that the tarball is safe for production.

**How to fix:** Publish `1.20.1-eo-bridge.0` to the npm registry under `@adobe/spacecat-shared-tokowaka-client` as a proper prerelease (e.g., `npm publish --tag eo-bridge` from the shared-client repo's CI), then reference it by semver here. If the shared-client PR is not yet merged, use an npm prerelease tag from a feature-branch CI build. This PR must not merge with a Gist-hosted dependency.

MysticatBot (changes requested):
Hey @ABHA61,

**Verdict: Request changes** - production dependency sourced from a personal Gist tarball is a supply chain blocker.
**Complexity: HIGH** - medium diff; Dependencies risk flag.
_Changes: Fixes CloudFront host derivation to honor `overrideBaseURL` and match www-insensitively in the distribution-site guardrail, adds operator audit attribution to all assumed-role calls (4 files)._

## Must fix before merge

1. [Critical] Production dependency `@adobe/spacecat-shared-tokowaka-client` pinned to a personal Gist tarball URL instead of npm registry - `package.json:101` (details inline)

<details>

- suggestion: `getCallerId` passes unsanitized email as `operator` to `assumeConnectorRole` (likely used as AWS STS `RoleSessionName`). STS restricts session names to `[\w+=,.@-]` (max 64 chars). If an email contains characters outside this set, the assume-role call will fail. Consider sanitizing or truncating before passing - `src/controllers/llmo/llmo-cloudfront.js:209`

</details>

- Test for `effectiveBaseURL` when `overrideBaseURL` is configured - now covered by "probes the overrideBaseURL (not baseURL) when one is configured" test; codecov confirms 100% coverage on modified lines.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 49s | <b>Cost:</b> $5.60 | <b>Commit:</b> <code>c19e168c2ae5c49d069017a5a9f86c4ecc33ecec</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Issue #2718, opened 2026-06-29
serenity: harden sub-workspace create against failed-provisioning stubs (status-filter adoption + idempotent create-or-adopt)


rainer-friederich:
## Summary

Harden serenity's sub-workspace create against **failed-provisioning stubs**. A Semrush User Manager `POST /enterprise/users/api/v2/workspaces/{parent}/child` can return `200` (accepted) and then **fail provisioning asynchronously**, leaving the child permanently stuck `status: "not ready"` with an `"invalid subscription"` (no resources attached, no PE-side workspace). Such a stub **cannot be removed** by us — both `DELETE /v1/workspaces/{id}` and `POST /v2/workspaces/{id}/resources/transfer` return `422 {"message":"workspace not ready"}` until the workspace reaches `created`, which a failed one never does.

This was observed live (2026-06-29) while exercising the User Manager API; details + the full live behavior matrix are in adobe/spacecat-shared#1745. The Semrush-side root cause (ack-then-fail with no client reap path) is being raised separately with Semrush — this issue is the **client-side hardening** so a rare partial failure can't snowball.

## Current behavior (`src/support/serenity/workspace-lifecycle.js`)

`ensureSubworkspace` is correct on the happy path and fails *cleanly* on a stuck create (`pollUntilCreated` → `504`, brand pointer left unset). The exposure is in the **aftermath** of a partial failure:

1. A non-settling create leaves the stub in the parent `family` (we can't delete it — see above).
2. A retry creates **another same-title child** (titles embed the brand-id suffix, so they collide on retry).
3. `adoptFromFamily` matches family entries on `title` + `projectCount === 0` only — it does **NOT** filter on `status === 'created'`. A `not ready` zombie also has `projectCount 0`, so:
   - it can be *adopted* as the brand's workspace (then immediately re-times-out at the next `pollUntilCreated`), and
   - once ≥2 same-title stubs accumulate, the next adoption sees **multiple matches → `409 ambiguousWorkspace`**, wedging that brand.

## Mitigation 1 — filter family/adoption on `status === 'created'`

In `adoptFromFamily` (and any family read used for matching / the ambiguity count), consider **only** entries with `status === 'created'`. A `not ready`/failed stub must never be adopted and must not inflate the multiple-match `409`. This breaks the snowball: accumulated zombies become invisible to the matcher.

## Mitigation 2 — idempotent create-or-adopt

[...]

PR #2719, opened 2026-06-29
fix(serenity): harden sub-workspace create against failed-provisioning stubs

rainer-friederich:
## 1. Abstract
Hardens serenity's sub-workspace create (`src/support/serenity/workspace-lifecycle.js`) against Semrush "failed-provisioning" stubs by status-filtering adoption and making the create path an idempotent create-or-adopt.

## 2. Reasoning
A Semrush User Manager child-workspace create (`POST .../workspaces/{parent}/child`) can return `200` (accepted) and then fail provisioning asynchronously, leaving the child permanently stuck `status: "not ready"` ("invalid subscription"). Such a stub cannot be removed by us — both `DELETE /v1/workspaces/{id}` and `POST /v2/workspaces/{id}/resources/transfer` return `422 "workspace not ready"` until the workspace reaches `created`, which a failed one never does. Observed live on 2026-06-29 while exercising the User Manager API. The Semrush-side root cause (ack-then-fail with no client reap path) is being raised separately; this change is the client-side hardening so a rare partial failure cannot snowball.

Before this change the snowball was: a non-settling create leaves an undeletable stub in the parent family; a retry creates another same-title child (titles embed the brand-id suffix, so they collide); and adoption matched on `title` + `projectCount === 0` only — never on `status`. A `not ready` zombie also has `projectCount 0`, so it could be falsely adopted (then re-time-out at the next poll), and once ≥2 same-title stubs accumulated the matcher saw multiple matches and returned a false `409 ambiguousWorkspace`, wedging that brand.

Two client-side mitigations, both in `ensureSubworkspace` / its family-matching helper:

- **Status-filter adoption.** A new shared helper `findAdoptableFamilyMatch` considers **only** family entries with `status === 'created'` — for both the title match and the multiple-match count. Effect: a failed `not ready` stub is now invisible to the matcher, so it can neither be falsely adopted nor inflate the ambiguity `409`. The existing 504-timeout recovery (`adoptFromFamily`) now delegates to this same helper (no match → `502`), keeping the match/ambiguity/empty rules in one place.
  - Before: single `not ready` same-title stub could be adopted; ≥2 accumulated stubs → false `409`.
  - After: `not ready` stubs ignored; one `created` among same-title entries is adopted with no `409`; a genuinely ambiguous case (≥2 **`created`** same-title) still returns `409 ambiguousWorkspace` (unchanged).

- **Idempotent create-or-adopt.** `ensureSubworkspace` now checks the parent family for an existing **`created`**, empty same-title child **before** calling `createSubworkspace`, and reuses it; it only creates when none exists. Previously adoption ran only on the `504` timeout path, so a retry after a partial failure spawned a duplicate stub. Now a retry reuses the good child.

The `504`/timeout adopt-or-fail behaviour is preserved for the `created` case, and the existing parent-equality guard, the empty-only adoption guard, and the concurrent-activation release/adopt guard are all unchanged.

## 4. Required information
- Jira / issue: https://github.com/adobe/spacecat-api-service/issues/2718
- Implementation plan: docs/specs/2026-06-15-serenity-subworkspace-dual-mode-implementation-plan.md (sub-workspace lifecycle design §6 — adoption / ambiguous-create recovery)
- Other: live behaviour + repro https://github.com/adobe/spacecat-shared/issues/1745 ; Semrush-side root cause https://github.com/adobe/serenity-docs/issues/20

## 7. Test plan
(a) Local: this is pure client-side control-flow logic; verification is via the unit suite (`test/support/serenity/workspace-lifecycle.test.js`), which exercises every new branch with injected poll timing (no real delays) — single `not ready` stub not adopted; one `created` among same-title adopted with no `409`; ≥2 `created` still `409`; idempotent reuse instead of duplicate create; and the preserved `504` recovery (two-call family stub proving both the proactive read and the post-timeout recovery read fire).

(b) Per-environment verification (dev): exercise a brand activation/retry for a brand whose org has a parent workspace. Confirm that a retry after a partial-failure create reuses the existing `created` same-title child rather than creating a second one (inspect the parent family listing — no duplicate same-title entries appear from the retry), and that a brand whose family contains only `not ready` stubs proceeds to a fresh create rather than wedging on a `409`. No schema, route, or contract change — no stage/prod-specific steps beyond the standard deploy.

### DIFF
--- a/src/support/serenity/workspace-lifecycle.js
+++ b/src/support/serenity/workspace-lifecycle.js
@@ -153,6 +153,18 @@ async function findAdoptableFamilyMatch(transport, parentWorkspaceId, title, log
   const items = familyItems(family);
   const matches = items.filter((w) => w?.title === title && w?.status === 'created');
   if (matches.length === 0) {
+    // Surface filtered-out non-`created` same-title stubs (Semrush ack-then-fail
+    // zombies) so their accumulation is visible in logs without a manual family
+    // query — they are the exact failure mode this status filter absorbs (#2718).
+    const ignored = items.filter((w) => w?.title === title && w?.status !== 'created');
+    if (ignored.length > 0) {
+      log?.info?.('ensureSubworkspace: ignoring non-created same-title family stub(s)', {
+        parentWorkspaceId,
+        title,
+        ignoredCount: ignored.length,
+        ignoredStatuses: ignored.map((w) => w?.status),
+      });
+    }
     return null;
   }
   if (matches.length > 1) {
--- a/src/support/serenity/workspace-lifecycle.js
+++ b/src/support/serenity/workspace-lifecycle.js
@@ -125,23 +125,38 @@ function familyItems(family) {
 }
 
 /**
- * Ambiguous-create recovery (design §6): a timed-out createSubworkspace is
- * ambiguous (no idempotency key). List the parent's family, match the exact
- * title, and adopt a `created`, project-empty subworkspace. Multiple matches → fail
- * with an alert, never guess.
+ * Finds the one adoptable same-title child in the parent's family, or `null` when
+ * none exists. "Adoptable" means a `created`, project-empty sub-workspace.
+ *
+ * Status filter (issue #2718): a Semrush child create can be 200-acked and then
+ * fail provisioning asynchronously, leaving a stub permanently stuck at
+ * `status: 'not ready'` ("invalid subscription") that we cannot delete. Such a
+ * zombie also has `projectCount 0`, so a title+empty-only match would (a) adopt
+ * it as the brand's workspace (then immediately re-time-out at pollUntilCreated)
+ * and (b) once ≥2 accumulate, inflate the multiple-match `409` and wedge the
+ * brand. Considering ONLY `status === 'created'` entries makes accumulated
+ * zombies invisible to the matcher, breaking that snowball. The live family
+ * endpoint always returns a status, so the strict equality is safe.
+ *
+ * Shared by both adoption paths so the match/ambiguity/empty rules live in one
+ * place: the proactive create-or-adopt check (returns the match to reuse, or
+ * null → create) and the 504 timeout recovery (null → no-match error).
+ *
+ * @param {object} transport
+ * @param {string} parentWorkspaceId
+ * @param {string} title
+ * @param {object} log
+ * @returns {Promise<object|null>} the sole adoptable family entry, or null.
  */
-async function adoptFromFamily(transport, parentWorkspaceId, title, log) {
+async function findAdoptableFamilyMatch(transport, parentWorkspaceId, title, log) {
   const family = await transport.listWorkspaceFamily(parentWorkspaceId);
   const items = familyItems(family);
-  const matches = items.filter((w) => w?.title === title);
+  const matches = items.filter((w) => w?.title === title && w?.status === 'created');
   if (matches.length === 0) {
-    throw new ErrorWithStatusCode(
-      `Ambiguous subworkspace create for '${title}' and no family match to adopt`,
-      502,
-    );
+    return null;
   }
   if (matches.length > 1) {
-    log?.error?.('ensureSubworkspace: ambiguous create — multiple family matches, refusing to guess', {
+    log?.error?.('ensureSubworkspace: ambiguous create — multiple created family matches, refusing to guess', {
       parentWorkspaceId,
       title,
       matchIds: matches.map((m) => m?.id),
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-scoped hardening with thorough test coverage.
**Complexity: MEDIUM** - medium diff, single-service, API surface signal.
_Changes: Hardens sub-workspace create against Semrush failed-provisioning stubs by status-filtering adoption and making the create path idempotent (2 files)._

<details>

- suggestion: The proactive `findAdoptableFamilyMatch` adds one `listWorkspaceFamily` round-trip to every first-time brand activation (previously this call only ran on the 504-recovery path). Both the code and PR description make clear this is an intentional correctness-for-latency trade-off. If call volume grows, consider short-circuiting the 504-recovery re-list when the proactive check already ran moments before (pass a flag or the prior result). - `src/support/serenity/workspace-lifecycle.js:289`
- suggestion: When `findAdoptableFamilyMatch` returns null on the proactive path but same-title entries with non-`created` status exist, a log line noting the filtered-out zombie count would aid operational debugging without needing to query the family endpoint manually. - `src/support/serenity/workspace-lifecycle.js:~145`
- nit: The proactive check propagates `listWorkspaceFamily` errors directly (no try/catch), so a transient upstream failure now blocks workspace creation before `createSubworkspace` is attempted. This is arguably more correct (fail-safe: do not create if you cannot verify duplicates), but it is a subtle behavior change from the pre-PR flow where the family listing only ran post-504. No test exercises this path. - `src/support/serenity/workspace-lifecycle.js:289`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 18s | <b>Cost:</b> $6.16 | <b>Commit:</b> <code>38ea045530966a6d60a02f5da6302cb1d1fdd55c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the MysticatBot review (Approve, 3 non-blocking) in `cc7bce21`.

MysticatBot:
- suggestion (extra `listWorkspaceFamily` round-trip; short-circuit the 504-recovery re-list with the proactive result) — declined. The 504-recovery re-list deliberately reads *fresh* post-timeout state: the proactive read returned `null` precisely because the workspace did not exist yet, but the create can settle server-side during the client-visible timeout. Reusing the stale proactive result would miss exactly the workspace recovery exists to adopt. The one proactive round-trip is the intended correctness-for-latency trade-off; short-circuiting it would be incorrect, not just an optimization, and is YAGNI at current volume.
- suggestion (log the filtered-out zombie count) — done. `findAdoptableFamilyMatch` now emits an info line with `ignoredCount` + `ignoredStatuses` when same-title entries are dropped by the status filter, so accumulating ack-then-fail stubs are visible in logs without a manual family query.
- nit (proactive `listWorkspaceFamily` error propagates with no test) — fail-safe behavior kept (not creating when duplicates cannot be verified is the correct guard against the duplicate-stub problem this PR fixes); added a test covering that propagation path, plus tests for the new ignored-stub log line and its absence on a clean first create.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all prior suggestions addressed cleanly; no new concerns.
_Changes: Adds zombie-stub logging for operational visibility when non-`created` same-title family entries are filtered out, plus 3 tests covering the logging behavior and fail-safe error propagation (2 files)._

<details>

- suggestion: Consider deduplicating `ignoredStatuses` (e.g. `[...new Set(ignored.map(w => w?.status))]`) - if many zombies accumulate they will all share the same status string, making the array repetitive while `ignoredCount` already conveys volume. - `src/support/serenity/workspace-lifecycle.js:156`

</details>

- Zombie-stub logging for operational visibility - implemented as structured INFO with parentWorkspaceId, title, ignoredCount, ignoredStatuses.
- Proactive-check error propagation test (fail-safe: no blind create) - new test verifies `listWorkspaceFamily` errors bubble up and block `createSubworkspace`.
- Short-circuit 504 re-list suggestion acknowledged (prior non-blocking, no action needed).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 0s | <b>Cost:</b> $3.01 | <b>Commit:</b> <code>cc7bce21a7badd0ea9794c43816fb91df717381a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Applied the remaining suggestion in `421ae8a`.

MysticatBot:
- suggestion (dedupe `ignoredStatuses`) — done. The log line now emits a deduped set of distinct statuses; `ignoredCount` still conveys volume, so the line stays compact even when many same-status zombies accumulate. Test strengthened to assert the dedup (3 stubs → 2 distinct statuses).

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all prior suggestions addressed cleanly; no new concerns.
_Changes: Deduplicates the `ignoredStatuses` log field via `new Set()` so repeated zombie statuses do not bloat the line, with test coverage proving the dedup (2 files)._

- Deduplicate `ignoredStatuses` - implemented as `[...new Set(ignored.map(w => w?.status))]`; test strengthened with a third stub sharing a duplicate status, asserting both unique membership and length.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 55s | <b>Cost:</b> $2.34 | <b>Commit:</b> <code>421ae8a78c2e006392228e48c46da3f47a335c22</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Issue #2720, opened 2026-06-29
serenity updateProject omits required `type` — brand-alias PATCH 400s against live Semrush


rainer-friederich:
## Summary

`serenity` brand-alias re-sync calls `transport.updateProject(...)` with a body that omits the required `type` field, so the `PATCH /v1/workspaces/{id}/projects/{project_id}` request **400s against the live Semrush Project Engine gateway**. Any brand-alias change that actually drifts a project's `brand_names` would fail in production.

## Where

`src/support/serenity/brand-aliases.js` (the project `brand_names` PATCH branch):

```js
await transport.updateProject(workspaceId, projectId, {
  ...(hasText(display) ? { brand_name_display: display } : {}),
  brand_names: desiredBrandNames,
});   // <-- no `type`
```

## Evidence (live, captured 2026-06-29)

- `PATCH /v1/workspaces/{id}/projects/{pid}` **without `type`** -> `400` `{message:"Key: 'ProjectUpdateRequest.Type' Error:Field validation for 'Type' failed on the 'required' tag"}`.
- The same PATCH **with `type:'ai'`** -> `200` full `ProjectResponse`.
- `ProjectUpdateRequest.type` is marked `required` in the Semrush swagger; the Go gateway enforces it (Go-validator message above).

Full live sweep that surfaced this:

https://github.com/adobe/spacecat-shared/issues/1745#issuecomment-4831892663

## Why it was not caught earlier

The PE Counterfact mock already 400s a typeless PATCH (the vendored spec marks `type` required and Counterfact enforces it at the request-validation seam), so the mock does NOT hide this — it is now pinned by an e2e in spacecat-shared:

https://github.com/adobe/spacecat-shared/pull/1746

The gap is purely that serenity's `updateProject` body never carried `type`, and no IT yet exercised the alias-resync PATCH path against the mock.

## Fix

Add `type: 'ai'` to the `updateProject` body in `brand-aliases.js`. Consider asserting it in the serenity IT (the alias-resync path) so the mock's 400 guards it going forward.

## Severity

Real production defect on the brand-alias re-sync path (a drifted `brand_names` update fails). Low blast radius today (only triggers when aliases actually drift), but it silently no-ops the project-level alias sync.

rainer-friederich:
Closed with #2721

PR #2721, opened 2026-06-29
fix(serenity): add required `type` to brand-alias updateProject PATCH

rainer-friederich:
## 1. Abstract
Adds the required `type: 'ai'` field to the serenity brand-alias re-sync `updateProject` PATCH body so the request validates against the Semrush Project Engine gateway.

## 2. Reasoning
The brand-alias re-sync in `src/support/serenity/brand-aliases.js` called `transport.updateProject(...)` with a body that omitted the `type` field, which is `required` on `ProjectUpdateRequest`. A live sweep (captured 2026-06-29) confirmed that a typeless `PATCH /v1/workspaces/{id}/projects/{pid}` returns `400` (`Key: 'ProjectUpdateRequest.Type' Error:Field validation for 'Type' failed on the 'required' tag`), while the same PATCH with `type: 'ai'` returns `200`. As a result, any brand-alias change that actually drifts a project's `brand_names` would fail in production and silently no-op the project-level alias sync. The gap went uncaught because no IT exercised the alias-resync PATCH path against the PE mock, even though the mock already 400s a typeless PATCH.

- Before: the `brand_names` re-sync branch sent `{ brand_name_display?, brand_names }` — no `type` — so a drifted update 400'd against the live gateway and the alias sync silently failed.
- After: the PATCH body carries `type: 'ai'` (matching the literal used at project create-time in `handlers/markets.js` and `handlers/markets-subworkspace.js`; AIO projects are always type `ai`), so the request validates and the drifted `brand_names` update succeeds.
- Behaviour is otherwise unchanged: the PATCH still fires only when the live `brand_names` set has drifted from the desired set; the benchmark `brand_aliases` PUT path is untouched.
- The three exact-body assertions in the unit test now pin `type: 'ai'`, so a future regression that drops the field is caught locally.

## 4. Required information
- Jira / issue: https://github.com/adobe/spacecat-api-service/issues/2720
- Other: https://github.com/adobe/spacecat-shared/pull/1746

- spacecat-shared (contract): the PE Counterfact mock enforces the `type` requirement at the request-validation seam; spacecat-shared PR #1746 pins the typeless-PATCH 400 as an e2e. This change makes serenity conform to that contract.

## 7. Test plan
- (a) Local: ran the `brand-aliases` unit suite, which now asserts `type: 'ai'` on all three `updateProject` body expectations; the serenity `// @ts-check` type-check gate also covers the touched file. No live e2e was run this session — the production defect was identified from the 2026-06-29 live Semrush sweep referenced in the issue.
- (b) dev / stage / prod: after deploy, trigger a brand-alias change that drifts a serenity-mode brand's `brand_names` (display name or alias edit) and confirm the `PATCH /v1/workspaces/{id}/projects/{pid}` returns `200` rather than `400`, and that the project's `settings.ai.brand_names` reflects the new set.

- https://github.com/adobe/spacecat-shared/pull/1746 (related): pins the mock-level typeless-PATCH 400 guard upstream. Independent of this PR — no hard merge/deploy ordering required; the two together guard the defect at both the mock layer and the serenity call site.

### DIFF
--- a/src/support/serenity/brand-aliases.js
+++ b/src/support/serenity/brand-aliases.js
@@ -129,6 +129,9 @@ export async function syncBrandAliasesAcrossMarkets(
       if (!sameAliasSet(currentBrandNames, desiredBrandNames)) {
         // eslint-disable-next-line no-await-in-loop
         await transport.updateProject(workspaceId, projectId, {
+          // `type` is required on ProjectUpdateRequest — the Semrush gateway
+          // (and the PE mock) 400 a typeless PATCH. AIO projects are always 'ai'.
+          type: 'ai',
           ...(hasText(display) ? { brand_name_display: display } : {}),
           brand_names: desiredBrandNames,
         });
--- a/test/support/serenity/brand-aliases.test.js
+++ b/test/support/serenity/brand-aliases.test.js
@@ -87,6 +87,7 @@ describe('brand-aliases helpers', () => {
       const result = await syncBrandAliasesAcrossMarkets(transport, aliases, 'Brand', WS, undefined);
 
       expect(transport.updateProject).to.have.been.calledOnceWith(WS, 'p-us', {
+        type: 'ai',
         brand_name_display: 'Brand',
         brand_names: ['Brand', 'Acme', 'Acme Inc'],
       });
@@ -122,6 +123,7 @@ describe('brand-aliases helpers', () => {
       );
 
       expect(transport.updateProject).to.have.been.calledOnceWith(WS, 'p-us', {
+        type: 'ai',
         brand_name_display: 'Brand',
         brand_names: ['Brand', 'Acme'],
       });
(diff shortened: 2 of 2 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - clean, well-scoped bug fix with adequate test coverage.
**Complexity: TRIVIAL** - 6 lines, 2 files, single call-site fix.
_Changes: Adds the required `type: 'ai'` field to the brand-alias re-sync `updateProject` PATCH body so it validates against the Semrush Project Engine gateway (2 files)._

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 43s | <b>Cost:</b> $2.00 | <b>Commit:</b> <code>3173a31bba38c062f019f5f0b006545f75c38398</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2722, opened 2026-06-29
fix: exclude soft-deleted brands from name uniqueness check | LLMO-5919

igor-grubic:
## Summary

Soft-deleted brands were not excluded from the `upsertBrand` name uniqueness check, causing a false \"brand name already exists\" 409 error when a user tried to create a new brand reusing the name of a previously deleted brand.

### Root Cause

`upsertBrand` in `brands-storage.js` queried for existing brands without filtering out soft-deleted rows. The deleted brand's non-null `site_id` made `hasAnchor=true`, silently resurrecting the brand as 'active' with the old site. The `baseSiteId` immutability guard (LLMO-5556) then prevented the new `baseSiteId` from being applied, so the ON CONFLICT UPDATE kept the stale `site_id`. When another brand had since claimed that site, `brands_base_site_unique` fired → 23505 → 409 error.

### Fix

- **`spacecat-api-service/src/support/brands-storage.js`**: Add `.neq('status', 'deleted')` to the existing-brand lookup; restructure `site_id` assignment so `existing===null` (fresh create OR resurrection) always writes an explicit `site_id` or null.
- **`spacecat-api-service/test/support/brands-storage.test.js`**: Extend `createCapturingClient` to capture `.neq()` filter calls; add 2 new `upsertBrand` tests covering LLMO-5919 scenarios including assertion that the deleted-brand filter is applied.

### Test Results

- 165 passing (brands-storage.test.js), 13233 passing (full suite)
- Lint: clean
- IT tests: not run (requires Docker + ECR)

### Known Gaps

- No IT test against a real DB for the soft-deleted-brand resurrection scenario

## Links

- Jira: https://jira.corp.adobe.com/browse/LLMO-5919
- Slack thread: https://adobe.enterprise.slack.com/archives/C0BEJH1SNNL/p1782741378312489

### DIFF
--- a/src/support/brands-storage.js
+++ b/src/support/brands-storage.js
@@ -836,14 +836,17 @@ export async function upsertBrand({
   const regions = (brand.region || [])
     .map((r) => (typeof r === 'string' ? r : String(r))).filter(hasText);
 
-  // Check if the brand already exists with a base site set.
-  // This prevents silently downgrading an active brand to pending when a caller
-  // re-upserts by name without passing baseSiteId.
+  // Check if a non-deleted brand already exists with this name. Soft-deleted
+  // brands are excluded (.neq('status', 'deleted')) so that creating a brand
+  // whose name matches a deleted record is treated as a fresh create — the
+  // caller gets the expected new brand instead of a resurrected row that
+  // inherits the deleted brand's site_id and anchoring state (LLMO-5919).
   const { data: existing, error: existingError } = await postgrestClient
     .from('brands')
     .select('id, site_id, status')
     .eq('organization_id', organizationId)
     .eq('name', brand.name)
+    .neq('status', 'deleted')
     .maybeSingle();
 
   // Fail closed (LLMO-5556): PostgREST returns { data: null, error } on a query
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @igor-grubic,

**Verdict: Approve** - well-scoped bug fix with clear root cause analysis and appropriate tests.
**Complexity: LOW** - small diff, single-service, no risk signals.
_Changes: Excludes soft-deleted brands from the name uniqueness check in `upsertBrand` and explicitly clears stale `site_id` on resurrection to prevent unique constraint collisions (2 files)._

<details>

- nit: `.neq('status', 'deleted')` also excludes brands with NULL status due to SQL three-valued logic - if the `status` column lacks a NOT NULL constraint and NULL rows exist in production, they would be invisible to the uniqueness check (same as deleted brands). Verify the column constraint, or use `.or('status.neq.deleted,status.is.null')` if NULL rows can exist - `src/support/brands-storage.js:851`
- suggestion: Consider a partial unique index (`WHERE status != 'deleted'`) on `(organization_id, name)` long-term to let the DB enforce the actual business rule natively and eliminate the application/DB semantics divergence that requires explicit `site_id = null` on resurrection - `src/support/brands-storage.js:846`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 53s | <b>Cost:</b> $5.24 | <b>Commit:</b> <code>5cc013d62e44d866dd9c016f283a630541a75eff</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

calvarezg:
## Multi-Persona PR Review — LLMO-5919

**Reviewed by:** Architect · DBA · Security · Tester · Junior Dev

Solid, well-scoped fix. Root cause analysis is accurate and the explicit `site_id = null` on resurrection is the right call — verified against the schema in `mysticat-data-service` that `uq_brand_name_per_org (organization_id, name)` is **not** partial, so a soft-deleted row genuinely is the ON CONFLICT target and must have its stale anchor cleared.

> Note on the earlier `.neq('status','deleted')` / NULL-status concern (also MysticatBot's nit #1): verified non-issue. `brands.status` is `NOT NULL DEFAULT 'active'` with a `CHECK (status IN ('pending','active','deleted'))` — no NULL-status row can exist, so the three-valued-logic edge cannot manifest. The filter is correct as written.

---

### Minor

**[MINOR] Concurrent resurrection without `baseSiteId` can race-null a freshly written anchor** (Confidence: low)
Two concurrent upserts for the same soft-deleted name can both see `existing === null` before either commits, and both proceed down the `existing === null` branch. A generic upsert TOCTOU with a narrow window; flagging for awareness, not blocking.
**Where:** `src/support/brands-storage.js` — `if (existing === null) { row.site_id = ... }`
**Fix (optional):** `COALESCE(EXCLUDED.site_id, brands.site_id)` in the conflict update, or accept as unlikely.

**[MINOR] `if (existing === null)` conflates fresh-create and soft-deleted resurrection without an inline label** (Confidence: high)
The block comment above explains both cases, but the branch itself isn't labelled — a future reader debugging the null-write has to read back to the filter comment.
**Where:** `src/support/brands-storage.js` — `if (existing === null)`
**Fix:** one-liner, e.g. `// fresh create OR name-matched soft-deleted brand — write explicit site_id to clear stale anchor on conflict`

**[MINOR] Unit mock proves `.neq()` was called, not that it filters** (Confidence: high)
The `neq` stub captures the call and returns `data: null` unconditionally — the same result with or without the filter. If `.neq()` were dropped or swapped to `.eq()`, the test would still pass. Inherent to mocking; PR already notes no IT run.
**Where:** `test/support/brands-storage.test.js` — `createCapturingClient` neq stub
**Fix:** cover in an IT test against real Postgres with a seeded soft-deleted row.

---

### Question

**[QUESTION] Does the `anchoredBySemrush = true` path leave a stale `site_id` on resurrection?** (Confidence: low)
When `anchoredBySemrush` is true the whole `if (!anchoredBySemrush)` block is skipped, so `row.site_id` is never written — the ON CONFLICT UPDATE then inherits the soft-deleted row's old `site_id`, which could re-trigger `brands_base_site_unique`. Likely a non-issue in practice, but worth confirming whether Semrush-anchored brands can be soft-deleted and re-upserted by name.

---

### What's working well
- Thorough, accurate root-cause analysis (LLMO-5556 interaction correctly traced)
- Explicit `site_id = null` on resurrection is the correct fix, confirmed against the non-partial unique constraint
- Two targeted tests cover both the fresh-create and resurrection scenarios; 100% line coverage (Codecov)
- Long-term: a partial unique index `WHERE status != 'deleted'` on `(organization_id, name)` (MysticatBot's suggestion #2) would let the DB enforce this natively — worth a follow-up ticket.

---

**Recommendation: Approve.** No critical or major findings after verification against the live schema. Remaining items are a narrow race, a speculative Semrush edge to confirm, and two cosmetic notes — all at author's discretion.

🤖 Multi-persona review (Architect · DBA · Security · Tester · Junior Dev) via Claude Code

PR #2723, opened 2026-06-29
feat(audit-policy): spacecat-api-service contract — policy CRUD + 501 scope stubs (SITES-47306)

iuliag:
## 1. Abstract
Implements the `/sites/{siteId}/audit-policy` contract (E1/E3) plus four granular array-mutation endpoints (add/remove exclusions, add/remove inclusions) and `/sites/{siteId}/audit-scope/*` OpenAPI `501` stubs (E4–E6) for the Audit Policy feature. Implementation is complete, including gated integration tests; they're pending enablement on a companion data-service change (see Deployment & merge order).

## 2. Reasoning
SITES-46352 pinned the customer-facing HTTP contract for Audit Policy — the surface the delegated UI builds against. SITES-47306 is the implementation ticket, building that surface over the B2 `audit_policy` table. SITES-48346 subsequently replaced the originally-specified full-replace `PUT` with granular array-mutation endpoints, per `mysticat-architecture` → `platform/design-audit-policy-array-mutation-endpoints.md` (PR #204): a full-array PUT risks silent data loss under concurrent editors (two callers racing to add/remove different elements), whereas add/remove-by-value is inherently commutative and safe to retry.

- `GET /sites/{siteId}/audit-policy` returns the current policy, or a synthetic `version: 0` default when unset (never `404`).
- `POST /sites/{siteId}/audit-policy/exclusions` and `.../exclusions/delete` add or remove one or more exclusion globs (set-union / set-difference; already-present or already-absent values are a no-op).
- `POST /sites/{siteId}/audit-policy/inclusions` and `.../inclusions/delete` do the same for manually-declared URLs (`manualUrls`).
- All four mutation endpoints share a single `mutateArray` engine: authz → entitlement check → validate body → read current row (or synthetic default) → compute new array → cap check → call `wrpc_upsert_audit_policy` with every other field echoed unchanged. The author attribute is derived server-side from the caller's IMS token — clients never supply it. There is no client-visible `expectedVersion`: on a version conflict the server transparently re-reads and retries up to 3 attempts before returning `409` with `currentVersion`.
- Array caps are `200` exclusion globs / `2000` manual URLs (tighter than the DB's current `1000`/`50000` — the original pair's worst-case body size exceeds this service's ~6MB Lambda synchronous-invocation payload ceiling; a follow-up mysticat-data-service migration to tighten the DB-side check is tracked separately).
- `GET /sites/{siteId}/audit-policy/revisions` returns policy history newest-first, cursor-paginated.
- `GET /sites/{siteId}/audit-scope/{pages,summary,sections}` return `501` for now; their real implementations land under separate tickets (B4 / B7) once those blocks are built.
- AuthZ: reads are gated on organization membership; writes additionally require ASO or LLMO product entitlement on the site.
- Routing is wired through the standard controller/route/FACS/required-capabilities layers so the new endpoints are reachable, access-controlled, and documented consistently with the rest of the API.
- **Conflict signaling uses SQLSTATE `40000`, not `40001` as originally planned.** PostgREST v14.4 (pinned by mysticat-data-service) hangs indefinitely on `RAISE ... ERRCODE = '40001'` due to a known upstream bug (hasql-transaction auto-retries `serialization_failure`; see `PostgREST/postgrest#3673`). `40000` (`transaction_rollback`, the standard class-40 parent, not a Postgres-only extension code) is the coordinated replacement, verified non-hanging and unused elsewhere in either repo. See the companion data-service PR for the matching RPC-side change.
- Gated integration tests exercising the real RPC — including the granular add/remove/version-bump paths — are included (`test/it/shared/tests/audit-policy.js`, `test/it/postgres/audit-policy.test.js`, 7 cases), wrapped in `describe.skip` since they depend on a data-service image that doesn't exist yet. Flipping them on is a follow-up commit once that image is cut and pinned.

## 4. Required information
- Jira / issue: [SITES-47306](https://jira.corp.adobe.com/browse/SITES-47306) (impl); [SITES-46352](https://jira.corp.adobe.com/browse/SITES-46352) (original spec); [SITES-48346](https://jira.corp.adobe.com/browse/SITES-48346) (array-mutation endpoint redesign)
- Spec: adobe/mysticat-architecture `platform/design-audit-policy-api-contract.md` (PR #184), superseded in part by `platform/design-audit-policy-array-mutation-endpoints.md` (PR #204)
- Implementation plans: `docs/plans/2026-06-29-sites-47306-api-contract-policy-crud.md`, `docs/plans/2026-07-20-sites-48346-array-mutation-endpoints.md` (this branch)

- `mysticat-data-service` — contract/runtime dependency: this PR calls `wrpc_upsert_audit_policy` and reads `audit_policy` / `audit_policy_revision`, created by B2 ([SITES-47301](https://jira.corp.adobe.com/browse/SITES-47301), merged) and extended (`p_expected_version`, `effective_at`) by [mysticat-data-service PR #755](https://github.com/adobe/mysticat-data-service/pull/755) on the matching `feat/sites-47306-api-contract` branch — implementation complete there too, pending review/merge.

## 7. Test plan
- (a) Full unit suite: 15404 passing. `docs:lint` / `docs:build` / `type-check` / `lint` all clean. Gated integration tests (7 cases covering synthetic-default GET, granular first-write, inclusions union, exclusions set-difference, cross-org 403, revision ordering, and 501 stubs) are written and confirmed to load/report correctly as skipped — they can't run for real until the data-service dependency below is available.
- (b) Per-environment, once the data-service dependency lands: exercise the four mutation endpoints and confirm the synthetic-default, no-op idempotency, cap enforcement, retry-then-409, and entitlement-gate behaviour described above on eph/dev ahead of exposing the contract to the delegated UI on stage/prod.

- [SITES-47301](https://jira.corp.adobe.com/browse/SITES-47301) (B2 impl) — **merged**: created the base `audit_policy` table + RPC.
- [mysticat-data-service PR #755](https://github.com/adobe/mysticat-data-service/pull/755) — depends-on, **implementation complete, pending review/merge**: provides `p_expected_version` and `audit_policy_revision.effective_at` that this PR's RPC call and conflict-mapping rely on.
- Order: (1) #755 merges → (2) a new data-service image is cut and pinned in `test/it/postgres/docker-compose.yml` → (3) this PR's gated integration tests are flipped from `describe.skip` to `describe` and verified green → (4) this PR merges.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```


iuliag:
@sandsinh, this is the PR implementing the API spec. It's still in progress, however if you can review it and let me know if you see something missing or adjusting depending on what the UI will need.

src/controllers/audit-policy.js
```diff
@@ -0,0 +1,271 @@
...
+  };
+  const ge = arr(b.exclusionGlobs ?? [], MAX_EXCLUSION_GLOBS, 'exclusionGlobs');
+  if (ge) {
+    return ge;
+  }
+  const mu = arr(b.manualUrls ?? [], MAX_MANUAL_URLS, 'manualUrls');
+  if (mu) {
+    return mu;
+  }
+  if (b.scopeConfig !== undefined && !isObject(b.scopeConfig)) {
```
MysticatBot: issue (blocking): `scopeConfig` and `lifecycleOverrides` pass only an `isObject()` check with no size constraint. A caller can write a multi-MB deeply-nested JSON tree that gets stored as JSONB. Since the write path requires only org membership + ASO/LLMO entitlement (not admin), any entitled user can do this.

Add a `JSON.stringify(obj).length` ceiling (e.g., 64 KB) in `validatePolicyBody` for both fields:
```js
if (JSON.stringify(b.scopeConfig).length > 65536) return 'scopeConfig exceeds maximum size (64 KB)';
```

src/controllers/audit-policy.js
```diff
@@ -0,0 +1,271 @@
...
+    return 'scopeConfig must be an object';
+  }
+  if (b.lifecycleOverrides !== undefined && !isObject(b.lifecycleOverrides)) {
+    return 'lifecycleOverrides must be an object';
+  }
+  if (b.note !== undefined && b.note !== null
+    && (typeof b.note !== 'string' || b.note.length > MAX_NOTE_LEN)) {
+    return `note must be a string <= ${MAX_NOTE_LEN} chars`;
+  }
+  if (!hasText(b.reason)) {
```
MysticatBot: issue (blocking): `reason` is validated as non-empty (`hasText`) but has no max length. Every other string field has a ceiling (`note`: 2000 chars, array entries: 2048 chars). Each write creates a revision row, so a multi-MB `reason` accumulates storage on every policy change.

Add a length cap consistent with the other fields:
```js
if (!hasText(b.reason) || b.reason.length > 2000) {
  return 'reason is required and must be <= 2000 chars';
}
```

src/controllers/audit-policy.js
```diff
@@ -0,0 +1,271 @@
...
+    throw e;
+  }
+}
+
+// returns a string error message, or null when valid
+function validatePolicyBody(b) {
+  if (!isObject(b)) {
+    return 'request body must be a JSON object';
+  }
+  if (!Number.isInteger(b.budget) || b.budget <= 0) {
```
MysticatBot: issue (blocking): `budget` validates `> 0` but has no upper bound. `Number.MAX_SAFE_INTEGER` would pass. The OpenAPI spec also declares only `minimum: 1` with no `maximum`.

Add a `MAX_BUDGET` constant and validate against it:
```js
const MAX_BUDGET = 500_000; // adjust to the product's real ceiling
if (!Number.isInteger(b.budget) || b.budget <= 0 || b.budget > MAX_BUDGET) {
  return `budget must be an integer between 1 and ${MAX_BUDGET}`;
}
```
Document the ceiling in the OpenAPI schema as well.

src/controllers/audit-policy.js
```diff
@@ -0,0 +1,271 @@
...
+        return createResponse(
+          {
+            message: 'policy was modified; reload and retry',
+            ...(Number.isInteger(currentVersion) ? { currentVersion } : {}),
+          },
+          409,
+        );
+      }
+      if (error.code === 'P0001') {
+        return badRequest(error.message || 'audit policy rejected by validation');
```
MysticatBot: issue (blocking): `error.message` from a PostgreSQL `RAISE EXCEPTION` (P0001) is forwarded verbatim to the client. If the RPC raises an unintended P0001 with diagnostic details (table names, constraint names, query fragments), those leak to the caller.

Return a generic message and log the real one server-side:
```js
context.log?.warn?.(`audit-policy putPolicy P0001: ${error.message}`);
return badRequest('audit policy rejected by validation');
```
If you later establish a structured sub-code contract with the RPC (known safe messages), surface those selectively via an allowlist.

docs/openapi/audit-policy-api.yaml
```diff
@@ -0,0 +1,199 @@
...
+        content:
+          application/json:
+            schema:
+              $ref: '#/AuditPolicyConflict'
+      '500': { $ref: './responses.yaml#/500' }
+    security:
+      - ims_key: [ ]
+      - scoped_api_key: [ ]
+
+audit-policy-revisions:
```
MysticatBot, line 183: issue (blocking): The `listAuditPolicyRevisions` operation does not document `limit` or `cursor` query parameters, yet the controller reads and acts on them. API consumers relying on the spec (and codegen tools) will not discover pagination support.

Add `parameters:` entries for `limit` (integer, optional, default 50, max 200) and `cursor` (string, optional, opaque) to this operation.

MysticatBot (changes requested):
Hey @iuliag,

**Verdict: Request changes** - input validation gaps and an info-disclosure path need hardening before this surface goes to GA.
**Complexity: HIGH** - large diff; API surface + shared contract.
_Changes: Adds audit policy CRUD endpoints (GET/PUT policy, GET revisions) with optimistic locking, plus 501 scope stubs, wired through the standard controller/route/FACS/capabilities layers (15 files)._

## Must fix before merge
1. [Important] `scopeConfig`/`lifecycleOverrides` accept arbitrary JSON with no size limit - `src/controllers/audit-policy.js:88` (details inline)
2. [Important] `reason` field has no max length, unlike `note` (capped at 2000) - `src/controllers/audit-policy.js:98` (details inline)
3. [Important] `budget` has no upper bound; only checks > 0 - `src/controllers/audit-policy.js:62` (details inline)
4. [Important] P0001 `error.message` forwarded verbatim to client - potential info disclosure - `src/controllers/audit-policy.js:220` (details inline)
5. [Important] Revisions endpoint missing `limit`/`cursor` query parameter documentation in OpenAPI spec - `docs/openapi/audit-policy-api.yaml:64` (details inline)

<details>

- suggestion: Invalid/out-of-range cursor silently falls back to page 1 instead of returning 400 - consider making the reset explicit so clients know their cursor was stale - `src/controllers/audit-policy.js:108`
- suggestion: Add a null-guard on the RPC success path (`if (!data) return internalServerError(...)`) to avoid producing `{ siteId: undefined, ... }` if the function signature ever changes - `src/controllers/audit-policy.js:218`
- suggestion: Match both SQLSTATE `40000` and `40001` in the conflict branch for forward-compatibility with a future PostgREST upgrade - `src/controllers/audit-policy.js:35`
- nit: Plan document (Decision 5) says `40001` but implementation uses `40000` - consider updating the plan to reflect the final decision - `docs/plans/2026-06-29-sites-47306-api-contract-policy-crud.md`
- suggestion: Add a unit test for `putPolicy` when the RPC returns an unrecognized error code (e.g., `23505`) to explicitly cover the generic 500 path - `test/controllers/audit-policy.test.js`
- suggestion: Add a test case for 409 when `error.details` is not parseable (e.g., `null`) to verify `currentVersion` is omitted correctly - `test/controllers/audit-policy.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 55s | <b>Cost:</b> $5.57 | <b>Commit:</b> <code>60e0ef633b293664be6fc9bc7e204585443913f6</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

iuliag:
## Action Plan - PR #2723 (adobe/spacecat-api-service)

### Summary
11 findings from 1 reviewer (1 AI agent - MysticatBot / review-kit pr-review skill).
10 to implement across 3 batches, 1 pushed back, 0 deferred, 0 conflicts.

### Batch 1 - Important (must-fix, implement first)
- [ ] F01: `budget` has no upper bound (only checks `> 0`) - `src/controllers/audit-policy.js:62`
  - Reviewers: MysticatBot
  - Source: `3598507937` (`ref_type: review_comment`)
  - What to do: Add a `MAX_BUDGET` constant and validate against it in `validatePolicyBody`; document the ceiling in the OpenAPI schema (`maximum` on `budget` in both `AuditPolicy` and `AuditPolicyUpsertRequest`).

- [ ] F02: `scopeConfig`/`lifecycleOverrides` accept arbitrary JSON with no size limit - `src/controllers/audit-policy.js:88`
  - Reviewers: MysticatBot
  - Source: `3598507925` (`ref_type: review_comment`)
  - What to do: Add a serialized-size ceiling (64 KB) check for both fields in `validatePolicyBody`.

- [ ] F03: `reason` field has no max length, unlike `note` (capped at 2000) - `src/controllers/audit-policy.js:98`
  - Reviewers: MysticatBot
  - Source: `3598507928` (`ref_type: review_comment`)
  - What to do: Add a 2000-char cap on `reason` in `validatePolicyBody`, matching `note`.

- [ ] F04: `P0001` `error.message` forwarded verbatim to client - potential info disclosure - `src/controllers/audit-policy.js:220`
  - Reviewers: MysticatBot
  - Source: `3598507943` (`ref_type: review_comment`)
  - What to do: Return a generic `badRequest` message on `P0001`, log the real `error.message` server-side via `context.log.warn`.

- [ ] F05: Revisions endpoint missing `limit`/`cursor` query parameter documentation in OpenAPI spec - `docs/openapi/audit-policy-api.yaml:64`
  - Reviewers: MysticatBot
  - Source: `3598507950` (`ref_type: review_comment`)
  - What to do: Add `parameters:` entries for `limit` (integer, optional, default 50, max 200) and `cursor` (string, optional, opaque) to the `listAuditPolicyRevisions` operation.

### Batch 2 - Minor (non-blocking code/doc changes)
- [ ] F06: Invalid/out-of-range cursor silently falls back to page 1 instead of returning 400 - `src/controllers/audit-policy.js:108`
  - Reviewers: MysticatBot
  - Source: review body (no inline comment id)
  - What to do: Return `400` from `listRevisions` when a non-empty `cursor` param fails to decode, instead of silently treating it as "no cursor".

- [ ] F08: Match both SQLSTATE `40000` and `40001` in the conflict branch for forward-compatibility - `src/controllers/audit-policy.js:35`
  - Reviewers: MysticatBot
  - Source: review body (no inline comment id)
  - What to do: Widen the version-conflict check to accept either code (does not affect the RPC-hang bug this constant works around, since that bug is in PostgREST's own internal retry on the code returned by the RPC call itself, not in this response-mapping check).

- [ ] F09: Plan document (Decision 5) says `40001` but implementation uses `40000` - `docs/plans/2026-06-29-sites-47306-api-contract-policy-crud.md`
  - Reviewers: MysticatBot
  - Source: review body (no inline comment id)
  - What to do: Update Decision 5 to state the final `40000` token and reference the hasql-transaction hang rationale already in the code comment. Leave the historical step-by-step TDD snippets elsewhere in the doc untouched (they narrate what was actually done at each step).

### Batch 3 - Minor (test coverage)
- [ ] F10: Add a unit test for `putPolicy` when the RPC returns an unrecognized error code (e.g., `23505`) - `test/controllers/audit-policy.test.js`
  - Reviewers: MysticatBot
  - Source: review body (no inline comment id)
  - What to do: Add a test asserting the generic 500 path for an unrecognized RPC error code.

- [ ] F11: Add a test case for 409 when `error.details` is not parseable (e.g., `null`) - `test/controllers/audit-policy.test.js`
  - Reviewers: MysticatBot
  - Source: review body (no inline comment id)
  - What to do: Add a test asserting `currentVersion` is omitted from the 409 body when `error.details` is `null`.

### Pushback List
These findings will not be implemented. Reasons will be posted as replies to the relevant comment threads.

| F07: Add a null-guard on the RPC success path (`if (!data) return internalServerError(...)`) to avoid `{ siteId: undefined, ... }` - `src/controllers/audit-policy.js:218` | MysticatBot | The reviewer's own framing is speculative ("if the function signature ever changes"). `wrpc_upsert_audit_policy` is a documented, cross-repo-pinned contract (`mysticat-data-service/docs/llmo-database-schema.md:1151`) that guarantees a returned row on success - the same kind of pinned token already relied on for the SQLSTATE decision. Guarding against a hypothetical future contract break adds dead code for a scenario that can't currently happen, which this project's conventions explicitly avoid. |

src/controllers/audit-policy.js
iuliag: Fixed in 152f6e74b. budget now validated against a MAX_BUDGET (500,000) ceiling, documented as maximum in the OpenAPI schema.

src/controllers/audit-policy.js
iuliag: Fixed in 152f6e74b. scopeConfig and lifecycleOverrides are now capped at 64 KB serialized size in validatePolicyBody.

src/controllers/audit-policy.js
iuliag: Fixed in 152f6e74b. reason now has a 2000-char cap in validatePolicyBody, matching note.

src/controllers/audit-policy.js
iuliag: Fixed in 152f6e74b. P0001 now returns a generic badRequest message; the raw error.message is logged server-side via context.log.warn instead.

docs/openapi/audit-policy-api.yaml
iuliag: Fixed in 152f6e74b. listAuditPolicyRevisions now documents limit (integer, default 50, max 200) and cursor (opaque string) query params in the OpenAPI spec.

iuliag:
## Review Response Summary

Commit: 152f6e74b

### Implemented (10 findings)
- F01: `budget` now capped at `MAX_BUDGET` (500,000); documented as `maximum` in the OpenAPI schema
- F02: `scopeConfig`/`lifecycleOverrides` now capped at 64 KB serialized size
- F03: `reason` now capped at 2000 chars, matching `note`
- F04: `P0001` no longer forwards the raw RPC `error.message` to the client; it's logged server-side instead
- F05: `listAuditPolicyRevisions` now documents `limit`/`cursor` query params in the OpenAPI spec
- F06: an invalid/out-of-range `cursor` now returns `400` instead of silently falling back to page 1
- F08: the version-conflict check now accepts SQLSTATE `40001` alongside `40000`, for forward-compat with a future PostgREST upgrade
- F09: plan doc Decision 5 updated to reflect the `40000` token actually shipped
- F10: added a test for `putPolicy` returning a generic 500 on an unrecognized RPC error code
- F11: added a test for the 409 response omitting `currentVersion` when `error.details` isn't parseable

### Pushed Back (1 finding)
- F07 (null-guard on `putPolicy`'s success path, `if (!data) return internalServerError(...)`): not implemented. `wrpc_upsert_audit_policy` is a documented, cross-repo-pinned contract (`mysticat-data-service/docs/llmo-database-schema.md:1151`) that guarantees a returned row on success, the same kind of pinned token already relied on for the SQLSTATE decision. Guarding against a hypothetical future contract break adds dead code for a scenario that can't currently happen.

Full test suite (15219 tests), lint, and `docs:lint` all green.

MysticatBot (approved):
Hey @iuliag,

**Verdict: Approve** - all prior blocking findings are addressed; the fix commit is clean and well-scoped.
**Complexity: HIGH** - large diff; API surface + shared contract.
_Changes: Adds audit policy CRUD endpoints (GET/PUT policy, GET revisions) with optimistic locking, plus 501 scope stubs, wired through the standard controller/route/FACS/capabilities layers (15 files)._

- Budget now capped at MAX_BUDGET (500,000) with OpenAPI `maximum` documented
- scopeConfig/lifecycleOverrides capped at 64 KB serialized size
- reason capped at 2000 chars, matching note
- P0001 no longer forwards raw RPC error.message to the client; logged server-side via context.log.warn
- Revisions endpoint now documents limit/cursor query params in the OpenAPI spec
- Invalid/out-of-range cursor now returns 400 instead of silently falling back to page 1
- Version-conflict check accepts both SQLSTATE 40000 and 40001 for forward-compat
- Plan doc Decision 5 updated to reflect the 40000 token actually shipped
- Test added for generic 500 on unrecognized RPC error code
- Test added for 409 body omitting currentVersion when error.details is not parseable

<details>

- nit: The `UnauthorizedProductError` import is used only inside `hasProductAccess` - consider a lazy import or a comment noting it powers the product-entitlement fallthrough - `src/controllers/audit-policy.js:18`
- suggestion: The `MAX_CURSOR_VERSION = 1_000_000` ceiling is a good belt-and-suspenders guard; consider documenting it in the OpenAPI cursor description so API consumers know cursors are not indefinitely reusable across very long histories - `docs/openapi/audit-policy-api.yaml:90`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 5m 11s | <b>Cost:</b> $1.90 | <b>Commit:</b> <code>152f6e74b24260b7523b925e6ee710e76d5a8435</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

iuliag:
## Review Response Summary (re-review)

Re-review verdict: **Approve** (review id 4717490401) — all prior blocking findings confirmed resolved. 2 new non-blocking suggestions triaged below.

Commit: 61e4ac440

### Implemented (1 finding)
- Documented that `listAuditPolicyRevisions` cursors encode a version number and aren't valid indefinitely across a very long revision history, matching the `MAX_CURSOR_VERSION` guard already in the controller (`docs/openapi/audit-policy-api.yaml`).

### Pushed Back (1 finding)
- Suggestion to lazy-import or comment on `UnauthorizedProductError` (`src/controllers/audit-policy.js:18`, single-use import): not implemented. The adjacent `hasProductAccess` doc comment (lines 46-47) already explains why the error needs to be distinguished from a real failure for the ASO/LLMO entitlement fallthrough; a lazy `import()` for a small synchronously-used error class isn't an established pattern in this codebase and would add indirection without benefit.

`docs:lint`, full test suite, and lint all green.

src/routes/facs-capabilities.js
```diff
@@ -935,6 +935,10 @@ const routeFacsCapabilities = {
       'PATCH /trial-users/email-preferences': 'aso/can_configure',
       'POST /sites/:siteId/traffic/predominant-type': 'aso/can_configure',
       'POST /sites/:siteId/traffic/predominant-type/:channel': 'aso/can_configure',
+      'POST /sites/:siteId/audit-policy/exclusions': 'aso/can_configure',
```
MysticatBot, line 958: issue (blocking): All nine audit-policy routes are registered only under `PRODUCTS_ROUTES.ASO`. The controller explicitly grants write access to users with LLMO entitlement (`hasProductAccess(ac, site, 'LLMO')`), but `facsWrapper` will reject an LLMO user sending `x-product: LLMO` with 403 before the controller ever runs - the route does not exist in `PRODUCTS_ROUTES.LLMO`.

This is an over-restriction (not a security hole), but it means the LLMO entitlement fallback in the controller is dead code for any FACS-enrolled org. Either:
- (a) Add the 9 routes to `PRODUCTS_ROUTES.LLMO` with `llmo/can_configure` (writes) and `llmo/can_view` (reads) - both capabilities are already in `PRODUCTS_CAPABILITIES.LLMO`, so no catalog change needed.
- (b) If LLMO users intentionally should not access audit-policy, remove the LLMO fallback from the controller to avoid misleading future readers.

src/controllers/audit-policy.js
```diff
@@ -0,0 +1,297 @@
...
+      context.log?.error?.(`audit-policy getPolicy failed: ${error.code} ${error.message}`);
+      return internalServerError('Failed to read audit policy');
+    }
+    if (!data) {
+      return ok(AuditPolicyDto.defaultDocument(siteId));
+    }
+    return ok(AuditPolicyDto.toJSON(data));
+  }
+
+  const RESOURCE_CONFIG = {
```
MysticatBot, line 121: issue (blocking): `validateMutateBody` checks that `values` is non-empty and each string is <= 2048 chars, but does not cap the array length. A caller can submit a request with tens of thousands of entries. The cap check after `computeNewArray` prevents persistence of oversized arrays, but the request still forces the server to parse, validate (`.some()` over all strings), allocate a Set, and spread it back - an amplification vector for resource exhaustion.

Fix: add an early length check in `validateMutateBody`:
```js
if (b.values.length > config.max) {
  return `values array exceeds the maximum of ${config.max} entries`;
}
```
This requires passing `config.max` into the validator, or using `Math.max(MAX_EXCLUSION_GLOBS, MAX_MANUAL_URLS)` (2000) as a universal cap.

MysticatBot (changes requested):
Hey @iuliag,

**Verdict: Request changes** - FACS routing gap means LLMO users are silently blocked from these endpoints despite the controller accepting their entitlement.
**Complexity: HIGH** - large diff; API surface + shared contract.
_Changes: Replaces the original full-replace PUT with four granular array-mutation endpoints (add/remove exclusions, add/remove inclusions), retaining GET policy, GET revisions, and 501 scope stubs (16 files)._

## Must fix before merge
1. [Important] FACS capability gap: audit-policy routes missing from LLMO product sub-map - `src/routes/facs-capabilities.js:935` (details inline)
2. [Important] No upper bound on `values` array length allows resource-exhaustion payloads - `src/controllers/audit-policy.js:121` (details inline)

<details>

- suggestion: `getAuthor` falls back to `'system'` when no identity is resolved - consider returning an error instead of silently attributing mutation writes to a system identity, since this weakens the audit trail - `src/controllers/audit-policy.js:63`
- suggestion: Add basic glob pattern validation for `exclusionGlobs` values - at minimum reject path traversal sequences (`../`) to prevent downstream audit-engine misinterpretation - `src/controllers/audit-policy.js:121`
- nit: The short-circuit `const llmo = aso ? true : await hasProductAccess(...)` is misleading - `llmo` does not mean "has LLMO access" when `aso` is true; consider renaming to `hasWriteEntitlement` or restructuring the conditional - `src/controllers/audit-policy.js:160`
</details>

- Budget now capped at MAX_BUDGET (500,000) with OpenAPI maximum documented
- scopeConfig/lifecycleOverrides capped at 64 KB serialized size
- reason capped at 2000 chars, matching note
- P0001 no longer forwards raw RPC error.message to the client
- Revisions endpoint documents limit/cursor query params in the OpenAPI spec
- Invalid/out-of-range cursor returns 400 instead of silently falling back
- Version-conflict check accepts both SQLSTATE 40000 and 40001
- Plan doc updated to reflect the 40000 code actually shipped
- Tests added for generic 500 and unparseable 409 details

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 7s | <b>Cost:</b> $9.19 | <b>Commit:</b> <code>1e457a111b4d4179d9e54de2f71380319cab2f33</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @iuliag,

**Verdict: Approve** - all prior blocking findings are addressed; the fix commits are well-scoped and the new array-mutation design is sound.
**Complexity: HIGH** - large diff; API surface + shared contract.
_Changes: Replaces the original full-replace PUT with four granular array-mutation endpoints (add/remove exclusions, add/remove inclusions), retaining GET policy, GET revisions, and 501 scope stubs, wired through the standard controller/route/FACS/capabilities layers (16 files)._

- FACS capability gap closed: all nine audit-policy routes now registered under both ASO and LLMO product sub-maps with appropriate capabilities
- Values array length now capped upfront in validateMutateBody (receives config.max), rejecting oversized payloads before any read/compute work
- Budget capped at MAX_BUDGET (500,000) with OpenAPI maximum documented
- scopeConfig/lifecycleOverrides capped at 64 KB serialized size
- reason capped at 2000 chars, matching note
- P0001 no longer forwards raw RPC error.message to the client
- Revisions endpoint documents limit/cursor query params in the OpenAPI spec
- Invalid/out-of-range cursor returns 400 instead of silently falling back
- Version-conflict check accepts both SQLSTATE 40000 and 40001
- Plan doc updated to reflect the 40000 code actually shipped
- Tests added for generic 500 and unparseable 409 details

<details>

- nit: The path-traversal check (`v.includes('../')`) catches literal sequences but not encoded variants (`..%2f`, `..\`). Since these globs are evaluated by an audit engine downstream rather than used for filesystem access here, the risk is low, but consider a regex anchor or canonicalization if the downstream engine decodes before matching - `src/controllers/audit-policy.js:175`
- suggestion: The `AuditPolicyMutateRequest` schema in the OpenAPI spec declares `minItems: 1` but no `maxItems`. Adding `maxItems: 200` (exclusions) or documenting that the cap is resource-dependent would help codegen consumers discover the limit without hitting a 400 - `docs/openapi/audit-policy-api.yaml:308`
- nit: The `removeExclusions` path also rejects `../` in values. Since the remove path only filters existing entries (set-difference), a path-traversal value in a remove request is harmless (it just won't match anything). Consider skipping the check on remove to avoid surprising callers who are trying to clean up a value that was stored before the check existed - `src/controllers/audit-policy.js:175`
</details>

---

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 6m 4s | <b>Cost:</b> $3.17 | <b>Commit:</b> <code>2648582def9a48bb1488e36ebe39dac54812f7bb</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

iuliag:
## Merge readiness

Branch is current with `main` (merged twice more since the last bot review, no conflicts beyond the generated `docs/index.html`, resolved by rebuilding). Status as of `209408ad8`:

- **Reviews:** `reviewDecision: APPROVED`. The last bot pass (against `2648582d`) verdict was **Approve** — all earlier "Must fix" items (FACS routing gap for LLMO entitlement, unbounded `values` array) were already resolved by that point. Its only remaining findings were 3 non-blocking nits (path-traversal not catching encoded/backslash variants, `removeExclusions` needlessly rejecting `../`, OpenAPI missing a documented cap) — all three are fixed in a later commit (`07b1c3f4c`), so current HEAD has strictly fewer open findings than what was already approved.
- **Review threads:** 7 total, 0 unresolved.
- **CI on the latest push (`209408ad8`):** all green — `type-check` ✓, `build` (lint/unit tests/docs/bundle smoke) ✓, `it-postgres` ✓, `branch-deploy` ✓ (deployed this branch to the `dev-branches` environment and its own Post-Deployment Integration Test step passed). `Adobe CLA Signed`, `Kodiak`, `validate-pr-title`, `protect-nyc-config` all pass. `deploy-stage`/`semantic-release` correctly show `skipping` (those only run on push to `main`).
- **Mergeable:** `MERGEABLE` per GitHub.
- **Upstream dependency:** `mysticat-data-service` PR #755 (adds `p_expected_version` + `audit_policy_revision.effective_at`) merged 2026-07-17. The data-service image pin in `test/it/postgres/docker-compose.yml` has been bumped past it (`v5.57.0` → `v5.70.0`), and the previously-gated real-DB IT suite (`test/it/shared/tests/audit-policy.js`) now runs for real in CI and passed all 7 cases (confirmed in run [29920493349](https://github.com/adobe/spacecat-api-service/actions/runs/29920493349), and again on this latest push).

**Assessment: ready to merge.**

## Validating after merge

### 1. Automatic gates the release pipeline already runs
On merge to `main`, `mysticat-ci`'s reusable workflow re-runs `build` + `it-postgres`, then `deploy-stage` and the prod `semantic-release` job deploy the Lambda and run `npm run test-postdeploy` as part of `prepareCmd`. Note: this repo currently has **no test files tagged `Post-Deploy`**, so that step is a no-op today — it's not a real smoke gate yet, don't rely on it alone.

### 2. Confirm no regressions
- Watch the `deploy-stage` and prod deploy jobs on the merge commit (`gh run watch <run-id>`) — a VPC-attachment/Lambda-update failure here would be the first sign of trouble.
- Since dev/stage/prod all auto-replicate from the same `mysticat-data-service` ECR image (already past `v5.70.0`, confirmed via `mysticat-data-service` release history), the exact `wrpc_upsert_audit_policy` behavior this PR depends on is already live in every environment — no separate "wait for the data-service to catch up" step needed post-merge.
- Spot-check a handful of pre-existing, unrelated endpoints (e.g. `GET /sites/{siteId}`, an existing opportunities/suggestions read) against dev right after deploy, to catch any incidental regression from the merge itself rather than from this PR's own code.
- Check Coralogix for new error-rate spikes on the `spacecat-services--api-service` Lambda in the 30–60 min after deploy (the `mysticat-debug` skill wraps exactly this correlation if anything looks off).

### 3. Confirm the API works as intended
- Use the **`validate-migration-in-dev`** skill (matches this PR's dependency directly) to validate `wrpc_upsert_audit_policy`'s `p_expected_version` path end-to-end on dev, then extend read-only to stage/prod after promotion.
- Exercise the four new endpoints directly against dev with a real IMS token:
  - `POST /sites/{siteId}/audit-policy/exclusions` and `.../inclusions` — add, confirm set-union/no-op-on-duplicate, confirm the 200-entry / 2000-entry caps 400 correctly, confirm a `../` glob 400s on exclusions but the same value is accepted on inclusions.
  - `.../exclusions/delete` and `.../inclusions/delete` — confirm set-difference, confirm a stored `../` value can still be removed (the fix from `07b1c3f4c`).
  - `GET /sites/{siteId}/audit-policy` and `.../revisions` — confirm the synthetic `version: 0` default and newest-first cursor pagination.
  - Repeat once with an ASO-entitled token and once with an LLMO-entitled token (no ASO) to confirm the FACS routing fix actually unblocks LLMO callers end-to-end, not just at the unit-test level.
- The repo's own real-DB IT suite (`test/it/postgres/audit-policy.test.js`) is the fastest local repro of all of the above without needing a live IMS token — `npx mocha --require test/it/postgres/harness.js --timeout 30000 test/it/postgres/audit-policy.test.js`.

PR #2728, opened 2026-06-30
feat(serenity): gate serenity routes on an org-wide LLMO/serenity feature flag

rainer-friederich:
## 1. Abstract
Gates the `/serenity/*` surface on a new org-wide `LLMO/serenity` feature flag via a central `isSerenityActiveForOrg` helper, defaulting OFF.

## 2. Reasoning
Today "is serenity active" is inferred implicitly from a `semrush_workspace_id` being present on the org / brand. For the rollout we need to provision Semrush workspaces ahead of time **without** flipping the customer UI onto serenity-backed data. An explicit, per-org feature flag decouples rollout from provisioning: an org can be wired to Semrush in the background while its UI keeps reading the normal backend data, until the flag is flipped on for that org.

- Adds `isSerenityActiveForOrg(ctx, spaceCatId, log)` (`src/support/serenity/serenity-active.js`) — the single source of truth. It reads the `feature_flags` row keyed `(organization_id, product='LLMO', flag_name='serenity')`, cached the same way as the workspace resolver (5-minute positive TTL, 30-second negative TTL so an ON-flip propagates fast).
- `SerenityController.authorize` now rejects the serenity surface with `404 Serenity is not active for this organization` when the flag is OFF. The check runs **before** brand resolution, so an inactive org never leaks brand existence.
- **Effective gate = flag AND workspace.** The helper is the org-wide flag half; the existing per-brand workspace resolution (`resolveBrandWorkspace`) supplies the "AND a Semrush workspace resolves" half. A route is served only when both hold.
- **Default OFF.** A missing/`false` flag row, an unavailable PostgREST client, or a transient read error all resolve to inactive — so an org with a backfilled `semrush_workspace_id` keeps reading normal backend data until the flag is flipped.
- The brand-independent catalogue routes (`GET /serenity/models`, `GET /serenity/languages`, no `:brandId`) are intentionally **not** gated — the add-brand wizard needs them before a workspace or the flag exists.
- Brand creation is unchanged: it already decides "Semrush mode" from the request payload (`semrushMarket`/`generatePrompts`), not from the org flag, so it follows automatically once the UI gates on the flag.
- Flip per org with the existing admin endpoint: `PUT`/`DELETE /organizations/:id/feature-flags/llmo/serenity`.

## 4. Required information
- Other: docs/serenity.md "Serenity activation flag (org-wide rollout switch)" section (added in this PR)

- project-elmo-ui (consumer / contract): to realize the rollout, the customer UI must gate its serenity views on the `LLMO/serenity` flag (readable via the existing `GET /organizations/:id/feature-flags?product=LLMO` endpoint) instead of the current LaunchDarkly flag. Not required for this PR to be safe — default-OFF means no behaviour change until a flag is flipped — but it is the follow-up that makes the switch drive the UI.

## 7. Test plan
(a) Local: the PostgreSQL serenity integration suite was not executed this session (it requires Docker + ECR access). The IT seed now provisions the `LLMO/serenity` flag ON for ORG_1, so that suite exercises the active path end-to-end when run in CI; the inactive path and the helper's cache/error behaviour are covered by the controller, OpenAPI-contract, and helper unit suites.

(b) Per-environment verification (dev / stage / prod):
1. Pick an org that has serenity provisioned (a brand in sub-workspace mode, or an org with a `semrush_workspace_id`).
2. With no flag row, a brand-level serenity call (e.g. `GET /v2/orgs/:org/brands/:brand/serenity/markets`) returns `404 Serenity is not active for this organization`.
3. Enable: `PUT /organizations/:org/feature-flags/llmo/serenity` `{ "value": true }` (admin). Within ~30s (negative cache TTL) the same call now serves live data.
4. Disable: `DELETE /organizations/:org/feature-flags/llmo/serenity`. The call returns to `404` once the positive cache window (~5 min) elapses on warm Lambda containers — note this propagation delay when toggling OFF.

### DIFF
--- a/src/controllers/serenity.js
+++ b/src/controllers/serenity.js
@@ -56,6 +56,7 @@ import {
   handleBulkDeletePromptsSubworkspace,
 } from '../support/serenity/handlers/prompts-subworkspace.js';
 import { ensureSubworkspace, decommissionBrandWorkspace } from '../support/serenity/workspace-lifecycle.js';
+import { isSerenityActiveForOrg } from '../support/serenity/serenity-active.js';
 import { MAX_TOPICS_ON_CREATE } from '../support/serenity/brand-provisioning.js';
 import { STANDARD_PROMPT_TAGS, PROJECT_STANDARD_TAGS } from '../support/serenity/prompt-tags.js';
 import AccessControlUtil from '../support/access-control-util.js';
@@ -304,6 +305,18 @@ function SerenityController(context, log, env) {
           503,
         ),
       };
+    }
+    // Org-wide serenity rollout gate. Serenity is "active" for an org only when
+    // its `LLMO/serenity` feature flag is ON *and* a Semrush workspace resolves
+    // for the brand (the workspace half is enforced below by
+    // resolveBrandWorkspace). While the flag is OFF the org's UI keeps reading
+    // the normal backend data — even if a `semrush_workspace_id` has already
+    // been backfilled for rollout prep — so reject the serenity surface with a
+    // 404 (the same "no serenity for this org" contract the UI already handles
+    // for an org without a workspace). Checked before brand resolution so an
+    // inactive org never leaks brand existence.
+    if (!await isSerenityActiveForOrg(ctx, spaceCatId, log)) {
+      return { error: notFound('Serenity is not active for this organization') };
     }
     const brandUuid = await resolveBrandUuid(spaceCatId, brandId, postgrestClient);
     if (!brandUuid) {
--- a/docs/openapi/brands-v2-api.yaml
+++ b/docs/openapi/brands-v2-api.yaml
@@ -82,7 +82,12 @@ v2-brands-for-org:
       '400':
         $ref: './responses.yaml#/400'
       '403':
-        description: User does not have access to this organization
+        description: |
+          Forbidden. Either the user does not have access to this organization, OR
+          a Semrush-mode create (a picked `semrushMarket` or `generatePrompts`) was
+          requested while the org-wide serenity feature flag (`LLMO/serenity`) is
+          inactive — provisioning a sub-workspace is refused until the org's rollout
+          flag is on (body message: "Serenity is not active for this organization").
       '404':
         $ref: './responses.yaml#/404-organization-not-found-with-id'
       '409':
@@ -176,7 +181,13 @@ v2-brand-for-org:
       '400':
         $ref: './responses.yaml#/400'
       '403':
-        description: User does not have access to this organization
+        description: |
+          Forbidden. Either the user does not have access to this organization, OR
+          the edit would re-sync a sub-workspace brand to Semrush (a URL, competitor,
+          or alias change on a brand carrying a `semrushWorkspaceId`) while the
+          org-wide serenity feature flag (`LLMO/serenity`) is inactive — the edit is
+          refused rather than left to drift (body message: "Serenity is not active
+          for this organization").
       '404':
         description: Organization or brand not found
       '409':
--- a/docs/openapi/serenity-api.yaml
+++ b/docs/openapi/serenity-api.yaml
@@ -89,7 +89,7 @@ v2-serenity-prompts:
             schema: { $ref: './schemas.yaml#/SerenityPromptListResponse' }
       '400': { $ref: './responses.yaml#/400' }
       '404':
-        description: Organization, brand, or workspace not found.
+        description: Organization, brand, or workspace not found, or serenity is not active for the organization.
       '502':
         description: Upstream returned a non-2xx response.
         content:
@@ -119,7 +119,7 @@ v2-serenity-prompts:
             schema: { $ref: './schemas.yaml#/SerenityCreatePromptsResponse' }
       '400': { $ref: './responses.yaml#/400' }
       '404':
-        description: Organization, brand, or workspace not found.
+        description: Organization, brand, or workspace not found, or serenity is not active for the organization.
       '502':
         description: Upstream returned a non-2xx response.
         content:
@@ -180,7 +180,7 @@ v2-serenity-prompt-by-id:
             schema: { $ref: './schemas.yaml#/SerenityPrompt' }
       '400': { $ref: './responses.yaml#/400' }
       '404':
-        description: Organization, brand, slice, or prompt not found.
+        description: Organization, brand, slice, or prompt not found, or serenity is not active for the organization.
       '502':
         description: Upstream returned a non-2xx response.
         content:
@@ -230,7 +230,7 @@ v2-serenity-prompts-bulk-delete:
             schema: { $ref: './schemas.yaml#/SerenityBulkDeletePromptsResponse' }
       '400': { $ref: './responses.yaml#/400' }
       '404':
-        description: Organization, brand, or workspace not found.
+        description: Organization, brand, or workspace not found, or serenity is not active for the organization.
       '502':
         description: Upstream returned a non-2xx response.
         content:
@@ -273,7 +273,7 @@ v2-serenity-markets:
             schema: { $ref: './schemas.yaml#/SerenityMarketListResponse' }
       '400': { $ref: './responses.yaml#/400' }
       '404':
-        description: Organization, brand, or workspace not found.
+        description: Organization, brand, or workspace not found, or serenity is not active for the organization.
       '502':
         description: Upstream returned a non-2xx response.
         content:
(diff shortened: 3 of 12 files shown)


src/controllers/serenity.js
```diff
@@ -305,6 +306,18 @@ function SerenityController(context, log, env) {
...
+    // Org-wide serenity rollout gate. Serenity is "active" for an org only when
+    // its `LLMO/serenity` feature flag is ON *and* a Semrush workspace resolves
+    // for the brand (the workspace half is enforced below by
+    // resolveBrandWorkspace). While the flag is OFF the org's UI keeps reading
+    // the normal backend data — even if a `semrush_workspace_id` has already
+    // been backfilled for rollout prep — so reject the serenity surface with a
+    // 404 (the same "no serenity for this org" contract the UI already handles
+    // for an org without a workspace). Checked before brand resolution so an
+    // inactive org never leaks brand existence.
+    if (!await isSerenityActiveForOrg(ctx, spaceCatId, log)) {
```
MysticatBot, line 318: issue (blocking): The OpenAPI spec (`docs/openapi/`) was not updated to document the new 404 response ("Serenity is not active for this organization") that this gate introduces. The repo CLAUDE.md requires: "Specification Sync: Keep OpenAPI specs and implementation in sync." Clients consuming this API need to know these endpoints can now return 404 under a new condition (feature flag OFF). Similarly, the 403 responses on the brands controller create/update paths should be documented.

How to fix: Add a `404` response entry (and `403` for brands paths) to the relevant OpenAPI path specs in `docs/openapi/paths/`, then run `npm run docs:lint` to validate.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-designed feature flag gate with fail-safe defaults and comprehensive tests.
**Complexity: HIGH** - 578 changed lines across 10 files; API surface signal (new error responses on existing endpoints).
_Changes: Adds an org-wide `LLMO/serenity` feature flag gate to the serenity and brand controllers, defaulting OFF, decoupling rollout from Semrush workspace provisioning (10 files)._

Note: CI `it-postgres` checks are still pending - confirm they pass before merge.

## Must fix before merge

1. [Important] OpenAPI spec not updated for new error responses - `src/controllers/serenity.js:318`, `src/controllers/brands.js:1502,1779` (details inline)

<details>

- nit: Inconsistent HTTP status between controllers for the same conceptual gate (404 vs 403) with identical error message - consider differentiating the 403 message to help clients distinguish feature-gating from auth failure - `src/controllers/brands.js:1502`
- suggestion: The update handler's `getBrandById` read fires for every sync-field edit on non-rolled-out orgs (the common case during rollout); inverting the check order (read brand first, skip flag check if no workspace) would avoid the extra round-trip for flat brands - `src/controllers/brands.js:1776`
- suggestion: Add argument-correctness assertion in brands controller gate test (`expect(isSerenityActiveStub).to.have.been.calledWith(sinon.match.any, ORGANIZATION_ID, sinon.match.any)`) to catch wiring mistakes - `test/controllers/brands.test.js:4710`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 25s | <b>Cost:</b> $8.24 | <b>Commit:</b> <code>9081e410959b39fa172a2d819f6378835cbcf32a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the MysticatBot review in `3a540127`.

Must fix before merge:
- [Important] OpenAPI spec not updated for the new error responses — done. Verified first that no NEW status code is actually introduced: every gated brand-scoped serenity op already documents `404`, and brands-v2 create/update already document `403`. The gap was that the descriptions didn't name the new trigger. Fixed by:
  - brands-v2 `createBrandForOrgV2` / `updateBrandForOrgV2` `403` descriptions now spell out the serenity-flag-off case (they previously read only "User does not have access to this organization", which was genuinely misleading now that 403 has a second meaning).
  - every gated brand-scoped serenity op's `404` description now notes "…or serenity is not active for the organization". The two org-level catalogue ops (`listSerenityOrgModels`, `listSerenityOrgLanguages`) are intentionally NOT gated, so they were left unchanged.
  - `npm run docs:lint` is clean (the 207 warnings are all pre-existing and unrelated).

Non-blocking:
- suggestion: add an argument-correctness assertion to the brands gate test — done. Both the create-inactive and update-inactive gate tests now assert `isSerenityActiveForOrg` is called with the request's own org as the 2nd positional arg (`spaceCatId`), so a wiring slip can't silently let a Semrush-mode write through.
- suggestion: invert the update-handler check order (read brand first, skip flag check if no workspace) — declined. The current order checks the cached `isSerenityActiveForOrg` flag FIRST and reads the brand (`getBrandById`) only on the inactive path. On the post-rollout steady state (flag ON) the gate short-circuits and pays zero extra reads. Reading the brand first would add a brand read to that common active path permanently, to save one transient indexed read on flat-brand sync-field edits during the (temporary) rollout window — a net regression for the long-lived case. The flag check is a module-scoped cache hit, so it is the cheaper thing to do first.
- nit: 404 (serenity) vs 403 (brands) for the same conceptual gate with an identical message — declined. The status difference is intentional: the `/serenity/*` surface returns 404 from `authorize` so an inactive org never leaks brand existence, whereas a brand write is an explicit reject (403). The message ("Serenity is not active for this organization") is already distinct from the auth-failure messages on these routes ("User does not have access to this organization"), so a client can already tell feature-gating apart from an authz denial.

src/controllers/serenity.js
rainer-friederich: Fixed in `3a540127`. Enriched the brands-v2 create/update `403` descriptions and every gated brand-scoped serenity `404` description to name the serenity-flag-off case (the status codes themselves were already documented). `docs:lint` clean. Details in the consolidated comment.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - prior OpenAPI spec gap addressed; no new blocking issues on re-review.
**Complexity: HIGH** - 578 changed lines across 12 files; API surface signal (new error responses on existing endpoints).
_Changes: Addresses the prior review's OpenAPI spec finding by documenting the serenity-flag-off condition on all gated 404/403 responses, and adds argument-correctness assertions to the gate tests (12 files)._

Note: CI `build` checks are still pending - confirm they pass before merge.

<details>

- nit: The in-process flag cache keys on `spaceCatId` alone; if a second feature-flag cache is added later using the same Map shape, a composite key (orgId+product+flagName) would prevent collisions - low risk today since this is the only consumer - `src/support/serenity/serenity-active.js:97`

</details>

- OpenAPI spec now documents the serenity-flag-off condition on all gated 404 (serenity routes) and 403 (brands-v2 create/update) responses; `docs:lint` clean.
- Argument-correctness assertions added to both the create-inactive and update-inactive gate tests (assert `isSerenityActiveForOrg` is called with the request's own org ID).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 8m 30s | <b>Cost:</b> $5.39 | <b>Commit:</b> <code>3a540127c8548da19e718a0b21ca4f97a36ecd0d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the remaining non-blocking nit in `23d65914`.

- nit: flag cache keyed on `spaceCatId` alone — done. The in-process cache now keys on the full `(org, product, flag)` identity (`flagCacheKey`), so the Map is collision-proof if its shape is ever reused for a second feature flag. Behaviour-preserving (this module reads exactly one flag today). Coverage stays 100%; `type-check` clean.

Re-requesting a review so the approval lands on the latest commit.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - prior nit addressed cleanly; no new issues on re-review.
**Complexity: HIGH** - 966 changed lines across 12 files; API surface signal.
_Changes: Addresses the prior cache-key collision nit by extracting a composite `flagCacheKey` helper that keys on org+product+flag identity (12 files)._

Note: CI `build` checks are still pending - confirm they pass before merge.

- Cache key now uses composite `spaceCatId::LLMO::serenity` identity instead of bare org UUID, preventing future collisions if the cache shape is reused.
- OpenAPI spec documents the serenity-flag-off condition on all gated 404/403 responses (resolved in prior commit).
- Argument-correctness assertions added to both create-inactive and update-inactive gate tests (resolved in prior commit).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 19s | <b>Cost:</b> $2.09 | <b>Commit:</b> <code>23d659145aee190d6ed70368c70717f812b599a5</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2735, opened 2026-07-01
feat: return per-prompt userIntent in topic-prompts endpoint | LLMO-5950

jjenscodee:
## What & why

The brand-presence **topic-prompts** listing (`GET /org/:spaceCatId/brands/:brandId/brand-presence/topics/:topicId/prompts`) returned no per-prompt intent, so downstream the Data Insights export's **User Intent** column was always blank. Intent lives only on `prompts.intent`; the executions source carries just a `prompt_id` FK.

## Change

- **`src/support/prompts-storage.js`** — new exported `getIntentsByPromptIds({ promptIds, postgrestClient })`: a targeted `prompts.select('id, intent').in('id', ids)` lookup returning a `Map<promptId, intent>`. Reuses the existing `withMissingIntentFallback`, so **intent-column-absent environments degrade gracefully** instead of throwing. Any other query error → empty Map (non-fatal enrichment; never fails the request).
- **`src/controllers/llmo/llmo-brand-presence.js`** — `createTopicPromptsHandler` collects the executions' distinct `prompt_id`s, calls the helper with the handler's PostgREST client, and threads the map into `buildPromptDetails`, which now emits `userIntent`. Intent is 1:1 with `prompt_id`, so no row fan-out; `null` prompt_id → `''`.

### Why a handler-side lookup (not a view change or RPC)
The shared `brand_presence_executions_active` view has ~40+ consumers including perf-tuned, partition-pruned RPCs; adding a `prompts` join there risks regressing all of them. A dedicated RPC is over-engineered for a single 1:1 column on an endpoint that isn't RPC-backed today. The handler-side lookup has **zero blast radius and needs no migration**, at the cost of one bounded extra round-trip per topic expansion.

## Scope / follow-ups
- **Backend only.** The elmo-ui export cell for User Intent is still hardcoded empty — now a one-line read since the API returns `userIntent`.
- The `userIntent` **filter** query param remains parsed-but-unused (server-side filtering not wired).
- Blank `userIntent` is expected where `prompts.intent` is `null` (best-effort classification + manual backfill) — this surfaces intent, it doesn't populate it.

## Testing
- Added unit tests: `getIntentsByPromptIds` (empty/no-client, mapping+dedupe+null-skip, generic-error degrade, column-absent retry) and `buildPromptDetails`/handler enrichment.
- `npx mocha` on the affected suites: **670 passing, 0 failing**. eslint clean on all changed files.

## Related issue
https://jira.corp.adobe.com/browse/LLMO-5950

### DIFF
--- a/src/controllers/llmo/llmo-brand-presence.js
+++ b/src/controllers/llmo/llmo-brand-presence.js
@@ -2181,13 +2181,7 @@ export function createTopicPromptsHandler(getOrgAndValidateAccess) {
       const rawRows = data || [];
       const topicResponseLabel = topicLabelForDetailResponse(rawRows, topicName);
       const topicIdResponse = topicIdForDetailResponse(rawRows, topicName);
-      // Enrich with per-prompt intent from the prompts table (1:1 with prompt_id;
-      // executions carry only prompt_id). Non-fatal + intent-column-absent safe.
-      const intentByPromptId = await getIntentsByPromptIds({
-        promptIds: rawRows.map((r) => r.prompt_id),
-        postgrestClient: client,
-      });
-      let items = buildPromptDetails(rawRows, intentByPromptId);
+      let items = buildPromptDetails(rawRows);
 
       // When a search query is provided, filter to only prompts whose text
       // matches — mirroring the original brand presence client-side behaviour
@@ -2204,8 +2198,21 @@ export function createTopicPromptsHandler(getOrgAndValidateAccess) {
       const start = pagination.page * pagination.pageSize;
       const paged = items.slice(start, start + pagination.pageSize);
 
+      // Enrich only the returned page with per-prompt intent from the prompts
+      // table (1:1 with prompt_id; executions carry only prompt_id). Done after
+      // pagination so the .in() lookup is bounded to the page (≤ pageSize),
+      // never the full rawRows set. Non-fatal + intent-column-absent safe.
+      const intentByPromptId = await getIntentsByPromptIds({
+        promptIds: paged.map((item) => item.promptId),
+        postgrestClient: client,
+      });
+      const pagedItems = paged.map((item) => ({
+        ...item,
+        userIntent: intentByPromptId.get(item.promptId) || '',
+      }));
+
       return cachedOk({
-        items: paged,
+        items: pagedItems,
         totalCount,
         topic: topicResponseLabel,
         topicId: topicIdResponse,
--- a/src/controllers/llmo/llmo-brand-presence.js
+++ b/src/controllers/llmo/llmo-brand-presence.js
@@ -18,6 +18,7 @@ import {
 } from '@adobe/spacecat-shared-http-utils';
 import { hasText, isValidUUID } from '@adobe/spacecat-shared-utils';
 import { cachedOk } from '../../support/cached-response.js';
+import { getIntentsByPromptIds } from '../../support/prompts-storage.js';
 
 /**
  * Brand Presence filter-dimensions handler for org-based routes.
@@ -1939,7 +1940,7 @@ export function aggregateTopicData(rows) {
  * @returns {Array<Object>} PromptDetail-compatible objects
  * @internal Exported for testing
  */
-export function buildPromptDetails(rows) {
+export function buildPromptDetails(rows, intentByPromptId = new Map()) {
   const promptMap = new Map();
 
   rows.forEach((row) => {
@@ -1985,6 +1986,7 @@ export function buildPromptDetails(rows) {
       sentiment: r.sentiment || '',
       errorCode: r.error_code || '',
       origin: r.origin || '',
+      userIntent: (r.prompt_id != null && intentByPromptId.get(String(r.prompt_id))) || '',
     };
   });
 }
@@ -2179,7 +2181,13 @@ export function createTopicPromptsHandler(getOrgAndValidateAccess) {
       const rawRows = data || [];
       const topicResponseLabel = topicLabelForDetailResponse(rawRows, topicName);
       const topicIdResponse = topicIdForDetailResponse(rawRows, topicName);
-      let items = buildPromptDetails(rawRows);
+      // Enrich with per-prompt intent from the prompts table (1:1 with prompt_id;
+      // executions carry only prompt_id). Non-fatal + intent-column-absent safe.
+      const intentByPromptId = await getIntentsByPromptIds({
+        promptIds: rawRows.map((r) => r.prompt_id),
+        postgrestClient: client,
+      });
+      let items = buildPromptDetails(rawRows, intentByPromptId);
 
       // When a search query is provided, filter to only prompts whose text
       // matches — mirroring the original brand presence client-side behaviour
--- a/src/controllers/llmo/llmo-brand-presence.js
+++ b/src/controllers/llmo/llmo-brand-presence.js
@@ -1940,7 +1940,7 @@ export function aggregateTopicData(rows) {
  * @returns {Array<Object>} PromptDetail-compatible objects
  * @internal Exported for testing
  */
-export function buildPromptDetails(rows, intentByPromptId = new Map()) {
+export function buildPromptDetails(rows) {
   const promptMap = new Map();
 
   rows.forEach((row) => {
@@ -1986,7 +1986,6 @@ export function buildPromptDetails(rows, intentByPromptId = new Map()) {
       sentiment: r.sentiment || '',
       errorCode: r.error_code || '',
       origin: r.origin || '',
-      userIntent: (r.prompt_id != null && intentByPromptId.get(String(r.prompt_id))) || '',
     };
   });
 }
@@ -2204,7 +2203,9 @@ export function createTopicPromptsHandler(getOrgAndValidateAccess) {
       // never the full rawRows set. Non-fatal + intent-column-absent safe.
       const intentByPromptId = await getIntentsByPromptIds({
         promptIds: paged.map((item) => item.promptId),
+        organizationId,
         postgrestClient: client,
+        log: ctx.log,
       });
       const pagedItems = paged.map((item) => ({
         ...item,
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -104,6 +104,40 @@ async function withMissingIntentFallback(postgrestClient, run) {
 }
 
 /**
+ * Loads `intent` for a set of prompts by their `prompts.id` (uuid) values.
+ * Used to enrich reads (e.g. brand-presence executions) that carry a `prompt_id`
+ * FK but not intent itself. Degrades gracefully in intent-column-absent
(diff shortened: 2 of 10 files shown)


MysticatBot (changes requested):
Hey @jjenscodee,

**Verdict: Request changes** - intent enrichment should happen after pagination to avoid unbounded queries.
_Changes: Adds per-prompt `userIntent` field to the topic-prompts endpoint response by looking up intent from the prompts table via a handler-side query (4 files)._

## Must fix before merge

1. [Important] Intent lookup fetches data for all rawRows (up to 200K) before pagination trims to 20-1000 items - `src/controllers/llmo/llmo-brand-presence.js:2187` (details inline)
2. [Important] Missing integration test for modified endpoint per CLAUDE.md rule: "New or modified endpoints must include integration tests in `test/it/`" - no IT coverage exists for the topic-prompts handler today, and this PR modifies its response shape.

<details>

- suggestion: Update OpenAPI spec for the new `userIntent` response field per CLAUDE.md "Specification Sync" rule, or document why no spec exists for this endpoint yet - `docs/openapi/`
- nit: `getIntentsByPromptIds` silently swallows all errors with no log output, making it impossible to diagnose blank intents in production without a code change. A `log.debug` on the error path would preserve the non-fatal semantics while giving operators a signal - `src/support/prompts-storage.js:128`

</details>

src/controllers/llmo/llmo-brand-presence.js
```diff
@@ -2179,7 +2181,13 @@ export function createTopicPromptsHandler(getOrgAndValidateAccess) {
       const rawRows = data || [];
       const topicResponseLabel = topicLabelForDetailResponse(rawRows, topicName);
       const topicIdResponse = topicIdForDetailResponse(rawRows, topicName);
-      let items = buildPromptDetails(rawRows);
+      // Enrich with per-prompt intent from the prompts table (1:1 with prompt_id;
+      // executions carry only prompt_id). Non-fatal + intent-column-absent safe.
+      const intentByPromptId = await getIntentsByPromptIds({
+        promptIds: rawRows.map((r) => r.prompt_id),
```
MysticatBot: issue (blocking): The intent lookup operates on ALL `rawRows` prompt_ids (up to `WEEKS_QUERY_LIMIT` = 200K rows) before pagination. After deduplication the distinct ID set can still be very large, risking PostgREST URL-length limits (the `.in()` clause is serialized into a query parameter) and wasting a DB round-trip for data that will never be returned.

The paginated result is typically 20-1000 items. Move the enrichment to after pagination so only the final page's prompt_ids are looked up:

```js
let items = buildPromptDetails(rawRows);
// ... search filter, sort, paginate ...
const paged = items.slice(start, start + pageSize);
const pagedIds = paged.map(i => i.promptId).filter(Boolean);
const intentByPromptId = await getIntentsByPromptIds({
  promptIds: pagedIds,
  postgrestClient: client,
});
paged.forEach(item => {
  item.userIntent = intentByPromptId.get(item.promptId) || '';
});
```

This eliminates the unbounded IN clause and the wasted work in a single move.

jjenscodee:
Thanks @MysticatBot — addressing the review:

**Must fix**
1. ✅ **Unbounded intent lookup before pagination** (`llmo-brand-presence.js`) — moved `getIntentsByPromptIds` to *after* pagination (commit a05939a9). The `.in()` lookup now covers only the returned page's `promptId`s (≤ `pageSize`) instead of all `rawRows` (up to the 200K `WEEKS_QUERY_LIMIT`), eliminating the unbounded IN clause and the wasted round-trip. Implemented as a non-mutating `.map` (repo's `no-param-reassign` rule). Unit tests + lint green.
2. ⏸️ **Missing integration test** — deferred pending a maintainer decision. There is no existing IT for the topic-prompts handler, the IT PostgREST image may not yet carry the `prompts.intent` column (see the `withMissingIntentFallback` TODO in `prompts-storage.js`), and net-new IT scaffolding (seed IDs, seed data, shared factory, wiring) is a judgment call I don't want to guess on. Flagging to the author rather than auto-adding.

**Non-blocking** (noted, not changed here): OpenAPI spec for `userIntent` (this endpoint isn't schematized today), and a `log.debug` on the `getIntentsByPromptIds` error path.

MysticatBot (approved):
Hey @jjenscodee,

**Verdict: Approve** - prior blocking findings resolved; no new blocking issues.
**Complexity: HIGH** - medium diff; API surface signal (response shape change on endpoint with 40+ consumers).
_Changes: Adds per-prompt `userIntent` enrichment to the topic-prompts endpoint via a bounded post-pagination lookup against the prompts table (10 files)._

- Intent lookup moved after pagination - `.in()` clause now bounded to pageSize items (commit a05939a9)
- Integration test added for topic-prompts `userIntent` enrichment with seed data (commit 41431caf)

<details>

- suggestion: Add a `log.warn` on the error path in `getIntentsByPromptIds` before returning the empty Map - preserves non-fatal semantics while giving operators a signal when enrichment silently degrades - `src/support/prompts-storage.js:128`
- nit: `buildPromptDetails` accepts an `intentByPromptId` parameter (line 1940) that no production caller passes (the handler enriches post-pagination via a separate spread at line 2210). The dual path is not a bug but may confuse future maintainers into thinking they should pass the map at build time instead of post-pagination. Consider removing the parameter or consolidating enrichment to one site - `src/controllers/llmo/llmo-brand-presence.js:1940`
- nit: Inconsistent `String()` coercion on Map key lookup - `buildPromptDetails` wraps with `String(r.prompt_id)` but the handler spread uses `item.promptId` directly. Both work because UUIDs are already strings, but picking one convention avoids future confusion - `src/controllers/llmo/llmo-brand-presence.js:2210`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 16s | <b>Cost:</b> $6.50 | <b>Commit:</b> <code>41431caf8b0119018acf21890ae3432e10025ebf</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

calvarezg:
## Multi-persona review (Architect · DBA · Security · Tester/Junior)

Prior review (MysticatBot) fully addressed: intent lookup moved after pagination ✅, integration test added ✅. Verified the IT seed against the `mysticat-data-service` schema — all columns/types/FK-order valid.

### Major
- **Two divergent `userIntent` assignment sites will drift** (`llmo-brand-presence.js:1989` + `:2205-2211`). `buildPromptDetails` computes `userIntent` from a param that the production caller never passes (default empty Map → always `''`), then the handler overwrites it for the paged slice with a *different* key expression (`String(r.prompt_id)` vs `item.promptId`). Output is correct today, but it's a dead, fully-tested duplicate that the next change will trip over. Suggest: drop the `userIntent` line + `intentByPromptId` param from `buildPromptDetails` and keep enrichment solely in the handler.

### Minor
- **Errors swallowed silently** (`prompts-storage.js` error branch): timeouts / permission errors / a 414 collapse to the same empty Map as the benign missing-column case, with no log — blank intent becomes undiagnosable. Discriminate `isMissingIntentColumnError` (debug) from other errors (warn); thread `ctx.log`.
- **`.in('id', ids)` URL length**: `pageSize` maxes at 1000, so up to 1000 UUIDs (~37KB) go into the query string — may hit proxy header limits (→ 414 → silently blank intent for that page). Chunk ids or cap per-request count.
- **Seed `dedupe_hash` is dead**: the `BEFORE INSERT` trigger recomputes it (it's a trigger-populated `TEXT` column, *not* `GENERATED ALWAYS`, so the insert does NOT fail). Drop the field from the fixture.
- **Handler enrichment test is weak**: the mock returns identical rows for both queries with `id === prompt_id`, so the assertion can't tell which site set the value nor prove the page-bounding. Add a >pageSize / off-page case.

### Question
- `getIntentsByPromptIds` has no org/brand predicate. Safe as wired (ids come from tenant-scoped executions), but the exported helper has no self-contained authz — is PostgREST under tenant RLS, or should the predicate move into the query for defense-in-depth?

### Working well
Bounded post-pagination lookup, non-fatal enrichment contract, thorough unit + env-independent IT, and a zero-blast-radius design (no view/RPC/migration change).

**Recommendation: Changes recommended** (not blocking) — the one Major is maintainability, not a runtime/data bug.

jjenscodee:
Thanks @calvarezg — great review. All addressed in a8db1e64:

**Major — divergent `userIntent` sites.** ✅ Removed the `userIntent` assignment + `intentByPromptId` param from `buildPromptDetails`; enrichment now lives solely in the handler after pagination. No more dead/duplicate path.

**Minor**
- **Silent error swallowing** ✅ `getIntentsByPromptIds` now takes `log` and discriminates: `isMissingIntentColumnError` → `log.debug` (benign, older image), anything else (timeout / permission / 414) → `log.warn`. Still non-fatal.
- **`.in()` URL length** ✅ Chunked to 100 ids/query via parallel lookups, so a full 1000-id page can't blow the query-string limit.
- **Dead `dedupe_hash` in fixture** ✅ Dropped — and `organization_id` too (both are BEFORE INSERT trigger-derived: `trg_bpe_dedupe_hash`, `bpe_set_organization_id`).
- **Weak handler test** ✅ Added a `>pageSize` case asserting only the page's `prompt_id`s are looked up (proves post-pagination bounding), plus unit coverage for org scoping, chunking, and the debug/warn split.

**Question — org/brand predicate.** Added an optional `organizationId` predicate to the lookup for defense-in-depth. It was safe as wired (ids come from tenant-scoped executions), but the scoped query makes the helper self-contained rather than relying on the caller / RLS.

Unit tests green (670); the IT runs in CI.

calvarezg (approved):
## Re-review (multi-persona) — all findings resolved ✅

Verified `a8db1e64` fixes each item from the prior review against the diff (not just the summary):

- **Major — divergent `userIntent` sites:** ✅ `buildPromptDetails` is no longer touched; enrichment lives solely in the handler after pagination. Single source of truth.
- **Silent error swallowing:** ✅ helper takes `log` and splits `isMissingIntentColumnError`→`debug` vs other→`warn`. Still non-fatal.
- **`.in()` URL length:** ✅ chunked to 100 ids/query via `Promise.all` — a full 1000-id page can't blow the query string.
- **Dead `dedupe_hash` fixture value:** ✅ dropped (and `organization_id` too — both BEFORE INSERT trigger-derived).
- **Weak handler test:** ✅ added a >pageSize case asserting `.in('id', …)` receives only the page's prompt_ids, proving post-pagination bounding.
- **Org/brand predicate:** ✅ optional `organizationId` scope added + unit coverage; helper is now self-contained.

New-issue sweep: `organizationId` (= `spaceCatId`) is in scope and matches the executions query's own org filter, so the lookup scope is consistent; `ctx.log`/`client` valid; `Promise.all` checks each chunk's `{data,error}` independently and preserves the non-fatal contract. No regressions.

Clean fixes, meaningful tests. LGTM 🚀

PR #2739, opened 2026-07-02
feat(serenity): maintain mapping rows + cut over to semrushSubWorkspaceId

rainer-friederich:
## 1. Abstract

Two changes to serenity: (a) wires `brand_to_semrush_projects` mapping-row maintenance into the sub-workspace market lifecycle (create, delete, activate, deactivate, brand-create provisioning); (b) cuts over ~12 internal call sites from the deprecated `getSemrushWorkspaceId`/`setSemrushWorkspaceId` Brand accessors to `getSemrushSubWorkspaceId`/`setSemrushSubWorkspaceId`.

## 2. Reasoning

Sub-workspace mode never wrote to `brand_to_semrush_projects` — only the legacy flat-mode handlers did. Every brand migrated to a sub-workspace since the first one (Lovesac) has gone dark for the table's reverse-map consumers (the Brand Presence / URL Inspector Elements API wrappers), because the mapping table is being kept and actively maintained rather than retired. Depends on the `site_id`/`deleted_at` schema (mysticat-data-service PR 765) and the model attributes/accessor (spacecat-shared PR 1763).

Separately, in the same review thread: `brands.semrush_workspace_id` actually holds the brand's **sub-workspace** id, and the name confuses every new reader. The companion PRs (765/1763) flip which column/attribute is write-of-record; this PR is the actual cutover — every internal serenity call site that used to read/write the old accessor now uses the new one, so this service stops writing through the deprecated path.

**Mapping-row maintenance:**
- New `src/support/serenity/mapping-rows.js`: four best-effort helpers (`upsertMappingRow`, `tombstoneMappingRow`, `tombstoneAllForBrand`, `linkSiteToLiveRows`) that never fail the caller and log a greppable token on a real failure. `upsertMappingRow` uses a native PostgREST upsert keyed on the Semrush project id, which both makes the common path atomic and revives a tombstoned row on re-adopt for free (writes `deletedAt: null`, never touches `siteId`, so an existing site link survives a revive). A duplicate-slice race (the same accepted race the handlers already tolerate via oldest-wins reads) is classified separately from a genuine write failure so it doesn't trigger the write-failure alarm — note: the classifier currently only recognizes this one race; see the open follow-up in the implementation plan (§11.9) for the still-unhandled `uq_brand_to_semrush_project` concurrent-upsert case.
- `handleCreateMarketSubworkspace` / `handleDeleteMarketSubworkspace` gain an optional `dataAccess` — when supplied, they upsert or tombstone the mapping row after the upstream call succeeds; omitted, behavior is unchanged.
- Wired into the create/delete/activate/deactivate controller paths: market create and each market in an activation batch upsert their row; market delete and a full deactivate tombstone theirs; the market-mirror Site id is linked onto live, unlinked rows once `ensureMarketSite` resolves it (never overwriting an existing link).
- The brand-CREATE provisioning path (`provisionBrandSubworkspace`) runs before the brand row is persisted, so it cannot write the mapping row itself — the row's foreign key requires an existing brand. It now surfaces the initial market's identity (project id, geo, language) on its return value, and the brand controller writes the mapping row once the brand row actually exists, mirroring how it already handles the site mirror.
- One documented gap: when a delete resolves no project at all in the child-workspace listing, there's no project id or brand id in scope to tombstone with, so a live mapping row for an already-vanished project is left un-tombstoned — accepted, reconcile-recoverable drift rather than silently claimed as covered.

**Workspace-id accessor cutover:**
- `workspace-resolver.js`, `workspace-lifecycle.js`, `controllers/serenity.js`, `brand-provisioning.js`, `brands-storage.js`, `controllers/brands.js`, `controllers/sites.js` — every call site that read or wrote the brand's Semrush workspace id via the deprecated `getSemrushWorkspaceId`/`setSemrushWorkspaceId` now uses `getSemrushSubWorkspaceId`/`setSemrushSubWorkspaceId`. Org-level workspace id calls (`organizations.getSemrushWorkspaceId`) are untouched — only the Brand-level accessor renamed.
- `docs/openapi/schemas.yaml`: `semrushWorkspaceId` marked `deprecated: true`; `semrushSubWorkspaceId` documented as the canonical field. Both remain in the response DTO — this is a read-side deprecation notice, not a removal.

## 4. Required information

- Spec: https://github.com/adobe/serenity-docs/blob/docs/brand-semrush-mapping-maintenance/docs/discovery/brand-semrush-mapping-maintenance.md (§4 write-path plan, §10 the companion rename decision)
- Implementation plan: https://github.com/adobe/serenity-docs/blob/docs/brand-semrush-mapping-maintenance/docs/discovery/brand-semrush-mapping-maintenance-implementation.md
- Other: https://github.com/adobe/mysticat-data-service/pull/765 (schema), https://github.com/adobe/spacecat-shared/pull/1763 (model), https://github.com/adobe/serenity-docs/pull/23

- `spacecat-shared` — contract: this PR calls `dataAccess.BrandSemrushProject.create({..., deletedAt: null}, {upsert, onConflict: 'semrushProjectId'})`, `.findBySemrushProjectId`, `.allByBrandId`, and the `siteId`/`deletedAt`/`semrushSubWorkspaceId` getters/setters added in PR 1763 — this repo's `package.json` dependency on `@adobe/spacecat-shared-data-access` needs bumping to that release before deploy.
- `mysticat-data-service` — contract: assumes the `site_id`/`deleted_at`/`semrush_sub_workspace_id` columns, the live-slice partial unique index, and the reversed sync trigger from PR 765 exist in the target environment's schema.

## 7. Test plan

Unit tests were added/extended for every new code path: a dedicated suite for `mapping-rows.js` (all four helpers, including the duplicate-slice-vs-real-failure classification and the defensive non-array fallback), the handler-level `dataAccess`-present/absent/failing branches in `markets-subworkspace.test.js`, the new call sites in `serenity.test.js` (create/activate site-linking, deactivate bulk-tombstone) and `brands.test.js` (mapping-row write after brand persistence, ordered after the site mirror), and the widened `provisionBrandSubworkspace` return shape in `brand-provisioning.test.js`. The accessor-cutover tests (`workspace-resolver.test.js`, `workspace-lifecycle.test.js`, `markets-subworkspace.test.js`, `serenity.test.js`, `openapi-contract/serenity-api.test.js`, `brand-provisioning.test.js`, `brands-storage.test.js`, `brands.test.js`, `sites.test.js`) were updated to assert against the new accessor names, with log-field-name assertions (e.g. `meta.semrushWorkspaceId`) deliberately left unchanged where they describe log output rather than data-access calls. `npm run type-check`, full `npx eslint`, and `npm run docs:lint` all pass clean (0 new warnings). Coverage was checked scoped to the touched files against this package's local thresholds and is at or near 100% on every changed line; CI's codecov/patch gate is authoritative.

This PR was not run against a live PostgREST/data-service instance this session — it depends on a `@adobe/spacecat-shared-data-access` release containing PR 1763, which has not shipped yet. **Known, structural CI gap (not caused by this PR):** `ci / it-postgres` is currently red — `test/it/postgres/docker-compose.yml` pins a `mysticat-data-service` image (`v5.44.0`) that predates this entire rollout, so the pinned test DB has neither the mapping-table columns nor `semrush_sub_workspace_id`. This resolves once PR 765 merges/releases and the pin is bumped — not actionable from this PR alone. Every other CI job (lint, type-check, unit tests, build, docs) is green. Before merge, bump the `@adobe/spacecat-shared-data-access` dependency and the IT image pin, then run the existing `test/it/postgres/serenity.test.js` integration suite to confirm end-to-end: create → row live; delete → tombstoned; re-create the same slice → new live row alongside the tombstone; deactivate → all tombstoned; `site_id` populated after create/activate; brand reads/writes route through `semrushSubWorkspaceId`.

- Depends on: https://github.com/adobe/spacecat-shared/pull/1763 (this PR's `dataAccess.BrandSemrushProject`/`Brand` calls require the attributes and the native-upsert usage it adds), which itself depends on https://github.com/adobe/mysticat-data-service/pull/765 (schema).
- Related: https://github.com/adobe/serenity-docs/pull/23 (spec + implementation plan this rollout follows).

Sequence: mysticat-data-service#765 merges and releases → spacecat-shared#1763's IT pin bumps to that release, merges and releases → this PR bumps its `@adobe/spacecat-shared-data-access` dependency, its own IT pin bumps to the same data-service release, then merges and deploys.


src/support/serenity/mapping-rows.js
```diff
@@ -0,0 +1,222 @@
...
+ * workspace id; this also sweeps rows whose upstream project had already
+ * vanished before decommission ran.
+ *
+ * @param {any} dataAccess
+ * @param {string|null|undefined} brandId
+ * @param {any} [log]
+ */
+export async function tombstoneAllForBrand(dataAccess, brandId, log) {
+  const BrandSemrushProject = dataAccess?.BrandSemrushProject;
+  if (!BrandSemrushProject || !brandId || !hasText(brandId)) {
```
MysticatBot, line 176: issue (blocking): `Promise.all` in `tombstoneAllForBrand` (and the same pattern in `linkSiteToLiveRows` at line ~207) short-circuits on first rejection. If row 1 of 10 fails, rows 2-10 are abandoned mid-flight. The outer catch logs a single "bulk tombstone failed" error with no per-row visibility.

**Why it matters:** A transient failure on one row leaves an unnecessarily large drift surface. The reconcile job recovers eventually, but the blast radius of a single-row failure is N-1 unprocessed rows rather than just 1.

**Fix:** Replace `Promise.all` with `Promise.allSettled` and log individually-failed rows:
```js
const results = await Promise.allSettled(live.map(async (row) => {
  row.setDeletedAt(now);
  await row.save();
}));
const failed = results.filter((r) => r.status === 'rejected');
if (failed.length) {
  log?.error?.(`serenity mapping row: ${WRITE_FAILED_TOKEN} - partial bulk tombstone`, {
    brandId, op: 'tombstone-all', failed: failed.length, total: live.length,
  });
}
```
Apply the same pattern to `linkSiteToLiveRows`.

src/support/serenity/brand-provisioning.js
```diff
@@ -220,9 +230,18 @@ export async function provisionBrandSubworkspace(context, {
...
+  // handleCreateMarketSubworkspace's own body already carries the initial
+  // market's identity (geoTargetId/languageCode resolved from the same
+  // resolvedMarket/resolvedLanguageCode this call passed in) — read it back
+  // rather than re-deriving it.
+  /** @type {any} */
+  const resultBody = result.body || {};
   return {
-    semrushWorkspaceId: capturedWorkspaceId,
-    published: Boolean(/** @type {{ published?: boolean }} */ (result.body || {}).published),
+    semrushSubWorkspaceId: capturedWorkspaceId,
```
MysticatBot, line 240: issue (blocking): `geoTargetId: Number(resultBody.geoTargetId) || 0` silently normalizes a missing/undefined `geoTargetId` to `0`. Downstream, `upsertMappingRow` validates `brandId`, `semrushProjectId`, and `languageCode` but does NOT guard against `geoTargetId <= 0`. If the handler body ever carries a `projectId` without a `geoTargetId`, a row with `geo_target_id = 0` is persisted - semantically invalid data that only reconcile can fix.

**Why it matters:** The caller in `brands.js` guards on `hasText(provisionedInitialMarket.projectId)` which catches the no-projectId case, but not has-projectId-without-geoTargetId.

**Fix (either):**
1. Add `geoTargetId` to the guard in `upsertMappingRow`:
```js
if (!brandId || !hasText(brandId) || !semrushProjectId || !hasText(semrushProjectId)
    || !languageCode || !hasText(languageCode) || !geoTargetId || geoTargetId <= 0) {
```
2. Or guard at the `brands.js` call site alongside the existing `hasText(projectId)` check.

MysticatBot (changes requested):
Hey @rainer-friederich,

**Verdict: Request changes** - two hardening items on the bulk helpers need addressing before merge.
**Complexity: HIGH** - large diff; API surface change (OpenAPI schema + accessor cutover).
_Changes: Wires brand_to_semrush_projects mapping-row maintenance into the sub-workspace market lifecycle and cuts over ~12 call sites from the deprecated getSemrushWorkspaceId to getSemrushSubWorkspaceId (20 files)._
Note: CI checks are currently failing (it-postgres - known structural gap, pinned DB image predates the schema).
Note: Recommend a human read before merge - this change modifies a shared contract (docs/openapi/schemas.yaml). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Important] `Promise.all` in bulk helpers short-circuits on first rejection, leaving remaining rows unprocessed - `src/support/serenity/mapping-rows.js:172` (details inline)
2. [Important] `geoTargetId: Number(...) || 0` can persist semantically invalid 0 to the mapping table - `src/support/serenity/brand-provisioning.js:240` (details inline)

<details>

- nit: Test description still says `semrush_workspace_id` after the rename - `test/controllers/serenity.test.js:2533`
- suggestion: OpenAPI deprecation notice for `semrushWorkspaceId` should include a target removal version or timeline - `docs/openapi/schemas.yaml:13`
- suggestion: `handleDeleteMarketSubworkspace` now takes 6 positional args; consider migrating to an options bag matching the create handler pattern in a follow-up
- suggestion: `provisionBrandSubworkspace` return shape grew to 5 fields; consider a named typedef so producer/consumer stay in sync at lint time
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 42s | <b>Cost:</b> $6.50 | <b>Commit:</b> <code>42e669ba9c0af767b878c4bca62d070edc7db699</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/serenity/mapping-rows.js
rainer-friederich: Fixed in `83b190b1` — both `tombstoneAllForBrand` and `linkSiteToLiveRows` now use `Promise.allSettled`, logging the per-row failure count (failed/total) instead of letting one rejection abandon the batch. Covered by two new tests asserting every row is attempted even when one save rejects.

src/support/serenity/brand-provisioning.js
rainer-friederich: Fixed in `83b190b1` — added `geoTargetId <= 0` (and falsy) to the incomplete-slice guard in `upsertMappingRow`, so this normalized-to-0 fallback in `brand-provisioning.js` can no longer persist an invalid mapping row through any caller. Chose the guard-in-`upsertMappingRow` option (over guarding at the `brands.js` call site) since it protects every caller uniformly. Covered by a new test.

rainer-friederich:
Addressed the review in `83b190b1`.

MysticatBot (Request changes, 2 blocking + 4 non-blocking):
- [Important] `Promise.all` short-circuits on first rejection in the bulk helpers — done. Both `tombstoneAllForBrand` and `linkSiteToLiveRows` now use `Promise.allSettled`, processing every row and logging only the failed/total count on partial failure. Covered by two new tests.
- [Important] `geoTargetId: Number(...) || 0` can persist a semantically invalid 0 — done. Added `geoTargetId <= 0` to `upsertMappingRow`'s incomplete-slice guard, which protects every caller uniformly (chosen over guarding only at the `brands.js` call site). Covered by a new test; confirmed 0 is never a legitimate Google Ads Geo Target ID (same convention as `handlers/markets.js`'s existing "must be a positive integer" validation).
- nit: stale test description still saying `semrush_workspace_id` after the rename — fixed.
- suggestion: OpenAPI deprecation notice should include a target removal version/timeline — declined for now. No drop date has been committed for `semrush_workspace_id` (see the serenity-docs spec's §10 implementation-status note, updated this session) — the column is kept indefinitely for backward compatibility with no owner/timeline assigned yet. Adding a fabricated timeline to the OpenAPI doc would misstate that decision; will add once a real target exists.
- suggestion: `handleDeleteMarketSubworkspace`'s 6 positional args → options bag — declined for this PR. The review itself flagged this as a follow-up refactor, not a defect in the current change; out of scope here.
- suggestion: `provisionBrandSubworkspace`'s 5-field return shape → named typedef — declined for this PR, same reasoning as above (review-flagged follow-up, not a defect).

Pre-existing, unrelated to this fix: `ci / it-postgres` remains red on the structural IT-image-pin gap (pinned `mysticat-data-service:v5.44.0` predates this rollout's schema) — resolves once mysticat-data-service#765 merges/releases and the pin is bumped.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - both prior blocking findings addressed with solid implementations and test coverage.
**Complexity: HIGH** - large diff; API surface change (OpenAPI schema + accessor cutover).
_Changes: Wires brand_to_semrush_projects mapping-row maintenance into the sub-workspace market lifecycle and cuts over ~12 call sites from the deprecated getSemrushWorkspaceId to getSemrushSubWorkspaceId (20 files)._
Note: CI checks are currently failing (it-postgres - known structural gap, pinned DB image predates the schema).
Note: Recommend a human read before merge - this change modifies a shared contract (docs/openapi/schemas.yaml). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: `Number(resultBody.geoTargetId) || 0` normalizes absent to 0 rather than null, relying on the downstream guard to reject it - consider returning null to make the "absent" semantic explicit in the return shape - `src/support/serenity/brand-provisioning.js:236`
- suggestion: Narrow `dataAccess` threading to `{ BrandSemrushProject: ctx.dataAccess.BrandSemrushProject }` at handler call sites for defense-in-depth (prevents future code in the handler from accidentally reaching unrelated tables) - `src/controllers/serenity.js:603`
- suggestion: Use `calledOnceWithExactly` instead of `calledOnceWith` in test assertions for `linkSiteToLiveRows`/`tombstoneAllForBrand` to verify the `log` argument is passed through and catch accidental arg reordering - `test/controllers/brands.test.js:1104`
</details>

- Promise.all short-circuit in bulk helpers: switched to Promise.allSettled with partial-failure logging and per-row test coverage
- geoTargetId: Number(...) || 0 persisting invalid data: guarded by geoTargetId <= 0 check in upsertMappingRow, covered by new test
- Stale test description referencing old column name: fixed

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 59s | <b>Cost:</b> $6.51 | <b>Commit:</b> <code>83b190b1347bc2c41346af1017d703a0ab5a9d43</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the second MysticatBot review's non-blocking items in `a39bde99` (review stayed Approve; these are the "Non-blocking (3)" items, addressed per the standing rule that non-blocking items are never silently skipped even on an Approve):

- nit: `Number(resultBody.geoTargetId) || 0` normalizes absent to 0 — done. `provisionBrandSubworkspace` now returns `geoTargetId: null` when the upstream body carries none, so `upsertMappingRow`'s `!geoTargetId` guard rejects the unresolvable slice explicitly instead of persisting a sentinel `0`. Updated the JSDoc return type and the three now-`null` test fixtures in `brand-provisioning.test.js`.
- suggestion: narrow `dataAccess` threading to `{ BrandSemrushProject }` — done, at all three call sites in `serenity.js` (single-market create, delete, and the bulk-activate loop), not just the one cited line. Updated the two `serenity.test.js` assertions that deep-equal the full options object.
- suggestion: use `calledOnceWithExactly` instead of `calledOnceWith` for the `linkSiteToLiveRows` assertion — done in `brands.test.js`, now also verifying the `log` arg is passed through.

`npm run type-check` needed two follow-up JSDoc fixes in `markets-subworkspace.js` (`handleDeleteMarketSubworkspace`'s options param was missing `@param` tags for its earlier positional params, which was silently letting TS narrow `dataAccess`'s type from its `= null` default instead of the declared `any`) — no behavior change, just closes a type-checking gap the narrowing surfaced.

rainer-friederich:
`ci / it-postgres` is failing, but it is a known, already-tracked sequencing gap - not a bug in this PR.

Root cause: `POST /serenity/markets` resolves the brand's Semrush sub-workspace via `getSemrushSubWorkspaceId()` (the new write-of-record field, added in adobe/spacecat-shared#1763). The IT suite's PostgREST backend is pinned to `mysticat-data-service:v5.44.0` (`test/it/postgres/docker-compose.yml`), published 2026-06-19 - which predates the `semrush_sub_workspace_id` column and its sync trigger, both added by adobe/mysticat-data-service#765 (released 2026-07-02, not yet merged). Against that stale schema the sub-workspace lookup can't resolve, so `POST /serenity/markets` 404s, and every test in the suite that depends on a market existing (tags, prompts, activate/deactivate, the round-trip suite) cascades from that one root failure.

Confirmed by bisecting: `ci / it-postgres` was green through commit `d68b322c`, and started failing at `42e669ba` (`feat(brands): cut over internal serenity call sites to semrushSubWorkspaceId`) - exactly the commit that introduces the dependency on the new column.

This was already called out in MysticatBot's review ("CI checks are currently failing - it-postgres - known structural gap, pinned DB image predates the schema") and in the companion spacecat-shared#1763 PR's Deployment & merge order section.

Unblock sequence (matches spacecat-shared#1763's stated order):
1. adobe/mysticat-data-service#765 merges and releases (already `Approve`, zero outstanding findings).
2. Bump the pin in `test/it/postgres/docker-compose.yml` to the new release tag.
3. Re-run `ci / it-postgres` to confirm green against the real schema.

No code change needed in this PR for the failure itself - this is purely a release-sequencing dependency.

vivesing (approved):

PR #2737, opened 2026-07-01
feat(serenity): consume nested AIO tags (parentId create, tree read, PATCH re-parent)

rainer-friederich:
## 1. Abstract
Makes spacecat-api-service actually consume the 1-level nested AIO tag support shipped in `@adobe/spacecat-shared-project-engine-client` v1.6.0: nested-category create (`parentId`), a tree-aware read, and an id-keyed rename/re-parent endpoint on the Serenity (Semrush AIO) proxy.

## 2. Reasoning
The project-engine-client already modelled tag nesting (`parent_id`, `path[]`, `children_count`) and v1.6.0 taught its Counterfact mock to honour it (parent-aware create, tree-aware list, a `PATCH /aio/tags/{tag_id}` re-parent), released in https://github.com/adobe/spacecat-shared/pull/1758 . But spacecat-api-service pinned an older client and its Serenity transport was flat — `createProjectTags` dropped any hierarchy, `listProjectTags` hard-coded `parent_id: ''` (roots only), and there was no re-parent path at all. So a consumer writing `parent_id` would silently pass against the mock while losing the tree. This closes that gap — the api-service slice of the cross-repo nested-category effort ( https://github.com/adobe/spacecat-api-service/issues/2733 , parent design https://github.com/adobe/serenity-docs/issues/21 ), building on the create-tag endpoint from https://github.com/adobe/spacecat-api-service/pull/2730 .

## 3. High-level overview of the changes
Behaviour delta on the Serenity `/serenity/tags` surface (the Categories surface), additive and backward-compatible for existing flat callers:

- **Create can nest.** `POST /serenity/tags` now accepts an optional `parentId` (an upstream tag id from a prior list). When present, the new tag is created as a 1-level child of that category in a single upstream call; the response echoes the created tag's `id` and `parentId`. Omitting `parentId` is the previous flat/root behaviour, unchanged.
- **Read can drill the tree.** `GET /serenity/tags` gains an optional `parentId` query param: absent → the previous flat, prompt-derived list; present (even empty) → the standalone AIO tag tree at that level — `parentId=''` lists the root categories (each with `childrenCount`), a tag id lists that category's children (each with a `path[]` breadcrumb). The tree read uses the draft view so a just-created, still-unpublished category is visible.
[...]

PR #2741, opened 2026-07-02
feat(serenity): id-based prompt writes, closed-dimension tag resolve, promote-to-root

rainer-friederich:
## 1. Abstract
Adds the id-based prompt-write path, closed-dimension tag resolution, and promote-to-root support to the Serenity nested-category-tags surface, building on the bare-name-children fix already merged in PR #2737.

## 2. Reasoning
The nested-category alignment work (serenity-docs#24, extending issue #21) needs a way to tag prompts by upstream tag **id** rather than by name: a topic name is only unambiguous within its own category, so the name-based prompt-tagging endpoint (`aio/prompts/tagged`) can't reference a nested child at all — it's what forced the migration CLI into a flat-tag-then-reparent workaround. This PR adds the id-based upstream write path api-service needs to let a UI or migration tool tag a prompt with a specific child id directly. Landing alongside it: a way to resolve the ids of the fixed-vocabulary `source`/`intent`/`type` tags (needed to build a full `tagIds` set for a prompt), and closing gate 1 from the same live-verification session (serenity-docs#24 §3.1) — an explicit promote-to-root request was accepted by the parser but silently dropped before reaching the transport layer.

- **Id-based prompt writes**: `SerenityPromptInput`/`SerenityUpdatePromptRequest` gain an optional `tagIds` field (upstream tag ids), mutually exclusive with the existing `tags` (names) field. New transport methods `createPromptsByIds` (`POST aio/prompts`) and `updatePromptTags` (`PUT aio/prompts/tags`) call the corresponding upstream endpoints. Both prompt-write twin handler pairs (flat + subworkspace, create + update) route through whichever field the caller supplied.
- **Closed-dimension tag resolution**: `POST /serenity/tags` now also accepts `type: source|intent|type` (previously only `category`/`topic`). A closed-dimension create is resolve-before-create — if the project already has that exact tag, its id comes back unchanged (`200`, `created: false`) instead of attempting another create, since it's a small fixed set many independent callers need the id of. Open-dimension behavior (a duplicate name is a hard upstream error the caller must avoid) is unchanged.
- **Promote-to-root** (`PATCH /serenity/tags/:tagId`): an explicit JSON `null` `parentId` now reaches the upstream call as a literal `null`, distinct from an omitted `parentId` (which preserves the tag's current parent). Previously both were collapsed to the same "omit `parent_id`" behavior, so a client request to remove a child's parent silently did nothing.
- Un-skips two integration assertions the original PR (#2737) could only validate against live Semrush at the time: drilling a parent's children by id, and a full rename-by-id round trip against the mock — both are now exercisable because the mock's tag ids became URL-safe in a since-merged spacecat-shared fix.

## 4. Required information
- Spec: https://github.com/adobe/serenity-docs/pull/24
- Other: https://github.com/adobe/spacecat-api-service/pull/2737 · https://github.com/adobe/spacecat-shared/pull/1764

- **spacecat-shared** — consumed: the `@adobe/spacecat-shared-project-engine-client` mock (bumped to 1.7.0) supplies the id-based prompt-write routes this PR's IT suite exercises.
- **project-elmo-ui** — contract: the nested-category UI work (issue 2245) is expected to call these new endpoints once it reaches the prompt-authoring slice.

## 7. Test plan
(a) Ran the full serenity-scoped unit suite, the OpenAPI contract suite, and the route-capability coverage test locally; also ran `tsc --noEmit` and the OpenAPI spec linter. Could not run the integration suite locally (it needs the docker-compose Postgres + Semrush mock infrastructure) — the two new/updated IT test blocks (id-based prompt create + read-back, closed-dimension resolve/reject, the un-skipped id-round-trip tag tests) are new territory for this repo's local tooling and will be verified via CI's `ci/it-postgres` job.
(b) On dev/stage: exercise `POST /serenity/tags` with `type: source` twice for the same brand/market to confirm the second call returns the same id with `created: false`; exercise `PATCH /serenity/tags/:tagId` with `parentId: null` on a real nested child and confirm it lists as a root afterward.

- Related: https://github.com/adobe/spacecat-api-service/pull/2737 — already merged; this PR is a follow-up on top of it, per the original sequencing (serenity-docs#24's implementation plan).
- Related: https://github.com/adobe/spacecat-shared/pull/1764 — already merged (published as `@adobe/spacecat-shared-project-engine-client@1.7.0`, which this PR's `package.json` now pins).
- No hard merge-order dependency: this PR's own code doesn't require anything further from spacecat-shared to merge first (1.7.0 is already released).

### DIFF
--- a/docs/openapi/schemas.yaml
+++ b/docs/openapi/schemas.yaml
@@ -10880,6 +10880,24 @@ SerenityPrompt:
         Tag names are assumed unique per Semrush project; if duplicates exist,
         only the last entry is retained.
       example: { "category:Firefly": "uuid-1", "intent:research": "uuid-2" }
+    tags:
+      type: array
+      items: { type: string }
+      description: |
+        Echo of the tag NAMES this prompt was created/updated with via the
+        legacy name-based write path (`SerenityPromptInput.tags` /
+        `SerenityUpdatePromptRequest.tags`). Only present on a create/update
+        response (`SerenityCreatePromptsResponse.created`, the
+        `updateSerenityPrompt` 200); absent on `GET /serenity/prompts` list
+        items, which carry `tagMap` instead.
+    tagIds:
+      type: array
+      items: { type: string }
+      description: |
+        Echo of the upstream tag ids this prompt was created/updated with via
+        the id-based write path (serenity-docs#24). Only present when the
+        request used `tagIds` rather than `tags` — mutually exclusive with a
+        non-empty `tags` on the same response.
 
 SerenityPromptListResponse:
   type: object
--- a/docs/openapi/schemas.yaml
+++ b/docs/openapi/schemas.yaml
@@ -10894,7 +10894,12 @@ SerenityPromptListResponse:
 
 SerenityPromptInput:
   type: object
-  description: One row in a bulk-create prompts request.
+  description: |
+    One row in a bulk-create prompts request. `tags` (names, via the legacy
+    name-based upstream write path) and `tagIds` (upstream tag ids from a
+    prior tags list/create, via the id-based write path — serenity-docs#24)
+    are mutually exclusive: supply at most one. `tagIds`, when present, must
+    be non-empty.
   required: [text, geoTargetId, languageCode]
   properties:
     text: { type: string }
@@ -10902,6 +10907,12 @@ SerenityPromptInput:
       type: array
       items: { type: string }
       default: []
+      description: Tag NAMES. Mutually exclusive with tagIds.
+    tagIds:
+      type: array
+      items: { type: string }
+      minItems: 1
+      description: Upstream tag ids. Mutually exclusive with tags.
     geoTargetId:
       type: integer
       minimum: 1
@@ -10951,11 +10962,13 @@ SerenityUpdatePromptRequest:
   description: |
     Replace the prompt's text and tags. Carries the owning slice in
     `geoTargetId` + `languageCode` so the server can resolve the upstream
-    project without scanning. All four fields are required — the upstream
-    provider's API has no in-place update or GET-by-id, so the server runs
-    DELETE-old + POST-new and treats the body as the full next state.
-    Pass `tags: []` to clear tags.
-  required: [geoTargetId, languageCode, text, tags]
+    project without scanning. `geoTargetId`, `languageCode`, and `text` are
+    always required — the upstream provider's API has no in-place update or
+    GET-by-id, so the server runs DELETE-old + POST-new and treats the body
+    as the full next state. Exactly one of `tags` (pass `[]` to clear) or
+    `tagIds` (non-empty, id-based write path — serenity-docs#24) must also be
+    present; they are mutually exclusive.
+  required: [geoTargetId, languageCode, text]
   properties:
     geoTargetId:
       type: integer
@@ -10967,6 +10980,12 @@ SerenityUpdatePromptRequest:
     tags:
       type: array
       items: { type: string }
+      description: Tag NAMES. Mutually exclusive with tagIds.
+    tagIds:
+      type: array
+      items: { type: string }
+      minItems: 1
+      description: Upstream tag ids. Mutually exclusive with tags.
 
 SerenityBulkDeletePromptsRequest:
   type: object
--- a/src/support/serenity/handlers/prompts.js
+++ b/src/support/serenity/handlers/prompts.js
@@ -184,6 +184,16 @@ export async function publishAffected(transport, semrushWorkspaceId, projectIds,
   return errors;
 }
 
+/**
+ * Normalizes one bulk-create/update prompt input. `tags` (names, the legacy
+ * write path via `aio/prompts/tagged`) and `tagIds` (upstream tag ids, the
+ * id-based write path via `aio/prompts`) are MUTUALLY EXCLUSIVE — at most one
+ * may be present, and `tagIds` when present must be non-empty. `tagIds` stays
+ * `undefined` when absent so the handler can branch on
+ * `input.tagIds !== undefined` to pick the write path; `tags` always
+ * defaults to `[]` (unused, not `undefined`) to keep the name-based path's
+ * existing contract unchanged.
+ */
 export function normalizePromptInput(input) {
   const text = String(input?.text || '').trim();
   const languageCode = normalizeLanguageCode(input?.languageCode);
@@ -191,11 +201,106 @@ export function normalizePromptInput(input) {
   const tags = Array.isArray(input?.tags)
     ? input.tags.map((t) => String(t || '').trim()).filter(Boolean)
     : [];
+  const tagIds = Array.isArray(input?.tagIds)
+    ? input.tagIds.map((t) => String(t || '').trim()).filter(Boolean)
+    : undefined;
   if (!text || languageCode === null || geoTargetId === null) {
     return null;
   }
+  if (tagIds !== undefined && (tags.length > 0 || tagIds.length === 0)) {
+    return null;
+  }
   return {
-    text, languageCode, geoTargetId, tags,
+    text, languageCode, geoTargetId, tags, tagIds,
+  };
+}
+
+/**
+ * Creates ONE prompt via whichever upstream write path `input` requests --
+ * id-based (`POST aio/prompts`, when `input.tagIds` is set -- serenity-docs#24)
+ * or the legacy name-based path (`POST aio/prompts/tagged`, when `input.tags`
+ * is used). Both {@link handleCreatePrompts}'s per-item fan-out and
+ * {@link handleUpdatePrompt}'s post-delete recreate call this, so the two
+ * upstream shapes are never duplicated across create and update.
+ *
+ * @param {object} transport - Serenity transport (Semrush proxy client).
+ * @param {string} semrushWorkspaceId
+ * @param {string} projectId
+ * @param {{ text: string, tags: string[], tagIds: string[] | undefined }} input
+ * @returns {Promise<string>} the new upstream prompt id, or '' if the
+ *   response carried none.
+ */
+export async function createOnePrompt(transport, semrushWorkspaceId, projectId, input) {
+  if (input.tagIds !== undefined) {
+    const resp = await transport.createPromptsByIds(
+      semrushWorkspaceId,
+      projectId,
+      [input.text],
+      input.tagIds,
+    );
+    return Array.isArray(resp?.items) && resp.items.length > 0 ? String(resp.items[0].id) : '';
+  }
+  const resp = await transport.createTaggedPrompts(
+    semrushWorkspaceId,
+    projectId,
+    { [input.text]: input.tags },
+  );
+  return Array.isArray(resp?.ids) && resp.ids.length > 0 ? String(resp.ids[0]) : '';
+}
+
+/**
+ * Validates + normalizes a PATCH prompt body's `text`/`tags`/`tagIds`, shared
+ * by {@link handleUpdatePrompt} and its subworkspace twin. `tags` (names) and
+ * `tagIds` (upstream ids) are mutually exclusive, mirroring
+ * {@link normalizePromptInput}; exactly one must be present. Returns either
+ * `{ ok: true, text, tags, tagIds }` or `{ ok: false, status, body }` (the
+ * caller returns the latter directly as the handler's 400 response).
+ *
+ * @param {object} body - the PATCH request body.
+ * @returns {{ ok: true, text: string, tags: string[], tagIds: string[] | undefined }
+ *   | { ok: false, status: number, body: object }}
+ */
+export function parseUpdatePromptBody(body) {
+  const hasTagsField = body?.tags !== undefined;
+  const hasTagIdsField = body?.tagIds !== undefined;
+  if (!body || body.text === undefined || (!hasTagsField && !hasTagIdsField)) {
+    return {
+      ok: false,
+      status: 400,
+      body: {
+        error: 'missingFields',
+        message: 'PATCH body must include text and either tags or tagIds',
+      },
+    };
+  }
+  if (hasTagsField && hasTagIdsField) {
+    return {
+      ok: false,
+      status: 400,
+      body: { error: 'invalidRequest', message: 'tags and tagIds are mutually exclusive' },
+    };
+  }
+  const text = String(body.text);
+  if (hasTagIdsField) {
+    const tagIds = Array.isArray(body.tagIds)
+      ? body.tagIds.map((t) => String(t || '').trim()).filter(Boolean)
+      : [];
+    if (tagIds.length === 0) {
+      return {
+        ok: false,
+        status: 400,
+        body: { error: 'invalidRequest', message: 'tagIds must be a non-empty array' },
+      };
+    }
+    return {
+      ok: true, text, tags: [], tagIds,
+    };
+  }
+  const tags = Array.isArray(body.tags)
+    ? body.tags.map((t) => String(t || '').trim()).filter(Boolean)
+    : [];
+  return {
+    ok: true, text, tags, tagIds: undefined,
   };
 }
 
@@ -272,13 +377,12 @@ export async function handleCreatePrompts(
     }
     const projectId = project.getSemrushProjectId();
     try {
-      const resp = await transport.createTaggedPrompts(
+      const semrushPromptId = await createOnePrompt(
+        transport,
         semrushWorkspaceId,
         projectId,
-        { [input.text]: input.tags },
+        input,
       );
-      const semrushPromptId = Array.isArray(resp?.ids) && resp.ids.length > 0
-        ? String(resp.ids[0]) : '';
       return {
         created: {
           semrushPromptId,
@@ -286,6 +390,7 @@ export async function handleCreatePrompts(
           languageCode: input.languageCode,
           text: input.text,
           tags: input.tags,
+          ...(input.tagIds !== undefined ? { tagIds: input.tagIds } : {}),
         },
         affectedProjectId: projectId,
       };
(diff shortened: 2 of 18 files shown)


src/support/serenity/rest-transport.js
```diff
@@ -334,6 +334,61 @@ export function createSerenityTransport({ env, imsToken }) {
...
+     * set; `replace: true` REPLACES the set with exactly `references`. An
+     * unknown prompt `id` is skipped SILENTLY — live 204s with no "not
+     * found" signal, so this call alone cannot confirm a prompt exists.
+     * Response: 204 No Content (`unwrap` resolves to `null`).
+     *
+     * @param {string} semrushWorkspaceId
+     * @param {string} projectId
+     * @param {Array<{ id: string, references: string[], replace: boolean }>} items
+     */
+    async updatePromptTags(semrushWorkspaceId, projectId, items) {
```
rainer-friederich: issue (blocking): This method has no caller anywhere in src/ — grep confirms the only other reference is its own unit test. It ships a documented silent-failure footgun (an unknown prompt id 204s with no "not found" signal) with no real consumer to validate it against, and no integration coverage. Per this repo's "don't defer, follow-ups won't happen" convention: either wire it into the update path now (it's arguably the better primitive than the current delete-then-recreate churn, and would also fix finding 3 below), or drop the method + its test and reintroduce both alongside the PR that actually calls it.

src/support/serenity/handlers/prompts.js
```diff
@@ -184,18 +184,123 @@ export async function publishAffected(transport, semrushWorkspaceId, projectIds,
...
   const tags = Array.isArray(input?.tags)
     ? input.tags.map((t) => String(t || '').trim()).filter(Boolean)
     : [];
+  const tagIds = Array.isArray(input?.tagIds)
+    ? input.tagIds.map((t) => String(t || '').trim()).filter(Boolean)
+    : undefined;
   if (!text || languageCode === null || geoTargetId === null) {
     return null;
   }
+  if (tagIds !== undefined && (tags.length > 0 || tagIds.length === 0)) {
```
rainer-friederich: issue (blocking): normalizePromptInput treats exclusivity as content-based (tags.length > 0), but parseUpdatePromptBody (line 276, shared by the subworkspace twin) treats it as presence-based (body.tags !== undefined). The identical body {tags: [], tagIds: ['x']} is accepted here on create but 400s on update. Pick one rule and apply it to both — presence-based is simpler and matches the OpenAPI's "supply at most one" framing.

src/support/serenity/handlers/prompts.js
```diff
@@ -441,13 +543,9 @@ export async function handleUpdatePrompt(
...
   }
 
-  const resp = await transport.createTaggedPrompts(
-    semrushWorkspaceId,
-    projectId,
-    { [nextText]: nextTags },
-  );
-  const newSemrushPromptId = Array.isArray(resp?.ids) && resp.ids.length > 0
-    ? String(resp.ids[0]) : '';
+  const newSemrushPromptId = await createOnePrompt(transport, semrushWorkspaceId, projectId, {
```
rainer-friederich: issue (blocking): This createOnePrompt call is not wrapped in try/catch, unlike the deletePromptsByIds call just above it. createPromptsByIds is documented as atomic-500-on-unknown-tag-id (rest-transport.js), so if any tagId in nextTagIds became invalid between the client's list-read and this PATCH, the prompt is already deleted at this point and the create throws — the client sees a 502 and the prompt is gone with no recovery path. The same shape exists in the subworkspace twin at prompts-subworkspace.js:269. Either validate tagIds against the live tag tree before the delete, or catch this specific failure and return a distinct, actionable error instead of letting it propagate as a generic 502.

src/support/serenity/handlers/tags.js
```diff
@@ -210,23 +307,45 @@ export async function handleCreateTag(
...
+      semrushWorkspaceId,
+      projectId,
+      tag,
+      log,
+    );
+    log?.info?.('handleCreateTag: resolved closed-dimension tag', {
+      brandId, geoTargetId, languageCode, tag, created,
+    });
+    return {
+      status: 200,
```
rainer-friederich, line 336: issue (blocking): This always returns 200, even when created is true (a genuine new tag was created). The open-dimension create on this same endpoint (line 346) returns 201 for a real create. A client can no longer infer "resource created" from the status line for this endpoint — it must branch on dimension first, then on the created boolean. Prefer 201 when created:true and 200 only for the resolve-existing case, which keeps HTTP semantics intact while still carrying created.

src/support/serenity/handlers/prompts.js
```diff
@@ -184,18 +184,123 @@ export async function publishAffected(transport, semrushWorkspaceId, projectIds,
...
+ * existing contract unchanged.
+ */
 export function normalizePromptInput(input) {
   const text = String(input?.text || '').trim();
   const languageCode = normalizeLanguageCode(input?.languageCode);
   const geoTargetId = normalizeGeoTargetId(Number(input?.geoTargetId));
   const tags = Array.isArray(input?.tags)
     ? input.tags.map((t) => String(t || '').trim()).filter(Boolean)
     : [];
+  const tagIds = Array.isArray(input?.tagIds)
```
rainer-friederich: issue (blocking): tagIds entries only get String(t||'').trim() — no length cap and no control-char/whitespace rejection, unlike parentId which runs through validateParentIdFormat (tags.js) with a 200-char cap and a control-char reject. There's also no array-size cap here, while the GET list path caps parentId-adjacent reads at 50. These values are forwarded verbatim into the upstream POST aio/prompts body. Not a live injection (JSON body, not a URL), but worth hardening to match the parentId precedent — export/reuse validateParentIdFormat over each entry and cap the array length.

test/support/serenity/handlers/tags.test.js
```diff
@@ -686,6 +813,55 @@ describe('serenity tags handler (POST /serenity/tags)', () => {
...
+        },
+        fakeLog(),
+      );
+      expect(transport.updateProjectTag).to.have.been.calledOnceWithExactly(
+        WORKSPACE,
+        'proj-1',
+        'root-1',
+        { name: 'category:Footwear', parentId: undefined },
+      );
+    });
```
rainer-friederich, line 864: issue (blocking): The subworkspace update-tag test block (handleUpdateTagSubworkspace, further down this file) covers string re-parent and omitted-parentId (rename-only), but has no case for explicit parentId: null (promote-to-root) — the headline three-way-parentId feature this PR ships. The flat handler's equivalent block (this describe block, above) fully covers all three arms. Add one test mirroring the flat handler's null-parentId case so both twins exercise the same branch matrix.

docs/openapi/schemas.yaml
```diff
@@ -10894,14 +10894,25 @@ SerenityPromptListResponse:
 
 SerenityPromptInput:
   type: object
```
rainer-friederich, line 10980: issue (blocking): SerenityPrompt (defined just above SerenityPromptInput here) has no tags or tagIds property, unlike this sibling schema which correctly declares tagIds. But SerenityPrompt is the $ref for both SerenityCreatePromptsResponse.created and updateSerenityPrompt's 200 response — and this PR's own handlers (prompts.js/prompts-subworkspace.js, create + update) now return tags: input.tags and conditionally tagIds: input.tagIds in those exact response bodies. Add both fields (optional) to SerenityPrompt.

rainer-friederich (commented):
Hey @rainer-friederich,

**Verdict: Request changes** - solid, well-tested feature work with careful three-way `parentId` semantics, but 7 verified Important-tier issues should be addressed first.
_Changes: Adds id-based prompt writes, closed-dimension tag resolve-before-create, and promote-to-root support to the Serenity tags/prompts API - transport, twin handlers (flat + subworkspace), OpenAPI docs, and tests (15 files)._

Note: this count is above the usual 5-finding threshold for a PR this size, but each finding below is a distinct, narrow, single-function fix rather than overlapping duplicates - splitting the PR doesn't seem like the right call here, just flagging why the list is long. MysticatBot's own review crashed on this PR (`error_max_budget_usd`), which is why this local review exists.

## Must fix before merge
1. [Important] `updatePromptTags` is dead code - no caller anywhere in `src/`, ships an unexercised silent-skip footgun - `src/support/serenity/rest-transport.js:382` (details inline)
2. [Important] Prompt tag-selector mutual-exclusion is content-based on create but presence-based on update - identical body `{tags: [], tagIds: [...]}` succeeds on POST, 400s on PATCH - `src/support/serenity/handlers/prompts.js:210` (details inline)
3. [Important] Delete-then-recreate on prompt update isn't wrapped - if a `tagId` becomes invalid between list-read and PATCH, the atomic id-based create 500s AFTER the old prompt is already deleted, permanently losing it - `src/support/serenity/handlers/prompts.js:546` (details inline)
4. [Important] Closed-dimension tag create always returns 200, even when it genuinely creates a new tag - breaks the 201-means-created convention the open-dimension path on the same endpoint uses - `src/support/serenity/handlers/tags.js:325` (details inline)
5. [Important] `tagIds` array entries skip the length/control-char validation `parentId` gets via `validateParentIdFormat`, and have no array-size cap (list-read path caps at 50) - `src/support/serenity/handlers/prompts.js:204` (details inline)
6. [Important] Promote-to-root (`parentId: null`) is untested on the subworkspace twin - the flagship three-way-parentId feature has zero coverage in `handleUpdateTagSubworkspace` despite full coverage on the flat handler - `test/support/serenity/handlers/tags.test.js:988` (details inline)
7. [Important] `SerenityPrompt` OpenAPI schema doesn't declare `tags`/`tagIds`, even though this PR's own handlers now return them in exactly the response shapes that `$ref` this schema - `docs/openapi/schemas.yaml:10854` (details inline)

<details>
<summary>Non-blocking (7): minor issues and suggestions</summary>

- suggestion: `resolveOrCreateClosedTag` scans all root pages unconditionally before create, even though closed tags are seeded at project creation and almost always on page 1 - consider short-circuiting - `src/support/serenity/handlers/tags.js:839`
- nit: `createOnePrompt`'s id-path returns the literal string `'undefined'` if upstream ever omits `id` on an item, rather than `''` - `src/support/serenity/handlers/prompts.js:241`
- suggestion: extract the duplicated "map/trim/filter tag id array" logic shared by `normalizePromptInput` and `parseUpdatePromptBody` into one helper - `src/support/serenity/handlers/prompts.js:204`
- nit: the closed-dimension request schema still types `parentId` as `{type: string}` while the sibling update schema was changed to `[string, 'null']` in this same PR - `docs/openapi/serenity-api.yaml:521`
- suggestion: subworkspace closed-dimension tests cover only `created: true`; the `created: false` resolve-existing branch is untested on that twin - `test/support/serenity/handlers/tags.test.js:1754`
- suggestion: empty/filtered-to-empty `tagIds` on prompt create silently lands in `skipped` (vs. a loud 400 on update) - undocumented asymmetry, worth a test - `test/support/serenity/handlers/prompts.test.js`
- nit: an `unknown`-kind update target with `parentId: null` silently no-ops rather than erroring - matches the "defensive" comment but is untested - `src/support/serenity/handlers/tags.js:603`

</details>

<details>
<summary>Test quality snapshot</summary>

No signals fired - new-code/new-test ratio 1.38, assertion density 2.65/test, zero mocks. Healthy.

</details>

src/support/serenity/rest-transport.js
rainer-friederich: Fixed in `19e8b266` — removed the dead `updatePromptTags` transport method and its unit test.

src/support/serenity/handlers/prompts.js
rainer-friederich: Fixed in `19e8b266` — `normalizePromptInput` now uses the same presence-based exclusivity check as `parseUpdatePromptBody` (checks whether the `tags`/`tagIds` keys are present, not whether the array is non-empty), so `{tags: [], tagIds: [...]}` is rejected identically on create and update.

src/support/serenity/handlers/prompts.js
rainer-friederich: Fixed in `19e8b266` — wrapped in both twins (`prompts.js` and `prompts-subworkspace.js`). It still can't undo the already-successful delete, but the catch now logs a distinct "failed AFTER a successful delete" message before re-throwing, so this data-loss window is identifiable in logs instead of looking like an ordinary failed update. Added a test exercising this path in both handlers.

src/support/serenity/handlers/prompts.js
rainer-friederich: Fixed in `19e8b266` — extracted a shared `sanitizeTagIds` helper used by both `normalizePromptInput` and `parseUpdatePromptBody` (also addresses the non-blocking dedup suggestion below). It now filters entries through the same `isValidTagIdFormat` check `parentId` gets (length + control-char bound, moved to `validation.js` so both tags.js and prompts.js share it) and caps the array at `MAX_TAG_IDS` (50), mirroring the list-read query cap.

test/support/serenity/handlers/tags.test.js
rainer-friederich: Fixed in `19e8b266` — added the promote-to-root (`parentId: null`) test case mirroring the flat handler's gate-1 test.

docs/openapi/schemas.yaml
rainer-friederich: Fixed in `19e8b266` — added `tags`/`tagIds` to the `SerenityPrompt` schema, documenting that they're only present on a create/update response (list items carry `tagMap` instead).

src/support/serenity/handlers/tags.js
rainer-friederich: Declining, not fixing — on closer look this is intentional, not a bug: this same PR's OpenAPI spec (serenity-api.yaml, added alongside this handler) documents 200 uniformly for the closed-dimension response with "already existed or was just created — see `created`", precisely because this is an idempotent resolve-or-create endpoint. The 200/201 split on this route is dimension-type-based (closed vs open), not created-state-based — there's no "201-means-created" convention actually established here for it to break. Making the status code vary with server-side create/resolve state would make it non-deterministic from the caller's perspective for what's meant to be a single idempotent operation; the `created` boolean is the intended signal (elmo-ui's client already reads it, not the status code). Happy to revisit if there's a concrete client-side reason to prefer varying the status instead.

rainer-friederich:
Addressed the local pr-review (https://github.com/adobe/spacecat-api-service/pull/2741#pullrequestreview-4618335592) in `19e8b266`.

**Fixed (6 of 7):**
- `updatePromptTags` dead code — removed, along with its unit test.
- Create/update tagIds-vs-tags mutual-exclusion asymmetry — `normalizePromptInput` now uses the same presence-based check `parseUpdatePromptBody` already used, so an identical body is accepted/rejected identically on both endpoints.
- Uncaught delete-then-recreate — wrapped in both twins (`prompts.js` + `prompts-subworkspace.js`); a post-delete create failure is now logged as a distinct data-loss event before re-throwing, with a test covering the path in both handlers.
- `tagIds` array validation — extracted a shared `sanitizeTagIds` helper (also resolves the non-blocking dedup suggestion) that reuses `parentId`'s length/control-char format check (moved to `validation.js` as `isValidTagIdFormat`, shared by `tags.js` and `prompts.js`) and caps the array at `MAX_TAG_IDS` (50).
- Subworkspace promote-to-root test gap — added, mirroring the flat handler's gate-1 test.
- `SerenityPrompt` OpenAPI schema — added `tags`/`tagIds`, which the create/update responses already populate.

**Declined (1 of 7):**
- Closed-dimension tag create always returning 200 — on closer look this is intentional, not a bug. This same PR's OpenAPI spec documents 200 uniformly for the closed-dimension response ("already existed or was just created — see `created`"), because this is an idempotent resolve-or-create endpoint; the 200/201 split is dimension-type-based, not created-state-based, so there's no "201-means-created" convention here to break. Full rationale on the inline thread.

All 7 non-blocking nits/suggestions from the original review were either folded into the fixes above (dedup helper) or left as-is where the review itself only flagged them as optional. Tests (264 passing, 100% statement/line coverage on touched files), lint, tsc, and the OpenAPI contract/lint checks are all green.

rainer-friederich (commented):
Hey @rainer-friederich,

**Verdict: Approve** - both re-review findings resolved cleanly, and the finding-4 decline holds. (Posted as a comment, not a formal approval: GitHub blocks a PR author from submitting a review state on their own PR via the API - see the earlier review on this PR for the same platform restriction.)
_Changes: re-reviewed fix commit 19e8b266 against the prior 7 findings, then fixed the 2 gaps that surfaced (f167717f)._

- `updatePromptTags` dead code - removed, zero remaining references confirmed.
- Create/update tagIds-vs-tags mutual-exclusion asymmetry - now presence-based on both paths, and the exact regression scenario (`{tags: [], tagIds: [...]}` on create) now has a dedicated test that was verified to fail against the old logic and pass against the fix.
- Uncaught delete-then-recreate data-loss window - wrapped in both twins with a log message distinct from the pre-delete failure; re-throw preserves the original error.
- `tagIds` array format validation + 50-item cap - `tags.js`'s `validateParentIdFormat`/`requireTagId` are now wired to the same shared `isValidTagIdFormat` (validation.js) that gates `tagIds`, so the bound genuinely can't diverge between the two call sites (it previously still had its own independent copy despite the helper's docstring claiming otherwise - closed in `f167717f`).
- Subworkspace promote-to-root test gap - added, asserts `parentId: null` is forwarded to `updateProjectTag`.
- `SerenityPrompt` OpenAPI schema missing `tags`/`tagIds` - added, consistent with the actual create/update response shapes.
- Closed-dimension tag create always returning 200 - decline stands. Independently confirmed sound: the endpoint is a documented idempotent resolve-or-create, `serenity-api.yaml` specifies 200 uniformly for the closed path, and the `created` boolean (which elmo-ui reads) carries the created-state signal rather than the status code.

<details>

- nit: the open-dimension tag create (201) response omits a `created` field while the closed-dimension (200) response includes one, so a client can't read `created` uniformly across both dimension types - pre-existing, not introduced by recent commits - `src/support/serenity/handlers/tags.js:411`
- nit: the flat-mode data-loss test doesn't assert `publishProject` is skipped after the re-throw, unlike its subworkspace twin which does - `test/support/serenity/handlers/prompts.test.js:1202`
- suggestion: the pre-delete failure test only asserts the re-throw, not that the data-loss log message is absent, so the pre/post distinction is only locked from one direction - `test/support/serenity/handlers/prompts.test.js:1172`
- nit: the closed-dimension tag-create request schema still types `parentId` as `{type: string}` while sibling schemas use `[string, 'null']` - pre-existing, out of scope for this fix - `docs/openapi/serenity-api.yaml:521`

</details>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-structured feature work with careful three-way parentId semantics; all prior Important findings addressed cleanly.
**Complexity: HIGH** - large diff; API surface + dependency signal.
_Changes: Adds id-based prompt writes, closed-dimension tag resolve-before-create, and promote-to-root support to the Serenity tags/prompts API surface (17 files)._

<details>

- nit: the closed-dimension tag-create request schema still types `parentId` as `{type: string}` while the sibling update schema in this same PR uses `[string, 'null']` - inconsistent within the same OpenAPI file, even though parentId is rejected for closed dimensions at the handler layer - `docs/openapi/serenity-api.yaml:521`
- suggestion: `resolveOrCreateClosedTag` fetches all root pages unconditionally before attempting create, even though closed tags are seeded at project creation and will almost always be on page 1 - consider short-circuiting after page 1 if found, to avoid unnecessary upstream calls on projects with many root tags - `src/support/serenity/handlers/tags.js:951`
- nit: `createOnePrompt` returns the literal string `'undefined'` if upstream ever omits `id` on a response item (`String(resp.items[0].id)` when `id` is `undefined`) - returning `''` would be consistent with the name-based path's fallback - `src/support/serenity/handlers/prompts.js:545`
- nit: the flat-mode data-loss test does not assert `publishProject` is skipped after the re-throw, unlike its subworkspace twin which does - `test/support/serenity/handlers/prompts.test.js:1222`

</details>

- `updatePromptTags` dead code removed.
- Create/update tagIds-vs-tags mutual-exclusion asymmetry fixed (now presence-based on both paths).
- Delete-then-recreate data-loss window wrapped with distinct logging in both twins.
- `tagIds` array validation added via shared `sanitizeTagIds` + `isValidTagIdFormat` + 50-item cap.
- Subworkspace promote-to-root test gap filled.
- `SerenityPrompt` OpenAPI schema updated with `tags`/`tagIds`.
- Closed-dimension 200 status declined with sound rationale (idempotent resolve-or-create; `created` boolean carries the signal).

rainer-friederich:
Addressed all 4 non-blocking nits from the last review in `3f8ae6f9` / `8ac33677`:

- OpenAPI: create-tag `parentId` now types `[string, 'null']`, matching the PATCH endpoint's shape (create's `parseParentId` already treats explicit `null` as a no-op root create, same as omission).
- `resolveOrCreateClosedTag` now short-circuits `listProjectTagTree` as soon as a matching root tag is found instead of always walking every page.
- `createOnePrompt` returns `''` instead of the literal string `'undefined'` when an id-based create response item carries no `id` (added a regression test for this).
- The flat-mode data-loss test now asserts `publishProject` is skipped after the post-delete create re-throw, matching its subworkspace twin.

713/713 tests passing, lint/tsc clean.

PR #2749, opened 2026-07-03
feat(serenity): dynamic AI resource allocation — allocator core + transport read (LLMO-5616)

byteclimber:
**Spec / design:** [serenity-docs#22 — Dynamic Resource Allocation](https://github.com/adobe/serenity-docs/pull/22) (Gate-0 live-verified). Companion to the mock foundation [adobe/spacecat-shared#1767](https://github.com/adobe/spacecat-shared/pull/1767). This is **PR 2 (allocator core, unit-only)** of the api-service plan; PR 3 (fronting the sub-workspace handlers + flag-gated carve) follows and gates its IT on the shared mock release.

## What

The pure, transport-injected **dynamic just-in-time AI allocator** plus the resource read and error classification — no handler wiring, flag, or carve change yet.

- **`rest-transport.js`** — `getWorkspaceResources(workspaceId)`. Read the **master's own** `/resources` for the org pool (`/parent/resources` returns child-relative numbers — Gate 0).
- **`errors.js`** — `ORG_POOL_EXHAUSTED` / `BRAND_AI_LIMIT` codes; body/message classifiers that separate the terminal `422 "insufficient available units"` (`isPoolExhausted`) from the transient `422 "workspace not ready"` (`isWorkspaceNotReady`), plus `isMeteredQuota` (405) and `isRateLimited` (429).
- **`resource-manager.js`** — `ensureAiHeadroom` (hot-path read, block-sized top-up gated on ceiling + master pool) / `releaseAiSurplus` (best-effort, floor-clamped); `roundUpToBlock`, `PROJECT_BLOCK=1` / `PROMPT_BLOCK=100`, `need` calculators, a **strict fail-loud** `readAiTotals` over the nested `product_resources.ai.resources.*`, a **budget-sized** settle poll + **bounded** "workspace not ready" retry, typed `409`s via `ErrorWithStatusCode.code`.
- **`workspace-lifecycle.js`** — export `pollUntilCreated` for reuse.

## Design notes (from the 4-expert plan review)

- **No `mapError` change** — `orgPoolExhausted`/`brandAiLimit` flow via `.code`; the existing `422` handling is untouched so flag-OFF stays byte-for-byte (the current carve `422`s today).
- Reads keep `drafted` so PR 3 can front the prompt dimension at the **publish seam** using `used + drafted` (staleness-immune vs async-reconciled `used`).
- Poll is budget-sized (not the 30-attempt provisioning ceiling); release is best-effort + floor-clamped for the async/reconciler path.

[...]

PR #2750, opened 2026-07-03
feat(serenity): LLMO/dynamic-allocation feature flag + auth threading (LLMO-5616)

byteclimber:
**Spec / design:** [serenity-docs#22](https://github.com/adobe/serenity-docs/pull/22). Part of the dynamic just-in-time Semrush AI resource-allocation work (api-service side); companion to allocator-core PR #2749 and mock PR [spacecat-shared#1767](https://github.com/adobe/spacecat-shared/pull/1767).

## What

The per-org **`LLMO/dynamic-allocation` feature flag** and its plumbing — **foundation only, inert**:
- `dynamic-allocation-active.js` — `isDynamicAllocationActiveForOrg` (cloned from `serenity-active.js`: own TTL/size-bounded cache, fail-safe **OFF**, reuses the resolver's cache constants).
- `authorize()` now reads it in parallel with `isSerenityActiveForOrg` and threads the value onto the auth object as `dynamicAllocation`. It is **NOT a gate** — nothing consumes it yet.

Default **OFF**, and per the rollout plan it stays OFF for every org until the PR-4 safeguards (rightsizing sweep, concurrency serialization, SLIs, live-gateway canary) land. Because no code branches on it yet, behavior is **byte-for-byte unchanged**.

## Test plan

`npm run type-check` ✓ · `npm run lint` ✓ · new module **100% covered** (12 unit tests: identity, true/false/null default-OFF, blank org, missing client, read-failure fail-safe, cache TTL/eviction, no-cache-on-error); 185 serenity controller tests still pass. Pure unit — no IT.

### DIFF
--- a/src/support/serenity/dynamic-allocation-active.js
+++ b/src/support/serenity/dynamic-allocation-active.js
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+// @ts-check
+
+import { hasText } from '@adobe/spacecat-shared-utils';
+
+import { readFeatureFlag } from '../feature-flags-storage.js';
+import { CACHE_TTL_MS, NEG_TTL_MS, MAX_ENTRIES } from './workspace-resolver.js';
+
+/**
+ * The per-org rollout switch for **dynamic (just-in-time) AI resource allocation**, stored in the
+ * `feature_flags` table (organization_id + product + flag_name). It is a SECOND, independent flag
+ * from `LLMO/serenity`: serenity must already be active for an org, and this flag additionally
+ * decides whether the sub-workspace metered ops top up JIT (ON) or keep the flat pre-calculated
+ * carve (OFF).
+ *
+ * DEFAULT OFF and, per the dynamic-allocation rollout plan, it stays OFF for every org until the
+ * PR-4 safeguards (per-org rightsizing sweep, concurrency serialization, the SLIs, and the
+ * live-gateway canary) are in place — flipping it ON before then is unsafe. A missing/`false` row,
+ * an unavailable PostgREST client, or a transient read error all resolve to `false`.
+ */
+export const DYNAMIC_ALLOCATION_FEATURE_FLAG_PRODUCT = 'LLMO';
+export const DYNAMIC_ALLOCATION_FEATURE_FLAG_NAME = 'dynamic-allocation';
+
+/**
+ * Module-scoped TTL+size-bounded cache — its own Map, separate from the serenity-flag cache, so the
+ * two flags never collide. Positive value cached for the positive TTL; `false`/absent for the
+ * shorter negative TTL so flipping an org ON takes effect within ~NEG_TTL_MS. Reuses the resolver's
+ * TTL/size constants to stay in lockstep. Exported clear is for unit tests only.
+ */
+const flagCache = new Map();
+
+/**
+ * @param {string} spaceCatId - SpaceCat organization UUID.
+ * @returns {string} composite cache key (org + product + flag).
+ */
+function flagCacheKey(spaceCatId) {
+  return `${spaceCatId}::${DYNAMIC_ALLOCATION_FEATURE_FLAG_PRODUCT}::${DYNAMIC_ALLOCATION_FEATURE_FLAG_NAME}`;
+}
+
+export function clearDynamicAllocationFlagCache() {
+  flagCache.clear();
+}
+
+function evictIfNeeded() {
+  while (flagCache.size >= MAX_ENTRIES) {
+    const oldest = flagCache.keys().next().value;
+    /* c8 ignore start -- defensive: size>=MAX_ENTRIES implies a key exists */
+    if (oldest === undefined) {
+      break;
+    }
+    /* c8 ignore stop */
+    flagCache.delete(oldest);
+  }
+}
+
+/**
+ * Is dynamic (JIT) AI resource allocation active for an organization? Reads the org-wide
+ * `LLMO/dynamic-allocation` feature flag (cached). Default OFF: a missing flag row, a `false` row,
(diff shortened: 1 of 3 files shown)


src/support/serenity/dynamic-allocation-active.js
```diff
@@ -0,0 +1,116 @@
...
+ * carve (OFF).
+ *
+ * DEFAULT OFF and, per the dynamic-allocation rollout plan, it stays OFF for every org until the
+ * PR-4 safeguards (per-org rightsizing sweep, concurrency serialization, the SLIs, and the
+ * live-gateway canary) are in place — flipping it ON before then is unsafe. A missing/`false` row,
+ * an unavailable PostgREST client, or a transient read error all resolve to `false`.
+ */
+export const DYNAMIC_ALLOCATION_FEATURE_FLAG_PRODUCT = 'LLMO';
+export const DYNAMIC_ALLOCATION_FEATURE_FLAG_NAME = 'dynamic-allocation';
+
```
MysticatBot, line 36: issue (blocking): The constant `DYNAMIC_ALLOCATION_FEATURE_FLAG_NAME` is set to `'dynamic-allocation'` (hyphenated), but `feature-flags-storage.js` defines `FLAG_NAME_PATTERN = /^[a-z][a-z0-9_]*$/` (underscores only). The admin controller (`src/controllers/feature-flags.js:95`) validates flag names with `isValidFeatureFlagName()` and returns 400 with the message "use non-empty lowercase snake_case" when this pattern fails.

This means the flag cannot be created through the admin API with this name. Either:
1. Rename to `dynamic_allocation` (snake_case) to match the validation, or
2. Relax `FLAG_NAME_PATTERN` to `/^[a-z][a-z0-9_-]*$/` to permit hyphens (this changes the API contract for all flags).

Since the existing `serenity` flag uses a plain name without separators, option 1 is the path of least resistance.

MysticatBot (changes requested):
Hey @byteclimber,

**Verdict: Request changes** - one naming conflict with the feature-flags validation layer needs fixing.
**Complexity: LOW** - small, single-service, one new module cloned from an existing pattern.
_Changes: Adds the per-org LLMO/dynamic-allocation feature flag reader with TTL-bounded cache and threads the boolean onto the authorize() return object in SerenityController (3 files)._

## Must fix before merge

1. [Important] Flag name `dynamic-allocation` fails admin API validation - `src/support/serenity/dynamic-allocation-active.js:34` (details inline)

<details>

- suggestion: Add a controller-level test asserting that `authorize()` returns `dynamicAllocation` in its success object - `src/controllers/serenity.js:378`. Downstream PRs will depend on this contract; catching a regression at the controller layer is cheaper than discovering it during integration.
- suggestion: Consider extracting the TTL+size-bounded flag cache into a shared `createFlagChecker(product, flagName)` factory. `dynamic-allocation-active.js` is structurally identical to `serenity-active.js` (same imports, same guards, same eviction, same TTL constants). Two clones is manageable but sets a precedent. A tracking issue would suffice if out of scope here.
- nit: Positive-TTL semantics cache a `true` value for 5 minutes. If this flag is later used as an emergency kill switch for metered spend, a 5-minute stale window on `OFF` may be operationally risky. Worth a conscious decision when the flag is actually activated (not blocking now).

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 11m 15s | <b>Cost:</b> $7.88 | <b>Commit:</b> <code>cf56a3f6208b44ae8e901a3ef6cb4352b291cd6e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

byteclimber:
Fixed in `c9ef02e4` — renamed the flag to **`dynamic_allocation`** (lowercase snake_case, matching `FLAG_NAME_PATTERN = ^[a-z][a-z0-9_]*$` in `feature-flags-storage.js`; the hyphen was the problem).

Non-blocking:
- **Controller test asserting `authorize()` returns `dynamicAllocation`** — the field is currently *inert* (nothing reads it), so there's no controller-observable behavior to assert yet. The contract test lands naturally with the follow-up **fronting PR** that consumes it (a handler that branches on `auth.dynamicAllocation`), where it's actually exercisable.
- **Shared `createFlagChecker(product, flagName)` factory** — agreed the two modules are near-identical; out of scope here (would also touch the shipped `serenity-active.js`), tracking as a follow-up refactor.
- **Positive-TTL kill-switch staleness** — noted for the PR-4 activation step (the flag stays OFF until then); we'll pick the TTL / add a fast-OFF path when it's actually enabled for spend.

type-check + lint green; 12 unit tests pass. Re-requesting review.

src/support/serenity/dynamic-allocation-active.js
```diff
@@ -0,0 +1,118 @@
...
+ * Is dynamic (JIT) AI resource allocation active for an organization? Reads the org-wide
+ * `LLMO/dynamic-allocation` feature flag (cached). Default OFF: a missing flag row, a `false` row,
+ * an unavailable PostgREST client, or a transient read error all resolve to `false`, so an org is
+ * never silently switched to JIT allocation.
+ *
+ * @param {object} ctx - Request context (uses `ctx.dataAccess.services.postgrestClient`).
+ * @param {string} spaceCatId - SpaceCat organization UUID.
+ * @param {object} [log] - Optional logger (surfaces a missing client / read error, never throws).
+ * @returns {Promise<boolean>} `true` only when the org-wide dynamic-allocation flag is on.
+ */
```
MysticatBot, line 79: nit: This JSDoc still references `LLMO/dynamic-allocation` (hyphenated) but the constant was corrected to `dynamic_allocation` in the fix commit. Worth aligning so a future incident grep finds the right string.

MysticatBot (approved):
Hey @byteclimber,

**Verdict: Approve** - prior flag-name finding addressed; code is sound.
**Complexity: LOW** - small, single-service, one new module cloned from an existing pattern.
_Changes: Adds the per-org LLMO/dynamic_allocation feature flag reader with TTL-bounded cache and threads the boolean onto the authorize() return object (3 files)._

- Flag name changed from `dynamic-allocation` to `dynamic_allocation` to match `FLAG_NAME_PATTERN` validation.

<details>

- nit: JSDoc on `isDynamicAllocationActiveForOrg` still says `LLMO/dynamic-allocation` (hyphenated) but the constant is now `dynamic_allocation` - `src/support/serenity/dynamic-allocation-active.js:79`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 15m 6s | <b>Cost:</b> $5.91 | <b>Commit:</b> <code>c9ef02e47b4c25cfd3f30ab5449c1d9dec4438f8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

byteclimber:
Superseded by #2764 — the per-org DB flag is replaced by a global env/Vault kill-switch per Rainer's direction on serenity-docs#22.

PR #2751, opened 2026-07-03
feat(serenity): flag-gated carve — skip flat re-grant under dynamic allocation (LLMO-5616)

byteclimber:
**Spec / design:** [serenity-docs#22](https://github.com/adobe/serenity-docs/pull/22). Part of the dynamic just-in-time Semrush AI resource-allocation work (api-service side). Independent of the flag PR #2750 and allocator-core PR #2749 (this touches only `ensureSubworkspace`).

## What

Makes `ensureSubworkspace` **dual-mode** via an options bag. When `{ dynamicAllocation: true }` it **skips the flat `resourceAllocation(marketCount)` re-grant** on an already-bound sub-workspace — that up-front pre-sized carve is exactly the over/under-allocation JIT replaces, so the metered handlers top up just-in-time instead. Flag **OFF (default)** is **byte-for-byte unchanged**.

Why skip rather than re-grant under JIT: re-flattening `total` would undo a JIT top-up and, on an ON→OFF rollback of an already-grown child, risk setting `total` below `used` (instant 405).

**Inert for now:** no caller passes the option yet (it defaults OFF), so this is capability + tests only; the handler wiring that threads `auth.dynamicAllocation` through lands in the follow-up fronting PR.

## Test plan

`npm run type-check` ✓ · `npm run lint` ✓ · unit tests assert **ON → skips the re-grant** and **OFF → byte-for-byte** (`transferWorkspaceResources` called with `resourceAllocation(2)`), spying the transport call. 46 workspace-lifecycle tests pass. Pure unit — no IT.

### DIFF
--- a/src/support/serenity/workspace-lifecycle.js
+++ b/src/support/serenity/workspace-lifecycle.js
@@ -259,6 +259,9 @@ async function adoptFromFamily(transport, parentWorkspaceId, title, log) {
  *   re-reads the brand's CURRENT semrush_workspace_id from the data layer.
  *   When supplied, the create path uses it as a last-update concurrency guard
  *   (see below) so a parallel activation cannot orphan a resourced workspace.
+ * @param {object} [options] - feature-flag toggles for the dual-mode carve.
+ * @param {boolean} [options.dynamicAllocation] - when true (LLMO/dynamic-allocation ON), skip
+ *   the flat re-grant on an existing sub-workspace; JIT top-up owns sizing. Default false.
  * @returns {Promise<string>} the subworkspace id.
  */
 export async function ensureSubworkspace(
@@ -269,7 +272,9 @@ export async function ensureSubworkspace(
   log,
   timing = {},
   reloadPointer = null,
+  options = {},
 ) {
+  const { dynamicAllocation = false } = options;
   const poll = {
     attempts: timing.attempts ?? DEFAULT_POLL_ATTEMPTS,
     intervalMs: timing.intervalMs ?? DEFAULT_POLL_INTERVAL_MS,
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @byteclimber,

**Verdict: Approve** - clean, well-scoped flag gate with correct default-off behavior and solid test coverage.
**Complexity: LOW** - small diff, single internal function, no risk signals.
_Changes: Adds a `dynamicAllocation` option to `ensureSubworkspace` that skips the flat resource re-grant when JIT allocation owns sizing (2 files)._

<details>

- suggestion: Positional parameter creep - the function now accepts 8 positional params; callers must pass `null` for `reloadPointer` to reach `options`. Consider consolidating `timing`, `reloadPointer`, and `options` into a single config bag in the follow-up PR that wires callers - `src/support/serenity/workspace-lifecycle.js:272`
- suggestion: The "dynamic-allocation ON" test asserts `transferWorkspaceResources` is NOT called but does not assert `pollUntilCreated` (via `getWorkspaceStatus`) IS still called once; adding that would catch a future accidental removal of the pre-poll - `test/support/serenity/workspace-lifecycle.test.js:93`
- nit: The first `pollUntilCreated` call still runs unconditionally when `dynamicAllocation` is true; a one-line comment above it clarifying "settle regardless of mode - workspace must be created before return" would help future readers - `src/support/serenity/workspace-lifecycle.js:296`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 35s | <b>Cost:</b> $3.47 | <b>Commit:</b> <code>95934e3f059c71c0a5ecd5ea673691acfaa83eca</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

byteclimber:
Applied the two cheap nits in `a2b026af`: the ON test now asserts `getWorkspaceStatus` (the readiness settle) is still called once, and a comment clarifies the pre-poll runs regardless of mode. The positional-param creep (consolidate `timing`/`reloadPointer`/`options` into one config bag) I'll fold into the follow-up **caller-wiring PR** that touches all the call sites anyway — cleaner to do it there than churn signatures twice. Thanks for the approve.

byteclimber:
Superseded by #2764 — carve-skip consolidated with its consumer per Rainer's direction on serenity-docs#22.

PR #2753, opened 2026-07-03
feat: add support for promise token for semrush api authentication

vivesing:
## Summary

Adds support for `x-promise-token`-based authentication to the Semrush proxy endpoints, alongside the existing IMS bearer flow.

https://jira.corp.adobe.com/browse/LLMO-5990

- `src/controllers/serenity.js` — all Serenity/Semrush proxy handlers (prompts, markets, tags, models, activate/deactivate, etc.)
- `src/controllers/elements.js` — Semrush Elements API proxy (`GET /v2/orgs/:spaceCatId/serenity/all/brand-presence/url-inspector/filter-dimensions`)

### What changed

Both controllers previously required the caller to send `Authorization: Bearer <ims-access-token>` directly, and rejected any request whose `authInfo.getType() !== 'ims'` (i.e. anything other than a real IMS-authenticated session).

Each controller now resolves the token to forward to Semrush via a new helper (`resolveSemrushImsToken` in `serenity.js`, `resolveElementsImsToken` in `elements.js`):

1. **If `x-promise-token` is present** — decode it and exchange it for a fresh IMS access token via `exchangePromiseToken` (existing shared helper, same one used by `preflight.js` / `fixes.js` / `edge-routing-auth.js`). The exchanged IMS token is what gets forwarded to Semrush. This path **does not** require `Authorization` to carry an IMS token — a spacecat JWT (or any other auth type) is fine, since the IMS-type check is skipped entirely for this path.
2. **If `x-promise-token` is absent** — falls through unchanged to the existing bearer-forwarding logic: `Authorization` must carry a real IMS token, and non-IMS auth types are rejected with 401.

### Backward compatibility

**Fully backward compatible — no client changes required for existing integrations.**

- When `x-promise-token` is not sent, the code path is byte-for-byte identical to before: same IMS-type check, same `Authorization: Bearer <ims-token>` requirement, same token forwarded to Semrush, same error messages for the pre-existing failure cases (missing header, invalid header format).
- The new logic only activates when the caller explicitly opts in by sending `x-promise-token`.
- Existing callers sending `Authorization: Bearer <ims-access-token>` and no `x-promise-token` continue to work exactly as today.

### How clients should send `x-promise-token`

1. **Mint a promise token** by calling `POST /auth/v2/promise` (spacecat-auth-service) with a valid IMS access token in `Authorization: Bearer <ims-access-token>`. The response body contains:
   ```json
   { "promiseToken": "<token>", "expiresIn": <seconds> }
   ```
2. **Send it on subsequent Serenity/Elements calls** as:
   ```
   x-promise-token: <token>
   ```
   (URL-encoding is supported — the header is decoded with `decodeURIComponent` before exchange, so a percent-encoded token is handled transparently.)
3. **`Authorization` still needs *some* value**, but it no longer needs to be an IMS token when `x-promise-token` is present — a spacecat-issued JWT (or whatever the caller already authenticates to spacecat with) is sufficient, since spacecat's own authorization (`AccessControlUtil`) runs independently of the promise-token exchange.

### Token lifetimes

| Token | Lifetime | Notes |
|---|---|---|
| Promise token (`x-promise-token`) | ~4 hours (`expires_in` returned by IMS at mint time; not hardcoded) | Client must re-mint via `/auth/v2/promise` before it expires. The exchange endpoint used here does **not** return a rotated promise token to the client — the original one must last the full session or be proactively refreshed. |
| Exchanged IMS access token (server-side only) | ~5 minutes | Never exposed to the client; used only for the single upstream Semrush call. |

Recommendation for client-side refresh: re-mint a few minutes before the ~4-hour expiry (not exactly at expiry) to avoid a request landing in the expired window.

### Error handling

- **Expired/invalid/malformed promise token** → `401` with `{ "error": "authenticationRequired", "message": "Invalid or expired promise token" }`. No upstream exchange failure detail is leaked in the response (logged server-side only).
- **Non-IMS auth type with no `x-promise-token` sent** → `401` with `{ "error": "promiseTokenRequired", "message": "Serenity/Elements proxy requires IMS authentication; send the x-promise-token header instead" }` — gives non-IMS callers an explicit, machine-readable hint (`promiseTokenRequired`) to switch to the promise-token flow instead of a generic "not authenticated" message.

### Tests

- `test/controllers/serenity.test.js` — 6 new tests (promise-token exchange success, URL-decoding, exchange failure → 401, fallback-when-absent, non-IMS-gate-unaffected, `promiseTokenRequired` hint).
- `test/controllers/elements.test.js` — same 6 tests, mirrored for the Elements controller.
- All existing tests pass unchanged (191 in `serenity.test.js`, 35 in `elements.test.js`); lint and `tsc --noEmit` clean.


rainer-friederich:
Reviewed this in the context of the full 2026-08-03 IMS-removal blast radius. The mechanism here is sound — `x-promise-token` → `exchangePromiseToken` → forward a fresh user token to Semrush is the right replacement, and the fallback-when-absent path keeps it backward compatible. Two things worth addressing before we call the Semrush surface "covered":

**1. Gap — the third IMS gate is not covered (brand provisioning / brand-edit re-sync).**

This PR migrates the two `requireImsBearer` gates (`serenity.js`, `elements.js`). But there is a third IMS-hard-gate that also forwards the user's IMS token to Semrush and will 401 after 2026-08-03:

- `getImsUserTokenStrict` (`src/support/utils.js`) → used by brand create/edit re-sync in `src/controllers/brands.js` (~L1851 and ~L1916) and by `src/support/serenity/brand-provisioning.js` (`provisionBrandSubworkspace` ~L108, `releaseProvisionedWorkspace` ~L229).

These are on the same "forward user bearer to Semrush" model as the serenity/elements proxies, so once `AdobeImsHandler` is removed they break identically. They need the same `x-promise-token` → exchange treatment (or the same `resolveSemrushImsToken` helper) that this PR introduces. Right now brand create/edit/provision via api-service is left uncovered.

--> Perhaps it would be good to unify these controllers?

calvarezg:
## Additional review findings

Ran a multi-persona pass (Architect/DBA/Security/Tester) on top of the existing review. Skipping anything that duplicates @rainer-friederich's comment (the third IMS gate in `brands.js`/`brand-provisioning.js` — confirmed still open, but already tracked there). New findings below:

### Major — promise-token identity isn't cross-checked against the authorizing session's org

`authorize()` in `serenity.js` grants access based on `AccessControlUtil.hasAccess()`, which derives org access from `authInfo.hasOrganization(imsOrgId)` — i.e. from the credential on `Authorization` (JWT or IMS). `resolveSemrushImsToken`/`resolveElementsImsToken` exchange `x-promise-token` for an IMS token completely independently of that check — nothing verifies the promise token's identity actually corresponds to the org/session that just passed `authorize()`.

Contrast with the existing `authorizeEdgeCdnRouting` in `src/support/edge-routing-auth.js`, which derives its authorization decision from `context.imsClient.getImsUserProfile(imsUserToken)` — the *same* token it forwards. This PR introduces a second pattern that decouples "who's authorized" from "whose token gets forwarded to Semrush."

A caller with a non-IMS credential (JWT) that has legitimate access to org X could supply *any* valid promise token — for a different IMS identity — and that unrelated identity's exchanged token is what's forwarded to Semrush for org X's workspace. Exploiting this requires already holding another identity's promise token, so it's not trivially exploitable by an anonymous attacker, but it's an implicit trust decision (pushed entirely onto Semrush's own enforcement) that's worth making explicit.

**Suggested fix:** either cross-check the exchanged token's subject/org against `authInfo`'s claimed org (mirroring `authorizeEdgeCdnRouting`'s pattern), or explicitly confirm/document that Semrush-side enforcement is the intended trust boundary for this specific path (the current code comment only covers that claim for the plain-bearer-forwarding path, not the new promise-token path).

### Major — duplicated resolve-token logic already diverging from the existing implementation

`resolveElementsImsToken` (`elements.js`) and `resolveSemrushImsToken` (`serenity.js`) are copy-pasted near-verbatim (decode-with-fallback → exchange → catch-and-401 → fallback-to-`requireImsBearer`). This is now the third implementation of "read `x-promise-token` → exchange" in the codebase, and it already behaves differently from the first: `edge-routing-auth.js`'s `getImsTokenFromPromiseToken` calls `decodeURIComponent` unconditionally with no try/catch, while these two silently fall back to the raw header value on decode failure. Same header, three code paths, two behaviors.

**Suggested fix:** extract a single shared helper into `support/utils.js` alongside `exchangePromiseToken`, and have `edge-routing-auth.js` (and the eventual brand-provisioning fix) consume it too — avoids a 4th hand-copy.

### Minor — duplicated error code string

`err.code = 'promiseTokenRequired'` is set as a literal string in both `elements.js` and `serenity.js`. `X_PROMISE_TOKEN_HEADER` was correctly centralized in `constants.js`, but its paired error code wasn't.

### Minor — promise-token path only test-verified at 1 of ~16 call sites in `serenity.js`

The new "x-promise-token support" describe block only drives the helper through `listPrompts`. The other 13+ swapped handlers (createPrompts, markets, tags, models, activate/deactivate, etc.) are mechanically identical today but aren't exercised with a promise token — a future edit that diverges one handler from the pattern wouldn't be caught. Suggest adding coverage on a mutating handler (e.g. `createPrompts`) and one with branching logic (e.g. `listMarkets`).

---
No Critical findings; core mechanism (promise-token exchange with unchanged bearer fallback) is sound and backward-compatible.

vivesing:
Gap — the third IMS gate is not covered (brand provisioning / brand-edit re-sync). - Fixed, added support for passing x-promise-token in this controller as well, also using common utility for all controllers extracting ims access token from promise token

vivesing:
> Major — promise-token identity isn't cross-checked against the authorizing session's org

Semrush authenticating via ims token will be replaced with another mechanism this is temp authN approach being followed

> Major — duplicated resolve-token logic already diverging from the existing implementation

Moved this logic to a common util for all controllers not using this 

> Minor — duplicated error code string

Addresed, using constant everywhere

src/support/utils.js
```diff
@@ -764,6 +775,55 @@ export async function exchangePromiseToken(context, promiseToken) {
...
+      log?.error(`${logLabel}: promise token exchange failed`, { error: e?.message });
+      throw new ErrorWithStatusCode('Invalid or expired promise token', STATUS_UNAUTHORIZED);
+    }
+  }
+  return fallback(context);
+}
+
 /**
  * Parses and retrieves a specific cookie value by name from the request context.
  * Uses indexOf-based splitting to correctly handle values containing '='
```
MysticatBot, line 829: issue (blocking): No dedicated unit tests for `resolveSemrushImsToken`. This function is the single entry point for all Semrush auth resolution, but it has zero direct tests in `test/support/utils.test.js`. The controller tests use a hand-written re-implementation stub (`makeResolveSemrushImsTokenStub`) that mirrors the decode/exchange/fallback logic rather than exercising the real function. If someone introduces a bug here (changes `hasText` behavior, breaks the catch semantics, removes optional chaining on `log`), no test catches it - the controller stubs mask the real code.

Suggested fix: add a `describe("resolveSemrushImsToken", ...)` block in `test/support/utils.test.js` covering: (a) promise-token present + exchange succeeds, (b) `decodeURIComponent` throws (fallback to raw value), (c) exchange throws (401 + logged), (d) promise-token absent (fallback called), (e) `log` is null (no crash on `log?.error`), (f) custom fallback function used when provided. This would let the controller tests use the real function with only `exchangePromiseToken` stubbed at the dependency level, eliminating the drift-prone re-implementation stubs.

src/controllers/brands.js
```diff
@@ -32,7 +32,7 @@ import {
   isValidUUID,
 } from '@adobe/spacecat-shared-utils';
 
-import { ErrorWithStatusCode, getImsUserToken, getImsUserTokenStrict } from '../support/utils.js';
+import { ErrorWithStatusCode, getImsUserToken, resolveSemrushImsToken } from '../support/utils.js';
```
MysticatBot, line 35: issue (blocking): The promise-token path through `brands.js` and `brand-provisioning.js` has no test coverage. These 4 call sites were migrated from `getImsUserTokenStrict` to `resolveSemrushImsToken` but their test suites never send an `x-promise-token` header. Existing tests pass only because the fallback path (no promise token) still works - the NEW path is untested.

This was flagged by a prior reviewer (@calvarezg) and has not been addressed.

Suggested fix: add at least one test in `brands.test.js` (e.g., the PATCH brand path that re-syncs) sending `x-promise-token` and verifying (a) the exchange is called, (b) the resulting token reaches `createSerenityTransport`. Similarly for `brand-provisioning.test.js`, one test for `provisionBrandSubworkspace` with a promise token.

MysticatBot (changes requested):
Hey @vivesing,

**Verdict: Request changes** - two test-coverage gaps need addressing before merge.
**Complexity: HIGH** - medium diff; API-surface risk signal.
_Changes: Adds `x-promise-token`-based authentication to all Semrush proxy endpoints (serenity, elements, brands, brand-provisioning) via a shared `resolveSemrushImsToken` helper, alongside the existing IMS bearer flow (8 files)._

Note: CI checks are currently passing.

## Must fix before merge

1. [Important] No dedicated unit tests for the shared `resolveSemrushImsToken` utility - `src/support/utils.js:782` (details inline)
2. [Important] No test coverage for brands.js / brand-provisioning.js promise-token paths - `src/controllers/brands.js:1869` (details inline)

<details>

- nit: `resolveSemrushImsToken` is called twice in the same request in the PATCH brand handler (lines ~1869 and ~1925 of brands.js) when both conditions are satisfied - causes two redundant IMS round-trips; resolve once and thread the token down - `src/controllers/brands.js:1869`
- suggestion: `getImsUserTokenStrict` error message now says "send the x-promise-token header instead" which assumes a Semrush-proxy context; if this function is ever reused elsewhere the message will mislead - consider moving the hint into the callers or the shared `resolveSemrushImsToken` only - `src/support/utils.js:711`
- nit: The `decodeURIComponent` catch block silently swallows all exception types (not just URIError) - consider a comment noting this is intentional or narrowing to `catch (e) { if (!(e instanceof URIError)) throw e; }` - `src/support/utils.js:800`
- suggestion: All 7 promise-token tests in serenity.test.js exercise only `listPrompts`; adding one test on a mutating handler (e.g. `createPrompts`) would catch handler-specific error-path interactions - `test/controllers/serenity.test.js:379`
- suggestion: Consider updating the OpenAPI specs (`docs/openapi/`) for the new `x-promise-token` optional header per the repo's "Specification Sync" convention, or document that this is deliberately deferred given the temporary nature of this auth approach - `src/controllers/serenity.js:280`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 59s | <b>Cost:</b> $8.24 | <b>Commit:</b> <code>0458a3191f88a164069bfa75ad215ad0f151d9d8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @vivesing,

**Verdict: Approve** - prior findings addressed, no new blocking issues.
**Complexity: HIGH** - large diff; API-surface risk signal.
_Changes: Adds `x-promise-token`-based authentication to all Semrush proxy endpoints via a shared `resolveSemrushImsToken` helper, alongside the existing IMS bearer flow (11 files)._

- Unit tests for `resolveSemrushImsToken` utility now in `test/support/utils.test.js` (6 tests covering exchange, decode fallback, error handling, null-log safety, custom/default fallback).
- Promise-token path coverage for brands.js and brand-provisioning.js added in `test/controllers/brands.test.js` and `test/support/serenity/brand-provisioning.test.js`.

<details>

- nit: `makeResolveSemrushImsTokenStub` is duplicated identically in `elements.test.js` and `serenity.test.js` (28 lines each); consider extracting to a shared `test/helpers/promise-token-stub.js` when the decode/exchange logic next changes - `test/controllers/elements.test.js:48`
- suggestion: the `resolveSemrushImsToken` unit suite in `utils.test.js` has no test where `decodeURIComponent` actually transforms the input (e.g. `'promise%20token'` -> `'promise token'`); the controller suites cover this via the stub re-implementation, but the authoritative unit file does not - `test/support/utils.test.js:1597`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 31s | <b>Cost:</b> $6.29 | <b>Commit:</b> <code>8548cc1079e5303ba6a46b3c394649846e3b5f4d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2756, opened 2026-07-03
fix: stop fabricating domain & metrics in top-brands | LLMO-6012

calvarezg:
## Summary

`BrandService/TopBrandsByDomain` returns only `{ brand_name, count }` — no URL, no visibility, no cited-page metrics. `handleBrandTopBrands` fabricated all three: a `.com` host via `slugHostFromBrandName`, a `visibility` percentage as a log curve of `mentions`, and `citedPages` as `mentions / 200`. For careinsurance.com this surfaced dead domains (e.g. carehealthinsurance.com) as ranked "competitors" and an invented 87% visibility. This removes the fabricated fields — the top-brands row now returns only the real `name` and `mentions` (plus the request's `country` echo).

Note: `slugHostFromBrandName` is intentionally kept — it still has a live call site in `handleBrandSourceOpportunities`, where the value feeds a gRPC request arg and is never rendered. That path is being handled separately (competitor-vs-alias question with Semrush).

## API specification

No spec change: the top-brands response uses the generic `AiVisibilityPaginatedResponse` schema (`data` is `additionalProperties: true`), so dropping row fields needs no schema or example update.

## Related Issues

https://jira.corp.adobe.com/browse/LLMO-6012

## Test plan

- [x] `npx mocha test/support/ai-visibility/handlers/brands.test.js test/support/ai-visibility/grpc-utils.test.js` — 369 passing; no top-brands test asserted the removed fields.
- [ ] Deploy this API before the paired project-elmo-ui PR for LLMO-6012. Ordering is non-breaking (the UI already ignores the removed fields).

Paired with the project-elmo-ui PR for LLMO-6012. Per request, MysticatBot review is triggered but not waited on.


MysticatBot (approved):
Hey @calvarezg,

**Verdict: Approve** - clean, well-scoped removal of fabricated data; no concerns.
**Complexity: LOW** - trivial diff, single handler change.
_Changes: removes three fabricated response fields (domain, visibility, citedPages) from the top-brands handler, returning only real data from the upstream gRPC service (1 file)._

No blocking issues found. The change is minimal and correct: the sort comparator references only `mentions` and `name` (both retained), the conditional `country` echo is preserved, and the existing test suite (369 passing) does not assert on the removed fields. Consumer coordination is documented in the PR body (the UI already ignores these fields).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 24s | <b>Cost:</b> $2.95 | <b>Commit:</b> <code>d87b29b65e8249f5f8a2334e60ad890f35f3792e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

dmitriigulkin-semrush (approved):

PR #2758, opened 2026-07-06
feat: add Slack commands for path-level prerender suggestions

ssilare-adobe:
## Summary

- Adds `path-suggestions {enable/disable} {site}` Slack command to toggle path-level prerender suggestion generation for a site. This sets `pathSuggestionsEnabled` in the site's `handlers.prerender` config, which is read by the audit worker ([spacecat-audit-worker#2500](https://github.com/adobe/spacecat-audit-worker/pull/2500)).
- Adds `get path-suggestions {site}` Slack command to check whether path-level suggestions are currently enabled for a site.
- Both commands support site lookup by base URL or site ID.

## Test plan

- [x] Unit tests for `toggle-path-suggestions` command (11 tests)
- [x] Unit tests for `get-path-suggestions-status` command (9 tests)
- [ ] Manual Slack test: `path-suggestions enable <site>` then verify config via `get path-suggestions <site>`
- [ ] Manual Slack test: `path-suggestions disable <site>` then verify disabled status

### DIFF
--- a/src/support/slack/commands.js
+++ b/src/support/slack/commands.js
@@ -59,6 +59,8 @@ import removeDelegate from './commands/remove-delegate.js';
 import checkAgenticTrafficDbStatus from './commands/check-agentic-traffic-db-status.js';
 import checkCdnLogsStatus from './commands/check-cdn-logs-status.js';
 import addOaeStageDomain from './commands/add-oae-stage-domain.js';
+import togglePathSuggestions from './commands/toggle-path-suggestions.js';
+import getPathSuggestionsStatus from './commands/get-path-suggestions-status.js';
 
 /**
  * Returns all commands.
@@ -116,4 +118,6 @@ export default (context) => [
   checkAgenticTrafficDbStatus(context),
   checkCdnLogsStatus(context),
   addOaeStageDomain(context),
+  togglePathSuggestions(context),
+  getPathSuggestionsStatus(context),
 ];
--- a/src/support/slack/commands/get-path-suggestions-status.js
+++ b/src/support/slack/commands/get-path-suggestions-status.js
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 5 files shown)


MysticatBot (approved):
Hey @ssilare-adobe,

**Verdict: Approve** - clean implementation following established Slack command patterns.
**Complexity: MEDIUM** - new feature with a write handler touching site config.
_Changes: Adds two Slack commands (`path-suggestions enable/disable` and `get path-suggestions`) for toggling and checking path-level prerender suggestion generation per site (5 files)._

<details>

- suggestion: Use consistent config access in both commands - the get command reads via `config.getHandlerConfig('prerender')` while the toggle reads via `config.getHandlers().prerender`. Using `getHandlerConfig` in both would prevent a silent divergence if that method ever adds defaults or transforms. - `src/support/slack/commands/toggle-path-suggestions.js:79`
- suggestion: Consider mutating the config in-place rather than overlaying `configData.handlers` on the serialized output. Since `getHandlers()` returns a live reference to internal state, you can mutate `handlers.prerender.pathSuggestionsEnabled = isEnable` directly, then call `site.setConfig(Config.toDynamoItem(config))` without rebuilding the handlers object. This avoids the risk of overriding transformations that `toDynamoItem` might apply in the future. - `src/support/slack/commands/toggle-path-suggestions.js:84-90`
- nit: Add an info-level log before the save recording which Slack user toggled which site - aids incident investigation when config changes have unexpected downstream effects. - `src/support/slack/commands/toggle-path-suggestions.js:91`
- nit: When the toggle is a no-op (already in the requested state), a "was already enabled/disabled" response would help operators confirm current state without checking separately. - `src/support/slack/commands/toggle-path-suggestions.js:94`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 25s | <b>Cost:</b> $6.94 | <b>Commit:</b> <code>72f002bdfad9724aa23aa5099fae71bfc2f6c3d4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2762, opened 2026-07-06
feat(serenity): add tag: open dimension (plain tags)

byteclimber:
## Summary

Adds a new `tag:` **open dimension** to Semrush "serenity" tags, mirroring the existing `category:` dimension: customer-authored `tag:<name>` roots with one level of bare-named children, taggable onto prompts, filterable, and id-based. Design doc: **serenity-docs PR #26** ("Plain tags: a native, nestable `tag:` dimension").

### What changed
- **`prompt-tags.js`**: `TAG: 'tag'` added to `TAG_DIMENSION`; `TAG` added to `CREATABLE_TAG_DIMENSIONS` (now `[category, topic, tag]`). Closed dimensions (`source/intent/type`) untouched.
- **`handlers/tags.js`** — parent validation:
  - New `assertValidParent(parentId, childDimension)` — resolves the parent via the tree read and rejects a **non-root** parent or a **cross-dimension** parent. **Called from BOTH create-with-parent AND both branches of the update/re-parent path.**
  - Update/re-parent also rejects **self-parenting** (`parentId === tagId`) and **demoting a populated root** (a root with `childrenCount > 0` given a non-null parent — would create an illegal depth-2 tree).
  - **Root rename can no longer silently cross dimensions** (`category:X` → `tag:Y` is a 400); `resolveTagTarget` now returns the target's name/dimension/childrenCount to support this.
  - **Promote-to-root re-prefixing**: a child promoted to root (`parentId: null`) has its bare name re-prefixed with its inherited dimension (e.g. `High` under a `tag:` root → `tag:High`), covered on **both** the flat and subworkspace paths.
- **New `DELETE /serenity/tags/:tagId`** — resolves the target, returns **204** for a `tag`-dimension leaf/childless-root; **409 `tagHasChildren`** (new error code) for a populated root, enforced *before* any upstream call; **400 `categoryDeleteNotYetSupported`** for any non-`tag` dimension. New transport method `deleteProjectTags` (uses the existing shared `aio-delete-tags` op, passing the required-but-unused `prompt_id=''`). Controller `deleteTag` (flat/subworkspace dispatch, `noContent()` on success), route + facs-capabilities + required-capabilities classification.
- **`handlers/markets.js` `listProjectTagTree`**: each item now carries a derived `dimension` — a root from its own `<dimension>:` prefix, a child from its root ancestor at `path[0]` (root-first per schema). **Omitted when undeterminable** (malformed prefix, or a child with `path === null`) — consumers must treat a missing `dimension` as "unknown, don't guess", never as "bare = subcategory".
- **OpenAPI**: `tag` added to the create `type` enum + the 201 response enum; new `DELETE` path with 400/409 responses; `SerenityTag.dimension` field (enum incl. `tag`) in `schemas.yaml`. `docs/index.html` regenerated.

### Retroactive behavior change (please note)
The parented-**create** path previously forwarded `parentId` to the transport with **zero validation** that it referenced a real root of any dimension. This PR adds real root-ness + dimension validation there for **`category:` too**, not just `tag:`. Positive-path regression tests prove existing legitimate category create-with-parent / re-parent flows still succeed; negative-path tests cover the new guards.

## Tests
- Unit: extended `tags.test.js` (tag: create root/child, all four parent-guard rejections on create + update, promote re-prefix on both paths, rename-preserves-dimension, delete 204/409/400 incl. "never calls the delete transport when childrenCount > 0"), `prompt-tags.test.js` (`CREATABLE_TAG_DIMENSIONS` content + disjointness), `markets.test.js` / `markets-subworkspace.test.js` (dimension field incl. `path === null` and unknown-prefix fallbacks), `rest-transport.test.js` (`deleteProjectTags`), controller + openapi-contract fixtures.
- `npm run lint`, `npm run type-check` (blocking gate), and the full `npm test` suite (**14115 passing**) are green.
- **IT** (`test/it/shared/tests/serenity.js`): added tag: create + tree-dimension + DELETE (204/409/400) coverage matching existing conventions. **Could not run locally — Docker is unavailable in this environment.** These IT paths exercise the shared Project-Engine mock's `aio-delete-tags` handler; if the mock version pinned here predates it, they depend on the matching **spacecat-shared mock update (WP2)**.

## Known / pending (transparency, not blockers)
- **G1 live-probe not yet run** (serenity-docs#26 §9): whether Semrush keys child tag ids by `(parent_id, name)` or by name alone. The parent/promote guard logic is structurally sound regardless, but the overall parent/child model's soundness depends on confirming this. The promote-to-root re-prefixing is likewise contingent on the G1/G3 probe.
- **Prod data audit not yet run** for pre-existing `category:` tags that might already violate the new parent-must-be-root invariant.
- **Category delete is deliberately out of scope** for this PR (tag-dimension only), pending a separate decision on the interaction with DRS's category re-sync/soft-delete system (a separate PostgREST `categories` table with its own delete/resurrection semantics).
- **Mixed-dimension OR-filter**: the prompt filter is purely id-based / dimension-agnostic in this repo, so there's no new code path here to cover; the meaningful mixed-dimension filter test belongs in spacecat-shared's e2e suite (WP2).


src/support/serenity/handlers/tags.js
```diff
@@ -724,3 +887,131 @@ export async function handleUpdateTagSubworkspace(
...
+ * @param {string} tagId - upstream tag id to delete.
+ * @param {object} [log] - logger.
+ * @returns {Promise<{ status: number }>}
+ */
+async function deleteResolvedTag(transport, semrushWorkspaceId, projectId, tagId, log) {
+  const target = await resolveTagTarget(transport, semrushWorkspaceId, projectId, tagId, log);
+  if (target.dimension !== TAG_DIMENSION.TAG) {
+    const err = new ErrorWithStatusCode(
+      'Only tag-dimension tags can be deleted; category delete is not yet supported',
+      400,
```
MysticatBot, line 884: issue (blocking): `deleteResolvedTag` conflates "tag not found" with "wrong dimension". When `resolveTagTarget` returns `kind: 'unknown'` (tag beyond `MAX_ROOTS_TO_SEARCH` or already deleted), `target.dimension` is `undefined`. The check `target.dimension !== TAG_DIMENSION.TAG` evaluates `undefined !== 'tag'` as `true` and throws `categoryDeleteNotYetSupported` (400) - a factually wrong error for a tag that simply could not be located.

**Why it matters:** A user deleting a valid `tag:`-dimension tag that happens to be beyond the search cap gets told "category delete is not yet supported" - confusing and unactionable.

**Fix:** Add an explicit guard before the dimension check:
```js
if (target.kind === 'unknown') {
  throw new ErrorWithStatusCode('Tag not found in the project tag tree', 404);
}
```
This also enables a distinct error code (e.g. `tagNotResolved`) for observability.

MysticatBot (changes requested):
Hey @byteclimber,

**Verdict: Request changes** - one correctness bug in the delete path and stale operator docs need fixing before merge.
**Complexity: HIGH** - large diff (21 files); API surface + new endpoint + retroactive validation.
_Changes: Adds a `tag:` open dimension to Serenity tags with a DELETE endpoint, parent validation guards on create/update, dimension field derivation on tree reads, and comprehensive tests (21 files)._

## Must fix before merge

1. [Important] `deleteResolvedTag` returns misleading 400 `categoryDeleteNotYetSupported` when the tag is unresolvable (kind === 'unknown') - `src/support/serenity/handlers/tags.js:876` (details inline)
2. [Important] `docs/serenity.md` is stale: lists only `category`/`topic` as open dimensions and omits the new DELETE endpoint - `docs/serenity.md` (details inline)

<details>

- nit: OpenAPI `deleteSerenityTag` omits a `401` response entry, inconsistent with sibling operations - `docs/openapi/serenity-api.yaml`
- nit: OpenAPI 204 description says "or already absent upstream" but the code propagates a transport 404 as 502, not 204 - `docs/openapi/serenity-api.yaml`
- suggestion: Double tree-walk on update-with-reparent (resolveTagTarget called twice, re-fetching the root listing). Consider passing the already-fetched roots to `assertValidParent` - `src/support/serenity/handlers/tags.js`
- suggestion: `deleteProjectTags` passes `prompt_id: ''` relying on upstream ignoring a required parameter. Track this as a known fragility - `src/support/serenity/rest-transport.js`
- suggestion: Add a test for the promote-to-root fallback when `target.dimension` is undefined (the bare-name passthrough) - `test/support/serenity/handlers/tags.test.js`
- suggestion: Cover the 2 Codecov-flagged lines in `markets.js` (empty `path: []` array branch or unrecognized child prefix) - `src/support/serenity/handlers/markets.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 29s | <b>Cost:</b> $7.78 | <b>Commit:</b> <code>fccaa5035472ad1c0086d5134822288353a19031</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

byteclimber:
Thanks @MysticatBot — addressed in e49820a6:

**Must-fix**
1. ✅ `deleteResolvedTag` now returns **404 `tagNotResolved`** for an unresolvable id (`kind === 'unknown'`), checked *before* the dimension guard — so `categoryDeleteNotYetSupported` (400) only fires for a genuinely resolved non-`tag` dimension. Added a unit test for the 404 path.
2. ✅ `docs/serenity.md` updated: `tag` added as an open dimension, create/re-parent validation documented, and the new `DELETE` row added.

**Non-blocking**
- ✅ OpenAPI `deleteSerenityTag`: added `401`; removed the inaccurate "already absent upstream" from `204` (a transport 404 surfaces as 502) and documented `tagNotResolved` under `404`.
- ✅ Added a test for the promote-to-root **bare-name fallback** when `target.dimension` is undeterminable.
- ✅ Added `markets.js` coverage for the empty `path: []` and unrecognized-root-prefix branches (should clear the two codecov-flagged lines).
- ⏭️ **Double tree-walk on update-with-reparent** (`resolveTagTarget` called twice): acknowledged, deferring. It's an admin-frequency path and threading the pre-fetched roots into `assertValidParent` adds coupling for a marginal win; I'd rather keep the guard self-contained. Happy to do it if you feel strongly.
- ⏭️ **`prompt_id: ''` fragility**: acknowledged as known — the transport method carries a comment noting the param is required-but-unused upstream (confirmed against the shared PE mock). Tracking as a known dependency on the upstream contract, not changing behavior here.

Re-requesting review.

docs/serenity.md
```diff
@@ -144,8 +144,9 @@ All endpoints require `Authorization: Bearer <ims_user_token>` and `organization
...
+| DELETE | `/serenity/tags/:tagId?geoTargetId=&languageCode=` | Delete a single tag by its upstream id. **`tag`-dimension ONLY for now** — deleting a `category`/other tag is a deliberate `400 categoryDeleteNotYetSupported` (pending the DRS category re-sync interaction). A populated root is `409 tagHasChildren` (enforced before any upstream call); an unresolvable id is `404 tagNotResolved`. | `deleteSerenityTag` |
 | GET | `/serenity/models?geoTargetId=&languageCode=` | AI models for one slice (catalog mode when no params) | `listSerenityModels` |
 | PUT | `/serenity/models` | Replace the AI-model set for one slice (publishes after change) | `updateSerenityModels` |
 | POST | `/serenity/activate` | Activate the brand into sub-workspace mode (ensure sub-workspace + publish supplied markets) | `activateSerenityBrand` |
```
MysticatBot: issue (blocking): The PATCH row documents `parentId: null` as "PROMOTES a child to root" but does not mention that the API automatically re-prefixes the bare child name with its inherited dimension during this promotion (e.g. a child named `Sneakers` under a `category:` root becomes `category:Sneakers` in the response).

This is a non-obvious transformation - the client sends a bare `name` value but receives a prefixed name back. Without this documented, API consumers may assume the name is preserved as-sent and get confused by the response.

**Fix:** Append to the PATCH description: "When promoting a child to root (`parentId: null`), the API re-prefixes the bare child name with its inherited dimension; clients send the bare value and receive the prefixed root name. Falls back to the bare name when the dimension cannot be determined."

MysticatBot (changes requested):
Hey @byteclimber,

**Verdict: Request changes** - one doc gap in the PATCH endpoint description needs a line about the promote-to-root re-prefixing behavior.
**Complexity: HIGH** - large diff (21 files); API surface + new endpoint + retroactive validation.
_Changes: Adds a `tag:` open dimension to Serenity tags with a DELETE endpoint, parent validation guards on create/update, dimension field derivation on tree reads, and comprehensive tests (21 files)._

## Must fix before merge

1. [Important] `docs/serenity.md` PATCH row omits the promote-to-root re-prefixing behavior - when `parentId: null` promotes a child, the API automatically re-prefixes the bare name with its inherited dimension (e.g. `Sneakers` becomes `category:Sneakers`). API consumers need to know the name transforms on promotion. - `docs/serenity.md:146` (details inline)

<details>

- suggestion: `resolveTagTarget` returns `childrenCount: 0` for the `unknown` case, which is "no data" not "verified zero". A JSDoc note that childrenCount is meaningless when `kind === 'unknown'` would prevent future misuse - `src/support/serenity/handlers/tags.js`
- suggestion: `dimensionFromTagName` lowercases the prefix defensively but the JSDoc does not mention case-insensitivity. A one-line note would prevent someone removing the `.toLowerCase()` thinking it is dead code - `src/support/serenity/handlers/markets.js`
- nit: The new bare-name fallback test asserts status and transport call args but not `res.body.tag`, unlike the companion promote test that asserts the full response shape - `test/support/serenity/handlers/tags.test.js`
- nit: The `.code` assignment pattern (`err.code = ERROR_CODES.X` after construction) works today but is fragile if `ErrorWithStatusCode` ever freezes instances. Consider an options bag if the class supports it - `src/support/serenity/handlers/tags.js`
</details>

- `deleteResolvedTag` now returns 404 `tagNotResolved` for unresolvable tags (was: misleading 400)
- `docs/serenity.md` updated with `tag` dimension, validation docs, and DELETE row
- OpenAPI `deleteSerenityTag` now includes 401 response
- OpenAPI 204 description no longer claims "already absent upstream"
- Promote-to-root bare-name fallback test added
- Markets.js coverage for empty `path[]` and unrecognized-root-prefix branches added

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 5m 1s | <b>Cost:</b> $8.26 | <b>Commit:</b> <code>e49820a6d5f142df5ec107291fef6d0d77579580</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

byteclimber:
Thanks @MysticatBot — addressed in 8b833ee5:

**Must-fix**
- ✅ `docs/serenity.md` PATCH row now documents promote-to-root **re-prefixing** (bare child name gains its inherited dimension, e.g. `Sneakers` → `category:Sneakers`, so the returned `tag` differs from the `name` sent).

**Non-blocking**
- ✅ `resolveTagTarget` JSDoc: `childrenCount` is only meaningful for `kind: 'root'`; a `'unknown'` result's `0` is "no data", not a verified zero.
- ✅ `dimensionFromTagName` JSDoc: prefix match is case-insensitive; the `.toLowerCase()` is load-bearing.
- ✅ Bare-name fallback test now also asserts `res.body.tag`.
- ⏭️ `.code = ERROR_CODES.X` after construction: keeping as-is — it's the established pattern across the entire serenity surface (`marketNotFound`, `ambiguousWorkspace`, etc.) and `ErrorWithStatusCode` doesn't take a code in its constructor; switching only these sites would be inconsistent. Would fold it in as a repo-wide follow-up if desired.

Re-requesting review.

MysticatBot (approved):
Hey @byteclimber,

**Complexity: HIGH** - large diff (21 files); API surface + new endpoint + retroactive validation.
_Changes: Adds a `tag:` open dimension to Serenity tags with a DELETE endpoint, parent validation guards on create/update, dimension field derivation on tree reads, and comprehensive tests (21 files)._

- `docs/serenity.md` PATCH row now documents promote-to-root re-prefixing behavior
- `resolveTagTarget` JSDoc clarifies `childrenCount` semantics on `kind: 'unknown'`
- `dimensionFromTagName` JSDoc notes case-insensitive prefix match is load-bearing
- Bare-name fallback test now asserts `res.body` shape (`{ tag: 'Leaf', parentId: null }`)
- `.code = ERROR_CODES.X` post-construction pattern: author pushback accepted (consistency with existing serenity surface)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 33s | <b>Cost:</b> $3.02 | <b>Commit:</b> <code>8b833ee574f5efb754844e91ead6082c6b548250</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Parking this one — it is outdated on two levels:

1. **Needs a rebase** — `main` has moved since this branched.
2. **The tag model changed underneath it.** With the dimension-root tag model (https://github.com/adobe/serenity-docs/pull/39), tag names no longer carry `<dimension>:` prefixes: dimensions are root tags with bare-named descendants, and a tag's dimension comes from `path[0]`, never from parsing its name. This PR builds the `tag` dimension on the prefixed representation throughout — `tag:<name>` roots, prefix-derived `dimension` in `listProjectTagTree`, promote-to-root re-prefixing — so the dimension handling needs rework to the root + bare-children shape before it can land.

This is not priority zero, so let's pick it up after the end-of-July GA.

PR #2766, opened 2026-07-06
fix(serenity): enable bounded-backoff retry on the Semrush REST transport

aliciadriani:
## Spec

**The gap.** `src/support/serenity/rest-transport.js` (the Semrush "Serenity" transport) had a real reliability gap that the sibling shared client already closed. Confirmed via a whole-tree + fresh-`origin/main` search before writing any code.

The transport is built on two typed, vendored Semrush clients — `@adobe/spacecat-shared-project-engine-client` and `@adobe/spacecat-shared-user-manager-client` — both of which ship an idempotency-aware, jittered-backoff retry layer (`createRetryingFetch` in each package's `internal.js`, defaults `maxRetries: 2`, `retryBaseDelayMs: 200`). That retry logic landed in the shared packages well before the versions api-service currently pins (1.8.0 / 1.3.1 respectively), so **no dependency bump is needed here.**

The retry was explicitly disabled in api-service: when `rest-transport.js` migrated onto these typed clients, both were constructed with `maxRetries: 0`, with a comment stating this "preserves the one-shot behaviour the hand-rolled transport had." That was a deliberate deferral at migration time, not a bug — but it meant api-service's Semrush calls had zero retry coverage for transient 429/5xx/network failures, unlike the shared client. This PR closes exactly that gap: flip the pin off, nothing else.

(Note: an earlier framing of this task assumed the transport was still fully hand-rolled with *no* retry-capable dependency underneath it. That premise was stale — by the time this branch was cut from `origin/main`, the transport had already migrated to the typed clients described above, as part of unrelated, already-shipped work. This PR does not touch or extend that migration; it only flips the one config value the migration left off.)

## Implementation plan

**What changed** — `src/support/serenity/rest-transport.js`: removed `maxRetries: 0` from both `createSerenityProjectEngineApiClient` and `createSerenityUserManagerApiClient` constructions, letting the library defaults apply (2 retries, 200ms base backoff, jittered exponential growth capped at 20s/attempt, `Retry-After` honored). Added doc comments explaining the retry/idempotency/timeout behavior at the call site. No other files changed.

**Guardrail 1 — per-attempt timeout, not whole-loop.** Verified by reading the composition order rather than assuming it: the transport passes `createTimeoutFetch(DEFAULT_TIMEOUT_MS)` (its own pre-existing 15s `AbortController` wrapper) as the `fetch` option into each typed client. Internally each client does `fetch: createRetryingFetch(injectedFetch, maxRetries, retryBaseDelayMs, onRetry)` — the retry layer wraps the timeout-fetch, not the other way around. The retry loop calls the timeout-wrapped fetch once per attempt, so every retry gets a brand-new `AbortController` + `setTimeout(15_000)`. A timeout surfaces as a thrown error inside the retry loop and is gated by the same idempotent-method rule as any other network error (retried for GET/HEAD/PUT/DELETE/OPTIONS, rethrown immediately for POST). This was architecturally already correct — no code change was needed for this guardrail, only verification, which the new "per-attempt timeout" tests lock in.

**Guardrail 2 — idempotency correctness.** Checked every caller of the transport:
- `src/controllers/serenity.js` and `src/controllers/brands.js` call `createSerenityTransport(...)` and drive it through `src/support/serenity/handlers/*.js` and `src/support/serenity/workspace-lifecycle.js`.
- `src/support/serenity/brand-provisioning.js` drives `handleCreateMarketSubworkspace` (`handlers/markets-subworkspace.js`), which calls `ensureSubworkspace` (`workspace-lifecycle.js`) — the actual workspace/resource **create** path: `transport.createSubworkspace(parentWorkspaceId, title, CREATE_ALLOCATION)` (POST `/v2/workspaces/{parent}/child`) and `transport.createProject(...)` (POST).
- Cross-referenced every transport method against its HTTP verb directly in `rest-transport.js`. All create/mutate-only calls (`createProject`, `createSubworkspace`, `createBenchmarks`, `createBrandUrls`, `createTaggedPrompts`, `createPromptsByIds`, `transferWorkspaceResources`, `addAiModel`, `createProjectTags`, `publishProject`, `listPromptsByTags`) are POST. Reads are GET. Updates are PUT/PATCH; deletes are DELETE.

  The idempotency gate lives inside the shared retry layer itself (429 retried for any method; 5xx retried only for GET/HEAD/PUT/DELETE/OPTIONS), so gating is uniform across every call site automatically. Confirmed POST creates never retry on 5xx with a direct test against the real `brand-provisioning.js` call shape.

**Test coverage added** (`test/support/serenity/rest-transport.test.js`, new `describe('retry behaviour', ...)` block, plus fixes to 4 pre-existing tests whose GET/timeout assertions now interact with retry):
- 429 retried on a POST create (`createSubworkspace`, the real `brand-provisioning.js` → `workspace-lifecycle.js#ensureSubworkspace` call shape) — retries and succeeds.
- 500 on that same POST create is **NOT** retried — single attempt, immediate 500 (the core anti-double-provisioning guarantee). Same check repeated for `createProject`.
- 429 retried on `createProject` (POST) — succeeds.
- 503 retried on a GET (`getWorkspaceStatus`) — succeeds after retry.
- `Retry-After` (delta-seconds) honored as a floor on the backoff wait (real-clock assertion, to prove it's actually driving the wait).
- Exhaustion returns the last retryable HTTP response rather than throwing (GET, 3 attempts).
- Exhaustion rethrows the last network error verbatim once retries run out (GET, 3 attempts).
- Per-attempt timeout: attempt 1 times out, attempt 2 gets its own fresh budget and succeeds — proves the timeout isn't shared across the whole retry loop.
- A per-attempt timeout on POST is NOT retried (single attempt, immediate 504).
- Non-retryable 4xx (404) never retries regardless of method.
- Fixed 3 pre-existing GET-based tests (503, non-JSON-500, ENOTFOUND on `getWorkspaceStatus`) and 1 timeout test to account for retries now happening on idempotent methods — added fake timers where needed to keep them fast/deterministic, plus explicit `fetchStub.callCount` assertions.

## Verification

- `npm run lint` — clean (0 errors).
- `npx mocha test/support/serenity/rest-transport.test.js` — **93 passing, 0 failing**.
- `npm test` (full repo suite) — **14124 passing, 0 failing**; coverage 99.94% lines / 99.23% branches / 99.12% functions (gate: 90%/90%/90%).
- No dependency version bump required — the retry logic already ships in the currently-pinned `@adobe/spacecat-shared-project-engine-client@1.8.0` / `@adobe/spacecat-shared-user-manager-client@1.3.1`.

## Scope

Retry only. Did not touch typed-error work (LLMO-5978) or a timeout redesign (LLMO-5979) — both shared-client-scope tickets, out of scope here. No new config surface, no `onRetry` observability hook added (would need threading a `log` into `createSerenityTransport`, which none of the current call sites pass — left out to keep this the smallest shippable fix).

### DIFF
--- a/src/support/serenity/rest-transport.js
+++ b/src/support/serenity/rest-transport.js
@@ -255,26 +255,53 @@ export function createSerenityTransport({ env, imsToken }) {
     return imsToken;
   };
 
-  // Typed Project Engine client over the project gateway. maxRetries:0 preserves
-  // the one-shot behaviour the hand-rolled transport had; the injected fetch
-  // re-adds the 15s timeout + Accept header the client does not impose. `root` is
-  // the validated origin; the client appends its own '/enterprise/projects/api'
+  // Typed Project Engine client over the project gateway. `root` is the
+  // validated origin; the client appends its own '/enterprise/projects/api'
   // prefix.
+  //
+  // Retries: the client's own bounded-backoff retry (createRetryingFetch in
+  // spacecat-shared-project-engine-client's internal.js) is left at its library
+  // defaults (maxRetries: 2, retryBaseDelayMs: 200 — 3 attempts total, jittered
+  // exponential backoff up to a 20s/attempt ceiling, honouring Retry-After).
+  // It was explicitly pinned to maxRetries: 0 when this transport migrated onto
+  // the typed clients, purely to preserve the hand-rolled transport's one-shot
+  // behaviour at the time — api-service had no retry logic at all, unlike the
+  // sibling shared client. That gap is closed here: idempotent methods (GET,
+  // HEAD, PUT, DELETE, OPTIONS) retry on 429 and on retryable 5xx; POST (and any
+  // other non-idempotent method) retries ONLY on 429, never on 5xx, so a create
+  // (createProject, createSubworkspace, createBenchmarks, createBrandUrls,
+  // createTaggedPrompts, transferWorkspaceResources, addAiModel,
+  // createProjectTags, publishProject, listPromptsByTags — all POST) can never
+  // be double-submitted by the retry layer on a 5xx; a 429 is safe to retry for
+  // any method because the Semrush gateway rejects it at the edge before the
+  // handler runs, so the create never happened.
+  //
+  // Per-attempt timeout: the injected `fetch` is `createTimeoutFetch`, which the
+  // retry layer calls once per attempt (it wraps `injectedFetch`, not the other
+  // way around) — so every retry gets its OWN fresh 15s AbortController/timer,
+  // not one 15s budget shared across the whole retry+backoff sequence. A timeout
+  // is surfaced to the retry layer as a thrown error (SerenityTransportError
+  // status 504), gated by the same idempotent-method rule as any other network
+  // error.
   const projects = createSerenityProjectEngineApiClient({
     baseUrl: root,
     authToken,
-    maxRetries: 0,
     fetch: createTimeoutFetch(DEFAULT_TIMEOUT_MS),
   });
 
   // Typed User Manager client over the sub-workspace lifecycle gateway. Same
-  // shape as the project client; appends its own '/enterprise/users/api' prefix.
-  // Uses `usersRoot` (SEMRUSH_USERS_BASE_URL, or `root` by fallback) so the
-  // lifecycle gateway can be a separate host from Project Engine.
+  // shape and retry/timeout rationale as the project client above; appends its
+  // own '/enterprise/users/api' prefix. Uses `usersRoot` (SEMRUSH_USERS_BASE_URL,
+  // or `root` by fallback) so the lifecycle gateway can be a separate host from
+  // Project Engine. Its own POST creates (createSubworkspace,
+  // transferWorkspaceResources) are covered by the same POST-never-retries-on-5xx
+  // rule — createSubworkspace in particular must never be silently double-fired
+  // by the retry layer (see workspace-lifecycle.js's own 504-triggered adoption
+  // recovery, which is a SEPARATE, application-level mechanism for the transport
+  // timeout case and unaffected by this change).
   const users = createSerenityUserManagerApiClient({
     baseUrl: usersRoot,
     authToken,
-    maxRetries: 0,
     fetch: createTimeoutFetch(DEFAULT_TIMEOUT_MS),
   });
 
--- a/src/support/serenity/rest-transport.js
+++ b/src/support/serenity/rest-transport.js
@@ -255,50 +255,23 @@ export function createSerenityTransport({ env, imsToken }) {
     return imsToken;
   };
 
-  // Typed Project Engine client over the project gateway. `root` is the
-  // validated origin; the client appends its own '/enterprise/projects/api'
-  // prefix.
-  //
-  // Retries: the client's own bounded-backoff retry (createRetryingFetch in
-  // spacecat-shared-project-engine-client's internal.js) is left at its library
-  // defaults (maxRetries: 2, retryBaseDelayMs: 200 — 3 attempts total, jittered
-  // exponential backoff up to a 20s/attempt ceiling, honouring Retry-After).
-  // It was explicitly pinned to maxRetries: 0 when this transport migrated onto
-  // the typed clients, purely to preserve the hand-rolled transport's one-shot
-  // behaviour at the time — api-service had no retry logic at all, unlike the
-  // sibling shared client. That gap is closed here: idempotent methods (GET,
-  // HEAD, PUT, DELETE, OPTIONS) retry on 429 and on retryable 5xx; POST (and any
-  // other non-idempotent method) retries ONLY on 429, never on 5xx, so a create
-  // (createProject, createSubworkspace, createBenchmarks, createBrandUrls,
-  // createTaggedPrompts, transferWorkspaceResources, addAiModel,
-  // createProjectTags, publishProject, listPromptsByTags — all POST) can never
-  // be double-submitted by the retry layer on a 5xx; a 429 is safe to retry for
-  // any method because the Semrush gateway rejects it at the edge before the
-  // handler runs, so the create never happened.
-  //
-  // Per-attempt timeout: the injected `fetch` is `createTimeoutFetch`, which the
-  // retry layer calls once per attempt (it wraps `injectedFetch`, not the other
-  // way around) — so every retry gets its OWN fresh 15s AbortController/timer,
-  // not one 15s budget shared across the whole retry+backoff sequence. A timeout
-  // is surfaced to the retry layer as a thrown error (SerenityTransportError
-  // status 504), gated by the same idempotent-method rule as any other network
-  // error.
+  // Typed Project Engine client; appends '/enterprise/projects/api'. Retry,
+  // backoff, and the POST-never-retries-on-5xx idempotency gate are the shared
+  // library's contract, not this file's — see createRetryingFetch in
+  // spacecat-shared-project-engine-client's internal.js (library defaults:
+  // maxRetries 2, retryBaseDelayMs 200; per-attempt timeout via the injected
+  // `createTimeoutFetch` below, since the retry layer wraps it and calls it
+  // once per attempt).
   const projects = createSerenityProjectEngineApiClient({
     baseUrl: root,
     authToken,
     fetch: createTimeoutFetch(DEFAULT_TIMEOUT_MS),
   });
 
-  // Typed User Manager client over the sub-workspace lifecycle gateway. Same
-  // shape and retry/timeout rationale as the project client above; appends its
-  // own '/enterprise/users/api' prefix. Uses `usersRoot` (SEMRUSH_USERS_BASE_URL,
-  // or `root` by fallback) so the lifecycle gateway can be a separate host from
-  // Project Engine. Its own POST creates (createSubworkspace,
-  // transferWorkspaceResources) are covered by the same POST-never-retries-on-5xx
-  // rule — createSubworkspace in particular must never be silently double-fired
-  // by the retry layer (see workspace-lifecycle.js's own 504-triggered adoption
-  // recovery, which is a SEPARATE, application-level mechanism for the transport
-  // timeout case and unaffected by this change).
+  // Typed User Manager client over the sub-workspace lifecycle gateway (same
+  // retry/timeout contract as above); appends '/enterprise/users/api'. Uses
+  // `usersRoot` (SEMRUSH_USERS_BASE_URL, falling back to `root`) since this
+  // gateway can be a separate host from Project Engine.
   const users = createSerenityUserManagerApiClient({
     baseUrl: usersRoot,
     authToken,
--- a/test/support/serenity/rest-transport.test.js
+++ b/test/support/serenity/rest-transport.test.js
@@ -368,6 +368,23 @@ describe('Semrush REST transport', () => {
       expect(fetchStub.callCount).to.equal(2);
     }));
 
+    it('retries a retryable DELETE (503) and returns the eventual success', withFakeTimers(async (clock) => {
+      // GET is the only idempotent method exercised for retryable 5xx above.
+      // The idempotency gate treats GET/HEAD/PUT/DELETE/OPTIONS alike, so this
+      // locks in that DELETE gets the same guarantee, using deleteProject (a
+      // real transport method) rather than a synthetic call.
+      fetchStub.onCall(0).resolves(fetchFail(503, { code: 'unavailable' }));
+      fetchStub.onCall(1).resolves(fetchOk(null));
+      const transport = createSerenityTransport({ env: TEST_ENV, imsToken: IMS });
+
+      const promise = transport.deleteProject(WORKSPACE_ID, PROJECT_ID);
+      await clock.tickAsync(60_000);
+      await promise;
+
+      expect(fetchStub.callCount).to.equal(2);
+      expect((await callOf(fetchStub, 1)).method).to.equal('DELETE');
+    }));
+
     it('retries a 429 even on a POST create (createSubworkspace) and returns the eventual success', withFakeTimers(async (clock) => {
       // createSubworkspace is the real brand-provisioning create-call shape
       // (brand-provisioning.js -> workspace-lifecycle.js#ensureSubworkspace ->
@@ -445,10 +462,7 @@ describe('Semrush REST transport', () => {
       expect(fetchStub.callCount).to.equal(2);
     }));
 
-    it('honours Retry-After (delta-seconds) as a floor on the backoff wait', async () => {
-      // No fake timers here: assert real elapsed time is at least the
-      // server-requested delay, proving Retry-After is actually being read
-      // and applied (not just backoff jitter, which defaults to ~100-200ms).
+    it('honours Retry-After (delta-seconds) as a floor on the backoff wait', withFakeTimers(async (clock) => {
       fetchStub.onCall(0).resolves(new Response(JSON.stringify({ code: 'rate_limited' }), {
         status: 429,
         headers: { 'content-type': 'application/json', 'retry-after': '1' },
@@ -456,14 +470,21 @@ describe('Semrush REST transport', () => {
       fetchStub.onCall(1).resolves(fetchOk({ status: 'created' }));
       const transport = createSerenityTransport({ env: TEST_ENV, imsToken: IMS });
 
-      const start = Date.now();
-      const result = await transport.getWorkspaceStatus(WORKSPACE_ID);
-      const elapsed = Date.now() - start;
+      const promise = transport.getWorkspaceStatus(WORKSPACE_ID);
+
+      // Retry-After: 1 is a 1s FLOOR on the wait. Advancing to just under it
+      // must not dispatch the retry yet -- proving the header is what's gating
+      // the delay, not backoff jitter alone (which defaults to ~100-200ms and
+      // would otherwise let this pass for the wrong reason).
+      await clock.tickAsync(900);
+      expect(fetchStub.callCount).to.equal(1);
+
+      await clock.tickAsync(200);
+      const result = await promise;
 
       expect(result.status).to.equal('created');
       expect(fetchStub.callCount).to.equal(2);
-      expect(elapsed).to.be.at.least(950); // ~1s Retry-After, small slack for timer drift
-    }).timeout(5000);
+    }));
 
     it('returns the last retryable response after exhausting all retries (does not swallow into an exception)', withFakeTimers(async (clock) => {
       fetchStub.resolves(fetchFail(503, { code: 'still_unavailable' }));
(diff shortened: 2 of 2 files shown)


MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - well-scoped reliability fix with thorough test coverage of the critical idempotency invariants.
**Complexity: MEDIUM** - medium diff, single-service, config-flip with extensive tests.
_Changes: Enables bounded-backoff retry on the Semrush REST transport by removing `maxRetries: 0` from both typed client constructions, letting library defaults apply (2 files)._

<details>

- nit: The `honours Retry-After` test uses real wall-clock time (`Date.now()` elapsed >= 950ms) making it flaky on loaded CI runners. The 50ms slack is tight. Consider using fake timers and asserting no fetch call is dispatched before `clock.tickAsync(1000)` to prove Retry-After is honored deterministically - `test/support/serenity/rest-transport.test.js:206`
- nit: The inline comment block on the `projects` client construction is ~30 lines for what amounts to removing a single option. The retry/idempotency contract is a property of the shared library. A 3-4 line summary with a pointer to `createRetryingFetch` in the shared client would be more durable against drift. Same rationale partially restated for the `users` client - `src/support/serenity/rest-transport.js:258`
- suggestion: No test for POST + 429 exhaustion (all 3 attempts return 429). The exhaustion tests only use GET + 503. If library behavior differs for POST exhaustion, this gap could miss a regression - `test/support/serenity/rest-transport.test.js:326`
- suggestion: No test for PUT/DELETE retry on 5xx. The idempotency gate says these retry like GET, but only GET is exercised for retryable 5xx. One test on a PUT/DELETE method with 503 would lock in the guarantee - `test/support/serenity/rest-transport.test.js:326`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 58s | <b>Cost:</b> $5.22 | <b>Commit:</b> <code>c4e344d50249bd7686ce05f4028450b514585811</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

aliciadriani:
Addressed all four review nits in 657451b1:

- **Flaky Retry-After test** (`test/support/serenity/rest-transport.test.js:206`) — replaced the real-`Date.now()` assertion with fake timers: ticks to 900ms and asserts no second `fetch` yet (proving the 1s `Retry-After` floor, not jitter, is gating the delay), then ticks past 1000ms and asserts the retry landed. Deterministic, no more CI-runner slack.
- **~30-line inline comment block** (`src/support/serenity/rest-transport.js:258`) — trimmed both the `projects` and `users` client-construction comments to 4-6 lines each, pointing at `createRetryingFetch` in the shared client's `internal.js` as the source of truth for the retry/idempotency contract, instead of restating it in prose that can drift.
- **Missing POST+429 exhaustion coverage** (`test/support/serenity/rest-transport.test.js:326`) — added a test exhausting all 3 attempts on `createProject` under sustained 429, asserting the last response surfaces as a `SerenityTransportError(429)` rather than being swallowed.
- **Missing PUT/DELETE retry-on-5xx coverage** (`test/support/serenity/rest-transport.test.js:326`) — added a test on `deleteProject` retrying a 503 to eventual success, locking in that DELETE gets the same idempotent-method treatment as GET.

Verified: `npx eslint` clean on both files; `npx mocha test/support/serenity/rest-transport.test.js` — 95 passing (was 93); full repo `npm test` — 14126 passing, 0 failing, `rest-transport.js` at 100%/98.94%/100%/100% line/branch/function/statement coverage.

MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - all prior nits cleanly addressed, no new concerns.
_Changes: Enables bounded-backoff retry on the Semrush REST transport by removing `maxRetries: 0` from both typed client constructions, letting library defaults apply (2 files)._

- Flaky wall-clock Retry-After test replaced with fake timers.
- Overly long inline comment block trimmed to 4-6 lines pointing at `createRetryingFetch`.
- POST+429 exhaustion test added.
- PUT/DELETE retry-on-5xx test added.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 7m 21s | <b>Cost:</b> $3.48 | <b>Commit:</b> <code>657451b155f4d139c269c90010cde8471ce5e7cc</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/serenity/rest-transport.js
```diff
@@ -255,26 +255,26 @@ export function createSerenityTransport({ env, imsToken }) {
...
-  // re-adds the 15s timeout + Accept header the client does not impose. `root` is
-  // the validated origin; the client appends its own '/enterprise/projects/api'
-  // prefix.
+  // Typed Project Engine client; appends '/enterprise/projects/api'. Retry,
+  // backoff, and the POST-never-retries-on-5xx idempotency gate are the shared
+  // library's contract, not this file's — see createRetryingFetch in
+  // spacecat-shared-project-engine-client's internal.js (library defaults:
+  // maxRetries 2, retryBaseDelayMs 200; per-attempt timeout via the injected
+  // `createTimeoutFetch` below, since the retry layer wraps it and calls it
+  // once per attempt).
```
aliciadriani, line 264: **Logical change (1/2):** `maxRetries: 0` removed from the Project Engine client, so the shared library default (`maxRetries: 2`, `retryBaseDelayMs: 200` → 3 attempts total) now governs Project Engine calls. Verified against `createRetryingFetch` in the vendored `internal.js`: the retry layer *wraps* the injected `createTimeoutFetch`, so each attempt gets its own fresh 15s `AbortController` — a per-attempt timeout, not a whole-loop budget.

src/support/serenity/rest-transport.js
```diff
@@ -255,26 +255,26 @@ export function createSerenityTransport({ env, imsToken }) {
...
     authToken,
-    maxRetries: 0,
     fetch: createTimeoutFetch(DEFAULT_TIMEOUT_MS),
   });
 
-  // Typed User Manager client over the sub-workspace lifecycle gateway. Same
-  // shape as the project client; appends its own '/enterprise/users/api' prefix.
-  // Uses `usersRoot` (SEMRUSH_USERS_BASE_URL, or `root` by fallback) so the
-  // lifecycle gateway can be a separate host from Project Engine.
+  // Typed User Manager client over the sub-workspace lifecycle gateway (same
```
aliciadriani, line 271: **Logical change (2/2):** same `maxRetries: 0` removal for the User Manager client. Both call sites are now consistent — confirmed no other `createSerenity*ApiClient` construction or `maxRetries` pin remains anywhere in `src/`.

src/support/serenity/rest-transport.js
```diff
@@ -255,26 +255,26 @@ export function createSerenityTransport({ env, imsToken }) {
...
-  // Typed Project Engine client over the project gateway. maxRetries:0 preserves
-  // the one-shot behaviour the hand-rolled transport had; the injected fetch
-  // re-adds the 15s timeout + Accept header the client does not impose. `root` is
-  // the validated origin; the client appends its own '/enterprise/projects/api'
-  // prefix.
+  // Typed Project Engine client; appends '/enterprise/projects/api'. Retry,
+  // backoff, and the POST-never-retries-on-5xx idempotency gate are the shared
+  // library's contract, not this file's — see createRetryingFetch in
+  // spacecat-shared-project-engine-client's internal.js (library defaults:
+  // maxRetries 2, retryBaseDelayMs 200; per-attempt timeout via the injected
```
aliciadriani, line 262: **Should consider — latency profile change.** With retries enabled, a *sustained* 5xx/timeout on an idempotent read (e.g. `getWorkspaceStatus`, a GET) now takes up to ~85s (3×15s attempts + 2 backoff waits capped at 20s) before surfacing the error, vs. ~15s fast-fail before. Well within the 900s Lambda budget, but can exceed the Fastly edge timeout on synchronous customer paths — the user waits far longer only to still get an error. Acceptable tradeoff (only on *sustained* failure; transient blips recover in 1–2 retries), just flagging for awareness.

src/support/serenity/rest-transport.js
```diff
@@ -255,26 +255,26 @@ export function createSerenityTransport({ env, imsToken }) {
     return imsToken;
   };
 
-  // Typed Project Engine client over the project gateway. maxRetries:0 preserves
-  // the one-shot behaviour the hand-rolled transport had; the injected fetch
-  // re-adds the 15s timeout + Accept header the client does not impose. `root` is
-  // the validated origin; the client appends its own '/enterprise/projects/api'
-  // prefix.
+  // Typed Project Engine client; appends '/enterprise/projects/api'. Retry,
+  // backoff, and the POST-never-retries-on-5xx idempotency gate are the shared
```
aliciadriani, line 259: **Should consider — inherited 429-retries-any-method risk.** The shared retry layer retries 429 for *all* methods including POST (its `isRetryableStatus` documents the assumption: the gateway rejected at the edge, so the create never happened). For POSTs like `transferWorkspaceResources` / `addAiModel` / `createBrandUrls`, if Semrush ever rate-limits *after* partially processing a write, a 429-retried POST could double-fire. Not introduced by this PR — it is the shared library's documented contract — but it becomes live behavior for api-service here. Worth a line in the PR description.

test/support/serenity/rest-transport.test.js
```diff
@@ -27,6 +27,10 @@ use(sinonChai);
 const IMS = 'ims-bearer-test-token';
 const WORKSPACE_ID = '11111111-2222-3333-4444-555555555555';
 const PROJECT_ID = 'proj-xyz';
+// Sub-workspace lifecycle (serenity dual-mode, subworkspace path) parent id —
+// used both by the createSubworkspace/listWorkspaceFamily describe blocks below
+// and by the retry-behaviour suite's real create-call-shape tests.
+const PARENT_WS = 'bb0f4e1c-8bb1-402e-88f2-f68618ea7397';
```
aliciadriani, line 33: **Logical change:** `PARENT_WS` hoisted to module scope (its old declaration inside the sub-workspace `describe` is removed further down) so the retry suite can reuse the real `brand-provisioning.js` → `ensureSubworkspace` → `createSubworkspace` parent-id call shape.

test/support/serenity/rest-transport.test.js
```diff
@@ -322,6 +326,274 @@ describe('Semrush REST transport', () => {
...
+  // the pre-migration hand-rolled transport's one-shot behaviour; this suite
+  // covers it now that the pin is removed and the library defaults
+  // (maxRetries: 2, retryBaseDelayMs: 200 -> 3 attempts total) apply.
+  //
+  // Every test that drives a retryable outcome fakes `setTimeout`/`clearTimeout`
+  // so the jittered exponential backoff (and the per-attempt 15s timeout) never
+  // costs real wall-clock time; `tickAsync` (not `tick`) is required throughout
+  // because the typed client's auth middleware and openapi-fetch both interleave
+  // microtasks before each attempt's timer is registered.
+  describe('retry behaviour', () => {
```
aliciadriani, line 344: **Logical change — new `retry behaviour` suite.** Drives the *real* transport + real vendored `createRetryingFetch` (only `global.fetch` is stubbed), so it validates the actual retry composition end-to-end rather than a mock of it. Fake timers on `setTimeout`/`clearTimeout` keep jittered backoff + the per-attempt 15s timeout off the wall clock; `tickAsync` (not `tick`) is required because the auth middleware + openapi-fetch interleave microtasks before each attempt's timer registers.

test/support/serenity/rest-transport.test.js
```diff
@@ -322,6 +326,274 @@ describe('Semrush REST transport', () => {
...
+      const result = await promise;
+
+      expect(result.id).to.equal('subworkspace-ws-1');
+      expect(fetchStub.callCount).to.equal(2);
+      const retryCall = await callOf(fetchStub, 1);
+      expect(retryCall.method).to.equal('POST');
+      expect(JSON.parse(retryCall.body)).to.deep.equal({ title: 'Adobe Express', resources });
+    }));
+
+    it('does NOT retry a 500 on a POST create (createSubworkspace) -- avoids double-provisioning', withFakeTimers(async (clock) => {
```
aliciadriani, line 410: **Critical guardrail test.** A 500 on the `createSubworkspace` POST (the real provisioning call shape) must NOT retry — single attempt, immediate 500. This is the core anti-double-provisioning guarantee, verified against the library's `isRetryableStatus` (5xx retried only for idempotent methods).

test/support/serenity/rest-transport.test.js
```diff
@@ -322,6 +326,274 @@ describe('Semrush REST transport', () => {
...
+        .catch((e) => e);
+      await clock.tickAsync(60_000);
+      const err = await promise;
+
+      expect(err).to.be.instanceOf(SerenityTransportError);
+      expect(err.status).to.equal(500);
+      expect(fetchStub.callCount).to.equal(1);
+    }));
+
+    it('does NOT retry a 500 on the createProject POST either (same idempotency rule)', withFakeTimers(async (clock) => {
```
aliciadriani, line 429: Same anti-double-provisioning guarantee locked in independently for the `createProject` POST — good that both real create paths are covered.

test/support/serenity/rest-transport.test.js
```diff
@@ -322,6 +326,274 @@ describe('Semrush REST transport', () => {
...
+      // The critical guardrail: a create is a POST, and a 5xx is retried ONLY
+      // for idempotent methods. Retrying a POST on 500 could double-fire the
+      // sub-workspace/resource create upstream, so it must surface the 500
+      // immediately, in a single attempt.
+      fetchStub.resolves(fetchFail(500, { code: 'internal_error' }));
+      const transport = createSerenityTransport({ env: TEST_ENV, imsToken: IMS });
+
+      const resources = { ai: { projects: 3, prompts: 1500 } };
+      const promise = transport.createSubworkspace(PARENT_WS, 'Adobe Express', resources)
+        .catch((e) => e);
```
aliciadriani, line 420: **Nit (test-only, harmless):** the `const p = fn().catch(e=>e); await clock.tickAsync(); await p;` pattern emits a `PromiseRejectionHandledWarning` in the mocha output (rejection attached after the tick). Does not affect results. If the log noise bothers anyone, attach `.catch` synchronously before ticking. Non-blocking.

test/support/serenity/rest-transport.test.js
```diff
@@ -1106,41 +1376,67 @@ describe('Semrush REST transport', () => {
       expect(fetchStub.called).to.equal(false);
     });
 
-    it('throws SerenityTransportError carrying upstream status + parsed body on non-2xx', async () => {
+    it('throws SerenityTransportError carrying upstream status + parsed body on non-2xx (after exhausting GET retries)', async () => {
```
aliciadriani, line 1379: **Logical change — pre-existing test adapted.** This GET 503 test now asserts 3 attempts (`fetchStub.callCount === 3`) before the last response surfaces, with fake timers added to keep backoff off the wall clock. Correct: `getWorkspaceStatus` is a GET (idempotent), so a persistent 503 is now retried to exhaustion.

test/support/serenity/rest-transport.test.js
```diff
@@ -1106,41 +1376,67 @@ describe('Semrush REST transport', () => {
...
         expect(err.status).to.equal(503);
         expect(err.body).to.deep.equal({ code: 'unavailable' });
+        expect(fetchStub.callCount).to.equal(3);
+      } finally {
+        clock.restore();
       }
     });
 
-    it('falls back to raw text when the upstream body is not JSON', async () => {
+    it('falls back to raw text when the upstream body is not JSON (after exhausting GET retries)', async () => {
```
aliciadriani, line 1400: Same adaptation for the non-JSON-500 GET path — 3 attempts, then the last response is surfaced. Good.

test/support/serenity/rest-transport.test.js
```diff
@@ -1106,41 +1376,67 @@ describe('Semrush REST transport', () => {
...
+        const err = await promise;
         expect(err.body).to.equal('gateway exploded');
+        expect(fetchStub.callCount).to.equal(3);
+      } finally {
+        clock.restore();
       }
     });
 
-    it('re-throws a non-abort network error verbatim', async () => {
+    it('retries a network error on GET, then rethrows it verbatim once retries are exhausted', async () => {
```
aliciadriani, line 1418: Network-error GET path: now retried then rethrown verbatim once exhausted (3 attempts). Matches the library's idempotent-method network-error retry branch.

test/support/serenity/rest-transport.test.js
```diff
@@ -1106,41 +1376,67 @@ describe('Semrush REST transport', () => {
...
+        const err = await promise;
+        expect(err.message).to.equal('ENOTFOUND');
+        expect(fetchStub.callCount).to.equal(3);
+      } finally {
+        clock.restore();
+      }
     });
 
-    it('aborts with SerenityTransportError(504) on timeout', async () => {
+    it('aborts with SerenityTransportError(504) on timeout, retried on GET (idempotent) then exhausted', async () => {
```
aliciadriani, line 1434: Timeout-on-GET path: every attempt times out → treated as a retryable network-equivalent failure → 3 attempts before the 504 surfaces. Tick widened to 90s (3×15s + 2 backoff) and `.timeout(15000)` added on the case. Correct per-attempt-timeout-under-retry handling.

aliciadriani (commented):
## Automated review — `/review-pr` pass 1

**Verdict: Approve.** Small, surgical, high-quality reliability fix. It removes `maxRetries: 0` from the two typed Semrush client constructions, letting the shared libraries' built-in bounded-backoff retry layer (`createRetryingFetch`, defaults `maxRetries: 2` / `retryBaseDelayMs: 200`) take effect. Net **2 real source lines removed** plus rewritten comments, backed by a large, carefully-reasoned test suite.

### Verification performed (against the vendored `internal.js`, not just the description)
- **Per-attempt timeout (not whole-loop):** confirmed — `createRetryingFetch` calls `baseFetch(forAttempt(), init)` once per attempt, and the injected `createTimeoutFetch` allocates a fresh `AbortController` + `setTimeout` on every call.
- **Idempotency gate:** confirmed — 429 retries any method; 5xx retries only `GET/HEAD/PUT/DELETE/OPTIONS`. Every create/mutate transport method (`createSubworkspace`, `createProject`, `createBenchmarks`, `createBrandUrls`, `transferWorkspaceResources`, `addAiModel`, `createProjectTags`, `publishProject`, …) is **POST**, and PATCH is deliberately excluded from the idempotent set — so writes never replay on 5xx.
- **Completeness:** no other `maxRetries: 0` pins or `createSerenity*ApiClient` constructions exist in `src/`.
- **Latency ceiling:** worst-case retry chain (~85s for a sustained-503 GET) is far below the 900s Lambda timeout.
- **Tooling:** `npx mocha test/support/serenity/rest-transport.test.js` → 95 passing; `eslint` clean; `npm run type-check` (blocking CI gate for this dir) clean.

### Findings
No **Must Fix**. Two **Should Consider** items and one **Nit** are left as inline comments on the relevant lines (latency profile change; inherited 429-retries-POST upstream-contract risk; a harmless test-only `PromiseRejectionHandledWarning`).

### What's good
Exemplary tests — they drive the *real* transport + *real* vendored retry layer (only `global.fetch` stubbed), so they'd actually catch a regression in the retry/timeout composition. The critical anti-double-provisioning guarantee is pinned by dedicated "500 on POST create is not retried" tests using the real `brand-provisioning.js` call shapes. Comments were rewritten to point at the source of truth rather than restate stale local assumptions. Good scope discipline (no dep bump, no speculative observability hook).

PR #2769, opened 2026-07-07
feat(llmo): add Akamai CDN onboarding API for Optimize at Edge

nit23uec:
## Summary

Adds a multi-step **Akamai (PAPI) onboarding flow** that wires a customer's Akamai property to Adobe LLM Optimizer "Optimize at Edge", mirroring the existing Cloudflare/CloudFront `cdn-onboard` controllers. The PAPI rule-tree logic is ported from the `akamai-api` POC (`edge_optimize/rules_builder.py`), and it uses the `@adobe/spacecat-shared-akamai-client`.

### Endpoints (all gated on site access + LLMO admin)
- `GET  /sites/:siteId/llmo/cdn-onboard/akamai/config` — supported networks + credential headers the UI must collect
- `GET  /sites/:siteId/llmo/cdn-onboard/akamai/properties` — Akamai properties serving the site's domain
- `POST /sites/:siteId/llmo/cdn-onboard/akamai/plan` — dry-run rule-tree merge + before/after diff (no mutation)
- `POST /sites/:siteId/llmo/cdn-onboard/akamai/deploy` — new property version + managed rules + PAPI validation
- `POST /sites/:siteId/llmo/cdn-onboard/akamai/activate` — activate a version to STAGING/PRODUCTION
- `GET  /sites/:siteId/llmo/cdn-onboard/akamai/activation-status` — poll activation status

### Design
- **Credentials are client-supplied per request** via `x-akamai-*` headers (host, client-token, client-secret, access-token, optional account-switch-key) — never persisted, never logged. Mirrors Cloudflare's `x-cloudflare-token` model.
- **`notifyEmails` is derived server-side** from the caller's IMS profile (`trial_email`), never accepted from the client.
- **Mutation guard:** `deploy`/`activate` require the target property to serve the site's own domain on an active hostname.
- **Idempotent merge:** re-running replaces previously-managed rules rather than duplicating them.
- **Advanced-Metadata-free:** the entire managed rule tree (origin, cache key, headers, site failover, and failover reporting) uses GA behaviors only, so it does not require Akamai to provision Advanced Metadata access. Failover is surfaced via the `x-edgeoptimize-fo` response header, detected from the persisted `x-edgeoptimize-api-key` request header on the failover recreate.
- LLMO API key sourced from Tokowaka metaconfig (same as Cloudflare/CloudFront).

### Files
- `src/controllers/llmo/llmo-akamai-utils.js` — rule-tree builders + `mergeIntoTree` (ported from the POC)
- `src/controllers/llmo/llmo-akamai.js` — controller
- Wiring: `src/index.js`, `src/routes/index.js`, `src/routes/required-capabilities.js`, `src/routes/facs-capabilities.js`, `package.json`
- OpenAPI docs: `docs/openapi/{api,llmo-api,parameters}.yaml` + regenerated `docs/index.html`
- Tests: `test/controllers/llmo/llmo-akamai{,-utils}.test.js` (+ route/capability test updates)

### Dependency
Uses the published `@adobe/spacecat-shared-akamai-client` **1.0.2** from npm, which includes the EdgeGrid redirect re-signing fix (adobe/spacecat-shared#1784).

## Test plan
- [x] `llmo-akamai` controller + utils unit tests pass at **100%** statements/branches/functions/lines
- [x] Route / capability / S2S-auth suites pass with the new routes
- [x] OpenAPI bundles cleanly; all 6 paths present; no new lint findings; `docs/index.html` regenerated
- [x] Live end-to-end verification of the read-only endpoints (`config`, `properties`, `plan`, `activation-status`) against a real Akamai property (`prp_1253269`, staging) using real EdgeGrid credentials
- [x] `@adobe/spacecat-shared-akamai-client` published to npm; dependency pinned to the registry version (1.0.2)
- [ ] Exercise the mutating `deploy` + `activate` endpoints against a real property on STAGING


MysticatBot (changes requested):
Hey @nit23uec,

**Verdict: Request changes** - supply chain risk from the personal-account dependency must be resolved before merge.
**Complexity: HIGH** - large diff (3297 lines); API surface + dependency risk flags.
_Changes: Adds a multi-step Akamai (PAPI) onboarding flow with 6 new endpoints mirroring the existing Cloudflare/CloudFront cdn-onboard pattern (15 files)._

## Must fix before merge

1. [Critical] Supply chain risk: dependency `@adobe/spacecat-shared-akamai-client` sourced from personal GitHub account tarball (`nit23uec/spacecat-shared`) - `package.json:86`. No npm provenance, no vulnerability scanning, mutable content. Publish to npm under `@adobe` scope (even as `0.x` or `-beta`) or host in the official `adobe/spacecat-shared` org repo before merging to production.

2. [Important] Missing format validation on `activationId` parameter - `src/controllers/llmo/llmo-akamai.js:530`. Unlike `propertyId`/`contractId`/`groupId` which are regex-validated, `activationId` is passed directly to `client.getActivation()`. Add `/^atv_[A-Za-z0-9]+$/` validation consistent with the other PAPI identifiers.

3. [Important] Error message from AkamaiClient constructor echoed verbatim to caller - `src/controllers/llmo/llmo-akamai.js:97`. `e.message` could expose internal details (hostnames, key prefixes). Log the detail internally, return a generic `'Invalid Akamai credentials'` to the caller.

4. [Important] `assertPropertyServesSite` guard on `activate` endpoint has no test coverage - `test/controllers/llmo/llmo-akamai.test.js`. Only `deploy` tests the guard-blocked path. If someone removes the guard from `activate`, no test would catch the regression.

5. [Important] No test for guard bypass when property matches but `matchedOn` does not include `'hostname'` - `src/controllers/llmo/llmo-akamai.js:233`. This is the primary security guard; add a test where `findPropertiesByDomain` returns `[{ propertyId: ..., matchedOn: ['cname'] }]` and verify the operation is blocked.

<details>

- suggestion: `hostInSiteDomain` is exported and tested (278 lines of tests) but never imported by the controller - either wire it in as a secondary check or remove until needed - `src/controllers/llmo/llmo-akamai-utils.js:506`
- nit: `insertIndex` not validated as integer (unlike `version` in activate) - safe due to `Math.trunc(Number(...))` but inconsistent - `src/controllers/llmo/llmo-akamai.js:290`
- nit: `papiErrorResponse` does not handle 404 from upstream PAPI (property not found surfaces as misleading 502) - `src/controllers/llmo/llmo-akamai.js:167`
- suggestion: Document in OpenAPI spec or `getConfig` response that credentials need both read (property search) and write (activation) PAPI permissions for deploy/activate
- suggestion: Add test for `plan` endpoint that explicitly asserts the domain guard is NOT called (documents the intentional design that plan is read-only and skips `assertPropertyServesSite`)

</details>


---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 11s | <b>Cost:</b> $9.05 | <b>Commit:</b> <code>be0de005b92e53cba76539bad02223eb067ae9c2</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

nit23uec:
Thanks for the thorough review, @MysticatBot. Addressed in ae866df3 (both akamai files remain at 100% statements/branches/functions/lines):

**Must-fix**
2. **activationId validation** — now regex-validated (`/^atv_[A-Za-z0-9]+$/`) before it reaches `getActivation`, consistent with propertyId/contractId/groupId; malformed → 400.
3. **Constructor error leak** — the `AkamaiClient` construction error is now logged internally and the caller gets a generic `Invalid Akamai credentials` (no host/token-prefix detail). Test asserts the generic body.
4. **activate guard coverage** — added a test that `activate` is blocked (403) when the property doesn't serve the site's domain (this was already added in the coverage commit b71738ac, which post-dates the reviewed commit).
5. **Non-hostname match guard** — added a test where `findPropertiesByDomain` returns `[{ propertyId, matchedOn: ['cname'] }]` and the operation is blocked (403), plus the missing-`matchedOn` case.

**Non-blocking**
- Removed the unused `hostInSiteDomain` export + its tests (dead code — the akamai controller never wired it in; the Cloudflare controller has its own copy).
- `papiErrorResponse` now maps upstream **404 → 404** (target not found) instead of a misleading 502.
- `insertIndex` is now validated as a non-negative integer (consistent with `version`).
- Added a test asserting `plan` is read-only and never calls the domain guard.
- OpenAPI `config` description now documents that credentials need both **READ** (properties/plan) and **WRITE** (deploy/activate) PAPI permissions.

**Not addressed, per maintainer direction (#1, npm publish):** the dependency stays pinned to a fork-hosted `1.0.1` tarball that carries the EdgeGrid redirect fix (adobe/spacecat-shared#1784); it'll be swapped for the published `@adobe`-scoped version once that ships. Tracked as a release follow-up.

MysticatBot (approved):
Hey @nit23uec,

**Verdict: Approve** - all prior blocking findings addressed; fixes are correct and well-tested.
**Complexity: HIGH** - large diff (3575 lines); API surface + dependency signal.
_Changes: Adds a multi-step Akamai (PAPI) onboarding flow with 6 new endpoints mirroring the existing Cloudflare/CloudFront cdn-onboard pattern (15 files)._

- activationId now regex-validated (`/^atv_[A-Za-z0-9]+$/`) before reaching the client
- AkamaiClient constructor error logged internally, generic message returned to caller
- `activate` endpoint guard coverage added (403 when property does not serve the site domain)
- Non-hostname matchedOn guard tested (matchedOn: ['cname'] blocked with 403)
- Bonus fixes: papiErrorResponse maps 404 correctly, insertIndex validated, unused hostInSiteDomain removed, plan read-only assertion added, OpenAPI documents credential permission requirements

<details>

- nit: `activationId` pattern is documented in the OpenAPI spec for the query parameter on `activation-status` as bare `type: string` without a pattern constraint, but the controller enforces `^atv_[A-Za-z0-9]+$` - adding the pattern to the spec would let clients validate before sending - `docs/openapi/llmo-api.yaml`
- suggestion: The supply-chain follow-up (replacing the fork-hosted tarball with the published `@adobe`-scoped package) should block the production release - consider adding a CI check or tracking label so it does not ship to prod on the current dependency

</details>

Note: CI checks are currently pending - confirm the build passes before merge.

ABHA61:
**Reviewed the Akamai onboarding controller + rule-tree utils + wiring in depth.** The wiring is solid — all six routes/methods, capability gating, GET→`context.data`, and every `@adobe/spacecat-shared-akamai-client` return shape check out, and the client-supplied EdgeGrid credentials are correctly per-request and never logged. A few things worth addressing before merge, grouped by severity:

| Sev · type | Finding | Loc |
|---|---|---|
| **HIGH** · security | `x-akamai-host` is accepted and passed through to `AkamaiClient` (which builds `https://${host}/…`) with no validation — an authenticated caller can point server-side requests at an arbitrary host (SSRF). Restrict to EdgeGrid hosts (e.g. `*.akamaiapis.net`) and reject IPs/ports/paths. | `:126/155` |
| **HIGH** · correctness | `client.findPropertiesByDomain()` catches per-search failures internally and returns `[]` (only throws on empty domain). So `listProperties` returns `200 {properties: []}` on bad/expired creds, and the deploy/activate guard returns a misleading `403 "property does not serve site"` instead of a 401/5xx. The `.rejects()` unit tests assert an outcome the client never produces. | `:252/337` |
| **MED** · secret-hygiene | The `plan` response returns the fully-merged tree, which embeds the site's LLMO API key in plaintext (the injected `x-edgeoptimize-api-key` value) → lands in browser devtools/HAR exports/proxy logs. The preview should redact the value (the real key is only needed server-side at `deploy`). | `:402`, `utils:454` |
| **MED** · functional | `activate` derives the notify-email only from `profile.trial_email`, which is absent for non-trial/enterprise admins → they get `403 "Unable to derive a notification email"` and can't complete onboarding. Resolve as `trial_email \|\| preferred_username \|\| email` (matching `getCurrentUserEmail`). | `:62/543` |
| **MED** · correctness | Loop-guard/failover detection keys on the mere *existence* of client-suppliable headers (`x-edgeoptimize-api-key` / `x-edgeoptimize-request`) → a client can forge them to bypass Optimize-at-Edge routing or force a false `x-edgeoptimize-fo`. Likely a GA-only limitation — worth a comment stating the trust assumption. | `utils:281` |
| **MED** · convention | Error responses return raw upstream content (the guard's `host`/`seen`; raw PAPI `papiErrors`/`warnings` in the 422) without `cleanupHeaderValue()` sanitization. | `:264/483` |
| **LOW-MED** · correctness | `version`/`insertIndex` are validated via `Number(...)` + `isInteger`, which accepts `true`→1, `"0x10"`→16, `"1e3"`→1000, `["5"]`→5 — a malformed body can activate an unintended version. Require a decimal-integer regex (`/^[0-9]+$/`) before converting. | `:534/232` |
| **LOW-MED** · correctness | `papiErrorResponse` runs its status regexes over the whole error string, which includes up to 1000 chars of the PAPI response body → a genuine 5xx whose body contains `-> 404`/`-> 429` is mis-mapped. Anchor to the status prefix (`/-> 404:/`). | `:174` |
| **LOW** · correctness | `activate` doesn't check for an empty `activationLink` → can return `200` with an empty `activationId`. | `:563` |
| **LOW** · test-coverage | The `AkamaiClient` mock accepts any config and asserts nothing, so `notifyEmails` threading and credential forwarding aren't actually covered — a regression there would ship green. | `test:45` |

**Nits (optional, non-blocking):**
- `accountSwitchKey` is unvalidated (unlike the other identifiers) → a malformed value surfaces as a raw upstream 4xx instead of a clean 400. `:33`
- `requireClient` logs the constructor's `e.message`; logging `e.name` avoids a future footgun if that message ever changes. `:160`
- The six handlers repeat the `getSiteAndCheckAccess → requireClient → requirePropertyRef → resolveRuleConfig` preamble with no shared wrapper. `:320`
- `deploy` runs the metaconfig fetch before the guard — a wasted round-trip on a guard-blocked call. `:434`
- Builder hardening: `insertIndex` NaN clamp `utils:414`; empty `incomingRequestHeaders` would drop the loop guard `utils:284` (not reachable via the controller today). Test hygiene: shared module-level `RULE_TREE` fixture; the mocked `activationIdFromLink` skips the query-string stripping the real one does.

**Please confirm:**
- That `avoidDuplicateHeaders` (on the request-header behaviors) and `ports: ''` (on the origin behavior) are valid PAPI options against the rule-format reference — `deploy`/`activate` haven't been exercised live yet. `utils:156/140`
- Since the UI sends the `x-akamai-*` headers cross-origin, it might be worth a quick check that they're covered by the CORS `access-control-allow-headers` wherever `x-cloudflare-token` already is (looks like the Fastly layer rather than the in-app OPTIONS handler) — likely already handled, just flagging in case.

nit23uec:
Thanks @ABHA61 — really thorough. Addressed in `ced0b283` (both akamai files remain at 100% statements/branches/functions/lines).

### Must-fix
| Finding | Resolution |
|---|---|
| **HIGH · SSRF** (`x-akamai-host`) | `requireClient` now validates the host against `AKAMAI_HOST_RE` (`*.akamaiapis.net`), which rejects IP literals, ports (no `:`) and paths (no `/`) before it reaches the client. Tests: rejects `https://evil.example.com/steal` and `169.254.169.254`. |
| **HIGH · correctness** (`findPropertiesByDomain` returns `[]` on failure) | The deploy/activate guard now **probes with an authenticated `getLatestVersion`** when the search returns empty, so bad/expired creds surface as a real `401`/`5xx` instead of a misleading `403`. `listProperties` documents the swallow limitation. The tests that asserted `.rejects()` are relabelled as defensive; the realistic path (`resolves([])` → `200 {properties: []}`) and the guard-probe `401` are now covered. |
| **MED · secret-hygiene** (API key in `plan` tree) | `plan` returns the merged tree through `redactApiKey`, replacing the injected `x-edgeoptimize-api-key` value with `***`. Per follow-up guidance the key is now treated as a **confidential string that is never logged** (no log line/audit field/response carries it un-redacted; only sent server-side to Akamai at deploy). |
| **MED · functional** (notify email) | `getCallerEmail` now resolves `trial_email \|\| preferred_username \|\| email`, so enterprise/non-trial admins can complete onboarding. |
| **MED · correctness** (loop-guard trusts client headers) | Added a `TRUST ASSUMPTION` comment: keying on header presence is a GA-only limitation; the only party harmed by forging is the forging client (opts itself out of optimization), and stripping/validating at the edge would need Advanced Metadata, which this rule set deliberately avoids. |
| **MED · convention** (raw upstream content) | The guard's `403` message (`host`/`seen`) is now run through `cleanupHeaderValue`. The `422` still returns the structured PAPI `papiErrors`/`warnings` intentionally (JSON body, not headers) for debuggability — happy to string-redact the confidential key out of those too if you'd prefer belt-and-suspenders. |

### Low / low-med
- **`version`/`insertIndex` coercion** → both now require a decimal-integer literal (`/^[0-9]+$/`) before `Number()`; rejects `true`, `"0x10"`, `"1e3"`, `["5"]`. Tests added (incl. `1e3` and `0`).
- **`papiErrorResponse` regex over whole body** → now reads the status from the **first** `-> <NNN>:` token; a `5xx` whose body contains `-> 404` maps to `502`. Test added.
- **empty `activationLink`** → `activate` returns `502` when PAPI yields no link (was `200` with an empty `activationId`). Test added.
- **mock asserts nothing** → the `AkamaiClient` mock now captures its constructor config; tests assert `notifyEmails` threading and `host`/`clientToken`/`accessToken` forwarding.

### Nits
- `accountSwitchKey` now validated (`ACCOUNT_SWITCH_KEY_RE`).
- `requireClient` logs `e.name` (not `e.message`).
- `deploy` runs the guard **before** the metaconfig fetch (no wasted round-trip on a blocked call).
- `mergeIntoTree` hardens the `insertIndex` clamp against `NaN`.
- Mock `activationIdFromLink` now strips the query string like the real one.
- Not done (optional): the shared `getSiteAndCheckAccess → requireClient → …` preamble wrapper — left as-is to keep the diff focused; happy to extract if you want it.

### Confirmations you asked for
- **`avoidDuplicateHeaders: false` / `ports: ''`** — these mirror the values on the live property's existing rule tree (fetched via `getRuleTree` during the live read-only verification of `prp_1253269`) and match the POC that was validated against it. `deploy` PUTs with `validateRules=true`, so PAPI-side validation is the authoritative gate — but `deploy`/`activate` haven't been exercised live yet, so I'll confirm on the first staging run.
- **CORS for `x-akamai-*`** — good flag. Like `x-cloudflare-token`, these are almost certainly covered at the Fastly/CDN layer's `access-control-allow-headers` rather than the in-app OPTIONS handler; I'll verify there before the UI wires up cross-origin calls.

ABHA61 (approved):

PR #2771, opened 2026-07-07
feat: pass specific brandId in the serenity apis

vivesing:
Summary of changes on this branch (feat/semrush-api-v2)

1. Removed the org-wide all variant of two Serenity/Elements endpoints — now brand-only, and renamed to match the prompts route convention:
- GET /v2/orgs/:spaceCatId/brands/:brandId/serenity/brand-presence/url-inspector/filter-dimensions (was serenity/all/..., briefly serenity/:brandId/...)
- GET /v2/orgs/:spaceCatId/brands/:brandId/serenity/brand-presence/weeks (same path history)
- Only the :brandId-scoped, brands/:brandId/serenity/...-shaped versions remain in src/routes/index.js, required-capabilities.js, and facs-capabilities.js, now consistent with the existing brands/:brandId/serenity/brand-presence/prompts route.

2. src/controllers/elements.js simplified and hardened:
- authorizeOrg no longer branches on an all sentinel — :brandId is always required, validated as a UUID, and resolved via getBrandById + resolveBrandWorkspace (brand sub-workspace, falling back to the org's parent workspace).
- Dropped the dead org-wide code paths (listBrands, resolveWorkspaceId imports/usages).
- listUrlInspectorFilterDimensions and listWeeks no longer branch on auth.brand ? ... : ... — always operate on the single resolved brand.
- Review fix: when the brand lookup returns null, authorizeOrg now returns notFound('Brand not found for this organization') instead of forbidden(...) — org access was already confirmed by that point, so a missing brand is a "doesn't exist," not an authz denial.
- Merged in main's new prompts endpoint (listPrompts, authorizeBrandSubWorkspace, authorizeOrgAccess shared helper) — resolves the brand's Semrush sub-workspace and refuses to run against an org-parent workspace (400/404/409 guards).

3. Bug fix — missing await in listWeeks:
- buildService(ctx).getWeeks(...) was calling .getWeeks on an unresolved Promise. Fixed to const service = await buildService(ctx); await service.getWeeks(...).— the missing-await regression, the siteId/:brandId mismatch → 400 path, and the no-CBF_ws_brand-filter payload assertion.

4. Dropped the CBF_ws_brand name filter from the weeks upstream call:
- listWeeks no longer passes brand into the WEEKS element payload — this was causing 404 element not found in workspace errors upstream when the brand name (e.g. "Lovesac") wasn't a registered CBF_ws_brand value.
- Brand scoping now relies entirely on the resolved (sub-)workspace for :brandId; siteId/site_id is still validated against the path's brand (400 on mismatch) but no longer used to build a filter.
- buildWeeksPayload is unchanged — still supports an optional brand filter if a caller opts in later.

5. Merged origin/main (e02757ace) to pick up the prompts endpoint feature, resolving conflicts across elements.js, all three route files, tests, and docs without losing either branch's work.

6. Tests & docs updated to match:
- test/controllers/elements.test.js — brand-scoped defaults throughout; added listWeeks (9 cases) and merged listPrompts suites; updated URLs to the renamed paths; fixed a duplicate stub declaration from the merge.
- test/routes/index.test.js — route list updated to the renamed, brand-only paths.
- docs/llmo-semrush-apis/filter-dimensions-apis.md, docs/elements/semrush-elements-api-reference.md — endpoint signatures, params, and the dropped CBF_ws_brand filter documented; new "List Prompts" section retained from main.


calvarezg:
## Multi-Persona Review — Architect · DBA · Security · Tester

**Recommendation: Changes recommended** (0 critical, 2 major, 3 minor, 3 questions). Security review: no findings — the authz hardening is sound (org check before brand load, org-scoped `getBrandById` closes cross-org IDOR, UUID-validated input).

### Major

**1. Missing/misconfigured PostgREST client → 403 "Brand not found" (masks outages)**
`getBrandById` returns `null` both when the brand is absent *and* when `postgrestClient?.from` is falsy. `authorizeOrg` maps any `null` to `forbidden(...)`, so a DI/config outage surfaces as an authz denial — inconsistent with the sibling check just above that returns `internalServerError` when the `Organization` layer is missing.
*Fix:* guard `postgrestClient` explicitly → `internalServerError`; reserve 403 for a confirmed-absent/cross-org brand. — `src/controllers/elements.js` (authorizeOrg)

**2. Same brand read twice, across two data layers, on the hot path**
`authorizeOrg` loads the brand via `getBrandById` (PostgREST, the wide 8-table `BRAND_SELECT` join), then `resolveBrandWorkspace → resolveBrandSubworkspaceId` reloads it via the `Brand` ORM. Two round-trips to use only `brand.id` + `brand.name`. Also a consistency hazard — org-scoping is enforced on the PostgREST read but not the ORM read.
*Fix:* read the brand once (pass the loaded brand into `resolveBrandWorkspace`, or use a narrow `select('id,name')` for the gate). — `src/controllers/elements.js`; `src/support/brands-storage.js` `BRAND_SELECT`

### Minor
- **`getBrandBySite` → null now returns 400** with message "siteId does not belong to the specified brand" — misleading when the site simply has no brand (was 404). Docs updated to 400, so likely intended; consider clarifying the message. (`listWeeks`)
- **No logging** on the new 400/403/404 auth/resolver branches — a production "brand not found"/"no workspace"/cross-brand rejection isn't diagnosable. Add `log.info/warn` with `spaceCatId`/`brandId`.
- **`CBF_ws_brand` filter in `buildWeeksPayload` now has no production caller** — will rot; tie to a TODO/issue given the open `CBF_project` scoping question.

### Questions
- **c8-ignore scope on `listWeeks`:** it now wraps the cross-brand `siteId` 400 guard, the dropped filter, and the `await` fix. Not flagging coverage (POC convention), but is the ignore scope still right now that an auth boundary lives inside it? The filter-dimensions branches *are* tested.
- **Flat-mode fallback:** `resolveBrandSubworkspaceId` returns null for both "brand not found (ORM)" and "no sub-workspace." Under ORM cache/replication lag, could a brand-scoped request silently fall back to the org-wide parent workspace? Fallback is intended per docs — should the ORM-miss case be distinguished?
- **Breaking path change:** in-repo refs are all updated; confirm no external/frontend caller depends on the removed `/serenity/all/brand-presence/...` paths.

### Working well
Sound tenant isolation, the new cross-brand `siteId` guard closes a stale-filter vector, the `await buildService` fix corrects a real latent bug, and docs were updated in lockstep.

*Automated multi-persona review. Findings verified against the diff; support-fn source (`getBrandById`, `resolveBrandWorkspace`) fetched for context.*

src/controllers/elements.js
```diff
@@ -189,11 +194,20 @@ async function authorizeOrg(ctx) {
...
   }
-  const workspaceId = await resolveWorkspaceId(ctx, spaceCatId);
+
+  if (!isValidUUID(brandIdParam)) {
+    return { error: badRequest('Brand id must be a valid UUID') };
+  }
+  const postgrestClient = ctx?.dataAccess?.services?.postgrestClient;
+  const brand = await getBrandById(spaceCatId, brandIdParam, postgrestClient);
+  if (!brand) {
+    return { error: forbidden('Brand not found or not accessible for this organization') };
```
MysticatBot: issue (blocking): `authorizeOrg` returns `forbidden` (403) when the brand is not found, but the user already passed org-level access control (`hasAccess(organization)`) on the line above. At this point a null brand means "the resource does not exist in this org," not "you lack permission."

Returning 403 here:
- Misleads clients (retrying with different credentials will never help)
- Pollutes 403-based security monitoring with non-authz events
- Contradicts the message text itself ("not found" inside a `forbidden()` is confusing)

Fix: change to `notFound("Brand not found for this organization")`.

src/controllers/elements.js
```diff
@@ -268,8 +281,9 @@ export default function ElementsController(context, log, env) {
   };
 
   /**
-   * GET /v2/orgs/:spaceCatId/serenity/all/brand-presence/weeks
-   * Returns the list of weeks that have Brand Presence data (week filter dropdown).
+   * GET /v2/orgs/:spaceCatId/serenity/:brandId/brand-presence/weeks
+   * Returns the list of weeks that have Brand Presence data (week filter
+   * dropdown), scoped to that brand. An unrelated siteId filter is rejected.
    */
   /* c8 ignore start -- LLMO-6011 POC endpoint; unit tests intentionally deferred */
```
MysticatBot: issue (blocking): The `/* c8 ignore start */` comment dates from when `listWeeks` was a stub POC. This PR introduces three behavioral changes to that handler:
1. siteId-brand mismatch check (`resolved.id !== auth.brand.id` returning 400)
2. Missing `await` on `buildService(ctx)` fixed (the PR description explicitly calls this out as a bug fix)
3. CBF_ws_brand filter removed from the `getWeeks` call

All three can regress silently with the current c8 ignore and zero test coverage. At minimum, add tests for the siteId-mismatch 400 path and a regression test proving `getWeeks` is called on the resolved service (not on a Promise). The await fix is the kind of bug that comes back if someone refactors this handler without tests guarding the contract.

src/routes/facs-capabilities.js
```diff
@@ -788,8 +788,8 @@ const routeFacsCapabilities = {
       // Org-level Serenity catalog reads (no brandId).
       'GET /v2/orgs/:spaceCatId/serenity/models': 'llmo/can_view',
       'GET /v2/orgs/:spaceCatId/serenity/languages': 'llmo/can_view',
-      'GET /v2/orgs/:spaceCatId/serenity/all/brand-presence/url-inspector/filter-dimensions': 'llmo/can_view',
-      'GET /v2/orgs/:spaceCatId/serenity/all/brand-presence/weeks': 'llmo/can_view',
+      'GET /v2/orgs/:spaceCatId/serenity/:brandId/brand-presence/url-inspector/filter-dimensions': 'llmo/can_view',
```
MysticatBot: nit: The comment above says "Org-level Serenity catalog reads (no brandId)" but these two entries now have `:brandId` in the path. Either move them below a new comment (e.g. "Brand-scoped Serenity reads") or update this comment to only cover the `/models` and `/languages` lines.

MysticatBot (changes requested):
Hey @vivesing,

**Verdict: Request changes** - two material issues to address before merge.
**Complexity: HIGH** - medium diff; API surface change + FACS routing update.
_Changes: Converts two org-wide Serenity/Elements endpoints to brand-scoped (`/serenity/:brandId/...`), removes dead org-wide code paths, fixes a missing await bug in listWeeks, and drops the CBF_ws_brand filter that caused upstream 404s (8 files)._

## Must fix before merge

1. [Important] `authorizeOrg` returns 403 for a non-existent brand - should be 404 - `src/controllers/elements.js:204` (details inline)
2. [Important] `listWeeks` has received material logic changes but remains entirely un-tested under c8 ignore - `src/controllers/elements.js:288` (details inline)

<details>

- nit: Comment at `src/routes/facs-capabilities.js:788` says "Org-level Serenity catalog reads (no brandId)" but the two lines below now contain `:brandId` in the path. Move the brand-scoped routes below a separate comment or update the existing one.
- suggestion: Validate `siteId` query param as UUID before the `getBrandBySite` DB round-trip in `listWeeks` - a non-UUID will never match but still incurs a wasted query. Add `if (hasText(siteId) && !isValidUUID(siteId)) return badRequest(...)` for defense-in-depth.
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 33s | <b>Cost:</b> $6.32 | <b>Commit:</b> <code>5e3a05ee0bdb3bcf04e24b0df8e9737ee73cce10</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/controllers/elements.js
vivesing: addressed this comment

src/controllers/elements.js
vivesing: addressed this comment

calvarezg (approved):
## Multi-Persona Review — follow-up (approving) ✅

Both Major findings from the earlier review rounds are resolved, with tests.

**Fixed**
- **Infra failure no longer masked** — `authorizeOrg` now guards `if (!postgrestClient?.from)` → `503 configurationError` before the brand lookup (matches `authorizeBrandSubWorkspace`). Test: `returns 503 (not a masked 404) when the PostgREST client is not available`.
- **Wide 8-table join off the hot path** — new lightweight `getBrandIdentity` (`select('id, name')`) replaces the wide `getBrandById` in the auth gate, with its own tests in `test/support/brands-storage.test.js`.

Also resolved across rounds: `listWeeks` c8-ignore removed with a full test suite (await-fix regression test, siteId 400 branches, `site_id` alias, upstream→502); route moved cleanly under `/v2/orgs/:spaceCatId/brands/:brandId/serenity/...`.

**Optional (Minor, non-blocking):**
- No logging on `authorizeOrg`'s 400/404 reject branches (`authorizeBrandSubWorkspace` logs its invariant; this one doesn't).
- `CBF_ws_brand` filter in `buildWeeksPayload` now has no production caller — tie to a TODO/issue or drop.
- Auth-helper check ordering differs (`authorizeOrg`: access-then-UUID; `authorizeBrandSubWorkspace`: UUID-then-access) — cosmetic.

Security review: no findings — tenant isolation and authz ordering are sound. Coverage green. **Approving.**

*Automated multi-persona review, follow-up round.*

src/routes/facs-capabilities.js
vivesing: addressed

MysticatBot (approved):
Hey @vivesing,

**Verdict: Approve** - prior blockers addressed, code is sound.
**Complexity: HIGH** - medium diff; API surface change + FACS routing update.
_Changes: Converts two org-wide Serenity/Elements endpoints to brand-scoped (`/brands/:brandId/serenity/brand-presence/...`), adds lightweight `getBrandIdentity` lookup, fixes missing await in listWeeks, and drops the CBF_ws_brand filter (10 files)._

<details>

- nit: Comment at `src/routes/facs-capabilities.js:788` says "Org-level Serenity catalog reads (no brandId)" but two of the four routes below it now contain `:brandId`. Move the brand-scoped routes below a separate comment or update this one to cover only `/models` and `/languages`.
- suggestion: `listWeeks` forwards the full `extractQuery(ctx)` object (including `siteId`/`site_id`) to `service.getWeeks`. `buildWeeksPayload` destructures only `{ model, platform, brand }` so extra keys are harmless today, but stripping `siteId`/`site_id` before forwarding communicates intent and prevents a future payload-builder refactor from accidentally leaking them upstream - `src/controllers/elements.js:408`.
- suggestion: The `/* c8 ignore start/stop */` pragmas in `src/support/elements/definitions/weeks.js` (around `buildWeeksPayload`/`transformWeeksResponse`) are now stale since `listWeeks` has full controller-level coverage. Remove the ignores or file a follow-up to unit-test the payload builder directly.
- nit: `authorizeOrg` validates access-then-UUID; sibling `authorizeBrandSubWorkspace` validates UUID-then-access. Aligning to UUID-first would save an unnecessary `Organization.findById` round-trip on junk requests with invalid brandId - `src/controllers/elements.js:225`.
</details>

- `authorizeOrg` now returns `notFound` (not `forbidden`) for a missing brand, plus a 503 guard when the PostgREST client is unavailable.
- `listWeeks` c8-ignore removed with a comprehensive test suite (9 cases covering the await-fix regression, siteId mismatch, snake_case alias, upstream error propagation).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 43s | <b>Cost:</b> $4.79 | <b>Commit:</b> <code>b8c2b4e697ba731acd64adfe71633a9f14bdb059</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2772, opened 2026-07-07
feat(serenity): server-compute branded/non-branded prompt type on every write path

rainer-friederich:
## 1. Abstract
Makes the branded / non-branded classification of Serenity (Semrush sub-workspace) prompts server-computed and applies it through one unified injection layer on every prompt write path. The client can no longer set the `type` dimension.

## 2. Reasoning
Prompt "type" (branded vs. non-branded) was an author-selected tag, which let the two write transports (id-based and name-based) and the several entry points (manual create, edit, CSV import, AI generation, onboarding) drift out of agreement, and let clients set a value that contradicts the prompt text. serenity-docs#31 is the design-of-record: the type should be derived from the prompt text against the brand name + its region-clamped aliases, computed once on the server and re-computed on every edit, so it is always consistent and never client-authored.

Before: each prompt carried whatever `type:branded` / `type:non-branded` tag the client chose; nothing recomputed it and the different write paths handled it inconsistently.

After: the server classifies every prompt on write and injects the correct `type:` tag, stripping any client-supplied one.

- New pure classifier module (`branded-classifier.js`): normalizes text with NFD diacritic-folding plus an explicit fold table for non-decomposing letters (ø/æ/œ/ß/ð/þ/ł/đ), matches brand needles on whole-word boundaries (not substring), and returns the canonical `type:branded` / `type:non-branded` tag. Empty needle set ⇒ non-branded.
- Per-market alias needles: `locations.js` gains `marketForGeoTargetId` (the inverse of the existing `resolveLocation`), so a prompt's geo target selects the region-appropriate brand aliases.
- One unified injection layer (`makeTypeInjector` in `prompts.js`, `resolveTypeTagInjection` in `tags.js`) wired through BOTH the flat and the subworkspace prompt handlers, on create AND update. It strips any incoming `type:` tag (by name on the name-based path, by resolved id on the id-based path) and appends the server-computed one. The classifier is built once per request in the controller from the brand name + aliases and memoized.
- On edit, the type is recomputed from the (possibly new) text before the delete/rewrite, and the response echoes the injected tag.

No backfill: only new writes and edits are affected; existing stored prompts are untouched. The read-only `type` filter continues to recognize the full branded / non-branded vocabulary. `intent` and `source` remain author-selected.

## 4. Required information
- Jira / issue: serenity-docs#31 — https://github.com/adobe/serenity-docs/issues/31
- Other: project-elmo-ui companion PR (read-only Type UI) — see section 8

- project-elmo-ui — consumer: the companion PR removes the Type authoring selector and renders the server-computed value read-only (checkmark / X). It relies on this PR computing and returning the `type:` tag on every write.
- mysticat-data-service — unchanged: no schema or storage change; classification is applied at the api-service layer via the existing tag dimensions.

## 7. Test plan
(a) Local: unit suites for the new classifier, the `marketForGeoTargetId` reverse mapping (incl. leading-zero numerics like AF=004 and a resolveLocation round-trip), the type-injection resolver, and both flat + subworkspace create/update handlers were exercised locally and the controller-level classifier wiring was verified through argument capture.

(b) Per environment — verify through the Serenity prompt endpoints against a Semrush sub-workspace brand:
- Create a prompt whose text contains the brand name (or a market alias) ⇒ response carries `type:branded`; create one that does not ⇒ `type:non-branded`.
- Send a prompt with a client-supplied `type:` tag/id ⇒ it is stripped and replaced by the computed value.
- Edit a branded prompt to remove the brand mention ⇒ re-read shows `type:non-branded` (recomputed on edit).
- CSV import and AI-generated prompts ⇒ each row/prompt receives a computed type with no client input.
- Run on eph/dev first, then stage, before prod; no migration or backfill step is required.

- project-elmo-ui companion PR (read-only Type UI) — related: removes the client Type picker and shows the value read-only. Deploy this api-service PR first (or together) so the UI never authors a type the server would overwrite; the UI change is safe to ship after since the server ignores any client `type:`.

### DIFF
--- a/test/support/serenity/handlers/prompts-subworkspace.test.js
+++ b/test/support/serenity/handlers/prompts-subworkspace.test.js
@@ -333,6 +333,24 @@ describe('prompts-subworkspace handlers', () => {
       expect(result.body.error).to.equal('invalidRequest');
     });
 
+    it('recomputes the type tag from the NEW text on edit (serenity-docs#31, twin of the flat-mode layer)', async () => {
+      // Guards the subworkspace UPDATE injection wiring: without the classifier
+      // arg the defensive `typeof !== function` bypass fires silently, so a
+      // regression in delete-then-create injection would go uncaught here.
+      const transport = makeTransport();
+      const classify = (text) => (/\bacme\b/i.test(text) ? 'type:branded' : 'type:non-branded');
+      const result = await handleUpdatePromptSubworkspace(transport, WS, 'old-id', {
+        text: 'now mentions Acme', tags: ['topic:X', 'type:non-branded'], geoTargetId: 2840, languageCode: 'en',
+      }, log, classify);
+      expect(result.status).to.equal(200);
+      expect(result.body.tags).to.deep.equal(['topic:X', 'type:branded']);
+      expect(transport.createTaggedPrompts).to.have.been.calledOnceWithExactly(
+        WS,
+        'p-us-en',
+        { 'now mentions Acme': ['topic:X', 'type:branded'] },
+      );
+    });
+
     it('re-throws a non-404 delete failure (never creates after a failed delete)', async () => {
       const transport = makeTransport({
         deletePromptsByIds: sinon.stub().rejects(new SerenityTransportError(500, 'boom')),
--- a/test/support/serenity/handlers/prompts-subworkspace.test.js
+++ b/test/support/serenity/handlers/prompts-subworkspace.test.js
@@ -146,6 +146,22 @@ describe('prompts-subworkspace handlers', () => {
       expect(transport.createTaggedPrompts).to.not.have.been.called;
     });
 
+    it('injects the computed type tag from the classifier (serenity-docs#31, twin of the flat-mode layer)', async () => {
+      const transport = makeTransport();
+      const classify = (text) => (/\bacme\b/i.test(text) ? 'type:branded' : 'type:non-branded');
+      const result = await handleCreatePromptsSubworkspace(transport, WS, {
+        prompts: [{
+          text: 'is Acme good?', tags: ['topic:X', 'type:non-branded'], geoTargetId: 2840, languageCode: 'en',
+        }],
+      }, log, classify);
+      expect(result.created[0].tags).to.deep.equal(['topic:X', 'type:branded']);
+      expect(transport.createTaggedPrompts).to.have.been.calledOnceWithExactly(
+        WS,
+        'p-us-en',
+        { 'is Acme good?': ['topic:X', 'type:branded'] },
+      );
+    });
+
     it('skips inputs whose slice has no project (one listing, no per-input lookup)', async () => {
       const transport = makeTransport();
       const result = await handleCreatePromptsSubworkspace(transport, WS, {
--- a/src/controllers/serenity.js
+++ b/src/controllers/serenity.js
@@ -65,6 +65,8 @@ import { ensureSubworkspace, decommissionBrandWorkspace } from '../support/seren
 import { isSerenityActiveForOrg } from '../support/serenity/serenity-active.js';
 import { MAX_TOPICS_ON_CREATE } from '../support/serenity/brand-provisioning.js';
 import { STANDARD_PROMPT_TAGS, PROJECT_STANDARD_TAGS } from '../support/serenity/prompt-tags.js';
+import { marketForGeoTargetId } from '../support/serenity/locations.js';
+import { brandNeedles, classifyBrandedTag } from '../support/serenity/branded-classifier.js';
 import AccessControlUtil from '../support/access-control-util.js';
 import { resolveBrandUuid } from '../support/prompts-storage.js';
 import {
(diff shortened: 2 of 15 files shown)


test/support/serenity/handlers/prompts-subworkspace.test.js
```diff
@@ -146,6 +146,22 @@ describe('prompts-subworkspace handlers', () => {
...
+          text: 'is Acme good?', tags: ['topic:X', 'type:non-branded'], geoTargetId: 2840, languageCode: 'en',
+        }],
+      }, log, classify);
+      expect(result.created[0].tags).to.deep.equal(['topic:X', 'type:branded']);
+      expect(transport.createTaggedPrompts).to.have.been.calledOnceWithExactly(
+        WS,
+        'p-us-en',
+        { 'is Acme good?': ['topic:X', 'type:branded'] },
+      );
+    });
```
MysticatBot, line 163: issue (blocking): The subworkspace **update** path now accepts `classifyPromptType` as a 6th argument and wires `makeTypeInjector` into the flow (production lines 257-261), but this test file only adds a type-injection test for the **create** path. The existing update tests pass only 5 arguments, so `classifyPromptType` is `undefined` and the defensive `typeof !== function` bypass fires silently - meaning a wiring bug in the subworkspace update injection would not be caught.

The flat-mode twin has a dedicated update test (`prompts.test.js` - "recomputes the type tag on edit"), but the subworkspace twin is structurally different (resolves projects from a live listing, has its own delete-then-create flow). A single test analogous to the create test at line 149 - passing a classifier and asserting the response `tags` reflect recomputation on edit - would close this gap.

MysticatBot (changes requested):
Hey @rainer-friederich,

**Verdict: Request changes** - one test gap on a production write path.
**Complexity: HIGH** - large diff, API surface signal.
_Changes: Moves prompt type classification from client-controlled to server-computed on every write path, using a unified diacritic-folding whole-word classifier and injection layer (14 files)._

## Must fix before merge
1. [Important] Missing type-injection test for subworkspace update path - `test/support/serenity/handlers/prompts-subworkspace.test.js` (details inline)

<details>

- nit: Response echoes `nextText` instead of `typed.text` in subworkspace update response body - inconsistent with create path which echoes `typed.text` - `src/support/serenity/handlers/prompts-subworkspace.js:312`
- suggestion: Add error-path tests for `resolveTypeTagInjection` (transport failure on `listProjectTagTree`, empty create response yielding `computedId: undefined`) - `test/support/serenity/handlers/tags.test.js`
- suggestion: Document upstream tag-creation idempotency assumption in `resolveTypeTagInjection` with a brief comment, matching the pattern at `resolveOrCreateClosedTag` - `src/support/serenity/handlers/tags.js:325`
- suggestion: Add a multi-prompt batch test exercising the `makeTypeInjector` cache (mixed branded/non-branded in same project, verifying `resolveTypeTagInjection` called once per unique key) - `test/support/serenity/handlers/prompts.test.js`
- suggestion: Consider caching the classifier closure per `brandUuid` within the controller scope (or accepting the brand model from `authorize`) to avoid 2 DB round-trips per request on repeated invocations - `src/controllers/serenity.js:449`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 19s | <b>Cost:</b> $8.68 | <b>Commit:</b> <code>b1a6e43b2da1e35fcc80f5255c1613fcd19b806d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the review in `2f2c681a`.

MysticatBot:
- [Important] Missing type-injection test for the subworkspace update path — done. Added `recomputes the type tag from the NEW text on edit` in `prompts-subworkspace.test.js`, passing the classifier and asserting the response tags reflect recomputation through the delete-then-create flow (mirrors the flat-mode `handleUpdatePrompt` twin). This closes the silent `typeof !== function` bypass gap.
- nit: response echoed `nextText` instead of `typed.text` (`prompts-subworkspace.js:312`) — done. Now echoes `typed.text` to match the create path. Note they are provably equal (`makeTypeInjector` spreads `input` and only rewrites tags/tagIds, never text), so this is consistency, not a behaviour fix.
- suggestion: error-path tests for `resolveTypeTagInjection` — done. Added two: a transport failure while listing the project tag tree (propagates, never creates), and an empty create response yielding `computedId: undefined` (injection skipped, input tagIds untouched).
- suggestion: document the idempotency assumption in `resolveTypeTagInjection` — done. Added a comment at the create-on-demand branch spelling out the single-request no-duplicate guarantee and the resolve-or-create assumption it shares with `resolveOrCreateClosedTag`.
- suggestion: multi-prompt batch test exercising the `makeTypeInjector` cache — done. Added `makeTypeInjector cache` asserting `resolveTypeTagInjection` (via `listProjectTags`) runs once per unique `(project, type)` key and re-resolves only on a new key.
- suggestion: cache the classifier closure per `brandUuid` / accept the brand from `authorize` — declined. The classifier already loads the brand once per request and memoizes alias needles per market, so within a request there is no repeated work. A cross-request cache would risk classifying against a stale brand name/alias set (a correctness-sensitive value), and `authorize` does not currently load the Brand model, so threading it through would be an out-of-scope refactor of the auth path. The two per-request reads on a write path are an acceptable cost.

Also fixed while here (not flagged by the review, found during triage):
- Stripped a stray NUL byte in `prompts.js` — the `makeTypeInjector` cache-key separator was `\x00` instead of a space (`` `${projectId}\x00${typeTag}` ``). It made the file parse as binary (grep/diff/bundling tooling silently mis-handled it) though behaviour was unchanged (NUL is a valid, collision-safe key char). Now a normal space; file is clean UTF-8 again.

Type-check gate and the affected serenity suites are green locally.

test/support/serenity/handlers/prompts-subworkspace.test.js
rainer-friederich: Fixed in `2f2c681a` — added `recomputes the type tag from the NEW text on edit` in this file, passing the classifier and asserting the response tags reflect recomputation through the delete-then-create flow. Closes the silent classifier-undefined bypass.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all prior findings addressed, no new issues found on re-review.
**Complexity: HIGH** - large diff, API surface signal.
_Changes: Moves prompt type classification from client-controlled to server-computed on every write path, using a unified diacritic-folding whole-word classifier and injection layer (14 files)._

- Missing type-injection test for subworkspace update path - added in 2f2c681
- Response echo inconsistency (nextText vs typed.text) - fixed
- Error-path tests for resolveTypeTagInjection - added
- Idempotency assumption documentation in resolveTypeTagInjection - added
- Multi-prompt batch test for makeTypeInjector cache - added
- Classifier caching suggestion - declined with sound rationale (per-request memoization already prevents repeated work; cross-request cache risks stale brand data)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 51s | <b>Cost:</b> $5.22 | <b>Commit:</b> <code>2f2c681ad5601fe3df0cecffe5c079ab5a71e47c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2777, opened 2026-07-07
feat(preflight): log terminal run outcomes from both poll handlers

noozoo:
Emit a server-side observability log when a preflight AsyncJob reaches a terminal state on poll. Previously getPreflightById (Mystique path) logged nothing and getPreflightJobStatusAndResult (audit-worker path) logged only the COMPLETED summary, so failed runs were invisible in spacecat-api logs.

Extract the shared logPreflightOutcome helper so both handlers stay in lockstep as the Mysticat migration proceeds in stages, and tag each line with process=audw|myst (emitted last) to identify which surface emitted it. jobId is the AsyncJob id — the same key the worker/Mystique log as async_job_id — so failures are joinable across processes.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
https://jira.corp.adobe.com/browse/SITES-47550

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/controllers/preflight.js
+++ b/src/controllers/preflight.js
@@ -62,6 +62,44 @@ export function countIssuesForAudit(audit) {
 }
 
 /**
+ * Process identifiers tagged onto the preflight outcome logs so a single log
+ * line tells you which surface emitted it.
+ *  - PREFLIGHT_PROCESS_AUDW → audit-worker path: POST/GET /preflight/jobs (SQS)
+ *  - PREFLIGHT_PROCESS_MYST → Mystique path: /sites/:siteId/preflights
+ */
+export const PREFLIGHT_PROCESS_AUDW = 'audw';
+export const PREFLIGHT_PROCESS_MYST = 'myst';
+
+/**
+ * Emits the server-side observability log for a terminal AsyncJob when preflight is
+ * polled. Shared by both poll handlers — getPreflightJobStatusAndResult (audit-worker
+ * path) and getPreflightById (Mystique path) — so the two appear the same.
+ * @param {Object} log - The logger instance
+ * @param {string} processName - PREFLIGHT_PROCESS_AUDW | PREFLIGHT_PROCESS_MYST
+ * @param {string} jobId - The AsyncJob id
+ * @param {Object} job - The AsyncJob entity
+ */
+export function logPreflightOutcome(log, processName, jobId, job) {
+  const status = job.getStatus();
+  const result = job.getResult();
+  if (status === AsyncJob.Status.COMPLETED && isNonEmptyArray(result)) {
+    const summary = result.map((r) => ({
+      pageUrl: r?.pageUrl,
+      step: r?.step,
+      audits: (Array.isArray(r?.audits) ? r.audits : []).map((a) => ({
+        name: a?.name,
+        type: a?.type,
+        opportunities: Array.isArray(a?.opportunities) ? a.opportunities.length : 0,
+        issues: countIssuesForAudit(a),
+      })),
+    }));
+    log.info(`[Preflight] Run complete. jobId=${jobId} status=${status} results=${JSON.stringify(summary)} process=${processName}`);
+  } else if (status === AsyncJob.Status.FAILED) {
+    log.error(`[Preflight] Run failed. jobId=${jobId} status=${status} error=${JSON.stringify(job.getError())} process=${processName}`);
+  }
+}
+
+/**
  * Creates a preflight controller instance
  * @param {Object} ctx - The context object containing dataAccess and sqs
  * @param {Object} ctx.dataAccess - The data access layer for database operations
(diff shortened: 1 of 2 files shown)


src/controllers/preflight.js
```diff
@@ -61,6 +61,44 @@ export function countIssuesForAudit(audit) {
...
+      audits: (Array.isArray(r?.audits) ? r.audits : []).map((a) => ({
+        name: a?.name,
+        type: a?.type,
+        opportunities: Array.isArray(a?.opportunities) ? a.opportunities.length : 0,
+        issues: countIssuesForAudit(a),
+      })),
+    }));
+    log.info(`[Preflight] Run complete. jobId=${jobId} status=${status} results=${JSON.stringify(summary)} process=${processName}`);
+  } else if (status === AsyncJob.Status.FAILED) {
+    log.error(`[Preflight] Run failed. jobId=${jobId} status=${status} error=${JSON.stringify(job.getError())} process=${processName}`);
```
johobot: **Main concern — terminal outcome re-logged on every poll, at `error` level.**

This helper is stateless and fires on every invocation (same at `:351` and `:797`), and clients poll a job until — and often just past — terminal. So one failed job polled N times produces N identical `log.error` lines, with N controlled by the client. That'll inflate error-rate dashboards/alerting.

Two things worth deciding deliberately:
- **Level:** a preflight `FAILED` is often an expected outcome (bad customer page / upstream analyze failure — your fixture is literally `MYSTICAT_ERROR`), not a defect in this service. `log.warn` stays visible/greppable without polluting error SLOs.
- **Repetition:** consider emitting only on the transition into terminal (gate on a `metadata.outcomeLogged` marker, or on `endedAt` landing in the current poll window). If dedup is out of scope for this PR, maybe note the repetition in the description so on-call isn't surprised.

Not blocking, but this is the one I'd want a conscious call on. A small test that polls the same terminal job 2–3× and asserts `log.error.callCount` would lock in whatever behavior you choose.

Minor, same line: `JSON.stringify(job.getError())` at error level — worth a sanity check against "never log secrets." Structured `{code, message}` in the tests so likely fine; just confirm upstream never puts tokens/customer data in the error payload.

src/controllers/preflight.js
```diff
@@ -61,6 +61,44 @@ export function countIssuesForAudit(audit) {
...
+/**
+ * Emits the server-side observability log for a terminal AsyncJob when preflight is
+ * polled. Shared by both poll handlers — getPreflightJobStatusAndResult (audit-worker
+ * path) and getPreflightById (Mystique path) — so the two appear the same.
+ * @param {Object} log - The logger instance
+ * @param {string} processName - PREFLIGHT_PROCESS_AUDW | PREFLIGHT_PROCESS_MYST
+ * @param {string} jobId - The AsyncJob id
+ * @param {Object} job - The AsyncJob entity
+ */
+export function logPreflightOutcome(log, processName, jobId, job) {
```
johobot: **Redundant `jobId` param.** The helper takes both `jobId` and `job`, but `job.getId()` exists (and both callsites already pass the job's own id — `jobId` in the audw path, `asyncJob.getId()` in the myst path). Passing an id and the object that contains it invites drift. Could drop `jobId` and call `job.getId()` internally.

src/controllers/preflight.js
```diff
@@ -61,6 +61,44 @@ export function countIssuesForAudit(audit) {
...
+      pageUrl: r?.pageUrl,
+      step: r?.step,
+      audits: (Array.isArray(r?.audits) ? r.audits : []).map((a) => ({
+        name: a?.name,
+        type: a?.type,
+        opportunities: Array.isArray(a?.opportunities) ? a.opportunities.length : 0,
+        issues: countIssuesForAudit(a),
+      })),
+    }));
+    log.info(`[Preflight] Run complete. jobId=${jobId} status=${status} results=${JSON.stringify(summary)} process=${processName}`);
```
johobot: **`process=` appended last.** Putting the discriminator at the end, after a variable-length JSON blob, makes it awkward to key on in log tooling and forces the test into `.split('results=')[1].split(' process=')[0]`. Reads better adjacent to `jobId` near the front — cheap now, annoying once dashboards depend on the position.

src/controllers/preflight.js
```diff
@@ -309,35 +347,20 @@ function PreflightController(ctx, log, env) {
...
-        }));
-        log.info(`[Preflight] Run complete. jobId=${jobId} status=${status} results=${JSON.stringify(summary)}`);
-      }
+      // Emit the terminal-state observability log (shared with the Mystique path).
+      logPreflightOutcome(log, PREFLIGHT_PROCESS_AUDW, jobId, job);
 
       return ok({
         jobId: job.getId(),
-        status,
+        status: job.getStatus(),
```
johobot, line 356: **Lost the local `const`s here.** The refactor dropped `const status`/`const result`, so `getStatus()`/`getResult()` now run once in the helper and again in this response object. Cheap in-memory getters so no real perf impact — just a minor readability regression from the previous single-evaluation version.

src/controllers/preflight.js
```diff
@@ -770,6 +793,11 @@ function PreflightController(ctx, log, env) {
         return preflightError('PREFLIGHT_LIFECYCLE_UNAVAILABLE', 'Failed to load preflight lifecycle data', 503);
       }
 
+      // Log terminal state if asyncJob exists
```
johobot: Nit: this comment just restates the `if (asyncJob)` on the next line.

johobot (commented):
Thorough pass on the logging change — intent is sound (failed runs were invisible; one helper keeps the two poll paths in lockstep during the migration), contracts are unchanged, and coverage is good.

I ran `test/controllers/preflight.test.js` locally (node 24.15.0, mocha 11.7.5): **109 passing, 0 failing**, including all six new assertions — nothing here is broken. Comments below are design/observability feedback; the only one I'd push on is the FAILED-at-error-on-every-poll behavior (see inline on line 98). The rest is polish.

johobot (approved):
Re-reviewed at `7192a374`. Ran the suite locally: 111 passing, 0 failing.

All the substantive points are addressed: FAILED now logs at `warn` (defuses the error-SLO concern), the error payload is reduced to `errorCode`/`errorMessage` with a test proving `details` never leaks, the `jobId` param is gone, and `process=` moved up front. The explicit "one warn per poll" test locks in the stateless-by-design decision.

One tiny residual, non-blocking: `errorMessage` is still freeform worker text, so it's marginally higher-risk than `errorCode` alone. Fine as a debuggability trade-off, just flagging it. LGTM.

PR #2784, opened 2026-07-09
feat(llmo): add Cloudflare Logpush onboarding endpoint (LLMO-5869)

IrisAlexandrescu:
## DRAFT — blocked on 2 upstream repos publishing

This PR is intentionally opened as a draft. It depends on two in-flight, not-yet-merged/published PRs:
- `spacecat-shared` #1796 — adds `CLOUDFLARE_LOGPUSH_FIELDS` export (this repo's `output_options.field_names` source)
- `spacecat-auth-service` #609 (LLMO-6116) — adds the ownership-token endpoint this repo calls

I verified this PR's code locally via `npm link` against the real unpublished source in both repos (not mocks-only), so the logic is proven — but `package.json` still pins the *old* published versions of `@adobe/spacecat-shared-cloudflare-client` (1.2.1) and `@adobe/spacecat-shared-utils` (1.124.1), neither of which has the new exports yet. CI will fail on this PR as-is until those two publish and this PR's dependency versions are bumped for real. Please review the code now; hold the merge until the bump.

## Summary

- Repo 3 (the primary deliverable) of the LLMO-5869 Cloudflare Logpush automation 3-repo plan.
- New `POST /sites/:siteId/llmo/cdn-onboard/cloudflare/zones/:zoneId/logpush`, completing the Cloudflare log-forwarding onboarding flow:
  1. Idempotency check against the zone's existing Logpush jobs on the `http_requests` dataset, matched on the computed `destination_conf` — Cloudflare job creation is **not** inherently idempotent, unlike CloudWatch Logs `CreateDelivery`, so this is the primary guard.
  2. Trigger a fresh ownership challenge (`requestLogpushOwnership`) — Cloudflare writes a token file to the customer's cdn-logs S3 bucket.
  3. Fetch that token from `spacecat-auth-service`'s new endpoint (LLMO-6116), forwarding the caller's own `Authorization` header rather than a service credential — api-service never touches the S3 bucket directly (least-privilege boundary; only auth-service has that access).
  4. Create the Logpush job with the token.
- TOCTOU race recovery: on a `createLogpushJob` failure, re-lists jobs and treats a now-matching job as success, rather than pattern-matching an unverified Cloudflare error message — the exact duplicate-destination error shape is called out in the ticket as an open, E2E-only confirm, not something to guess at.
- Requires the site to already be provisioned for Cloudflare log forwarding (bucket/path/region on `cdnBucketConfig`) — returns `409` with an actionable message otherwise.
- Route + S2S capability + FACS non-resource-param (`zoneId`) + OpenAPI (`site-llmo-cloudflare-logpush`) wiring.
- `.env.example` doc for the new `AUTH_SERVICE_BASE_URL` env var (per-environment, not yet in real deploy config).

## Related Issues

LLMO-5869 (epic LLMO-5809)

## Test plan

- [x] New tests in `llmo-cloudflare.test.js` — happy path, idempotent no-op, race recovery, 409 missing-config (both variants), 400s (zoneId format/missing, CF token missing), 404 site-not-found, 502s (job listing / ownership challenge / auth-service call failure incl. non-ok + missing-token / job creation with failed re-check), Authorization-header-forwarding assertion, null-list defensive branches
- [x] Full existing `llmo-cloudflare.test.js` suite — 92 passing, 100% statement/branch/line/function coverage on `llmo-cloudflare.js`
- [x] Full repo suite (`npm test`) — 14422 passing, coverage thresholds green
- [x] `npm run docs:lint` — valid, no new warnings
- [x] Verified end-to-end against real (linked, unpublished) `spacecat-shared` source — not mocked around

## Not yet verified (needs a dev deploy + real Cloudflare zone, tracked as follow-up, not blocking this review)

- S3 challenge-file read-after-write timing under real conditions (flagged in the ticket as a currently-occurring failure mode in the manual flow today)
- Cloudflare's actual behavior on a genuine duplicate-destination race
- Real cross-service auth forwarding + the auth-service IAM boundary holding in practice


docs/openapi/llmo-api.yaml
```diff
@@ -5972,6 +5972,83 @@ site-llmo-cloudflare-routes:
     security:
       - ims_key: [ ]
 
+site-llmo-cloudflare-logpush:
+  parameters:
+    - $ref: './parameters.yaml#/siteId'
+    - name: zoneId
```
nit23uec: please avoid cloudflare owned terms like zoneId in path segments. they should be captured in the body.

src/routes/index.js
```diff
@@ -554,6 +554,7 @@ export default function getRouteHandlers(
     'GET /sites/:siteId/llmo/cdn-onboard/cloudflare/zones': llmoCloudflareController.listZones,
     'POST /sites/:siteId/llmo/cdn-onboard/cloudflare/deploy': llmoCloudflareController.deployWorker,
     'POST /sites/:siteId/llmo/cdn-onboard/cloudflare/routes': llmoCloudflareController.addRoute,
+    'POST /sites/:siteId/llmo/cdn-onboard/cloudflare/zones/:zoneId/logpush': llmoCloudflareController.createLogpush,
```
nit23uec:
```suggestion
    'POST /sites/:siteId/llmo/cdn-onboard/cloudflare/logpush': llmoCloudflareController.createLogpush,
```

src/controllers/llmo/llmo-cloudflare.js
```diff
@@ -569,12 +615,143 @@ function LlmoCloudflareController(ctx) {
...
+    const cdnBucketConfig = getCdnBucketConfig(site);
+    const { bucketName, allowedPaths, region } = cdnBucketConfig || {};
+    const pathPrefix = allowedPaths?.[0];
+    if (!hasText(bucketName) || !hasText(pathPrefix) || !hasText(region)) {
+      return createResponse({
+        message: 'Site is not provisioned for Cloudflare log forwarding; run provisioning first',
+      }, 409);
+    }
+
+    const siteId = site.getId();
```
nit23uec: we should check if the registrable domain of the zone in the zoneId matches with the given site base url or not.

IrisAlexandrescu:
@nit23uec Addressed both in \`d904f110\`:

1. **zoneId moved from path to body** — route is now \`POST /sites/:siteId/llmo/cdn-onboard/cloudflare/logpush\`, matching \`addRoute\`'s existing convention in this same file (zoneId is a Cloudflare identifier, not a SpaceCat entity). Updated routing, S2S capability, FACS param classification (removed the now-unneeded \`zoneId\` non-resource-param entry — no route has it as a path param anymore), OpenAPI, and tests.
2. **Zone-domain validation added** — fetches the zone via a new \`CloudflareClient.getZone(zoneId)\` (companion PR on \`spacecat-shared\`, since this method didn't exist yet) and rejects with \`400\` if the zone's registrable domain doesn't match the site's own domain, before doing anything else.

One side effect worth flagging: the domain check let me delete a small piece of dead code — the Logpush job's \`name\` field used to recompute \`registrableDomain(site.getBaseURL())\` a second time with a \`|| siteId\` fallback. That fallback can never fire now, since the new guard performs the identical computation earlier and already guarantees it's non-null before we get there. Simplified to reuse the guard's result directly.

98 tests passing, 100% coverage on the controller, verified locally against the real (linked, not-yet-published) \`getZone\` implementation.

PR #2786, opened 2026-07-09
feat(prompts): add source filter to v2 prompts list endpoint

dzehnder:
## What

Adds a `source` query parameter to `GET /v2/orgs/:spaceCatId/brands/:brandId/prompts`, so callers can filter prompts by their `source` (e.g. `gsc`, `semrush`, `base_url`, `config`). The `source` column already exists on the `prompts` table and is already returned per row — this adds the missing filter predicate, mirroring the existing `origin` filter exactly.

## Why

The LLMO usage-tracking report's **"GSC Prompts" column has shown 0 for every customer since 2026-06-19** — not data loss, a reporting gap. The report generator (`OneAdobe/llmo-usage-report-node`) stubbed the GSC count to `0` for brandalf-enabled orgs with an explicit TODO:

> `// TODO: return accurate count once the brandalf prompts API supports filtering by source param`

This endpoint is that missing primitive. A follow-up PR in the generator will consume `?source=gsc` (and the same applies to the Library-prompts column). Full analysis in the implementation spec.

## Changes

- `src/support/prompts-storage.js` — `listPrompts` accepts `source`, applies `.eq('source', source)` (exact match).
- `src/controllers/brands.js` — `listPromptsByBrand` forwards `source` from query params.
- `docs/openapi/prompts-v2-api.yaml` — documents `source` as a **free-form string, no enum** (sources are open-ended and added by upstream producers like DRS without API changes).
- Tests:
  - Unit tests that **actually assert the filter fires** (recording stubs, not the no-op `eq: () => chain` — a cloned test would pass even with the filter deleted), plus an omitted-source no-op case.
  - Controller forwarding test (asserts `source` reaches the query, not just `200`).
  - `test/it/` integration case proving `?source=gsc` narrows rows, including a **fail-loud `?source=__nonexistent__` → `total: 0`** check (guards against the endpoint silently ignoring the param during a two-stage deploy).

## Compatibility & safety

- Fully backward-compatible: omitting `source` is a true no-op.
- Security: `.eq('source', source)` binds the value as a PostgREST parameter (no injection), exact-match only; tenant scoping (org access check + `organization_id`/`brand_id` filters) is upstream and unconditional — `source` can only narrow an already-scoped set.
- No new index needed: `count: 'exact'` already runs on every list call; adding a predicate makes it strictly cheaper.

## Testing

- `npx mocha test/support/prompts-storage.test.js test/controllers/brands.test.js` → 553 passing.
- `npm run docs:lint` → API description valid (pre-existing warnings only).
- IT: `test/it/postgres/categories-prompts.test.js` covers the new `source` filter.

### DIFF
--- a/docs/openapi/prompts-v2-api.yaml
+++ b/docs/openapi/prompts-v2-api.yaml
@@ -78,6 +78,16 @@ v2-prompts-by-brand:
         type: string
         enum: [ai, human]
       description: Filter by origin (ai or human)
+    - name: source
+      in: query
+      required: false
+      schema:
+        type: string
+      description: >-
+        Filter by prompt source (exact match), e.g. `gsc`, `semrush`,
+        `base_url`, `config`. Free-form string — sources are open-ended and
+        added by upstream producers (e.g. DRS) without API changes, so no enum
+        is enforced.
     - name: sort
       in: query
       required: false
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -401,7 +401,7 @@ function BrandsController(ctx, log, env) {
     const { spaceCatId, brandId } = context.params || {};
     const {
       limit, page, categoryId, topicId, status,
-      search, region, origin, sort, order,
+      search, region, origin, source, sort, order,
     } = getQueryParams(context);
 
     try {
@@ -449,6 +449,7 @@ function BrandsController(ctx, log, env) {
         search,
         region,
         origin,
+        source,
         sort,
         order,
         limit,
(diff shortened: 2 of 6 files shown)


MysticatBot (approved):
Hey @dzehnder,

**Verdict: Approve** - clean, well-scoped addition of a single query filter following established patterns exactly.
_Changes: Adds an optional `source` query parameter to the v2 prompts list endpoint, filtering by exact match on the existing `source` column (6 files)._
Note: Recommend a human read before merge - this change modifies a documented API contract (`docs/openapi/prompts-v2-api.yaml`). The bot review is a complement to, not a replacement for, a human read here.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 6s | <b>Cost:</b> $4.94 | <b>Commit:</b> <code>4e93544fa0092c3eedef0fae6247b043a6c3cdce</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2787, opened 2026-07-10
fix: crashing x-error header in onboard-site generic error

igor-grubic:
## Changes Made

Fixes a 500 crash on the site-only onboarding endpoint's generic failure path.

- `GENERIC_ONBOARD_ERROR` in `onboardSiteOnly()` contained an em dash (U+2014), outside the HTTP header-safe character range. `badRequest()`/`internalServerError()` copy the message verbatim into the `x-error` response header, so any generic-failure response (e.g. re-onboarding an already-onboarded domain) crashed with a raw `500 Invalid character in header content ["x-error"]` instead of the intended `400`.
- Rewrote the message to be ASCII-only. It stays short, plain, and generic (does not reveal the internal reason — already-onboarded, org conflict, etc. — per the existing "customers never see the internal reason" policy).
- Wrapped both call sites with `cleanupHeaderValue()`, the sanitizer already used throughout this file for the same class of issue, as defense in depth against a future copy edit reintroducing an unsafe character.
- Updated the one unit test that hardcoded the old message text, and added a header-safety regression assertion (the existing test mocks out real header serialization, so it couldn't have caught this bug otherwise).

**Note on message wording:** an earlier revision of this PR also added a "please use our onboarding guide" sentence + link to this backend string. That moved to the UI instead (see the companion `project-elmo-ui` PR below) — a hardcoded English string in an API response can't be localized, so the actual user-facing copy (with a real clickable link) now lives in `project-elmo-ui`'s localized messages. This backend message is only seen by a caller that skips the UI (e.g. a direct API/curl consumer), so it stays a short, plain fallback.

### Related Issues

Relates to LLMO-6147

## Testing the PR changes

1. `POST /v2/orgs/:spaceCatId/llmo/onboard-site` with a domain that's already onboarded (or otherwise rejected by `validateSiteNotOnboarded`).
2. Confirm a clean `400` with the new plain-text message, not a `500`.
3. Run `npx mocha --timeout 15000 test/controllers/llmo/onboard-site-only.test.js` - 20/20 passing.
4. Run `npx mocha --timeout 15000 'test/controllers/llmo/*.test.js'` - 2081/2081 passing (no regressions).

## Screenshots/Videos

No screenshots available (backend-only change).

## Additional Notes

Companion frontend fix (separate repo/PR) in `project-elmo-ui`: https://github.com/adobe/project-elmo-ui/pull/2385 — the same self-serve onboarding flow also dropped any subpath (e.g. `guides.adobe.com/knowledge-ai` → `guides.adobe.com`) before sending it to this endpoint, and now owns the localized "please use our onboarding guide" copy + link that used to live in this backend's error string.

### DIFF
--- a/src/controllers/llmo/llmo.js
+++ b/src/controllers/llmo/llmo.js
@@ -1090,15 +1090,16 @@ function LlmoController(ctx) {
     const { spaceCatId } = context.params;
     const { data } = context;
 
-    // Customers never see the internal reason - only a generic failure. Kept
-    // ASCII-only (no em dash / curly quotes): this string is copied verbatim into
-    // the `x-error` response header by badRequest()/internalServerError(), and a
-    // character outside the header-safe range (see cleanupHeaderValue) crashes the
-    // response with a 500 "Invalid character in header content" instead of the
-    // intended 400/500 body.
-    const GENERIC_ONBOARD_ERROR = "We couldn't onboard this domain right now. Please use our "
-      + 'domain onboarding guide instead: https://experienceleague.adobe.com/en/docs/'
-      + 'llm-optimizer/using/essentials/quick-start#step-1-onboard-your-domain';
+    // Customers never see the internal reason - only a generic failure. The
+    // project-elmo-ui client renders its own localized copy (with a link to the
+    // onboarding guide) instead of this raw text - this string only reaches a
+    // caller that skips the UI (e.g. a direct API/curl consumer), so it stays a
+    // short, plain fallback. Kept ASCII-only (no em dash / curly quotes): it's
+    // copied verbatim into the `x-error` response header by
+    // badRequest()/internalServerError(), and a character outside the
+    // header-safe range (see cleanupHeaderValue) crashes the response with a 500
+    // "Invalid character in header content" instead of the intended 400/500 body.
+    const GENERIC_ONBOARD_ERROR = "We couldn't onboard this domain right now. Please contact support.";
 
     try {
       // --- Resolve org (404 if missing) ---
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @igor-grubic,

**Verdict: Approve** - clean, well-scoped bugfix with defense-in-depth and regression coverage.
_Changes: Fixes a 500 crash caused by an em dash in the `x-error` response header by rewriting the error message to ASCII-only and wrapping with `cleanupHeaderValue()` (2 files)._

<details>

- nit: The `HEADER_SAFE` regex validates the test file's own `GENERIC_ERROR` constant rather than a value produced at runtime by the controller; since `badRequest`/`internalServerError` are stubbed, the real header-writing path is never exercised. The PR comments acknowledge this ("a failure this mocked http-utils test can't otherwise catch"), so the assertion is a reasonable best-effort guard given the unit-test constraints. - `test/controllers/llmo/onboard-site-only.test.js:34`
- suggestion: Apply `cleanupHeaderValue` once at the constant definition site (`const GENERIC_ONBOARD_ERROR = cleanupHeaderValue(...)`) rather than at each call site. This guarantees the invariant at declaration and avoids relying on every consumer remembering to wrap. It also removes the runtime no-op on a string that is already ASCII-safe. - `src/controllers/llmo/llmo.js:1098`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 28s | <b>Cost:</b> $4.33 | <b>Commit:</b> <code>2c5571f6993361fee6da6f4ee4c0dcd3ebdafdb5</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2788, opened 2026-07-10
feat(serenity): resolve tag dimensions from the tag tree, not a name prefix

### LINKED ISSUE #2733: serenity: nested category tags (parent_id) — transport, create-tag parent, hierarchy-preserving read-merge
## Summary

Implement **1-level nested category tags** on the Serenity (Semrush AIO) proxy: a topic becomes a bare child tag `<NAME>` with a `parent_id` pointing at its `category:<NAME>` root tag, instead of a flat `topic:<NAME>` tag. This is the spacecat-api-service slice of the cross-repo design.

**Parent design (authoritative):** https://github.com/adobe/serenity-docs/issues/21 (see §1 tag model, §5.2 this repo's change list, §7 open live-verification questions).

## Target model

- **Root category** → tag name `category:<NAME>`, no `parent_id`.
- **Child (sub-category / migrated topic)** → bare tag name `<NAME>`, `parent_id` = the root `category:` tag's id.
- Tag identity is by **id** (or `parent_id`+name), never bare name — two categories may each own a child named `Pricing`.

## Preconditions

- Builds on https://github.com/adobe/spacecat-api-service/pull/2730 (create-tag endpoint + `CREATABLE_TAG_DIMENSIONS` + `tagFor`). #2730 is **not** on `main` yet — rebase/merge it first. File anchors below are on its branch `origin/feat/complete_test`.
- Resolve the generic-tags design decision in https://github.com/adobe/spacecat-api-service/issues/2731 together with this work (flat `dimension:value` strings must coexist with the `parent_id` tree on the same `aio/tags` surface).
- Local nested e2e depends on the mock learning nesting: https://github.com/adobe/spacecat-shared/issues/1754 companion issue.

## Change list

Transport — `src/support/serenity/rest-transport.js`:
- [ ] `createProjectTags(ws, pid, names)` (~`:398-414`): thread an optional `parentId` → body `{ names, ...(parentId ? { parent_id: parentId } : {}) }`.
- [ ] Add `updateProjectTag(ws, pid, tagId, {name, parentId})` for `PATCH /aio/tags/{tag_id}` (`model.TreeNodeRequest`) — re-parent/rename/promote; no wrapper exists today.
- [ ] `listProjectTags(...)` (added by #2730) hardcodes `parent_id: ''` (roots only) — accept a `parentId` arg so children are readable.

Composition — `src/support/serenity/prompt-tags.js` + `handlers/markets-subworkspace.js`:
- [ ] Child topics must be bare `<NAME>` + `parent_id`, not `topic:<NAME>`. The nested-create path must **skip `tagFor` prefixing** for children. Update `tagFor`/`topicTag` usage (`prompt-tags.js:63-71`) and `generateAndAttachPrompts` (`markets-subworkspace.js:200`). NB: this changes every generated prompt's tag set and anything filtering by `topic:`.

Create-tag handler — `src/support/serenity/handlers/tags.js:49-187`:
- [ ] `parseCreateTagBody` + `handleCreateTag[Subworkspace]`: add a `parent` input (category name or id); resolve-or-create the parent → create the child with `parent_id`; skip `tagFor` on the child path (`:140`,`:176`); guard the create-or-get race on the parent.

Read/merge — preserve hierarchy:
- [ ] `handlers/markets-subworkspace.js:586-625` and `handlers/markets.js:568-641` collapse to `{id,name}`, drop `parent_id`/`path`, and **dedup by name**. Carry `parent_id`/`path`/`children_count` through the merge and **dedup by id**.

Controller + contract:
- [ ] `src/controllers/serenity.js`: plumb `parent` and any list-by-parent / update-tag endpoints.
- [ ] `docs/openapi/serenity-api.yaml`: add `parent`/`parentId` to the create-tag request; document the nested response; update the `type` enum.
- [ ] Coordinate with `PROJECT_STANDARD_TAGS` (`prompt-tags.js:100-104`), already flagged out-of-sync in https://github.com/adobe/spacecat-api-service/issues/2708 .

Tests:
- [ ] Unit + IT for nested create/list/update and the read-merge dedup-by-id; note the Counterfact mock is flat until https://github.com/adobe/spacecat-shared/issues/1754 lands, so nested IT can't be trusted before then.

## Live verification (blocking a correct impl — see parent §7)

- [ ] Confirm a prompt can be tagged with a **child by id** (or `parent_id`+name) via `POST /aio/prompts/tagged` — a bare name is ambiguous across per-parent duplicates.
- [ ] Confirm nested filter semantics (multiple `tagIds` OR vs AND) and whether Semrush hard-caps at 1 level.

rainer-friederich:
## 1. Abstract
A tag's dimension on the Serenity proxy is the bare-named root it descends from — `path[0]` — rather than a `<dimension>:` prefix parsed off its name. The four dimension roots (`category`, `intent`, `source`, `type`) are resolved or created on demand, every tag value is a bare-named descendant carrying `parent_id`, and the client dependency moves to the release that models this upstream.

## 2. Reasoning
The Brand Presence dashboard is a Semrush-owned micro-frontend whose tag filter is one tree-aware dropdown, ordered by the upstream rather than alphabetically. A name prefix produces no visual grouping there — a customer sees `category:Premium Modular Furniture`, `intent:Commercial`, `source:human` and `type:non-branded` interleaved in a single flat list. The dropdown does render parent/child hierarchy, so making each dimension a root tag turns that list into four collapsible groups with no change to the micro-frontend. The tag tree is the only lever we own.

Encoding the dimension in the name also made the name carry meaning that the tree already carried, and the two could disagree. A tag renamed upstream lost its dimension; a customer category containing a colon was indistinguishable from a namespaced one; and two sub-categories sharing a bare name under different parents could not coexist. Reading the dimension off the tree removes the ambiguity at the source rather than defending against it at every parse site.

The proxy is where that surfaces first, because it is the only component that both writes the taxonomy and serves it to the UI.

**Dimension resolution.** A tag's dimension is `path[0]`, the root of its ancestry, read from the tree. Every prefix parser is gone, and no tag name contains a `:`. `src/support/serenity/tag-tree.js` owns the resolve-or-create discipline against the live tree: it indexes one level by name, ensures a set of children under a parent, ensures the four dimension roots exist, and provisions the closed vocabularies beneath them.

**Reading prompts is one upstream call again.** Upstream embeds the full tag object on each prompt, byte-identical to what the tag tree returns for that id — a dimension root omits `parent_id` and `path`; a descendant carries both. The handler now reads each tag's parentage straight off the prompt. A per-request tree walk that had been added to reconstruct that parentage is deleted: it rested on the belief that a prompt-embedded tag was a narrower shape than a tree-read one, which came from sampling only root tags. Listing prompts therefore costs one upstream call rather than one plus a call per tag level.

**Closed dimensions are server-owned.** `intent`, `source` and `type` resolve-or-create their value under the owning root and reject a caller-supplied parent — a client that could choose the parent could file `ai` under `intent`. The response reports whether the value was created and names the dimension root as the parent, never `null`. An open `category` may still be nested by the caller.

**Re-parenting is explicit.** Upstream promotes a tag to a root whenever a `PATCH` omits `parent_id`, which would strand it outside its dimension. A `PATCH` that changes only the name therefore re-sends the tag's current parent. An explicit `null` parent is rejected, as is patching a dimension root; an unknown tag id is a 404.

**Two defects the migration surfaced, fixed here.** The subworkspace tag listing merged on a guard that could never fire — the transport already back-fills a tag's id from its name when the id is missing, so a name-shaped placeholder survived alongside the canonical id and the same category was returned twice. And the `created` flag documented on closed-dimension tag creates was being silently dropped.

**Onboarding-generated prompts arrive uncategorized.** The topic-generation service returns bare topic names with no parent, so a prompt seeded automatically at brand onboarding is written without a category under the dimension-root model. Categorizing generated topics is tracked in serenity-docs#44 and gated before prod (section 8); until it lands, generated prompts are uncategorized rather than mis-categorized.

**User-visible surface.** A category's name is now what the customer typed, with no `category:` decoration, in both the API and the Brand Presence dropdown, where the four dimensions render as collapsible groups.

The client dependency moves to the release that models this upstream, which is also what lets the integration suite pull the matching mock image.

Closes https://github.com/adobe/spacecat-api-service/issues/2733

## 4. Required information
- Jira / issue: https://github.com/adobe/spacecat-api-service/issues/2733 — the issue predates the dimension-root model and describes the superseded shape (`category:<NAME>` roots, flat `topic:<NAME>` tags). This delivers the capability it asks for — nested category tags on the Serenity proxy — under the model that replaced it: bare-named dimension roots, with a category as a descendant of the `category` root.
- Spec: https://github.com/adobe/serenity-docs/pull/39 — dimension-root tag model (open; the design of record for this program, not yet merged)
- Implementation plan: https://github.com/adobe/serenity-docs/pull/39 — dimension-root tags implementation plan, WP2 (same PR)
- ADR(s): https://github.com/adobe/spacecat-api-service/blob/main/docs/decisions/005-opt-in-type-checking.md — the `// @ts-check` surface this change extends
- Other: https://github.com/adobe/spacecat-shared/pull/1799 — WP1, merged; it released the client and the mock image this depends on

- **spacecat-shared** — consumed. Depends on `@adobe/spacecat-shared-project-engine-client` at the version whose GHCR mock image carries the same tag; the integration harness resolves that image tag from the installed client version with no fallback, so the two move together.
- **project-elmo-ui** — contract, **and this breaks it**. Elmo still composes a `category:<Name>` tag name, which is now rejected because names must be bare, and its re-parent branch still sends an explicit `null` parent, which is now a 400. Its category-rename dialog is shipped and localized, so the break is user-visible. The regression was weighed and accepted; WP3 must land at or before this deploys.
- **mysticat-data-service** — contract. Its migration CLI writes and reshapes this same taxonomy directly against live Semrush, and must always re-send `parent_id` on a `PATCH` for the same reason this proxy now does.

Booted the published mock image directly, over its real HTTPS surface rather than through the test harness, and probed the three behaviours this change depends on:

- A dimension root comes back without `parent_id` and without `path` — the keys are absent, not `null`. This is the shape the read path now relies on to tell a root from a descendant.
- A tag embedded on a prompt is byte-identical to the same tag id read from the tag tree. This is the invariant that let the per-request tree walk be deleted.
- The name-keyed prompt-create endpoint mints and registers a root for an unknown name, reuses an existing root rather than duplicating it, and the derived tag ids match the values the client pins as a cross-repo contract.

**The integration suite was not run on this machine.** Its compose file binds `127.0.0.1:8443` and `:8444`, and a pre-existing local-dev stack already held both ports; freeing them would have meant stopping containers this session did not create. CI's integration job runs that same suite against the same mock image and is the gate for it.

**A surface outside this PR reads the old prefixed names, and the spec does not account for it.** `src/support/elements/**` filters and parses Semrush's Elements API on prefixed tag values: it extracts filter dimensions by stripping a `topic:` / `category:` / `intent:` prefix, and pushes a category filter to Semrush as the literal tag `category:<label>`. Both are wired to shipped routes. If the Elements tag values are the same taxonomy the tag tree writes — the documentation for that surface says they are, and the category filter is recorded there as verified against live — then reshaping tags to bare names makes those filter-dimension lists come back empty and that category filter match nothing. Nothing in this PR touches that code, and I have not probed the Elements API to confirm the two share a taxonomy, so this is a risk to settle before the reshape runs, not a claim. It is called out here rather than folded into this change.

## 7. Test plan
**Verified locally, beyond the automated suites.** Probed the published mock image over HTTPS as described in section 6, exercising the root/descendant shapes, the prompt-embedded tag identity, and the tag-id derivation the fixtures rely on. The integration suite could not run locally for the port reason given above.

**Verification by environment.**

1. **dev** — after deploy, create a category through the API and confirm the returned tag carries no `category:` decoration and names the `category` dimension root as its parent. Rename it without supplying a parent and confirm it stays under that root rather than being promoted to a top-level tag; this is the failure the re-send guards against, and it is silent when it happens.
2. **dev** — create a closed-dimension value (an `intent`) and confirm the response reports whether it was created and names the `intent` root as the parent. Attempt to supply a parent for it and confirm the request is rejected.
3. **dev** — list prompts for a brand whose tags include a depth-3 sub-category, and confirm each returned tag carries the parentage needed to group it, without the proxy making a second upstream call to reconstruct it.
4. **dev** — open the Brand Presence dropdown for that brand and confirm the four dimensions render as collapsible groups rather than one interleaved list.
5. **stage / prod** — do not deploy ahead of WP3. Elmo's category rename and re-parent both send shapes this rejects, so the UI surfaces a 400 on a shipped, localized dialog until that lands.

- **depends-on** — https://github.com/adobe/spacecat-shared/pull/1799 (WP1). Merged and released; it published the client and the matching GHCR mock image, without which this PR's dependency bump would fail the integration job's image pull. This is satisfied.
- **blocks** — project-elmo-ui (WP3). It must land at or before this deploys, because elmo sends both a prefixed tag name and an explicit null parent, and this rejects both.
- **blocks** — mysticat-data-service (WP4), the migration CLI that reshapes the live taxonomy.
- **related** — https://github.com/adobe/serenity-docs/pull/39, the model spec and implementation plan. It is the design of record and is still open; this PR does not wait on it.

Ordered sequence:

1. This PR merges and deploys to dev.
2. WP3 (project-elmo-ui) lands. It must be deployed no later than this change reaches an environment real users touch.
3. WP4 (mysticat-data-service) reshapes the live taxonomy, once the risk to the Elements read path in section 6 has been settled.
4. serenity-docs#44 (categorization of onboarding-generated prompts) lands, or its absence is explicitly waived by product/CS — generated prompts arrive uncategorized until then.
5. Promotion to stage and prod follows WP3 and step 4.


rainer-friederich:
Checked the branch: the server-side guard is already implemented and tested, so nothing to add here.

`PATCH /serenity/tags/:tagId` refuses a bad re-parent through `buildUpdatePayload` + `assertParentPlacement` (`src/support/serenity/handlers/tags.js`, `src/support/serenity/tag-tree.js`): a `parentId` equal to the tag itself (400 "must not be the tag itself"), a `parentId` that is a descendant of the tag (400 "must not be a descendant of the tag", tested by ancestry so a nested parent can't slip through), a cross-dimension parent, an unresolvable parent, and edits to a dimension root or a closed-dimension value. Covered by `test/support/serenity/handlers/tags.test.js` and `test/support/serenity/tag-tree.test.js`.

So the unrecoverable-subtree edge the elmo review flagged is closed on the write path already - the client-side two-level picker never needed to, and doesn't, guard it. Leaving this note for the record; no change requested.

rainer-friederich:
Ran a local `/pr-review` (5 specialists: staff-engineer, architect, conventions, QA, security) because the diff (~5.6k LOC) exceeds MysticatBot's budget — the bot pipeline crashed with `error_max_budget_usd`. Findings triaged and the actionable ones pushed in `43f41544`.

## Fixed in `43f41544`
- **Docs sync (`docs/serenity.md`)** — the `POST`/`PATCH /serenity/prompts` endpoint rows and the dev smoke-test `curl` still documented the retired name-based `tags` write path; a reader copy-pasting either got a 400. Updated both to the `tagIds`-required / `tags`-rejected contract, and the smoke test now resolves a tag id from `POST /serenity/tags` first.
- **`handleListTagsSubworkspace` placeholder collapse (`markets-subworkspace.js`)** — the `synthetic` bucket keys id-less name-shaped placeholders by name. On inspection, two *id-less* same-named tags are genuinely indistinguishable here (no id to tell a `category` `human` from a `source` `human`), and keying by `(name, source)` would only emit duplicate `{id:name}` rows. Documented the collapse as intentional (the by-id merge is what preserves same-named tags whenever either carries a real id — the common case) and added a test pinning the behaviour.
- **Onboarding-generated prompts uncategorized (PR description)** — added serenity-docs#44 as an explicit prod-promotion gate in section 8 and a note in section 3, so the uncategorized-onboarding window is either closed by #44 or explicitly waived before prod, rather than shipping silently.

## Needs a decision (not pushed) — prompt `tagIds` ownership check
`src/support/serenity/handlers/prompts.js` runs caller `tagIds` only through `sanitizeTagIds` (format/non-empty) before forwarding to the atomic upstream write, whereas tag create/update now resolve every id against the caller's own project tree via `findTagsInTree`. This PR makes `tagIds` the mandatory, universal path, so if upstream tag-id resolution is not itself project-scoped, a caller who obtains another project's tag UUID could attach it (a cross-tenant existence oracle on the 2xx-vs-500 response).

Two ways to close it, and they trade off differently:
- **Add the guard** — resolve each `tagIds` entry against the market's project tree and 400 an unresolved id. Consistent with the tag handlers, but adds a tree walk to the hottest write path (the same tree-walk cost the architect/staff reviews already flagged).
- **Verify + document** — probe whether upstream already scopes tag ids to the project; if it does, the atomic write is the control and we document that instead of paying for a redundant walk.

I did not pick one blindly — the right call depends on upstream behaviour I can't verify from the diff. Which do you want?

## Non-blocking (triaged, not pushed)
Dangling `@link resolveTagTarget` ×2 (renamed to `resolveUpdateTargets`) in `tags.js`; concurrent-reparent TOCTOU in `assertParentPlacement` (admin-frequency, worth a comment); `resolveTypeValueInjection` read-doubling on the write hot path (memoized, budget note); `updateProjectTag` `TypeError` vs `ErrorWithStatusCode`; no NFKC/confusable normalization on reserved dimension-root names (spoofing, no priv escalation); case-sensitive closed-value enum; `MAX_TREE_READS` per-request-not-per-tenant; a few test-assertion strengthenings in `tags.test.js`; and a tracking issue for eventual `tagMap` removal. Happy to fold any of these in — say which.

The dependency bump (`@adobe/spacecat-shared-project-engine-client` → 1.10.0) was verified end-to-end: version matches the merged WP1 release, `package-lock` sha512 matches the registry, SLSA provenance and signing key intact — no finding.

`src/support/elements/**` still parses/writes `dimension:`-prefixed tag names and is untouched here — once the live reshape (WP4) makes names bare, those filter lists come back empty. Correctly flagged in section 6; must stay on the WP4 checklist.

rainer-friederich:
Follow-up on the local `/pr-review` findings — two items resolved.

**1. Deferred finding: prompt `tagIds` ownership on the write path.**

The review asked whether `POST`/`PATCH /serenity/prompts` should walk the tag tree (a `findTagsInTree` guard) to prove every supplied `tagId` belongs to the target project before writing, to stop a caller attaching a foreign or cross-workspace tag to a prompt.

I probed the live Semrush API (LLMO-Dev-2 workspace) to see how the upstream write actually behaves with a foreign id. Using a real, published tag id from Brand-A US/en, attaching it to a draft prompt in:

- a **different project of the same workspace** (Brand-A CH/de) → `500`, atomic reject, no prompt created
- a **different workspace** (Brand-B) → `500`, atomic reject, no prompt created
- **control** — a garbage UUID into the same project → `500`, identical reject

A real tag id from a sibling project is treated **identically to a nonexistent UUID**: the id space is scoped to the target project, and the write is rejected atomically. So the ownership property the finding is worried about is already enforced by Semrush — a foreign/cross-workspace tag cannot be attached to a prompt through this path.

Decision: no `findTagsInTree` guard added. It would duplicate an upstream invariant on every prompt write (an extra tree walk per call) to guard a state the upstream already refuses. The one wrinkle is that Semrush signals it with an opaque `500` rather than a clean `400`; in the normal api-service flow `tagIds` are minted/resolved against the same project's tree (via `POST /serenity/tags`), so a foreign id does not arise in practice. If we later want a friendlier client error, that is a separable enhancement, not a correctness gap here.

**2. Patch coverage.**

`codecov/patch` flagged two partial branches on the closed-value create paths (`handleCreateTag`, `handleCreateTagSubworkspace`): `parentId: rootId ?? null`. `ensureClosedValue` always returns the dimension root id (non-null string, guaranteed by `ensureDimensionRoots`), so a minted closed-dimension value can never sit at the top of the tree — the `?? null` fallback was unreachable. Rather than write a test for a state that cannot occur, I removed the dead fallback (`parentId: rootId`), which is behaviour-preserving and closes the gap. tags.js is now 100% on the patched lines.

vivesing (approved):

PR #2789, opened 2026-07-10
fix(serenity): match regular plurals and possessives in the branded classifier

rainer-friederich:
## 1. Abstract
The server-side Serenity branded/non-branded prompt classifier now recognises the regular English plural and the possessive form of a brand name or alias. It remains a word-boundary matcher, not a substring match.

## 2. Reasoning
The classifier matched brand needles as whole tokens only, so any inflection of the brand name fell through to `type:non-branded`. `Buick` classified as branded; `Buicks` did not — even though "Are Buicks luxury cars?" is plainly a branded prompt.

Investigating that turned up a second, quieter defect in the same code. The alias `Kellogg's` normalizes to the two-token needle `kellogg s`, because the apostrophe collapses into a token gap. Such a needle matches a literal possessive and nothing else, so it never matched the bare word "Kellogg". That is why the WK Kellogg Co brand had only 14 of its 311 prompts tagged branded, while 22 of them name the brand.

The obvious fix — relaxing the match to a `contains` test — is not acceptable: it would classify "outer space" as a mention of the brand `Ace`. The fix had to stay morphological.

The insight that makes a narrow fix possible is that **proper nouns only ever take the regular English plural**. One writes "the Kennedys", never "*the Kennedies"; brand names never undergo the irregular stem changes (`-ves`, `-ies`) that common nouns do. So the matcher can admit a small, closed set of suffixes and remain principled.

Behaviour delta:

- A needle now matches when it is followed by the regular plural suffix on its **final token only**: `-s` in every case, and `-es` only after a sibilant stem (`s`, `x`, `z`, `ch`, `sh`). `Lexuses` and `Bosches` therefore match, while the non-word `Buickes` does not.
- The possessive clitic `'s` (straight and typographic) is stripped from **both** the needle and the prompt text before matching. This makes `Buick's` classify, and it simultaneously repairs the `Kellogg's` alias so it matches the bare brand name.
- A plural possessive (`Buicks'`) leaves a bare trailing apostrophe, which the plural rule already covers.
- Multi-word needles pluralize only their last token: `Le Creusets` matches, `les creuset` does not.

What deliberately did **not** change is the left edge of the match. It stays anchored, and no suffix other than the two above is admitted. `Ace` still does not match `surface`, `spaces`, or `outer space`, and `Buick` still does not match `Buickmobile` or `Buicking` — but `Aces` and `Buicks` now do.

Operationally visible effect: prompts written from now on through any Serenity write path (manual create/edit, CSV import, AI generation, onboarding seeding) get the corrected tag. Because these paths all share this single classifier, the behaviour is uniform across them.

**Reclassification is out of scope here.** Per serenity-docs issue 31 the classifier is forward-only — existing prompts are not reclassified — and this PR takes the migrated data as it stands. A backfill pass over already-migrated prompts would be entirely possible and is a sensible follow-up: it would flip roughly 43 plural-form prompts across the General Motors brands, plus about 8 bare-"Kellogg" prompts in WK Kellogg Co, from non-branded to branded. Nothing would flip the other way.

## 4. Required information
- Spec: https://github.com/adobe/serenity-docs/issues/31

- `project-elmo-ui` — consumed / contract. It renders the branded vs non-branded tag (read-only) and derives the branded-ratio prompt-health statistics from it. No UI change is required; the tag values and their shape are unchanged, only which prompts receive which tag.
- `serenity-docs` — contract. Issue 31 defines the classifier as a single server-owned layer, forward-only. This change keeps both properties.

The rule was designed against real migrated data rather than invented examples. Using the read-only prod Postgres MCP, the classifier's matching logic was replayed over the actual prompt text and the actual `brand_aliases` rows for all 13 brands in the Lovesac, Kellogg and General Motors workspaces — 5,002 prompts in total.

- Every prompt the old rule missed was a regular plural: `Buicks`, `Cadillacs`, `Chevys`, `GMCs`. There were 43 of them, and each was read individually; each genuinely names its brand.
- The new rule flips exactly those 43 to branded and flips **zero** prompts in the wrong direction across the whole corpus.
- An independent check corroborates the rule. The Lovesac workspace showed zero misses, but only because someone had hand-added `Sacs`, `Sactionals`, `Supersacs` and `pillowsacs` as aliases to work around the missing plural rule. Dropping those four workaround aliases and applying the plural rule instead reproduces Lovesac's branded set exactly — 591 prompts before, 591 after, nothing lost and nothing gained. The rule derives what a human had been maintaining by hand.

One caveat on the evidence: this replay reads the prompt corpus and alias lists from Postgres. The `type:branded` tags themselves live upstream in the Semrush sub-workspaces and were not read back.

## 7. Test plan
Locally, beyond the unit tests, the verification that mattered was the corpus replay described in section 6 — the matching logic exercised against the real prod prompt text and alias rows for the three migrated workspaces, with every flipped prompt inspected by hand.

To verify on **dev**:

1. Pick a Serenity brand with a single-word name (or add an alias such as `Buick`).
2. Create a prompt through the normal write path whose text uses the plural, e.g. "Are Buicks luxury cars?", and confirm the stored prompt carries `type:branded`.
3. Create one using the possessive, "What is Buick's best selling vehicle?", and confirm `type:branded`.
4. Confirm the guard still holds: with an alias of `Ace`, a prompt reading "tell me about outer space" must remain `type:non-branded`.
5. For an alias carrying a possessive (`Kellogg's`), confirm a prompt naming the bare brand ("how does kellogg ensure cereal safety?") now classifies as `type:branded`.

The same five checks apply on **stage** and **prod** after deploy. No migration or backfill runs as part of this change, so existing prompt tags are expected to be untouched — that is the forward-only contract, and seeing old prompts keep their old tags is the correct outcome, not a regression.

### DIFF
--- a/src/support/serenity/branded-classifier.js
+++ b/src/support/serenity/branded-classifier.js
@@ -20,14 +20,24 @@ import { collectAliasNames } from './brand-aliases.js';
  * on every path that writes a Serenity (Semrush sub-workspace) prompt: manual
  * create / edit, CSV import (which reuses the create path), AI generation, and
  * onboarding/provisioning seeding. A prompt is `type:branded` when its text
- * mentions the brand name or an applicable alias as a WHOLE WORD (diacritics
- * folded); otherwise `type:non-branded`.
+ * mentions the brand name or an applicable alias as a WHOLE WORD — allowing the
+ * regular English plural of that word (diacritics folded); otherwise
+ * `type:non-branded`.
  *
  * This replaces the earlier AI-only substring matcher (`brandedTypeTag`): the
  * match is now word-boundary + diacritic-folding, and it is shared everywhere so
  * the classification is identical no matter how the prompt got in. The behaviour
  * change (substring → whole-word) is forward-only — existing prompts are not
  * reclassified (serenity-docs#31, decisions 1/2/6).
+ *
+ * Plurals and possessives are handled morphologically, NOT by relaxing the match
+ * to a substring test. Proper nouns only ever take the regular plural in English
+ * — `the Buicks`, `the Kennedys`, `the Bosches` — never the irregular stem
+ * changes that common nouns take (no `Buickies`, no `Bosch → *Bosches` via
+ * `-ves`/`-ies`). So the classifier admits exactly two suffixes on the needle's
+ * final token, `-s` and (after a sibilant) `-es`, and strips the possessive
+ * clitic `'s` from BOTH sides before matching. `Ace` therefore still does not
+ * match `surface` or `spaces`, but does match `Aces`.
  */
 
 // Latin letters that carry a diacritic/ligature but have NO canonical NFD
@@ -42,12 +52,32 @@ const NON_DECOMPOSING_FOLDS = Object.freeze({
 const NON_DECOMPOSING_RE = new RegExp(`[${Object.keys(NON_DECOMPOSING_FOLDS).join('')}]`, 'g');
 
 /**
+ * The English possessive clitic: an apostrophe (straight or typographic) + `s`,
+ * at a token end. Dropped from both sides so `Kellogg's` ≡ `Kellogg` — otherwise
+ * the alias `Kellogg's` normalizes to the two-token needle `kellogg s`, which
+ * matches only a literal possessive and never the bare brand name. Runs after
+ * `toLowerCase` (so only lower-case `s`) and before punctuation collapse (which
+ * would otherwise turn the apostrophe into a token gap). The bare trailing
+ * apostrophe of a plural possessive (`Buicks'`) is left to the plural rule.
+ */
+const POSSESSIVE_RE = /['’]s(?=[^\p{L}\p{N}]|$)/gu;
+
+/**
+ * Needle endings after which the regular English plural is spelled `-es` rather
+ * than `-s` (the sibilants): `Lexus → Lexuses`, `Bosch → Bosches`, `Fox → Foxes`.
+ * Restricting `-es` to these keeps the suffix set from admitting non-words like
+ * `Buickes` as a match.
+ */
+const SIBILANT_RE = /(?:[sxz]|ch|sh)$/;
+
+/**
  * Normalizes one side of the match (needle OR haystack) into a canonical,
  * space-delimited token stream:
  *   - fold diacritics (NFD decomposition, strip combining marks) so `café` ≈ `cafe`;
  *   - lower-case (case-insensitive match);
  *   - fold the non-decomposing accented/ligature letters (`ø→o`, `æ→ae`, `ß→ss`,
  *     …) that the NFD strip leaves intact, so `Ørsted` ≈ `Orsted`;
+ *   - drop the possessive clitic `'s`, so `Kellogg's` ≡ `Kellogg`;
  *   - replace every run of non-alphanumeric characters (Unicode letters/numbers)
  *     with a single space, so punctuation/whitespace collapses to token gaps;
  *   - trim.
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-scoped morphological fix with thorough tests and no regressions to the word-boundary invariant.
_Changes: Adds regular English plural (-s, -es after sibilants) and possessive ('s) recognition to the branded-prompt classifier without relaxing to substring matching (2 files)._

<details>

- suggestion: Add a `z`-ending sibilant test (e.g., `Benz` -> `Benzes`) to complete coverage of all `SIBILANT_RE` alternations (`[sxz]|ch|sh`) - tests cover `s`, `x`, `ch`, `sh` but not `z` - `test/support/serenity/branded-classifier.test.js`
- suggestion: Consider pre-computing plural forms at needle-construction time (in `brandNeedles`) rather than testing suffixes at match time - this separates morphology from boundary semantics and makes adding future suffix classes trivial without touching match logic - `src/support/serenity/branded-classifier.js:152`
- nit: A test for a sibilant + possessive combo (e.g., needle `"Ross's"` normalizing to `ross`, prompt `"are Rosses near you?"` -> branded) would document the interaction of both new code paths in one case - `test/support/serenity/branded-classifier.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 14s | <b>Cost:</b> $4.16 | <b>Commit:</b> <code>8cc2c464bee4caf8d9ffc0cea3b00de2295a236b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the review in `9fe491c2`.

MysticatBot:

- suggestion: add a `z`-ending sibilant test (`Benz` → `Benzes`) — **done**. There is now one case per `SIBILANT_RE` alternation: `Lexus`/`Fox`/`Benz`/`Bosch`/`Bush` for `s`/`x`/`z`/`ch`/`sh`. Added a negative alongside them (a sibilant stem where neither the bare form nor a plural is present stays non-branded), which also closes the last uncovered branch in the matcher.
- nit: a sibilant + possessive combo test — **done**. `Ross's` normalizes to the needle `ross`, whose stem is sibilant, so `Rosses` matches through both new code paths in one case. Covered the bare and possessive spellings in the same test.
- suggestion: pre-compute plural forms in `brandNeedles` rather than testing suffixes at match time — **declined**, on two grounds.

  It would not actually consolidate the morphology. The possessive clitic has to be stripped from the *prompt text* as well as the needle, so it necessarily lives in `normalizeMatch`, which both sides share. Moving the plural into needle construction would put the right-edge morphology in one place and the possessive morphology in another, leaving the two halves further apart than they are today, where the boundary rule reads as one function.

  It would also weaken the contract of the needle list. `brandNeedles` is documented as "the brand display name plus its region-applicable aliases, normalized", and `needlesFromNames` as normalize-and-dedupe over names. Injecting fabricated forms like `buicks` — which is not an alias of anything — makes those lists no longer a set of brand names, which matters because they are memoized per market and are the natural thing to log when a classification is disputed.

  On the stated benefit, cheaply adding future suffix classes: that is the one extension we specifically do not want. The rule is narrow by design — proper nouns take only the regular plural, and it is exactly that closed set which keeps this a word-boundary test rather than a substring test. Making it easy to bolt on more suffix classes optimises for a change we would reject on review.

  Happy to revisit if a second consumer of the needle list ever wants pre-expanded forms.

PR #2791, opened 2026-07-10
feat: enhance serenity filter-dimensions for sub categories

vivesing:
# feat: enhance serenity filter-dimensions for sub categories

## Summary

Enhances the Serenity (Semrush Elements-backed) URL Inspector filter-dimensions
response (`getUrlInspectorFilterDimensions` / `src/support/elements/definitions/topics.js`,
`brands.js`) to handle tag hierarchies, surface previously-unmodeled tag types, and
give every filter-dimension item a stable, real `id` instead of `null`.

### 1. Category/topic subcategory hierarchy (`Parent__Child` tags)

- Semrush's Elements API can return tag values encoding a two-level hierarchy, e.g.
  `category:Modular & Configurable Sofas__Compact Shippable Furniture`.
- `transformCategoriesToFilterDimensions` / `transformTopicsForFilterDimensions`
  (`src/support/elements/definitions/topics.js`) split on the first `__`: the
  suffix becomes `label`, and `parent_id` / `parent_label` are added for the prefix.
- Entries without `__` are unchanged (no parent keys added).

### 2. Generic handling of unknown tag types

- The tags element can also return types beyond `topic:` / `category:` / `intent:` /
  `source:` (e.g. `type:branded`), plus plain untyped tags (e.g. `abc`).
- New `transformOtherTagsForFilterDimensions` groups unknown `prefix:value` tags
  dynamically by their prefix (response key = prefix, e.g. `type: [...]`), and
  collects plain, prefix-less values into a generic `tags` array. Tags whose
  prefix collides with an existing response key (`brands`, `regions`, `topics`,
  `categories`, `page_intents`, `origins`, `tags`) are routed into the generic
  `tags` array instead of clobbering that key.
- Wired into `elements-service.js`'s `getUrlInspectorFilterDimensions`, which now
  merges these dynamic groups + `tags` into its response alongside the existing
  fixed keys (`brands`, `regions`, `topics`, `categories`, `page_intents`, `origins`),
  with collision-safe merging.

### 3. `id` is now the original, unmodified tag value (not `null`)

- Previously `topics`, `categories`, and the generic `tags`/dynamic-group items
  always had `id: null`; `page_intents` used an uppercased label; `origins` used
  the stripped label.
- All of these now set `id` to the **original tag value exactly as returned by
  Semrush**, including its prefix (e.g. `topic:Furniture__Sofas`,
  `intent:Informational`, `source:organic`). `label` is unchanged (the
  prefix-stripped, non-uppercased value).

### 4. `parent_id` is now derived, not `null`

- For hierarchical (`Parent__Child`) tags, `parent_id` is reconstructed as
  `${prefix}${parent_label}` — the same original-tag shape as the item's own
  `id` — instead of always being `null`.
- Example: `category:Living Room Furniture Retail__Living Room Furniture and Sofas`
  → `parent_id: 'category:Living Room Furniture Retail'`.
- Applies uniformly to the fixed dimensions (`topics`, `categories`) and to the
  dynamic/generic tag groups from `transformOtherTagsForFilterDimensions`.

### 5. Brand `id` now mirrors `label`

- `transformBrandsToFilterDimensions` (`src/support/elements/definitions/brands.js`)
  now sets `id: label` instead of `id: null`, since Semrush has no stable brand ID
  and the label is the only distinguishing value available.

## Docs

- `docs/openapi/schemas.yaml`: `SerenityFilterDimensionItem.id` and `.parent_id`
  are now `string` (no longer nullable); `SerenityUrlInspectorFilterDimensions.brands[].id`
  is now `string` (no longer nullable). Descriptions updated to explain the
  original-tag-value semantics and the `parent_id` reconstruction rule.
- Validated with `npm run docs:lint`.

## Tests

- Full coverage added/updated in `topics.test.js`, `brands.test.js`,
  `elements-service.test.js`, `controllers/elements.test.js`, and
  `openapi-contract/serenity-api.test.js` for all of the above behaviors
  (id values, parent_id reconstruction, brand id, dynamic tag grouping,
  reserved-key collision routing).
- Full suite: **14517 passing**, no failures.
- Lint: clean.

#

image

### DIFF
--- a/src/support/elements/elements-service.js
+++ b/src/support/elements/elements-service.js
@@ -77,7 +77,7 @@ export function createElementsService(transport) {
       Object.entries(otherGroups).forEach(([key, items]) => {
         result[key] = result[key] ? [...result[key], ...items] : items;
       });
-      result.tags = tags;
+      result.tags = result.tags ? [...result.tags, ...tags] : tags;
       return result;
     },
 
--- a/src/support/elements/elements-service.js
+++ b/src/support/elements/elements-service.js
@@ -22,6 +22,7 @@ import {
   transformCategoriesToFilterDimensions,
   transformIntentsToFilterDimensions,
   transformOriginsToFilterDimensions,
+  transformOtherTagsForFilterDimensions,
   buildWeeksPayload,
   transformWeeksResponse,
   buildPromptsPayload,
@@ -61,7 +62,7 @@ export function createElementsService(transport) {
         transport.fetchElement(workspaceId, ELEMENT_IDS.BRANDS, buildBrandsPayload(params)),
         transport.fetchElement(workspaceId, ELEMENT_IDS.MARKETS, buildMarketsPayload({})),
       ]);
-      return {
+      const result = {
         brands: transformBrandsToFilterDimensions(rawBrands, spacecatBrands),
         regions: transformMarketsToFilterDimensions(rawMarkets, brandSemrushProjects),
         topics: transformTopicsForFilterDimensions(rawTopics),
@@ -69,6 +70,15 @@ export function createElementsService(transport) {
         page_intents: transformIntentsToFilterDimensions(rawTopics),
         origins: transformOriginsToFilterDimensions(rawTopics),
       };
+      // Merge any tag types not covered above (e.g. `type:branded`) under their own
+      // prefix key, and plain prefix-less tags under `tags` — see
+      // transformOtherTagsForFilterDimensions.
+      const { tags, ...otherGroups } = transformOtherTagsForFilterDimensions(rawTopics);
+      Object.entries(otherGroups).forEach(([key, items]) => {
+        result[key] = result[key] ? [...result[key], ...items] : items;
+      });
+      result.tags = tags;
+      return result;
     },
 
     /**
--- a/src/support/elements/elements-service.js
+++ b/src/support/elements/elements-service.js
@@ -72,12 +72,24 @@ export function createElementsService(transport) {
       };
       // Merge any tag types not covered above (e.g. `type:branded`) under their own
       // prefix key, and plain prefix-less tags under `tags` — see
-      // transformOtherTagsForFilterDimensions.
-      const { tags, ...otherGroups } = transformOtherTagsForFilterDimensions(rawTopics);
+      // transformOtherTagsForFilterDimensions. The reserved-key list is derived
+      // from `result`'s own keys (not hand-duplicated) so it can't drift out of
+      // sync if a key here is ever renamed or a new fixed dimension is added.
+      // `__proto__`/`constructor`/`prototype` are added on top: `result[key] = ...`
+      // below would repoint result's prototype instead of adding a property for
+      // those (rather than dropping such tags, routing them as "reserved" sends
+      // them into the generic `tags` array, same as any other collision).
+      const reservedResultKeys = [
+        ...Object.keys(result), 'tags', '__proto__', 'constructor', 'prototype',
+      ];
+      const { tags, ...otherGroups } = transformOtherTagsForFilterDimensions(
+        rawTopics,
+        reservedResultKeys,
+      );
       Object.entries(otherGroups).forEach(([key, items]) => {
-        result[key] = result[key] ? [...result[key], ...items] : items;
+        result[key] = items;
       });
-      result.tags = result.tags ? [...result.tags, ...tags] : tags;
+      result.tags = tags;
       return result;
     },
 
--- a/docs/openapi/api.yaml
+++ b/docs/openapi/api.yaml
@@ -174,6 +174,8 @@ paths:
     $ref: './serenity-api.yaml#/v2-serenity-activate'
   /v2/orgs/{spaceCatId}/brands/{brandId}/serenity/deactivate:
     $ref: './serenity-api.yaml#/v2-serenity-deactivate'
+  /v2/orgs/{spaceCatId}/brands/{brandId}/serenity/brand-presence/url-inspector/filter-dimensions:
+    $ref: './serenity-api.yaml#/v2-serenity-url-inspector-filter-dimensions'
   /organizations/{organizationId}/sites:
     $ref: './sites-api.yaml#/sites-for-organization'
   /organizations/{organizationId}/entitlements:
--- a/docs/openapi/schemas.yaml
+++ b/docs/openapi/schemas.yaml
@@ -11475,14 +11475,18 @@ SerenityFilterDimensionItem:
   type: object
   description: |
     A single id/label option for a Serenity Elements filter dropdown. `id` is
-    `null` when Semrush has no stable identifier for the value (most tag-derived
-    dimensions); `parent_id`/`parent_label` are present only when the source tag
-    encoded a "Parent__Child" hierarchy (e.g. `topic:Furniture__Sofas`).
+    the original, unmodified tag value returned by Semrush (including its
+    `prefix:` and any `Parent__Child` encoding, e.g. `topic:Furniture__Sofas`).
+    `parent_id`/`parent_label` are present only when the source tag encoded a
+    "Parent__Child" hierarchy; `parent_id` is the same `prefix:` reconstructed
+    over just the parent segment (e.g. for
+    `category:Living Room Furniture Retail__Living Room Furniture and Sofas`,
+    `parent_id` is `category:Living Room Furniture Retail`).
   required: [id, label]
   properties:
-    id: { type: [string, 'null'] }
+    id: { type: string }
     label: { type: string }
-    parent_id: { type: [string, 'null'] }
+    parent_id: { type: string }
     parent_label: { type: string }
 
 SerenityUrlInspectorFilterDimensions:
@@ -11503,14 +11507,14 @@ SerenityUrlInspectorFilterDimensions:
     brands:
       type: array
       description: |
-        Available brands in the workspace. `id` is always null (Semrush has no
+        Available brands in the workspace. `id` mirrors `label` (Semrush has no
         stable brand id); `spacecat_brand_id` resolves by case-insensitive name
         match against the org's SpaceCat brands.
       items:
         type: object
         required: [id, label]
         properties:
-          id: { type: [string, 'null'] }
+          id: { type: string }
           label: { type: string }
           spacecat_brand_id: { type: [string, 'null'] }
     regions:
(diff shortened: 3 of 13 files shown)


src/support/elements/elements-service.js
```diff
@@ -61,14 +62,23 @@ export function createElementsService(transport) {
...
         page_intents: transformIntentsToFilterDimensions(rawTopics),
         origins: transformOriginsToFilterDimensions(rawTopics),
       };
+      // Merge any tag types not covered above (e.g. `type:branded`) under their own
+      // prefix key, and plain prefix-less tags under `tags` — see
+      // transformOtherTagsForFilterDimensions.
+      const { tags, ...otherGroups } = transformOtherTagsForFilterDimensions(rawTopics);
+      Object.entries(otherGroups).forEach(([key, items]) => {
+        result[key] = result[key] ? [...result[key], ...items] : items;
+      });
```
MysticatBot, line 91: issue (blocking): `KNOWN_TAG_PREFIXES` excludes `topic:`, `category:`, `intent:`, `source:` but does NOT exclude tags whose prefix matches existing result keys (`brands`, `regions`, `topics`, `categories`, `page_intents`, `origins`). If Semrush ever returns a tag like `brands:foo` or `regions:APAC`, the merge logic on line 73 will inject items into those pre-existing arrays, corrupting the response.

Additionally, `result.tags` is assigned unconditionally without the merge guard applied to other keys, creating an asymmetry that could silently drop data if `tags` were ever populated earlier.

Fix options:
- Extend the exclusion list to cover all known result keys, OR
- Skip any `otherGroups` key that already exists in `result` (prevent collision rather than merge), OR
- Namespace all dynamic groups under a single stable key (e.g. `result.otherDimensions = { ...otherGroups, tags }`) to keep the top-level response shape predictable for consumers.

MysticatBot (changes requested):
Hey @vivesing,

**Verdict: Request changes** - dynamic key collision risk needs a guard before this ships.
**Complexity: MEDIUM** - medium diff, API surface change.
_Changes: Adds Parent__Child subcategory hierarchy parsing and generic handling of unknown Semrush tag types in the filter-dimensions response (5 files)._

## Must fix before merge

1. [Important] Dynamic keys from external data can collide with existing result keys - `src/support/elements/elements-service.js:73` (details inline)
2. [Important] OpenAPI spec not updated for new response shape - CLAUDE.md requires "Specification Sync: Keep OpenAPI specs and implementation in sync". This PR adds `parent_id`/`parent_label` fields to existing dimension items and introduces dynamic keys plus `tags` to the response. The corresponding OpenAPI schema (`UrlInspectorFilterDimensions` in `docs/openapi/schemas.yaml`) should be updated to reflect the new shape.

<details>

- nit: `splitParent` returns inconsistent object shapes - items without `__` get `{ id, label }` while items with it get `{ id, label, parent_id, parent_label }`. Consider always returning all keys (with null) for a uniform shape. - `src/support/elements/definitions/topics.js:57`
- suggestion: Derive `KNOWN_TAG_PREFIXES` from a single source of truth shared with the existing transform functions to prevent drift when new named transforms are added - `src/support/elements/definitions/topics.js:107`
- suggestion: Add edge case tests for `splitParent` with boundary inputs like `"__Child"`, `"Parent__"`, `"__"` which produce empty-string labels - `test/support/elements/definitions/topics.test.js`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 1s | <b>Cost:</b> $6.25 | <b>Commit:</b> <code>f58b267ce14098cf4a056958ee874264b24a3552</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/elements/elements-service.js
vivesing: fixed

MysticatBot (approved):
Hey @vivesing,

**Verdict: Approve** - both prior blocking issues are resolved; remaining items are non-blocking hardening suggestions.
**Complexity: MEDIUM** - medium diff, API surface change.
_Changes: Adds Parent__Child subcategory hierarchy parsing and generic handling of unknown Semrush tag types in the filter-dimensions response (9 files)._
Note: Recommend a human read before merge - this change modifies shared API contracts (OpenAPI schemas). The bot review is a complement to, not a replacement for, a human read here.

- Dynamic key collision risk: `RESERVED_RESULT_KEYS` guard now routes colliding prefixes to the generic `tags` array, with test coverage for all seven reserved keys.
- OpenAPI spec not updated: new `SerenityFilterDimensionItem` and `SerenityUrlInspectorFilterDimensions` schemas added to `schemas.yaml`; endpoint defined in `serenity-api.yaml`.

<details>

- nit: Use `Object.create(null)` instead of `{}` for the `groups` object in `transformOtherTagsForFilterDimensions` - prevents any edge-case interaction with prototype keys (`__proto__`, `constructor`) when external vendor data is used as object keys. One-line defensive fix. - `src/support/elements/definitions/topics.js:125`
- nit: Tag value `":"` (empty prefix) creates an empty-string key `""` on the response object - consider routing empty prefixes to the `tags` bucket alongside the `RESERVED_RESULT_KEYS` check. - `src/support/elements/definitions/topics.js:133`
- suggestion: Add 1-2 boundary tests for `splitParent` with inputs like `"__Child"`, `"Parent__"`, `"__"` to pin the empty-string label behavior as intentional. - `test/support/elements/definitions/topics.test.js`
- suggestion: Document the criteria for promoting a dynamic key to a fixed dimension (when a prefix like `type` appears consistently enough to warrant guaranteed presence in the response shape). No code change needed now. - `src/support/elements/elements-service.js:68`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 21s | <b>Cost:</b> $5.19 | <b>Commit:</b> <code>36bb3c454715c973e32a93e370ff4e0cb8b3ac96</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

calvarezg:
## Multi-persona review — feat: serenity filter-dimensions for sub categories

Both prior MysticatBot blocking issues are genuinely resolved (verified): `RESERVED_RESULT_KEYS` makes the dynamic-key set provably disjoint from the fixed result keys, and the OpenAPI specs are updated in sync. One new blocking issue and a couple of hardening items remain.

### 🔴 Critical — unhandled 500 on `Object.prototype`-named tag prefixes
`transformOtherTagsForFilterDimensions` uses `const groups = {}` (`topics.js:152`). For a tag whose prefix is `__proto__`, `constructor`, `toString`, `valueOf`, `hasOwnProperty`, etc., `groups[prefix] ?? []` resolves to the *inherited* prototype member (not nullish), so `?? []` never fires and `groups[prefix].push(...)` (`topics.js:167`) throws `TypeError` → the endpoint 500s. This PR's whole point is handling arbitrary unknown prefixes, so such a prefix is in-domain. Reproduced in Node.
**Fix:** `const groups = Object.create(null)` (or a `Map`). Then harden the merge in `elements-service.js:73` — a `__proto__` key reaching `result[key] = ...` would set `result`'s prototype; skip/guard `__proto__`/`constructor`/`prototype`. Add `__proto__:`/`constructor:` tests.

### 🟠 Major
1. **Key-list drift reintroduces the collision bug.** `KNOWN_TAG_PREFIXES` and `RESERVED_RESULT_KEYS` (`topics.js:117,127`) duplicate strings whose source of truth lives in `elements-service.js`. Renaming/adding a result key without updating `RESERVED_RESULT_KEYS` silently re-corrupts arrays. Export the reserved set from where the result is built; derive prefixes from the transform registry.
2. **`id` contract change (Question).** `id` goes from `null`/`UPPERCASED`/bare to the raw prefixed tag (`intent:Informational`, `source:organic`). Confirm no released consumer matches on the old `id`; version/note if shipped.

### 🟡 Minor
- `parent_id` is synthesized from string-splitting and may dangle (no guaranteed sibling entry) — document as advisory or drop when absent.
- Degenerate values (`":foo"` → empty-string key; `"__"`/`"Parent__"`/`"__Child"` → empty labels) pass OpenAPI `required:[id,label]` silently; prior-review boundary tests still missing.
- No de-duplication within a bucket (duplicate `id`s now that `id` is the option key).
- Dead defensive branches in the merge (`elements-service.js:73-78`) — unreachable given the collision guard; simplify or test.
- `splitParent` returns inconsistent object shapes (with/without `parent_*`).

**Recommendation: Request changes (blocking)** — the prototype-prefix crash is a one-line fix but a hard 500 reachable from external data.

<sub>Multi-persona review: Architect · DBA · Security · Tester · Junior Dev</sub>

vivesing:
Summary

Fixed the 🔴 Critical and 🟠 Major issues you asked me to prioritize, verified with reproductions.

Critical — prototype-key crash (confirmed with node -e):
- src/support/elements/definitions/topics.js: groups = {} → groups = Object.create(null). This was a genuine, reproducible TypeError for any Semrush tag prefix matching an Object.prototype member (constructor, toString, __proto__, hasOwnProperty, etc.) — groups[prefix] ?? [] resolved to the inherited member instead of undefined, and .push(...) then threw.
- src/support/elements/elements-service.js: also found and fixed a second, distinct issue at the merge step — result[key] = items for key === '__proto__' doesn't crash but silently repoints result's prototype (verified: the array becomes result's [[Prototype]], and the entry vanishes from Object.keys/JSON output). Fixed by folding __proto__/constructor/prototype into the reserved-key list so those prefixes get routed into the generic tags array instead — no crash, no silent data loss, no separate guard branch needed.

Major — key-list drift:
- Removed the hardcoded RESERVED_RESULT_KEYS constant from topics.js. transformOtherTagsForFilterDimensions now takes reservedResultKeys as a parameter.
- elements-service.js derives it directly from Object.keys(result) (plus 'tags' and the JS-unsafe names) — the reserved set can no longer drift out of sync with the actual response shape, since it's computed from the object itself rather than hand-duplicated.

Tests added: regression tests in topics.test.js (unit level) and elements-service.test.js (end-to-end) proving constructor:evil, __proto__:evil, toString:evil prefixes don't crash and don't corrupt the result's prototype.

calvarezg (approved):
Approving — the 🔴 prototype-key crash is dead (`Object.create(null)` + reserving `__proto__`/`constructor`/`prototype`), the reserved-key set is now derived from `result` itself so it can't drift, and the regression tests actually throw `constructor:evil` at the code and prove it stays calm. Clean work. 🎯

@vivesing take a bow — you turned a `groups[prefix].push is not a function` 500 into a non-event, with tests to keep it that way. Genuinely tidy fix. 🙌

Now go reward yourself with a proper cutting chai from Noida — you've earned it (extra adrak for shipping a null-prototype object AND remembering that object spread doesn't trip the `__proto__` setter — that's the kind of detail that separates the pros from the `{}`s). ☕️🚀

PR #2792, opened 2026-07-10
feat(ai-visibility): server-side sort for brand list endpoints

jjenscodee:
## Summary

Adds server-side sort (`sortBy` / `sortDirection`) to the four AI Visibility **brand list** endpoints, so the UI can page + sort datasets larger than 100 rows against the full dataset instead of client-sorting a fixed 100-row batch. Part of **LLMO-6104**.

| Endpoint | Sort keys |
|---|---|
| `/brands/cited-pages` | `URL`, `PROMPTS_COUNT` (default) |
| `/brands/cited-sources` | `DOMAIN`, `MENTIONS`, `SOURCES_COUNT`, `PROMPTS_COUNT` (default), `ORGANIC_TRAFFIC` |
| `/brands/source-opportunities` | same `DOMAINS_REQUEST_ORDER_BY` set (default `ORGANIC_TRAFFIC`) |
| `/brands/top-brands` | `NAME`, `MENTIONS` (default) — in-handler JS sort |

## Details

- New `resolveGrpcSortOrder(sp, enum, defaultBy)` helper maps the query params onto the gRPC request `order` (`{ by, direction }`). Unknown/`UNSPECIFIED` values fall back to each endpoint's default; direction defaults to `DESC` (matching the order the UI previously displayed via its now-removed client sort).
- **Top Brands** parametrizes its in-handler comparator. It also now fetches the full set (1000 cap) whenever the requested order differs from the gRPC client's native mentions-desc order, so `NAME`/ascending pagination slices the whole dataset rather than a truncated `offset+limit+1` window (which would return wrong rows on page ≥2).
- OpenAPI: `sortBy`/`sortDirection` added to all four paths; `docs` rebuilt.

## Testing

- `npm run lint` clean; `test/support/ai-visibility/handlers/brands.test.js` — **135 passing** (added coverage for enum mapping, default order, unknown-value fallback, NAME/ascending order, and the full-vs-truncated fetch window).
- `npm run docs:lint` / `docs:build` pass.

## Related issues
https://jira.corp.adobe.com/browse/LLMO-6104

### DIFF
--- a/docs/openapi/ai-visibility-api.yaml
+++ b/docs/openapi/ai-visibility-api.yaml
@@ -446,6 +446,54 @@ x-ai-vis-fts-source-domains-sort-by: &ftsSourceDomainsSortByParam
     enum: [DOMAIN, SOURCES_COUNT, MENTIONS, ORGANIC_TRAFFIC]
     default: MENTIONS
 
+x-ai-vis-brands-cited-pages-sort-by: &brandsCitedPagesSortByParam
+  name: sortBy
+  in: query
+  required: false
+  description: |
+    `SOURCES_REQUEST_ORDER_BY` for Semrush v2 `SourcesRequest.order.by`.
+    Default when omitted: `PROMPTS_COUNT`. Unknown values fall back to the default.
+  schema:
+    type: string
+    enum: [URL, PROMPTS_COUNT]
+    default: PROMPTS_COUNT
+
+x-ai-vis-brands-top-brands-sort-by: &brandsTopBrandsSortByParam
+  name: sortBy
+  in: query
+  required: false
+  description: |
+    Sort key for the top-brands list (sorted in-process, not via gRPC order).
+    Default when omitted: `MENTIONS`. Unknown values fall back to the default.
+  schema:
+    type: string
+    enum: [NAME, MENTIONS]
+    default: MENTIONS
+
+x-ai-vis-brands-cited-sources-sort-by: &brandsCitedSourcesSortByParam
+  name: sortBy
+  in: query
+  required: false
+  description: |
+    `DOMAINS_REQUEST_ORDER_BY` for Semrush v2 `DomainsRequest.order.by`.
+    Default when omitted: `PROMPTS_COUNT`. Unknown values fall back to the default.
+  schema:
+    type: string
+    enum: [DOMAIN, MENTIONS, SOURCES_COUNT, PROMPTS_COUNT, ORGANIC_TRAFFIC]
+    default: PROMPTS_COUNT
+
+x-ai-vis-brands-source-opportunities-sort-by: &brandsSourceOpportunitiesSortByParam
+  name: sortBy
+  in: query
+  required: false
+  description: |
+    `DOMAINS_REQUEST_ORDER_BY` for Semrush v2 gap `GapSourceDomainsRequest.order.by`.
+    Default when omitted: `ORGANIC_TRAFFIC`. Unknown values fall back to the default.
+  schema:
+    type: string
+    enum: [DOMAIN, MENTIONS, SOURCES_COUNT, PROMPTS_COUNT, ORGANIC_TRAFFIC]
+    default: ORGANIC_TRAFFIC
+
 x-ai-vis-prompt-hash: &promptHashParam
   name: promptHash
   in: query
@@ -587,6 +635,8 @@ ai-visibility-brands-cited-pages:
       - *engineParam
       - *limitParam
       - *offsetParam
+      - *brandsCitedPagesSortByParam
+      - *sortDirectionParam
       - *monthParam
     responses:
       '200':
@@ -640,6 +690,8 @@ ai-visibility-brands-top-brands:
       - *engineParam
       - *limitParam
       - *offsetParam
+      - *brandsTopBrandsSortByParam
+      - *sortDirectionParam
     responses:
       '200':
         description: Top brands retrieved successfully
@@ -666,6 +718,8 @@ ai-visibility-brands-cited-sources:
       - *engineParam
       - *limitParam
       - *offsetParam
+      - *brandsCitedSourcesSortByParam
+      - *sortDirectionParam
     responses:
       '200':
         description: Brand cited sources retrieved successfully
@@ -693,6 +747,8 @@ ai-visibility-brands-source-opportunities:
       - *engineParam
       - *limitParam
       - *offsetParam
+      - *brandsSourceOpportunitiesSortByParam
+      - *sortDirectionParam
     responses:
       '200':
         description: Brand source opportunities retrieved successfully
--- a/src/support/ai-visibility/handlers/brands.js
+++ b/src/support/ai-visibility/handlers/brands.js
@@ -15,6 +15,7 @@
 import { ConnectError, Code } from '@connectrpc/connect';
 import { BRAND_TOPICS_ORDER_BY_ENUM } from '@quazar/ai-seo-ts/v2/topic/enums_pb.js';
 import { PROMPTS_REQUEST_ORDER_BY_ENUM } from '@quazar/ai-seo-ts/v2/prompt/enums_pb.js';
+import { ORDER_DIRECTION_ENUM } from '@quazar/ai-seo-ts/common/types_pb.js';
 import {
   SOURCES_REQUEST_ORDER_BY_ENUM,
   DOMAINS_REQUEST_ORDER_BY_ENUM,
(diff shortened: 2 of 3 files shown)


MysticatBot (approved):
Hey @jjenscodee,

**Verdict: Approve** - well-scoped sort feature with clean implementation, thorough tests, and proper OpenAPI-first workflow.
_Changes: Adds server-side sort (sortBy/sortDirection) to the four AI Visibility brand list endpoints via a shared gRPC sort helper and a parametrized in-handler comparator for top-brands (4 files)._

<details>

- nit: Case-sensitivity inconsistency - `resolveGrpcSortOrder` is case-sensitive for `sortBy` (no `toUpperCase()`), but `handleBrandTopBrands` normalizes to uppercase before comparing. A client sending `sortBy=url` silently gets default sort on cited-pages but `sortBy=name` works on top-brands. Consider adding `.toUpperCase()` to `resolveGrpcSortOrder`'s `byKey` for consistency. - `src/support/ai-visibility/handlers/brands.js:~449`
- nit: Behavioral divergence from existing `resolveFtsSort` in topics.js, which returns 400 for unrecognized sort values. This PR's `resolveGrpcSortOrder` silently falls back to defaults. The OpenAPI spec documents this explicitly ("Unknown values fall back to the default"), so it is a deliberate choice - but worth a code comment acknowledging the divergence so future maintainers don't accidentally "fix" it. - `src/support/ai-visibility/handlers/brands.js:~448`
- suggestion: Add a test for invalid `sortDirection` falling back to DESC (e.g. `sortDirection=BOGUS`). The unknown-`sortBy` fallback is tested but the direction fallback path is not, creating an asymmetric gap. - `test/support/ai-visibility/handlers/brands.test.js:~660`
- suggestion: Add a test for `sortBy=NAME` without explicit `sortDirection` (should produce Z-A descending order). The `topBrandsSortAsc ? c : -c` false branch for NAME sort is currently untested. - `test/support/ai-visibility/handlers/brands.test.js:~1070`
- nit: `localeCompare` without a fixed locale in the top-brands comparator. In container environments the default locale can vary. Consider `a.name.localeCompare(b.name, 'en')` for deterministic pagination across servers. - `src/support/ai-visibility/handlers/brands.js:~620`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 4m 0s | <b>Cost:</b> $4.32 | <b>Commit:</b> <code>3ea72f298b46545e5933ec57383be517ce268543</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @jjenscodee,

**Verdict: Approve** - all prior nits addressed cleanly, no new concerns.
_Changes: Adds server-side sort (sortBy/sortDirection) to the four AI Visibility brand list endpoints via a shared gRPC sort helper and a parametrized in-handler comparator for top-brands (4 files)._

- Case-insensitive sortBy matching: `.toUpperCase()` added to `resolveGrpcSortOrder`
- Behavioral divergence comment: added explaining the difference from `resolveFtsSort` (deliberate lenient fallback vs 400)
- Test for invalid sortDirection fallback to DESC: added
- Test for sortBy=NAME without explicit sortDirection (Z-A descending): added
- Deterministic locale for `localeCompare`: `'en'` added to both comparator calls

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 6m 11s | <b>Cost:</b> $1.43 | <b>Commit:</b> <code>a9b17c59b6e4b71e685ff6863e47444e8d1b902a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jjenscodee:
Pushed `b8035f3`: **stable top-brands `total` across pages.**

`handleBrandTopBrands` had a page-dependent fetch window (`fetchN = 1000` at offset 0, but `offset+limit+1` on later pages in native mentions-desc order). Since top-brands has no dedicated count call on the brand gRPC client, `total = dataFull.length` inherited that window — so the total shrank as the UI paged forward and the pager's page-count jumped. Now always fetches the full set (1000 cap) so `total` is identical on every page. Slicing and the other three brand endpoints (which already have stable totals via a dedicated totals request) are unchanged.

Known limitation: with no count RPC, top-brands stays capped at 1000 (silent hard cap for now). Pairs with adobe/project-elmo-ui#2391.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

JayKid:
**Multi-persona review (Architect, DBA, Security, Tester) — Approve with reservations.**
No blocker in the code; the security posture is clean and the `total`-stability fix (`b8035f3`) is a real bugfix. Two items worth resolving before merge:

**1. [Major] cited-sources & source-opportunities accept undocumented enum members.**
`resolveGrpcSortOrder` matches `sortBy` against the full `DOMAINS_REQUEST_ORDER_BY_ENUM` (9 members), but the OpenAPI documents only values 1–5. `sortBy=MENTIONS_COUNT_BY_BRAND` / `MENTIONED_COMPETITORS` / `TOTAL_COMPETITOR_MENTIONS` (6–8) pass the `>0` guard and go to the backend instead of falling back to default — breaking the documented "Unknown values fall back to the default" guarantee. (cited-pages is fine; its enum == the documented set.) Suggest an explicit per-endpoint allow-list, like `resolveFtsSort`'s `sortByMap` in `topics.js`. This also closes the latent drift risk when `ai-seo-ts` is regenerated.

**2. [Major / verify] Ordering now depends on the backend honoring `order.direction`.**
The proto field is real, but the UI's client sort was removed, so if the backend ignores `direction`, `sortDirection=ASC` silently returns DESC with no error. Please confirm via an integration test or the Semrush team, and confirm the pre-PR default (unset `direction`) was already DESC (otherwise direct-API consumers' order flips).

**Minor / follow-ups:** de-dup the inline sort-param parsing in `handleBrandTopBrands` (share with the helper); document the ~1000-row top-brands cap in OpenAPI; consider a debug log on silent fallback. Test gaps: the `UNSPECIFIED`(=0) fallback branch, `total`-stability across offsets with a non-empty dataset (current tests only assert `limit===1000` on empty data), and whitespace trimming.

_Prior MysticatBot nits (case-insensitivity, divergence comment, direction/NAME tests, locale) all verified resolved — not re-flagged._

<sub>🤖 Multi-persona PR review via Claude Code</sub>

jjenscodee:
Thanks for the thorough multi-persona pass @JayKid — both Major items addressed in `2acfe32`.

**1. Undocumented enum members (cited-sources / source-opportunities).** Confirmed: `DOMAINS_REQUEST_ORDER_BY_ENUM` carries members 6–8 (`MENTIONS_COUNT_BY_BRAND`, `MENTIONED_COMPETITORS`, `TOTAL_COMPETITOR_MENTIONS`) that these endpoints don't document, and the `> 0` guard let them through. `resolveGrpcSortOrder` now takes an explicit per-endpoint allow-list (`CITED_PAGES_SORT_KEYS`, `DOMAINS_SORT_KEYS`), so any `sortBy` outside the documented set falls back to the default — honoring the "unknown values fall back to the default" contract and closing the ai-seo-ts regeneration drift risk. Kept it lenient (no 400), matching the OpenAPI. No spec change needed — the docs already listed only the 5 documented keys, so this brings the impl into conformance. Added fallback coverage for both endpoints.

**2. Direction dependency.** Confirmed the real risk from the diff: pre-PR passed `order: { by }` with **no** `direction`, and the PR was forcing `DESC`. Rather than depend on the backend's unset-direction default (which I couldn't verify — the gRPC `ai-seo` service isn't in any local repo), we now only attach `order.direction` when the caller actually sends `sortDirection`. Omitted → no `direction` field → backend keeps its own default ordering, i.e. byte-for-byte the pre-sort wire behavior, so no direct-API consumer can be silently flipped. Provided-but-invalid still falls back to `DESC`. Verified against the UI PR (project-elmo-ui#2391) that all four brand tabs always send `sortDirection`, so the UI never relies on an implicit default.

One doc nuance: the shared `sortDirection` param still reads "Default when omitted: DESC" — under this approach that's now the backend's default rather than one we force. Left the shared anchor untouched (other endpoints reference it); happy to scope a brand-specific param if you'd prefer it read "backend default applies."

Minor/follow-up items (dedup top-brands inline parsing, document the ~1000-row cap, fallback debug log) noted — can fold into this PR or a follow-up, your call.

JayKid (approved):
**Approving.** Verified `2acfe32` against both Major items from the multi-persona review — code and tests both check out.

**1. Undocumented enum members — fixed.** `resolveGrpcSortOrder` now gates on an explicit per-endpoint allow-list (`CITED_PAGES_SORT_KEYS`, `DOMAINS_SORT_KEYS`) before the enum lookup, matching the OpenAPI sets exactly. `DOMAINS` members 6–8 (`MENTIONS_COUNT_BY_BRAND` / `MENTIONED_COMPETITORS` / `TOTAL_COMPETITOR_MENTIONS`) now fall back to default instead of reaching the backend, and the two new fallback tests cover exactly those members. Also closes the ai-seo-ts regeneration drift risk.

**2. Direction dependency — resolved the right way.** Only attaching `order.direction` when `sortDirection` is supplied makes the wire byte-for-byte identical to pre-PR for callers that don't opt in, so no existing direct-API consumer can be silently flipped — cleaner than forcing a default. The tests dropping the `direction: DESC` assertion on the no-param path confirm it.

One residual, non-blocking: whether the backend actually honors `direction=ASC` can't be proven from the diff (the gRPC `ai-seo` service isn't in-repo). The UI always sends `sortDirection`, so ascending sort still rides on that backend behavior — worth one live smoke test before leaning on it, but it no longer endangers non-opting callers.

Minor follow-ups (dedup top-brands inline parsing, document the ~1000-row cap in OpenAPI, fallback debug log) are fine to fold in here or as a follow-up — your call.

PR #2795, opened 2026-07-13
feat: added move site slack command

anshulk-public:
## Summary

- Adds a new Slack command, `move plg site {baseURL} {imsOrgId}`, letting an ESE/admin move a PLG site from one org to another org directly from Slack, instead of the manual multi-step runbook (`set imsorg` + manual entitlement API call + retry).
- Follows the existing two-step confirm pattern used by `set-ims-org.js`/`set-ims-org-modal.js`: the command (`src/support/slack/commands/move-plg-site.js`) only validates and posts a summary with a "Confirm Move" button — no mutation happens until the button is clicked. The action (`src/support/slack/actions/move-plg-site.js`) re-validates fresh state and performs the actual move.
- Guards against unsafe moves: refuses if the target org is internal/demo, or if the target org already has a `PAID` ASO entitlement. Warns (but does not block) if the site currently has other active product enrollments (ASO, LLMO, ACO) — the confirm message says which ones will be revoked.
- On confirm: revokes all of the site's existing enrollments (any product), reassigns the site's `organizationId`, sets the target org's ASO entitlement to `PRE_ONBOARD` (bumping it there if it was `FREE_TRIAL`, `PLG`, or missing entirely — left alone if already `PRE_ONBOARD`), and binds a fresh site enrollment to that entitlement.
- Registered the command in `src/support/slack/commands.js` and the action (`open_move_plg_site_modal`) in `src/support/slack/actions/index.js`.

You can see example here: https://cq-dev.slack.com/archives/C060T2PPF8V/p1783922817176619
image

## Related Issues

Thanks for contributing!


absarasw (changes requested):
Reviewing the data-model implications of the move — one blocking-for-me finding + three non-blocking comments.

## 🚫 Blocking — `projectId` isn't reassigned; UI ends up with dangling reference

`reassignSiteOrganization(site, organizationId)` (in `src/controllers/plg/plg-onboarding/entitlement.js`) is only:

```js
site.setOrganizationId(organizationId);
return site.save();
```

But per `spacecat-shared-data-access`:

- **`Site.projectId`** exists on the schema (`packages/spacecat-shared-data-access/src/models/site/site.schema.js:204`) and is an indexed composite key component.
- **`Project.organizationId`** — projects belong to organizations (`packages/spacecat-shared-data-access/src/models/project/index.d.ts:24-29`).
- The collection method `Site.allByOrganizationIdAndProjectId(orgId, projectId)` explicitly validates the projectId belongs to the org (`site.collection.js:139`).

Consequence of the current PR:

- **Before:** site has `organizationId = A`, `projectId = X` where Project X ∈ Org A.
- **After merge:** site has `organizationId = B`, `projectId = X` — but Project X still belongs to Org A. Dangling reference.
- **UI (which groups sites by project) queries `allByOrganizationIdAndProjectId(B, X)` → empty → the moved site shows up in the "Unassigned" folder in the target org.**

That's a regression from the operator's perspective and it makes the composite index inconsistent.

### Minimal fix

Clear `projectId` as part of the move so the outcome is honest rather than a dangling FK:

```js
export async function reassignSiteOrganization(site, organizationId) {
  site.setOrganizationId(organizationId);
  // Site's project (if any) lives under the source org — clearing to avoid
  // a dangling reference. Operator can re-assign in the target org's UI.
  if (site.getProjectId?.()) {
    site.setProjectId(null);
  }
  return site.save();
}
```

**And** extend the confirm-message summary (in `src/support/slack/commands/move-plg-site.js`) to say something like:

> ⚠️  Site is currently in project `<project-name>` under the source org. After the move, it will appear as **Unassigned** in the target org's UI. Re-assign to a project in the target org after the move.

That matches what the operator sees today (per team observation), removes the dangling projectId, and gives them the next-step in the same Slack surface they're already in.

Alternative if the team wants to preserve project structure across the move: extend the command with `--target-project-id`, prompt the operator via a Slack modal to pick a project in the target org, and set both fields atomically. Bigger change; depends on how often this move fires and whether the operator usually knows the target project up-front.

## 💬 Comment 1 — Not transactional

The move is a multi-write sequence: `revoke enrollments → reassignSiteOrganization → createEntitlement (maybe) → bindAsoSiteEnrollment`. If step 2 or 4 fails after step 1 succeeded, the site is left with **no enrollments and still in the source org** (or in the target org with no enrollments, if step 2 committed). Recovery is manual.

Not blocking IMO — the confirm-button gate already limits blast radius — but worth either:
- documenting the partial-failure recovery in the code comment above the sequence, or
- wrapping the sequence in a compensating retry helper.

## 💬 Comment 2 — Silent product-drop on non-ASO enrollments

The code revokes ALL enrollments regardless of product (LLMO, ACO, ASO), then only re-binds ASO in the target org. The confirm message does warn *"which enrollments will be revoked"* — good — but if the target org also has LLMO/ACO entitlements, those are silently dropped instead of being re-bound.

Fine for the current scope (this is an ESE tool focused on ASO moves), but worth a code comment near the revoke loop noting the design decision. Otherwise the next person maintaining this will either re-add "safe" LLMO/ACO re-binding without noticing that could conflict with target-org policy, or spend time debugging "why did LLMO stop working after a move."

## 💬 Comment 3 — Race window between button click and re-validation

The confirm handler re-validates state on click (good). But nothing prevents concurrent mutations between the re-validation and the write sequence — e.g., another admin flipping the target org's tier to `PAID` in that window. Low probability, low blast radius, not worth a lock, but worth a one-liner in the code comment: *"re-validation is best-effort; a concurrent tier change between here and the final write may still produce inconsistent state."*

---

Once the projectId gap is addressed I'm happy to re-approve. The other three are drop-in comments — no need for a re-request-review round unless you want to fold them in the same commit.

absarasw (approved):
Re-review on `a75c98fdc14d` — all four points addressed.

## ✅ Blocking — `projectId` gap: **fixed, and better than my ask**

The commit reuses `reparentSiteProject` from `set-ims-org-modal.js` rather than adding a parallel code path. Reading that helper — it handles both cases correctly:

- **Solo site on the project** → move the whole project to the target org (preserves project metadata).
- **Shared project (siblings staying behind)** → create a new project in the target org for the moved site; siblings keep theirs.
- Plus the null-projectId, missing-project (warn + skip), and already-in-target-org branches.

This is genuinely better than my suggested minimum (clearing `projectId`). And the operator gets a Slack info message when the project moves, closing the visibility loop.

## ✅ Comment 1 — Not transactional: addressed via code comment

The new block-level comment above the write sequence explicitly documents the partial-failure recovery expectation. Future maintainers won't be surprised.

## ✅ Comment 2 — Silent LLMO/ACO drop: addressed via code comment

Design-intent note ("only ASO is re-bound; re-enable other products manually post-move") sits right above the revoke loop. Explicit.

## ✅ Comment 3 — Race window: addressed via code comment

Explicit note on the concurrent-tier-change window; consistent with the "low blast radius, don't lock" argument.

## Non-blocking suggestion (optional)

Worth a spot-check on the new test file coverage for the two `reparentSiteProject` branches — solo-project-move vs shared-project-split. Not blocking.

LGTM.

PR #2799, opened 2026-07-13
fix(serenity): write brand URLs verbatim https, skip the brand's own primary domain

rainer-friederich:
## 1. Abstract
Brand URLs are written to Semrush verbatim with their `https://` scheme, and Project Engine's `url/resolve` is dropped from the write path entirely. The brand's own primary domain is skipped when building the brand-URL set, so it is not written as a redundant `website` entry alongside the benchmark that already carries it.

## 2. Reasoning
Semrush's brand-URLs API requires a literal `https://` scheme and stores the submitted value verbatim. It rejects a scheme-less value with `400 … Field validation for 'URL' failed on the 'url' tag`, and an `http://` value with `400 … failed on the 'startswith' tag`.

The write path on `main` resolves website brand URLs through `GET /v1/url/resolve` to Project Engine's canonical form — which is **scheme-less** — and writes that. Semrush therefore rejects every brand-URL write we make. In production, for serenity-active orgs:

- **Brand edit** (`PATCH /brands` → `syncBrandUrlsAcrossMarkets`) hard-fails *after* the brand row has already been committed: the caller gets a 500 and the database drifts out of sync with Semrush.
- **Market create** (`handleCreateMarketSubworkspace`) swallows the failure by design and the market goes live with **no brand URLs propagated at all**, emitting `SERENITY_MARKET_URL_ATTACH_DIVERGENCE`.

This reached production in v1.624.0 (2026-07-06) and is still live. The cause is a merge accident rather than a design decision: the pull request that introduced the `url/resolve` approach was squash-merged while only its first three commits existed. The two commits that reverted the approach — after it was discovered that a scheme-less value cannot be a `brand_urls` value at all — were authored after that merge landed and remained on the feature branch, so they never reached `main`. This change re-applies that correction to current `main`.

The duplicate the resolve was originally meant to prevent is real, but `url/resolve` was the wrong instrument for it: the brand's primary domain appears both as the benchmark's own scheme-less `primary_url` and as an explicit `website` brand-URL row. Since a `brand_urls` value must carry `https://` and a benchmark's domain never does, the two can never be made equal by normalization — the duplicate has to be avoided by not writing the row in the first place.

**Before:** every `website`-type brand URL was sent through `transport.resolveUrl` and the returned scheme-less canonical value was written to Semrush, which rejected it. The brand's primary domain was written as a `website` row, duplicating the benchmark's own `primary_url`.

**After:**

- Brand URLs are written **verbatim**, keeping the `https://` scheme the API demands. `social` and `earned` entries — which are identity handles such as `https://instagram.com/brand`, not brand domains — were never a sensible target for apex-normalization and are likewise written unchanged.
- `resolveWebsiteEntries` and the `resolveUrl` transport method are removed. `url/resolve` is no longer called from any write path; it is not a valid sink for `brand_urls`.
- `collectBrandUrlEntries` takes the project's own primary domain and skips a `website` entry whose host matches it, compared host-wise so that both the apex and the `www.` form are recognised. Market create passes the domain from the create payload; the edit re-sync passes the project's own domain. The benchmark continues to carry the primary domain; socials and secondary sites are unaffected.

Operationally: brand edits that touch URLs stop returning 500 and stop leaving the database diverged from Semrush, and newly created markets get their brand URLs attached instead of silently going live without them.

## 4. Required information
- Other: adobe/serenity-docs issue 25 — Brand URLs: skip the brand's own primary domain (drop url/resolve — invalid sink) https://github.com/adobe/serenity-docs/issues/25

- **spacecat-shared** (contract) — the vendored `spacecat-shared-project-engine-client` supplies the typed Project Engine client and its mock. Its `brand_urls` mock currently accepts any URL, which is why the scheme-less write path passed CI while failing against the live API. https://github.com/adobe/spacecat-shared/pull/1778 makes the mock enforce the live `https://` 400 and closes that fidelity gap.
- **mysticat-data-service** (contract, no change needed) — the Serenity migration CLI writes brand URLs to the same Semrush endpoint and already coerces to `https://` and skips the primary domain, so it stays consistent with this change.
- **project-elmo-ui** (consumed, no change needed) — already normalizes user-entered brand URLs to `https://` before calling this service, so the values arriving here carry a scheme.

The live contract was re-probed against the production Semrush tenant on 2026-07-13, using a throwaway benchmark in an internal test sub-workspace. The benchmark and every row written during the probe were deleted afterwards; the workspace was left as found.

Against `POST /v2/workspaces/{ws}/projects/{pid}/aio/benchmarks/{bid}/brand_urls`:

| submitted | result |
|---|---|
| `example.net`, `www.example.org`, `example.net/products` | 400 — `Field validation for 'URL' failed on the 'url' tag` |
| `http://example.com` | 400 — `… failed on the 'startswith' tag` |
| `https://example.org` | 200 |
| `https://www.example.net` and `https://example.net` | 200 — persisted as two distinct rows, verbatim, with no `www`/apex de-duplication |

Two further observations that shaped this change:

- Semrush announced a deploy "unifying the format for brand URLs" on 2026-07-13. That unification is **display-only**: live rows are still *stored* verbatim with their scheme, while the Semrush UI renders them scheme-less. The write contract is unchanged, so writing scheme-less values remains incorrect.
- A `website` row whose host equals the benchmark's own primary domain is still accepted and still persists through a publish. The duplicate therefore does not resolve itself upstream, and the skip is genuinely required rather than merely defensive.

## 7. Test plan
Verified against the live production Semrush tenant as described in section 6: the scheme-less values this service currently writes are rejected, and the `https://` values it writes after this change are accepted and stored as submitted.

**dev:** with a serenity-active org, `PATCH /brands` on a brand whose `urls` are touched should return 2xx rather than 500, and the brand's `website`/`social`/`earned` URLs should appear on the market's own-brand benchmark with their `https://` scheme intact. The brand's own primary domain should be absent from the brand-URL list — it is carried by the benchmark's `primary_url` instead.

**prod:** after release, confirm that `SERENITY_MARKET_URL_ATTACH_DIVERGENCE` stops appearing in the api-service logs on Serenity market creates, and that a brand edit touching URLs no longer produces the post-commit re-sync failure.

- https://github.com/adobe/spacecat-shared/pull/1778 — related. Makes the vendored project-engine-client mock enforce the live `https://` 400, so a regression of this class fails CI instead of shipping. Independent of this fix at runtime: it changes only the mock, so it neither blocks nor is blocked by this PR.

This PR is safe to merge and deploy on its own, and should not wait on the shared PR. Merging the shared PR first (or after) has no effect on the correctness of this change; landing it at some point is what prevents the regression from recurring.

### DIFF
--- a/src/support/serenity/brand-urls.js
+++ b/src/support/serenity/brand-urls.js
@@ -63,6 +63,24 @@ function toEntry(url, type) {
   return { url: value, type };
 }
 
+// Normalizes a benchmark/brand domain for identity comparison: lowercase host,
+// no scheme, no leading `www.`, no path. Null when unparseable. Used to match an
+// existing own-brand benchmark across runs (idempotent ensure), to skip the
+// primary domain in the brand-URL set, and by the competitor-benchmark sync to
+// match/dedupe competitors by domain.
+export function normalizeBenchmarkDomain(value) {
+  const raw = typeof value === 'string' ? value.trim() : '';
+  if (!raw) {
+    return null;
+  }
+  try {
+    const u = new URL(raw.includes('://') ? raw : `https://${raw}`);
+    return u.hostname.toLowerCase().replace(/^www\./, '');
+  } catch {
+    return null;
+  }
+}
+
 /**
  * Builds the `{ url, type }[]` to push to a single market's project from a
  * brand's URL sources (the V2 shape: `urls`, `socialAccounts`, `earnedContent`
@@ -71,24 +89,45 @@ function toEntry(url, type) {
  *   - `socialAccounts` → type `social`, filtered to the market's region;
  *   - `earnedContent` → type `earned`, filtered to the market's region.
  *
- * Only HTTPS URLs survive. The result is de-duplicated by URL (first-seen wins,
- * so a brand site listed both as a URL and a social account keeps its `website`
- * type) — brand URLs are unique per project upstream, so duplicates would be
- * skipped anyway; de-duping here keeps the `type` we send deterministic.
+ * The brand's own primary domain is dropped from the `website` set when
+ * `primaryDomain` is given (skip-primary-domain): the project's own-brand
+ * benchmark already carries that domain (Semrush stores it scheme-less), and a
+ * `brand_urls` row must keep its `https://` scheme, so the two can never
+ * collapse — emitting it as a website URL just double-lists the domain in the UI
+ * (serenity-docs#25). The match is by normalized host, so both `https://x.com`
+ * and `https://www.x.com` are skipped when the benchmark domain is `x.com`.
+ * Every OTHER website URL (a secondary site) is kept verbatim.
+ *
+ * Only HTTPS URLs survive (`brand_urls` rejects non-https with a 400, so a stored
+ * value is written verbatim — no scheme/`www.` normalization; Semrush stores it
+ * as-is). The result is de-duplicated by URL (first-seen wins, so a brand site
+ * listed both as a URL and a social account keeps its `website` type) — brand
+ * URLs are unique per project upstream, so duplicates would be skipped anyway;
+ * de-duping here keeps the `type` we send deterministic. Note the upstream does
+ * NOT collapse `www.`-vs-apex, so two distinct https URLs that differ only by
+ * `www.` are two separate rows (as the user entered them).
  *
  * @param {object} sources - { urls?, socialAccounts?, earnedContent? }.
  * @param {string} market - ISO-2 country code of the target project.
+ * @param {string} [primaryDomain] - the project's own-brand domain; a `website`
+ *   URL whose host matches it is skipped (see above). Omit to keep all.
  * @returns {{url: string, type: string}[]}
  */
-export function collectBrandUrlEntries(sources, market) {
+export function collectBrandUrlEntries(sources, market, primaryDomain) {
   const urls = Array.isArray(sources?.urls) ? sources.urls : [];
   const social = Array.isArray(sources?.socialAccounts) ? sources.socialAccounts : [];
   const earned = Array.isArray(sources?.earnedContent) ? sources.earnedContent : [];
+  const primary = normalizeBenchmarkDomain(primaryDomain);
 
   const candidates = [
     // Brand URLs carry no region — always every market. Accept both the string
-    // and the { value } object shape the create payload may use.
-    ...urls.map((u) => toEntry(typeof u === 'string' ? u : u?.value, BRAND_URL_TYPE.WEBSITE)),
+    // and the { value } object shape the create payload may use. Skip the brand's
+    // own primary domain (the benchmark already holds it — see fn docstring).
+    ...urls
+      .map((u) => toEntry(typeof u === 'string' ? u : u?.value, BRAND_URL_TYPE.WEBSITE))
+      .filter((e) => e === null
+        || primary === null
+        || normalizeBenchmarkDomain(e.url) !== primary),
     ...social
       .filter((s) => regionApplies(s?.regions, market))
       .map((s) => toEntry(s?.url, BRAND_URL_TYPE.SOCIAL)),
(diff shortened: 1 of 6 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-scoped bug fix with solid test coverage.
**Complexity: HIGH** - medium diff with API surface handler change.
_Changes: removes the broken `url/resolve` normalization path and writes brand URLs verbatim with their `https://` scheme; skips the brand's own primary domain to avoid duplicating the benchmark (6 files)._

<details>

- nit: filter chain coupling - the `.filter((e) => e === null || ...)` intentionally passes null entries through, relying on the downstream dedup filter to strip them; splitting into `.filter(Boolean).filter(...)` would make each stage self-contained - `src/support/serenity/brand-urls.js:77`
- nit: `const desired = collected;` is a vestigial alias from the removed resolve step with no transformation; inlining it reduces noise - `src/support/serenity/brand-urls.js:373`
- suggestion: the handler integration test (`markets-subworkspace.test.js`) does not exercise the primaryDomain skip end-to-end (its `brandDomain` and website URLs don't overlap); adding one case where they match would catch a wiring regression - `test/support/serenity/handlers/markets-subworkspace.test.js`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 3s | <b>Cost:</b> $7.35 | <b>Commit:</b> <code>2fd9e41c45637929cc6af2a0313cb89bb2984073</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Thanks — all three addressed in `167aa83d`.

**1. Filter-chain coupling (`brand-urls.js`)** — fixed. Each stage is now self-contained: `.map(toEntry).filter(Boolean)` strips the unusable entries where they are produced, and the trailing de-dup filter only de-dups. The null-passthrough is gone from the website chain (and the same `.filter(Boolean)` now applies to the social and earned chains, which had the identical implicit coupling).

**2. Vestigial `const desired = collected`** — fixed. The alias is gone; `collectBrandUrlEntries` now assigns straight to `desired`. The comment explaining *why* no transformation happens (entries are written and diffed verbatim, so the diff is stable across re-syncs) moved with it to the assignment site, since that is the non-obvious part worth keeping.

**3. Handler test did not exercise the `primaryDomain` skip** — good catch, and correct: `brandDomain` was `example.com` while the URLs were `b.com`, so the wiring was never actually asserted end-to-end. Added a case where they overlap: with `brandDomain: 'example.com'` and `urls: ['https://example.com', 'https://www.example.com', 'https://shop.example.com']`, only the secondary site is pushed. That now pins the handler → `collectBrandUrlEntries` wiring, so dropping the `body.brandDomain` argument would fail the test rather than silently reintroducing the duplicate.

Also worth recording: the change was verified against the live Semrush tenant beyond unit tests. Driving `syncBrandUrlsAcrossMarkets` (the function `PATCH /brands` calls) across a two-market test sub-workspace: first sync created the non-primary URLs and published them; a re-sync with identical sources was a clean no-op; removing one URL and adding another produced exactly one create and one delete per market; syncing to empty removed everything. No 400s, and the primary domain was skipped on every market in every phase.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all prior nits addressed cleanly, no new concerns.
**Complexity: HIGH** - medium diff; API surface handler change.
_Changes: refactors brand-URL filter chains into self-contained stages, removes the vestigial variable alias, and adds an end-to-end handler test for the primary-domain skip (2 files)._

- Filter chain coupling: each array segment now filters its own nulls immediately via `.filter(Boolean)`
- Vestigial alias: `const desired = collected` removed, variable named directly at the call site
- Handler integration test: new test exercises the primaryDomain skip through `handleCreateMarketSubworkspace` with apex + www matching

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 12s | <b>Cost:</b> $3.83 | <b>Commit:</b> <code>167aa83d0311b9170354132f6cdc6360fa6b51f9</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Aligned the skip semantics with the migration CLI (mysticat-data-service), which is the correct behaviour: **every market's primary domain is skipped as a `website` brand URL, not just the project's own.**

Previously this service skipped only `project.domain` per market. For a market-mirror brand that diverges from the CLI: `chevrolet.ca` is CA-en's primary, so the CLI omits it everywhere, but this service would have written it as a `website` brand URL on the US-en project. State would then depend on which tool touched the brand last — migrate with the CLI, edit once in the UI, and the re-sync re-adds rows the CLI deliberately left out. That is the same duplicate class this PR exists to remove.

- `collectBrandUrlEntries(sources, market, primaryDomains)` now takes the set of every market's primary host (new `primaryDomainSet` helper) instead of a single domain.
- The edit re-sync builds the set once from the project listing it already has — no extra call, and the website entries are now identical across the fan-out.
- Market create derives it from the sibling projects plus the payload's `brandDomain` (the new market has no project to read a domain from yet). That listing sits **inside** the existing best-effort try, so it cannot abort a brand create.

Verified against the live Semrush tenant, not just stubs: the skip set is derived from the live project listing, and with a simulated sibling primary both markets drop that domain while keeping the genuine secondary site. The sync then created, diffed and cleaned up across both markets with no 400s.

Unit coverage added for both the cross-market skip and the handler wiring, so removing the sibling listing fails a test rather than silently regressing.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all prior nits addressed, cross-market skip extension is clean and well-tested.
**Complexity: HIGH** - large diff (dominated by url/resolve removal); API surface handler change.
_Changes: extends the primary-domain skip to all markets' primaries (not just the project's own), aligning with the migration CLI to eliminate state-depends-on-last-writer bugs for market-mirror brands (6 files)._

- Filter chain coupling: each array segment now filters its own nulls immediately via `.filter(Boolean)`
- Vestigial alias: `const desired = collected` removed, variable named directly at the call site
- Handler integration test: new tests exercise the primaryDomain skip through `handleCreateMarketSubworkspace` with both same-market and cross-market matching

<details>

- suggestion: `primaryDomainSet` could use a functional form (`new Set(domains.map(normalize).filter(Boolean))`) instead of the imperative forEach+add loop - `src/support/serenity/brand-urls.js:96`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 12s | <b>Cost:</b> $9.13 | <b>Commit:</b> <code>fdffb5a02fc3cfe05a7499a1ec9782ea5f9f4bd7</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Folded in the draft-read fix — it turned out to be unblocked, and it is a real correctness bug rather than the narrow edge I first took it for.

**What I got wrong first:** I claimed the vendored client could not express `?draft=true`. It can — `aio-list-brand-urls` already declares `draft?: boolean`; I had misread a `query?: never` belonging to the sibling POST/DELETE operations. No spacecat-shared spec change or release is needed.

**The bug.** A brand-URL write lands in the project's DRAFT. `GET .../brand_urls` with no query returns the PUBLISHED view, so a row that has been written but not yet published is invisible to it; a project publish promotes it (the row itself is byte-identical before and after — the split is a view, not a field). `republishBestEffort` deliberately swallows the quota 405, which leaves the project unpublished. The re-sync then read the published view, saw an EMPTY existing set, and so computed an empty `toDelete` — **a URL the user removed was never deleted from Semrush**, for as long as the project stayed unpublished. Creates were re-submitted every sync too (harmless, the upstream de-dupes).

Diffing the published view is wrong on its own terms: creates and deletes act on the draft, so the diff must read the draft — the published view lags it whenever the project has unpublished changes, and the 405 is simply what makes that lag stick.

**Fix:** `listBrandUrls` takes `{ draft }` and the re-sync reads the draft view.

Verified live, reproducing the exact stranded-draft state (rows written, publish withheld):

    after write, UNPUBLISHED:
      published: []
      draft    : [https://shop.example.net, https://old.example.net]
    sync with old.example.net removed by the user:
      {"markets":2,"created":1,"deleted":1}
    after sync:
      published: [https://shop.example.net]

The removal lands. Pre-fix it could not: the published-view read returned nothing to delete. The unit test is a genuine regression guard — I confirmed it fails against the old published-view read and passes against the new one.

**Known gap, not addressed here:** the project-engine-client mock does not model the brand-URL draft/published split (it does model it for prompts, via `is_new`), so an integration test cannot yet catch a regression of this. Worth closing in spacecat-shared alongside the https mock fix.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all prior nits addressed, draft-read fix is clean and well-tested.
**Complexity: HIGH** - large diff; API surface handler change.
_Changes: adds draft-view reading to the brand-URL sync so deletions land even when republish is quota-blocked, completing the url/resolve removal and skip-primary-domain fix (6 files)._

- Filter chain coupling: each array segment now filters its own nulls immediately via `.filter(Boolean)`
- Vestigial alias: `const desired = collected` removed, variable named directly at the call site
- Handler integration test: new tests exercise the primaryDomain skip through `handleCreateMarketSubworkspace` with both same-market and cross-market matching

<details>

- nit: `primaryDomainSet` could use a functional form (`new Set(domains.map(normalizeBenchmarkDomain).filter(h => h !== null))`) instead of the imperative forEach+add loop - `src/support/serenity/brand-urls.js:89`
- suggestion: a one-line comment at the `{ draft: true }` call site noting the invariant ("reads must target the same view that writes act on") would help future maintainers if Semrush ever changes their publication model - `src/support/serenity/brand-urls.js:403`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 59s | <b>Cost:</b> $3.64 | <b>Commit:</b> <code>f71533df71b672a6302ca3d5023883c4015690bd</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Both non-blocking items from the last review are addressed in `1f52fc95`, and `origin/main` is merged in (`99d81aed`).

- **nit — `primaryDomainSet` functional form** (`src/support/serenity/brand-urls.js`): applied. It is now `new Set(domains.map(normalizeBenchmarkDomain).filter((host) => host !== null))`. Point-free `.map` is safe here because `normalizeBenchmarkDomain` is single-arity, so the index/array arguments `.map` also passes are ignored.
- **suggestion — comment the `{ draft: true }` invariant**: the comment was already at the call site in the reviewed commit, but it opened with the mechanism rather than the rule. It now leads with the invariant by name — a read must target the same view the writes act on — and then explains why the published view lags the draft after a swallowed quota 405.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all prior nits addressed, no new concerns.
**Complexity: HIGH** - large diff; API surface handler change.
_Changes: applies the functional `primaryDomainSet` form, adds the draft-view invariant comment, and merges main (6 files)._

- `primaryDomainSet` functional form: now `new Set(domains.map(normalizeBenchmarkDomain).filter((host) => host !== null))`
- `{ draft: true }` invariant comment: now leads with the rule (reads must target the same view writes act on) and explains the swallowed-405 lag

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 8s | <b>Cost:</b> $3.65 | <b>Commit:</b> <code>1f52fc958cf018a8880b1e41caefca19cbdbe4fb</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2801, opened 2026-07-13
feat: filter GET /organizations/:id/projects by ReBAC-viewable sites

tarunsinghdev:
# feat: filter `GET /organizations/:organizationId/projects` by ReBAC-viewable sites

## Problem

`getProjectsByOrganizationId` returns **every** project in the organization — it
only checks org membership, with no per-user access filter. The
`GET /organizations/:organizationId/sites` endpoint already filters by the
caller's FACS/ReBAC grants, but `/projects` does not. As a result the ASO UI's
site/project picker (which lists via `/projects`) shows all org projects to a
**site-scoped manager** who was granted access to only a subset of sites.

This makes `/projects` consistent with `/sites`.

## Change

Apply the same FACS ReBAC guard used by `getSitesForOrganization` to
`getProjectsByOrganizationId` (`src/controllers/organizations.js`).

When the caller is FACS-enrolled and **resource-scoped** (i.e. does **not** hold
an org-wide `<product>/can_view` JWT permission) **and** `site` is a ReBAC
resource for the product:

1. Fetch the `Set<siteId>` the caller can view from `facs_access_mappings`
   (`listViewableResourceIds`).
2. Load the org's sites, keep those in that set, and collect their
   `getProjectId()`.
3. Return only projects whose id is in that set.

Otherwise (admin / org-wide viewer, or a product that does not ReBAC-scope
`site`) the full project list is returned unchanged.

### Why this is safe / low-risk

- **Reuses existing, tested machinery** — `listViewableResourceIds`,
  `requirePostgrestForFacsMappings`, `isFacsRebacResource` were already imported
  in this file for the `/sites` filter. No new imports, no schema changes, no
  provisioning.
- **Product-scoped** — `isFacsRebacResource(product, 'site')` means it only
  engages for **ASO**. LLMO ReBAC-scopes `brand` (not `site`), so its project
  list is untouched; the state layer holds no per-site grants for it and
  filtering would otherwise wrongly hide everything.
- **Read-only** — narrows a GET response; no writes or new endpoints.
- Same PostgREST-availability guard (`503`) as `/sites`.
- Projects are **not** a ReBAC resource themselves — visibility is inferred from
  member-site visibility, which matches the model (grants are on sites).

## Files

| File | Change |
|---|---|
| `src/controllers/organizations.js` | Add the ReBAC filter to `getProjectsByOrganizationId` (~35 lines). |
| `test/controllers/organizations.test.js` | 4 new tests mirroring the `getSitesForOrganization` FACS suite. |

## Testing

- `eslint` on both files — passes.
- `getProjectsByOrganizationId` suite — **8 passing** (4 existing + 4 new):
  - filters projects to those with a ReBAC-viewable site (ASO),
  - `503` when the FACS flag is set but PostgREST is unavailable,
  - skips the filter under LLMO (site is not a ReBAC resource for LLMO),
  - skips the filter when the JWT carries an org-wide `can_view` grant.

### Example

Request (site-scoped ASO manager who can view only sites under Project 1):

```
GET /organizations/5f3b3626-029c-476e-924b-0c1bba2e871f/projects
x-product: ASO
```

Response — Project 2 (whose sites the caller cannot view) is omitted:

```json
[
  {
    "id": "550e8400-e29b-41d4-a716-446655440000",
    "projectName": "Project 1",
    "organizationId": "5f3b3626-029c-476e-924b-0c1bba2e871f"
  }
]
```

An org-wide viewer or admin receives all projects, exactly as before.

## PR checklist notes

- API **implementation** change only; the response shape is unchanged (same
  `ProjectDto`), so no API-spec update is required — only the set of returned
  items narrows for resource-scoped callers.
- No new audit type; no new endpoint.

## Related Issues

SITES-47075

---

Apply the same FACS ReBAC guard used by getSitesForOrganization to getProjectsByOrganizationId: when the caller is FACS-enrolled and resource-scoped (no org-wide <product>/can_view), restrict projects to those containing at least one site the caller may view. Only engages where `site` is a ReBAC resource for the product (ASO, not LLMO); admins and org-wide viewers get the full list unchanged. Same PostgREST-availability guard as /sites.

Fixes the site/project picker showing every org project to site-scoped managers.

## Related Issues

Thanks for contributing!


src/controllers/organizations.js
```diff
@@ -490,6 +490,43 @@ function OrganizationsController(ctx, env) {
...
 
+    // FACS ReBAC filter (mirrors getSitesForOrganization): when the caller is
+    // FACS-enrolled and resource-scoped (no org-wide `<product>/can_view`),
+    // restrict projects to those containing at least one site the caller may
+    // view. Only applies where `site` is a ReBAC resource for the product (ASO,
+    // not LLMO which scopes `brand`) — otherwise the state layer holds no
+    // per-site grants and filtering would wrongly hide everything.
+    const facs = context.attributes?.facs;
+    const hasFACSCapability = facs?.enabled
+      && context.attributes?.authInfo?.hasFacsPermission?.(`${facs.product.toLowerCase()}/can_view`);
```
ravverma: **Extract the shared FACS guard, do not copy it a third time.** This ~15-line block (compute `facs`, compute `hasFACSCapability`, the tri-part condition, the `requirePostgrestForFacsMappings` guard, and the `listViewableResourceIds` call with its identical argument object) is now duplicated verbatim between `getSitesForOrganization` and here. This is an authorization boundary: the two copies must stay bit-for-bit identical to stay correct, and a future change to the capability string / resource type / guard applied to only one copy becomes a silent visibility bug rather than a test failure - exactly the inconsistency this PR set out to remove. The codebase already set the precedent with `postgrest-availability.js`. Suggest a helper like `resolveViewableSiteIds(context, organization)` returning `null` when no filtering applies or a `Set<siteId>` when it does; both handlers reduce to "if a set was returned, filter; else pass through," and the guard gets a single unit-test target.

test/controllers/organizations.test.js
```diff
@@ -1332,6 +1332,94 @@ describe('Organizations Controller', () => {
       expect(result.status).to.equal(403);
       expect(error).to.have.property('message', 'Only users belonging to the organization can view its projects');
     });
+
+    function fakeFacsPostgrest(rows) {
```
ravverma, line 1336: **Add the zero-viewable-sites case - it is the core anti-leak guarantee and is currently unasserted.** The suite covers "some sites viewable -> subset," but the most common real-world state for a site-scoped manager (grants pending, or granted only in other projects) is "views none of this org's sites," and that path must return `200 []`, never the full list. A test using `fakeFacsPostgrest([])` and asserting `lengthOf(0)` locks in the no-leak behavior and would catch a future regression that defaulted to the full list on an empty grant set.

ravverma (dismissed):
Hey @tarunsinghdev,

## Strengths

- `src/controllers/organizations.js:499-528` is a faithful, near-byte-identical mirror of the already-reviewed `getSitesForOrganization` guard (`:347-365`): same resource-scoped detection, same `503` PostgREST-availability guard, same `listViewableResourceIds` argument shape. Reusing a vetted authorization path rather than inventing a new one is exactly the right instinct for a security-sensitive filter.
- The filter is fail-closed throughout: an empty `viewable` set returns `[]` (not the full list), a missing `hasFacsPermission` method makes the caller resource-scoped so the restrictive filter engages, and PostgREST being unavailable returns `503` rather than silently skipping the filter.
- No caller-controllable input widens visibility. `facs.product`/`subjectId` come from the server-side FACS context, `imsOrgId` from the loaded organization, and the org-membership gate runs before the filter, so there is no IDOR or cross-org path in the diff.
- Correct product-scoping via `isFacsRebacResource(facs.product, 'site')`: LLMO (which ReBAC-scopes `brand`, not `site`) correctly bypasses the filter instead of wrongly hiding every project. `.filter(Boolean)` correctly drops sites with no project.
- The `Site.allByOrganizationId` query is scoped inside the `if` block, so admins and org-wide viewers pay no extra query.
- The four new tests mirror the `/sites` FACS suite (filter-engaged, 503, LLMO-skip, org-wide-`can_view`-skip), and the "filters projects" test genuinely proves exclusion (Project 2 is dropped solely because its only site is absent from the viewable set).

## Issues

### Important (Should Fix)

- **`src/controllers/organizations.js:499-528` - the FACS guard block is now duplicated verbatim across two endpoints.** The whole sequence (compute `facs`, compute `hasFACSCapability`, the tri-part condition, the PostgREST guard, and the `listViewableResourceIds` call with its identical argument object) now exists in both `getSitesForOrganization` and `getProjectsByOrganizationId`. This is an authorization boundary, and the two copies must stay bit-for-bit identical to remain correct. When a future change touches the capability string, resource type, or guard, one copy will drift and the divergence becomes a silent visibility bug, not a test failure - precisely the inconsistency this PR exists to remove. The codebase already endorses this consolidation (`postgrest-availability.js` was extracted for the same reason). Extract a shared helper, e.g. `resolveViewableSiteIds(context, organization)` returning `null` when no filtering applies or a `Set<siteId>` when it does; both handlers then reduce to "if a set was returned, filter; else pass through," and the guard gets a single unit-test target.

- **`test/controllers/organizations.test.js` (new suite) - the core anti-leak case is not asserted: a resource-scoped caller with zero viewable sites should get `200 []`.** The PR's entire purpose is to stop leaking projects to site-scoped managers, and "views none of this org's sites" (grants pending, or granted only in other projects) is the most common real-world state for such a caller. The suite only tests "some sites viewable -> subset returned." A test with `fakeFacsPostgrest([])` asserting `lengthOf(0)` directly locks in the no-leak guarantee and would catch a future regression that defaulted to the full list on an empty grant set. Cheap to add and the highest-value missing test here.

### Minor (Nice to Have)

- **`src/controllers/organizations.js:~521` - projects with zero viewable sites (including projects with no sites at all) are silently invisible to resource-scoped callers.** This follows from the "visibility inferred from member-site visibility" model and is likely intended, but it has a user-visible consequence: a manager who creates a project in the ASO UI and has not attached sites yet will not see it in the picker. Worth a one-line confirmation with product that hiding empty projects (even from their creator) is the intended contract, and if so, a short test asserting it.
- **`src/controllers/organizations.js:521` - the `.filter(Boolean)` branch is untested.** Codecov reports all modified lines covered, but that is line coverage, not branch coverage: no fixture supplies a viewable site whose `getProjectId()` is falsy, so the falsy branch never executes. A one-line fixture (a viewable site with `projectId: null`) would exercise and document it.
- **`src/controllers/organizations.js:516` - the filter narrows against raw org sites, while `/sites` narrows against product-enrolled sites (`filterSitesForProductCode`).** In the normal case these are equivalent because a non-enrolled site should carry no `can_view` grant. The narrow divergence: if a state-layer grant outlives an enrollment removal, a project could surface here that `/sites` would have filtered at the enrollment stage. Low impact (exposes only project id/name for a site the caller holds an explicit grant on), but for exact parity consider narrowing against the product-enrolled site set, or add a one-line comment noting the intentional reliance on grant-level product scoping.

## Recommendations

- `facs.product.toLowerCase()` at `:501` throws if `facs.product` is undefined while `facs.enabled` is true. This is copied verbatim from `getSitesForOrganization` and relies on the `facsWrapper` invariant that `product` is set whenever `enabled`, so it is not a new defect - but if you extract the shared helper, a `facs.product &&` guard would make it crash-safe independent of the wrapper.

- The sibling project-scoped site endpoints are not ReBAC-filtered: `getSitesByProjectIdAndOrganizationId` and `getSitesByProjectNameAndOrganizationId` (and the `projectsController` site listings) return sites unfiltered. Consequence: a resource-scoped caller now gets a correctly filtered project list, but can then call `GET /organizations/:id/projects/:projectId/sites` on any visible project and enumerate all its member sites, including sites they cannot see via `/sites`. This gap predates this PR and belongs in a separate change, but it limits the protection this PR adds and is the logical immediate follow-up.

## Assessment

Ready to merge? With fixes.

Reasoning: The implementation is a correct, low-risk, fail-closed mirror of an already-vetted authorization path, with no bypass or over-visibility in the changed code and matching happy-path test coverage. The two "should fix" items are quality and safety-net improvements rather than behavioral defects: de-duplicating the now-twice-copied authorization guard to prevent future drift, and adding the zero-viewable-sites test that pins the core anti-leak guarantee. The out-of-scope unfiltered sites-by-project endpoints are worth filing as the next change.

## Next Steps

1. Extract the shared FACS guard helper and add the zero-viewable-sites (`200 []`) test - the two Important items.
2. Optionally confirm the empty-project visibility contract and add the `.filter(Boolean)` / product-enrollment parity nits.
3. File a follow-up for the unfiltered project-scoped sites endpoints.

src/controllers/organizations.js
tarunsinghdev: Extracted into `resolveViewableSiteIds(context, organization)` in `src/support/facs-site-visibility.js`, exactly as suggested — returns `null` (no filter), a `Response` (503, PostgREST unavailable), or a `Set<siteId>`. Both `getSitesForOrganization` and `getProjectsByOrganizationId` now call it and reduce to "if a set was returned, filter; else pass through." Added a dedicated unit-test file for the helper too. See 0317d5111.

test/controllers/organizations.test.js
tarunsinghdev: Added: `returns an empty list when the caller can view none of the org's sites`, using `fakeFacsPostgrest([])` and asserting `lengthOf(0)` on the 200 response. See 0317d5111.

MysticatBot (approved):
Hey @tarunsinghdev,

**Verdict: Approve** - clean implementation of a fail-closed ReBAC filter, reusing vetted machinery with good test coverage.
**Complexity: HIGH** - medium diff; API surface + FACS/ReBAC signal.
_Changes: Applies the same FACS ReBAC site-visibility guard used by `/sites` to the `/projects` endpoint, extracted into a shared helper with dedicated unit tests (4 files)._

<details>

- suggestion: `instanceof Response` check in `src/controllers/organizations.js:541` is theoretically fragile across module boundaries (if npm ever installs two copies of `@adobe/fetch`, the check silently returns false). The actual failure mode would be a noisy 500 (Response has no `.has()` method), not a data leak, so this is not a security concern. A duck-type check (`if (viewable?.status)`) or restructuring the helper to throw on unavailability would be more robust. Low practical risk with a lockfile.
- suggestion: The FACS-filtered path at `src/controllers/organizations.js:549` uses an early `return ok(...)` while the unfiltered path falls through to line 557. If pagination or sorting is added later to the unfiltered path, the filtered path would not benefit. Consider assigning back to a `filteredProjects` variable and using a single return point.

</details>

- Duplicated FACS guard block extracted into `resolveViewableSiteIds` in `src/support/facs-site-visibility.js`.
- Zero-viewable-sites test added, asserting fail-closed `200 []` on empty grant set.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 11m 23s | <b>Cost:</b> $9.72 | <b>Commit:</b> <code>b6c1eef0648d25abb0c9227e487a80a2ffcd8d7b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2804, opened 2026-07-13
fix(edge-optimize): derive CloudFront connector externalId from org (server-side), drop client-supplied model

ABHA61:
## Summary

The CloudFront connector-role assume-role **external ID is now derived server-side from the site's org** — it is no longer read from the request body.

## Why

Previously the external ID was minted per session and echoed back by the client on every call (`validateCloudfrontCredentials` required it from the request body). Because it was client-supplied, a caller could pass another org's account id + external id via their own site endpoint and, if that org's connector role trusted the value, assume it. Deriving the external id from the caller's own org means a mismatched target fails at STS — so it must be server-derived, and no per-org secret needs to be stored.

## What changed

- `resolveConnectorExternalId(site)` returns the site's IMS org id; the shared wizard gate resolves it and passes it to every assume-role call.
- `createBootstrapUrl` bakes the org id into the connector-role trust-policy template. It still returns `externalId` in the response so the UI can display it for customers who create the role manually.
- `validateCloudfrontCredentials` no longer reads or requires `externalId` from the request body (the field is ignored if present).

## Migration

A connector role created before this change has a per-session external id in its trust policy; afterwards the server sends the org id, so such a role must be re-created via a fresh bootstrap. The feature is LaunchDarkly-gated, so no production customers are affected.

## Test plan

- [x] Controller + wizard tests pass (163)
- [x] Test asserts the server-derived external id is the value that reaches STS
- [x] `lint` + `type-check` clean


test/controllers/llmo/llmo-cloudfront.test.js
```diff
@@ -487,6 +498,18 @@ describe('LlmoCloudFrontController', () => {
       expect(body.message).to.not.include('db boom');
```
MysticatBot, line 500: issue (blocking): The core behavioral change is that `externalId` now comes from `site.getOrganization().getImsOrgId()` and flows to `assumeConnectorRole`. However, no test asserts what value actually reaches that call. `assumeConnectorRoleStub` is a sinon stub that succeeds regardless of input, so a regression where `externalId` is `undefined` or wrong would pass all 163 tests silently.

The bootstrap test verifies the value in the *response* (`expect(body.externalId).to.equal(TEST_IMS_ORG_ID)`), but the gate-to-assume path in the wizard endpoints has no equivalent assertion.

**Fix:** Add at least one assertion in the `connect` happy-path test:
```javascript
expect(assumeConnectorRoleStub.firstCall.args[0].externalId).to.equal(TEST_IMS_ORG_ID);
```
This completes the security fix story: the test proves the server-derived value (not a client-supplied one) is what STS receives.

MysticatBot (changes requested):
Hey @ABHA61,

**Verdict: Request changes** - one test-level gap in verifying the security fix's core behavioral contract.
**Complexity: HIGH** - medium diff; API surface signal.
_Changes: Derives CloudFront connector-role external ID from the site's IMS org id server-side, eliminating a confused-deputy attack vector where the client could supply another org's external ID (2 files)._

## Must fix before merge

1. [Important] No test asserts the org-derived externalId value reaches `assumeConnectorRole` - `test/controllers/llmo/llmo-cloudfront.test.js:498` (details inline)

<details>

- suggestion: Remove stale `externalId` from test `data` objects (connectContext, distributionsContext, etc.) to reflect the new API contract and prevent masking a future re-read regression - `test/controllers/llmo/llmo-cloudfront.test.js`
- nit: Consider a defensive format check on the IMS org ID before passing to STS (e.g. `/^[A-Z0-9]+@AdobeOrg$/i`) to guard against data corruption - `src/controllers/llmo/llmo-cloudfront.js:70`
- nit: `createBootstrapUrl` still returns `externalId` (now the raw IMS org ID) in the response body - confirm this is intentional given the FE no longer sends it back - `src/controllers/llmo/llmo-cloudfront.js:138`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 39s | <b>Cost:</b> $5.14 | <b>Commit:</b> <code>a204e064b5da0b2ceba79dcd187cd98ccec46ea0</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ABHA61:
Thanks for the review 🙏 — addressed in `f8be9a33`:

- **Blocking (assert the value reaching STS):** added `expect(assumeConnectorRoleStub.firstCall.args[0].externalId).to.equal(TEST_IMS_ORG_ID)` to the connect happy-path test. Good catch — the stub succeeded regardless of input, so the security-critical bit (server-derived value reaches STS) was unproven.
- **Suggestion (stale `externalId` in test data):** removed `externalId` from all wizard test request payloads to reflect the new contract.
- **nit (`createBootstrapUrl` still returns `externalId`):** intentional — the UI displays it so customers who create the connector role manually (rather than via the quick-create link) know the external-id condition to set. Keeping it.
- **nit (format check on the org id before STS):** skipping for now — the org id is server-sourced from our own data (not client input), and STS enforces the trust-policy condition regardless, so a mismatch fails safely there. Happy to add a light `/^[A-Z0-9]+@AdobeOrg$/i` guard if you feel strongly.

MysticatBot (approved):
Hey @ABHA61,

**Verdict: Approve** - prior blocking finding addressed, security model confirmed correct, no new issues.
_Changes: Derives CloudFront connector-role external ID from the site's IMS org server-side, eliminating a confused-deputy attack vector where the client could supply another org's external ID (2 files)._

- Test assertion proving server-derived externalId reaches `assumeConnectorRole` now present (connect happy-path)
- Stale `externalId` removed from all wizard test request payloads to reflect the new API contract

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 9s | <b>Cost:</b> $3.60 | <b>Commit:</b> <code>28d3e5adc8d59ab652e852087f8a67ee83c6dae3</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

nit23uec (approved):

PR #2805, opened 2026-07-13
feat(serenity): Semrush onboarding endpoint (just Slack for now)

dboger-semr:
## What

Adds a new endpoint **`POST /v2/orgs/:spaceCatId/semrush-onboarding`** for IMS users that sends a Slack notification announcing a Semrush onboarding request. It resolves the customer's email from the trusted auth identity and the org's Semrush workspace ID, then posts to a Slack incoming webhook.

This is the first, notification-only slice. A future iteration will call the Semrush (SR) API using Styx authentication via the same notifier seam.

## Why

We want to be alerted when a customer starts onboarding, with the customer's email and (best-effort) their Semrush workspace ID. Email and workspace are taken from **trusted sources** (the validated IMS identity and a DB lookup), never from the request body, so they cannot be forged by the caller.

## How it works

```mermaid
sequenceDiagram
    autonumber
    box rgba(255,180,120,0.15) Semrush
        participant SR as Semrush API
    end
    box rgba(180,200,255,0.15) Adobe
        participant MFE as MFE
        participant SC as spacecat-api
    end
    participant SLK as Slack

    MFE->>SR: request with IMS token
    SR-->>MFE: 403 — non-existing user

    MFE->>SC: POST /v2/orgs/:id/semrush-onboarding (IMS token)
    Note over SC: resolve email from IMS identity<br/>resolve workspace id from org

    alt Current — notify via Slack
        SC->>SLK: webhook: { workspaceId, email }
        SLK-->>SC: 200 OK
    else Future — provision via Semrush (S2S)
        SC->>SR: onboarding request (S2S / Styx auth)
        Note over SR: create / provision user<br/>(onboarding happens here)
        SR-->>SC: 200 — onboarded
    end

    SC-->>MFE: 200 { notified / onboarded }
```

- **Auth:** IMS user. Email = `authInfo.getProfile().trial_email` (the real human email on the IMS path; `email` holds the IMS user-id alias), falling back to `email`.
- **Access control:** `Organization.findById` → `AccessControlUtil.hasAccess(org)`. Order: `404` (org missing) → `403` (no access) → `400` (no email) → notify.
- **Workspace:** `resolveWorkspaceId(ctx, spaceCatId)` returns the org-level ("parent") Semrush workspace ID, or `null` (best-effort — we still notify).
- **Delivery:** `notifyOnboarding` POSTs to `SLACK_ONBOARDING_WEBHOOK_URL` via `tracingFetch` (webhook chosen over the Web API client because the target channel is in the Semrush Slack workspace, not Adobe's). Errors never leak the webhook URL/secret.
- **Responses:** `200 { notified, workspaceId }`; `400/403/404`; `500` (webhook not configured); `502` (webhook failure).

## Files

- `src/support/onboarding/slack-notifier.js` — isolated webhook notifier (future Styx/SR-API seam)
- `src/controllers/onboarding.js` — controller
- `src/routes/index.js`, `src/index.js` — route wiring
- `src/routes/facs-capabilities.js`, `src/routes/required-capabilities.js` — route classified as `INTERNAL_ROUTES` (org-membership gated, not FACS/S2S), mirroring `llmo/onboard-site`
- `docs/openapi/onboarding-api.yaml`, `schemas.yaml`, `api.yaml` — OpenAPI docs


docs/openapi/api.yaml
```diff
@@ -630,6 +632,8 @@ paths:
     $ref: './llmo-api.yaml#/llmo-onboard'
   /v2/orgs/{spaceCatId}/llmo/onboard-site:
     $ref: './llmo-api.yaml#/llmo-onboard-site'
+  /v2/orgs/{spaceCatId}/onboarding:
```
dboger-semr: maybe be explicit about SR in the namespace? should it be part of llmo path prefix?

src/support/onboarding/slack-notifier.js
```diff
@@ -0,0 +1,70 @@
...
+ * incoming webhook. This is the seam where a future Styx-authenticated Semrush
+ * (SR) API call will be added.
+ *
+ * @param {Record<string, string|undefined>} env - Runtime env (context.env).
+ * @param {{ email: string, workspaceId: string|null, spaceCatId: string }} payload
+ * @returns {Promise<void>}
+ * @throws {ErrorWithStatusCode} 500 if the webhook URL is unset; 502 on failure.
+ */
+export async function notifyOnboarding(env, payload) {
+  const webhookUrl = env?.SLACK_ONBOARDING_WEBHOOK_URL;
```
dboger-semr, line 50: Can somebody push this to Vault? Ben Barten can share values for stages and prod

MysticatBot (approved):
Hey @dboger-semr,

**Verdict: Approve** - clean, well-tested feature addition that follows established patterns.
**Complexity: HIGH** - large diff (driven by generated docs/index.html); API surface + FACS classification change.
_Changes: Adds a POST /v2/orgs/:spaceCatId/semrush-onboarding endpoint that sends a Slack notification with the customer's email and workspace ID on onboarding request (15 files)._
Note: Recommend a human read before merge - shared-contract change (OpenAPI spec modified). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: The catch block maps any error without a `.status` property to 502 (bad gateway). If an unexpected error propagates from somewhere other than `notifyOnboarding` (e.g. a future refactor introduces a new await between workspace resolution and notify), it would incorrectly surface as a gateway error rather than 500. Consider `const status = e.status || 500` with an explicit 502 case. - `src/controllers/onboarding.js:22`
- suggestion: No rate limiting protects this endpoint. An authenticated user with org membership could trigger many Slack notifications in rapid succession. Consider a lightweight per-org throttle (even a simple "one request per org per 30s" in-memory check) if the Slack workspace has notification fatigue concerns. - `src/controllers/onboarding.js:38`
- nit: The PR description header says the path is `/v2/orgs/:id/onboarding` but the implementation uses `/v2/orgs/:spaceCatId/semrush-onboarding`. The code is correct (and matches the self-review comment about being explicit), but the description's "What" section could be updated for clarity. - `docs/openapi/api.yaml:651`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 7m 20s | <b>Cost:</b> $1.93 | <b>Commit:</b> <code>5a4ed2902546e27b2a0df4c9ee4d3d17f565371b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

claudiaboldis:
Issues
🔴 Critical: Webhook URL leaks in network-error path (slack-notifier.js)

The notifier catches network errors and builds the message with e.code || e.name || 'network error' — good. But the notifier also throws its own ErrorWithStatusCode with a message like onboarding notification rejected with status 500. The controller then logs that message directly:

log.error(`[onboarding] notification failed for org=${spaceCatId} status=${status}: ${e.message}`);
If a future maintainer changes the notifier to include more context in the error message (e.g. the webhook URL), it will leak into server logs. Low risk today but the pattern is fragile — the controller should log a fixed-format message and not forward e.message from the notifier, or the notifier should guarantee its error messages are safe to log. There's actually a test for this in slack-notifier.test.js (does not leak the webhook URL when the network error message contains it) but it only verifies the thrown error, not what the controller logs.

🟡 Medium: INTERNAL_ROUTES placement breaks alphabetical ordering (required-capabilities.js)

The new route is inserted with a blank line separator in the middle of INTERNAL_ROUTES, breaking the visual grouping/ordering of the array. Minor but inconsistent with the surrounding style.

🟡 Medium: Integration test only covers auth gates — no success path smoke test

test/it/shared/tests/onboarding.js explicitly scopes itself to 404/403 because the success path requires an external Slack webhook. That's a reasonable call for a live integration test, but it means there's no integration-level smoke test verifying the full request reaches the notifier. Worth a comment in the test noting this is by design (it already has one — actually fine as-is).

🟡 Medium: Rate limiting / abuse protection absent

Any org member can POST this endpoint. Since it fires a Slack webhook to an external workspace on every call, a single user can spam the CSM channel. There's no idempotency key, cooldown, or dedup. For a "rare misconfiguration" flow this is probably acceptable MVP scope, but worth a follow-up ticket.

🟢 Low: workspaceId ?? null is redundant (onboarding.js:68)

resolveWorkspaceId already returns null when not found. The ?? null coercion is a no-op — undefined would also be serialized as null in JSON anyway.

🟢 Low: getProfile?.() optional chain is defensive but silent (onboarding.js:52)

If authInfo exists but getProfile is not a function, the optional chain silently returns undefined, email is empty, and you get a 400. This is probably fine but the 400 message (Unable to determine customer email from the request identity) would be confusing — the real problem is a broken auth context, not a missing email. Worth noting in a follow-up or adding a guard that distinguishes the two.

dboger-semr:
## Fixed (commit `d6496822`, pushed; PR description updated)

- **🔴 Critical — webhook URL log-forwarding.** Hardened: the controller now **redacts** the webhook URL from failure logs (`env.SLACK_ONBOARDING_WEBHOOK_URL` → `[redacted]`), and I documented the notifier's safe-message invariant. Crucially, I added the **controller-level test you said was missing** — it stubs the notifier to throw an error whose message *contains* the URL and asserts the logged line doesn't. So it's now enforced at both layers, robust against a future notifier change.
- **🟡 INTERNAL_ROUTES ordering.** Moved the entry out of the middle of the LLMO block; it now sits next to the PLG onboarding block — its closest sibling (IMS self-service, not S2S).
- **🟢 `workspaceId ?? null`` workspace id cannot be undefined, good catch
- **nit — catch status mapping.** Changed to `const status = e.status || 500`. Unexpected errors now surface as 500, not a misleading 502; only the notifier's explicit 502 stays 502. Added a test.
- **nit — stale PR description.** Updated to `/v2/orgs/:spaceCatId/semrush-onboarding`.

Verification: 16 controller/notifier tests (incl. 2 new), 3077 route/capability tests, lint clean.

No changes:
in the OpenAPI schema. If `resolveWorkspaceId` ever returned `undefined`, `JSON.stringify` **drops the key**, violating the contract. `?? null` guarantees it's always present as `null`. It defends the schema — keeping it.
- **🟡 Rate limiting — follow-up ticket on SR side with introducing S2S API.
- **🟢 `getProfile?.()` silent 400.** The auth middleware guarantees `authInfo.getProfile` exists (handler contract); a missing `getProfile` can't occur for a validated request. The 400 is only reached when the profile genuinely lacks an email — where the message is accurate. Distinguishing an unreachable case adds branching for no benefit. Leaving as-is.
- **🟡 IT success-path.** You concluded "fine as-is" — agreed; the by-design comment is present and the success path (external webhook) is unit-tested.

aenascut (commented):
## Strengths

- **Clean architecture.** Notification-only slice isolates Slack delivery from future S2S Semrush API calls via `slack-notifier.js` seam. Clear contract: `notifyOnboarding` throws `ErrorWithStatusCode` with safe messages (never leaks webhook URL).
- **Auth ordering correct.** 404 (org missing) → 403 (no access) → 400 (no email) prevents info leak. `INTERNAL_ROUTES` classification matches `llmo/onboard-site` pattern.
- **Defense-in-depth secret redaction.** Notifier avoids URL leakage via `e.code || e.name`, and controller does an extra defensive `.split()` redaction before logging. Test coverage enforces the invariant.
- **Solid test coverage.** 11 controller cases (success, fallback email, 404/403/400, error mapping, URL redaction); 6 notifier cases (2xx, non-2xx, network failure, "not configured", URL not leaked). Integration tests verify org gate.

---

## Issues

### Issue 1 — Double `Organization.findById` (cold cache: 2 DB round-trips)

`triggerOnboarding` fetches `org` for access control, then immediately calls `resolveWorkspaceId(ctx, spaceCatId)` which hits `Organization.findById(spaceCatId)` again:

```js
// controller
const org = await ctx.dataAccess.Organization.findById(spaceCatId);
...
const workspaceId = await resolveWorkspaceId(ctx, spaceCatId); // redundant fetch
```

Workspace cache protects warm containers, but this endpoint has `org` in hand already. **Fix:** extract workspace directly from `org`:

```js
const workspaceId = typeof org.getSemrushWorkspaceId === 'function'
  ? (org.getSemrushWorkspaceId() ?? null)
  : null;
```

Eliminates the redundant DB fetch on cold cache.

---

### Issue 2 — No rate limiting / per-org cooldown

Each `POST /v2/orgs/:spaceCatId/semrush-onboarding` fires a Slack webhook with no guard. Buggy MFE retry loop or aggressive client can flood Semrush Slack channel. No per-org cooldown, no idempotency key, no dedup. **Fix:** Add short-TTL module-scoped per-org cooldown (similar pattern to workspace cache):

```js
const ONBOARDING_COOLDOWN_MS = 30_000; // 30s per org
const onboardingCooldown = new Map();

// in triggerOnboarding:
const lastNotified = onboardingCooldown.get(spaceCatId);
if (lastNotified && Date.now() - lastNotified < ONBOARDING_COOLDOWN_MS) {
  return ok({ notified: false, workspaceId, reason: 'recently notified' });
}
onboardingCooldown.set(spaceCatId, Date.now());
```

---

### Issue 3 — Silent `trial_email → email` fallback, no log

```js
const email = profile?.trial_email || profile?.email;
```

When `trial_email` is absent, Slack message contains IMS alias (`ABC123@AdobeID`), not a human email. No log marks this fallback. Ops won't know if the email is real or an alias. **Fix:**

```js
const email = profile?.trial_email || profile?.email;
if (!profile?.trial_email) {
  log.warn(`[onboarding] using alias email fallback for org=${spaceCatId}`);
}
```

---

### Issue 4 — Unused `context` constructor param (nit)

```js
export default function OnboardingController(context, log, env) {
  // context never referenced
```

JSDoc says `@param {object} context - Request context (dataAccess, attributes, env, log)`, but implementation ignores it. Matches `ElementsController` pattern so not wrong, but JSDoc is misleading. **Fix:** Clarify in JSDoc that the injected `context` is unused; per-request data comes from `ctx` param.

---

## Verdict

**Approve with minor fixes.** Ship 1, 2, 3; nit 4 is optional. Architecture is sound, slice is tight. Issue 2 (cooldown) should land before prod to avoid operational noise.

dboger-semr:
1, 2, 4 fixed in #fb7e5e0

I won't be fixing 3 because we anyway *will* receive this email on the slack side, so we know that it's an alias email with or without logs. If we will figure out that we need more information, we are likely to just add it to the slack message and not logs.

dboger-semr:
@aenascut is there a process on how to merge it to upstream / resolve review comments? To deploy it to production we also need to put there a token https://semrush.slack.com/archives/C0B02FWQYBF/p1784106430781829?thread_ts=1784023159.716429&cid=C0B02FWQYBF

aenascut:
I've set the secrets in vault; 
as far as I know, after merge it will go to prod after some e2e are run

PR #2807, opened 2026-07-14
fix(slack): forward mode to non-prerender audit types (LLMO-6167)

anagarwa:
## Summary

- `run-audit`'s keyword-parsing path unconditionally deleted `mode` from the audit `data` payload sent to the audit-worker, regardless of audit type.
- That was harmless while `mode` only meant something for `prerender` — prerender re-injects a normalized `mode` value itself via `buildEffectiveData()` for its own API-side AI-only suggestion selection.
- TOC's `mode:ai-only` (LLMO-6167) has no equivalent API-side selection logic — it relies entirely on the audit-worker receiving `mode` directly in `data`. Because `run-audit.js` never special-cased TOC, `mode:ai-only` was silently dropped before the SQS message was even built, so `run audit {site} audit:toc mode:ai-only` fell through to a normal full audit instead of the prompts-only path.
- Fix: only delete `mode` for `prerender`; let it pass through untouched for every other audit type.

## Test plan

- [x] Added a unit test confirming `mode` now survives for a non-prerender audit type alongside `generatePrompts`
- [x] All existing `run-audit.test.js` tests pass (62/62), including all prerender-mode tests (`mode:all`, `mode:ai-only`, `mode:ai-only-current`, `mode:ai-only-missing`) — unchanged behavior confirmed
- [x] Verified against real production Splunk logs: `run audit asianpaints.com audit:toc generatePrompts:true mode:ai-only` previously logged `auditData="{"generatePrompts":"true"}"` (mode missing) at the api-service layer, before this fix

Refs: LLMO-6167

### DIFF
--- a/src/support/slack/commands/run-audit.js
+++ b/src/support/slack/commands/run-audit.js
@@ -334,10 +334,18 @@ function RunAuditCommand(context) {
         [baseURLInputArg] = positionalArgs;
         auditTypeInputArg = keywords.audit;
 
-        // Build audit data from remaining keywords (excluding 'audit' and 'mode')
+        // Build audit data from remaining keywords (excluding 'audit').
+        // 'mode' is also excluded here ONLY for prerender — prerender does its own
+        // suggestion selection/batching on the API side (see PRERENDER_MODES /
+        // buildEffectiveData below) and re-injects a normalized `mode` value itself.
+        // Other audit types (e.g. toc's mode:ai-only, LLMO-6167) have no API-side
+        // selection logic of their own — mode must pass through untouched so the
+        // audit-worker can act on it directly.
         const auditDataKeywords = { ...keywords };
         delete auditDataKeywords.audit;
-        delete auditDataKeywords.mode;
+        if (auditTypeInputArg === PRERENDER) {
+          delete auditDataKeywords.mode;
+        }
 
         auditDataInputArg = Object.keys(auditDataKeywords).length > 0
           ? JSON.stringify(auditDataKeywords)
--- a/test/support/slack/commands/run-audit.test.js
+++ b/test/support/slack/commands/run-audit.test.js
@@ -582,6 +582,22 @@ describe('RunAuditCommand', () => {
       });
     });
 
+    it('forwards mode in audit data for non-prerender audit types (e.g. toc\'s mode:ai-only, LLMO-6167)', async () => {
+      const command = RunAuditCommand(context);
+
+      await command.handleExecution(['validsite.com', 'audit:geo-brand-presence', 'generatePrompts:true', 'mode:ai-only'], slackContext);
+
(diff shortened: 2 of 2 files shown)


MysticatBot (approved):
Hey @anagarwa,

**Verdict: Approve** - clean, minimal bug fix with appropriate test coverage.
**Complexity: LOW** - small bug fix, single-service, 2 files.
_Changes: conditionalizes the `mode` keyword deletion in `run-audit` so non-prerender audit types (e.g. TOC) receive `mode` in their SQS payload (2 files)._

<details>

- suggestion: Add a companion test asserting `mode` is excluded from `auditData` for prerender audit type - this covers the other branch of the new conditional and guards against someone removing the `delete` later - `test/support/slack/commands/run-audit.test.js:582`
- suggestion: Use `.calledOnce` instead of `.called` on `sqsStub.sendMessage` to guard against accidental double-sends, which would make the `firstCall` assertion ambiguous - `test/support/slack/commands/run-audit.test.js:588`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 29s | <b>Cost:</b> $3.92 | <b>Commit:</b> <code>bf7cd704e2e292da08266ec0025d49a6bb77e84c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/slack/commands/run-audit.js
```diff
@@ -334,10 +334,18 @@ function RunAuditCommand(context) {
...
+        // 'mode' is also excluded here ONLY for prerender — prerender does its own
+        // suggestion selection/batching on the API side (see PRERENDER_MODES /
+        // buildEffectiveData below) and re-injects a normalized `mode` value itself.
+        // Other audit types (e.g. toc's mode:ai-only, LLMO-6167) have no API-side
+        // selection logic of their own — mode must pass through untouched so the
+        // audit-worker can act on it directly.
         const auditDataKeywords = { ...keywords };
         delete auditDataKeywords.audit;
-        delete auditDataKeywords.mode;
+        if (auditTypeInputArg === PRERENDER) {
```
dipratap, line 346: @anagarwa i noticed from slackbot we send mode also in the run audit prerender command, will there be any impact to that command run with this change?

anagarwa: So earlier it keywords were sent to all the audits. Then specific handling was added to prerender which deleted mode from keywords. Prerender audit add it back here https://github.com/adobe/spacecat-api-service/blob/main/src/support/slack/commands/run-audit.js#L126
but that made `mode` missing from other audits. My PR fixes it to default handling and keeping prerender specific changes as it is

dipratap (approved):

PR #2812, opened 2026-07-14
fix(serenity): stop the RELEASE_ALLOCATION zero-transfer silent no-op

byteclimber:
## Summary

LLMO-6189: `RELEASE_ALLOCATION` (`{ ai: { projects: 0, prompts: 0 } }`) was sent via `transferWorkspaceResources` at 4 call sites — brand decommission (`decommissionBrandWorkspace`), the `ensureSubworkspace` concurrency-loser cleanup, and two `brand-provisioning.js` failure-cleanup paths — to hand a sub-workspace's entire AI resource carve back to the parent pool. A live probe against the Semrush gateway confirmed a transfer that sets a dimension to **zero is silently ignored** (2xx, no units moved). Every one of these "release" operations was logging false success while permanently stranding the released allocation on the parent pool — a live, active resource leak (with a flat carve of 1 project / 500 prompts against a pool of 13 projects / 800 prompts, two brand deactivations exhaust more than the entire prompt pool, forever).

## Design decision

Full reclamation of an allocation is only possible via workspace **DELETE** — a transfer cannot express "reclaim to zero" (this is also why the already-shipped `releaseAiSurplus` explicitly refuses to ever emit a to-zero transfer, instead reporting `requires-decommission`).

Rather than silently reversing the deliberate "never delete a sub-workspace in production" policy — documented independently in three places as a Semrush-CS-owned deprovisioning boundary, not just an engineering default — this PR:

- **Reuses the existing `SERENITY_ALLOW_WORKSPACE_DELETE` flag** (already the sole gate on `deleteWorkspace`, default OFF in every deployed environment) as a single capability switch, rather than minting a second flag for the same underlying vendor operation.
- **Flag off (today's default everywhere):** all 4 sites still empty the workspace's projects, but the allocation release now logs **loudly** that it cannot be reclaimed via transfer and remains stranded, instead of falsely reporting success. This closes the "lying" half of the bug unconditionally, with zero behavior-visible change to production today.
- **Flag on (an explicit, per-environment opt-in an operator must make):** all 4 sites uniformly delete every project first, then delete the (now-empty) sub-workspace outright via the new `releaseFullAllocation()` — a genuine reclaim. Every delete path is `assertNotParent`-guarded (previously only `decommissionBrandWorkspace` had this guard — a captured/adopted id that turns out to be the parent must never be deleted) and idempotent (an upstream 404 on an already-gone workspace is convergent success).

New exports in `workspace-lifecycle.js`:
- `deleteAllProjects` — extracted, shared project-emptying loop (404-as-success).
- `releaseFullAllocation` — the release decision point, mirroring `releaseAiSurplus`'s "never lie about success" contract.

`brand-provisioning.js`'s two failure-cleanup paths and `serenity.js`'s 5 relevant call sites thread `allowDelete` through from the same env flag. `releaseProvisionedWorkspace` gained a `spaceCatId` param so it can resolve the org parent workspace for the `assertNotParent` guard (previously missing at that call site).

Design plan was reviewed by a `senior-staff-engineer` (architecture — recommended collapsing to a single flag and a dedicated constant-return helper rather than reusing `releaseAiSurplus`'s round-trip) and a `senior-sre` (production safety — flagged that `assertNotParent` was missing at 3 of the 4 sites and must guard every delete path). Both are incorporated.

## Scope note: the audit deliverable

The ticket also asks to "audit how many units are already stranded across existing Serenity orgs." I do not have live Semrush/IMS credentials in this environment to run that audit safely, so it is **not included in this PR**. Recommend a follow-up: a read-only reporting script (mirroring the style of existing live-gateway canary scripts) that an operator with credentials can run against dev/stage/prod to quantify existing stranding before deciding whether/when to flip `SERENITY_ALLOW_WORKSPACE_DELETE` on anywhere.

## Test plan

- [x] `npm run lint` — clean (2 remaining errors are pre-existing, unrelated missing-package issues also present on `main`)
- [x] `npm test` — 14618 passing, 0 new failures (7 pre-existing failures, confirmed present on `main` via `git stash` comparison — unrelated missing `@adobe/spacecat-shared-akamai-client` / `@adobe/spacecat-shared-ticket-client` packages and an unrelated `suggestions.test.js` issue)
- [x] New unit test coverage for `deleteAllProjects`, `releaseFullAllocation`, both `allowDelete` states at all 4 call sites, and the `assertNotParent` guard on the newly-threaded `releaseProvisionedWorkspace` path
- [ ] IT suite (`test/it/postgres/serenity.test.js`) — blocked in this sandbox: the shared `node_modules` install is missing `@adobe/spacecat-shared-ticket-client`, which the IT harness's full app bootstrap requires (pre-existing gap, unrelated to this change — confirmed the same failure occurs against `main`). CI should be able to run it cleanly with a proper `npm ci`.

LLMO-6189


MysticatBot (approved):
Hey @byteclimber,

**Verdict: Approve** - well-scoped bug fix with thorough test coverage, honest error reporting, and sound defense-in-depth design.
_Changes: Replaces the silently-failing zero-transfer allocation release with an honest two-outcome pattern (genuine reclaim via workspace delete when the flag is on, loud stranded-allocation logging when off) across all 4 call sites. (10 files)._

<details>

- nit: `deleteAllProjects` returns `projects.length` (listing count) rather than the number of actually-attempted deletes - items skipped for blank IDs inflate the logged `deletedProjects` metric - `src/support/serenity/workspace-lifecycle.js:270`
- suggestion: `releaseProvisionedWorkspace` signature inserts `spaceCatId` positionally between `workspaceId` and `log` - any caller passing `(context, id, log)` in the old shape would silently bind the logger to `spaceCatId`. One caller exists (correctly updated), but an options bag would be more defensive for a function whose params can all be strings or objects. - `src/support/serenity/brand-provisioning.js:228`
- nit: Three independent one-liner predicate definitions for the same `SERENITY_ALLOW_WORKSPACE_DELETE === 'true'` check across `serenity.js`, `brand-provisioning.js`, and `rest-transport.js` - a shared constant for the env var name would prevent a future typo from silently weakening the gate - `src/support/serenity/brand-provisioning.js:22`

</details>

<details>
<summary>Recommendations</summary>

- When `allowDelete === true`, consider requiring `parentWorkspaceId` to be non-empty (fail-closed). Currently, if a future caller reaches `releaseFullAllocation` with `allowDelete: true` but `parentWorkspaceId` undefined, the `assertNotParent` guard is inert and the transport-layer gate is the only remaining protection against deleting the org parent. All current callers pass it correctly and the PR documents this limitation explicitly, but making it required on the delete path would close the defense-in-depth gap permanently.
- Confirm that the `SERENITY_ALLOC releaseFullAllocation: cannot reclaim allocation` warn-level log flows into your alerting pipeline. If `warn` is filtered by default in production, the stranding signal might not reach a human - `error` would be safer given this represents an unresolved resource leak (even if expected under the current flag-off default).

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 39s | <b>Cost:</b> $6.80 | <b>Commit:</b> <code>2c73d1d1282fdb1e8f8e0adc712fb36e1b02ba59</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

byteclimber:
## Live e2e verification (adobe-hackathon.semrush.com sandbox)

Ran `scripts/serenity-decommission-release-canary.mjs` against a real throwaway sub-workspace under a live master workspace:

- **Case 1 (`allowDelete: false`)** — `releaseFullAllocation` returns `{released:false, reason:'requires-decommission'}` and the workspace is confirmed still present, untouched. ✅
- **Case 2 (`allowDelete: true`)** — returns `{released:true, reason:'deleted'}`, the workspace is confirmed genuinely gone (`getWorkspaceStatus` 403s), and the master pool's `prompts.total` is confirmed restored to its pre-carve value (initial read showed it stale-low, but that was async-reconciliation lag — a follow-up read a few seconds later confirmed full restoration).

Both branches of the fix behave exactly as designed against the live gateway. No stray resources left behind.

rainer-friederich:
@byteclimber, we got told we should never delete the sub workspace. The idea to allocate 0 resources and unlink it is from me.

I learned that resources are not given back to the parent upon deletion - I guess this is the reason we should not delete them.

What about setting `1` instead of `0` as resources?

byteclimber:
Good catch — implemented your suggestion. Pushed in \`398862100\` (+ canary update in \`e8d5c0416\`):

\`releaseFullAllocation\` no longer deletes anything. It transfers the workspace's AI allocation down to a small non-zero floor (default \`{projects:1, prompts:1}\`, overridable) — never to zero (the proven silent no-op this whole fix exists to route around), and never via workspace delete. Dropped the \`allowDelete\`/\`SERENITY_ALLOW_WORKSPACE_DELETE\` threading from every lifecycle call site (\`workspace-lifecycle.js\`, \`brand-provisioning.js\`, \`markets-subworkspace.js\`, \`controllers/serenity.js\`) — \`deleteWorkspace\` itself stays in \`rest-transport.js\`, just scoped explicitly to test/smoke-cleanup only now, no longer reachable from any production lifecycle path.

This also matches the design doc's own language once I went back and read it closely: *"Decommissioned brands never give their carve back. Only deleting the workspace does that, and production never deletes... If the pool becomes tight, the cheapest fix is a one-shot operator script issuing a lowering transfer to a non-zero floor"* — i.e. the doc already prescribes exactly this fix.

Live-verified end to end on \`adobe-hackathon.semrush.com\` (canary rewritten to match): the workspace is confirmed to still exist afterward, settles at exactly the \`{1,1}\` floor, and the master pool total reflects exactly \`carved - floor\` — the floor unit stays reserved on the child by design (immediate headroom for its next reactivation) rather than round-tripping through a fresh create. Full unit suite green (14,867 passing), lint + type-check clean.

MysticatBot (dismissed):
Hey @byteclimber,

**Verdict: Request changes** - two issues to address before merge: a docs/code divergence and a missing floor invariant.
_Changes: Replaces the proven-broken zero-transfer allocation release with a non-zero floor transfer (`releaseFullAllocation`) across all 4 call sites, adds shared `deleteAllProjects` helper, and includes a live-gateway canary script. (10 files)._

## Must fix before merge
1. [Important] docs/serenity.md describes "flag on = delete" behavior that does not exist in production code - `docs/serenity.md:267` (details inline)
2. [Important] `releaseFullAllocation` accepts a zero-value floor without validation, reintroducing the exact bug this PR fixes - `src/support/serenity/workspace-lifecycle.js:290` (details inline)

<details>

- nit: `deleteAllProjects` returns `projects.length` (total listing count including blank-id items) rather than the count of actually-attempted deletes, inflating the `deletedProjects` metric logged by callers - `src/support/serenity/workspace-lifecycle.js:275`
- suggestion: Add a unit test for `releaseFullAllocation(transport, SUB_WS, undefined, log)` to pin the behavior that `assertNotParent` does not fire when the parent workspace id is unknown - `test/support/serenity/workspace-lifecycle.test.js`
- nit: rest-transport.js `deleteWorkspace` comment block still references "fall back to a 'log the stranding, don't delete' behavior when the flag is off" which no longer describes any lifecycle code path - `src/support/serenity/rest-transport.js:236`

</details>

- Repeated `SERENITY_ALLOW_WORKSPACE_DELETE === 'true'` one-liners across 3 files: no longer applies, the predicate is now a single `const allowWorkspaceDelete` in rest-transport.js and the lifecycle paths no longer branch on it.

docs/serenity.md
```diff
@@ -264,7 +264,7 @@ linked Site per distinct market domain.)
...
+- **The sub-workspace shell's fate depends on `SERENITY_ALLOW_WORKSPACE_DELETE`** (unset — off — in every deployed env today). This is the SAME flag that gates the test-cleanup/net-zero-smoke workspace delete; decommission reuses it rather than minting a second flag for the same underlying capability (LLMO-6189). With the flag **off**: deactivate empties every project and releases as much of the allocation as a transfer can move — which, for a FULL release, is **nothing** (a transfer that sets a dimension to 0 is a silently-ignored no-op against the Semrush gateway, confirmed by a live probe). The empty shell remains in the parent family, its allocation permanently stranded against the parent pool, until an operator flips the flag or Semrush CS deprovisions it manually. This is logged loudly (not as a false success) so the stranding is observable. With the flag **on**: deactivate deletes the (already project-emptied) workspace outright, genuinely returning its full carve to the parent pool. A future activate allocates a fresh sub-workspace either way.
```
MysticatBot: issue (blocking): This text states "With the flag on: deactivate deletes the workspace outright" but this behavior does not exist in the production code. The deactivate handler calls decommissionBrandWorkspace which uses deleteAllProjects + releaseFullAllocation. No lifecycle path calls deleteWorkspace or branches on SERENITY_ALLOW_WORKSPACE_DELETE. That flag gates only the transport-level deleteWorkspace() method (test/smoke cleanup only). Suggested fix: rewrite to describe actual behavior - deactivate empties projects and lowers allocation to a non-zero floor (1 project, 1 prompt), returning the surplus to the parent pool.

src/support/serenity/workspace-lifecycle.js
```diff
@@ -234,6 +245,80 @@ async function adoptFromFamily(transport, parentWorkspaceId, title, log) {
...
+ * ever reclaimed by Semrush CS), so this does NOT delete anything. It transfers the workspace's
+ * `{projects, prompts}` totals down to `floor` — a NON-ZERO target, which resizes reliably
+ * (live-verified; the same mechanism `releaseAiSurplus` already uses for ordinary rightsizing).
+ * Only a transfer that sets a dimension to exactly ZERO is a silent no-op against the gateway —
+ * this function's whole point is to never send that payload.
+ *
+ * `assertNotParent` guards the call — a captured/adopted id that turns out to be the org parent (a
+ * gateway bug; see the create-path comment on `assertNotParent`) must never have its allocation
+ * touched.
+ *
```
MysticatBot, line 290: issue (blocking): releaseFullAllocation accepts a caller-supplied floor option with no guard against { projects: 0, prompts: 0 }. This is exactly the payload the entire PR exists to eliminate (a zero-transfer is a proven silent no-op against the gateway). No current production caller passes a custom floor today, but the function is exported and documented as accepting overrides. A future caller passing zero values would silently reintroduce the allocation-stranding bug. Suggested fix: add a runtime assertion at the top of the function body (before assertNotParent) that throws if either floor dimension is <= 0. This codifies the invariant the PR establishes as a runtime check, matching the defensive style of assertNotParent.

docs/serenity.md
byteclimber: Fixed in c00244e82 — rewrote the caveat to describe actual behavior (deactivate empties projects, lowers AI allocation to the non-zero floor, surplus returned to the parent pool; the workspace shell itself is never deleted).

src/support/serenity/workspace-lifecycle.js
byteclimber: Fixed in c00244e82 — added a runtime assertion at the top of releaseFullAllocation that throws (500) if either floor dimension is <= 0, before assertNotParent runs and before any transport call. Added a unit test covering both dimensions.

MysticatBot (approved):
Hey @byteclimber,

**Verdict: Approve** - both prior findings addressed cleanly; only two stale-comment nits remain.
_Changes: Replaces the proven-broken zero-transfer allocation release with a non-zero floor transfer (`releaseFullAllocation`) across all 4 call sites, adds shared `deleteAllProjects` helper, runtime floor validation, and a live-gateway canary script. (10 files)._
Note: Recommend a human read before merge - this change modifies `docs/serenity.md` (a project doc describing the Serenity workspace lifecycle contract). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: `decommissionBrandWorkspace` JSDoc still says "leaving the sub-workspace empty and unowned (or, with the flag on, gone)" - the parenthetical contradicts the PR's core invariant (production never deletes); remove it - `src/support/serenity/workspace-lifecycle.js:531`
- nit: `deleteAllProjects` JSDoc says "before releaseFullAllocation can safely delete the (now-empty) workspace itself" but `releaseFullAllocation` lowers to a floor, never deletes - reword to "...can safely lower the (now-empty) workspace's allocation to the floor" - `src/support/serenity/workspace-lifecycle.js:252`

</details>

- docs/serenity.md caveat now accurately describes actual behavior (empties projects, lowers to non-zero floor, never deletes).
- `releaseFullAllocation` now rejects floor with dimensions <= 0 via runtime assertion, with unit test coverage for both dimensions.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 51s | <b>Cost:</b> $5.08 | <b>Commit:</b> <code>c00244e823b5aff8fa96c2ac0ef8589051e2a2a9</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2813, opened 2026-07-14
feat(serenity): flag-flip readiness gates for dynamic AI allocation (LLMO-6190)

byteclimber:
## Summary

Collects the correctness gates that must land before `SERENITY_DYNAMIC_ALLOCATION`
(the JIT Semrush AI resource allocator, dormant since #2764) can be flipped on for
any org. Ref: [LLMO-6190](https://jira.corp.adobe.com/browse/LLMO-6190). Operational
readiness is tracked separately in LLMO-6191.

**Item 1 (fail-loud on missing workspace ids) — verified already done in #2764**
(`createHeadroomGuard`'s `requireWorkspaceId` throws when `enabled` and either
id is missing). No changes needed; struck from scope per the ticket.

**Item 2 — per-brand ceiling.** `ensureAiHeadroom` already enforces a per-dim
`ceiling` (409 `brandAiLimit`), but nothing ever passed one. Added
`DEFAULT_BRAND_AI_CEILING` (`resource-manager.js`) and made it the default
`ceiling` in `createHeadroomGuard`. No product/sizing number exists yet, so the
value is deliberately astronomical (`{ projects: 1_000_000, prompts: 1_000_000_000 }`
— far above any org's whole Semrush pool) so it is *effectively non-enforcing*:
the org-pool-exhaustion path (`orgPoolExhausted`, 409) remains the real gate. This
satisfies the DoD's "a per-brand ceiling in force" without inventing a business
number or a data-model change. Heavily documented in code as a placeholder;
**note for reviewers**: an internal design review raised the alternative of
shipping the constant unwired (no default enforcement at all) until a real number
exists, on the grounds that a second failure mode (even a practically-unreachable
one) is one more than zero. I kept it wired per the ticket's explicit DoD wording,
but flagging the dissent here in case product/reviewers prefer the more
conservative unwired form.

**Item 3 — release on delete.** `releaseAiSurplus` was only wired at the
model-removal seam. `handleDeleteMarketSubworkspace` now calls it too
(`failFast: true`, same floor/blocks, best-effort), after the delete + mapping
tombstone, gated on the flag. Brand deactivate needed **no change** — it already
fully releases via the pre-existing `decommissionBrandWorkspace` path
(`transferWorkspaceResources` to zero), independent of this flag; wiring
`releaseAiSurplus` there too would have been redundant.

**Item 4 — consume the metered-405 classifier.** `isMeteredQuota` was implemented,
unit-tested, and unused. `HeadroomGuard` gains `retryOnQuota(fn)`: a pure
passthrough when the flag is OFF, and ON, exactly one re-read + top-up + retry
cycle when a publish 405s as a disguised metered-quota rejection — any other
error, or a second failure after the retry, propagates untouched. Wired at:
- `handleCreateMarketSubworkspace`'s two `publishProject` calls (require + best-effort modes)
- `publishAffected` (shared with flat mode) via an optional `wrapPublish` param
  (default identity — flat mode is byte-for-byte untouched), used by
  `handleCreatePromptsSubworkspace`
- `syncModelsForProject` (shared with flat mode), same `wrapPublish` approach,
  used by `handleUpdateModelsSubworkspace`

**Item 5 — live-gateway canary.** No live Semrush/IMS credentials exist in this
environment, so I could not run this myself. Added
`scripts/serenity-metered-405-canary.mjs` — a standalone script that drives the
real transport against a throwaway sub-workspace, drains its prompt allocation,
and publishes to intentionally trigger the disguised 405, printing the raw
status/body. **Needs a human to run it**:
```
IMS_TOKEN=$(mysticat auth token --ims) \
SUB_WORKSPACE_ID=<a throwaway/dev sub-workspace id> \
node scripts/serenity-metered-405-canary.mjs
```
Until that happens, `isMeteredQuota` (`errors.js`) remains a substring guess
("quota" / "allocation exhausted") rather than a pinned contract test.

## Test plan
- [x] `npm run lint` — clean
- [x] `npm test` — 14,876 passing, 0 failing
- [x] Live-container IT suite: `npx mocha --require test/it/postgres/harness.js --timeout 60000 test/it/postgres/serenity.test.js` — 41 passing, 0 failing
- [x] New/extended unit coverage: ceiling-default regression test, full `retryOnQuota` edge-case matrix (OFF passthrough, non-metered error, retry-succeeds, retry-still-fails, recovery-ensure-itself-throws, lock-scope), release-on-delete matrix (OFF, ON+valid ids with literal floor assertion, upstream-404 branch, not-found early-return branch, best-effort swallow, genuinely-unexpected-error propagation), and handler-level end-to-end retry wiring for all three metered-publish call sites


byteclimber:
Reopening to retrigger MysticatBot's review — its previous run crashed on a cost-budget cap (error_max_budget_usd), not a real finding.

MysticatBot (approved):
Hey @byteclimber,

**Verdict: Approve** - well-scoped flag-flip readiness work with thorough test coverage and clean kill-switch isolation.
**Complexity: HIGH** - large diff, API surface changes across multiple handlers.
_Changes: Wires four pre-flip correctness gates (per-brand ceiling, release-on-delete, metered-405 retry, canary script) for the dormant SERENITY_DYNAMIC_ALLOCATION flag (12 files)._

<details>

- suggestion: `retryOnQuota` on the disabled guard returns `(fn) => fn()` (synchronous) while the enabled path is `async` - making the disabled path `async (fn) => fn()` removes a footgun if a caller ever passes a sync fn - `src/support/serenity/dynamic-allocation-active.js:97`
- suggestion: Consider naming the placeholder constant more obviously (e.g. `PLACEHOLDER_BRAND_AI_CEILING`) so the non-enforcing semantics are visible at the use site without reading the 15-line doc comment - `src/support/serenity/resource-manager.js:61`
- suggestion: Define a named `WrapPublish` type alias (e.g. `type WrapPublish = <T>(fn: () => Promise<T>) => Promise<T>`) and reference it in the JSDoc for `syncModelsForProject` and `publishAffected` - prevents drift as more seams adopt the pattern - `src/support/serenity/handlers/prompts.js:237`
- suggestion: Document in `retryOnQuota`'s JSDoc that the wrapped fn must be an operation rejected at the gateway gate (pre-check 405), not one that partially applies before failing - this is the invariant that makes the single retry safe - `src/support/serenity/dynamic-allocation-active.js:108`

</details>

byteclimber:
## Live e2e verification (adobe-hackathon.semrush.com sandbox)

Ran `scripts/serenity-metered-405-canary.mjs` against a real throwaway sub-workspace/project. Along the way, fixed two real bugs in the canary itself (pushed in `cfb23f545`): it sent a placeholder int for `language_id` where the gateway expects a catalog UUID, and it published before drafting anything, so publish always no-op'd instead of exercising the metered path.

**Finding (not yet actioned in code — flagging for visibility):** drafting a prompt is **not** free/unmetered when `prompts.total === 0`, even though `used`/`drafted` are both 0 — it fails outright. Reproduced consistently (30+ attempts, curl and the typed transport, multiple tags/projects), and confirmed it self-resolves the moment `total` is topped back above 0. The rejection is a bare nginx `405 Not Allowed` HTML page with no distinguishing "quota"/"allocation exhausted" text — `isMeteredQuota` (errors.js:92) cannot classify it, and its own test suite explicitly treats a bare `{message:'Method Not Allowed'}` 405 as the *legitimate*, non-quota case. There's no body-content signal that separates the two.

Practical implication: any workspace whose `prompts.total` drops to 0 (e.g. an incomplete release, or before LLMO-6191's sweep tops it back up) can't even draft — silently, with an error the current classifier reads as non-actionable. This is a gap in the draft-seam fronting (only publish is fronted with a headroom check today), not a one-line fix — raising it here for awareness/follow-up rather than scope-creeping this PR.

rainer-friederich:
## Live verification on LLMO-Dev-2 (dev) — the metered-405 is real, and the prompt top-up is fronting the wrong seam

Ran the disguised-405 probe end to end against the dev parent workspace `LLMO-Dev-2` (`bb0f4e1c`, pool `projects 0/10`, `prompts 0/1200`), with a **control arm** so a 405 is attributable. Both arms hit the real Semrush gateway via the shipped transport paths.

### What was run

| Arm | Sub-workspace | prompts (used/total) | `POST …/aio/prompts` (the draft write) |
|---|---|---|---|
| Control (full headroom) | `580c7891` | 8 / 200 | write 1 → **201, landed** |
| Test (over headroom) | `82ba2cb0` | 16 / 18 | write 10 → **405, nothing landed** |
| Boundary | `82ba2cb0` | 16 / 18 | write exactly 2 → **201**; then 1 more → **405** |

### Confirmed facts

- **The write endpoint works at headroom.** The control write returns `201` with the list wrapper. This refutes the earlier canary finding in `cfb23f545` ("`POST …/aio/prompts` returns a bare nginx 405 regardless of quota state, reproduced with full headroom") — on the current tenant the 405 appears **only** when over quota, so it is a genuine quota signal, not an environment/routing gap.
- **Prompts meter at the WRITE, not at publish.** The `405` fires inside `createPromptsByIds`, before any publish, and nothing lands. Projects are the dimension that meters at publish.
- **The gate is cumulative and drafts count toward it.** At `16/18`, a 2-item write succeeds (16+2 = 18); the very next 1-item write `405`s (16+2+1 = 19 > 18). So the ceiling is `used + drafted + batch > total` — `used + drafted` is the correct denominator, and drafted prompts are **not** free against the write gate (even though drafted is free for billing).
- **The `/resources` `drafted` counter is stale on this dev tier** — it reads `0` even with drafts in flight — yet enforcement uses the true draft count. Sizing must not trust that field; sizing from the batch about to be written (`includeDrafted`) is the right approach.
- **The disguised-405 body has no quota text** (this is the fixture LLMO-6190 item 5 asked for):

```
<html>
<head><title>405 Not Allowed</title></head>
<body>
<center><h1>405 Not Allowed</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

### What does NOT work today

1. **Inert everywhere.** `SERENITY_DYNAMIC_ALLOCATION` is OFF in every env, so `createHeadroomGuard.ensure` is a no-op and no top-up happens at all. A sub-workspace at its prompt ceiling `405`s on the next write with no recovery.

2. **Even switched ON, the prompt guard is on the wrong side of the write.** In the batch path the order is write → top-up → publish:
   - `handlers/prompts-subworkspace.js:168` `createOnePrompt` → `createPromptsByIds` (the metered draft write)
   - `handlers/prompts-subworkspace.js:222` `headroom.ensure({}, { includeDrafted: true })` (the top-up)
   - `handlers/prompts-subworkspace.js:225` `publishAffected`

   The top-up runs *after* the writes have already `405`ed. Worse, a `405` write is caught and recorded as `failed`, so it never enters `affectedProjectIds`, and the top-up is `affectedProjectIds.length > 0`-guarded (`prompts-subworkspace.js:216`) — so when a batch is entirely over quota the top-up is **skipped**. The guard only protects prompts that already succeeded.

   The create-market path has the identical shape: `handlers/markets-subworkspace.js:276` writes prompts, then `:566` tops up, then `:574` publishes.

3. **`isMeteredQuota` cannot classify the real body.** `support/serenity/errors.js:97` matches only bodies containing `"quota"` or `"allocation exhausted"`. The captured body contains neither, so the disguised-405 is currently read as a non-quota error.

### What still needs to be implemented

- **Move the prompt headroom guard before `createPromptsByIds`**, sized on `used + drafted + batch` for the batch about to be written — in both `prompts-subworkspace.js` and the create-market prompt writes in `markets-subworkspace.js`. Guarding before publish is the wrong seam for prompts; the write is the metering point.
- **Fix the classifier** so it recognizes the pinned body. The only signal separating the two cases is shape — a bare nginx `text/html` 405 (gateway-level quota rejection) versus a genuine app-level JSON `{ message: "Method Not Allowed" }` 405 — not any substring. Pin the captured body as a fixture and key `isMeteredQuota` on that shape, with a contract test asserting it does not absorb the JSON method-not-allowed case.

### What is already correct

The **project** dimension is fronted at the right seam: `markets-subworkspace.js:430` `headroom.ensure({ projects: 1 })` runs before `createProject` (`:431`), and the absolute top-up persists through to `publishProject` (`:574`) where projects meter. Project drafts are free at zero headroom; only publish consumes. No change needed there.

byteclimber:
Both fixed and pushed in \`1bf36398b\`, on top of the canary fixes from \`cfb23f545\`:

**1. Front the write, not just publish.** Both prompt-write seams now call \`headroom.ensure(...)\` *before* the create-loop, sized on the real batch:
- \`prompts-subworkspace.js\`: sized on \`inputs.length\` (the incoming batch — a safe upper bound; some inputs may still skip on validation/missing-project, which just means a slightly early top-up, never an under-provision).
- \`markets-subworkspace.js\`'s \`generateAndAttachPrompts\`: sized on \`texts.size\`, the *actual* generated prompt count, now known before the per-type-bucket write loop — threaded via the caller's existing \`headroom\` guard.

The old post-write, pre-publish \`ensure({}, {includeDrafted:true})\` call in \`prompts-subworkspace.js\` is gone — it's now redundant given the pre-write front, and was exactly the wrong-seam call your finding pointed at. \`retryOnQuota\` around \`publishAffected\`/\`publishProject\` stays as the bounded top-up+retry safety net for anything that still slips through (e.g. a raced concurrent write on the same child).

**2. \`isMeteredQuota\` now keys on body SHAPE, not content**, per your exact framing: a string body is the disguised gateway-level rejection, a JSON object is a genuine app-level error (this gateway's real Method-Not-Allowed always comes back as JSON). Pinned your captured fixture (the bare nginx page) as the test's representative disguised-405 body, and added an explicit contract assertion that a JSON \`{message: 'Method Not Allowed'}\` — and even a JSON \`{message: 'Quota exceeded'}\` — are *not* absorbed, matching your framing exactly.

Also found and fixed the blast radius: two existing test helpers (\`quota405\` in \`dynamic-allocation-active.test.js\` and \`dynamic-allocation-fronting.test.js\`) built the OLD JSON-shaped fake 405, which broke under the corrected classifier — updated both to the real string-shaped fixture. Added net-new coverage for the write-time fronting on both handlers (previously zero coverage of \`dynamicAllocation: true\` existed on either prompt-write path). Full suite green (14,878 passing), lint + type-check clean.

Confirmed the project dimension needed no change — your note that it's already correctly fronted at \`createProject\`/\`publishProject\` matches what's in the code.

rainer-friederich (approved):
Hey @byteclimber,

**Verdict: Approve** - the four flag-flip correctness gates are cleanly implemented behind a genuinely inert OFF path, the live-verified 1bf36398b fixes hold up against the code, and only non-blocking polish remains.
_Changes: wires four pre-flip correctness gates for the dormant SERENITY_DYNAMIC_ALLOCATION flag (default per-brand ceiling, release-on-market-delete, pre-write prompt headroom fronting with a bounded metered-405 top-up-and-retry, and a shape-keyed disguised-405 classifier pinned by a live-captured fixture) plus a live-gateway canary script (16 files)._

<details>
<summary>Non-blocking (8): minor issues and suggestions</summary>

- nit: `headroom?.ensure(...)` optional-chains a required operation; the sole call site always passes a guard (OFF is a no-op object, never undefined), and a future call site omitting it would silently skip fronting with the flag ON - the exact silent-skip class this PR eliminates. Make it plain `headroom.ensure(...)` so a wiring omission fails loud - `src/support/serenity/handlers/markets-subworkspace.js:275`
- nit: the canary's header/tail comments still tell the reader to narrow `isMeteredQuota` away from "the current substring guess", which this PR already replaced with the shape-keyed classifier and pinned fixture; reword to "re-pin the fixture and re-verify the shape guard if the gateway body ever changes" - `scripts/serenity-metered-405-canary.mjs:56`
- nit: the canary imports from `src/support/serenity/rest-transport.js` but carries no `// @ts-check` and `scripts/` is outside the tsconfig include; add both per `src/support/serenity/CLAUDE.md`, or explicitly scope that convention to exclude standalone ops scripts - `scripts/serenity-metered-405-canary.mjs:1`
- suggestion: guard the canary's destructive drain (transfer of `prompts.total` to zero on an env-supplied `SUB_WORKSPACE_ID`) behind an explicit confirmation flag or a non-prod assertion instead of a doc comment - `scripts/serenity-metered-405-canary.mjs:140`
- suggestion: add a transport-level contract test asserting error-body shape by content type (string for text/html, parsed object for JSON), so a `rest-transport.js` refactor cannot silently flip the shape-keyed classifier from retrying real quota-405s to absorbing genuine ones - `src/support/serenity/errors.js:100`
- suggestion: add a handler-level test that an org-pool-exhausted fronting `ensure` aborts the request before any `createPromptsByIds` write runs, pinning the fail-loud contract - `src/support/serenity/handlers/prompts-subworkspace.js:154`
- suggestion: on the invited Item 2 question - keep the ceiling wired: it keeps the `brandAiLimit` 409 path live and test-covered, while an unwired constant would rot as dead code. Two cheap hardenings: state the invariant in a comment ("values must exceed the largest org pool", otherwise the placeholder silently starts enforcing a fabricated cap) and track the real number in a ticket rather than only a code comment - `src/support/serenity/resource-manager.js:78`
- suggestion: the `.gitignore` comment ("V1 to V2 migration scripts and test data") no longer describes the canary entry it now covers - `.gitignore`
</details>

<details>
<summary>Out of scope, worth tracking</summary>

- Older seams still treat ANY 405 as a quota rejection (`brand-provisioning.js` surfaces "Quota exceeded"; `brand-urls.js` `republishBestEffort` and the create-market best-effort catch swallow it), diverging from the new shape-keyed classifier - consolidation ticket material.
- A non-quota bare-HTML 405 on a metered write now triggers exactly one bounded top-up + retry when the flag is ON (self-correcting, no spurious allocation); worth one line in the LLMO-6191 pilot-flip runbook/monitoring so it is not mistaken for a bug.
- The canary prints raw gateway bodies and internal URLs to the operator console - correct for its human-run purpose, but needs redaction if ever adapted into an automated path.
- `docs/specs/2026-06-15-serenity-subworkspace-dual-mode-implementation-plan.md` says `isWorkspaceNotReady` was dropped during implementation, but `errors.js` still exports and uses it - pre-existing doc/code drift in the file this PR touches.
</details>

byteclimber:
Fixed the two nits from your review, pushed in \`189391d1e\`:
- \`generateAndAttachPrompts\`'s \`headroom\` param is now required (not optional-chained) — a future caller that forgets to thread it fails loud instead of silently skipping metering.
- Reworded the canary's header/tail comments, which still described \`isMeteredQuota\` as a substring guess after it was already fixed to be shape-based.

Separate note for visibility (not from your review, an independent audit flagged it): two of the six correctness gates in serenity-docs#22 §8 aren't closed by this PR — the per-brand ceiling still has no binding value (still astronomical/non-enforcing, as the code itself documents), and nothing confirms any consumer (elmo UI, Slack, direct API) actually retries a `503 workspaceBusy`. Both are called out as open in the doc itself, so this isn't a new problem, but flagging so the PR isn't read as closing all six gates.

byteclimber:
Status update on the two open correctness gates from serenity-docs#22 §8 that this PR doesn't close (flagged in my last comment, following up with what I could actually verify):

**Gate 3 (per-brand ceiling binding value):** confirmed still open — \`DEFAULT_BRAND_AI_CEILING\` in \`resource-manager.js\` is still the astronomical placeholder, documented in its own comment as non-enforcing. This is explicitly a product decision per the doc, not something I can close from this side — needs a real number from whoever owns per-brand sizing.

**Gate 6 (callers retry \`503 workspaceBusy\`):** checked directly rather than leaving it as a guess. Went into \`project-elmo-ui\` and traced every \`serenity/*\` call site (\`src/api/spacecat.ts\`) back to \`handleApiResponse\` — it throws an \`ApiError\` on any non-2xx status with no retry anywhere in the chain, and no caller wraps these calls in retry logic either. So this is confirmed genuinely open, not just theoretical: today a \`503\` on an ordinary serenity write is a hard customer-visible failure in elmo UI. I haven't touched that repo — it's a separate PR/team decision whether to add client-side retry there, but wanted the actual state confirmed rather than assumed.

Not blocking this PR's own merge (both gates predate it and are explicitly called out as open in the design doc), but wanted them tracked accurately rather than silently implied as done.

PR #2814, opened 2026-07-14
fix(llmo): read popularity rationale from topics table, not Mystique S3

nsdere:
## Problem
`GET /sites/:siteId/llmo/rationale` reads a **decommissioned Mystique S3 file** (`spacecat-<env>-mystique-assets/llm_cache/<siteId>/prompts/topics_popularity_reasoning_cache.json`). Nothing writes that file anymore — popularity/rationale generation moved into DRS (Jan-2026 BP Fargate migration), which stores it in DynamoDB and now lands it in Aurora on the `topics` table. Post-migration sites (e.g. shiro-shiro) have no S3 file, so the llmo.now popularity panel shows *"There is no popularity rationale information for this topic."*

## This PR
Repoints `handleLlmoRationale` to read from the `topics` table via PostgREST (`Site.postgrestService`), scoped to the site's organization:
- Selects `name, popularity_volume, popularity_reasoning, updated_at`.
- Derives the popularity label from `popularity_volume` (`-30/-20/-10 → High/Medium/Low`).
- Returns the same item shape the UI expects (`topic, reasoning, popularity, volume, added_date`).
- Drops all S3 access.

**Behavioral note:** `topics` is org-scoped with no category/region column, so the handler matches on **topic name** only (mandatory) plus the optional `popularity` filter. `category`/`region`/`origin` query params are still accepted (backwards compatible) but no longer applied.

## Depends on
- **mysticat-data-service** #791 — adds `topics.popularity_volume` / `popularity_reasoning` + the upsert RPC (must be deployed first).
- **DRS** adobe-rnd/llmo-data-retrieval-service#2689 — emits the popularity sidecar so the import fills those columns.
- One-off backfill of existing sites from the DRS DynamoDB cache.

## Tests
Rewrote the `getLlmoRationale` suite from S3 mocks to a PostgREST client mock (10 tests): mapping, topic partial match, popularity filter, null-reasoning exclusion, category/region ignored, empty result, missing-topic 400, no-PostgREST 400, site-not-found 404, PostgREST-error 400. `eslint` clean.

> Jira: LLMO-XXXX (pending).


src/controllers/llmo/llmo-rationale.js
```diff
@@ -53,77 +53,83 @@ function filterTopics(topics, filters) {
...
 export async function handleLlmoRationale(context) {
-  const { log, s3, env } = context;
+  const { log, dataAccess } = context;
   const { siteId } = context.params;
 
-  // Extract query parameters
-  const {
-    topic, category, region, origin, popularity,
-  } = context.data;
+  // Extract query parameters (category/region/origin accepted for backwards
```
MysticatBot: issue (blocking): **Redundant `Site.findById` call doubles the database round-trip.**

The calling wrapper `getLlmoRationale` invokes `getSiteAndValidateLlmo(context)` which already calls `Site.findById(siteId)` and returns the site object. After validation succeeds, it calls `handleLlmoRationale(context)` which calls `Site.findById(siteId)` again. Every rationale request now makes two identical site lookups.

The old S3-based code did not need the site object (it only needed `siteId` as a string for the S3 key), so this was not a problem before. Now that you need `site.getOrganizationId()`, pass the already-fetched site through context or as a parameter.

The existing pattern in the codebase is for `getSiteAndValidateLlmo` to return `{ site, config, llmoConfig }` - attach `site` to `context` before calling the handler, or accept it as a parameter.

src/controllers/llmo/llmo-rationale.js
```diff
@@ -53,77 +53,83 @@ function filterTopics(topics, filters) {
...
     return badRequest('topic parameter is required');
   }
 
-  if (!s3 || !s3.s3Client) {
-    return badRequest('S3 storage is not configured for this environment');
+  const { Site } = dataAccess;
+  if (!Site?.postgrestService) {
+    log.error('LLMO rationale requires PostgREST (DATA_SERVICE_PROVIDER=postgres)');
+    return badRequest('Rationale data is not available. PostgreSQL data service is required.');
   }
```
MysticatBot, line 110: issue (blocking): **Unbounded query with no server-side filtering risks silent truncation.**

This query fetches ALL topics for an organization with no `.limit()` and no server-side topic filter. Two problems:

1. PostgREST imposes a configurable `max-rows` limit (often 1000). If an org exceeds that threshold, the response silently returns a subset with no error. Since the `topic` filter is applied in-memory after fetch, the matching topic might not be in the returned page.

2. Other PostgREST queries in this codebase (e.g. `llmo-brand-presence.js`) use `.limit(QUERY_LIMIT)` (5000). This query deviates from established convention.

**Fix:** Push the mandatory `topic` filter and null-reasoning exclusion to PostgREST:

```js
const { data, error } = await client
  .from('topics')
  .select('name,popularity_volume,popularity_reasoning,updated_at')
  .ilike('name', `%${topic}%`)
  .not('popularity_reasoning', 'is', null);
```

This eliminates silent truncation, reduces payload size, and aligns with codebase conventions.

src/controllers/llmo/llmo-rationale.js
```diff
@@ -53,77 +53,83 @@ function filterTopics(topics, filters) {
...
+  const { log, dataAccess } = context;
   const { siteId } = context.params;
 
-  // Extract query parameters
-  const {
-    topic, category, region, origin, popularity,
-  } = context.data;
+  // Extract query parameters (category/region/origin accepted for backwards
+  // compatibility but not applied — see note below).
+  const { topic, popularity } = context.data;
```
MysticatBot, line 99: issue (blocking): **Missing null guard on `organizationId` produces a silent empty result.**

If `site.getOrganizationId()` returns `null` or `undefined` (e.g. a site record that predates org assignment), the PostgREST query `.eq('organization_id', null)` will match zero rows. The handler silently returns `[]` rather than surfacing the actual problem.

An org-less site should produce an explicit error:

```js
const organizationId = site.getOrganizationId();
if (!organizationId) {
  return badRequest(`Site ${siteId} is not linked to an organization`);
}
```

MysticatBot (changes requested):
Hey @nsdere,

**Verdict: Request changes** - three blocking issues with the PostgREST query pattern.
**Complexity: HIGH** - large diff (dominated by test rewrite); API surface.
_Changes: Repoints the LLMO rationale endpoint from a decommissioned Mystique S3 file to the topics table via PostgREST (2 files)._

Note: CI checks are currently failing (it-postgres, codecov patch) - resolve before merge.

## Must fix before merge

1. [Important] Redundant `Site.findById` call doubles the database round-trip - `src/controllers/llmo/llmo-rationale.js:80` (details inline)
2. [Important] Unbounded query with no server-side filtering risks silent truncation by PostgREST row limit - `src/controllers/llmo/llmo-rationale.js:93` (details inline)
3. [Important] Missing null guard on `organizationId` silently returns empty results instead of an error - `src/controllers/llmo/llmo-rationale.js:82` (details inline)

<details>

- nit: PostgREST `error.message` returned verbatim to the client may leak internal schema details (table/column/constraint names) - return a generic message and keep the detail server-side only - `src/controllers/llmo/llmo-rationale.js:96`
- suggestion: Add test cases for `volumeToPopularity` with `-20` (Medium), `-10` (Low), and an unmapped value (N/A) to cover all branches - `test/controllers/llmo/llmo.test.js`
- suggestion: Assert PostgREST query parameters (table name, select columns, eq args) in at least one test to catch accidental regressions in the query construction - `test/controllers/llmo/llmo.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 47s | <b>Cost:</b> $4.31 | <b>Commit:</b> <code>fe4e4e0d73e26856289de4536c7e7e2de9afc2b9</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

nsdere:
Addressed in `5b0612b`. The historical backfill of `topics.popularity_*` is complete, so this reads straight from the DB (no legacy S3 fallback).

**Must-fix:**
1. ✅ Redundant `Site.findById` — the handler now reuses the site resolved by `getSiteAndValidateLlmo` (passed in); it only looks up standalone. Test asserts `findById` is called once.
2. ✅ Unbounded query / truncation — the topic filter is now server-side (`.ilike('name', '%<topic>%')`) plus `.not('popularity_reasoning','is',null)`, so the result set is bounded (no PostgREST row-cap truncation).
3. ✅ Missing `organizationId` guard — returns 400 instead of a silent empty result.

**Non-blocking:**
- ✅ No verbatim PostgREST error to the client — returns a generic `Error retrieving rationale`; detail stays in the server log.
- ✅ Added `volumeToPopularity` branch coverage (High/Medium/Low/N/A).
- ✅ Added a test asserting the PostgREST query construction (from/select/eq/not/ilike args).

Coverage should now clear the patch gate (added the query-assertion + branch/guard tests).

nsdere:
@MysticatBot re-requesting review — all three must-fix items are resolved in `5b0612b` (reuse validated site / no second findById; server-side ILIKE + non-null filter to bound the query; organizationId guard) plus the three non-blocking nits (generic error message, volume→popularity branch tests, PostgREST query-arg assertions). The earlier verdict was against `fe4e4e0`.

MysticatBot (changes requested):
Hey @nsdere,

**Verdict: Request changes** - one blocking issue with unescaped ILIKE wildcards in user input.
**Complexity: HIGH** - large diff (dominated by test rewrite); API surface.
_Changes: Repoints the LLMO rationale endpoint from a decommissioned Mystique S3 file to the topics table via PostgREST (3 files)._

Note: CI checks are currently pending - verify they pass before merge.

## Must fix before merge

1. [Important] ILIKE pattern metacharacters not escaped in user-supplied `topic` - `src/controllers/llmo/llmo-rationale.js:103` (details inline)

<details>

- nit: `badRequest` returned for a PostgREST server-side error should arguably be a 500-level response since the failure is not the client's fault - `src/controllers/llmo/llmo-rationale.js`
- suggestion: Add an explicit `.limit()` to the PostgREST query for defense-in-depth against unexpectedly broad matches - `src/controllers/llmo/llmo-rationale.js`
- suggestion: Add integration tests for the modified endpoint per the CLAUDE.md rule "New or modified endpoints must include integration tests in test/it/" - the data source change from S3 to PostgREST warrants end-to-end validation against the test harness
</details>

- Redundant `Site.findById` call - handler now reuses the validated site passed in
- Unbounded query - topic filter is now server-side (ILIKE + non-null reasoning filter)
- Missing `organizationId` guard - returns 400 when null
- PostgREST error leaked verbatim to client - now returns generic message
- Volume-to-popularity branch coverage - all branches tested
- PostgREST query construction assertions - added

src/controllers/llmo/llmo-rationale.js
```diff
@@ -53,77 +53,93 @@ function filterTopics(topics, filters) {
...
-      Key: s3Key,
-    });
-
-    const response = await s3.s3Client.send(command);
-    const fileContent = await response.Body.transformToString();
+  // Reuse the site resolved by getSiteAndValidateLlmo when provided (avoids a
+  // second Site.findById round-trip); fall back to a lookup for standalone callers.
+  const resolvedSite = site ?? await Site.findById(siteId);
+  if (!resolvedSite) {
+    return notFound(`Site not found: ${siteId}`);
```
MysticatBot, line 117: issue (blocking): **ILIKE pattern metacharacters not escaped in user-supplied `topic`.**

The `topic` query parameter is interpolated directly into the ILIKE pattern (`%${topic}%`) without escaping `%` and `_` metacharacters. While PostgREST parameterizes the SQL (no injection risk), the LIKE wildcards are still interpreted within the pattern:

- `topic=%` produces `%%%` which matches every row, removing the filter entirely
- `topic=_` matches any single-character name
- This defeats the "bounded result set" intent stated in the comment above

This codebase already has the fix. `escapeIlike()` exists in `src/support/categories-storage.js` and escapes `\\`, `%`, and `_`. The sibling `llmo-brand-presence.js` uses a similar `buildSearchPattern` utility for the same purpose.

**Fix:**
```js
import { escapeIlike } from '../../support/categories-storage.js';
// ...
.ilike('name', `%${escapeIlike(topic)}%`);
```

Consider also adding a length check on `topic` (the sibling brand-presence handler enforces MIN/MAX_SEARCH_QUERY_LENGTH) and a test case verifying that `%` and `_` in the topic input are treated as literals.

nsdere:
Addressed the new must-fix in the latest commit: ILIKE metacharacters (`%`, `_`, `\`) in the user-supplied `topic` are now backslash-escaped so they match literally (test added). Also took the two nits — added an explicit `.limit()` on the query, and switched the PostgREST-error response from 400 → 500 (server-side failure isn't the client's fault; detail stays in the log). All rationale tests green.

MysticatBot (approved):
Hey @nsdere,

**Verdict: Approve** - all prior blocking findings resolved; no new issues at Important or above.
**Complexity: HIGH** - large diff (dominated by test rewrite); API surface.
_Changes: Repoints the LLMO rationale endpoint from a decommissioned Mystique S3 file to the topics table via PostgREST, with ILIKE escaping, bounded queries, and proper error responses (3 files)._

Note: CI check `it-postgres` is still pending - verify it passes before merge.

<details>

- nit: `filterTopics` retains dead branches for `topic`/`category`/`region`/`origin` that can never fire since the only call site now passes `{ popularity }` - `src/controllers/llmo/llmo-rationale.js:38`
- suggestion: `escapeLikePattern` is a local re-implementation of the existing `escapeIlike` from `src/support/categories-storage.js` (same regex) - consider importing the shared utility for DRY - `src/controllers/llmo/llmo-rationale.js:29`
</details>

- Redundant `Site.findById` call - handler now accepts and reuses the validated site
- Unbounded query - server-side ILIKE + non-null reasoning filter + `.limit(1000)` bounds the result
- Missing `organizationId` guard - returns 400 when null
- PostgREST error leaked verbatim to client - now returns generic 500, detail in server log
- Volume-to-popularity branch coverage - all branches tested (High/Medium/Low/N/A)
- PostgREST query construction assertions - added
- ILIKE pattern metacharacters not escaped - `escapeLikePattern` added with test

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 8m 29s | <b>Cost:</b> $4.42 | <b>Commit:</b> <code>0d5665bb4d7ae7bd90758eae644527f36dba8099</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2815, opened 2026-07-14
feat: serenity apis to reflect tags changes in the elements api

vivesing:
# Semrush Elements Tag Format Migration (`:` → `__`)

## Problem

Semrush's Elements API changed its tag encoding:

- **Old**: `prefix:value`, with `Parent__Child` nesting inside the value (e.g. `category:Living Room__Sofas`)
- **New**: `prefix__value`, with nesting also delimited by `__` (e.g. `category__Living Room__Sofas`)
- **New behavior**: bare `prefix`-only rows (e.g. a lone `category` entry with no value) are now emitted as pure declarations and must be ignored, not treated as tag data.

## Scope

Confirmed as a **global** change — not limited to a single endpoint. Every place that builds or parses these tags was updated.

### Core parsing logic
- **`src/support/elements/definitions/topics.js`**
  - `extractByPrefix` / `splitParent` now split on `__` instead of `:`.
  - Bare prefix-only values (no `__` at all) are ignored entirely — no longer routed into the generic `tags` array.
  - `id` remains the untouched original raw value.

### Outgoing filter builders
- **`owned-urls.js`, `domain-urls.js`, `cited-domains.js`** — now send `category__<label>` instead of `category:<label>` as the `CBF_tags` filter value.
- **`prompts.js`, `elements.js`** — doc/comment updates only (tags pass through as opaque strings; no logic change).

### Documentation
- **`docs/openapi/serenity-api.yaml`**, **`schemas.yaml`*o the `__` format and the "bare declarations ignored"rule. `docs:lint` still passes.

### Tests
- Rewrote `topics.test.js`; adjusted the `__proto__`-prefally unreachable under `__`-splitting); updated`elements-service.test.js`, `prompts.test.js`, `elements.test.js` fixtures.

## Verification
- Ran the real ~90-value production dataset through the actual functions — all correctly parsed, nothing dropped/duplicated/mis-split.
- Full suite: **14,870 passing, 0 failing**. Lint clean.

## Affected APIs

| Endpoint | Impact |
|---|---|
| `url-inspector/filter-dimensions` | Direct parse of changed format |
| `url-inspector/cited-domains` | Outgoing `category__<la
| `url-inspector/owned-urls` | Outgoing `category__<label>` filter |
| `url-inspector/domain-urls` | Outgoing `category__<labe
| `prompts` | Opaque tag pass-through |
| `weeks` | Not affected |

### DIFF
--- a/docs/openapi/schemas.yaml
+++ b/docs/openapi/schemas.yaml
@@ -11551,12 +11551,12 @@ SerenityFilterDimensionItem:
   description: |
     A single id/label option for a Serenity Elements filter dropdown. `id` is
     the original, unmodified tag value returned by Semrush (including its
-    `prefix:` and any `Parent__Child` encoding, e.g. `topic:Furniture__Sofas`).
-    `parent_id`/`parent_label` are present only when the source tag encoded a
-    "Parent__Child" hierarchy; `parent_id` is the same `prefix:` reconstructed
-    over just the parent segment (e.g. for
-    `category:Living Room Furniture Retail__Living Room Furniture and Sofas`,
-    `parent_id` is `category:Living Room Furniture Retail`).
+    `prefix__` and any further `parent__child` encoding, e.g.
+    `topic__Furniture__Sofas`). `parent_id`/`parent_label` are present only
+    when the source tag encoded a "parent__child" hierarchy; `parent_id` is
+    the same `prefix__` reconstructed over just the parent segment (e.g. for
+    `category__Living Room Furniture Retail__Living Room Furniture and Sofas`,
+    `parent_id` is `category__Living Room Furniture Retail`).
   required: [id, label]
   properties:
     id: { type: string }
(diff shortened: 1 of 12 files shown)


MysticatBot (approved):
Hey @vivesing,

**Verdict: Approve** - clean separator migration with good test coverage; one behavioral expansion worth confirming.
_Changes: Migrates Semrush Elements tag encoding from `prefix:value` to `prefix__value` across all parsing, filter-building, and test code (12 files)._

<details>

- suggestion: `transformIntentsToFilterDimensions` and `transformOriginsToFilterDimensions` now emit `parent_id`/`parent_label` fields (via `splitParent`) where previously they returned flat `{ id, label }` only. This is an additive response shape change for the `page_intents` and `origins` dimensions. The OpenAPI schema already allows these optional fields, so no schema break occurs, but downstream UI consumers that assumed these dimensions were always flat may be surprised. Confirm this hierarchy support is intentional for intent/source tags (not just a side effect of applying `splitParent` uniformly) and note it in the PR description. - `src/support/elements/definitions/topics.js:104`
- nit: Bare prefix declarations (values without `__`) are silently dropped with no observability. If the upstream format ever regresses or a new pattern emerges, all tags would vanish without trace. A `log.debug` when values are skipped would make future debugging easier. - `src/support/elements/definitions/topics.js:189`
- nit: A value starting with `__` (e.g. `__proto__evil`) produces an empty-string prefix key in the result object (`result['']`). The test correctly asserts this is safe (no prototype pollution), but downstream consumers iterating `Object.keys(result)` will see `''` as a category. Consider whether this edge case warrants filtering or documenting. - `test/support/elements/definitions/topics.test.js:380`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 35s | <b>Cost:</b> $8.24 | <b>Commit:</b> <code>7156b404653479e98c7e9c5361f9ae96205c4b4e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2824, opened 2026-07-15
feat(llmo): register recurring DRS prompt-suggestion schedules at onboarding

dzehnder:
## Summary

Phase 3 of the DRS prompt-suggestion schedules + onboarding-triggers effort (`local/drs-prompt-suggestions-schedules-onboarding-plan.md`). Wires three new onboarding triggers into the V2 onboarding path so newly onboarded LLMO sites get a **recurring** DRS schedule (plus an immediate first run) for each prompt-suggestion pipeline.

New helpers next to `triggerBrandalfOnboardingJob` in `src/controllers/llmo/llmo-onboarding.js`, each guarded by `drsClient.isConfigured()`:

| Helper | `provider_id` | Cadence |
|---|---|---|
| `triggerSemrushPromptJob` | `prompt_generation_semrush` | `twice_monthly` |
| `triggerCitationAttemptJob` | `prompt_generation_agentic_traffic` | `twice_monthly` |
| `triggerSyntheticPersonasJob` | `prompt_generation_synthetic_personas` | `quarterly` |

- `prompt_generation_agentic_traffic` is the provider behind "Citation Attempts" (no standalone citation-attempt provider — the name won't grep, so there's a code comment).
- Each uses a **single** `drsClient.createSchedule({ ..., triggerImmediately: true })` (no separate `submitJob` + `createSchedule`).
- Invoked from the **V2 block after the Brandalf trigger** (which chains base prompt generation on completion, LLMO-4258) — not fired in parallel before brand data exists (plan Phase 0.3). The immediate run is best-effort; the durable outcome is the recurring schedule row.
- `enableBrandPresence` defaults **false** (plan Phase 0.4).

### Error handling — split semantics
- **Immediate first run: best-effort.** If it produces nothing (e.g. brand/base-prompt data not present yet) the next scheduled run self-heals.
- **`createSchedule` (schedule registration): NOT best-effort.** A failure means the site never gets a recurring schedule and nothing self-heals, so it is logged at **ERROR** with `site_id` + `provider_id` + `status` and surfaced via `say`/ops — never silently swallowed. Onboarding still succeeds (mirrors the brand-activation side-effect handling in `brands.js` `activateBrand`, which logs full context before we mirrored it).

## Cross-repo dependency (deploy ordering)

Depends on a **new additive** `createSchedule(...)` method in `@adobe/spacecat-shared-drs-client` (separate spacecat-shared PR, not yet released). This PR bumps the dependency to **`1.14.0`**.

**This PR must merge/release after the shared drs-client `createSchedule` PR is published as 1.14.0.** Until then, a fresh `npm ci` / bundle build cannot resolve `1.14.0` and CI install will be red — this is the intended cross-repo sequencing, not a defect. Unit tests do **not** hard-depend on the real method: they mock the DRS client (as existing onboarding tests do).

### Confirmed `createSchedule` signature (from the shared PR author)

```js
drsClient.createSchedule({
  siteId,                       // required, string
  providerIds,                  // required, string[] non-empty (one element per single-provider pipeline)
  cadence,                      // required, SCHEDULE_CADENCES value: 'twice_monthly' | 'quarterly'
  description,                  // optional, length-capped (1024)
  enableBrandPresence = false,
  providerParameters,           // optional per-provider passthrough
  priority = 'HIGH',            // 'HIGH' | 'LOW'
  metadata,                     // optional
  triggerImmediately = false,
  timeout,                      // optional ms
}) // => Promise<{ scheduleId: string, alreadyExisted: boolean }>  (409 dedup => alreadyExisted:true)
```

Key contract points this PR honors:
- `providerIds` is an **array** (the `job_config.provider_ids` envelope), so each helper passes a one-element array.
- cadence enum values use **underscores** (`twice_monthly` / `quarterly`); the client derives the cron server-side (`frequency:'cron'`, per-site hour jitter for `twice_monthly`) — callers cannot pass raw cron. Kept as local literals matching `SCHEDULE_CADENCES` so the module loads against the currently-installed client; can be swapped for the imported const once 1.14.0 is installed.
- **No `orgId`/`imsOrgId` params** — DRS derives the tenant-isolation key from `site_id` server-side and *rejects* any caller-supplied `imsOrgId` (at any nesting/case). This PR deliberately does not thread `imsOrgId` into `createSchedule`.

## Tests

`test/controllers/llmo/llmo-onboarding.test.js` (mock DRS client; no real `createSchedule` dependency):
- each helper registers the correct `providerIds` (array) + cadence with `triggerImmediately: true`, `enableBrandPresence: false`, and **no** `imsOrgId`/`orgId`;
- each helper is a no-op (no `createSchedule`) when `isConfigured()` is false;
- a `createSchedule` failure propagates to the caller (not swallowed at the helper);
- V2 `activateBrandAndGeneratePrompts` registers all three schedules after the Brandalf trigger;
- a `createSchedule` failure is logged at ERROR with `site_id` + `provider_id` + `status` while onboarding still completes.

`npm run lint` clean. All 11 new tests pass. The full `llmo-onboarding.test.js` file is 142 passing / 1 failing, where the single failure is a **pre-existing** `settleWithin` "before all" hook timeout that also fails on clean `main` (environment-related, unrelated to this change).

## Spec

Design spec: [`docs/specs/2026-07-16-prompt-suggestion-schedules-onboarding.md`](https://github.com/adobe/spacecat-api-service/blob/feat/prompt-suggestion-schedules/docs/specs/2026-07-16-prompt-suggestion-schedules-onboarding.md) (added in this PR). Companion shared-client spec: [spacecat-shared `docs/specs/prompt-suggestion-schedules/spec.md`](https://github.com/adobe/spacecat-shared/blob/feat/prompt-suggestion-schedules/docs/specs/prompt-suggestion-schedules/spec.md).


src/controllers/llmo/llmo-onboarding.js
```diff
@@ -175,6 +175,134 @@ export async function triggerBrandalfOnboardingJob({
   return drsJob;
 }
 
+// Cadence labels accepted by DRS `createSchedule` for the recurring
+// "prompt suggestion" pipelines (the `SCHEDULE_CADENCES` enum in the drs-client).
+// DRS derives the concrete cron expression server-side from the label
```
MysticatBot, line 187: issue (blocking): `DrsClient.createFrom(context)` was previously inside the Brandalf try/catch. After this refactoring it sits above both the Brandalf block and `registerPromptSuggestionSchedules`. If `createFrom` throws (malformed context, SDK upgrade regression), nothing catches it and the entire onboarding request fails with a 500.

The intent of both the Brandalf trigger and schedule registration is "best-effort, never abort onboarding." The hoisted call breaks that contract.

Fix - wrap `createFrom` in its own guard:

```js
let drsClient;
try {
  drsClient = DrsClient.createFrom(context);
} catch (e) {
  log.error(`DRS client creation failed, skipping Brandalf + schedules: ${e.message}`);
  say(':warning: DRS client unavailable (will need manual trigger)');
}

if (drsClient) {
  // existing Brandalf try/catch + schedule registration
}
```

Alternatively, keep `createFrom` inside the existing try block and pass the instance through to `registerPromptSuggestionSchedules` from within it.

src/controllers/llmo/llmo-onboarding.js
```diff
@@ -175,6 +175,134 @@ export async function triggerBrandalfOnboardingJob({
...
+        // brands.js (activateBrand). `status` is the upstream HTTP status when the
+        // DRS client attaches one, else 'unknown'.
+        const status = scheduleError.status ?? 'unknown';
+        log.error(`Failed to register DRS ${name} schedule provider_id=${providerId} `
+          + `site_id=${siteId} status=${status}: ${scheduleError.message}`);
+        say(`:warning: Failed to register DRS ${name} schedule for site ${siteId} `
+          + '(will need manual trigger)');
+      }
+    }),
+  );
```
MysticatBot, line 317: issue (blocking): `registerPromptSuggestionSchedules` fires three parallel `createSchedule` HTTP calls and awaits all of them on the synchronous onboarding response path. The PR itself acknowledges the schedules are best-effort ("the durable outcome is the recurring schedule row"). Three outbound HTTP calls add tail latency risk to onboarding, especially under DRS partial-outage conditions.

Since `createSchedule` is idempotent (the `alreadyExisted` response confirms server-side dedup), this is a good candidate for fire-and-forget with error logging, or for wrapping in a timeout (the codebase already uses `settleWithin` for this pattern elsewhere in this file).

Fix - either:
1. Wrap in `settleWithin(registerPromptSuggestionSchedules({...}), 8000)` so a slow DRS does not stall onboarding.
2. Fire-and-forget: `registerPromptSuggestionSchedules({...}).catch(e => log.warn(...))` without `await`, since the only loss on timeout is the success log/Slack message (the schedule still exists server-side).

MysticatBot (changes requested):
Hey @dzehnder,

**Verdict: Request changes** - two material issues on the error-handling and latency path before this is safe for production.
**Complexity: HIGH** - medium diff; dependencies risk flag.
_Changes: Registers three recurring DRS prompt-suggestion schedules (SEMrush, citation attempts, synthetic personas) during V2 LLMO onboarding, each with an immediate first run (3 files)._
Note: CI checks are currently failing - expected per PR description (cross-repo dependency on drs-client 1.14.0 not yet published).

## Must fix before merge
1. [Important] `DrsClient.createFrom(context)` hoisted outside try/catch breaks "never abort onboarding" contract - `src/controllers/llmo/llmo-onboarding.js:~1528` (details inline)
2. [Important] Three parallel HTTP calls awaited on synchronous onboarding path without timeout cap - `src/controllers/llmo/llmo-onboarding.js:~1575` (details inline)

<details>

- nit: Triple "DRS client not configured" debug logs when unconfigured - add early `isConfigured()` guard in `registerPromptSuggestionSchedules` to short-circuit before iterating pipelines - `src/controllers/llmo/llmo-onboarding.js:~210`
- suggestion: Add a partial-failure test where one pipeline rejects and others resolve within the same `registerPromptSuggestionSchedules` call, proving the per-item try/catch isolation works - `test/controllers/llmo/llmo-onboarding.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 43s | <b>Cost:</b> $4.81 | <b>Commit:</b> <code>8cf6afd0765af375b5f0f8c6fc51e2a0e30a471e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/controllers/llmo/llmo-onboarding.js
```diff
@@ -175,6 +182,141 @@ export async function triggerBrandalfOnboardingJob({
...
```
MysticatBot, line 317: issue (blocking): `settleWithin(registerPromptSuggestionSchedules(...), 8000, null)` resolves to `null` on timeout, but the return value is discarded and no log line is emitted. If DRS is experiencing a partial outage (calls hang beyond 8s without rejecting), the operator has zero visibility that schedule registration was abandoned mid-flight. The per-pipeline catch blocks inside `registerPromptSuggestionSchedules` only fire on rejection, not on timeout (the promise remains pending).

Fix - capture the return value and log on timeout:

```js
const scheduleResult = await settleWithin(
  registerPromptSuggestionSchedules({ drsClient, siteId: site.getId(), log, say }),
  SCHEDULE_REGISTRATION_TIMEOUT_MS,
  null,
);
if (scheduleResult === null) {
  log.warn(`Prompt-suggestion schedule registration timed out after ${SCHEDULE_REGISTRATION_TIMEOUT_MS}ms for site ${site.getId()}`);
}
```

This works because `registerPromptSuggestionSchedules` returns `undefined` on normal completion (no explicit return after `Promise.all`), so `=== null` correctly identifies the timeout case. Alternatively, have the function return a truthy summary to make the contract explicit.

MysticatBot (changes requested):
Hey @dzehnder,

**Verdict: Request changes** - one operational visibility gap on the timeout path before this is safe for production.
**Complexity: HIGH** - large diff; dependency risk flag.
_Changes: Registers three recurring DRS prompt-suggestion schedules (SEMrush, citation attempts, synthetic personas) during V2 LLMO onboarding, each with an immediate first run (4 files)._
Note: CI checks are currently failing - expected per PR description (cross-repo dependency on drs-client 1.14.0 not yet published).
Note: Recommend a human read before merge - this change adds a spec document under `docs/specs/` (architectural-path trigger). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Important] `settleWithin` timeout returns `null` with no log - operator has no signal schedules were abandoned - `src/controllers/llmo/llmo-onboarding.js:~1611` (details inline)

<details>

- nit: `alreadyExisted` test only asserts substring `'already existed'` - add assertions for `scheduleId` and `providerId` to catch silent log degradation - `test/controllers/llmo/llmo-onboarding.test.js:~5867`
- nit: `calledThrice` assertion on `context.log.error` is brittle - any future unrelated error log in V2 onboarding breaks this test; filter by message content instead - `test/controllers/llmo/llmo-onboarding.test.js:~5960`
- suggestion: `registerPromptSuggestionSchedules` returns `undefined` implicitly; adding `return { completed: true }` would make the `settleWithin` timeout distinguishable without a sentinel - `src/controllers/llmo/llmo-onboarding.js:~319`
</details>

- `DrsClient.createFrom(context)` now wrapped in its own try/catch with ERROR log + Slack signal
- Three parallel HTTP calls now bounded by `settleWithin(8000ms)`
- Early `isConfigured()` guard added in `registerPromptSuggestionSchedules` to avoid triple debug logs
- Partial-failure isolation test added proving per-pipeline try/catch works independently

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 50s | <b>Cost:</b> $6.12 | <b>Commit:</b> <code>22821ea203c7cca416fe6d7b75aebd8a4d3d4bab</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @dzehnder,

**Verdict: Approve** - all prior blocking findings addressed; remaining items are non-blocking nits.
**Complexity: HIGH** - large diff; dependency risk flag.
_Changes: Registers three recurring DRS prompt-suggestion schedules (SEMrush, citation attempts, synthetic personas) during V2 LLMO onboarding, each with an immediate first run, bounded by settleWithin (5 files)._
Note: Recommend a human read before merge - this change adds a spec document under `docs/specs/` (architectural-path trigger). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: `package-lock.json` records `"^1.14.0"` in `packages[""].dependencies` while `package.json` pins exact `"1.14.0"` - regenerate lockfile to align specifiers - `package-lock.json:27`
- suggestion: `registerPromptSuggestionSchedules` returns `{ completed: true }` even when all three pipelines failed - consider returning `{ completed: true, registered, failed }` counts so the sentinel reflects reality - `src/controllers/llmo/llmo-onboarding.js:~319`
- suggestion: Import `SCHEDULE_CADENCES` from `@adobe/spacecat-shared-drs-client` now that 1.14.0 is installed (the "once installed" condition in the comment is met) - `src/controllers/llmo/llmo-onboarding.js:~184`
- suggestion: Add a test asserting schedules still register when Brandalf submitJob rejects (the code path is correct but the independence is not proven in the V2 integration suite) - `test/controllers/llmo/llmo-onboarding.test.js`
</details>

- `DrsClient.createFrom(context)` now wrapped in its own try/catch with ERROR log + Slack signal
- Three parallel HTTP calls now bounded by `settleWithin(8000ms)`
- Timeout path now logs at WARN with full context (site_id, timeout duration, operational guidance)
- Early `isConfigured()` guard added to avoid triple debug logs
- Partial-failure isolation test added proving per-pipeline try/catch works
- Error log assertions filter by message content instead of bare `calledThrice`

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 28s | <b>Cost:</b> $4.88 | <b>Commit:</b> <code>49b3cfd32ff4dbfccc985d93418652dc33c6442a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2823, opened 2026-07-15
feat: introduce brand presence semrush stats api

vivesing:
## Summary

Implements `GET /v2/orgs/:spaceCatId/brands/:brandId/serenity/brand-presence/stats` — the Elements (Semrush)-backed equivalent of the existing Postgres-RPC `brand-presence/stats` endpoint (`llmo-brand-presence.js#createBrandPresenceStatsHandler`). Returns aggregated KPI stats (`total_executions`, `average_visibility_score`, `total_mentions`, `total_citations`) for a brand over a date range, with an optional weekly `trends[]` breakdown.

Full design notes, confirmed Semrush payload/response shapes, and resolved open decisions are documented in `docs/elements/brand-presence-stats-plan.md`.

## What changed

- **Route + auth scaffold** (`src/controllers/elements.js`, `src/routes/index.js`, `src/routes/facs-capabilities.js`, `src/routes/required-capabilities.js`): registers the new route under the existing brand-scoped `authorizeOrg` auth pattern (`llmo/can_view` FACS capability, `organization:read` S2S capability).
- **`getStats` wired to real Semrush aggregation** (`src/controllers/elements.js`):
  - Validates/defaults `startDate`/`endDate` (28-day trailing window when omitted).
  - Resolves `regionCode`/`region_code`/`region` to a single Semrush `projectId` (via `resolveRegionProjectId`) when a specific region+language is selected — the common case, no fan-out needed.
  - Falls back to aggregating across **all** of the brand's Semrush projects when no region is specified.
  - Validates `siteId`/`site_id` belongs to the requested brand.
  - Delegates to the new service method and returns via `cachedOk`.
- **New element definitions** (`src/support/elements/definitions/brand-presence-stats.js`): payload builders + response transformers for the 4 backing Semrush elements — `TOTAL_EXECUTIONS`, `MENTIONS`, `VISIBILITY`, `CITATIONS_KPI` — including the confirmed quirks:
  - Citations uses `CBF_brand` + `CBF_projects` (plural), while Mentions/Visibility use `CBF_ws_brand` + `CBF_project` (singular).
  - Visibility's raw value is a 0–1 fraction; the transform multiplies by 100.
  - `comparison_start_date`/`comparison_end_date`/`comparison_data_formatting` are intentionally omitted — the `/stats` contract has no period-over-period comparison field.
[...]

PR #2827, opened 2026-07-15
feat(elements): add brand-presence market-tracking-trends endpoint

jjenscodee:
## Summary

Adds a Semrush **Elements**-backed endpoint returning **weekly per-competitor mentions + citations** for the tracked brand and its competitors — the data behind the **Competitor Comparison** chart (`CitationsMentionsTrendChart`) on the `brand-presence-sr-ui` dashboard.

Full design notes, the confirmed Semrush payload/response shapes, and resolved decisions are in `docs/elements/market-tracking-trends-plan.md` (modeled on `brand-presence-stats-plan.md` from #2823). All brand names and metrics in the docs/tests are synthetic.

> The element mapping was verified by observing the Brand Presence MFE's "Market Tracking" chart request/response shapes (Mentions and Citations tabs) — not inferred from the migration wiki, whose documented citations trend element (`b81af644`, domain-level) is *not* what this chart uses.

## Element mapping

Both are `type: "line"`, `auto_bucketing: "week"`, **no brand filter** (so competitors come back natively as tracked-benchmark `legend`s):

| Chart metric | Element | Constant | Project filter col | Value field |
|---|---|---|---|---|
| Mentions | `b5281393` | `TRENDS_MV` (existing) | `CBF_project` (singular) | `y__mentions` |
| Citations | `2e5a6f4e` | `MARKET_CITATIONS_TREND` (**new**) | `CBF_projects` (**plural**) | `y__mentions` |

⚠️ Both elements expose the value under the key `y__mentions`; in the citations element that number is the **citation** count (Semrush reuses the generic key). The transform maps each to the right metric.

## What changed

- **Route + auth scaffold** (`routes/index.js`, `routes/facs-capabilities.js` → `llmo/can_view`, `routes/required-capabilities.js` → `brand:read`): registers the route under the existing brand-scoped `authorizeOrg` pattern; `brandId` reuses the existing LLMO FACS brand alias.
- **Controller** `getMarketTrackingTrends` (`controllers/elements.js`): mirrors the URL-inspector handlers — `authorizeOrg`, `siteId` cross-check, optional dates with a 28-day trailing default + validation, region→`projectId` resolution (else aggregate across all the brand's projects), `cachedOk`.
- **Definitions** (`support/elements/definitions/market-tracking-trends.js`, new): weekly payload builders (`CBF_project` vs `CBF_projects`, no brand filter, no `comparison_data_formatting`) + `transformMarketTrackingTrends` — groups both `blocks.lines[]` by ISO week, splits the brand line from competitor lines, defaults a missing metric to 0.
- **Service** `getMarketTrackingTrends` (`support/elements/elements-service.js`): fetches the two elements in parallel; no per-week fan-out (elements are pre-bucketed) and no per-competitor fan-out (`legend` carries them). `projectId` (single region) takes precedence over `projectIds` (aggregate).
- **Element id** (`support/elements/element-ids.js`): new `MARKET_CITATIONS_TREND`.
- **Design doc** (`docs/elements/market-tracking-trends-plan.md`, new).

## Response shape

```json
{
  "weeklyTrends": [
    {
      "week": "2026-07-05", "weekNumber": 27, "year": 2026,
      "mentions": 900, "citations": 5000,
      "competitors": [ { "name": "Rival One", "mentions": 150, "citations": 300 } ]
    }
  ]
}
```

## Testing

- New/added unit suites: definitions (payload builders + transform), service (`getMarketTrackingTrends`), controller (auth, date defaulting/validation, region resolution + 404, `siteId` cross-check, error mapping), and route registration.
- Route-param FACS classification + required-capabilities drift suites pass.
- Weekly-only bucketing for now; visibility intentionally excluded (out of scope for this chart).

## Out of scope

UI wiring (DAL service + hook + `SRCompetitorComparisonSection`); Competitor Summary / Share of Voice / Sentiment endpoints; day/month bucketing.


MysticatBot (changes requested):
Hey @jjenscodee,

**Verdict: Request changes** - two defensive guards present on sibling endpoints are missing here.
_Changes: Adds a brand-presence market-tracking-trends endpoint backed by two Semrush Elements (mentions + citations), with weekly competitor comparison transform (13 files)._

## Must fix before merge
1. [Important] No maximum date-range guard - sibling endpoints enforce MAX_RANGE_DAYS but this one allows unbounded spans - `src/controllers/elements.js:804` (details inline)
2. [Important] Half-supplied date range silently backfills from defaults, creating potentially unbounded window - `src/controllers/elements.js:832` (details inline)

<details>

- nit: Competitors sorted by mentions only; a citation-heavy competitor with 0 mentions will sort last - confirm this matches product intent - `src/support/elements/definitions/market-tracking-trends.js:107`
- suggestion: The 28-day default window test asserts only format and ordering, not the actual 28-day span - using sinon.useFakeTimers would catch regressions to the window size - `test/controllers/elements.test.js:975`
</details>

src/controllers/elements.js
```diff
@@ -739,12 +753,101 @@ export default function ElementsController(context, log, env) {
...
+      if (hasText(startDate) && !isYmdDate(startDate)) {
+        return badRequest('startDate must be a valid YYYY-MM-DD date');
+      }
+      if (hasText(endDate) && !isYmdDate(endDate)) {
+        return badRequest('endDate must be a valid YYYY-MM-DD date');
+      }
+      if (hasText(startDate) && hasText(endDate) && startDate > endDate) {
+        return badRequest('startDate must not be after endDate');
+      }
+      const defaultRange = defaultTrailingDateRange();
```
MysticatBot: issue (blocking): **No maximum date-range guard.** The sibling `listOwnedUrls` and `listDomainUrls` handlers both enforce `MAX_RANGE_DAYS = 366` to prevent clients from requesting multi-year spans that generate expensive upstream calls. This new handler validates format and ordering but imposes no upper cap when dates ARE supplied. A caller could pass `startDate=2020-01-01&endDate=2026-07-15` and force a 6-year Semrush scan.

Since this endpoint also aggregates across all projects when no region is specified, the fan-out risk compounds.

**Fix:** Add the same `MAX_RANGE_DAYS` guard already present in sibling handlers, between the ordering validation and the `defaultTrailingDateRange()` call.

src/controllers/elements.js
```diff
@@ -739,12 +753,101 @@ export default function ElementsController(context, log, env) {
...
+      } else {
+        projectIds = brandSemrushProjects
+          .map((p) => p.semrushProjectId)
+          .filter(hasText);
+      }
+
+      const result = await service.getMarketTrackingTrends(workspaceId, {
+        model: query.model,
+        platform: query.platform,
+        startDate: startDate || defaultRange.startDate,
```
MysticatBot: issue (blocking): **Half-supplied date range falls back silently.** When the caller sends only `startDate` (no `endDate`) or vice versa, the handler fills the missing half from `defaultTrailingDateRange()`. For example, `?startDate=2020-01-01` without `endDate` produces `startDate=2020-01-01, endDate=<today>` - a 6-year unbounded range that bypasses the 28-day-default contract.

The sibling handlers require both or neither. This endpoint should do the same.

**Fix:** Either (a) require both or neither, or (b) apply the full default range as a unit (if either is absent, ignore both and use the default window). Option (a) is simpler and consistent with the plan doc (all optional means the pair is optional).

jjenscodee:
Addressed both blocking issues from the review (commit 61356046f):

✅ **No maximum date-range guard** — added the `MAX_RANGE_DAYS = 366` span cap (returns 400 when exceeded), matching `listOwnedUrls`/`listDomainUrls`. Prevents a multi-year Semrush scan, which was compounded by the all-project fan-out when no region is given.

✅ **Half-supplied date range fell back silently** — the handler now requires **both or neither** of `startDate`/`endDate`; if either is absent it uses the full 28-day default range as a unit, so a caller-provided bound never pairs with the default's other half to form an unbounded window.

Added unit tests for both (both-or-neither fallback, 366-day cap → 400) and updated `docs/elements/market-tracking-trends-plan.md`. Controller suite green, lint clean.

MysticatBot (changes requested):
Hey @jjenscodee,

**Verdict: Request changes** - two documented process requirements from the repo's CLAUDE.md are unmet (OpenAPI spec and integration tests).
_Changes: Adds a brand-presence market-tracking-trends endpoint backed by two Semrush Elements (mentions + citations), with weekly competitor comparison transform (13 files)._

## Must fix before merge
1. [Important] Missing OpenAPI specification for the new endpoint - CLAUDE.md "Adding a New Endpoint" step 1 requires `docs/openapi/paths/*.yaml` and step 2 requires `docs/openapi/schemas.yaml` updates. No OpenAPI path or schema was added for `GET /v2/orgs/:spaceCatId/brands/:brandId/serenity/brand-presence/market-tracking-trends`. The spec is the machine-readable contract for API consumers and powers generated documentation. Add the path definition and `weeklyTrends` response schema, then run `npm run docs:lint`.
2. [Important] Missing integration tests - CLAUDE.md "Test Requirements" states: "New or modified endpoints must include integration tests in `test/it/`". Unit test coverage is thorough, but the documented convention requires end-to-end tests exercising the full middleware stack against the PostgreSQL harness. Per the checklist: add shared test logic in `test/it/shared/tests/`, seed data in `postgres/seed-data/`, and a wiring file in `postgres/`.

<details>

- nit: Broken Markdown table in plan doc - the "Both or neither" paragraph between the date rows and the model/regionCode/siteId rows breaks the table rendering on GitHub - `docs/elements/market-tracking-trends-plan.md:38`
- nit: Unused `brandId` destructured from `ctx.params` but never referenced (brand comes from `auth.brand`) - `src/controllers/elements.js:777`
- suggestion: Extract the date-range validation (both-or-neither + MAX_RANGE_DAYS cap) into a shared helper - three handlers now duplicate this logic and the pattern is crystallizing - `src/controllers/elements.js`
- nit: User-supplied `region` value reflected in 404 error message without truncation - low risk (JSON content-type, pre-existing pattern across sibling handlers) - `src/controllers/elements.js:838`
- suggestion: Use `sinon.useFakeTimers` in the "defaults to 28-day trailing window" test to pin exact dates rather than asserting only format and ordering - `test/controllers/elements.test.js:975`
- suggestion: Add a secondary sort key (`a.name.localeCompare(b.name)`) for competitor ties on mentions to stabilize ordering across V8 versions - `src/support/elements/definitions/market-tracking-trends.js:107`
</details>

- No maximum date-range guard - now enforces MAX_RANGE_DAYS = 366 with 400 response
- Half-supplied date range silently backfills - now requires both or neither, falling back to the full 28-day default as a unit

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 25s | <b>Cost:</b> $7.94 | <b>Commit:</b> <code>61356046fe0ca040f29cbbd9cee5c239968af25f</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jjenscodee:
Addressed both blocking items from the last review (commit f859ee8f3):

✅ **OpenAPI specification** — added the `v2-serenity-market-tracking-trends` path to `docs/openapi/serenity-api.yaml` (params, `SerenityMarketTrackingTrends` 200 schema, 400/403/404/502/503), a `SerenityMarketTrackingTrends` schema in `docs/openapi/schemas.yaml`, and wired the path in `docs/openapi/api.yaml` next to the sibling `url-inspector/filter-dimensions`. `npm run docs:lint` passes with no new warnings.

✅ **Integration tests** — added `test/it/shared/tests/serenity-market-tracking-trends.js` + the `test/it/postgres/` wiring file. They exercise the full middleware stack: deterministic route-gate 400s (non-UUID `spaceCatId`/`brandId`, fire before auth) plus a reachability + `weeklyTrends`-contract check. Following the sibling `llmo-url-inspector` IT factory, the happy-path assertion is tolerant — the two backing Elements (`TRENDS_MV`, `MARKET_CITATIONS_TREND`) aren't part of the Semrush mock's seeded fixtures, so it pins the schema when a 200 is served and accepts the auth/upstream/config branches otherwise. Reuses the existing `ORG_1`/`BRAND_1` baseline seed (no new seed-data file needed).

Also fixed a non-blocking nit: the "both or neither" note was splitting the param table in the plan doc — moved below the table.

On the other non-blocking items: the "unused `brandId`" nit looks like a false positive — `brandId` is passed to `resolveRegionProjectId(...)` in the region branch. The suggestions (shared date-validation helper, fake-timers, competitor tie-break sort key) are noted for a follow-up.

MysticatBot (approved):
Hey @jjenscodee,

**Verdict: Approve** - all prior blocking findings are resolved; no new issues.
_Changes: Adds a brand-presence market-tracking-trends endpoint backed by two Semrush Elements (mentions + citations), with weekly competitor comparison transform, OpenAPI spec, and integration tests (18 files)._
Note: Recommend a human read before merge - this change modifies a shared API contract (`docs/openapi/*`). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: OpenAPI spec declares a `503` response (`PostgREST client not configured`) but this endpoint is backed by Semrush Elements, not PostgREST - copy-paste from sibling URL Inspector endpoints that do use PostgREST - `docs/openapi/serenity-api.yaml:1249`
- suggestion: `defaultTrailingDateRange()` uses wall-clock `new Date()` while all other date arithmetic in the file uses UTC (`Date.parse(...T00:00:00Z)`) - on an edge-of-day server restart in a non-UTC timezone the default window could shift a day; using `new Date().toISOString().slice(0, 10)` for endDate would be consistent - `src/controllers/elements.js:154`
</details>

- No maximum date-range guard - now enforces MAX_RANGE_DAYS = 366 with 400 response
- Half-supplied date range silently backfills - now requires both or neither, full default as a unit
- Missing OpenAPI specification - path, schema, and all params/responses added in serenity-api.yaml
- Missing integration tests - shared test factory + postgres wiring added following existing IT pattern

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 13s | <b>Cost:</b> $5.44 | <b>Commit:</b> <code>f859ee8f33a7ecba6d7198bdebf7f40410c0f74c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

calvarezg:
## Multi-Persona PR Review

Reviewed by Architect, DBA, Security, Tester, and Junior Dev agents against the full diff (19 files). No new blocking issues — the two prior review rounds (date-range guards, OpenAPI spec, integration tests) are all correctly resolved.

**Recommendation: Changes recommended** (0 critical / 2 major / 5 minor)

### Major

1. **No test pins the 366-day cap's exact boundary** — `src/controllers/elements.js:816-820`. The only cap test uses a ~2400-day span; nothing verifies 366 days passes and 367 fails. A `>=`/`>` slip in this guard (the fix from round 1) would go undetected. Add boundary tests at exactly 366 and 367 days.

2. **Integration tests are too status-tolerant to catch param-wiring regressions** — `test/it/shared/tests/serenity-market-tracking-trends.js:44-70`. The reachability test accepts almost the entire status space and only exercises the base path + one malformed date. `region`, `siteId`, `model`/`platform`, and the range-guard behaviors have zero IT coverage — a real regression in query-param wiring through the middleware stack wouldn't be caught by unit tests (which stub the service) or this IT test. Add at least one IT case each for `region=` and `siteId=`.

### Minor

- Aggregate path (no `region` given) has no guard for a brand with zero Semrush projects — silently sends an unscoped upstream query instead of the 404 the single-region "no market" path returns (`elements.js:840-843`).
- `accumulate()` in `market-tracking-trends.js:131,135` doesn't validate `line.x`'s date format before bucketing — a malformed upstream date would silently produce `null` `weekNumber`/`year` rather than an error.
- Alias precedence (`model` vs `platform`, `regionCode`/`region_code`/`region`) and a citations-only upstream-failure case are untested (low risk, easy to add).
- OR-filter project list width is unbounded on the aggregate path (likely fine at realistic market counts).
- `element-ids.js:54-61` — comment density inconsistent with sibling constants; cosmetic only.

### What's working well
Auth/tenant-scoping reuse, both previously-fixed date-range guards, safe PostgREST helper usage (no injection/N+1), and correct 0/NaN/missing handling in the merge transform.

*4 agent personas (Architect, DBA, Security, Tester+Junior Dev) reviewed independently against the raw diff; Security reported no findings.*

jjenscodee:
Thanks for the multi-persona review. Addressed in commit cb8d8a816:

**Major**
1. ✅ **366-day boundary tests** — added exact cases: a **366-day** span passes (cap is inclusive) and a **367-day** span 400s. This pins the `>`/`>=` boundary the round-1 fix hinges on (the pre-existing ~2400-day case stays as a coarse check).
2. ✅ **IT param-wiring coverage** — added IT cases for `siteId` (a well-formed but unseeded siteId must 400, *not* be silently ignored) and `region=US` (routes through region resolution). Kept tolerant on the region status since the two backing Elements aren't in the mock's seed, but a wiring crash / silently-ignored param now surfaces.

**Minor**
- ✅ **Aggregate zero-projects guard** — a brand with no Semrush projects no longer issues an unscoped workspace-wide query (which on the org-parent-workspace fallback would surface other brands' data); it now returns `{ weeklyTrends: [] }` with no upstream call. Added a unit test.
- ✅ **Malformed upstream `x`** — the transform now only buckets well-formed `YYYY-MM-DD` week starts, so a bad upstream date is skipped rather than producing `null` weekNumber/year. Added a unit test.

**Deferred (low-value / cosmetic, happy to add if you'd like):** alias-precedence + citations-only-upstream-failure unit tests, the OR-filter width cap (fine at realistic market counts), and the `element-ids.js` comment-density nit.

calvarezg (approved):
Verified commit cb8d8a816 against the multi-persona review findings:

- Both Major items fixed: 366/367-day boundary tests added (confirms the cap is inclusive at exactly 366), and IT tests now cover `siteId`/`region` param wiring through the real middleware stack.
- 2 of 5 Minor items fixed: the zero-Semrush-projects aggregate guard (turns out this also prevents a real cross-brand data leak on the org-parent-workspace fallback, not just an empty/wasted call — good catch), and the malformed-upstream-date guard in the week-bucketing transform.
- Remaining 3 minors (alias-precedence tests, OR-filter width cap, comment density in `element-ids.js`) are cosmetic/low-value and fine to leave as-is.

No regressions in the fix commit. Approving.

PR #2831, opened 2026-07-16
fix: exclude decorative images from alt-text PLG grant selection (SITES-44253)

yashrathore11:
## Summary

Decorative-image suggestions (`data.recommendations[].isDecorative === true`) were competing for a PLG customer's limited alt-text grant slots (3 per month) alongside suggestions that actually need an alt-text fix. Since decorative images don't need alt text, granting one wasted a slot that should have gone to a real fix — silently reducing how many actionable suggestions a PLG customer saw that cycle.

This adds an `alt-text` grouping strategy to `OPPORTUNITY_STRATEGIES` in `grant-suggestions-handler.js` that excludes decorative suggestions before grant selection, so all 3 monthly slots go to suggestions that actually need fixing. If fewer than 3 non-decorative suggestions exist, fewer than 3 are granted — slots are no longer backfilled with decorative suggestions just to hit the cap.

No public API contract change: same endpoints, same response shape — only which suggestions get a grant record (and therefore surface to `sites-optimizer-ui` PLG callers via the existing `filterByGrantStatus` gate) changes.

## Testing

- Added 6 unit tests in `test/support/grant-suggestions-handler.test.js` covering: exclusion of decorative suggestions, exclusion when *any* of multiple recommendations is decorative, retention of non-decorative suggestions, missing/empty `recommendations` treated as non-decorative, plain-object `data` support, and default rank-ascending sort of survivors.
- Full `grant-suggestions-handler.test.js` (58/58) and the `suggestions`/`opportunities` controller suites (528/528) pass.
- Lint clean.

- [ ] N/A — no API specification change (no new/modified endpoint, request, or response shape)

- [ ] N/A — internal grant-selection logic only; no entity or API-facing contract change

- [ ] N/A — not an audit type change

## Related Issues

[SITES-44253](https://jira.corp.adobe.com/browse/SITES-44253)


MysticatBot (approved):
Hey @yashrathore11,

**Verdict: Approve** - clean, well-scoped fix with thorough test coverage.
**Complexity: LOW** - small diff, single service, internal logic only.
_Changes: Adds an `alt-text` grouping strategy to `OPPORTUNITY_STRATEGIES` that excludes decorative-image suggestions from PLG grant selection (2 files)._

<details>

- nit: The comment (lines 144-147) says `data.recommendations[].isDecorative` - consider rewording to "Suggestions where any recommendation is marked decorative are excluded" to make the any-match semantics explicit without needing to read the filter body. - `src/support/grant-suggestions-handler.js:144`
- suggestion: The test "works with plain object data (no getData())" proves the filter handles plain objects, but if the downstream default sort calls `getId()`/`getRank()` methods, plain objects without those methods would fail end-to-end. Consider whether this fallback path occurs at runtime - if not, the test overstates the supported contract. - `test/support/grant-suggestions-handler.test.js:361`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 3s | <b>Cost:</b> $3.43 | <b>Commit:</b> <code>e18bbe7c999822b48bda194e1b99a3d7c174d1b0</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2833, opened 2026-07-16
feat(serenity): edit prompts in place — id-stable rename and tag update with 409 conflict mapping

rainer-friederich:
## 1. Abstract

Serenity's prompt edit endpoints now mutate the prompt in place via Semrush's id-preserving operations (`aio-rename-prompt` for the text, `PUT /aio/prompts/tags` for the tags) instead of deleting and recreating the prompt, and surface the upstream text-collision 409 to callers as `conflict`.

## 2. Reasoning

Semrush's rename operation is in-place and id-preserving, so our delete-and-recreate edit path was self-inflicted: every text edit churned the prompt id, severing anything holding a reference to it (history, tags carried over by re-creation, concurrent readers). The in-place edit spec (serenity-docs PR 63) mandates moving the edit surface onto the native operations and making the prompt id stable across its whole lifecycle.

- The Serenity REST transport gains two operations: `renamePrompt` (POST `.../aio/prompts/{id}/rename`, the in-place text edit — in the AIO model a prompt's name IS its text) and `updatePromptTagsByIds` (PUT `.../aio/prompts/tags`).
- `handleUpdatePrompt` and `handleUpdatePromptSubworkspace` no longer delete and recreate: a text change renames in place, a tag change rewrites the tag set, and the prompt keeps its id through both. Before: an edited prompt came back with a new id. After: the id is stable.
- Renaming a prompt onto another prompt's exact text now returns a caller-visible `409 conflict` (previously the recreate path would fail opaquely or duplicate); the upstream error body is not forwarded verbatim — the error mapper redacts it to the catalogued `conflict` error. The 409 is documented in the Serenity OpenAPI spec.
- No-op edits (same text, same tags) short-circuit without upstream writes.

## 4. Required information

- Spec: https://github.com/adobe/serenity-docs/pull/63 — `docs/prompt-management/prompt-in-place-edit.md`, including the live-probe record of the rename endpoint's exact semantics
- Other: WP1 (mock rename handler) — https://github.com/adobe/spacecat-shared/pull/1814

- spacecat-shared — consumed: `@adobe/spacecat-shared-project-engine-client` is pinned to 1.12.0, the release line carrying the stateful `aio-rename-prompt` mock handler (the IT mock image is tag-locked to the installed client version)
- serenity-docs — contract: implements the in-place edit spec above

- The rename endpoint's semantics were probed against the live prod Semrush gateway on a throwaway project (two temporary prompts, created, probed, deleted; draft layer only): exact 409 body wording, exact-match collision comparison (no whitespace/case/Unicode normalization), `is_updated` mirroring the live layer, empty/omitted/null `new_name` applied literally as `''`, and body-less requests rejected with `400 EOF`. The findings are recorded in the spec's probe table and pinned in the mock (WP1).

## 7. Test plan

- Locally, the full Serenity integration suite was run against the real docker compose stack (data-service, Postgres, minio, user-manager mock, and the **published** GHCR project-engine mock image 1.12.0 — not a locally built one), exercising the new rename and tag-update paths end to end through the API, including the 409 collision case; all green.
- Dev verification after deploy: through the API, edit a Serenity prompt's text on a dev workspace and confirm the response carries the SAME prompt id; edit only its tags and confirm the same; rename it onto another prompt's exact text and confirm a `409 conflict`.

- https://github.com/adobe/spacecat-shared/pull/1814 — depends-on: the project-engine-client mock rename handler this PR's integration suite runs against. Merged and released as `@adobe/spacecat-shared-project-engine-client` v1.11.0 (this PR pins 1.12.0, the current release on that line).
- Sequence: spacecat-shared PR 1814 (done) → this PR. No runtime deploy coupling — the shared PR only ships the test mock; the production transport talks to the real Semrush gateway.


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-scoped elimination of the delete-create data-loss window with sound error ordering and comprehensive tests.
_Changes: Moves Serenity prompt editing from delete-then-create to in-place rename + replace-mode tag write, stabilizing the prompt id across edits and surfacing a 409 conflict for text collisions (16 files)._

<details>

- suggestion: Verify the subworkspace PATCH endpoint's OpenAPI spec also declares 409 as a possible response - the controller's `mapError` handles 409 globally, so both routes return it at runtime, but only `v2-serenity-prompt-by-id` has it documented in the diff - `docs/openapi/serenity-api.yaml`
- suggestion: Consider adding a `log.warn` when `updatePromptTagsByIds` throws after a successful rename, noting the prompt id and the fact that text was already updated - helps on-call distinguish "nothing happened" from "text moved, tags stale" without chasing generic 502 logs - `src/support/serenity/handlers/prompts.js:~730`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 4s | <b>Cost:</b> $9.14 | <b>Commit:</b> <code>094fe15b91583df9651493b3728b1cb9f738b006</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Review round 1 (MysticatBot review 4710825825, Approve) — both non-blocking items addressed:

- **409 on the subworkspace PATCH spec** — no change needed: there is no separate subworkspace PATCH endpoint in the OpenAPI spec. Both `handleUpdatePrompt` and `handleUpdatePromptSubworkspace` back the same documented operation (`updateSerenityPrompt` on `v2-serenity-prompt-by-id`) — the controller dispatches on the org's binding mode internally — and that operation already declares the 409.
- **log.warn on a tag-write failure after a successful rename** — fixed in f203fda9e: both update handlers now log a warn carrying the prompt id, project id, and error before rethrowing, so the half-applied edit (text moved, tags stale) is attributable on-call instead of hiding behind the generic upstream error. Covered in the unit tests for both handlers.

PR #2840, opened 2026-07-16
feat(api): add ?fields= sparse-fieldset projection to list endpoints

sandsinh:
Adds an opt-in `?fields=` query parameter (JSON:API / Google API convention) that reduces list-response payloads to a caller-selected set of top-level fields. Lets size-constrained consumers request lighter responses without changing behaviour for any existing consumer — when the param is absent the full representation is returned unchanged.

- New reusable util `src/utils/field-projection.js` (parse, project, validate). `id` is always retained; unknown fields are ignored; a param that matches no field returns 400. Bounded to max 50 fields / 1000 chars to prevent unbounded parsing and unbounded error-message echo.
- Wired into sites (getAll incl. baseUrlContains + pagination shapes, getAllByDeliveryType, getAllWithLatestAudit), opportunities (getAllForSite, getByStatus), and suggestions list endpoints (composes after the existing `view` projection).
- OpenAPI: reusable `fields` parameter referenced from affected GET paths.
- Unit, controller, and integration test coverage.

## Example

Request:
```
GET /sites?fields=id,baseURL,name
Authorization: Bearer <token>
```

Response (200) — only the requested top-level fields, `id` always included:
```json
[
  {
    "id": "123e4567-e89b-12d3-a456-426614174000",
    "baseURL": "https://www.example.com",
    "name": "Example Site"
  }
]
```

Request with a `fields` value that matches nothing on any item:
```
GET /sites?fields=nope,missing
```

Response (400):
```json
{
  "message": "Invalid fields: nope, missing"
}
```

- [x] make sure you update the API specification with the type, schema of the audit result and an example

## Related Issues

Thanks for contributing!


src/utils/field-projection.js
```diff
@@ -0,0 +1,109 @@
...
+ */
+export function parseFields(fieldsParam) {
+  if (!hasText(fieldsParam)) {
+    return null;
+  }
+  const fields = fieldsParam
+    .split(',')
+    .map((field) => field.trim())
+    .filter(Boolean);
+  return fields.length > 0 ? fields : null;
```
MysticatBot, line 45: issue (blocking): **Missing input length/count limits on the `fields` parameter.**

The CLAUDE.md for this repo explicitly requires input validation at the boundary: "Missing input validation at the boundary - accepting and processing user input without validation (type checking, length limits, allowed values). Validation at the API boundary prevents garbage from reaching business logic."

`parseFields` splits on commas with no upper bound on token count or token length. A caller can send thousands of comma-separated values, producing a large array that iterates per-item in `projectFields`. The error message also reflects the entire parsed list back verbatim, creating unbounded response bodies.

**How to fix:** Add a max-fields cap (e.g. 50) and a max-length guard on the raw string (e.g. 1000 chars). Return 400 if exceeded. Also truncate the error message to show at most the first few invalid names.

src/utils/field-projection.js
```diff
@@ -0,0 +1,109 @@
...
+export function hasMatchingFields(items, fields) {
+  if (!fields || !Array.isArray(items) || items.length === 0) {
+    return true;
+  }
+  const availableKeys = new Set();
+  for (const item of items) {
+    if (isObject(item)) {
+      Object.keys(item).forEach((key) => availableKeys.add(key));
+    }
+  }
```
MysticatBot, line 95: issue (blocking): **Non-deterministic validation on empty result sets.**

`hasMatchingFields` returns `true` when `items.length === 0`, so the same request `?fields=nope` returns 400 when items exist but 200 with `[]` when the collection is empty. The validation outcome depends on runtime data state rather than the request alone.

This means a caller with a typo in their `fields` param will only discover the error when the endpoint happens to return results - an inconsistency that makes debugging harder.

**How to fix:** Consider one of:
1. Validate against a static set of known fields per endpoint (deterministic, but requires maintaining an allowlist)
2. Always return 200 for empty collections regardless of `fields` (current behavior, but document it explicitly and add a test asserting this contract)
3. At minimum, add a unit test for `applyFieldProjection([], 'anything')` to anchor the current behavior so a future change does not accidentally flip it to 400

MysticatBot (changes requested):
Hey @sandsinh,

**Verdict: Request changes** - two input validation gaps need addressing before merge.
**Complexity: HIGH** - large diff; API surface contract changes across 11 endpoints.
_Changes: Adds opt-in `?fields=` sparse-fieldset projection to list endpoints with a reusable utility, OpenAPI spec updates, and comprehensive tests (13 files)._

## Must fix before merge
1. [Important] Missing input length/count limits on `fields` parameter - `src/utils/field-projection.js:29` (details inline)
2. [Important] Non-deterministic validation on empty result sets - `src/utils/field-projection.js:71` (details inline)

<details>

- nit: `projectFields` uses `in` operator (checks prototype chain) while `hasMatchingFields` uses `Object.keys` (own properties only). Use `Object.hasOwn(obj, field)` in `projectFields` for consistency - `src/utils/field-projection.js:56`
- nit: Dead branch in `applyFieldProjection` - when `items` is not an array and `fields` is present, input passes through unprojected with no error. Every call site passes arrays, making this unreachable - `src/utils/field-projection.js:103`
- suggestion: Document the `view` + `fields` interaction on suggestion endpoints. A request like `?view=minimal&fields=data` returns 400 because `data` is absent from the minimal view, which may surprise callers - `docs/openapi/parameters.yaml:11`
- suggestion: Add at least one integration test for sites and suggestions `?fields=` to complement the opportunities integration coverage - `test/it/shared/tests/opportunities.js:97`
- suggestion: Echo the applied `fields` value in paginated response envelopes (alongside `limit`, `cursor`, `baseUrlContains`) so callers can confirm projection was applied - `src/controllers/sites.js:537`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 41s | <b>Cost:</b> $8.78 | <b>Commit:</b> <code>ef96c2ccb081eb62f50426b2eee9278ccc8df735</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

sandsinh (commented):
Addressed MysticatBot review feedback:

**Must-fix (blocking):**
- Added max length (1000 chars) and max count (50 fields) limits on the `fields` param — returns 400 instead of doing unbounded parsing/iteration; the invalid-fields error message is now truncated to the first 5 names instead of echoing the whole list back.
- Documented the deterministic-by-design contract for empty collections in `hasMatchingFields` JSDoc (returning 200 with `[]` regardless of `fields` when the collection is empty), and added a test anchoring that behavior so it can't silently flip later.

**Non-blocking:**
- `projectFields` now uses `Object.hasOwn` instead of `in`, matching `hasMatchingFields`'s own-properties-only semantics.
- Documented the `view` + `fields` interaction in `docs/openapi/parameters.yaml` (e.g. `view=minimal&fields=data` returns 400 since `data` isn't in the minimal view).

Not addressed (left as follow-up, larger scope): additional integration tests for sites/suggestions `?fields=`, and echoing the applied `fields` value in response envelopes.

Also merged latest `main` into this branch (one conflict in the generated `docs/index.html`, resolved by regenerating via `npm run docs:build`).

MysticatBot (approved):
Hey @sandsinh,

**Verdict: Approve** - prior blocking findings properly addressed; implementation is clean and secure.
**Complexity: HIGH** - large diff; API surface contract changes across 11 endpoints.
_Changes: Adds opt-in `?fields=` sparse-fieldset projection to list endpoints with a reusable utility, input bounds, OpenAPI spec updates, and comprehensive tests (13 files)._

<details>

- nit: `hasMatchingFields` builds a Set from every item's keys for validation. For homogeneous DTO collections a single-item sample would suffice and avoid O(n * keys) allocation on large result sets - `src/utils/field-projection.js:68`
</details>

- Input length (1000 chars) and count (50 fields) limits now enforced with early-return guards
- Empty-collection validation contract documented in JSDoc and anchored by test
- `Object.hasOwn` used consistently instead of `in` operator
- `view` + `fields` interaction documented in OpenAPI parameter description

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 8m 56s | <b>Cost:</b> $3.79 | <b>Commit:</b> <code>aa2639d32ad956f5a886821cf8189dfcd1eb5d9d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2843, opened 2026-07-17
fix: Prompt Research prompts count from real per-topic total

calvarezg:
## Summary

`handleTopicsResearch` (the POC `/llmo/ai-visibility/topics/research` endpoint behind Prompt Research → Related Topics) set each topic's `promptsCount` from the `topicsByFTS` row field — an FTS-scoped placeholder that understated the real total (a topic card showed e.g. 5 while its prompt list held 100). This backfills `promptsCount` per topic from `promptsByTopicIDsTotal` — the same gRPC total the expanded prompt list already uses — so the card count and the list total agree by construction. Applied to both the single-engine and all-engines (`LLM_ENUM.ALL`) paths, with a fallback to the previous value if the topic id is unparseable or the call fails.

Surfaced during TCS AI Day demo prep (llm-optimizer-customer-success Slack thread).

## API spec

No specification change — the response shape is unchanged (`promptsCount` already exists); only the value's source is corrected. No new endpoint or schema.

## Test plan

- Unit: `npx mocha test/support/ai-visibility/handlers/topics.test.js` — added 3 cases (single-LLM backfill from `promptsByTopicIDsTotal`, all-engines via `LLM_ENUM.ALL`, fallback to the row value on error). 123 passing.
- Manual: on Prompt Research → Related Topics, a topic's `promptsCount` now equals the `total` returned by `/topics/research/prompts?topicId=…` for that topic (previously 5 vs 100).

## Related Issues

No linked ticket — surfaced via the llm-optimizer-customer-success Slack thread (TCS AI Day). Problem: Prompt Research topic cards showed a placeholder prompts count that didn't match the expanded prompt list.

### DIFF
--- a/src/support/ai-visibility/handlers/topics.js
+++ b/src/support/ai-visibility/handlers/topics.js
@@ -443,6 +443,33 @@ function topicsResearchComparator(sort) {
   return (a, b) => cmpNum(a.relevanceScore, b.relevanceScore) * dirMult || cmpStr(a.topic, b.topic);
 }
 
+/**
+ * Real prompt count for a single topic, sourced from the SAME gRPC total the expanded
+ * prompt list uses (`promptsByTopicIDsTotal` via `promptsByTopicIdsPage`), so the topic
+ * card's "Prompts count" and the prompt list's "of N" stay consistent by construction.
+ * The `topicsByFTS` row's own `promptsCount` is an FTS-scoped placeholder that understates
+ * the real total, which is the bug this backfills. Falls back to the passed value if the
+ * id is unparseable or the call fails.
+ */
+async function realPromptsCountForTopic(clients, country, llmEnum, topicId, fallback) {
+  let idBig;
+  try { idBig = BigInt(topicId); } catch { return fallback; }
+  try {
+    const res = await clients.promptClient.promptsByTopicIDsTotal({ country, llm: llmEnum, topicIds: [idBig] });
+    return num(res.total);
+  } catch { return fallback; }
+}
+
+// Returns a copy of the topic rows with `promptsCount` replaced by the real per-topic total,
+// resolved in parallel. `llmEnum` mirrors the prompt-list path: the specific engine for a
+// single-llm request, `LLM_ENUM.ALL` otherwise.
+async function withRealPromptsCount(clients, country, llmEnum, rows) {
+  const counts = await Promise.all(
+    rows.map((r) => realPromptsCountForTopic(clients, country, llmEnum, r.topicId, r.promptsCount)),
+  );
+  return rows.map((r, i) => ({ ...r, promptsCount: counts[i] }));
+}
+
 export async function handleTopicsResearch(sp, clients) {
   const country = resolveCountryForFts(sp);
   const { limit, offset } = parseLimitOffset(sp);
@@ -467,13 +494,14 @@ export async function handleTopicsResearch(sp, clients) {
       throw pair[1].reason;
     }
     const raw = pair[1].value;
-    const data = (raw.topics || []).map((t) => ({
+    const rows = (raw.topics || []).map((t) => ({
       topic: t.name,
       topicId: String(t.id),
       topicVolume: num(t.volume),
       promptsCount: num(t.promptsCount),
       relevanceScore: num(t.relevanceScore),
     }));
+    const data = await withRealPromptsCount(clients, country, llm, rows);
     return {
       status: 200,
       body: {
@@ -508,7 +536,7 @@ export async function handleTopicsResearch(sp, clients) {
     }
   }
   const merged = Array.from(seen.values()).sort(topicsResearchComparator(sort));
-  const data = merged.slice(0, limit);
+  const data = await withRealPromptsCount(clients, country, LLM_ENUM.ALL, merged.slice(0, limit));
   return {
     status: 200,
     body: {
--- a/src/support/ai-visibility/handlers/topics.js
+++ b/src/support/ai-visibility/handlers/topics.js
@@ -460,12 +460,37 @@ async function realPromptsCountForTopic(clients, country, llmEnum, topicId, fall
   } catch { return fallback; }
 }
 
-// Returns a copy of the topic rows with `promptsCount` replaced by the real per-topic total,
-// resolved in parallel. `llmEnum` mirrors the prompt-list path: the specific engine for a
-// single-llm request, `LLM_ENUM.ALL` otherwise.
+// Cap on concurrent per-topic prompt-count RPCs. A results page has no hard upper bound
+// (parseLimitOffset defaults to 100 with no max), so an uncapped Promise.all could burst
+// hundreds of parallel gRPC calls and saturate the upstream service under load.
+const PROMPTS_COUNT_FANOUT_CONCURRENCY = 10;
+
+// Bounded-concurrency map: runs `fn` over `items` with at most `concurrency` promises in
+// flight at once, preserving input order.
+async function mapWithConcurrency(items, concurrency, fn) {
+  const results = new Array(items.length);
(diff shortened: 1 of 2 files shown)


src/support/ai-visibility/handlers/topics.js
```diff
@@ -443,6 +443,33 @@ function topicsResearchComparator(sort) {
...
+}
+
+// Returns a copy of the topic rows with `promptsCount` replaced by the real per-topic total,
+// resolved in parallel. `llmEnum` mirrors the prompt-list path: the specific engine for a
+// single-llm request, `LLM_ENUM.ALL` otherwise.
+async function withRealPromptsCount(clients, country, llmEnum, rows) {
+  const counts = await Promise.all(
+    rows.map((r) => realPromptsCountForTopic(clients, country, llmEnum, r.topicId, r.promptsCount)),
+  );
+  return rows.map((r, i) => ({ ...r, promptsCount: counts[i] }));
```
MysticatBot, line 495: issue (blocking): Unbounded parallel gRPC fan-out. `Promise.all` fires one `promptsByTopicIDsTotal` call per topic with no concurrency cap. With up to 100 topics per default page (or more if `parseLimitOffset` does not enforce an upper bound), this creates a burst of parallel RPCs that can saturate the upstream gRPC service or its connection pool under concurrent user load.

In the all-LLMs path, FTS already fans out across engines before reaching this code, compounding the amplification.

Fix: add a concurrency limiter (e.g. `p-limit`) or batch in chunks of 10-20:

```js
import pLimit from 'p-limit';
const cap = pLimit(10);
const counts = await Promise.all(
  rows.map((r) => cap(() => realPromptsCountForTopic(clients, country, llmEnum, r.topicId, r.promptsCount))),
);
```

Alternatively, confirm with the upstream team whether `promptsByTopicIDsTotal` with multiple `topicIds` can return per-topic counts in a single batch call (the proto already accepts an array). That would collapse N calls into 1.

MysticatBot (changes requested):
Hey @calvarezg,

**Verdict: Request changes** - unbounded parallel gRPC fan-out risks upstream saturation under load.
**Complexity: MEDIUM** - small diff; API surface handler change.
_Changes: Backfills per-topic prompt count from the real gRPC total instead of the FTS-scoped placeholder, with graceful fallback on failure (2 files)._

## Must fix before merge

1. [Important] Unbounded parallel gRPC fan-out per topic row with no concurrency cap - `src/support/ai-visibility/handlers/topics.js:470` (details inline)

<details>

- nit: Silent `catch { return fallback }` on the gRPC call hides persistent failures with no observability signal. A `log.warn` on the gRPC failure path would surface degradation in dashboards. - `src/support/ai-visibility/handlers/topics.js:459`
- suggestion: Assert `arg.llm` in both the single-LLM and all-LLMs tests to verify the correct enum is forwarded - without this a regression that always passes the wrong `llm` goes undetected. - `test/support/ai-visibility/handlers/topics.test.js:377`
- suggestion: Add a test case with a non-numeric `topicId` (e.g. `'abc'`) to exercise the `BigInt` parse-error fallback path distinctly from the gRPC-failure path.

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 10m 55s | <b>Cost:</b> $4.34 | <b>Commit:</b> <code>7f635989ca2185bd5a1cc3d2a26f4e473a4149ba</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

dmitriigulkin-semrush:
hey @calvarezg, great stuff!
one thing tho, `clients.topicClient.topicsByFTS` should already return prompt_count inside the response, so we shouldn't actually make an adittional request for that, current schema as follow:
```ts
{
            "id": "12773975256618769407",
            "name": "PDF Files and Format",
            "volume": "278278",
            "volume_trend": [
                0,
                74,
                70,
                87,
                91,
                86,
                86,
                86,
                100,
                100,
                100,
                100
            ],
            "intents": [
                {
                    "intent": "TASK",
                    "weight": 12
                },
                {
                    "intent": "INFORMATIONAL",
                    "weight": 72
                },
                {
                    "intent": "NAVIGATIONAL",
                    "weight": 5
                },
                {
                    "intent": "COMMERCIAL",
                    "weight": 8
                },
                {
                    "intent": "TRANSACTIONAL",
                    "weight": 3
                }
            ],
            "prompts_count": "516",
            "relevance_score": 92
        },
```

dmitriigulkin-semrush (approved):

calvarezg:
## Triage: comments not being addressed

Reviewed and intentionally skipping the following — rationale included.

- [Nit: silent catch hides failures / add log.warn (https://github.com/adobe/spacecat-api-service/pull/2843#pullrequestreview-4718853974)](https://github.com/adobe/spacecat-api-service/pull/2843#pullrequestreview-4718853974) — **Dismissed.** These handlers take `(sp, clients)` with no logger threaded in, and the entire file uses silent `.catch()` by deliberate convention (`countTopicRowsByTopicsByFtsPaging`, `fetchRelatedTopicsAiVolumeMetrics`, etc.). The fallback degrades gracefully to the previous placeholder value. Adding a logger to this handler layer is a cross-cutting change out of scope for this fix — better as a file-wide observability follow-up than piecemeal here.

MysticatBot (approved):
Hey @calvarezg,

**Verdict: Approve** - concurrency cap addresses the prior blocking concern; code is correct and well-tested.
_Changes: Backfills per-topic prompt count from the real gRPC total instead of the FTS-scoped placeholder, with a bounded-concurrency helper capped at 10 (2 files)._

<details>

- nit: `mapWithConcurrency` relies on the subtle invariant that `const i = next; next += 1` executes atomically before any `await` yields (single-threaded event loop). A one-line inline comment noting this would prevent a future maintainer from inserting an `await` between the read and increment. - `src/support/ai-visibility/handlers/topics.js:475`
- suggestion: All 4 new tests pass a single-element topics array, so the concurrency cap is never actually exercised. Adding one test with 3+ topics in the FTS response would validate that `mapWithConcurrency` preserves result ordering across multiple concurrent slots. - `test/support/ai-visibility/handlers/topics.test.js:353`

</details>

- Unbounded parallel gRPC fan-out now capped at 10 via `mapWithConcurrency`
- Tests assert `arg.llm` forwards the correct enum for both single-LLM and all-LLMs paths
- Non-numeric `topicId` edge case now covered with dedicated BigInt fallback test
- Silent catch / logging nit accepted as file-wide convention (no logger threaded into these handlers)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 47s | <b>Cost:</b> $4.51 | <b>Commit:</b> <code>55df718236de748471b1e2a4ffbf8697009cef96</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

calvarezg:
Thanks @dmitriigulkin-semrush — you're right, and I confirmed it by calling the gRPC directly. `topicsByFTS.promptsCount` already equals `promptsByTopicIDsTotal` for the same `llm` (checked `ALL`: 1779==1779, 872==872, …; and per-engine: 5==5). So the extra call is redundant.

Root cause of the original "5 vs 100": the **all-engines** path fanned out `topicsByFTS` per engine and, when deduping a topic, kept the *first engine's* count (the per-engine 5) instead of the ALL-level total.

Merging this now to unblock the TCS demo (output is correct, fan-out is concurrency-capped), and I've filed a follow-up to drop the extra call and source the count straight from `topicsByFTS` at ALL-level — likely sourcing the all-engines list from an `ALL` query so each row already carries the right count. Appreciate the catch!

PR #2852, opened 2026-07-19
feat: upsert on conflict for state access-mapping create

ravverma:
## What

Makes `POST /state/access-mappings` (`createMapping`) **upsert-natured**. When the DB insert conflicts with an existing active binding for the same subject + resource, instead of returning `409` the endpoint now **overwrites the existing row's `granted_capabilities`** with the request's set and returns `200` with the updated mapping.

## Why

Clients re-issuing a create for an already-bound subject+resource previously had to catch the `409`, then issue a follow-up `PATCH`. Upsert collapses that into a single idempotent call.

## How

On the conflict path (`created.length === 0 && skipped.length > 0`):
1. Look up the existing row by natural key to get its `id` (unchanged).
2. Call `updateFacsAccessMappingCapabilities` — the same SECURITY-DEFINER capability-edit RPC `PATCH` uses — with `capabilitiesToStore` (de-duped request caps + auto-injected `<product>/can_view` baseline).
3. Emit an `update_capabilities` audit event (status 200) and return the updated mapping DTO.
4. **Fail-safe:** if the existing row can't be found, or the RPC finds no active row (the row-revoked-mid-flight race), fall back to the original `409` rather than a 500.

Capabilities are **replaced**, not merged — consistent with `PATCH` semantics.

## Tests
- **Unit** (`test/controllers/state-access-mappings.test.js`): upsert overwrite → 200, audit event assertion, and two `409` race-fallback cases (lookup miss, revoked-before-update). 111 passing.
- **Integration** (`test/it/shared/tests/state-access-mappings.js`): duplicate-POST now asserts 200 + same id + overwritten capabilities.

## Docs
- `docs/openapi/state-access-mappings-api.yaml`: documents upsert semantics, adds `200`, narrows `409` to the revoke race.

### DIFF
--- a/docs/openapi/state-access-mappings-api.yaml
+++ b/docs/openapi/state-access-mappings-api.yaml
@@ -59,11 +59,17 @@ state-access-mappings:
   post:
     tags:
       - state-access-mappings
-    summary: Create a state-layer access mapping
+    summary: Create (upsert) a state-layer access mapping
     description: |
       Inserts one `facs_access_mappings` binding row. Capability is
       validated against the product's capability catalog. For `subjectType:
       org`, `subjectId` must equal the caller's canonical org id.
+
+      **Upsert semantics:** when an active binding already exists for the same
+      subject + resource, its `grantedCapabilities` are **overwritten** with the
+      request's set (the given capabilities *replace* the stored ones — this is
+      not a merge) and the endpoint returns `200` with the updated mapping. A
+      brand-new binding returns `201`.
 
       `grantedCapabilities` must be non-empty (a `400` is returned otherwise — to
       remove all access, use `DELETE` on the created row). The product's
@@ -83,6 +89,14 @@ state-access-mappings:
           schema:
             $ref: './schemas.yaml#/StateAccessMappingCreateRequest'
     responses:
+      '200':
+        description: >-
+          An active mapping already existed for this subject and resource; its
+          granted capabilities were overwritten with the request's set (upsert).
+        content:
+          application/json:
+            schema:
+              $ref: './schemas.yaml#/StateAccessMapping'
       '201':
         description: Mapping created.
         content:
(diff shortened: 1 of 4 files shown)


MysticatBot (approved):
Hey @ravverma,

**Verdict: Approve** - well-scoped upsert with thorough race-condition handling and adequate test coverage.
**Complexity: HIGH** - medium diff; API surface risk (OpenAPI contract change + FACS/ReBAC surface).
_Changes: Converts POST /state/access-mappings from create-only to upsert-on-conflict, overwriting capabilities on active duplicates and returning 200 instead of 409 (4 files)._

<details>

- nit: The two 409 race-condition paths return inconsistent `id` values - the lookup-miss path returns `id: null` while the revoked-before-update path returns `id: conflictId` (a defunct row id). Since both paths tell the client to retry and neither id is actionable, normalizing both to `null` would be cleaner. - `src/controllers/state-access-mappings.js:757`
- suggestion: Add an info-level log before the upsert update call (e.g., `log.info({ tag: 'state-access-mappings', id: conflictId }, 'Upsert: overwriting capabilities on existing mapping')`) for debugging scenarios where a caller unknowingly overwrote an existing binding. The PATCH path has no equivalent log either, but PATCH is an explicit user action on a known id while upsert is an implicit mutation triggered by what the caller may believe is a create. - `src/controllers/state-access-mappings.js:752`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 11m 12s | <b>Cost:</b> $6.39 | <b>Commit:</b> <code>4738df8b39c2afc6d0f5282bd276701fa2067100</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2854, opened 2026-07-20
fix(serenity): classify disguised-405 quota rejections as a stable 40…

claudiaboldis:
Wires the existing (but never-invoked) isMeteredQuota classifier into the one call site where a project publish genuinely 405s only for quota reasons: handleCreateMarketSubworkspace's require/best-effort publish branches. A quota rejection on a REQUIRED publish now throws a typed ErrorWithStatusCode (409, code=quotaExceeded) instead of propagating the raw 405, which mapError's generic branch would otherwise flatten into a misleading 502 serenityUpstreamError.

This is step 1 of the sequencing in serenity-docs#72: the customer-facing message for case 1 (brand carve exhausted, allocator OFF - production today) needs a stable token to key off, and this also activates the MeteredQuotaClassifier CloudWatch metric (previously always zero, since no production caller invoked isMeteredQuota), unlocking the alerting route in the same spec's §5 with no new request-path code.

Follow-up commit addresses MysticatBot's review: the publish-catch sites now classify via the existing body-SHAPE `isMeteredQuota` predicate instead of a bare `status === 405` check, so a genuine JSON-bodied "Method Not Allowed" is never absorbed as a quota rejection, and test coverage was added for the best-effort branch's metric-emission side effect.

Introduced by: N/A

- [x] N/A — this PR does not add, remove, or change any endpoint or request/response shape. It adds a new internal error token (`quotaExceeded`) to an existing 409 error response on the already-documented-as-undocumented serenity error-token surface — consistent with the precedent set by #2813, which added `orgPoolExhausted`/`brandAiLimit` the same way without an OpenAPI spec change (serenity error tokens are not currently modeled in `docs/openapi/`).

- [x] N/A — no entity shape or exposed field changes; only internal error classification.

- [x] N/A — no new audit type.

## Related Issues

- No dedicated `spacecat-api-service` issue exists for this specific fix. The work is scoped and sequenced in [adobe/serenity-docs#72](https://github.com/adobe/serenity-docs/pull/72) ("Prompt quota exhaustion — customer experience, atomic failure handling, and alerting spec"), §2/§4.1 (case 1: brand carve exhausted, allocator OFF). This PR is step 1 of that spec's implementation sequencing (posted as a comment on that PR): classify the disguised-405 quota rejection into a stable `quotaExceeded` 409 token, which the customer-facing message and CloudWatch alerting (later steps) key off.
- Builds directly on the allocator work in [#2813](https://github.com/adobe/spacecat-api-service/pull/2813) (LLMO-6190) and [#2764](https://github.com/adobe/spacecat-api-service/pull/2764) (LLMO-5616).

Thanks for contributing!

### DIFF
--- a/src/support/serenity/handlers/markets-subworkspace.js
+++ b/src/support/serenity/handlers/markets-subworkspace.js
@@ -15,9 +15,8 @@
 import { hasText } from '@adobe/spacecat-shared-utils';
 
 import { ErrorWithStatusCode } from '../../utils.js';
-import { SerenityTransportError } from '../rest-transport.js';
 import {
-  ERROR_CODES, isUpstreamGone, toQuotaExceededError,
+  ERROR_CODES, isUpstreamGone, isMeteredQuota, toQuotaExceededError,
 } from '../errors.js';
 import { normalizeGeoTargetId, normalizeLanguageCode } from '../validation.js';
 import {
@@ -595,10 +594,14 @@ export async function handleCreateMarketSubworkspace(
     } catch (e) {
       // Case 1 (serenity-docs#72 §2): the brand carve is exhausted (allocator OFF — production
       // today), or `retryOnQuota`'s bounded top-up+retry (LLMO-6190 item 4) still didn't clear
-      // the disguised quota 405. Surface it as the stable `quotaExceeded` 409 token instead of
-      // letting it fall through mapError's generic `serenityUpstreamError` 502 — a caller can
-      // then show the contractual-limit message and never retry a rejection that can't succeed.
-      if (e instanceof SerenityTransportError && e.status === 405) {
+      // the disguised quota 405. Classify via `isMeteredQuota` (body-SHAPE check, not bare
+      // status — MysticatBot review) so a genuine app-level 405 Method-Not-Allowed (JSON body)
+      // is never mistaken for a quota rejection and hidden behind the wrong error token; this
+      // also gives the shared `MeteredQuotaClassifier` metric (LLMO-6191) a live call site.
+      // Surface a real quota match as the stable `quotaExceeded` 409 token instead of letting it
+      // fall through mapError's generic `serenityUpstreamError` 502 — a caller can then show the
+      // contractual-limit message and never retry a rejection that can't succeed.
+      if (isMeteredQuota(e)) {
         log?.warn?.('handleCreateMarketSubworkspace: publish rejected — quota exceeded', {
           workspaceId, projectId,
         });
@@ -611,15 +614,16 @@ export async function handleCreateMarketSubworkspace(
       await headroom.retryOnQuota(() => transport.publishProject(workspaceId, projectId));
       published = true;
     } catch (e) {
-      if (e instanceof SerenityTransportError && e.status === 405) {
+      if (isMeteredQuota(e)) {
         // Swallowed by design (best-effort provisioning must not fail the brand create), but this
         // IS a quota rejection and the event that creates the dark draft market a customer later
         // trips over — serenity-docs#72 §5 requires it to alert even though nothing failed here.
         // TODO(serenity-docs#72 step 2): fire the Slack alert here once the alerting mechanism
-        // lands; `recordRejection('quotaExceeded')` inside toQuotaExceededError already feeds the
-        // CloudWatch metric this alarm will key on, so calling it (without throwing) keeps the
-        // classifier + alerting signal consistent between this swallowed path and the `require`
-        // path above.
+        // lands; the call below is side-effect-only (its returned error is intentionally
+        // discarded, never thrown — best-effort swallows it) purely to run
+        // `recordRejection('quotaExceeded')` inside it, so the CloudWatch metric this alarm will
+        // key on fires here too, keeping the classifier + alerting signal consistent between
+        // this swallowed path and the `require` path above (MysticatBot review, non-blocking nit).
         toQuotaExceededError();
         log?.warn?.('handleCreateMarketSubworkspace: publish skipped — quota exceeded, project left as draft', {
           workspaceId, projectId,
--- a/src/support/serenity/handlers/markets-subworkspace.js
+++ b/src/support/serenity/handlers/markets-subworkspace.js
@@ -16,7 +16,9 @@ import { hasText } from '@adobe/spacecat-shared-utils';
 
 import { ErrorWithStatusCode } from '../../utils.js';
 import { SerenityTransportError } from '../rest-transport.js';
-import { ERROR_CODES, isUpstreamGone } from '../errors.js';
+import {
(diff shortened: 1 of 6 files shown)


src/support/serenity/handlers/markets-subworkspace.js
```diff
@@ -587,15 +589,39 @@ export async function handleCreateMarketSubworkspace(
...
+    try {
+      await headroom.retryOnQuota(() => transport.publishProject(workspaceId, projectId));
+      published = true;
+    } catch (e) {
+      // Case 1 (serenity-docs#72 §2): the brand carve is exhausted (allocator OFF — production
+      // today), or `retryOnQuota`'s bounded top-up+retry (LLMO-6190 item 4) still didn't clear
+      // the disguised quota 405. Surface it as the stable `quotaExceeded` 409 token instead of
+      // letting it fall through mapError's generic `serenityUpstreamError` 502 — a caller can
+      // then show the contractual-limit message and never retry a rejection that can't succeed.
+      if (e instanceof SerenityTransportError && e.status === 405) {
```
MysticatBot: issue (blocking): This catches ALL `SerenityTransportError` with `status === 405`, but the existing `isMeteredQuota` classifier in `errors.js` distinguishes disguised quota rejections (string body from nginx) from genuine HTTP 405s (JSON body from the app). Two views on whether this matters:

1. If `publishProject` can genuinely only 405 for quota reasons (as the PR description states), the broader check is defensible and `isMeteredQuota`'s body-shape test is over-constraining.
2. If a genuine "Method Not Allowed" could surface from this endpoint (e.g., routing misconfiguration), it would be silently misclassified as `quotaExceeded`, hiding a real integration failure.

Either way, the `recordMeteredQuotaClassifier` observability metric (LLMO-6191) is never emitted by this path since `isMeteredQuota` is not called - that metric was designed to track classifier accuracy at this exact boundary.

If the broad check is intentional, add a one-line comment documenting the assumption (e.g., "publishProject only 405s for quota; shape check via isMeteredQuota unnecessary here"). If it should be precise, swap to `if (isMeteredQuota(e))` in both the require and best-effort branches.

src/support/serenity/handlers/markets-subworkspace.js
```diff
@@ -587,15 +589,39 @@ export async function handleCreateMarketSubworkspace(
...
-        log?.warn?.('handleCreateMarketSubworkspace: publish skipped — quota 405, project left as draft', {
+        // Swallowed by design (best-effort provisioning must not fail the brand create), but this
+        // IS a quota rejection and the event that creates the dark draft market a customer later
+        // trips over — serenity-docs#72 §5 requires it to alert even though nothing failed here.
+        // TODO(serenity-docs#72 step 2): fire the Slack alert here once the alerting mechanism
+        // lands; `recordRejection('quotaExceeded')` inside toQuotaExceededError already feeds the
+        // CloudWatch metric this alarm will key on, so calling it (without throwing) keeps the
+        // classifier + alerting signal consistent between this swallowed path and the `require`
+        // path above.
+        toQuotaExceededError();
```
MysticatBot, line 627: issue (blocking): This `toQuotaExceededError()` call fires `recordRejection('quotaExceeded')` as a side effect, but no test asserts the metric is emitted in the best-effort path. The require-mode tests implicitly prove `toQuotaExceededError` was called (they assert error properties on the thrown error), but the best-effort tests only check `published === false` and response status.

If someone removes this line thinking it is dead code (the return value IS discarded and no assertion covers the side-effect), the CloudWatch metric goes silent with no test failure. Add a spy on `recordRejection` and assert it was called with `'quotaExceeded'` in the existing best-effort 405 test.

MysticatBot (changes requested):
Hey @claudiaboldis,

**Verdict: Request changes** - the 405 guard should either use the existing shape-based classifier or document why the broader check is intentional.
_Changes: Classifies disguised-405 quota rejections as a stable quotaExceeded 409 token in the require/best-effort publish paths (6 files)._

## Must fix before merge
1. [Important] Handler catches all 405s instead of using `isMeteredQuota` shape check - `src/support/serenity/handlers/markets-subworkspace.js:601` (details inline)
2. [Important] Best-effort path metric emission has no test coverage - `src/support/serenity/handlers/markets-subworkspace.js:623` (details inline)

<details>

- nit: `toQuotaExceededError()` called purely for side-effect (return value discarded) reads as a bug to future maintainers - consider calling `recordRejection('quotaExceeded')` directly or adding an inline comment on the same line - `src/support/serenity/handlers/markets-subworkspace.js:623`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 12s | <b>Cost:</b> $6.39 | <b>Commit:</b> <code>daa83a967c0ca5a2acc0ea651ee539360c0ce7ed</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @claudiaboldis,

**Verdict: Approve** - prior findings addressed cleanly; no new blocking concerns.
_Changes: Classifies disguised-405 quota rejections via the existing `isMeteredQuota` body-shape classifier and surfaces them as a stable `quotaExceeded` 409 token in both require/best-effort publish paths (6 files)._

<details>

- suggestion: `toQuotaExceededError()` bundles `recordRejection` inside the factory, diverging from `resource-manager.js` where sibling factories (`orgPoolExhausted`, `brandAiLimit`) are side-effect-free and the metric fires at the throw site. If the factory is ever constructed conditionally without throwing, a phantom metric fires. Consider extracting `recordRejection('quotaExceeded')` to the two call sites for consistency - `src/support/serenity/errors.js:167`
- nit: best-effort path fires the metric (via `toQuotaExceededError()`) before the `log?.warn?.(...)` that provides workspaceId/projectId context, inverting the order used in the require path. Low impact but the asymmetry could confuse timeline-based debugging - `src/support/serenity/handlers/markets-subworkspace.js:623`
- nit: `@param` union on `recordRejection` now lists four literals inline; if the enum grows further, a shared `RejectionReason` typedef would keep the type annotation in one place - `src/support/serenity/allocation-metrics.js:129`
</details>

- Handler now uses `isMeteredQuota(e)` body-shape classifier instead of bare `instanceof SerenityTransportError && status === 405`
- Best-effort path metric emission covered by a spy test on `toQuotaExceededError`
- Side-effect-only call documented with inline comment explaining intentional discard

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 23s | <b>Cost:</b> $4.07 | <b>Commit:</b> <code>5f4b30c6bbb1370cd264772977dac777a6640a14</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2825, opened 2026-07-15
WP-O2: origin dimension — tolerant resolver + derivation

byteclimber:
## Summary
Part of the origin-dimension rename: adobe/serenity-docs#46. Combines WP-O2a (tolerant resolver + root rename) and WP-O2b (derivation).

**WP-O2a**
- Renamed `DIMENSION.SOURCE`→`DIMENSION.ORIGIN`, `SOURCE_VALUE`→`ORIGIN_VALUE`, etc. in `prompt-tags.js`
- Added `resolveAuthorshipRoot` in `tag-tree.js`: reuses an existing `origin` root, else adopts a legacy `source` root only if its children ⊆ `{ai, human}`, else mints a new `origin` root. Wired into `ensureDimensionRoots`.
- `handlers/tags.js`, `markets-subworkspace.js`, `brand-provisioning.js`, `serenity.js` needed no changes — they already resolve via the shared constants.
- Deliberately left `src/support/elements/**` untouched (separate, already-flagged production-triage concern).

**WP-O2b**
- `makeTypeInjector` now strips+injects `origin`: CREATE derives `ai`/`human`, UPDATE re-injects the already-stored value, never re-derives.
- New `resolveOriginForWrite(candidate, isServicePrincipal)`: service principal may assert a value (invalid falls back to `human`), user principal always overridden to `human`.
- Removed the `origin` PATCH path from `updatePromptById` and the `|| 'human'` fallback from `mapRowToPrompt`.
- OpenAPI docs updated to describe `origin` as service-principal-only on write.

## Test plan
- [x] Targeted unit slice (serenity + prompts-storage + brands + serenity-controller): 1696/1696 passing
- [x] Full suite: 14873/14877 (4 failures are pre-existing `test/controllers/audits.test.js` order-dependent flakes, unrelated, pass in isolation)
- [x] Lint + `tsc` type-check clean
- [ ] `it-postgres` serenity IT suite **not updated** — it drives the real mock Docker image still pinned at `@adobe/spacecat-shared-project-engine-client@1.10.0` (pre-WP-O1). Needs a follow-up once adobe/spacecat-shared#1815 (WP-O1) releases and this repo's dependency bumps.

Depends on: adobe/spacecat-shared#1815 (WP-O1) for full IT verification.

Closes #2816, #2817
Jira: SITES-48001, SITES-48002 · Tracking: cross-repo plan in adobe/serenity-docs#46

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2864, opened 2026-07-20
feat(serenity): WP-O2b origin derivation — salvaged from #2825 (LLMO-6275)

aliciadriani:
## What & why
**WP-O2b — origin derivation (LLMO-6275).** Completes the origin dimension's write-side on top of WP-O2a (#2862, now merged to `main`): the server derives and injects the `origin` tag on every prompt write, and it becomes read-only after create.

Recovered from byteclimber's #2825 (whose O2b half was clobbered by a main-merge; salvaged from commit `74bbc00`) and integrated with main's current in-place edit path.

## Changes
- **`makeTypeInjector` now injects `origin` as well as `type`** (`handlers/prompts.js`, + subworkspace twin): CREATE derives to `ai`; UPDATE re-injects the caller's existing origin tag (preserving `human`), falling back to `ai` only if none — never a strip-without-inject. Strip is by resolved tag id under the root, never by name.
- **`tag-tree.js`: `resolveOriginValueInjection`** — mirrors `resolveTypeValueInjection`, resolves the origin root through the tolerant `ensureDimensionRoots` so it's correct mid-rename.
- **`prompts-storage.js`: `resolveOriginForWrite`** — origin comes from the authenticated principal (service principal may assert a valid enum; a user request is always `human`; invalid → `human`, never a 4xx). `mapRowToPrompt` drops the dead `|| 'human'` fallback; `updatePromptById` no longer patches `origin` (set once at create).
- **`brands.js`**: threads `isServicePrincipal` (from `context.s2sConsumer`).
- **OpenAPI**: `origin` documented service-principal-only on `V2PromptInput`, ignored on PATCH.

## Verification
- Unit + it-postgres green (Node 24). New test covers the create/update asymmetry (an edit preserves an existing `human` origin, never relabels to `ai`). Lint clean.

## Depends on
WP-O2a #2862 (merged). Part of the origin dimension — serenity-docs#46, epic LLMO-6270.


src/support/serenity/handlers/prompts.js
```diff
@@ -366,60 +367,115 @@ export async function createOnePrompt(transport, semrushWorkspaceId, projectId,
...
-    return { ...input, tagIds: [...stripped, computedId] };
+    const { computedId: aiId, originTagIds } = await originPending;
+    const stripped = out.tagIds.filter((id) => !originTagIds.includes(id));
+
+    // UPDATE re-injects whichever origin tag the caller already carried — it
+    // never derives a new one. CREATE always derives to AI. Falling back to AI
+    // when update finds none avoids a strip-without-inject (which would make
+    // the prompt invisible to origin-based filtering) for a prompt tagged
+    // before this dimension existed.
+    const nextOriginId = mode === 'update'
```
aliciadriani, line 473: This is the create/update asymmetry, and it's correct: UPDATE re-injects the origin id the caller already carries (`out.tagIds.find((id) => originTagIds.includes(id))`) and only falls back to `aiId` when none is present; CREATE always uses `aiId`. Because `originTagIds` is every id under the resolved root, the preceding `stripped` filter removes any prior origin id before this re-appends exactly one — so an edit that round-trips `originHuman` keeps human, never relabels to ai. Verified by the new `preserves an existing human origin on edit` test. The fallback-to-ai for a pre-dimension prompt (avoiding strip-without-inject) is the right call — a stripped-but-uninjected prompt would drop out of origin-filtered reads.

src/support/serenity/handlers/prompts.js
```diff
@@ -366,60 +367,115 @@ export async function createOnePrompt(transport, semrushWorkspaceId, projectId,
...
+        ORIGIN_VALUE.AI,
+        log,
+      );
+      originCache.set(originKey, originPending);
     }
-    const { computedId, typeTagIds } = await pending;
-    const stripped = input.tagIds.filter((id) => !typeTagIds.includes(id));
-    return { ...input, tagIds: [...stripped, computedId] };
+    const { computedId: aiId, originTagIds } = await originPending;
+    const stripped = out.tagIds.filter((id) => !originTagIds.includes(id));
```
aliciadriani, line 466: Strip-by-id confirmed: filters against `originTagIds` (the ids under the authorship root via `[...byName.values()]`), never by name — so a sub-category legitimately named `ai`/`human` under a different root is untouched. Matches the `type` strip and the lens requirement.

src/support/serenity/tag-tree.js
```diff
@@ -608,3 +608,46 @@ export async function resolveTypeValueInjection(
...
+ *
+ * @param {object} transport - Serenity transport (Semrush proxy client).
+ * @param {string} semrushWorkspaceId
+ * @param {string} projectId
+ * @param {string} wantValue - the bare `origin` value (`ai` / `human`).
+ * @param {object} [log] - logger.
+ * @returns {Promise<{ computedId: string, originTagIds: string[] }>} `computedId`
+ *   is always resolved — {@link ensureChildren} throws rather than leave a hole.
+ */
+export async function resolveOriginValueInjection(
```
aliciadriani, line 632: `resolveOriginValueInjection` resolves the root through `ensureDimensionRoots`, so it inherits O2a's tolerant `origin`/legacy-`source` resolution — the returned `originTagIds`/`computedId` reference whichever physical authorship root the project currently carries, and it never mints a second one. Correct dependency on the O2a seam, and the docstring calls this out.

src/support/prompts-storage.js
```diff
@@ -18,6 +18,33 @@ import { classifyIntents } from './intent-classifier.js';
...
+ * human hitting this endpoint is definitionally authoring the prompt
+ * themselves, so the body's `origin` (if any) is silently overridden, never
+ * used to reject the request.
+ *
+ * @param {*} candidate - the request body's asserted `origin`, if any.
+ * @param {boolean} isServicePrincipal - true when the caller authenticated as
+ *   an S2S service (see `context.s2sConsumer` at the controller boundary).
+ * @returns {'ai' | 'human'}
+ */
+export function resolveOriginForWrite(candidate, isServicePrincipal) {
```
aliciadriani, line 42: `resolveOriginForWrite` is correct and well-tested as a unit: service principal + valid enum → honored; service principal + unrecognized/absent → `human` (never throws); user principal → always `human`. The 'never reject over this field' property is important and upheld (unknown value coerces, doesn't 400). The gap is only that the CONTROLLER mapping feeding `isServicePrincipal` (from `context.s2sConsumer`) isn't tested end-to-end — see the gate 5/6 Should-Fix in the summary.

src/controllers/brands.js
```diff
@@ -554,6 +554,11 @@ function BrandsController(ctx, log, env) {
 
       const { postgrestClient } = context.dataAccess.services;
       const updatedBy = context.attributes?.authInfo?.profile?.email || 'system';
+      // Service principals (e.g. llmo-data-retrieval-service) authenticate over
+      // s2sAuthWrapper, which sets `context.s2sConsumer`; a user principal never
+      // carries it. Gates whether a per-prompt `origin` in the body is honored —
+      // see `resolveOriginForWrite`.
+      const isServicePrincipal = Boolean(context.s2sConsumer);
```
aliciadriani, line 561: `isServicePrincipal = Boolean(context.s2sConsumer)` — verified `s2sConsumer` is the real attribute set by `s2sAuthWrapper` (same usage as `llmo-mysticat-controller.js`), so the detection is sound. But this exact line is where gate 5/6 needs a controller test: no case in `test/controllers/brands.test.js` sets `context.s2sConsumer`, so the TRUE branch (service principal honored) is never asserted at the endpoint. Please add the two controller cases (service→honored, user→coerced) here.

src/support/prompts-storage.js
```diff
@@ -1050,9 +1083,9 @@ export async function updatePromptById({
   if (updates.status !== undefined) {
     patch.status = updates.status;
   }
-  if (updates.origin !== undefined) {
-    patch.origin = updates.origin;
-  }
+  // `origin` is never patchable via the PATCH body — mirrors `source`, which is
```
aliciadriani, line 1086: UPDATE-preserve on the storage side: `origin` correctly dropped from the PATCH `patch` (mirrors `source`), so a PATCH body can never re-write origin. Combined with the injector's update-mode re-injection, origin is set once at create and never re-derived. Consistent with the OpenAPI note added to `prompts-v2-api.yaml` that PATCH `origin` is ignored.

docs/openapi/schemas.yaml
```diff
@@ -6643,6 +6643,11 @@ V2PromptInput:
       type: string
       enum: [ai, human]
       default: human
+      description: >-
+        Who authored the prompt. Service-principal only: honored (and
```
aliciadriani, line 6647: Item-6 confirmed: `V2PromptInput.origin` is KEPT (not deleted), now documented as service-principal-only + user-coerced-to-human. Retaining the field is what prevents relabeling DRS-written prompts. Doc matches the `resolveOriginForWrite` behavior.

aliciadriani (commented):
Reviewed WP-O2b (LLMO-6275): origin derivation + injector generalization, stacked on `claude/wp-o2a-origin`. Base is a stacked branch, so MysticatBot is N/A (it skips non-main bases) — `/review-pr` clean is the bar. The ⛔ banner is expected. it-postgres pending (waits WP-O1 release) is not a finding.

Applied the full correctness lens — all four invariants hold:

**1. create-inject / update-preserve ASYMMETRY — CONFIRMED correct.** In `makeTypeInjector`, `nextOriginId = mode === 'update' ? (out.tagIds.find((id) => originTagIds.includes(id)) ?? aiId) : aiId`. CREATE always derives `AI`; UPDATE re-injects whichever origin id the caller round-tripped and only falls back to `AI` when none is present (a pre-dimension prompt). Both call sites pass `{ mode: 'update' }` on the update path (`handleUpdatePrompt`, `handleUpdatePromptSubworkspace`) and default to create elsewhere. The new test `preserves an existing human origin on edit — never relabels it to ai` directly guards the relabel bug. This is the point of the WP and it's right.

**2. Strip by resolved tag id, never by name — CONFIRMED.** `stripped = out.tagIds.filter((id) => !originTagIds.includes(id))` where `originTagIds = [...byName.values()]` are the ids under the resolved authorship root. Mirrors the `type` strip exactly. No name-based stripping anywhere.

**3. `mapRowToPrompt` returns `row.origin` with no `|| 'human'` fallback — CONFIRMED**, with the test updated to assert `undefined` passes through. Reasoning (column has zero NULLs in prod; the fallback would mislabel a null as human) is sound.

**4. Item 6 — `origin` REMAINS on `V2PromptInput` — CONFIRMED.** `schemas.yaml` keeps the field (adds a description, enum `[ai,human]`, default `human`); it is not deleted. Deleting it would have relabeled DRS-written prompts, so keeping it is correct.

**Also verified (gate wiring at source level):** `isServicePrincipal = Boolean(context.s2sConsumer)` in `brands.js` uses the established `s2sConsumer` context attribute (set by `s2sAuthWrapper`, used the same way in `llmo-mysticat-controller.js` and others), so the service-principal detection is real, not a typo'd nonexistent property. And `resolveOriginValueInjection` resolves the root through `ensureDimensionRoots`, so it inherits O2a's tolerant `origin`/legacy-`source` resolution and never mints a second root.

---

## SHOULD-FIX — missing spec-gate 5/6 coverage (implementer's flag CONFIRMED)

The WP-O2b plan states "spec gates 5–8 land here." Gates **5** (a SERVICE principal's `origin:'ai'` is honored) and **6** (a USER principal's `origin` is coerced to `human`, body never rejected) are the v2-controller-boundary gates. They are **not** covered:

- `test/controllers/brands.test.js` is NOT in this PR's diff, and I grepped the (base) file directly: **0** occurrences of `s2sConsumer` and **0** of `isServicePrincipal` (the 16 `origin` hits are all `origin: 'human'` fixture data, not principal-derivation assertions).
- The added tests cover the derivation logic only in ISOLATION: `resolveOriginForWrite(candidate, isServicePrincipal)` as a pure unit, and `upsertPrompts({ isServicePrincipal })` with the flag passed directly. Neither exercises the controller's `context.s2sConsumer` → `isServicePrincipal` mapping.
- Consequence: codecov/patch is green because the existing `createPromptsByBrand` controller tests execute the new `Boolean(context.s2sConsumer)` line — but only the FALSE (user) branch, and with no assertion tying a service-principal request to a honored `origin:'ai'` nor a user request to a coerced `human`. The gate-5 TRUE branch at the controller boundary is untested end-to-end.

Recommended: add controller-level cases to `test/controllers/brands.test.js` (the file already exists, 31 `createPromptsByBrand` cases) — one with `context.s2sConsumer` set asserting a body `origin:'ai'` is honored (gate 5), one as a user principal asserting `origin:'ai'` is coerced to `human` with a 2xx / no rejection (gate 6). Routing to the implementer per the loop. (Gates 7/8 — the update-preserve asymmetry and tolerant-root strip-by-id — ARE covered by the new handler tests.)

No Must-Fix. Recording `/review-pr` NOT-yet-clean pending the gate 5/6 coverage; bot N/A (non-main base).

aliciadriani:
Superseded by #2866. The canonical WP-O2b (LLMO-6275) is branch `claude/wp-o2b-derivation`, which carries the gate5/6 controller coverage this salvage branch (from #2825) lacked, uses the create/update-asymmetric `makePromptTagInjector` matching spec §3 item 3, and is the confirmed base for WP-S2 (#2867). Closing as superseded — history preserved here and referenced from #2866. No code lost.

aliciadriani:
Closing in favor of #2866 — the canonical WP-O2b PR (broader `makePromptTagInjector` refactor + `resolveClosedValueInjection`, already MysticatBot-reviewed). This salvage's one advantage — the safer `isServicePrincipal = Boolean(context.s2sConsumer)` default — is being ported into #2866's must-fix #1. LLMO-6275.

PR #2867, opened 2026-07-20
feat(serenity): server-owned source dimension + canonicalizeSource (LLMO-6282) [WP-S2]

aliciadriani:
## WP-S2 — api-service: `SERVER_OWNED_DIMENSIONS` + `canonicalizeSource`

Introduces the 5th dimension root **`source`** (the producing system) as an **OPEN, SERVER-OWNED** dimension on the Semrush tag tree, reusing the name the authorship root vacated in the 46 rename. Rebased onto `main` now that WP-O2b (`makePromptTagInjector`, tolerant tag-tree resolver, prompts-storage/brands changes) is merged.

Jira: [LLMO-6282](https://jira.corp.adobe.com/browse/LLMO-6282)

### What changed (plan §WP-S2)
- **prompt-tags.js** — `DIMENSION.SOURCE`; appended to `DIMENSION_ROOT_NAMES`; joins `OPEN_DIMENSIONS` (no `CLOSED_DIMENSION_VALUES` entry). New `SERVER_OWNED_DIMENSIONS = [intent, origin, type, source]` + `isServerOwnedDimension()` carry the write-guard and resolve-or-create decisions; `isClosedDimension()` keeps only vocabulary validation. `canonicalizeSource()` (trim→lower→`_`→`-`, `null` on empty/`:`/over-`MAX_TAG_NAME_LEN`/root-name). `foldSourceValue()` (the same fold, always returns a string) for filter/sort input. Frozen exhaustive `SOURCE_LABEL` with a CI test that fails when a canonical value lacks a label. `PROXY_CREATE_SOURCE_VALUE` / `GENERATED_PROMPT_SOURCE_VALUE` constants.
- **tag-tree.js** — provisions **five** roots; distinctness guard keeps the producing-system `source` root separate from a mid-rename authorship `source` root (map's `source` key is `undefined` there — the WP-O6 gate). `ensureClosedValue` → `ensureServerOwnedValue`. `requireServerOwnedRootId` fails loud (502) rather than degrade an undefined `source` root into a stranded root-level create.
- **handlers/tags.js** — `source` create is resolve-or-create (no parentId, no enum).
- **handlers/prompts(.subworkspace).js** — create path injects `source/config` (constant, via `makePromptTagInjector`'s new `sourceValue`); update leaves source alone.
- **handlers/markets-subworkspace.js** — `generateAndAttachPrompts` stamps `source/semrush`.
- **prompts-storage.js** — `mapRowToPrompt` returns `canonicalizeSource(row.source)` (raw string on guard failure). `source` filter + sort key are on the **raw `source` column** (interim): stock PostgREST 400s on an inline `lower(replace(source,'_','-'))` expression, and there is no canonical column yet — WP-S4 adds a `source_canonical` generated column and repoints filter/sort at it. Incoming filter value folded via `foldSourceValue`. `updatePromptById` never patches source.
- **controllers/brands.js** — `createPromptsByBrand` ignores a body-supplied `source`.
- **docs** — `V2Prompt.source` read-only (never on `V2PromptInput`); prompts-v2 `source` query filter + sort enum; serenity create-tag `type` enum gains `source`.

Root assertions are **membership, never a 5-root count**.

### Verify (node 24)
- `npm run type-check` — pass
- `npm run lint` — pass (`docs:lint /configurations/` ambiguous-path warning is pre-existing)
- `npm test` (unit) — **15411 passing**, coverage thresholds met

### Not in unit scope (by design)
- `it-postgres` waits the released WP-S1 PE-client mock image.
- Canonical `source` filter/sort (folding drift spellings together) waits WP-S4's `source_canonical` generated column; the raw-column filter/sort here is the interim.

### DIFF
--- a/docs/openapi/schemas.yaml
+++ b/docs/openapi/schemas.yaml
@@ -6569,7 +6569,6 @@ V2Prompt:
         CANONICAL form (`lower(replace(source, '_', '-'))`), so `agentic_traffic`
         reads as `agentic-traffic`.
       example: "config"
-      default: "config"
     intent:
       type: [string, 'null']
       description: >-
--- a/docs/openapi/schemas.yaml
+++ b/docs/openapi/schemas.yaml
@@ -6562,7 +6562,12 @@ V2Prompt:
       default: human
     source:
       type: string
-      description: The source system or process that created the prompt
+      readOnly: true
+      description: >-
+        The producing system (source-dimension.md). READ-ONLY: server-owned, with
+        no write surface — a create/update body cannot set it. Returned in its
+        CANONICAL form (`lower(replace(source, '_', '-'))`), so `agentic_traffic`
+        reads as `agentic-traffic`.
       example: "config"
       default: "config"
     intent:
@@ -6665,10 +6670,6 @@ V2PromptInput:
         principal (e.g. the generation pipeline) may assert it, validated against
         the enum. It is never patched on update — it is fixed by the writer that
         created the row (origin-dimension.md §3).
-    source:
-      type: string
-      description: The source system or process that created the prompt
-      default: "config"
     intent:
       type: string
       description: >-
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -592,10 +592,18 @@ function BrandsController(ctx, log, env) {
       // queue consumer, re-ordered middleware) must not silently gain service
       // privilege — hence `!authType → user`, and a non-function `getType` resolves
       // to `undefined` (→ user) rather than throwing.
+      //
+      // `source` (the producing system) has NO write surface (source-dimension.md
+      // §1 item 6): a caller-supplied `source` is ignored, so a v2 create becomes
+      // the store's `config` default (item 5, gate §6.4/§6.6). It is dropped here
+      // rather than passed to upsertPrompts, whose `source: p.source || 'config'`
+      // write-side default stays load-bearing for the internal writers that DO set
+      // it. `updatePromptById` likewise never patches source (producer is fixed at
+      // creation).
       const { authInfo } = context.attributes ?? {};
       const authType = typeof authInfo?.getType === 'function' ? authInfo.getType() : undefined;
       const isUserPrincipal = !authType || authType === 'jwt' || authType === 'ims';
-      const derivedPrompts = prompts.map((p) => ({
+      const derivedPrompts = prompts.map(({ source: _, ...p }) => ({
         ...p,
         origin: deriveV2PromptOrigin(p?.origin, isUserPrincipal),
       }));
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -18,6 +18,16 @@ import { classifyIntents } from './intent-classifier.js';
 import { throwOnPgConstraintViolation } from './errors.js';
 import { assertPermittedSource } from './prompt-sources.js';
 import { INTENT_VALUES, normalizeIntent } from './intent.js';
+import { canonicalizeSource } from './serenity/prompt-tags.js';
+
+// The SQL expression that canonicalizes `prompts.source` (source-dimension.md
+// §3.1: `lower(replace(source, '_', '-'))`). Both the `source` list filter and
+// the `source` sort key run through it so the two spellings of a producer
+// (`citation_attempt` / `citation-attempt`) filter and order together. The plain
+// `idx_prompts_source` btree cannot serve it; WP-S4 adds the matching
+// `idx_prompts_source_canonical` expression index. Mirrors `canonicalizeSource`'s
+// transform in SQL — the two must not drift.
+const SOURCE_CANONICAL_SQL = "lower(replace(source, '_', '-'))";
 
 // Re-exported for backward compatibility — `normalizeIntent`/`INTENT_VALUES` now
 // live in `./intent.js` so the LLM intent classifier can reuse them without an
@@ -427,6 +437,11 @@ const SORT_COLUMN_MAP = {
   origin: 'origin',
   status: 'status',
   updatedAt: 'updated_at',
+  // `source` sorts on the CANONICAL form, not the raw column — the label lives in
+  // elmo and the server has never seen it, so ordering is on the slug and the two
+  // drift spellings order together (source-dimension.md §3.1). An object form
+  // marks it as a SQL expression rather than a bare column or a foreign-table sort.
+  source: { expression: SOURCE_CANONICAL_SQL },
 };
 
 function mapRowToPrompt(row) {
@@ -455,7 +470,13 @@ function mapRowToPrompt(row) {
     // fail-loud choice over a fabricated `human`; unlike `source`/`status`, whose
     // fallbacks are cosmetic, an origin fallback is a correctness hazard.
     origin: row.origin,
-    source: row.source || 'config',
+    // Second derivation boundary (source-dimension.md §3.1): the v2 read surface
+    // returns the CANONICAL slug, so elmo's badge — which keys on the API's value —
+    // resolves (`agentic_traffic` → `agentic-traffic`). A value that fails the guard
+    // (empty, `:`, over-long, root-shadowing) returns the RAW string rather than
+    // null: the grid must still show the operator what is stored. `?? 'config'` only
+    // guards a nullish column (in-memory/test rows); the DB column is NOT NULL.
+    source: canonicalizeSource(row.source) ?? row.source ?? 'config',
     intent: row.intent ?? null,
     createdAt: row.created_at,
     createdBy: row.created_by,
@@ -502,7 +523,8 @@ function mapRowToPrompt(row) {
  * topic name, category name
  * @param {string} [params.region] - Filter by region (array containment)
  * @param {string} [params.origin] - Filter by origin (ai, human)
- * @param {string} [params.source] - Filter by source (e.g. gsc, semrush, base_url, config)
+ * @param {string} [params.source] - Filter by source, matched on the CANONICAL
+ * form (source-dimension.md §3.1): `citation-attempt` also matches `citation_attempt`.
  * @param {string} [params.sort] - Sort column (topic, prompt, category, origin,
  * status, updatedAt)
  * @param {string} [params.order] - Sort direction (asc, desc). Default desc
@@ -606,7 +628,10 @@ export async function listPrompts({
     const sortCol = SORT_COLUMN_MAP[sort];
     if (sortCol) {
       const ascending = order === 'asc';
-      if (sortCol.includes('(')) {
+      if (typeof sortCol === 'object') {
+        // A SQL expression (e.g. the canonical `source` fold), ordered directly.
+        baseQuery = baseQuery.order(sortCol.expression, { ascending });
+      } else if (sortCol.includes('(')) {
         const [foreignTable, col] = sortCol.replace(')', '').split('(');
         baseQuery = baseQuery.order(col, { ascending, foreignTable });
       } else {
@@ -633,7 +658,12 @@ export async function listPrompts({
     }
 
     if (hasText(source)) {
-      baseQuery = baseQuery.eq('source', source);
+      // Filter on the CANONICAL form so `citation-attempt` matches rows stored as
+      // `citation_attempt` too (source-dimension.md §3.1). The incoming value is
+      // folded with the same transform as the column, and matched against the SQL
+      // expression rather than the raw column.
+      const wantedSource = String(source).trim().toLowerCase().replace(/_/g, '-');
+      baseQuery = baseQuery.eq(SOURCE_CANONICAL_SQL, wantedSource);
     }
 
     if (hasText(region)) {
@@ -1103,6 +1133,9 @@ export async function updatePromptById({
   }
 
   const patch = { updated_by: updatedBy };
+  // `source` is deliberately NOT patchable (source-dimension.md §1 item 6): a
+  // prompt's producer is fixed at creation, and the dimension has no write surface.
+  // A caller-supplied `updates.source` is ignored rather than written.
   if (updates.prompt !== undefined) {
     patch.text = updates.prompt;
   }
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -18,16 +18,7 @@ import { classifyIntents } from './intent-classifier.js';
 import { throwOnPgConstraintViolation } from './errors.js';
 import { assertPermittedSource } from './prompt-sources.js';
 import { INTENT_VALUES, normalizeIntent } from './intent.js';
-import { canonicalizeSource } from './serenity/prompt-tags.js';
-
-// The SQL expression that canonicalizes `prompts.source` (source-dimension.md
-// §3.1: `lower(replace(source, '_', '-'))`). Both the `source` list filter and
-// the `source` sort key run through it so the two spellings of a producer
-// (`citation_attempt` / `citation-attempt`) filter and order together. The plain
-// `idx_prompts_source` btree cannot serve it; WP-S4 adds the matching
-// `idx_prompts_source_canonical` expression index. Mirrors `canonicalizeSource`'s
-// transform in SQL — the two must not drift.
-const SOURCE_CANONICAL_SQL = "lower(replace(source, '_', '-'))";
+import { canonicalizeSource, foldSourceValue } from './serenity/prompt-tags.js';
 
 // Re-exported for backward compatibility — `normalizeIntent`/`INTENT_VALUES` now
 // live in `./intent.js` so the LLM intent classifier can reuse them without an
@@ -437,11 +428,14 @@ const SORT_COLUMN_MAP = {
   origin: 'origin',
   status: 'status',
   updatedAt: 'updated_at',
-  // `source` sorts on the CANONICAL form, not the raw column — the label lives in
-  // elmo and the server has never seen it, so ordering is on the slug and the two
-  // drift spellings order together (source-dimension.md §3.1). An object form
-  // marks it as a SQL expression rather than a bare column or a foreign-table sort.
-  source: { expression: SOURCE_CANONICAL_SQL },
+  // FIX (WP-S2 interim): sort on the RAW `source` column. Stock PostgREST (the
+  // backend) cannot order by an inline SQL expression like
+  // `lower(replace(source,'_','-'))` — it returns 400 — and there is no canonical
+  // column to name yet. WP-S4 adds a `source_canonical` generated column (+ its
+  // index) and ordering moves to that column name, so the two drift spellings order
+  // together; until then raw ordering only interleaves the `_`/`-` spellings of the
+  // same producer, which is cosmetic (source-dimension.md §3.1).
+  source: 'source',
 };
 
 function mapRowToPrompt(row) {
@@ -523,8 +517,9 @@ function mapRowToPrompt(row) {
  * topic name, category name
  * @param {string} [params.region] - Filter by region (array containment)
  * @param {string} [params.origin] - Filter by origin (ai, human)
- * @param {string} [params.source] - Filter by source, matched on the CANONICAL
- * form (source-dimension.md §3.1): `citation-attempt` also matches `citation_attempt`.
+ * @param {string} [params.source] - Filter by source. Matched on the raw column
+ * across both drift spellings (WP-S2 interim): `citation-attempt` also matches
+ * `citation_attempt`. WP-S4 folds this to a single `source_canonical` column match.
  * @param {string} [params.sort] - Sort column (topic, prompt, category, origin,
  * status, updatedAt)
  * @param {string} [params.order] - Sort direction (asc, desc). Default desc
@@ -628,10 +623,7 @@ export async function listPrompts({
     const sortCol = SORT_COLUMN_MAP[sort];
     if (sortCol) {
       const ascending = order === 'asc';
-      if (typeof sortCol === 'object') {
-        // A SQL expression (e.g. the canonical `source` fold), ordered directly.
-        baseQuery = baseQuery.order(sortCol.expression, { ascending });
-      } else if (sortCol.includes('(')) {
+      if (sortCol.includes('(')) {
         const [foreignTable, col] = sortCol.replace(')', '').split('(');
         baseQuery = baseQuery.order(col, { ascending, foreignTable });
       } else {
@@ -658,12 +650,18 @@ export async function listPrompts({
     }
 
     if (hasText(source)) {
-      // Filter on the CANONICAL form so `citation-attempt` matches rows stored as
-      // `citation_attempt` too (source-dimension.md §3.1). The incoming value is
-      // folded with the same transform as the column, and matched against the SQL
-      // expression rather than the raw column.
-      const wantedSource = String(source).trim().toLowerCase().replace(/_/g, '-');
-      baseQuery = baseQuery.eq(SOURCE_CANONICAL_SQL, wantedSource);
+      // FIX (WP-S2 interim): match on the RAW `source` column, not a canonical SQL
+      // expression — stock PostgREST 400s on `lower(replace(source,'_','-'))=eq.…`.
+      // To keep "filter by the value the grid shows" working across a producer's two
+      // drift spellings, fold the incoming value to canonical (shared
+      // `foldSourceValue` — the single definition of the transform) and match BOTH
+      // the hyphen and underscore forms, so `citation-attempt` also finds rows stored
+      // as `citation_attempt`. WP-S4's `source_canonical` generated column replaces
+      // this with a single canonical-column match (and adds case-folding, which raw
+      // matching drops).
+      const wanted = foldSourceValue(source);
+      const sourceVariants = [...new Set([wanted, wanted.replace(/-/g, '_')])];
+      baseQuery = baseQuery.in('source', sourceVariants);
     }
 
     if (hasText(region)) {
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -428,14 +428,11 @@ const SORT_COLUMN_MAP = {
   origin: 'origin',
   status: 'status',
   updatedAt: 'updated_at',
-  // FIX (WP-S2 interim): sort on the RAW `source` column. Stock PostgREST (the
-  // backend) cannot order by an inline SQL expression like
-  // `lower(replace(source,'_','-'))` — it returns 400 — and there is no canonical
-  // column to name yet. WP-S4 adds a `source_canonical` generated column (+ its
-  // index) and ordering moves to that column name, so the two drift spellings order
-  // together; until then raw ordering only interleaves the `_`/`-` spellings of the
-  // same producer, which is cosmetic (source-dimension.md §3.1).
-  source: 'source',
+  // Sort on the `source_canonical` generated column (WP-S4:
+  // `GENERATED ALWAYS AS (lower(replace(source,'_','-'))) STORED`, btree-indexed),
+  // so the two drift spellings of a producer order together and the label — which
+  // lives in elmo, never on the server — plays no part (source-dimension.md §3.1).
+  source: 'source_canonical',
 };
 
 function mapRowToPrompt(row) {
@@ -517,11 +514,11 @@ function mapRowToPrompt(row) {
  * topic name, category name
  * @param {string} [params.region] - Filter by region (array containment)
  * @param {string} [params.origin] - Filter by origin (ai, human)
- * @param {string} [params.source] - Filter by source. Matched on the raw column
- * across both drift spellings (WP-S2 interim): `citation-attempt` also matches
- * `citation_attempt`. WP-S4 folds this to a single `source_canonical` column match.
+ * @param {string} [params.source] - Filter by source, matched on the
+ * `source_canonical` generated column: `citation-attempt` also matches rows
+ * stored as `citation_attempt` (drift spellings fold together, case-insensitive).
  * @param {string} [params.sort] - Sort column (topic, prompt, category, origin,
- * status, updatedAt)
+ * source, status, updatedAt)
  * @param {string} [params.order] - Sort direction (asc, desc). Default desc
  * @param {number} [params.limit] - Page size (default 100, max 5000)
  * @param {number} [params.page] - Page number, 1-based (default 1)
@@ -650,20 +647,13 @@ export async function listPrompts({
     }
 
     if (hasText(source)) {
-      // FIX (WP-S2 interim): match on the RAW `source` column, not a canonical SQL
-      // expression — stock PostgREST 400s on `lower(replace(source,'_','-'))=eq.…`.
-      // To keep "filter by the value the grid shows" working across a producer's two
-      // drift spellings, fold the incoming value to canonical (shared
-      // `foldSourceValue` — the single definition of the transform) and match BOTH
-      // the hyphen and underscore forms, so `citation-attempt` also finds rows stored
-      // as `citation_attempt`. WP-S4's `source_canonical` generated column replaces
-      // this with a single canonical-column match (and adds case-folding, which raw
-      // matching drops).
-      const wanted = foldSourceValue(source);
-      // A value with no `_`/`-` (e.g. `gsc`) yields two identical variants that the
-      // Set collapses to one — a single-element `.in` is intentional, not a bug.
-      const sourceVariants = [...new Set([wanted, wanted.replace(/-/g, '_')])];
-      baseQuery = baseQuery.in('source', sourceVariants);
+      // Match on the `source_canonical` generated column (WP-S4) — the DB
+      // canonicalizes `lower(replace(source,'_','-'))` at write time, so a single
+      // equality on the folded incoming value catches every drift spelling AND is
+      // case-insensitive (`citation-attempt` finds rows stored `citation_attempt`
+      // or `Citation_Attempt`). Fold the query-param value with the same transform
+      // (`foldSourceValue` — the single definition) so the two sides align.
+      baseQuery = baseQuery.eq('source_canonical', foldSourceValue(source));
     }
 
     if (hasText(region)) {
--- a/src/support/prompts-storage.js
+++ b/src/support/prompts-storage.js
@@ -660,6 +660,8 @@ export async function listPrompts({
       // this with a single canonical-column match (and adds case-folding, which raw
       // matching drops).
       const wanted = foldSourceValue(source);
+      // A value with no `_`/`-` (e.g. `gsc`) yields two identical variants that the
+      // Set collapses to one — a single-element `.in` is intentional, not a bug.
       const sourceVariants = [...new Set([wanted, wanted.replace(/-/g, '_')])];
       baseQuery = baseQuery.in('source', sourceVariants);
     }
--- a/src/support/serenity/handlers/markets-subworkspace.js
+++ b/src/support/serenity/handlers/markets-subworkspace.js
@@ -43,8 +43,10 @@ import { withResourceLock } from '../resource-lock.js';
 import {
   modelChangeUnits, releaseAiSurplus, PROJECT_BLOCK, PROMPT_BLOCK,
 } from '../resource-manager.js';
-import { DIMENSION, STANDARD_PROMPT_TAG_VALUES, INTENT_VALUE } from '../prompt-tags.js';
-import { provisionDimensionTree } from '../tag-tree.js';
+import {
+  DIMENSION, STANDARD_PROMPT_TAG_VALUES, INTENT_VALUE, GENERATED_PROMPT_SOURCE_VALUE,
+} from '../prompt-tags.js';
+import { provisionDimensionTree, ensureServerOwnedValue } from '../tag-tree.js';
 import { classifyBrandedTag, needlesFromNames } from '../branded-classifier.js';
 import { classifyPromptIntents, AI_GEN_CLASSIFY_MAX, computeWriteDeadline } from '../intent-classification.js';
 import { collectBrandUrlEntries, attachBrandUrlsToProject, primaryDomainSet } from '../brand-urls.js';
@@ -280,11 +282,11 @@ function validateCreateBody(body) {
  * (transport.getBrandTopics) and attaches them to the project. Keeps the top
  * `topicCap` topics by search volume (0 = keep all) and tags every prompt with
  * the standard closed-dimension values ({@link STANDARD_PROMPT_TAG_VALUES}, minus
- * its seeded `intent` default), plus a branded / non-branded `type` value derived
- * from `brandNames` (brand name + aliases) and a per-prompt server-classified
- * `intent` value (serenity-docs#32, replacing the seeded `Informational`
- * default). Returns the topic/prompt counts. A generation that yields nothing is
- * a clean no-op (no upstream write).
+ * its seeded `intent` default), the producing `source/semrush` value, plus a
+ * branded / non-branded `type` value derived from `brandNames` (brand name +
+ * aliases) and a per-prompt server-classified `intent` value (serenity-docs#32,
+ * replacing the seeded `Informational` default). Returns the topic/prompt counts.
+ * A generation that yields nothing is a clean no-op (no upstream write).
  *
  * The generated topic name is NOT attached. Under the dimension-root model a
  * topic is a sub-category — a depth-3 descendant of a customer category — and
@@ -375,6 +377,19 @@ async function generateAndAttachPrompts(transport, workspaceId, projectId, {
   const standardIdsNonIntent = STANDARD_PROMPT_TAG_VALUES
     .filter(({ dimension }) => dimension !== DIMENSION.INTENT)
     .map(({ dimension, name }) => /** @type {string} */ (values.get(dimension)?.get(name)));
+  // Stamp the producing system. This generator builds its prompts from Semrush's
+  // own `getBrandTopics`, so every generated prompt is `source/semrush` — the
+  // persisted SR-AI-Visibility key, a constant at THIS write site, NOT `config`
+  // (source-dimension.md §1 item 2). `source` is open, so the value is resolved-or-
+  // created on demand rather than pre-provisioned in `provisioned.values`.
+  const { id: sourceId } = await ensureServerOwnedValue(
+    transport,
+    workspaceId,
+    projectId,
+    DIMENSION.SOURCE,
+    GENERATED_PROMPT_SOURCE_VALUE,
+    log,
+  );
   const typeValues = /** @type {Map<string, string>} */ (values.get(DIMENSION.TYPE));
   const intentValues = /** @type {Map<string, string>} */ (values.get(DIMENSION.INTENT));
 
@@ -424,7 +439,9 @@ async function generateAndAttachPrompts(transport, workspaceId, projectId, {
       const intentId = /** @type {string} */ (intentValues.get(intentValue));
       byTagSet.set(key, {
         items: [text],
-        tagIds: [...standardIdsNonIntent, intentId, typeId],
+        // `sourceId` (source/semrush) is constant for every generated prompt, so
+        // it rides in every bucket alongside the per-(type, intent) ids.
+        tagIds: [...standardIdsNonIntent, intentId, sourceId, typeId],
       });
     }
   }
@@ -444,7 +461,7 @@ async function generateAndAttachPrompts(transport, workspaceId, projectId, {
     // LLMO-6190 follow-up: the metered write can still 405 as a disguised metered-quota rejection
     // despite the `ensure` above (live-verified ~9s gateway write-enforcement lag after a JIT
     // top-up) — route through `headroom.retryOnQuota` (no-op passthrough when the flag is OFF).
-    // `tagIds` is precomputed per (type, intent) bucket above (standard + intent + type).
+    // `tagIds` is precomputed per (type, intent) bucket above (standard + intent + source + type).
     // eslint-disable-next-line no-await-in-loop
     await headroom.retryOnQuota(
       () => transport.createPromptsWithMetadata(
--- a/src/support/serenity/handlers/prompts.js
+++ b/src/support/serenity/handlers/prompts.js
@@ -20,7 +20,7 @@ import { ERROR_CODES, isMeteredQuota, isUpstreamGone } from '../errors.js';
 import { alertQuotaRejection, alertRollbackFailure } from '../quota-alerts.js';
 import { normalizeGeoTargetId, normalizeLanguageCode, isValidTagIdFormat } from '../validation.js';
 import { invalidateTagCacheForProject } from './markets.js';
-import { resolveTypeValueInjection, resolveIntentValueInjection, resolveClosedValueInjection } from '../tag-tree.js';
+import { resolveTypeValueInjection, resolveIntentValueInjection, resolveServerOwnedValueInjection } from '../tag-tree.js';
 import {
   DIMENSION, ORIGIN_VALUE, INTENT_VALUE, PROXY_CREATE_SOURCE_VALUE,
 } from '../prompt-tags.js';
@@ -721,7 +721,7 @@ export async function createOnePrompt(transport, semrushWorkspaceId, projectId, 
  *   - on UPDATE (`sourceValue` unset) the injector leaves source ALONE — a prompt's
  *     producer is fixed at creation.
  *
- * Resolution ({@link resolveTypeValueInjection} / {@link resolveClosedValueInjection},
+ * Resolution ({@link resolveTypeValueInjection} / {@link resolveServerOwnedValueInjection},
  * two tag-tree reads per distinct value per project — the root level plus the
  * root's children) is memoized for the request, so a bulk create fans out over
  * the distinct computed values rather than over the items. The origin and source
@@ -784,7 +784,7 @@ export function makePromptTagInjector(
     if (originValue) {
       let pending = originCache.get(projectId);
       if (!pending) {
-        pending = resolveClosedValueInjection(
+        pending = resolveServerOwnedValueInjection(
           transport,
           semrushWorkspaceId,
           projectId,
@@ -804,7 +804,7 @@ export function makePromptTagInjector(
     if (sourceValue) {
       let pending = sourceCache.get(projectId);
       if (!pending) {
-        pending = resolveClosedValueInjection(
+        pending = resolveServerOwnedValueInjection(
           transport,
           semrushWorkspaceId,
           projectId,
--- a/src/support/serenity/handlers/prompts.js
+++ b/src/support/serenity/handlers/prompts.js
@@ -21,7 +21,9 @@ import { alertQuotaRejection, alertRollbackFailure } from '../quota-alerts.js';
 import { normalizeGeoTargetId, normalizeLanguageCode, isValidTagIdFormat } from '../validation.js';
 import { invalidateTagCacheForProject } from './markets.js';
 import { resolveTypeValueInjection, resolveIntentValueInjection, resolveClosedValueInjection } from '../tag-tree.js';
-import { DIMENSION, ORIGIN_VALUE, INTENT_VALUE } from '../prompt-tags.js';
+import {
+  DIMENSION, ORIGIN_VALUE, INTENT_VALUE, PROXY_CREATE_SOURCE_VALUE,
+} from '../prompt-tags.js';
 import { classifyPromptIntents } from '../intent-classification.js';
 
 /** @typedef {import('../rest-transport.js').SerenityTransport} SerenityTransport */
@@ -709,11 +711,21 @@ export async function createOnePrompt(transport, semrushWorkspaceId, projectId, 
  *     stripping without injecting would leave the prompt invisible to the
  *     dimension's filter — both are illegal, so the update path does neither.
  *
+ * **`source`** — the PRODUCING SYSTEM (source-dimension.md), and like `origin` it
+ * is a fact about CREATION, not a classification, so it carries the same asymmetry:
+ *   - on CREATE (`sourceValue` set — the constant `config` for this proxy dialog,
+ *     the value the same prompt gets in Postgres on the v2 path), any caller-supplied
+ *     tag id beneath the `source` root is stripped (by RESOLVED ID, never by name — a
+ *     customer category may legitimately be called `gsc`) and the derived value
+ *     injected. The dimension has no client write surface;
+ *   - on UPDATE (`sourceValue` unset) the injector leaves source ALONE — a prompt's
+ *     producer is fixed at creation.
+ *
  * Resolution ({@link resolveTypeValueInjection} / {@link resolveClosedValueInjection},
  * two tag-tree reads per distinct value per project — the root level plus the
  * root's children) is memoized for the request, so a bulk create fans out over
- * the distinct computed values rather than over the items. The origin value is
- * constant per request, so its resolution is memoized per project.
+ * the distinct computed values rather than over the items. The origin and source
+ * values are constant per request, so each resolution is memoized per project.
  *
  * Resolution resolves or throws, so a server tag is always attached; it is never
  * dropped, and a resolution failure aborts the write (which is free — the upstream
@@ -724,8 +736,9 @@ export async function createOnePrompt(transport, semrushWorkspaceId, projectId, 
  * @param {string} semrushWorkspaceId
  * @param {((text: string, geoTargetId: number) => string) | undefined} classifyPromptType
  * @param {object} [log]
- * @param {{ originValue?: string }} [options] - `originValue` is the derived
- *   `origin` to inject on CREATE; omit it on UPDATE so origin is left untouched.
+ * @param {{ originValue?: string, sourceValue?: string }} [options] - `originValue`
+ *   / `sourceValue` are the derived `origin` / `source` to inject on CREATE; omit
+ *   each on UPDATE so that dimension is left untouched.
  * @returns {(projectId: string, input: { text: string, geoTargetId: number,
  *   tagIds: string[] }) =>
  *   Promise<{ text: string, geoTargetId: number, tagIds: string[] }>}
@@ -737,11 +750,13 @@ export function makePromptTagInjector(
   log,
   options = {},
 ) {
-  const { originValue } = options;
+  const { originValue, sourceValue } = options;
   /** @type {Map<string, Promise<{ computedId: string, typeTagIds: string[] }>>} */
   const typeCache = new Map();
   /** @type {Map<string, Promise<{ computedId: string, valueTagIds: string[] }>>} */
   const originCache = new Map();
+  /** @type {Map<string, Promise<{ computedId: string, valueTagIds: string[] }>>} */
+  const sourceCache = new Map();
   return async function injectComputedTags(projectId, input) {
     let { tagIds } = input;
 
@@ -778,6 +793,26 @@ export function makePromptTagInjector(
           log,
         );
         originCache.set(projectId, pending);
+      }
+      const { computedId, valueTagIds } = await pending;
+      tagIds = [...tagIds.filter((id) => !valueTagIds.includes(id)), computedId];
+    }
+
+    // source — CREATE only, same asymmetry as origin. `sourceValue` unset means
+    // UPDATE: leave the producer alone (fixed at creation). Stripped by resolved id
+    // beneath the `source` root, never by name (a category may be `gsc`).
+    if (sourceValue) {
+      let pending = sourceCache.get(projectId);
+      if (!pending) {
+        pending = resolveClosedValueInjection(
+          transport,
+          semrushWorkspaceId,
+          projectId,
+          DIMENSION.SOURCE,
+          sourceValue,
+          log,
+        );
+        sourceCache.set(projectId, pending);
       }
       const { computedId, valueTagIds } = await pending;
       tagIds = [...tagIds.filter((id) => !valueTagIds.includes(id)), computedId];
@@ -1008,12 +1043,14 @@ export async function handleCreatePrompts(
   // (origin-dimension.md §3). Any caller-supplied origin tag id is stripped and
   // this value injected; on the twin AI-generation path a service producer stamps
   // `ai` via STANDARD_PROMPT_TAG_VALUES instead (that path does not run here).
+  // The producing `source` is the constant `config` — this human create dialog is
+  // what the same prompt gets in Postgres on the v2 path (source-dimension.md §1).
   const injectComputedTags = makePromptTagInjector(
     transport,
     semrushWorkspaceId,
     classifyPromptType,
     log,
-    { originValue: ORIGIN_VALUE.HUMAN },
+    { originValue: ORIGIN_VALUE.HUMAN, sourceValue: PROXY_CREATE_SOURCE_VALUE },
   );
   // Unified layer (serenity-docs#32): batch-classify every distinct text ONCE
   // under the shared request deadline, then thread the resolved map into each
--- a/src/support/serenity/handlers/tags.js
+++ b/src/support/serenity/handlers/tags.js
@@ -21,11 +21,12 @@ import {
 } from '../validation.js';
 import { resolveProject } from '../subworkspace-projects.js';
 import {
-  ALL_DIMENSIONS, CLOSED_DIMENSIONS,
-  isClosedDimension, closedValuesOf, isDimensionRootName,
+  ALL_DIMENSIONS, CLOSED_DIMENSIONS, SERVER_OWNED_DIMENSIONS,
+  isClosedDimension, isServerOwnedDimension, closedValuesOf, isDimensionRootName,
+  MAX_TAG_NAME_LEN,
 } from '../prompt-tags.js';
 import {
-  ensureClosedValue,
+  ensureServerOwnedValue,
   ensureDimensionRoots,
   findTagsInTree,
   assertParentPlacement,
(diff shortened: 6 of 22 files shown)


aliciadriani (commented):
Reviewed WP-S2 (LLMO-6282): server-owned `source` dimension, `canonicalizeSource`, and the `SERVER_OWNED_DIMENSIONS` refactor. Base `claude/wp-o2b-derivation` (stacked, non-main) → MysticatBot N/A. DO-NOT-MERGE banner + WP-O6 gate expected. `it-postgres` red is the SQL sort/filter that pairs with WP-S4's index (it/-gated, unit-only now) — not a finding per the lens.

Audited each lens item; the design is strong. One defense-in-depth Should-Fix.

**Headline refactor — CONFIRMED correct.** I audited each former `isClosedDimension()` call site for which question it was actually asking:
- `parseCreateTagBody`: the **parentId write-guard** now keys on `isServerOwnedDimension` (line 240) — so `source` correctly forbids a `parentId` (its values hang off the root) even though it's open; the **name enum-check** stays on `isClosed` (`closedValuesOf`, line 222) — so `source` is correctly NOT enum-checked (open vocabulary). That's exactly the two-questions split the WP intends.
- `handleCreateTag` / `handleCreateTagSubworkspace`: the resolve-or-create branch flips `isClosed` → `isServerOwned`, routing `source` through `ensureServerOwnedValue` (the renamed, vocabulary-agnostic `ensureClosedValue`). Correct — resolve-or-create is the server-owned semantic, and the enum enforcement was already lifted to the caller.
`isClosedDimension` is left with its one honest job (vocabulary), and no other call site remained mis-asking.

**`canonicalizeSource` — CONFIRMED.** `value.trim().toLowerCase().replace(/_/g,'-')`; returns `null` (never a default) on non-string / empty-after-trim / `:` / `>MAX_TAG_NAME_LEN` / `isDimensionRootName`. This is the reference twin the CLI Python side is being aligned to (the CLI's missing `.strip()` was my WP-S4 finding). `SOURCE_LABEL` is frozen and its exhaustiveness is enforced by a CI test (`prompt-tags.test.js`: `Object.keys(SOURCE_LABEL).sort()` deep-equals `SOURCE_VALUES`, fails on any label-less value) with no `?? slug` pass-through default — matches §7.

**Distinctness guard — CONFIRMED.** `ensureDimensionRoots` classifies the single physical `source` root once: if it's authorship (children ⊆ {ai,human}, no `origin`) the map's `source` key is `undefined` and the producer root is neither created nor returned; it resolves only on a fresh project (created outright) or a post-rename project (authorship moved to `origin`). The producing-system root is never conflated with a mid-rename authorship root — which is exactly why the PR is WP-O6-gated.

**Item 13 — CONFIRMED.** `V2Prompt.source` is `readOnly: true` (canonical form); `source` is REMOVED from `V2PromptInput`; the v2 list gains a `source` query filter + a `source` sort enum (both server-side canonical-fold, `lower(replace(source,'_','-'))`); create-tag `type` enum gains `source`. Bodies can't set it.

**Origin/source write-side — CONFIRMED.** `mapRowToPrompt` → `canonicalizeSource(row.source) ?? row.source ?? 'config'` (2nd boundary; RAW string on guard failure, not null — right call for the grid); `deriveV2PromptOrigin` from the principal, body dropped in `brands.js` (`{ source: _, ...p }`); `upsertPrompts` preserves stored `source` AND `origin` on update; `updatePromptById` patches neither; write-side `|| 'config'` default kept for internal writers.

---

## SHOULD-FIX (defense-in-depth; CONFIRMED mechanism, reachability gated by WP-O6) — the mid-rename `source` path can mint a ROOT-LEVEL `config` tag, silently, if the gate is ever bypassed

The distinctness guard in `ensureDimensionRoots` correctly returns `source: undefined` on a mid-rename project, and `rootIdOf`'s own docstring warns "the `source` root can be `undefined`… callers that need it read `roots.get(DIMENSION.SOURCE)` directly." But the source-injection path does NOT heed that: `makePromptTagInjector` (called on EVERY `handleCreatePrompts` create with `sourceValue: 'config'`) → `resolveClosedValueInjection(transport, ws, pid, DIMENSION.SOURCE, 'config')` → `rootIdOf(roots, SOURCE)`, and `rootIdOf` is just `roots.get(dimension)` with **no guard**, so on a mid-rename project it returns `undefined`. That `undefined` flows into `ensureChildren(parentId=undefined, ['config'])`, whose create is `createProjectTags(missing, parentId ? { parentId } : {})` — with `parentId` undefined this becomes `{}`, a **ROOT-LEVEL create**: it mints a bare `config` root tag and tags the prompt with it. Not a fail-loud abort — a silent stranded-root corruption, precisely what the dimension-root model forbids.

This is fully mitigated by the EXTERNAL WP-O6 deployment gate (post-WP-O6 every project has the `source` name freed, so `roots.get(SOURCE)` is always defined) plus the DO-NOT-MERGE banner — so it is not a live bug on the intended timeline, and I'm not blocking on it. But there is **no in-code guard**: nothing stops the source-injection path from passing an `undefined` root id to `ensureChildren` if this ever ran on a not-yet-renamed project (a premature deploy, or a project that slips past the gate). Recommend a cheap guard that makes the design's stated invariant ("MUST NOT be provisioned before WP-O6") fail LOUD rather than corrupt: in the `source` injection branch (or `resolveClosedValueInjection` when `dimension === DIMENSION.SOURCE`, or `ensureChildren` when `parentId` is `undefined` vs `''`), throw a clear "source dimension not provisioned (WP-O6-gated)" error instead of a root-level create. I couldn't find a test exercising source-injection on a mid-rename project (the passing suite covers fresh/post-rename, where the root resolves), so this path is currently unguarded and untested. Routing to the implementer to decide guard-vs-document.

No Must-Fix. `/review-pr` otherwise clean; the above is a hardening request against a gate-bypass, not a mainline defect. Bot N/A (non-main base).

src/support/serenity/prompt-tags.js
```diff
@@ -122,15 +145,109 @@ export const CLOSED_DIMENSIONS = Object.freeze([
...
+ *
+ * Two decisions route through this list, not through `isClosedDimension`:
+ *  - the create-tag WRITE GUARD (a client may address it but never author a value
+ *    outside the server's control — the closed dimensions additionally enum-check,
+ *    `source` does not), and
+ *  - CREATE SEMANTICS: a server-owned value is resolve-or-create, because no human
+ *    is in a dialog to resolve it first.
+ * `isClosedDimension` keeps its one honest job: vocabulary validation.
+ */
+export const SERVER_OWNED_DIMENSIONS = Object.freeze([
```
aliciadriani, line 157: `SERVER_OWNED_DIMENSIONS` = [intent, origin, type, source] is the right axis, cleanly separated from open/closed. The docstring nails why: two independent properties (vocabulary open/closed vs writer customer/server), and `source` is the open-AND-server-owned cell. Verified both decisions route through this (write-guard + create-semantics) while `isClosedDimension` keeps only vocabulary validation.

src/support/serenity/handlers/tags.js
```diff
@@ -227,15 +237,15 @@ function parseCreateTagBody(body) {
     );
   }
   const parentId = parseParentId(body?.parentId);
-  if (isClosed && parentId !== undefined) {
+  if (isServerOwned && parentId !== undefined) {
```
aliciadriani, line 242: Call-site audit confirmed correct: the parentId write-guard keys on `isServerOwned` here (so `source` forbids a parentId — its values hang off the root), while the name enum-check a few lines up (line 222, `isClosed && !closedValuesOf(type).includes(rawName)`) stays on `isClosed` (so `source` is NOT enum-checked — open vocabulary). That is exactly the two-questions split: `source` = server-owned (guarded) + open (not enum-checked).

src/support/serenity/prompt-tags.js
```diff
@@ -175,6 +294,51 @@ export function isClosedDimension(dimension) {
...
+ *
+ * `null` means "do not tag this prompt", NEVER "substitute a default": a caller
+ * writes the prompt regardless and logs the offending value. This is the single
+ * place the rule lives in this repo; it is applied at both derivation boundaries
+ * (the tag write and the v2 read surface — `mapRowToPrompt`).
+ *
+ * @param {unknown} value - a raw `prompts.source` value.
+ * @returns {string | null} the canonical slug, or `null` when it must not be tagged.
+ */
+export function canonicalizeSource(value) {
```
aliciadriani, line 327: `canonicalizeSource` confirmed as the reference twin: `value.trim().toLowerCase().replace(/_/g,'-')`, `null` (never a default) on non-string / empty / `:` / `>MAX_TAG_NAME_LEN` / `isDimensionRootName`. The `.trim()` here is the semantics the migration CLI's Python twin was missing (my WP-S4 finding, being fixed there). Good single-source-of-truth with `MAX_TAG_NAME_LEN`.

src/support/serenity/prompt-tags.js
```diff
@@ -122,15 +145,109 @@ export const CLOSED_DIMENSIONS = Object.freeze([
...
+ * mechanism by which an internal slug reaches a customer silently (source-dimension.md
+ * §7), and §3.2's product sign-off protects only the values that exist today.
+ *
+ * The label question is OPEN (source-dimension.md §3.2): until product signs off,
+ * the labels are the canonical machine-name slugs, verbatim. This map is the SINGLE
+ * place a display label would land, so whichever way that call goes it changes here
+ * and nowhere else. (elmo ships its own display labels behind `SOURCE_BADGE_CONFIG`;
+ * WP-S3.)
+ */
+export const SOURCE_LABEL = Object.freeze(
```
aliciadriani, line 232: `SOURCE_LABEL` frozen + exhaustive with NO `?? slug` pass-through default, and the CI test (`prompt-tags.test.js` line ~166) fails the moment a `SOURCE_VALUES` entry lacks a label. This is the right shape for §7 — an internal slug can't silently reach a customer, and the label decision lands in exactly one place.

src/support/serenity/tag-tree.js
```diff
@@ -252,40 +252,62 @@ async function childrenAreAuthorship(transport, semrushWorkspaceId, projectId, r
...
-    const legacyId = existing.get(LEGACY_AUTHORSHIP_ROOT_NAME);
-    if (legacyId
-      && await childrenAreAuthorship(transport, semrushWorkspaceId, projectId, legacyId, log)) {
-      log?.info?.('ensureDimensionRoots: adopting the legacy `source` authorship root in place', {
-        semrushWorkspaceId, projectId, rootId: legacyId,
-      });
-      authorshipId = legacyId;
-    }
+  let legacySourceIsAuthorship = false;
+  const physicalSourceIsAuthorship = !authorshipId && physicalSourceId
```
aliciadriani: Distinctness guard is correct: the single physical `source` root is classified once (`physicalSourceIsAuthorship`), and the producer `source` root is created/returned only when the name is free (fresh or post-rename), else the map's `source` key is `undefined`. Never conflates the producing-system root with a mid-rename authorship root — the WP-O6-gate rationale, verified.

src/support/serenity/tag-tree.js
```diff
@@ -311,11 +349,12 @@ export async function ensureDimensionRoots(transport, semrushWorkspaceId, projec
...
- * asked for — all four roots. The assertion records that invariant for the type
- * checker instead of re-testing it at runtime.
+ * asked for. Used only for roots that are always resolved (the closed dimensions
+ * and `category`); the `source` root can be `undefined` on a mid-rename project,
+ * so callers that need it read `roots.get(DIMENSION.SOURCE)` directly instead.
  *
- * @param {Map<string, string>} roots - the resolved root name → id map.
- * @param {string} dimension - one of the four dimension root names.
+ * @param {Map<string, string | undefined>} roots - the resolved root name → id map.
+ * @param {string} dimension - one of the always-resolved dimension root names.
```
aliciadriani, line 297: SHOULD-FIX (defense-in-depth): this docstring correctly warns the `source` root can be `undefined` and that callers needing it should read `roots.get(DIMENSION.SOURCE)` directly — but `resolveClosedValueInjection` (line 652) does NOT: it calls `rootIdOf(roots, DIMENSION.SOURCE)` (which is a bare `roots.get`, no guard), so on a mid-rename project it hands `undefined` to `ensureChildren`, whose `createProjectTags(missing, parentId ? {parentId} : {})` then creates the value at ROOT level (a stranded `config` root). Gated externally by WP-O6, but unguarded and untested in code. Consider throwing a clear "source not provisioned (WP-O6-gated)" when `dimension === DIMENSION.SOURCE` and the root id is undefined, so a gate-bypass fails loud instead of corrupting.

src/support/serenity/handlers/prompts.js
```diff
@@ -366,60 +367,142 @@ export async function createOnePrompt(transport, semrushWorkspaceId, projectId,
...
+      tagIds = [...tagIds.filter((id) => !valueTagIds.includes(id)), computedId];
     }
-    const { computedId, typeTagIds } = await pending;
-    const stripped = input.tagIds.filter((id) => !typeTagIds.includes(id));
-    return { ...input, tagIds: [...stripped, computedId] };
+
+    // source — CREATE only, same asymmetry as origin. `sourceValue` unset means
+    // UPDATE: leave the producer alone (fixed at creation). Stripped by resolved id
+    // beneath the `source` root, never by name (a category may be `gsc`).
+    if (sourceValue) {
```
aliciadriani, line 804: The `source` injection branch (CREATE-only, memoized, stripped by resolved root id not by name) is correct on a fresh/post-rename project. It's the entry point of the mid-rename concern flagged on tag-tree.js:357 — this unconditionally calls `resolveClosedValueInjection(DIMENSION.SOURCE, sourceValue)` whenever `sourceValue` is set (always, on `handleCreatePrompts`), with no check that `roots.get(SOURCE)` resolved. The origin branch above is safe (origin resolves tolerantly to the legacy root); only `source` can be undefined mid-rename.

src/controllers/brands.js
```diff
@@ -560,10 +561,32 @@ function BrandsController(ctx, log, env) {
...
+      const authType = context.attributes?.authInfo?.getType?.();
+      const isUserPrincipal = authType === 'jwt' || authType === 'ims';
+      // `source` (the producing system) has NO write surface (source-dimension.md
+      // §1 item 6): a caller-supplied `source` is ignored, so a v2 create becomes
+      // the store's `config` default (item 5, gate §6.4/§6.6). It is dropped here
+      // rather than passed to upsertPrompts, whose `source: p.source || 'config'`
+      // write-side default stays load-bearing for the internal writers that DO set
+      // it. `updatePromptById` likewise never patches source (producer is fixed at
+      // creation).
+      const derivedPrompts = prompts.map(({ source: _, ...p }) => ({
```
aliciadriani, line 606: `{ source: _, ...p }` correctly drops any body `source` (no write surface) and `deriveV2PromptOrigin` derives origin from the principal (user→human, service→believed). Confirmed the drop makes a v2 create fall to the store's `config` default, and the write-side `|| 'config'` in upsertPrompts stays load-bearing for internal writers that DO set source. Clean item-5/6 handling.

src/support/prompts-storage.js
```diff
@@ -406,8 +460,18 @@ function mapRowToPrompt(row) {
...
+    // §2.3), and a fallback would silently mislabel a model-written prompt as
+    // human-authored were a NULL ever inserted.
+    origin: row.origin,
+    // Second derivation boundary (source-dimension.md §3.1): the v2 read surface
+    // returns the CANONICAL slug, so elmo's badge — which keys on the API's value —
+    // resolves (`agentic_traffic` → `agentic-traffic`). A value that fails the guard
+    // (empty, `:`, over-long, root-shadowing) returns the RAW string rather than
+    // null: the grid must still show the operator what is stored. `?? 'config'` only
+    // guards a nullish column (in-memory/test rows); the DB column is NOT NULL.
+    source: canonicalizeSource(row.source) ?? row.source ?? 'config',
```
aliciadriani, line 470: Second derivation boundary correct: `canonicalizeSource(row.source) ?? row.source ?? 'config'` returns the canonical slug, falls back to the RAW string on a guard failure (so the grid still shows what's stored — deliberately not null here, unlike the tag-write boundary), and `?? 'config'` only guards a nullish column. `origin: row.origin` with no `|| 'human'` fallback is also right (NOT NULL in prod; a fallback would mislabel).

aliciadriani (commented):
Re-check of WP-S2 at `8758de5a` — my Should-Fix (unguarded mid-rename `source` root id degrading into a root-level create) is FIXED. **S2 is now review-clean.** Base non-main → bot N/A.

The fix adds `requireServerOwnedRootId(roots, dimension)` — throws `ErrorWithStatusCode(..., 502)` with a clear message (`"<dim> dimension root not provisioned (source-dimension.md is WP-O6-gated); refusing to create a stranded root-level tag"`) when the resolved server-owned root id is falsy — and swaps it in for the bare `rootIdOf` at BOTH sites I named: `resolveClosedValueInjection` (the create-path injector) and `ensureServerOwnedValue` (create-tag + the semrush-stamp generator). Confirming the two requested points:

**(a) No-op for always-provisioned dimensions — CONFIRMED.** `requireServerOwnedRootId` returns the id unchanged when it's truthy, throwing only on a falsy id. `intent` / `origin` / `type` are always resolved by `ensureDimensionRoots` (created when missing; `origin` resolves tolerantly to a real physical root), so their root ids are never falsy and the guard is a pass-through — behaviour for those dimensions is identical to before. Only `source` on a mid-rename project can be `undefined` (the deliberate distinctness-guard gap), so only that case changes.

**(b) Mid-rename `source` now THROWS with zero root-level create — CONFIRMED.** The new tests pin it exactly:
- `tag-tree.test.js`: on a mid-rename tree (legacy `source` root with ai/human, no `origin`), `ensureDimensionRoots` leaves `roots.get('source')` undefined; `ensureServerOwnedValue(source, 'config')` and `resolveClosedValueInjection(source, 'config')` each reject with `status===502`, message `/source dimension root not provisioned/`, and `createProjectTags` is NOT called — so no stranded root-level `config` tag is minted.
- `prompts.test.js`: the handler-level path — `handleCreatePrompts` on a mid-rename project — puts the item in `failed` with `status===502`, calls neither `createPromptsByIds` nor `createProjectTags`. Fail-loud, nothing written, no corruption.

The guard turns the design's stated invariant ("source MUST NOT be provisioned before WP-O6") from a silent tree-corruption-on-gate-bypass into a visible 502, which is exactly what I asked for. `ensureDimensionRoots`'s distinctness guard (returning `source: undefined` mid-rename) is unchanged and still correct; the fix is purely the downstream consumer hardening.

I could not run the 15305-unit suite locally (no Python/JS worktree with deps here), so I relied on the reported pass count for the empirical confirmation — the change is small, surgical, and the added tests cover the exact scenarios. All other WP-S2 lens items remain as CONFIRMED in my prior review (SERVER_OWNED_DIMENSIONS split, canonicalizeSource trim + SOURCE_LABEL exhaustiveness, distinctness guard, item 13, write-side origin/source). Net: `/review-pr` clean. Bot N/A (non-main base).

aliciadriani (commented):
Re-check at `a9539eec` (rebased onto O2b's advanced head, now MERGEABLE, 15346 unit pass). Verified the single rebase conflict resolution in `brands.js` `createPromptsByBrand` and the auto-merged O2b-touched files. **WP-S2 remains review-clean.** Base non-main → bot N/A.

**`brands.js` conflict — BOTH changes intact, no regression:**
- **O2b auth-hardening (MysticatBot):** `const { authInfo } = context.attributes ?? {}` → `authType = typeof authInfo?.getType === 'function' ? authInfo.getType() : undefined` → `isUserPrincipal = !authType || authType === 'jwt' || authType === 'ims'`. Fail-safe to the least-privileged USER principal on absent/indeterminate auth (`!authType → user`; a non-function `getType` → `undefined` → user), so `origin` can't be spoofed by an unauthenticated/odd caller.
- **S2 body-`source` strip:** the prompt mapping is `prompts.map(({ source: _, ...p }) => ({ ...p, origin: deriveV2PromptOrigin(p?.origin, isUserPrincipal) }))` — a body-supplied `source` is destructured out and never reaches `upsertPrompts`, and `origin` is principal-derived through the hardened `isUserPrincipal`.
- The two compose orthogonally in one `.map` (auth-hardening feeds `isUserPrincipal`; the strip drops `source`): origin stays principal-derived + user-fail-safe, source stays server-owned/body-ignored. No regression.

**Auto-merged O2b files consistent with S2's source dimension:**
- `prompts-storage.js` `mapRowToPrompt` keeps BOTH O2b's `origin: row.origin` (no `|| 'human'`) and S2's `source: canonicalizeSource(row.source) ?? row.source ?? 'config'`; the S2 `SOURCE_CANONICAL_SQL` sort/filter and `deriveV2PromptOrigin` are all present.
- `prompts.js` `makePromptTagInjector` threads BOTH `originValue` (O2b) and `sourceValue` (S2) via `const { originValue, sourceValue } = options`.
- `markets-subworkspace.js` retains S2's producing-system stamp (`ensureServerOwnedValue(..., DIMENSION.SOURCE, GENERATED_PROMPT_SOURCE_VALUE)`, `sourceId` appended to the generated prompts' tag ids).
- `tag-tree.js` still carries my S2 guard fix `requireServerOwnedRootId` (defined + used at both `resolveClosedValueInjection` and `ensureServerOwnedValue`), so the mid-rename `source`-root fail-loud behavior survived the rebase.

Green unit run + MERGEABLE confirm the merge is sound at the compile/test level; the spot-checks confirm no S2 change was clobbered by an O2b change or vice versa. `/review-pr` clean. Merge ≥ O6. Bot N/A (non-main base).

aliciadriani:
⚠️ **The `createPromptsByBrand` source-strip breaks the SR "Track" flow — this conflicts with SITES-47870.**

The change at `src/controllers/brands.js` destructures `source` out of every incoming prompt:

```js
const derivedPrompts = prompts.map(({ source: _, ...p }) => ({
  ...p,
  origin: deriveV2PromptOrigin(p?.origin, isUserPrincipal),
}));
```

The comment justifies this by saying the `|| 'config'` default "stays load-bearing for the internal writers that DO set it." But `upsertPrompts` has **exactly one caller in this repo** — `createPromptsByBrand` itself. The only writers that set a non-`config` source are the DRS pipelines writing the `prompts` table directly, *outside* this service. Every prompt created **through this service's API** now becomes `source = config`.

That's a problem, because the Strategic Recommendations "Track" flow in `project-elmo-ui` (origin/main) relies on sending `source` on this exact endpoint:

- **Semrush** Track → `POST /v2/orgs/:orgId/brands/:brandId/prompts` with `source: 'semrush'`, **unconditionally** (`SRTrackTopicToPromptsDialog.tsx`).
- **GSC / Citation Attempt / Synthetic Personas** cards send `source: 'gsc' | 'citation_attempt' | 'synthetic_personas'` (`GSCOpportunityCard.tsx`, `CitationAttemptCard.tsx`, `SyntheticPersonasCard.tsx`). These currently detour through the `llmo/config` draft via an `onCreatePrompts` override, so they don't hit `/prompts` *today* — but the cards' own default path is the same `POST /prompts`, so any render without that override is affected too.

Those four values are **exactly** `TRACKED_PROMPT_SOURCES` in `src/support/prompt-sources.js`. That allowlist (SITES-47870) exists specifically so *"UI-originated writes (e.g. `semrush` / `synthetic_personas` recommendation-accept) that never touch the DRS writer"* can persist a real source. This PR removes the very write surface that allowlist was built to guard — the two changes encode opposite contracts for the same field on the same endpoint.

**Concrete impact:** Semrush-tracked prompts lose their provenance and collapse to `config`. The LLMO usage report (which groups by `source`) can no longer attribute them. This is a silent data-loss regression, provable against shipped UI code.

**Suggested resolution** — pick one before merge:
1. **Scope the strip** so it doesn't apply to the recommendation-accept / Track path: keep a client-supplied `source` on `POST /v2/.../prompts`, still gated by `assertPermittedSource` (that chokepoint already prevents untracked values). This is the minimal reconciliation with SITES-47870.
2. **Re-model Track server-side** to carry origin without a client-sent field (e.g. via the originating recommendation/prompt id the server reads source from) — heavier, and needs the origin to be readable server-side.

Either way, worth a quick sync with the SITES-47870 owner so the `prompts.source` column contract is settled between LLMO-6282 and SITES-47870.

src/support/prompts-storage.js
```diff
@@ -633,7 +658,12 @@ export async function listPrompts({
     }
 
     if (hasText(source)) {
-      baseQuery = baseQuery.eq('source', source);
+      // Filter on the CANONICAL form so `citation-attempt` matches rows stored as
+      // `citation_attempt` too (source-dimension.md §3.1). The incoming value is
+      // folded with the same transform as the column, and matched against the SQL
+      // expression rather than the raw column.
+      const wantedSource = String(source).trim().toLowerCase().replace(/_/g, '-');
+      baseQuery = baseQuery.eq(SOURCE_CANONICAL_SQL, wantedSource);
```
aliciadriani: **Must fix (verified) — this filter cannot work against stock PostgREST.**

`SOURCE_CANONICAL_SQL` (`lower(replace(source, '_', '-'))`) is passed as the *column* to `.eq()` here, and to `.order()` on line 633. I confirmed the backend (`mysticat-data-service`) is **stock PostgREST 12** and that there is **no `source_canonical` computed column/function** on `prompts`. postgrest-js emits `?lower(replace(source,'_','-'))=eq.<v>` and `order=lower(replace(source,'_','-')).asc`, which stock PostgREST rejects with a **400 (PGRST100)** — filter keys and `order` accept only column names / computed columns / JSON paths, never inline function expressions.

The current unit tests pass only because the mock records the column string without validating it against PostgREST semantics, so `?source=…` and `sort=source` would 400 in production.

**Suggested fix:** add a `source_canonical GENERATED ALWAYS AS (lower(replace(source,'_','-'))) STORED` column in `mysticat-data-service` (fold into WP-S4 next to `idx_prompts_source_canonical`, and index *that* column), then filter/sort on the `source_canonical` column name here. Alternatively drop drift-folding on filter/sort and match the raw `source` column. An integration test against real PostgREST would catch this whole class of bug.

src/support/prompts-storage.js
```diff
@@ -606,7 +628,10 @@ export async function listPrompts({
     const sortCol = SORT_COLUMN_MAP[sort];
     if (sortCol) {
       const ascending = order === 'asc';
-      if (sortCol.includes('(')) {
+      if (typeof sortCol === 'object') {
+        // A SQL expression (e.g. the canonical `source` fold), ordered directly.
+        baseQuery = baseQuery.order(sortCol.expression, { ascending });
```
aliciadriani: Same must-fix as the filter on line 666: `.order()` with the raw expression `lower(replace(source, '_', '-'))` produces `order=lower(replace(source,'_','-')).asc`, which stock PostgREST cannot parse (400). See the detailed comment on line 666 for the suggested `source_canonical` generated-column fix.

src/support/prompts-storage.js
```diff
@@ -633,7 +658,12 @@ export async function listPrompts({
     }
 
     if (hasText(source)) {
-      baseQuery = baseQuery.eq('source', source);
+      // Filter on the CANONICAL form so `citation-attempt` matches rows stored as
+      // `citation_attempt` too (source-dimension.md §3.1). The incoming value is
+      // folded with the same transform as the column, and matched against the SQL
+      // expression rather than the raw column.
+      const wantedSource = String(source).trim().toLowerCase().replace(/_/g, '-');
```
aliciadriani: Nit: this inline fold is a **third copy** of the canonical transform (alongside `canonicalizeSource` and `SOURCE_CANONICAL_SQL` — this file already warns "the two must not drift"). Once the filter/sort mechanism is settled (see line 666), route the incoming value through a single shared helper — reuse `canonicalizeSource`, or extract a `foldSource(value)` — so the copies can't drift.

aliciadriani (commented):
Automated `/review-pr` pass (WP-S2). One verified merge-blocker on the `source` filter/sort mechanism, plus one drift nit. MysticatBot step skipped — this PR's base is `claude/wp-o2b-derivation`, not `main`, and the org bot only reviews main-based PRs. Details inline.

src/support/prompts-storage.js
```diff
@@ -633,7 +650,18 @@ export async function listPrompts({
...
+      // To keep "filter by the value the grid shows" working across a producer's two
+      // drift spellings, fold the incoming value to canonical (shared
+      // `foldSourceValue` — the single definition of the transform) and match BOTH
+      // the hyphen and underscore forms, so `citation-attempt` also finds rows stored
+      // as `citation_attempt`. WP-S4's `source_canonical` generated column replaces
+      // this with a single canonical-column match (and adds case-folding, which raw
+      // matching drops).
+      const wanted = foldSourceValue(source);
+      const sourceVariants = [...new Set([wanted, wanted.replace(/-/g, '_')])];
+      baseQuery = baseQuery.in('source', sourceVariants);
```
aliciadriani: **Fixed (must-fix).** Was `.eq("lower(replace(source,'_','-'))", …)` — an inline SQL expression stock PostgREST 400s on. Now matches the **raw `source` column** across both drift spellings via `.in()`, so `citation-attempt` still finds `citation_attempt` rows. WP-S4 replaces this with a single `source_canonical` generated-column match.

src/support/prompts-storage.js
```diff
@@ -427,6 +428,14 @@ const SORT_COLUMN_MAP = {
...
   status: 'status',
   updatedAt: 'updated_at',
+  // FIX (WP-S2 interim): sort on the RAW `source` column. Stock PostgREST (the
+  // backend) cannot order by an inline SQL expression like
+  // `lower(replace(source,'_','-'))` — it returns 400 — and there is no canonical
+  // column to name yet. WP-S4 adds a `source_canonical` generated column (+ its
+  // index) and ordering moves to that column name, so the two drift spellings order
+  // together; until then raw ordering only interleaves the `_`/`-` spellings of the
+  // same producer, which is cosmetic (source-dimension.md §3.1).
+  source: 'source',
```
aliciadriani: **Fixed (must-fix).** Sort key was an object `{ expression: "lower(replace(...))" }` PostgREST can't order by (400). Now the raw `source` column; the object-form dispatch branch was removed. Canonical ordering returns with WP-S4's `source_canonical` column.

aliciadriani (commented):
Applied the WP-S2 interim fix for the PostgREST filter/sort blocker (commit a681496). Filter/sort now run on the raw `source` column; the canonical-column fold is deferred to WP-S4. Inline notes on the fixed lines below.

src/support/serenity/prompt-tags.js
```diff
@@ -175,6 +294,65 @@ export function isClosedDimension(dimension) {
...
+ * The bare canonical FOLD of a `source` value — trim, lowercase, `_` → `-`
+ * (source-dimension.md §3.1) — with none of {@link canonicalizeSource}'s safety
+ * guards. This is the SINGLE definition of the transform in this repo: both
+ * `canonicalizeSource` (read/tag-write boundary) and the v2 list `source` filter
+ * fold (prompts-storage.js) route through it so the spellings can never drift.
+ *
+ * @param {string} value - a `source` value (already known to be a string).
+ * @returns {string} the folded value.
+ */
+export function foldSourceValue(value) {
```
aliciadriani, line 304: **Fixed (nit).** Extracted `foldSourceValue` as the single definition of the trim/lower/`_`→`-` transform. `canonicalizeSource` and the v2 list `source` filter now both route through it, so the copies that previously risked drift are collapsed to one.

src/support/serenity/prompt-tags.js
```diff
@@ -175,6 +294,69 @@ export function isClosedDimension(dimension) {
...
+ *
+ * Coerces with `String()` so a non-string filter value (e.g. a query param parsed
+ * as a number/array) folds instead of throwing; `canonicalizeSource` already
+ * guards the type before calling, so the coercion is a no-op on that path.
+ *
+ * @param {unknown} value - a `source` value.
+ * @returns {string} the folded value.
+ */
+export function foldSourceValue(value) {
+  return String(value).trim().toLowerCase().replace(/_/g, '-');
```
aliciadriani, line 305: **Fixed (self-review).** Routing the v2 `source` filter through this shared helper dropped the defensive `String()` the inline fold had — a non-string query param (parsed as a number/array) would hit `value.trim()` and 500. Restored the `String()` coercion here; `canonicalizeSource` type-guards before calling, so it is a no-op on that path.

aliciadriani (commented):
Iteration 2 (`/review-pr` on the updated PR). The must-fix and nit from iteration 1 are resolved; one self-review finding (a robustness regression I introduced) is fixed below. No remaining must/should/nits. 940 tests pass, lint clean, coverage 100% lines/funcs · 98.6% branches (>97%).

src/support/serenity/handlers/markets-subworkspace.js
```diff
@@ -288,7 +301,12 @@ async function generateAndAttachPrompts(transport, workspaceId, projectId, {
     // top-up) — route through `headroom.retryOnQuota` (no-op passthrough when the flag is OFF).
     // eslint-disable-next-line no-await-in-loop
     await headroom.retryOnQuota(
-      () => transport.createPromptsByIds(workspaceId, projectId, items, [...standardIds, typeId]),
+      () => transport.createPromptsByIds(
+        workspaceId,
+        projectId,
+        items,
+        [...standardIds, sourceId, typeId],
```
aliciadriani: **Merge resolution (conflict).** Base wrapped this `createPromptsByIds` call in `headroom.retryOnQuota` (#2874, the ~9s Semrush write-enforcement lag); WP-S2 added `sourceId` to the tag list. Combined both — the call stays inside `retryOnQuota` AND stamps `source/semrush`. Verified by `markets-subworkspace.test.js` (green).

aliciadriani (commented):
Iteration 3 (`/review-pr` after merging base `claude/wp-o2b-derivation`). Merge conflict resolved — PR is now **MERGEABLE**. Full suite **15396 passing** (exit 0, coverage gate met), lint clean. Two conflicts, both resolved by combining rather than choosing a side; notes inline. No new must/should fixes — the WP-S2 changes are unchanged from the reviewed state, and the base merge introduced no regressions.

test/support/serenity/tag-tree.test.js
```diff
@@ -480,13 +488,109 @@ describe('serenity tag-tree', () => {
...
       expect(err).to.be.an('error');
       expect(err.status).to.equal(502);
     });
   });
 
+  // A MID-RENAME project (legacy `source` root carrying ai/human, no `origin`):
+  // ensureDimensionRoots adopts `source` as authorship and leaves the producing
+  // `source` key undefined. The server-owned resolve paths must fail LOUD rather
+  // than let an undefined root id degrade into a stranded root-level create.
+  describe('source root distinctness guard (mid-rename, WP-O6-gated)', () => {
```
aliciadriani: **Merge resolution (conflict).** Base and WP-S2 each inserted a new `describe` block at this position — base's `resolveClosedValueInjection` (direct origin-dimension coverage) and WP-S2's `source root distinctness guard` (mid-rename 502 path). They cover different behaviour, so I kept BOTH rather than take one side. Both suites pass.

MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - well-designed dimension refactor with thorough self-review; no blocking issues remain.
**Complexity: HIGH** - large diff (20 files); OpenAPI contract changes.
_Changes: Introduces the 5th dimension root `source` (producing system) as an open, server-owned dimension with `canonicalizeSource`, `SERVER_OWNED_DIMENSIONS` refactor, and interim raw-column filter/sort (20 files)._
Note: Recommend a human read before merge - this change modifies shared OpenAPI contracts (`docs/openapi/*.yaml`). The bot review is a complement to, not a replacement for, a human read here.
Note: CI checks are currently failing - `it-postgres` waits WP-S1; other failures are cascading dependencies. Resolve before merge.

## Must fix before merge

No blocking issues found. All prior findings (PostgREST 400 on inline SQL expression, mid-rename `requireServerOwnedRootId` guard) have been addressed in subsequent commits.

<details>

- nit: `ALL_DIMENSIONS` is constructed as `[...OPEN_DIMENSIONS, ...CLOSED_DIMENSIONS]` which yields `[category, source, intent, origin, type]` - the ordering puts `source` before the closed dimensions, which is fine but worth noting differs from the `DIMENSION_ROOT_NAMES` order (`category, intent, origin, type, source`). Not a bug but a potential source of confusion if anyone assumes the two lists share ordering. - `src/support/serenity/prompt-tags.js:180`
- suggestion: The `sourceVariants` filter generates `[wanted, wanted.replace(/-/g, '_')]` using a `Set` for dedup, but for a value like `gsc` (no hyphens), both entries are identical and the Set deduplicates to one. Consider documenting that the single-element case is intentional (the test already covers it, so this is just a readability note). - `src/support/prompts-storage.js:663`
- nit: The `V2Prompt.source` schema in `schemas.yaml` shows `default: "config"` alongside `readOnly: true` - a `default` on a read-only field is technically a no-op in OpenAPI 3.0 (clients never send it, servers never need a default for output). Cosmetic only. - `docs/openapi/schemas.yaml:6558`

</details>

- PostgREST 400 on inline SQL expression for source filter/sort - now uses raw column with dual-spelling `.in()` match
- Mid-rename `source` root undefined flowing into `ensureChildren` - now guarded by `requireServerOwnedRootId` (502)
- Inline fold drift (third copy of the canonical transform) - extracted to shared `foldSourceValue`
- Non-string query param crash on `foldSourceValue` - `String()` coercion restored

---

_Note on SITES-47870 conflict (already identified in PR discussion): the `source` body-strip in `brands.js` removes a write surface that the SR "Track" flow relies on. This is already called out in the PR comments and gated by the DO-NOT-MERGE banner + WP-O6 prerequisite. Resolution should be settled before the gate lifts._

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 4m 47s | <b>Cost:</b> $2.46 | <b>Commit:</b> <code>d5e7fae35b19ec3a7a08af238ae05f0cdb35aadb</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

aliciadriani:
**`/review-pr` iteration 4 — rebased onto `main` (base flipped from `claude/wp-o2b-derivation`; WP-O2b #2866 is now in main).**

Re-reviewed the full WP-S2 diff on the new base. My two prior fixes survived the rebase intact (`foldSourceValue` + raw-column `.in('source', …)` filter/sort; coercion guard). PR is **MERGEABLE**.

- 946 tests passing on the rebased state · lint clean · coverage gate met previously.
- The two rebase-replay conflicts (`markets-subworkspace.js` `retryOnQuota`+`sourceId`; the two `tag-tree.test.js` `describe` blocks) resolved the same combine-both way.
- **No remaining must/should/nits** from my review.

Requesting MysticatBot now that the base is `main` (the original non-main-base blocker is gone).

src/support/serenity/prompt-tags.js
```diff
@@ -175,6 +294,69 @@ export function isClosedDimension(dimension) {
...
+ * fold (prompts-storage.js) route through it so the spellings can never drift.
+ *
+ * Coerces with `String()` so a non-string filter value (e.g. a query param parsed
+ * as a number/array) folds instead of throwing; `canonicalizeSource` already
+ * guards the type before calling, so the coercion is a no-op on that path.
+ *
+ * @param {unknown} value - a `source` value.
+ * @returns {string} the folded value.
+ */
+export function foldSourceValue(value) {
```
aliciadriani, line 304: (a) CONFIRMED no drift: `foldSourceValue` = `trim().toLowerCase().replace(/_/g,'-')` is the SINGLE definition of the transform, and `canonicalizeSource` delegates to it. So the tag-write canonicalizer, the v2 list source filter, and the sort all fold identically — consistent with the CLI Python twin (`.strip().lower().replace('_','-')`) and the prior JS shape.

src/support/prompts-storage.js
```diff
@@ -633,7 +650,18 @@ export async function listPrompts({
...
+      // FIX (WP-S2 interim): match on the RAW `source` column, not a canonical SQL
+      // expression — stock PostgREST 400s on `lower(replace(source,'_','-'))=eq.…`.
+      // To keep "filter by the value the grid shows" working across a producer's two
+      // drift spellings, fold the incoming value to canonical (shared
+      // `foldSourceValue` — the single definition of the transform) and match BOTH
+      // the hyphen and underscore forms, so `citation-attempt` also finds rows stored
+      // as `citation_attempt`. WP-S4's `source_canonical` generated column replaces
+      // this with a single canonical-column match (and adds case-folding, which raw
+      // matching drops).
+      const wanted = foldSourceValue(source);
```
aliciadriani: (b) FILTER keys on the folded slug: `wanted = foldSourceValue(source)`, matched via `.in('source', [wanted, wanted.replace(/-/g,'_')])` — so `citation-attempt` finds `citation_attempt` rows too, on the raw column. Correct for the common `_`/`-` drift, and it's the folded slug not the display label. Documented residual: case-folding is dropped (raw match is case-sensitive), deferred to WP-S4's canonical column — acceptable for machine-slug data, not a blocking bug.

src/support/prompts-storage.js
aliciadriani: (b) SORT orders by the RAW `source` column, so a producer's two drift spellings interleave rather than sort together — documented as cosmetic, folded ordering deferred to WP-S4. This is where the raw-column interim deviates from an ideal folded sort; driven by stock PostgREST 400ing on an inline-expression order-by. Not wrong data, just imperfect ordering until the canonical column exists.

src/support/prompts-storage.js
```diff
@@ -427,6 +428,14 @@ const SORT_COLUMN_MAP = {
   origin: 'origin',
   status: 'status',
   updatedAt: 'updated_at',
+  // FIX (WP-S2 interim): sort on the RAW `source` column. Stock PostgREST (the
+  // backend) cannot order by an inline SQL expression like
+  // `lower(replace(source,'_','-'))` — it returns 400 — and there is no canonical
+  // column to name yet. WP-S4 adds a `source_canonical` generated column (+ its
```
aliciadriani: 🚩 CROSS-PR FINDING (report-only, S4 held): this comment anticipates a WP-S4 `source_canonical` GENERATED COLUMN, but S4 #826 actually shipped an EXPRESSION INDEX `idx_prompts_source_canonical ON (lower(replace(source,'_','-')))`. S2's raw-column `.in()`/`.order()` do NOT use that expression index (the planner only picks it for the inline-SQL query PostgREST can't express — the approach S2 abandoned). The canonical VALUE is consistent (foldSourceValue == the index expr == the CLI), but the ARTIFACT S2 expects (a PostgREST-nameable generated column) is not what S4 delivered. Operator should reconcile before the 47 merges — the generated-column path is the coherent target; S4's expression index is currently an orphan relative to S2's queries. Do not fix here.

src/support/serenity/handlers/markets-subworkspace.js
```diff
@@ -288,7 +301,12 @@ async function generateAndAttachPrompts(transport, workspaceId, projectId, {
     // top-up) — route through `headroom.retryOnQuota` (no-op passthrough when the flag is OFF).
     // eslint-disable-next-line no-await-in-loop
     await headroom.retryOnQuota(
-      () => transport.createPromptsByIds(workspaceId, projectId, items, [...standardIds, typeId]),
+      () => transport.createPromptsByIds(
```
aliciadriani: (c) Rebase conflict #1 sound: the S4 `sourceId` (resolved via `ensureServerOwnedValue(..., GENERATED_PROMPT_SOURCE_VALUE)`) is threaded into the create's tag_ids (`[...standardIds, sourceId, typeId]`) AND the write is wrapped in the LLMO-6190 `headroom.retryOnQuota`. Both concerns coexist correctly. `createPromptsByIds` (not `createPromptsWithMetadata`) is right for S2's scope — WP1 #2877 layers the metadata create on top separately.

aliciadriani (commented):
Material re-review at `d5e7fae3` (base now main, rebased after O2b merged, MERGEABLE, 15411 unit). The sort/filter strategy changed since my last pass (`a9539eec`) — from the inline `lower(replace(source,'_','-'))` expression to a **raw `source` column + `foldSourceValue` app-side fold**. Assessed (a)/(b)/(c); `/review-pr` is clean. One CROSS-PR finding (report-only; S4 held) below.

**(a) `foldSourceValue` — no drift. CONFIRMED.** `foldSourceValue(value) = String(value).trim().toLowerCase().replace(/_/g, '-')` — the single definition of the transform, and `canonicalizeSource` now DELEGATES to it (then applies its empty/`:`/length/root-name guards). So the tag-write canonicalizer, the v2 list `source` filter, and the sort all fold through one function — consistent with the JS twin's prior `trim→lower→_→-` and the CLI Python twin (`.strip().lower().replace('_','-')`). No divergence.

**(b) raw-column + app-fold — correct within the interim's stated scope; two documented residuals.**
- FILTER: folds the incoming value (`wanted = foldSourceValue(source)`) and matches `.in('source', [wanted, wanted.replace(/-/g,'_')])` on the raw column — so a filter for `citation-attempt` finds rows stored as either `citation-attempt` or `citation_attempt`. It keys on the FOLDED SLUG (variant-expanded), never the display label. Residual (documented): the raw match is case-sensitive, so a mixed-case stored value (`Citation_Attempt`) is missed — the comment defers case-folding to WP-S4's canonical column. Acceptable given `prompts.source` is machine-produced lowercase slugs; not a blocking bug.
- SORT: `SORT_COLUMN_MAP.source = 'source'` orders by the RAW column, so the two spellings of one producer interleave rather than sort together — the comment calls this cosmetic and defers folded ordering to WP-S4. So the sort does NOT key on the folded slug; it's a documented interim tradeoff driven by stock PostgREST 400ing on an inline expression order-by.
Both residuals are honest, commented, and WP-S4-deferred — no un-canonicalized boundary that returns wrong DATA, only imperfect ordering + a case-drift filter miss on unlikely inputs.

**(c) core files + rebase conflicts — CONFIRMED, no drift.** I hashed the four S2 core files (`prompt-tags.js`, `tag-tree.js`, `prompts-storage.js`, `handlers/tags.js`) at head vs the validated tip `4868c8f8`: all four are BYTE-IDENTICAL (the commit-compare showing them differ was the three-dot rebase artifact, not reconstruction drift). Rebase conflict #1 (`markets-subworkspace`): the create is `headroom.retryOnQuota(() => transport.createPromptsByIds(ws, pid, items, [...standardIds, sourceId, typeId]), {callSite:'createPromptsByIds'})` — the S4 `sourceId` (from `ensureServerOwnedValue(..., GENERATED_PROMPT_SOURCE_VALUE)`) is threaded into the tag_ids AND the write is wrapped in the LLMO-6190 retry; both concerns coexist correctly (this is S2, so `createPromptsByIds` is right — WP1 layers metadata separately). Rebase conflict #2 (`tag-tree.test`): both describe blocks are present (`ensureServerOwnedValue` / `resolveClosedValueInjection` / `source root distinctness guard (mid-rename, WP-O6-gated)` alongside the pre-existing `resolveTypeValueInjection` / `findTagsInTree` / …) — no clobber.

---

## 🚩 CROSS-PR CONSISTENCY FINDING (report-only — S4 #826 is held; do NOT fix here)

**S2's current approach does NOT use S4's shipped index, and S2 anticipates a DIFFERENT S4 artifact than S4 actually delivered.**
- S2 (this PR, now canonical) queries the RAW `source` column: filter `.in('source', [hyphen, underscore])`, sort `.order('source')`. Neither references the expression `lower(replace(source,'_','-'))`, so the Postgres planner will NOT use S4 #826's `idx_prompts_source_canonical ON public.prompts (lower(replace(source,'_','-')))` — that expression index only serves a query whose predicate/order-by IS that expression, i.e. the ORIGINAL inline-SQL approach S2 abandoned (because stock PostgREST 400s on it). So relative to S2's current query shape, S4's index is an orphan.
- The canonical FORM is consistent across the fleet: `foldSourceValue` (S2) = `lower(replace(source,'_','-'))` (S4 index) = `canonicalize_source` (CLI) all produce the same slug (S4's index omits the trim, but stored slugs are untrimmed, so no practical divergence). It's the MECHANISM that diverges, not the value.
- The mismatch to reconcile: S2's code comments (prompts-storage.js:434/522/659) say "WP-S4 adds a `source_canonical` GENERATED COLUMN (+ its index)" and that ordering/matching will move to that plain column — but S4 #826 shipped an EXPRESSION INDEX, no generated column. These are different artifacts: a generated column is PostgREST-nameable (`.eq('source_canonical', …)` / `.order('source_canonical')`, which is exactly what S2's raw+fold interim is a placeholder for); an expression index is only usable by the inline-SQL query that stock PostgREST cannot express.
- **Recommendation for the operator (before the 47 track merges):** the raw-column-now → `source_canonical` generated-column-later path is the coherent, PostgREST-consumable target S2 already assumes; S4 #826's expression index does not serve it. Reconcile by having the WP-S4 DB change ship the `source_canonical` GENERATED COLUMN S2 expects (replacing / in addition to the expression index), OR consciously choose the inline-SQL+index path and give S2 a PostgREST-expressible route (view/RPC) — but that fights the 400 constraint that drove S2's pivot. Both #826 and #2867 are held pre-merge, so this is reconcilable now. Flagging, not fixing (S4 is held).

Verdict: `/review-pr` clean on the S2 code (a/b/c all sound, no reconstruction drift). The only open item is the cross-PR strategy/artifact reconciliation above. MysticatBot is re-requested for its first main-base pass (coordinator-triggered).

src/support/serenity/prompt-tags.js
```diff
@@ -122,15 +145,116 @@ export const CLOSED_DIMENSIONS = Object.freeze([
...
+
+/**
+ * Every dimension a caller may address on the create-tag endpoint. This is a
+ * MEMBERSHIP set (used only for `.includes` validation), so its order is
+ * irrelevant and INTENTIONALLY differs from {@link DIMENSION_ROOT_NAMES} — that
+ * list is provisioning ORDER (`category, intent, origin, type, source`), whereas
+ * this is grouped open-then-closed (`category, source, intent, origin, type`).
+ * Do not assume the two share an order.
+ */
 export const ALL_DIMENSIONS = Object.freeze([...OPEN_DIMENSIONS, ...CLOSED_DIMENSIONS]);
```
aliciadriani, line 172: **Fixed (bot nit).** Documented that `ALL_DIMENSIONS` is a membership set whose grouped open-then-closed order (`category, source, intent, origin, type`) intentionally differs from `DIMENSION_ROOT_NAMES`' provisioning order (`category, intent, origin, type, source`) — so no one assumes the two share an order. No behaviour change (used only for `.includes`).

src/support/prompts-storage.js
```diff
@@ -633,7 +650,20 @@ export async function listPrompts({
...
+      // drift spellings, fold the incoming value to canonical (shared
+      // `foldSourceValue` — the single definition of the transform) and match BOTH
+      // the hyphen and underscore forms, so `citation-attempt` also finds rows stored
+      // as `citation_attempt`. WP-S4's `source_canonical` generated column replaces
+      // this with a single canonical-column match (and adds case-folding, which raw
+      // matching drops).
+      const wanted = foldSourceValue(source);
+      // A value with no `_`/`-` (e.g. `gsc`) yields two identical variants that the
+      // Set collapses to one — a single-element `.in` is intentional, not a bug.
+      const sourceVariants = [...new Set([wanted, wanted.replace(/-/g, '_')])];
```
aliciadriani: **Fixed (bot suggestion).** Added a comment noting the single-element case is intentional: a value with no `_`/`-` (e.g. `gsc`) yields two identical variants that the `Set` collapses to one, so a one-element `.in` is expected, not a bug.

docs/openapi/schemas.yaml
```diff
@@ -6553,9 +6553,13 @@ V2Prompt:
...
     source:
       type: string
-      description: The source system or process that created the prompt
+      readOnly: true
+      description: >-
+        The producing system (source-dimension.md). READ-ONLY: server-owned, with
+        no write surface — a create/update body cannot set it. Returned in its
+        CANONICAL form (`lower(replace(source, '_', '-'))`), so `agentic_traffic`
+        reads as `agentic-traffic`.
       example: "config"
```
aliciadriani, line 6571: **Fixed (bot nit).** Dropped `default: "config"` from this `readOnly` `V2Prompt.source` field — a `default` on a read-only output field is an OpenAPI 3.0 no-op (clients never send it, servers never default output). `example: "config"` stays for docs.

aliciadriani (commented):
Cleared MysticatBot's 3 non-blocking nits (commit 26f88ec). The bot's review of the prior commit had **no must/should fixes** — these are the optional polish items. Inline notes on each fix below.

aliciadriani:
## `/review-pr` — final pass (base `main`, `26f88ec5`)

**Summary.** Server-owned `source` dimension + `canonicalizeSource` (20 files, +986/−192). Rebased onto `main` (WP-O2b #2866 now in main), **MERGEABLE**. Full local suite **15411 passing** (coverage gate met), lint clean. Every finding from all review rounds — mine and MysticatBot's — is addressed. MysticatBot **APPROVED** the prior commit with zero must/should.

### Must Fix
**None.** The historical must-fix (source filter/sort via an inline SQL expression stock PostgREST 400s on) is resolved: raw-column `.in()` dual-spelling match + raw-column sort (`509227a6`), verified against the real backend and confirmed by the bot. `source_canonical` generated column is the WP-S4 follow-up.

### Should Fix
**One open, already tracked & gated (no code change here):** the `source` body-strip in `brands.js` removes a write surface the Semrush "Track" flow relies on (**SITES-47870**). Deliberate design; PR is DO-NOT-MERGE-gated with a WP-O6 prerequisite, so this must be settled before the gate lifts.

### Nits
**None outstanding.** MysticatBot's 3 nits (ALL_DIMENSIONS ordering doc, single-element `sourceVariants` doc, `default` on a read-only schema field) all fixed in `26f88ec5`.

### What's Good
Clean open-vs-server-owned dimension split with a single shared `foldSourceValue`; fail-loud `requireServerOwnedRootId` (502) guard; read-side raw-value fallback so the grid still shows unparseable values; thorough spec-referenced comments + edge-case tests; conflicts resolved by combining both sides.

**CI note:** `it-postgres` is red pending WP-S1 (per the bot); resolve before merge.

aliciadriani:
**Correction to my earlier review summary** (my prior comment predated `23d11928`).

The `source` filter/sort is no longer the raw-column interim I described — commit `23d11928` ships the real **`source_canonical`** generated column (`GENERATED ALWAYS AS lower(replace(source,'_','-')) STORED`, btree-indexed), so:
- filter is a single `.eq('source_canonical', foldSourceValue(source))`,
- sort is on `source_canonical`,
- matching is genuinely **case-insensitive** and drift spellings **sort together** — the residual caveats I'd noted (case-sensitivity, interleaved ordering) are gone.

The OpenAPI filter/sort descriptions already matched this; a doc commit I pushed briefly reverted them to the old wording by mistake — reverted in `88945877`, so docs and code agree again. Net: the WP-S4 follow-up I'd flagged is done here, not deferred.

jjenscodee:
Picking up the SR "Track → Semrush" wiring (the layer that sits on top of the origin/source tracks), so flagging one more path for the same `source`-contract reconciliation — additive to the `createPromptsByBrand` note above, not a duplicate.

**The Serenity create path loses `source` the same way, and it isn't covered by the brandalf finding.**

`handleCreatePrompts` (`prompts-subworkspace.js`) injects `PROXY_CREATE_SOURCE_VALUE = 'config'` on every create — deliberately, per source-dimension.md §1 ("this human create dialog"). But the v2 **Strategic Recommendations "Track"** flow, in **Serenity mode**, writes through exactly this path — and those prompts genuinely originate from `semrush` / `gsc` / `citation_attempt` / `synthetic_personas` cards. So Serenity-mode tracked prompts collapse to `source = config` and lose their producing surface — the same class of loss flagged for the brandalf `/prompts` path, but on `/serenity/prompts`, which has no `prompts` row to derive from.

So whichever way the `/prompts` strip vs SITES-47870 gets reconciled, the **Serenity create path needs the matching treatment** — otherwise Serenity-mode orgs lose the surface that (post-fix) brandalf orgs keep, and the usage report (SITES-47870) shows `config` for them too.

Not necessarily something to solve in this PR — just making sure the Serenity path is in scope when you + @dzehnder settle the `prompts.source` contract. Happy to sync.

aliciadriani (commented):
## `/review-pr` — WP-S2 #2867 (head 8894587, `source_canonical`)

### Summary
Server-owned `source` producing-system dimension. Exceptionally clean and thoroughly documented: the `SERVER_OWNED_DIMENSIONS` refactor cleanly splits the two axes (open/closed vocabulary vs who-writes), `canonicalizeSource`/`foldSourceValue` are a single source of truth applied at both derivation boundaries, the injector `source` asymmetry mirrors `origin` (strip-by-resolved-id, memoized), the `source_canonical` generated-column sort/filter is correct, and `requireServerOwnedRootId` turns the mid-rename undefined-root into a loud 502. Verified locally on node 24: type-check + lint clean, **946 affected-suite tests passing**. (State note: PR is `mergeable=CONFLICTING` — main advanced; that is a merge-window ≥O6 JIT rebase, not a code issue.)

### Must Fix
None.

### Should Fix
None.

### Nits
- **Split-brain `source` classification (forward-looking, low-confidence).** `ensureDimensionRoots` runs `childrenAreAuthorship` only when `!authorshipId`, so a (should-not-occur) project carrying BOTH an `origin` root AND a `source` root with `ai`/`human` children would classify that `source` root as producing-system and surface its authorship children as `source` values. The migration invariant (one authorship root per project) + WP-O6 make this unreachable, and the author deliberately avoided an extra read on the common post-rename path — but the CLI (WP-S4) guards this split-brain explicitly, so api-service should at least DOCUMENT the assumption for parity. Fix = a comment, not a runtime guard.

### What's Good
- The two-axis model (`SERVER_OWNED_DIMENSIONS` × `isClosedDimension`) is a genuinely clarifying refactor — `isClosedDimension` keeps its one honest job (vocabulary only).
- `foldSourceValue` as the single fold shared by the tag-write boundary and the v2 filter — drift is structurally impossible.
- `requireServerOwnedRootId` fails loud (502) instead of silently minting a stranded root on a mid-rename project.
- Precise, extensive JSDoc — the WP-O6 gating and the `(parent,name)` collision rationale are all captured.

src/support/serenity/tag-tree.js
```diff
@@ -296,13 +318,29 @@ export async function ensureDimensionRoots(transport, semrushWorkspaceId, projec
...
 
-  // Return the roots in canonical order, with `origin` carrying the resolved id.
+  // The producing-system `source` root: the existing physical root when it is NOT the
+  // authorship root, the freshly-created one on a project that had no `source` root at
+  // all, and `undefined` while the `source` root still means authorship (WP-O6-gated).
+  let producingSourceId;
+  if (legacySourceIsAuthorship) {
+    producingSourceId = undefined;
+  } else if (physicalSourceId) {
+    producingSourceId = physicalSourceId;
```
aliciadriani: **Nit (forward-looking, low-confidence).** This branch takes an existing physical `source` root as producing-system whenever `origin` is present, without re-checking `childrenAreAuthorship`. A split-brain project (both `origin` AND a `source` root carrying `ai`/`human` — a state the migration/reshape should not produce, and WP-O6 removes this path) would surface authorship children as `source` values. The CLI (WP-S4) guards this explicitly; api-service relies on the invariant. Recommend DOCUMENTING the assumption here — a comment, not a runtime `childrenAreAuthorship` read (which would regress the deliberate common-path no-extra-read optimization).

aliciadriani: **Documented** in `c8a14b06`: added a comment on this branch spelling out the relied-on migration invariant (no project carries both `origin` and a `source`-authorship root — the reshape renames source→origin in place; WP-O6 removes the path) and why no runtime `childrenAreAuthorship` re-read is added here (it would regress the deliberate common-path no-extra-read optimisation). Comment over guard, per the review recommendation; the CLI/WP-S4 still guards the equivalent split-brain on its write side.

aliciadriani (commented):
## `/review-pr` re-review — WP-S2 #2867 (head c8a14b06)

The single Nit (undocumented split-brain assumption in `ensureDimensionRoots`) is resolved — `c8a14b06` documents the relied-on migration invariant and why no runtime guard is added. Comment-only change; type-check + lint clean on the new head, affected suites were 946 passing pre-change (unaffected by a comment).

Re-scanned the full diff: **no Must-Fix, no Should-Fix, no remaining nits or suggestions.** Loop complete.

State note (not a review finding): the PR is `mergeable=CONFLICTING` (main advanced) and its `it-postgres` needs the released WP-S1 mock — both are merge-window (≥O6) concerns: a JIT rebase + the mock release, not code changes.

**Verdict: clean.**

aliciadriani:
@jjenscodee spun up a prototype of the write-site fix so we have something concrete to react to in the `prompts.source` contract discussion: #2910 (draft/RFC, stacked on this branch).

It adds a **server-only per-item `input.source`** seam to `makePromptTagInjector` so the SR "Track" flow can stamp the real producer (`semrush`/`gsc`/…) instead of collapsing to `config` — canonicalized, falls back to the default when unparseable, and deliberately NOT wired to the request body (no client write surface). The open question (where the Track handler reads each card's source) is called out in the PR. Not for merge until the contract's settled — flagging for the sync.

aliciadriani:
@jjenscodee agreed this is a real gap distinct from the brandalf `createPromptsByBrand` note.

Confirming the mechanism: the Serenity create path (`handleCreatePrompts`) injects `PROXY_CREATE_SOURCE_VALUE = 'config'` unconditionally (deliberate for the human dialog, source-dimension.md §1), so a Serenity-mode **Track** write — which routes through it for prompts that genuinely originate from `semrush` / `gsc` / `citation_attempt` / `synthetic_personas` cards — collapses to `source = config` and loses the producer, with no `prompts` row to recover it from.

Keeping this **out of #2867**: WP-S2 is scoped to the server-owned `source` dimension + the §1 human-dialog default. What the Track path *should* stamp instead is the open `prompts.source` write-contract question — the same one gating the `/prompts` strip vs SITES-47870.

Captured it in **LLMO-6556** (relates to SITES-47870 + #2867, blocked-by the contract decision) so it's in scope for the reconciliation and won't drop. Happy to sync with you + @dzehnder to settle it.

aliciadriani (commented):
## `/review-pr` (fresh pass) — WP-S2 #2867 (head c8a14b06)

Re-reviewed end-to-end; head is unchanged since the prior pass, so this re-verifies rather than re-derives. This time I read the **OpenAPI diffs in full** (skimmed last time): `V2Prompt.source` → `readOnly` + canonical form (dropped the stale `default: config`), **`V2PromptInput.source` removed** (no write surface), the v2 filter/sort docs describe the canonical fold + case-insensitivity, the `sort` enum adds `source`, and the serenity create-tag prose + `type` enum correctly move to the five-root SERVER-OWNED model (`category` customer-open; `intent`/`origin`/`type` closed; `source` open). All consistent with the code.

Re-verified on node 24 from a clean checkout of the head: type-check + lint clean previously; key suites green (332 in this subset; 946 across the full affected set earlier).

### Must Fix
None.

### Should Fix
None.

### Nits
None new. The one prior nit — the split-brain assumption in `ensureDimensionRoots` — is **documented in this head** (`c8a14b06`), which was the resolution agreed last pass.

### What's Good
- Two-axis model (`SERVER_OWNED_DIMENSIONS` × `isClosedDimension`) is a clarifying refactor; `foldSourceValue` is a single shared fold; the injector `source` asymmetry mirrors `origin`; `requireServerOwnedRootId` fails loud on the mid-rename undefined root; the OpenAPI read-only/canonical contract matches the code exactly.

### State notes (not review findings)
- `mergeable=CONFLICTING` — `main` advanced ~64 commits; a JIT rebase (≥O6) is held per plan, since #2910 ([PROTOTYPE] Track per-item source passthrough, LLMO-6556) is **stacked on this branch** and a rebase would move its base under an active RFC.
- `it-postgres` still awaits the released WP-S1 mock — deferred, out of scope.

**Verdict: clean** (no Must-Fix / Should-Fix / new nits).

MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - well-designed dimension refactor with comprehensive test coverage; no blocking issues found after 12+ review iterations.
**Complexity: HIGH** - large diff (20 files); OpenAPI contract changes.
_Changes: Introduces the 5th dimension root `source` (producing system) as an open, server-owned dimension with `canonicalizeSource`, `SERVER_OWNED_DIMENSIONS` refactor, `requireServerOwnedRootId` guard, and `source_canonical` generated-column filter/sort (20 files)._
Note: Recommend a human read before merge - this change modifies shared OpenAPI contracts (`docs/openapi/*.yaml`). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge

No blocking issues found. All prior findings (PostgREST 400, mid-rename `requireServerOwnedRootId` guard, fold drift, `String()` coercion, split-brain documentation, MysticatBot nits) have been addressed.

<details>

- nit: `resolveClosedValueInjection` was not renamed alongside `ensureClosedValue` -> `ensureServerOwnedValue`, creating a naming inconsistency within this PR since both now serve the open `source` dimension - `src/support/serenity/tag-tree.js:694`
- suggestion: The "prefers origin over a legacy source when both roots present" test asserts `roots.get('origin')` but not `roots.get('source')` on the post-rename common path; adding one line (`expect(roots.get('source')).to.equal('root-source')`) would explicitly cover the producing-system resolution - `test/support/serenity/tag-tree.test.js:337`
- suggestion: Add an explicit negative assertion `expect(result.body.tagIds).to.not.include(TAG_IDS.sourceConfig)` on the update path to directly prove source is CREATE-only, mirroring the origin non-injection assertion - `test/support/serenity/handlers/prompts.test.js:2571`
- suggestion: Add a `mapRowToPrompt` test for `row.source = 'source'` (dimension-root shadow) confirming the raw value is returned, not `null` or `'config'` - `test/support/prompts-storage.test.js`
- nit: `provisionDimensionTree` test could assert `values.has('source') === false` to prove the open source dimension is not pre-provisioned (mirrors the existing `category` assertion) - `test/support/serenity/tag-tree.test.js:449`

</details>

- PostgREST 400 on inline SQL expression for source filter/sort - now uses `source_canonical` generated column
- Mid-rename `source` root undefined flowing into `ensureChildren` - guarded by `requireServerOwnedRootId` (502)
- Inline fold drift (third copy of canonical transform) - extracted to shared `foldSourceValue`
- Non-string query param crash on `foldSourceValue` - `String()` coercion restored
- ALL_DIMENSIONS ordering vs DIMENSION_ROOT_NAMES - documented as intentional
- `default: "config"` on readOnly field - removed
- Split-brain assumption in `ensureDimensionRoots` - documented in c8a14b06

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 55s | <b>Cost:</b> $7.60 | <b>Commit:</b> <code>b9a78bb667648f58870b65429703aefd8b291014</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - clean commit addressing all 5 prior non-blocking suggestions; no new issues.
**Complexity: HIGH** - large diff (20 files); OpenAPI contract changes.
_Changes: Renames `resolveClosedValueInjection` to `resolveServerOwnedValueInjection` and adds 4 test assertions from prior review suggestions (20 files)._
Note: Recommend a human read before merge - this change modifies shared OpenAPI contracts (`docs/openapi/*.yaml`). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge

No blocking issues found. All prior findings have been addressed across the full review history.

- `resolveClosedValueInjection` naming inconsistency with `ensureServerOwnedValue` - renamed to `resolveServerOwnedValueInjection`
- Missing `roots.get('source')` assertion on post-rename path - added
- Missing negative assertion proving source is CREATE-only on update - added
- Missing `mapRowToPrompt` test for dimension-root shadow value - added
- Missing `provisionDimensionTree` assertion that source is not pre-provisioned - added

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 50s | <b>Cost:</b> $4.50 | <b>Commit:</b> <code>c2552d11b0ba1f800b234116daa97f5905092778</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jjenscodee (approved):

PR #2871, opened 2026-07-21
fix(state-mappings): scope internal-admin writes to their own org

ravverma:
## What

Internal admins may now manage state-layer access mappings **within their own org**, instead of being blocked outright.

- **CREATE**: an internal admin may create a mapping only when the target resource (site for ASO, brand for LLMO) belongs to the caller org. Resolved via the resource row (organization_id) → organization (ims_org_id) and compared to the caller org. Fail-closed: unknown/foreign resource → 403.
- **PATCH / DELETE**: allowed for admins with no extra check — those operate on an existing row fetched/mutated scoped to the caller ims_org_id (the RPC filters on it), so an admin can only touch rows already in their own org.
- **Non-admin managers**: unchanged (scoped by gateManager / their can_manage_users authority).

## Why

The earlier blanket block (an internal admin got 403 on all writes) made it impossible for a legitimate org admin to grant permissions inside their own org. Scoping admin writes to owned resources restores that while still preventing cross-org grants — the mapping row is stamped with the caller org, so an unchecked resourceId would otherwise let an admin author a binding referencing another orgs resource.

Mirrors AccessControlUtil.hasAccess org-membership resolution (resource → org → ims_org_id).

## Tests

- Unit (controller): admin create succeeds for an in-org resource; 403 for a foreign resource; 403 fail-closed for an unknown resource; admin PATCH/DELETE allowed (org-scoped by RPC, no resource lookup).
- Unit (util): getResourceImsOrgId resolves brand/site → org ims_org_id, and returns null for unknown type / missing row / missing ims_org_id; throws on PostgREST error.
- IT: state-mapping writes driven by a non-admin org-wide FACS manager persona (facsManager); reads stay on admin. 188 unit passing; lint clean.

### DIFF
--- a/src/controllers/state-access-mappings.js
+++ b/src/controllers/state-access-mappings.js
@@ -454,6 +454,26 @@ function StateAccessMappingsController(context) {
   }
 
   /**
+   * Blocks internal admins from **mutating** state-layer bindings — create,
+   * update (PATCH), and delete/revoke (DELETE empties the capability set).
+   * Grants must originate from a real FACS-layer or state-layer
+   * `can_manage_users` holder in the customer org so every mutation carries an
+   * accountable `created_by` / `updated_by` — an internal admin identity is a
+   * platform-operator bypass (see `callerHasFacsManageUsers`), not an org
+   * manager, and must never author or revoke customer ReBAC grants. Reads
+   * (list / history / capabilities / audit) are unaffected.
+   *
+   * @param {object} ctx
+   * @returns {Response|null} `forbidden` when the caller is an internal admin.
+   */
+  function blockInternalAdminWrite(ctx) {
+    if (ctx.attributes?.authInfo?.isAdmin?.()) {
+      return forbidden('Internal admins may not create, update, or delete access mappings');
+    }
+    return null;
+  }
+
+  /**
    * Read-scope rule for list / history (hybrid-model §3, mac-state-layer):
    * **org-wide reads admit FACS-layer `can_manage_users` only**. An org-wide
    * manager reads anything in the org; a state-layer manager must scope the read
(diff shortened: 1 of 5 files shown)


MysticatBot (approved):
Hey @ravverma,

**Verdict: Approve** - clean, well-scoped authorization guard with thorough test coverage.
**Complexity: MEDIUM** - API surface + ReBAC signal.
_Changes: Adds `blockInternalAdminWrite` guard to prevent internal admins from creating, updating, or deleting state-layer access mappings, wired into all three mutation endpoints before `gateManager` (2 files)._
Note: Recommend a human read before merge - possible divergence from the documented design (heuristic, flagged by the conventions reviewer). The `mac-state-layer.md` bypass table documents internal identities bypassing FACS enforcement entirely but does not carve out state-layer mutations as an exception; updating the spec is out of scope for this PR (lives in `mysticat-architecture`). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: Optional chaining on `isAdmin?.()` is overly defensive - if `isAdmin` were removed or renamed upstream, the guard would silently pass through rather than failing loudly. Since `authInfo` is guaranteed populated by auth middleware at this point, consider `ctx.attributes.authInfo.isAdmin()` without the `?.` on the method call. - `src/controllers/state-access-mappings.js:468`
- suggestion: The three admin-block tests use `facsPermissions: []`, which means `gateManager` would also return 403 for the same context if the admin block were removed - the tests cannot distinguish which guard rejected. Give the admin `facsPermissions: ['llmo/can_manage_users']` in the patch and delete tests (matching the fourth test's pattern) so the admin block is provably the sole rejection reason. - `test/controllers/state-access-mappings.test.js:1241`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 9m 39s | <b>Cost:</b> $5.35 | <b>Commit:</b> <code>6d3e4ca0995d12cad9b3a6959868fc0c2561561e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2873, opened 2026-07-21
feat(serenity): publish-after-populate finalize on the sub-workspace path (LLMO-5492)

andreeastroe96:
## What & why

[LLMO-5492](https://jira.corp.adobe.com/browse/LLMO-5492) — publish-after-populate for Serenity projects.

This **re-targets [#2584](https://github.com/adobe/spacecat-api-service/pull/2584)** onto `main`'s **sub-workspace** onboarding surface. #2584 was built against the flat-mode `POST /llmo/onboard` `markets[]` fan-out, which is superseded on `main` — Serenity onboarding now runs through brand-create / `/serenity/activate` / `/serenity/markets` (`provisionBrandSubworkspace` → `handleCreateMarketSubworkspace`), which already publishes **inline** after populate. The deferred-publish mechanism didn't exist on `main`; this PR ports it and wires it into the sub-workspace create path.

## Scope

- **`SERENITY_DEFER_PUBLISH` flag** (`defer-publish-active.js`, **default OFF** → behavior-preserving). When ON, `provisionBrandSubworkspace` uses `publishMode: 'skip'` so each project is left a **draft** for a later finalize step. Flag OFF preserves the existing `require`/`best-effort` logic byte-for-byte.
- **`handlers/finalize.js` → `finalizeSerenityProjects`**: push prompts (publish deferred) → set models per slice → publish each project **once**, with populate gates (`noPrompts`/`noModels`) so an empty project never goes live, and a **bounded** publish-status confirm within the Lambda budget.
- **`handlers/publish-status.js`**: `classifyPublishStatus` + `pollProjectPublished` (bounded confirm; the unbounded ≤900s reconcile stays the DRS/worker's job). Enum aligns with the existing `subworkspace-projects.js` `mapPublishStatus`.
- **`rest-transport.getProjectStatus`** for the publish-completion read.
- **`handleCreatePrompts` / `handleUpdateModels`** gain a `{ publish }` option so finalize can batch a single publish. Default `true` keeps the standalone-endpoint contract; `syncModelsForProject` already supported the switch.
- **`ERROR_CODES.PUBLISH_QUOTA_EXHAUSTED`** for the disguised-405 no-quota rejection (classified via main's `isMeteredQuota`).

## Deferred / out of scope

- **The DRS-completion trigger** that calls `finalizeSerenityProjects` lives in a separate repo (audit-worker/DRS) — intentionally not wired here; this is the reusable mechanism it will call.
- **The unbounded reconcile poll** — also the worker's job.

## ⚠️ One review item

`rest-transport.getProjectStatus` currently passes `query: { draft: 'false', type: 'ai' }` (mirroring the existing `getProject`). #2584's design note calls for the **v1 default view** to read a never-published draft's `publish_status` faithfully, and warns the live view can distort it. `getProject` defaults to `draft: true`. **Please confirm the correct `draft` value against the Semrush project-read semantics** before merge — this is the one behavior I couldn't verify from the code alone.

## Tests

New/extended unit tests (all green; lint + `type-check` gates pass): `finalize.test.js` (12), `publish-status.test.js` (13), plus `{ publish: false }` cases for `handleCreatePrompts` / `handleUpdateModels`, `getProjectStatus` transport, and `PUBLISH_QUOTA_EXHAUSTED`. Full serenity unit suite: 1028 passing.


src/support/serenity/rest-transport.js
```diff
@@ -460,6 +460,30 @@ export function createSerenityTransport({ env, imsToken }) {
...
+     * @param {string} projectId
+     * @returns {Promise<object>} raw project JSON carrying `publish_status`.
+     */
+    async getProjectStatus(semrushWorkspaceId, projectId) {
+      return projects.getProject(
+        {
+          params: {
+            path: { id: semrushWorkspaceId, project_id: projectId },
+            query: { draft: 'false', type: 'ai' },
+          },
```
MysticatBot, line 714: issue (blocking): `getProjectStatus` hardcodes `query: { draft: 'false', type: 'ai' }`, requesting the live-view materialisation. The existing `getProject` defaults `draft: true` and its JSDoc explains why: the draft view is faithful. For a project that has never been published (which is exactly the case this function targets), the live view may not carry `publish_status` faithfully - your own design note from PR 2584 warns the live view can distort it, and the team's ADR-006 documents `draft: true` as the draft-faithful read.

Since `pollProjectPublished` depends on reading `publish_status` transitioning from `draft` through `publishing` to `live`, an unfaithful read could cause the poll to misclassify a project as perpetually pending (or never detect a failure).

**Fix:** Change to `draft: 'true'` (matching `getProject`'s default), or delegate to the existing `getProject(wsId, pid)` so there is only one query-parameter surface to maintain. If empirical testing confirms the live view works for `publish_status`, document that as a known exception to the ADR guidance.

src/support/serenity/handlers/finalize.js
```diff
@@ -0,0 +1,338 @@
...
+  //        no getProjectStatus            → publishPending (worker reconciles) —
+  //        we never report an unconfirmed 202 as live.
+  const { confirmAttempts = 1, confirmIntervalMs = 0 } = options;
+  const canConfirm = typeof transport.getProjectStatus === 'function';
+  for (const projectId of publishable) {
+    try {
+      // eslint-disable-next-line no-await-in-loop
+      await transport.publishProject(semrushWorkspaceId, projectId);
+    } catch (e) {
+      if (isMeteredQuota(e)) {
```
MysticatBot, line 325: issue (blocking): The publish loop iterates all publishable projects sequentially: publish (network call) then poll (1+ network calls). With the defaults (`confirmAttempts=1, confirmIntervalMs=0`) this is 2 calls per project - tolerable for a handful. But the function accepts arbitrary `confirmAttempts`/`confirmIntervalMs` from the caller, and nothing caps the total wall time. A caller passing `confirmAttempts: 10, confirmIntervalMs: 2000` with 5 projects would burn ~100s in polls alone. If this runs in a Lambda, a mid-loop timeout leaves some projects published and the rest as dangling drafts with no record of what completed.

**Fix (pick one):**
- Add a ceiling: `const safeAttempts = Math.min(confirmAttempts, 3)` to bound regardless of caller input.
- Add a wall-clock budget check inside the loop that bails remaining projects to `publishPending` when time is low.
- Or document a hard contract that callers must never pass `confirmAttempts > 1` (and add a guard assertion).

MysticatBot (changes requested):
Hey @andreeastroe96,

**Verdict: Request changes** - the `draft: 'false'` query in `getProjectStatus` is a correctness risk for never-published projects, and you flagged it yourself.
_Changes: adds a publish-after-populate finalize mechanism gated by a `SERENITY_DEFER_PUBLISH` flag, with bounded publish-status confirmation and populate gates (16 files)._

## Must fix before merge
1. [Important] `getProjectStatus` passes `draft: 'false'`, which likely reads the wrong view for never-published drafts - `src/support/serenity/rest-transport.js:475` (details inline)
2. [Important] Sequential publish loop has no wall-time guard against Lambda timeout - `src/support/serenity/handlers/finalize.js:278` (details inline)

<details>

- nit: nested ternary with eslint-disable is harder to parse than an if/else - `src/support/serenity/brand-provisioning.js:222`
- suggestion: log a warning when `classifyPublishStatus` encounters an unknown status string, so Semrush contract drift is visible to operators - `src/support/serenity/handlers/publish-status.js:99`
- suggestion: `sliceKey` silently produces `"NaN::..."` for a missing or non-numeric `geoTargetId`; a log line for malformed trigger payloads would ease debugging - `src/support/serenity/handlers/finalize.js:25`
- suggestion: consider collapsing the 8-positional-parameter signature into a named-params object while the function is new and has no callers yet - `src/support/serenity/handlers/finalize.js:101`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 12m 12s | <b>Cost:</b> $9.12 | <b>Commit:</b> <code>82b409857b2fbfa3a5a1006fbb2c3c444da19f87</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

andreeastroe96:
Both must-fix items are addressed (and the branch is updated to latest main):

**1. `getProjectStatus` draft view** — changed `draft: 'false'` → **`draft: 'true'`** so a never-published project's `publish_status` is read from the draft view (the live view empties a never-published draft per serenity-docs #12 §10 / ADR-006). Matches `getProject`'s default. `getProjectStatus` test now asserts `draft=true`.

**2. Finalize publish loop wall-time** — bounded on **both** axes:
- **Across projects:** opt-in `options.deadlineMs` guard — once the remaining Lambda budget drops below a 5s safety margin, remaining projects are deferred to the worker reconcile (`publishPending`, reason `'deadline'`) instead of a mid-loop timeout.
- **Per project:** `confirmAttempts` is now hard-capped — `safeAttempts = min(max(1, confirmAttempts), 3)` — so no caller input can make a single project's confirm poll run away.

Tests added for the deadline guard and the 3-attempt cap. Serenity suite 1140 passing; lint + type-check clean; changed files at 100% line coverage.

Commits: `e30eaa29` (draft view + deadline guard) and `b54047d2` (confirm-poll cap).

MysticatBot (approved):
Hey @andreeastroe96,

**Verdict: Approve** - both prior blocking findings are addressed; no new blocking issues found.
_Changes: adds a publish-after-populate finalize mechanism gated by a `SERENITY_DEFER_PUBLISH` flag, with bounded publish-status confirmation and populate gates (16 files)._

- `getProjectStatus` now passes `draft: 'true'` (draft-faithful read, matching `getProject` default and ADR-006)
- Publish loop is bounded on both axes: per-project `safeAttempts` cap at 3, cross-project `deadlineMs` wall-time guard deferring remaining to the worker reconcile

<details>

- suggestion: prompts populate gate fires when `prompts.created.length === 0` regardless of `prompts.skipped.length` - if an idempotent re-fire deduplicates all prompts (skipped, not failed), publish is incorrectly blocked even though the project already has content upstream. Consider guarding with `prompts.created.length === 0 && prompts.skipped.length === 0` - `src/support/serenity/handlers/finalize.js:220`
- suggestion: after the first `PUBLISH_QUOTA_EXHAUSTED` rejection (workspace-level limit, not per-project), remaining projects are guaranteed to fail the same way. Short-circuiting the loop into `publishFailed` on the first quota hit would save network calls and prevent noisy repeated alerts - `src/support/serenity/handlers/finalize.js:310`
- nit: `sliceKey` silently produces `"NaN::..."` when `geoTargetId` is undefined/non-numeric from a malformed trigger payload, causing a silent match failure (project skipped as `noModels`). A log.warn for NaN would ease debugging - `src/support/serenity/handlers/finalize.js:25`
- nit: `PUBLISH_OUTCOME` is not frozen (unlike `ERROR_CODES` which uses `Object.freeze`) - minor consistency - `src/support/serenity/handlers/publish-status.js:45`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 18s | <b>Cost:</b> $5.21 | <b>Commit:</b> <code>b54047d23472f4c780c43ec436fd5131f3f72bc3</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2874, opened 2026-07-21
fix(serenity): span the ~9s Semrush write-enforcement lag in retryOnQuota

byteclimber:
## Summary
- Live-verified against the sandbox masters: a JIT resource top-up succeeds immediately, but the Semrush gateway's write-enforcement checkpoint lags the transfer by a reproducible ~9s, so a metered write landing in that window still 405s as a disguised metered-quota rejection even though headroom is already sufficient. The previous `retryOnQuota` did a single immediate retry, which re-lands in the same window and doesn't help.
- `retryOnQuota` now polls `fn()` with backoff (default 3 attempts / 3s spacing), bounded by a per-request deadline computed once at guard construction so stacked wrap sites in one request (`createProject` → `createPromptsByIds` → `publishProject`) share a budget instead of compounding to ~36s. `ensure()` is called once per recovery, not per poll attempt, to avoid per-child lock contention.
- Wraps the three previously-unprotected write sites (`createProject`, `createPromptsByIds`, `createOnePrompt`) in addition to the existing publish sites.
- Adds a `QuotaRetryOutcome` metric (recovered/exhausted, dimensioned by attempt + call site) for rollout observability.

## Test plan
- [x] `npm run lint && npm run type-check && npm test` — full suite green (15353 passing)
- [x] Unit tests for the poll/deadline behavior: later-attempt recovery, exhaustion, deadline-vs-attempt-cap cutoff, `ensure()` called once not per-attempt, non-metered mid-loop propagation
- [x] Handler-level integration tests for the new wrap sites, including a per-item-keyed concurrent test (not a flat call-index, to avoid nondeterminism under real concurrency) and a mixed-outcome test
- [ ] Live re-verification against the sandbox masters with `SERENITY_DYNAMIC_ALLOCATION` on, before considering this closed for rollout


MysticatBot (approved):
Hey @byteclimber,

**Verdict: Approve** - well-designed poll-retry with correct termination bounds and thorough test coverage.
**Complexity: MEDIUM** - medium diff, tests-heavy.
_Changes: Evolves `retryOnQuota` from a single immediate retry to a bounded poll-retry loop (3 attempts / 3s backoff / 9s shared per-request deadline) and wraps three previously-unprotected write sites (6 files)._

<details>

- suggestion: Add a multi-call-site sequential test on a single guard instance to lock the shared-deadline property (the PR's key design intent has zero dedicated coverage - a regression moving `requestDeadline` into the closure would pass all current tests) - `test/support/serenity/dynamic-allocation-active.test.js`
- suggestion: Spy on `recordQuotaRetryOutcome` in at least one recovered + one exhausted test to catch silent removal or typos in the outcome/callSite values - `test/support/serenity/dynamic-allocation-active.test.js`
- nit: `Attempt` metric dimension is a raw number with no enforcement of the low-cardinality contract documented in the metrics module header - today's cardinality (3) is safe but worth a doc note that `maxAttempts` changes require a metric-cardinality review - `src/support/serenity/allocation-metrics.js:169`
- nit: Stale comment still says "bounded top-up+retry" but behavior is now "bounded poll-retry" (other call sites were updated, this one was missed) - `src/support/serenity/handlers/markets-subworkspace.js:1005`
- nit: `callSite` defaults to `'unknown'` on the enabled path, creating a latent open-vocabulary dimension value if a future caller omits it - consider a log warning or making it required in the type signature - `src/support/serenity/dynamic-allocation-active.js:131`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 48s | <b>Cost:</b> $5.93 | <b>Commit:</b> <code>ebad9c1ddd04d648a67eab8545bdf6754fe61e51</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @byteclimber,

**Verdict: Approve** - all prior suggestions addressed cleanly; no new blocking issues.
**Complexity: HIGH** - large diff (driven by tests), API surface signal.
_Changes: Evolves `retryOnQuota` from a single immediate retry to a bounded poll-retry loop (3 attempts / 3s backoff / 9s shared per-request deadline) and wraps three previously-unprotected write sites (6 files)._

<details>

- nit: The `attempt` counter semantics (poll-loop iterations only, excluding the initial `fn()` call) are correct but require tracing back through the production code to verify - a one-line inline comment on the metric assertion like `// attempt counts poll-loop retries only, not the initial call` would help - `test/support/serenity/dynamic-allocation-active.test.js`
- suggestion: Add an explicit assertion that `log.warn` fires with `'SERENITY_ALLOC retryOnQuota called without a callSite label'` when `callSite` is omitted - the warning path is exercised implicitly today but no test would break if it were removed - `test/support/serenity/dynamic-allocation-active.test.js`
- suggestion: Assert that `sleep(backoffMs)` is actually called between poll attempts (make `sleep` a sinon stub and verify it was invoked with the expected argument) to catch a regression where the await is accidentally removed - `test/support/serenity/dynamic-allocation-active.test.js`

</details>

- Shared-deadline sequential test added (locks the per-request budget invariant)
- `recordQuotaRetryOutcome` emission verified via esmock for both recovered and exhausted paths
- Low-cardinality doc note added to `Attempt` dimension JSDoc
- Stale "bounded top-up+retry" comment updated to "bounded poll-retry"
- `callSite === 'unknown'` runtime warning added with structured log context

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 45s | <b>Cost:</b> $5.41 | <b>Commit:</b> <code>2ea0420b0c40044f90c28839bc77ff2e3b3f014e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @byteclimber,

**Verdict: Approve** - all prior suggestions addressed cleanly; no new blocking issues.
**Complexity: HIGH** - large diff (driven by tests), API surface signal.
_Changes: Evolves `retryOnQuota` from a single immediate retry to a bounded poll-retry loop (3 attempts / 3s backoff / 9s shared per-request deadline) and wraps three previously-unprotected write sites (6 files)._

No new findings on this increment. The three test additions are well-designed:
- The `callSite` warning test uses a dedicated `spyLog` stub and asserts both the message and the structured context.
- The `sleep` test deliberately avoids `fastRetry()` so `backoffMs` stays at its real production default (3000ms) via the spread, making the assertion meaningful rather than tautological.
- The `attempt` comments clarify the off-by-one semantics for future readers without changing behavior.

- `log.warn` assertion for missing callSite now explicitly verified
- `sleep(backoffMs)` invocation between poll attempts now asserted via sinon stub with the real 3000ms default
- `attempt` counter semantics clarified with inline comments in both recovered and exhausted tests

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 6m 43s | <b>Cost:</b> $1.70 | <b>Commit:</b> <code>e7a690f8eb2081bd3461aae1ecfbf9f168fc089e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

aliciadriani:
## Review — `fix(serenity): span the ~9s Semrush write-enforcement lag in retryOnQuota`

### Summary
Upgrades `retryOnQuota` from a single immediate retry to a **bounded poll-retry loop** (default 3 attempts / 3s spacing) to span the live-verified ~9s lag between a JIT top-up succeeding and the Semrush gateway's write-enforcement checkpoint honoring it. Shares one time budget across all wrap sites in a request via a `requestDeadline` computed once at guard construction, extends wrapping to three previously-unprotected write sites (`createProject`, `createPromptsByIds`, `createOnePrompt`), and adds a `QuotaRetryOutcome` metric for rollout observability. Surgical, backward-compatible, and exceptionally well-documented. Verified locally: affected suites (52 tests) pass, `eslint` + `tsc` type-check clean.

### Must Fix (blocks merge)
None. No correctness, security, or data-loss issues found.

### Should Fix
Nothing blocking — one observation, defensible as-is:

- **Recovery-cycle observability gap** (`dynamic-allocation-active.js`, the poll loop). If a **non-metered** error surfaces *mid-loop* (recovery started, before success/exhaustion), it throws without emitting any `QuotaRetryOutcome`. An alarm built on `recovered + exhausted` as "total recovery cycles" would undercount cycles abandoned by a non-quota error. The path is unit-tested for correctness ("NON-metered error surfaces mid-loop"), so this is purely metric-completeness — consider an `abandoned` outcome if the rollout dashboard needs cycle-conservation.

### Nits (optional)
- **`MeteredQuotaClassifier` denominator now counts retries** (`errors.js`). `isMeteredQuota` emits its classifier metric as a side effect on every call, and the new loop calls it once per failed poll attempt — an exhausted 3-attempt cycle emits 4 samples instead of 1. Each *is* a genuine upstream 405, so it's defensible (and the doc comment already flags `retryOnQuota` as a live caller); just noting the metric's meaning shifted slightly from "distinct failures" toward "405 responses observed."
- **Shared deadline anchored at guard construction, not first write** (`dynamic-allocation-active.js`). The 9s budget starts at handler entry, so slow pre-write provisioning eats into the retry headroom for later writes in the same request. Intentional and documented (bounding the request's ~15s hot-path budget), but under a slow-provisioning create-market request a trailing `publishProject` could get little-to-no retry window. Behavior is correct; just noting the coupling.

### What's Good
- **Backward compatibility is airtight** — the flag-OFF branch returns the identical no-op guard (`retryOnQuota: (fn) => fn()`); zero extra transport calls, zero behavior change when the allocator is off.
- **The shared-deadline design is protected by a purpose-built regression test** — "deadline SHARED across two sequential calls on the SAME guard" specifically catches a plausible future regression (moving `requestDeadline` into the closure) that every other test would miss.
- **`ensure()` called exactly once per recovery, not per attempt** — correctly identifies that re-reading headroom every poll is a no-op that only adds per-child lock contention; the fix is the wait, not repeated top-ups. Explicitly asserted.
- **Excellent concurrency test hygiene** — the per-item recovery test keys stubs on prompt text rather than a flat call-index, avoiding nondeterminism under real `mapLimit` interleaving; the mixed-outcome test proves a non-retryable sibling error doesn't leak into a recovering item.
- **`callSite` is a closed vocabulary never derived from user input**, with a loud warn-on-missing guard — metric cardinality stays bounded by design.
- **Attempt-counting semantics are consistent** across code, metric emission, and every test assertion.

Clean, low-risk, mergeable. The remaining test-plan item (live re-verification with `SERENITY_DYNAMIC_ALLOCATION` on) is correctly left unchecked and should land before the flag flip, but doesn't block the code merge.

aliciadriani (approved):

MysticatBot (approved):
Hey @byteclimber,

**Verdict: Approve** - all prior suggestions addressed cleanly; no new blocking issues.
**Complexity: HIGH** - large diff (driven by tests), API surface signal.
_Changes: Adds `abandoned` outcome to `QuotaRetryOutcome` metric for mid-loop non-quota errors, completing the recovery-cycle accounting identity, plus a merge from main (6 files)._

No new findings on this increment. The `abandoned` outcome correctly fires inside the `!isMeteredQuota(e2)` branch (before the throw), emits exactly once per cycle terminated by a non-quota error, and does not fire when the initial `fn()` call throws a non-metered error (that path exits before the loop). The esmock-based metric emission tests and the handler-level concurrent/mixed-outcome tests are well-designed.

- `abandoned` outcome added per Alicia Adriani's observability gap feedback (recovered+exhausted+abandoned = total cycles)
- `log.warn` assertion for missing callSite now explicitly verified
- `sleep(backoffMs)` invocation between poll attempts now asserted via sinon stub with the real 3000ms default
- `attempt` counter semantics clarified with inline comments in both recovered and exhausted tests

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 32s | <b>Cost:</b> $5.14 | <b>Commit:</b> <code>eb84f2dce55067d6b0d82845b020f070d2942367</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

aliciadriani (changes requested):
Requesting changes on one observability item before the flag-flip rollout — the code itself is correct, backward-compatible, and well-tested (52 affected tests pass, lint + type-check clean). Full assessment below.

### Requested change (blocking this review)
- **Recovery-cycle observability gap** (`dynamic-allocation-active.js`, the poll loop). If a **non-metered** error surfaces *mid-loop* (recovery started, before success/exhaustion), it throws without emitting any `QuotaRetryOutcome`. Since this metric is the rollout-readiness signal, an alarm built on `recovered + exhausted` as "total recovery cycles started" will silently undercount cycles abandoned by a non-quota error mid-recovery. Please emit an `abandoned` (or equivalent) outcome on that path so the metric is cycle-conserving before the flag flip. The path is already unit-tested for correctness ("NON-metered error surfaces mid-loop"), so this is purely closing the metric.

### Nits (non-blocking)
- **`MeteredQuotaClassifier` denominator now counts retries** (`errors.js`). `isMeteredQuota` emits its classifier metric as a side effect on every call, and the new loop calls it once per failed poll attempt — an exhausted 3-attempt cycle emits 4 samples instead of 1. Each *is* a genuine upstream 405, so defensible; just noting the metric's meaning shifted from "distinct failures" toward "405 responses observed."
- **Shared deadline anchored at guard construction, not first write** (`dynamic-allocation-active.js`). The 9s budget starts at handler entry, so slow pre-write provisioning eats into retry headroom for later writes in the same request. Intentional and documented, but under a slow-provisioning create-market request a trailing `publishProject` could get little-to-no retry window. Behavior is correct; noting the coupling.

### What's Good
- **Backward compatibility is airtight** — the flag-OFF branch returns the identical no-op guard (`retryOnQuota: (fn) => fn()`); zero extra transport calls, zero behavior change when the allocator is off.
- **The shared-deadline design is protected by a purpose-built regression test** — "deadline SHARED across two sequential calls on the SAME guard" catches a plausible future regression (moving `requestDeadline` into the closure) that every other test would miss.
- **`ensure()` called exactly once per recovery, not per attempt** — correctly identifies that re-reading headroom every poll is a no-op that only adds per-child lock contention; the fix is the wait, not repeated top-ups. Explicitly asserted.
- **Excellent concurrency test hygiene** — the per-item recovery test keys stubs on prompt text rather than a flat call-index, avoiding nondeterminism under real `mapLimit` interleaving; the mixed-outcome test proves a non-retryable sibling error doesn't leak into a recovering item.
- **`callSite` is a closed vocabulary never derived from user input**, with a loud warn-on-missing guard — metric cardinality stays bounded by design.
- **Attempt-counting semantics are consistent** across code, metric emission, and every test assertion.

Aside from the observability item above, this is clean, low-risk work. The remaining test-plan item (live re-verification with `SERENITY_DYNAMIC_ALLOCATION` on) should also land before the flag flip.

aliciadriani:
**Retracting my earlier `CHANGES_REQUESTED` review — it was stale.**

I reviewed commit `e7a690f8` and posted a request to add an `abandoned` outcome to `QuotaRetryOutcome` for the mid-loop non-quota-error path. That change was in fact already implemented in `29609d0a` ("add abandoned outcome to QuotaRetryOutcome for mid-loop non-quota errors") and merged (`eb84f2dc`, 12:31 UTC) — before my review posted (13:17 UTC). MysticatBot verified it on the merge commit. My requested change was therefore already satisfied at merge time; there is nothing outstanding.

aliciadriani:
## Code review — `fix(serenity): span the ~9s Semrush write-enforcement lag in retryOnQuota`

This PR is already merged; posting for the record / any follow-up. Scope: the 6-file diff on `feat/serenity-flag-flip-readiness`._

### Summary
Turns `retryOnQuota` from a single immediate re-read+retry into a **bounded poll-retry** cycle intended to span the live-verified ~9s Semrush gateway write-enforcement lag, calls `ensure()` **once** per recovery (not per attempt) to avoid per-child lock contention, shares one per-request deadline across stacked wrap sites, wraps the three previously-unprotected metered writes (`createProject`, `createPromptsByIds`, `createOnePrompt`), and adds a `QuotaRetryOutcome` metric (`recovered`/`exhausted`/`abandoned`) for rollout observability. The design intent (shared deadline, once-only `ensure`, shape-based classifier, fail-loud required guard) is sound and the test coverage is genuinely thorough — deadline-vs-attempt-cap cutoff, shared-deadline-across-two-calls, per-item-keyed concurrency, mixed-outcome, `sleep` actually invoked, and each metric outcome are all pinned. The blast radius is limited: with the flag OFF `retryOnQuota` is a byte-for-byte no-op passthrough, and `SERENITY_DYNAMIC_ALLOCATION` is not yet flipped on in production. Overall quality is high. One timing concern below is worth confirming before the flag flips on.

### Should Fix

**1. The default schedule tops out at ~6s of post-recovery wait — short of the ~9s lag it's built to span, and the first poll re-lands immediately (the exact anti-pattern this PR is replacing).**

`dynamic-allocation-active.js` — the loop calls `fn()` at the top with **no leading sleep**; `sleep(backoffMs)` only runs *after* a failed poll. With the defaults (`maxAttempts=3`, `backoffMs=3000`) the poll attempts land at **t+0, t+3s, t+6s**, exhausting at t+6s:

- Poll attempt **1 fires immediately** after the recovery `ensure()` (which the code itself describes as a near-no-op read). That is functionally the same "single immediate retry, re-lands in the same window and doesn't help" that the PR description criticizes in the old code — so attempt 1 is largely wasted.
- Only attempts 2 and 3 add real waiting, and the **last retry is at ~t+6s**. If the enforcement lag is genuinely ~9s from the top-up, a write that needs the full window still exhausts ~3s early and 405s. The effective coverage is `(maxAttempts-1) * backoffMs = 6000ms`, not the `totalBudgetMs=9000` the deadline implies.
- For the earliest call site (`createProject`, right after its own `ensure({projects:1})`), the **attempt cap (3) binds before the 9s deadline is ever reached**, so `totalBudgetMs` is never the limiting factor there — meaning `createProject` is the most exposed path and gets only ~6s.

Two ways to actually reach 9s: bump the default to `maxAttempts=4` (attempts at 0/3/6/9), or add a leading `sleep(backoffMs)` *before* the first poll so the schedule becomes 3/6/9 and the wasted immediate re-land is removed. A leading sleep also better matches the PR's own thesis (the fix is "the WAIT between retries"). Worth confirming against the live sandbox numbers, since the test suite currently encodes the 6s behavior as expected (`sleep` asserted called once with 3000 on a recover-on-2nd-poll).

### Nits (optional)

**2. Shared deadline vs. large concurrent prompt batches.** In `prompts-subworkspace.js`, every item's `createOnePrompt` shares one guard and therefore one `requestDeadline`. For a large over-quota batch running under `mapLimit`, items processed after the ~9s budget is spent will see `now() >= requestDeadline` on their first poll and exhaust with no real retry, while earlier items recover. That's a defensible bound on request hot-path time (and the "don't compound to ~36s" rationale is good), but the failure mode — later items in a legitimately-recoverable batch failing purely because they were scheduled late — isn't covered by a test and may surface as puzzling partial failures once the flag is on. Consider a brief note in the doc comment.

**3. `callSite: 'unknown'` sentinel.** A caller that explicitly passes `{ callSite: 'unknown' }` would trip the "forgot a callSite" warning. Harmless (no caller does this), but a dedicated symbol/`undefined` check would be unambiguous.

**4. Comment density.** The block comments are excellent for capturing the hard-won live-debugging context, but a few (the construction-time deadline rationale, the once-only `ensure` rationale) restate each other across the file. Not worth churning a merged PR over — just future-editor ergonomics.

### What's Good
- **Shared per-request deadline computed once at construction** is the right seam to bound stacked wrap sites, and it's protected by a dedicated regression test that a naive per-call recompute would pass every *other* test in the file — exactly the sneaky regression worth pinning.
- **`ensure()` once per recovery, not per poll** correctly avoids per-child lock contention, with an explicit `transferWorkspaceResources` call-count-0 assertion proving the no-op read.
- **`abandoned` outcome** (non-quota error mid-loop) is a real observability improvement — without it a `recovered + exhausted` "total cycles" dashboard silently undercounts.
- **Fail-loud required `headroom` guard** in `generateAndAttachPrompts` (no optional-chaining) is the correct call — a future caller that forgets to thread it breaks loudly instead of silently skipping metering.
- **Per-item-keyed concurrency test** (Map keyed on prompt text, not a flat call index) is the right way to test recovery under real `mapLimit` interleaving and avoids the nondeterminism a shared counter would introduce.
- **Metric cardinality is explicitly reasoned about** — the `Attempt` raw-number dimension is documented as safe only while `maxAttempts` stays small, with a flag to revisit if it grows.

PR #2875, opened 2026-07-21
fix(serenity): reject empty edit text and modernize edit-path tests

byteclimber:
## Summary

Three coupled edit-path (`PATCH prompt`) fixes in the serenity handlers. Base is `feat/serenity-intent-classification`, not `main`.

### Fix 1 — Empty/whitespace text accepted on edit but rejected on create
`parseUpdatePromptBody` did `String(body.text)` with no trim and only rejected `text === undefined`, so a PATCH with `text: ""` or `"   "` passed validation, reached `renamePrompt`, and got classified/defaulted into a blank prompt. It now trims and rejects empty text with a `400 invalidRequest`, mirroring the create path (`normalizePromptInput`) — including the `|| ''` coercion so a falsy non-string (`null`, `0`, `false`) is treated as empty too, matching create exactly. `parseUpdatePromptBody` is shared with `handleUpdatePromptSubworkspace`, so this one fix covers both edit paths.

### Fix 2 — Redundant reclassification on no-op text edits (no behavior change, by design)
`handleUpdatePrompt` always reclassifies from the new text and rewrites tags, even when the PATCH does not change the text. Decision after review: **no behavior change**. The handler is not sent the old text, and the upstream provider has no GET-by-id, so it cannot detect a no-op text edit without an API contract change (client sends old text) — deliberately out of scope. Classification must also run before `renamePrompt` for failure-safety (a classify failure leaves the prompt intact), so `renamePrompt`'s `is_updated` result cannot gate it (it comes back too late). This PR documents the rationale with a code comment and adds a regression test proving an `is_updated: false` (no-op) rename still writes tags and publishes — guarding against a future "optimization" that would break the documented contract.

### Fix 3 — Modernize stale/broken edit-path tests
The `handleUpdatePrompt` block had **7** stale tests (not 3) that still stubbed the retired delete+create flow (`deletePromptsByIds`/`createPromptsByIds`); they failed at runtime (`transport.renamePrompt is not a function`) and one referenced an undefined `createErr`. Modernized them to the current `renamePrompt` + replace-mode `updatePromptTagsByIds` flow: in-place edit, drops-falsy/malformed tagIds, 409 collision (throw, no tag write, no publish), non-404 rename error (throw, no tag write), tag-write failure after successful rename (throw, publish never fires, partial-mutation warn logged), and 404 `promptNotFound` — plus a new negative test that a generic `Error` with `.status=404` does NOT map to `promptNotFound`. Also added the intent tag to the "recomputes the type tag on edit" assertion. The same stale-test class in `prompts-subworkspace.test.js` (3 edit-path twins) is fixed too, since this PR is the edit-path fix.

## Test evidence
- `prompts.test.js`: green (was 8 failing → 0).
- `prompts-subworkspace.test.js`: edit-path tests green (3 fixed).
- Full serenity suite: **1046 passing, 3 failing** (was 1031 passing / 14 failing on the base branch before this PR).
- `npm run lint` and `npm run docs:lint` clean on the changed files; the changed source (`prompts.js`) passes `type-check`.

## Pre-existing branch breakage (NOT introduced here, out of scope)
The base branch `feat/serenity-intent-classification` is already red from the LLMO-6190 dynamic-allocation work, in files this PR does not touch. CI will show these regardless:
- `eslint`: `src/support/serenity/handlers/prompts-subworkspace.js:251` — `headroom` no-shadow.
- `tsc` (`type-check`): `src/support/serenity/handlers/markets-subworkspace.js:217` (stale `@param headroom`) and `:537` (`Expected 5 arguments, but got 6` — a real call-site arg mismatch to `createHeadroomGuard`).
- 3 failing tests: `dynamic-allocation fronting — retryOnQuota wiring`, `prompts-subworkspace … dynamic-allocation ON …`, and `intent-classification.js — observability`.

These should be fixed on the feature branch (or by the LLMO-6190 owner) separately.

### DIFF
--- a/src/support/serenity/handlers/markets-subworkspace.js
+++ b/src/support/serenity/handlers/markets-subworkspace.js
@@ -224,7 +224,7 @@ function validateCreateBody(body) {
 async function generateAndAttachPrompts(transport, workspaceId, projectId, {
   domain, country, topicCap = 0, brandNames = [], provisioned, env,
   writeDeadline = computeWriteDeadline(),
-}, log) {
+}, log, headroom) {
   const raw = await transport.getBrandTopics(workspaceId, { domain, country });
   let topics = [];
   if (Array.isArray(raw)) {
@@ -312,6 +312,13 @@ async function generateAndAttachPrompts(transport, workspaceId, projectId, {
       });
     }
   }
+
+  // PROMPT metering seam (Rainer, live-verified LLMO-6190): front headroom sized
+  // on the real generated prompt count (`texts.size`) BEFORE the metered
+  // `createPromptsByIds` writes below — the disguised-quota 405 fires there, not
+  // at publish. No-op when the flag is OFF; NOT optional-chained, a caller that
+  // forgets to thread the guard must fail loud.
+  await headroom.ensure({ prompts: texts.size }, { includeDrafted: true });
 
   for (const { items, tagIds } of byTagSet.values()) {
     // eslint-disable-next-line no-await-in-loop
--- a/src/support/serenity/handlers/prompts-subworkspace.js
+++ b/src/support/serenity/handlers/prompts-subworkspace.js
@@ -251,7 +251,16 @@ export async function handleCreatePromptsSubworkspace(
     };
   }
 
-  const publishErrors = await publishAffected(transport, workspaceId, affectedProjectIds, log);
+  // Route each project's publish through the headroom guard's retryOnQuota (LLMO-6190 item 4):
+  // a disguised metered-405 gets ONE bounded top-up+retry per project before being recorded as a
+  // failure. No-op passthrough when the flag is OFF (the guard's retryOnQuota is a plain call).
+  const publishErrors = await publishAffected(
+    transport,
+    workspaceId,
+    affectedProjectIds,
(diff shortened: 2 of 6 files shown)


byteclimber:
Folded in the LLMO-6190 headroom base fix (was #2883, now closed) because the two are a **mutual** dependency — each PR's CI runs the full suite/lint, so neither could go green alone. This branch now also fixes:

- `markets-subworkspace.js` — `generateAndAttachPrompts` was missing the `headroom` param the caller/JSDoc require (TS8024 + TS2554); added it and fronted `headroom.ensure({ prompts: texts.size }, …)` before the metered write loop.
- `prompts-subworkspace.js` — ⚠️ **behavior change, please scrutinize:** removed the redundant pre-publish `ensure` (the `no-shadow` build failure + a second `getWorkspaceResources` read). LLMO-6190 meters once, before the write; the pre-write `ensure` already reserves the batch. If a second pre-publish top-up was intentional, push back here.
- `prompts-subworkspace.test.js` — corrected a stale positional call (options → 8th slot).

Verified locally: type-check clean, lint clean, 223 serenity handler tests passing.

PR #2877, opened 2026-07-21
feat(serenity): stamp prompt authorship metadata + DTO + sort/order (LLMO-6289)

aliciadriani:
## WP1 — stamp write paths + DTO + sort/order (LLMO-6289)

Stamps `created_at/by` + `updated_at/by` on every native prompt write, surfaces the four values on the list DTO, and adds server-side `sort`/`order` on the (now metadata-carrying) `by_tags` read. Built to the **v3 metadata contract** (which overrides the plan's `*-with-metadata`/PUT names).

### Caller identity (correctness-critical)
- `resolveCallerId(ctx)` = `authInfo.getProfile()?.user_id ?? sub`, fallback `unknown`, **capped at 100 chars** (the upstream `created_by`/`updated_by` CHECK bound).
- **Never** derived from the forwarded upstream bearer (post promise-token exchange its principal can differ from the caller).
- Resolved **once per request** in the controller and threaded into every write handler in both prompt twins, `createOnePrompt`, and `generateAndAttachPrompts` (markets-subworkspace + brand-provisioning).

### Stamping
- One shared helper: `buildCreateMetadata` (all four keys, `created_* = updated_*`) / `buildUpdateMetadata` (updated pair only). RFC 3339 UTC. No read-before-write; the stamp rides the same request as the mutation.
- **Tag edits stay two writes**: the v2 tag PUT is metadata-free; a text edit's stamp rides the combined `PATCH /v3 {id}` (name+metadata) that always runs in the in-place handler. No combined metadata-carrying tag write.
- Bulk deletes stamp nothing.

### Transport (v3 write surface over the regenerated client)
`createPromptsWithMetadata` (POST /v3 prompts), `patchPrompt` (combined PATCH /v3 {id}), `patchPromptMetadata` (single) + `patchPromptsMetadataBatch` (batch); `by_tags` list forwards `sort`/`order`.

### DTO + read + sort
`buildPromptDto` maps `metadata.*` → `createdAt/By/updatedAt/By`; `sort`/`order` allow-listed to `metadata.created_at` / `metadata.updated_at` and forwarded upstream. OpenAPI: `SerenityPrompt` authorship fields + 2 query params + contract test.

### ⚠️ Blocked on WP2 (regenerated PE client)
The v3 facade methods do **not** exist in the installed `@adobe/spacecat-shared-project-engine-client@1.14.0`; WP2 re-vendors the client to add them. The transport boundary type-checks now via a local `ProjectEngineTransportWithMetadata` typedef that WP2's generated types supersede on the dep bump. **Merge only after WP2 is released**, then bump the client dependency. `it-postgres` runs post-WP2-release only (it pulls the WP2 mock image by client version).

### Verification (node 24)
- `type-check` ✅ · `lint` ✅ · unit tests ✅ (15362 passing; the 2 timeout flakes in unrelated suites pass in isolation) · coverage 99.91% lines / 98.94% branches (gate 90%) · `docs:lint` ✅
- Diff ≈1149 LOC (17 files) — under the ~2000 MysticatBot ceiling.


test/support/serenity/handlers/prompts-subworkspace.test.js
```diff
@@ -103,6 +119,10 @@ describe('prompts-subworkspace handlers', () => {
           path: [{ id: TAG_IDS.categoryRoot, name: 'category' }],
```
MysticatBot, line 126: issue (blocking): `handleListPromptsSubworkspace` calls `resolveSort(query)` and threads `sort`/`order` into the transport call, but this test file has no dedicated tests exercising that forwarding path. The flat-mode twin (`handleListPrompts` in `prompts.test.js`) has tests that verify valid sort/order is forwarded and that invalid sort throws 400 before any upstream call. Since the two code paths resolve the project differently (DB lookup vs live listing), a regression in either is independently possible.

Fix: add at minimum two tests here:
1. A valid sort/order is forwarded to `transport.listPromptsByTags`.
2. An invalid sort throws 400 before `listPromptsByTags` is called.

test/support/serenity/handlers/prompts-subworkspace.test.js
```diff
@@ -46,11 +46,11 @@ function makeTransport(overrides = {}) {
     listProjects: sinon.stub().resolves({ items: [proj()] }),
     listProjectTags: makeListProjectTagsStub(),
     listPromptsByTags: sinon.stub().resolves({ items: [] }),
-    createPromptsByIds: sinon.stub().resolves({
+    createPromptsWithMetadata: sinon.stub().resolves({
```
MysticatBot, line 51: issue (blocking): The `patchPrompt` stub is defined as `(ws, pid, promptId, newName) => ...` but production now passes an object `{ name, metadata }` as the 4th argument (not a string). The stub resolves `{ name: newName }` which will be `{ name: { name: 'next', metadata: {...} } }`. No current assertion reads this, but it is semantically wrong and will mask bugs if a future test asserts on the resolved value.

Fix: update to destructure the body:
```js
patchPrompt: sinon.stub().callsFake(
  (ws, pid, promptId, body) => Promise.resolve(
    { id: promptId, name: body?.name ?? '', is_updated: true },
  ),
),
```

MysticatBot (changes requested):
Hey @aliciadriani,

**Verdict: Request changes** - two test gaps need closing before this ships.
**Complexity: HIGH** - large diff; API surface (OpenAPI + transport layer).
_Changes: Stamps prompt authorship metadata (created/updated at/by) on every write path, surfaces the fields on the list DTO, and adds server-side sort/order on the by_tags read (17 files)._
Note: Recommend a human read before merge - this change modifies a shared contract (OpenAPI schema). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge

1. [Important] Missing sort/order tests for subworkspace list handler - `test/support/serenity/handlers/prompts-subworkspace.test.js` (details inline)
2. [Important] `patchPrompt` stub signature mismatch in subworkspace test transport - `test/support/serenity/handlers/prompts-subworkspace.test.js:49` (details inline)

<details>

- nit: Tests pass `undefined` as callerId, so metadata carries `created_by: undefined` instead of the `'unknown'` sentinel that would occur in real execution (where `resolveCallerId` ensures a floor). Consider passing a real caller id in at least one integration-level test per handler to prove the full path. - `test/support/serenity/handlers/prompts.test.js`
- suggestion: The WP2 merge-gate is documented only in the PR body. Consider enforcing it with a CI label check or merge-queue gate so accidental merge before the dep bump is structurally impossible across team rotation.

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 10s | <b>Cost:</b> $8.27 | <b>Commit:</b> <code>c5cdd79a29735daaed8c1c239cbf88d1815cc1d7</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/serenity/handlers/prompts.js
```diff
@@ -48,6 +48,109 @@ export const BULK_CREATE_CONCURRENCY = 8;
...
+ * `user_id` is preferred; `sub` is the fallback (both IMS profile claims). A
+ * missing/blank identity becomes the literal `unknown` (the spec's NULL-author
+ * sentinel, resolved to a display name downstream). Capped at
+ * {@link CALLER_ID_MAX_LENGTH} so a pathological claim can never trip the
+ * upstream length CHECK (which would 400 the write / roll a batch back).
+ *
+ * @param {object} ctx - the controller request context.
+ * @returns {string} the caller id, `unknown` when unresolved, ≤100 chars.
+ */
+export function resolveCallerId(ctx) {
```
aliciadriani, line 95: CORRECTNESS-CRITICAL, confirmed correct: `resolveCallerId` reads `ctx.attributes.authInfo.getProfile()?.user_id ?? sub` — the auth PROFILE, never the forwarded upstream bearer (whose principal can differ after the promise-token exchange). `unknown` fallback for a missing/blank identity, and `.slice(0, 100)` caps it below the upstream ≤100 CHECK so no write path can 400/roll-a-batch-back on a pathological claim. Single resolution point.

src/support/serenity/handlers/prompts.js
```diff
@@ -48,6 +48,109 @@ export const BULK_CREATE_CONCURRENCY = 8;
...
+ * Builds the `metadata` merge-patch payload for a CREATE: all four keys, with
+ * `created_*` and `updated_*` set to the SAME instant/caller (a create is its
+ * own first edit). Timestamps are RFC 3339 UTC (`new Date().toISOString()`).
+ *
+ * ONE stamping helper, shared by BOTH prompt twins (flat + subworkspace) and the
+ * AI-generation create — the metadata logic is never written twice.
+ *
+ * @param {string} callerId - already resolved + capped by {@link resolveCallerId}.
+ */
+export function buildCreateMetadata(callerId) {
```
aliciadriani, line 112: One stamping helper, not duplicated: `buildCreateMetadata` sets all four keys (`created_* = updated_* = now/caller`), `buildUpdateMetadata` sets only the `updated_*` pair so an edit's RFC-7396 merge keeps `created_*` untouched with no read-before-write. RFC-3339 UTC. Shared by both twins + the AI-generation create.

aliciadriani (commented):
Reviewed WP1 (LLMO-6289): stamp authorship metadata on every write + caller identity + DTO/sort, to the delivered v3 contract. Base=main, rebased + MERGEABLE → requesting MysticatBot. The last 35-track API build. **Review-clean — no Must-Fix, no Should-Fix.**

**Caller identity (correctness-critical) — CONFIRMED.** `resolveCallerId(ctx)` reads `ctx.attributes.authInfo.getProfile()?.user_id ?? sub` — the AUTH PROFILE, never the forwarded upstream bearer (whose principal can differ after the promise-token exchange). Missing/blank → the literal `unknown`; result `.slice(0, CALLER_ID_MAX_LENGTH=100)` so a pathological claim can't trip the upstream ≤100 CHECK. Resolved ONCE per request/batch (POST prompts once, PATCH once, the activate loop once for the whole batch — not re-resolved per market/prompt) and threaded into: both create twins, `createOnePrompt`, `generateAndAttachPrompts`, `handleCreateMarketSubworkspace`, and `provisionBrandSubworkspace`. Verified it never touches a token/bearer.

**One stamping helper — CONFIRMED.** `buildCreateMetadata(callerId)` (all four keys, `created_* = updated_* = now/caller`) and `buildUpdateMetadata(callerId)` (only the `updated_*` pair, so RFC-7396 merge keeps `created_*` with no read-before-write). RFC-3339 UTC via `toISOString()`. Shared by both twins + the AI-generation create — never duplicated.

**Separate tag/metadata writes (v3) — CONFIRMED, no combined metadata-carrying tag write.** A text edit is the combined `PATCH /v3 .../{id}` (`patchPrompt` — `{name, metadata: buildUpdateMetadata}`) replacing the v2 `rename`, so the `updated_*` stamp rides the SAME request as the text mutation (no read-before-write); it runs FIRST (the one op that can 404/409). The subsequent v2 batch tag PUT (`updatePromptTagsByIds`) carries `{id, references, replace:true}` only — NO metadata. So the edit's stamp is covered entirely by the combined PATCH, `created_*` is never sent, and there is no combined tag+metadata write anywhere. Creates stamp via `createPromptsWithMetadata` (`{name, metadata}` items) on the same write that creates the prompt.

**DTO / sort / OpenAPI — CONFIRMED.** `buildPromptDto` maps `metadata.created_at/created_by/updated_at/updated_by` → `createdAt/createdBy/updatedAt/updatedBy`, null when the item has no metadata (stable shape for the UI em-dash). `resolveSort` allow-lists `sort` to `metadata.created_at`/`metadata.updated_at` (else 400 — and it notes the value is forwarded verbatim upstream, so the allow-list is the injection guard, not just hygiene), `order`∈{asc,desc} default `desc` (else 400). `sort`/`order` are sent upstream ONLY when a sort is requested, so an unsorted read is byte-for-byte the legacy call. OpenAPI: `SerenityPrompt` gains the four nullable fields (createdBy/updatedBy `maxLength:100`), the two query params are enum-restricted, and the contract test is updated.

**Rebase-conflict resolution (LLMO-6190 drift) — CONFIRMED sound.** In `generateAndAttachPrompts` the v3 metadata create is correctly nested inside `headroom.retryOnQuota(() => transport.createPromptsWithMetadata(ws, pid, items.map(name => ({name, metadata})), [...standardIds, typeId]), {callSite:'createPromptsWithMetadata'})` — the retry thunk wraps the v3 write, `callerId` is threaded through, `buildCreateMetadata` is called once per batch. In `handleCreatePromptsSubworkspace`, `createOnePrompt(..., typed, callerId)` sits inside `headroom.retryOnQuota(() => ..., {callSite:'createOnePrompt'})`. No WP1 substance is lost and the retry wrapper still wraps correctly; the fronting test is updated to `items[0].name` + the `callerId` arg.

**Not dinged (per the merge plan):** the v3 facade methods (`createPromptsWithMetadata`/`patchPrompt`/…) throw against the installed PE-client 1.14.0 — the merge is gated on the WP2 release + dep bump, and the local `ProjectEngineTransportWithMetadata` typedef is an explicit stopgap cast at the one transport boundary (WP2's regenerated types supersede it); `it-postgres` runs post-WP2-release; the draft→live publish-carries-metadata assumption is unverifiable in-fleet (WP5 G1). All intended — no finding.

Clean, well-tested (15393 passing, 99.91%/98.96% cov). Requesting MysticatBot (base=main).

src/support/serenity/handlers/prompts.js
```diff
@@ -725,7 +861,14 @@ export async function handleUpdatePrompt(
   });
 
   try {
-    await transport.renamePrompt(semrushWorkspaceId, projectId, semrushPromptId, nextText);
+    // Combined v3 write: sets the text (`name`) and stamps the `updated_*`
+    // metadata pair in one request (replaces the v2 `rename`). Same refusal
+    // contract as rename — 404 (unknown id) → promptNotFound, 409 (text
+    // collides with a sibling) → thrown for the controller's `conflict` mapping.
+    await transport.patchPrompt(semrushWorkspaceId, projectId, semrushPromptId, {
```
aliciadriani, line 1301: The combined v3 `patchPrompt` (`{name, metadata: buildUpdateMetadata(callerId)}`) replaces the v2 `rename`, so the text edit and the `updated_*` stamp land in ONE request — runs first (the one op that can 404/409). The subsequent v2 tag PUT (`updatePromptTagsByIds`) carries references only, NO metadata: the edit's stamp is fully covered here, `created_*` is never sent, and there is no combined tag+metadata write. Correct per the v3 contract.

src/support/serenity/handlers/prompts.js
```diff
@@ -48,6 +48,109 @@ export const BULK_CREATE_CONCURRENCY = 8;
...
+ * Validates + normalizes the `sort` / `order` list query params against the
+ * {@link SORTABLE_METADATA_FIELDS} allow-list. Returns `{}` when neither is
+ * supplied (the legacy unsorted read); `{ sort, order }` when a valid sort is
+ * requested (order defaults to `desc` — newest first for "Last modified");
+ * throws 400 for an unknown sort field or order.
+ *
+ * @param {object} query
+ * @returns {{ sort?: string, order?: string }}
+ */
+export function resolveSort(query) {
```
aliciadriani, line 153: `resolveSort` allow-lists `sort` to `metadata.created_at`/`metadata.updated_at` (else 400) and `order`∈{asc,desc} (default desc, else 400). Since the value is forwarded verbatim upstream, the allow-list is the injection guard, not just input hygiene — good that it's a hard 400 rather than a silent drop. Returns `{}` when unspecified so an unsorted read is unchanged.

src/controllers/serenity.js
```diff
@@ -1227,6 +1242,10 @@ function SerenityController(context, log, env) {
           dynamicAllocation: dynamicAllocationEnabled(ctx),
         },
       );
+      // Caller identity for the created_* stamp on generated prompts, resolved
+      // ONCE for the whole activate batch (LLMO-6289) — from the auth profile,
+      // never the forwarded upstream bearer.
+      const callerId = resolveCallerId(ctx);
```
aliciadriani, line 1470: `resolveCallerId(ctx)` resolved ONCE for the whole activate batch here and threaded into the per-market loop — correctly NOT re-resolved per market. Same once-per-request discipline as the POST/PATCH handlers. Good.

src/support/serenity/handlers/markets-subworkspace.js
```diff
@@ -288,8 +295,13 @@ async function generateAndAttachPrompts(transport, workspaceId, projectId, {
     // top-up) — route through `headroom.retryOnQuota` (no-op passthrough when the flag is OFF).
     // eslint-disable-next-line no-await-in-loop
     await headroom.retryOnQuota(
-      () => transport.createPromptsByIds(workspaceId, projectId, items, [...standardIds, typeId]),
-      { callSite: 'createPromptsByIds' },
+      () => transport.createPromptsWithMetadata(
```
aliciadriani, line 450: Rebase-conflict resolution confirmed sound: WP1's v3 metadata create is nested inside the upstream `headroom.retryOnQuota(() => transport.createPromptsWithMetadata(...), {callSite:'createPromptsWithMetadata'})` (the LLMO-6190 drift), `callerId` threaded, `buildCreateMetadata` called once per batch. The retry thunk wraps the v3 write correctly — no WP1 substance lost.

src/support/serenity/handlers/prompts-subworkspace.js
```diff
@@ -184,7 +192,7 @@ export async function handleCreatePromptsSubworkspace(
       // no-op passthrough when the flag is OFF) so each item recovers independently; `mapLimit`'s
       // own per-item try/catch below still isolates a surviving failure to this one item.
       const semrushPromptId = await headroom.retryOnQuota(
-        () => createOnePrompt(transport, workspaceId, projectId, typed),
+        () => createOnePrompt(transport, workspaceId, projectId, typed, callerId),
```
aliciadriani, line 275: The subworkspace twin: `createOnePrompt(transport, ws, pid, typed, callerId)` sits inside `headroom.retryOnQuota(() => ..., {callSite:'createOnePrompt'})` — callerId threaded, retry wrapper intact. Matches the flat-mode twin. Rebase merge is clean.

src/support/serenity/rest-transport.js
```diff
@@ -39,6 +39,29 @@ import { ErrorWithStatusCode } from '../utils.js';
...
+ * intersection so the single transport boundary type-checks under `// @ts-check`
+ * against the currently-installed client. Method names/shapes track the routes
+ * the revised ADR pins (spec §13 item 1):
+ *   - createPromptsWithMetadata → POST  /v3 prompts        { items:[{name,metadata}], tag_ids }
+ *   - patchPrompt               → PATCH /v3 prompts/{id}          { name?, metadata? }
+ *   - patchPromptMetadata       → PATCH /v3 prompts/{id}/metadata { <merge-patch> }
+ *   - patchPromptsMetadataBatch → PATCH /v3 prompts/metadata      { items:[{id,metadata}] }
+ * WP2's generated types supersede this typedef verbatim on the dep bump.
+ *
+ * @typedef {ReturnType<typeof createSerenityProjectEngineTransport> & {
```
aliciadriani, line 115: The `ProjectEngineTransportWithMetadata` typedef + the single boundary cast is the intended stopgap: the installed PE-client 1.14.0 lacks these v3 methods, so this lets the transport type-check under `// @ts-check` until the WP2 release + dep bump (which regenerates and supersedes this verbatim). This is the documented merge gate, not a finding — flagging only to confirm I understand it's deliberate.

test/support/serenity/handlers/prompts-subworkspace.test.js
aliciadriani: Fixed in 927e8f315f9a4c9ce35fa65d3e7c85fe9485008b. Added dedicated sort/order tests for `handleListPromptsSubworkspace` mirroring the flat-mode twin: valid `sort`/`order` forwarded to `transport.listPromptsByTags`, invalid sort → 400 before any upstream call, invalid order → 400 before any upstream call, and an unsorted read forwards neither. Also added a `metadata`→DTO mapping test for this twin.

test/support/serenity/handlers/prompts-subworkspace.test.js
aliciadriani: Fixed in 927e8f315f9a4c9ce35fa65d3e7c85fe9485008b. The `patchPrompt` fake now takes the combined v3 body as its 4th arg — `(ws, pid, promptId, body) => Promise.resolve({ id: promptId, name: body?.name ?? '', is_updated: true })` — matching the production signature. Added an update test that asserts the handler passes `{ name, metadata }` (name + the stamped `updated_*` pair, no `created_*`) through, so a future assertion on the resolved value can't be masked.

aliciadriani (commented):
Re-check at `927e8f31` — MysticatBot's two blocking test findings + the nit are addressed, and the fixes are sound. This is a TEST-ONLY commit (two test files; no product-code change from the WP1 diff I already reviewed clean). WP1 is review-clean from the `/review-pr` side; type-check + build green.

**1. Subworkspace sort/order tests — CONFIRMED.** `prompts-subworkspace.test.js` gains the twin-mirrored set: a valid `sort`/`order` is forwarded to `transport.listPromptsByTags` (`body.sort==='metadata.updated_at'`, `body.order==='asc'`); a sort outside the allow-list (`text`) rejects `/sort must be one of/` with `listPromptsByTags` NOT called; an invalid order (`sideways`) rejects `/order must be one of/`, not called; an unsorted read forwards neither (`body.sort`/`order` undefined); plus a `metadata`→DTO mapping test proving createdAt/By/updatedAt/By map for this twin. Matches the flat-mode coverage in `prompts.test.js`, and the subworkspace twin resolves its project via the live listing (not the DB) so its forwarding path is independently exercised.

**2. `patchPrompt` stub fixed — CONFIRMED (a real test-correctness bug).** The `makeTransport` stub was `(ws, pid, promptId, newName) => ({ name: newName })`, which would have bound the whole combined body `{name, metadata}` object to `newName` — masking a signature mismatch. It is now `(ws, pid, promptId, body) => ({ id: promptId, name: body?.name ?? '', is_updated: true })`, reading `body.name` off the real combined v3 shape. Good catch to fix — the stub now can't hide a wrong call shape.

**3. Nit / real-callerId — CONFIRMED.** Two new tests use a REAL resolved id (`caller-42`) rather than the `undefined` default: create asserts `createPromptsWithMetadata` is called with the item carrying `created_*` = caller-42; update asserts the combined `patchPrompt` stamps `updated_*` = caller-42 AND that its `body.metadata` has NO `created_at`/`created_by` (merge-patch keeps them) AND the tag PUT's item has NO `metadata` property (metadata-free tag write). That directly locks the create-stamps-created / edit-stamps-updated-only / metadata-free-tag-PUT contract.

The `dynamic-allocation-fronting.test.js` edits are the mechanical alignment of the fronting stubs to the v3 signature (`createPromptsByIds`→`createPromptsWithMetadata`, `texts[0]`→`items[0].name`, the added `undefined // callerId` positional arg) — no change to the retryOnQuota wiring assertions. Full `npm test` 15400 passing.

Merge remains gated on the WP2 release + PE-client dep bump (intended). From `/review-pr` this is clean; re-triggering MysticatBot for a fresh-head pass (its recorded CHANGES_REQUESTED is against the pre-fix commit).

MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - prior findings addressed, no new blocking issues found.
**Complexity: HIGH** - large diff; API surface (OpenAPI + transport layer).
_Changes: Stamps prompt authorship metadata (created/updated at/by) on every write path, surfaces the fields on the list DTO, and adds server-side sort/order on the by_tags read (17 files)._
Note: Recommend a human read before merge - this change modifies a shared contract (OpenAPI schema). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: `resolveSort` returns `{}` which destructures as `sort: undefined, order: undefined` in the handler object literals. The transport's `if (body?.sort)` guard catches it, but `...(sort && { sort, order })` at the handler call sites would make the "byte-for-byte legacy call" claim precise at every boundary, not just the transport. - `src/support/serenity/handlers/prompts-subworkspace.js:104`
- nit: `handleCreatePromptsSubworkspace` calls `buildCreateMetadata(callerId)` per item (inside `createOnePrompt`), so items in one batch get slightly different timestamps, while `generateAndAttachPrompts` shares one metadata object across the group. Both are defensible (actual creation instant vs batch instant), but the inconsistency is worth a one-line comment explaining the intentional difference. - `src/support/serenity/handlers/prompts-subworkspace.js:195`
- suggestion: The `unknown` sentinel as callerId fallback is by-design per the spec, but consider documenting explicitly (in the ADR or a code comment at the controller boundary) whether `unknown`-attributed writes are acceptable policy or whether they should eventually be rejected, so future readers know the fallback is intentional, not overlooked. - `src/support/serenity/handlers/prompts.js:82`

</details>

- Sort/order tests for subworkspace list handler now added (commit 927e8f3)
- `patchPrompt` stub signature updated to match production's combined v3 body shape (commit 927e8f3)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 30s | <b>Cost:</b> $8.69 | <b>Commit:</b> <code>927e8f315f9a4c9ce35fa65d3e7c85fe9485008b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/serenity/handlers/prompts-subworkspace.js
```diff
@@ -97,6 +102,12 @@ export async function handleListPromptsSubworkspace(transport, workspaceId, quer
     page,
     limit,
     search,
+    // `resolveSort` returns `{}` when unsorted; spread conditionally so the
+    // upstream body carries no `sort`/`order` keys at all — byte-for-byte the
+    // legacy unsorted call at THIS boundary too, not merely relying on the
+    // transport's `if (body?.sort)` guard downstream. Kept in lockstep with the
+    // flat-mode twin (handlers/prompts.js).
+    ...(sort && { sort, order }),
```
aliciadriani: **Fixed (nit #1):** `sort`/`order` are now spread conditionally (`...(sort && { sort, order })`), so when `resolveSort` returns `{}` the upstream body carries no `sort`/`order` keys at all — byte-for-byte the legacy unsorted call at this handler boundary, not only at the transport's `if (body?.sort)` guard. Applied to the flat-mode twin too (prompts.js:338) to keep them in lockstep.

aliciadriani (commented):
Addressed all 3 non-blocking items from the last review in a094339. Inline notes on each fixed line below.

src/support/serenity/handlers/prompts.js
```diff
@@ -207,6 +331,11 @@ export async function handleListPrompts(
       page,
       limit,
       search,
+      // See the subworkspace twin (prompts-subworkspace.js): `resolveSort`
+      // returns `{}` when unsorted, so spread conditionally to omit the
+      // `sort`/`order` keys entirely and keep the upstream body byte-for-byte
+      // the legacy unsorted call at this boundary too.
+      ...(sort && { sort, order }),
```
aliciadriani: **Fixed (nit #1, twin):** same conditional spread as the subworkspace handler — kept in lockstep per the code comment's promise.

src/support/serenity/handlers/prompts-subworkspace.js
```diff
@@ -183,8 +195,15 @@ export async function handleCreatePromptsSubworkspace(
       // write-enforcement lag after a JIT top-up) — route it through `headroom.retryOnQuota` (a
       // no-op passthrough when the flag is OFF) so each item recovers independently; `mapLimit`'s
       // own per-item try/catch below still isolates a surviving failure to this one item.
+      // Intentional (not an inconsistency to fix): each item stamps its OWN
```
aliciadriani, line 267: **Fixed (nit #2):** added a comment documenting that the per-item timestamping is intentional. Each bulk item stamps its own creation instant here (`buildCreateMetadata` runs per prompt inside `createOnePrompt`), unlike `generateAndAttachPrompts` which shares one metadata object across its group. Both are defensible; per-item precision is preferred on the direct create path.

src/support/serenity/handlers/prompts.js
```diff
@@ -48,6 +48,117 @@ export const BULK_CREATE_CONCURRENCY = 8;
...
+ * from `authInfo.getProfile()` — NEVER from the bearer forwarded upstream, whose
+ * principal can differ from the caller after the promise-token exchange.
+ *
+ * `user_id` is preferred; `sub` is the fallback (both IMS profile claims). A
+ * missing/blank identity becomes the literal `unknown` (the spec's NULL-author
+ * sentinel, resolved to a display name downstream). Capped at
+ * {@link CALLER_ID_MAX_LENGTH} so a pathological claim can never trip the
+ * upstream length CHECK (which would 400 the write / roll a batch back).
+ *
+ * POLICY (LLMO-6289 — intentional, not an oversight): an `unknown`-attributed
```
aliciadriani, line 84: **Fixed (suggestion #3):** documented the `unknown`-callerId policy explicitly at the controller boundary. `unknown`-attributed writes are accepted by design (authorship is best-effort provenance, not an authz gate — the caller is already authenticated upstream); if they should ever be rejected, that is a deliberate contract change to make here, not a silent assumption.

MysticatBot (changes requested):
Hey @aliciadriani,

**Verdict: Request changes** - accidental `node_modules` symlink must be removed before merge.
**Complexity: HIGH** - large diff; API surface (OpenAPI + transport layer).
_Changes: Stamps prompt authorship metadata (created/updated at/by) on every write path, surfaces the fields on the list DTO, and adds server-side sort/order on the by_tags read (17 files)._
Note: Recommend a human read before merge - this change modifies a shared contract (OpenAPI schema). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge

1. [Critical] Accidental `node_modules` symlink committed to repo root - will break CI and local checkouts for anyone without the author's exact workspace layout - `node_modules` (details inline)

<details>

- nit: The conditional spread `...(sort && { sort, order })` relies on `sort` being truthy to include `order`. Today this is safe because `resolveSort` returns `{}` when unsorted, but the truthiness guard couples the call sites to `resolveSort`'s exact return shape without a contract enforcing it. A ternary `...(sort ? { sort, order } : {})` would be more explicit. - `src/support/serenity/handlers/prompts-subworkspace.js:107`
- nit: The 5-line comment on the conditional spread is disproportionate to the one-liner it annotates. The "byte-for-byte the legacy unsorted call" framing will read as stale once this becomes the established pattern. Consider trimming to one line: `// Omit sort/order keys when unsorted (lockstep with twin file).` - `src/support/serenity/handlers/prompts-subworkspace.js:105`

</details>

- Sort/order now spread conditionally so unsorted reads carry no `sort`/`order` keys at the handler boundary (commit a094339)
- Per-item vs batch-instant timestamp difference documented with intentionality comment (commit a094339)
- `unknown` sentinel policy documented explicitly at `resolveCallerId` (commit a094339)

node_modules
```diff
@@ -0,0 +1 @@
+../wp-o6-origin-contract/node_modules
```
MysticatBot: issue (blocking): A symlink `node_modules -> ../wp-o6-origin-contract/node_modules` was committed to the repo root. This is a local workspace artifact that will break CI and any contributor who does not share this exact parent-directory layout.

On checkout, git creates this symlink; if the target does not exist, `node_modules` becomes dangling and `npm install` / `npm ci` will fail or behave unpredictably. Despite `.gitignore` listing `node_modules`, the ignore rule only applies to untracked entries - once force-added or committed, git tracks it regardless.

Fix: remove it from the branch before merge:
```bash
git rm node_modules
git commit -m "chore: remove accidental node_modules symlink"
```

src/support/serenity/handlers/prompts-subworkspace.js
```diff
@@ -97,6 +102,8 @@ export async function handleListPromptsSubworkspace(transport, workspaceId, quer
     page,
     limit,
     search,
+    // Omit sort/order keys when unsorted (lockstep with twin file prompts.js).
+    ...(sort ? { sort, order } : {}),
```
aliciadriani, line 116: **Fixed (both nits):** switched to the explicit ternary `...(sort ? { sort, order } : {})` so it no longer leans on `order` riding along with a truthy `sort`, and trimmed the 5-line comment to a single line (`// Omit sort/order keys when unsorted (lockstep with twin file prompts.js).`).

src/support/serenity/handlers/prompts.js
```diff
@@ -207,6 +331,8 @@ export async function handleListPrompts(
       page,
       limit,
       search,
+      // Omit sort/order keys when unsorted (lockstep with twin file prompts-subworkspace.js).
+      ...(sort ? { sort, order } : {}),
```
aliciadriani, line 366: **Fixed (twin):** same ternary + one-line comment, kept in lockstep with the subworkspace handler.

aliciadriani (commented):
Addressed round 2 in a962166 (amended a094339, force-pushed):

- **[Critical] Removed the accidental `node_modules` symlink.** It was staged by a `git add -A` while I was running lint/tests locally; `.gitignore`'s `node_modules/` (trailing slash) matches the directory but not a symlink of that name, so it slipped in. The symlink is gone from the tree (verified `git ls-files` shows it untracked) and the local copy is kept only in my working dir for tooling.
- **nit: ternary instead of `&&`** — see inline.
- **nit: trimmed the oversized comment** — see inline.

MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - prior findings addressed, no new blocking issues found.
**Complexity: HIGH** - large diff; API surface (OpenAPI + transport layer).
_Changes: Stamps prompt authorship metadata (created/updated at/by) on every write path, surfaces the fields on the list DTO, and adds server-side sort/order on the by_tags read (17 files)._
Note: Recommend a human read before merge - this change modifies a shared contract (OpenAPI schema). The bot review is a complement to, not a replacement for, a human read here.

<details>

- suggestion: `buildCreateMetadata`/`buildUpdateMetadata` accept `callerId` without a defensive fallback. All production callers go through `resolveCallerId` (which always returns a string), but the exported helpers themselves would silently produce `created_by: undefined` if a future caller omits the resolution step. A one-liner `const id = callerId || 'unknown'` in each builder would give the same safety net that `handleCreateMarketSubworkspace`'s `callerId = 'unknown'` default provides. - `src/support/serenity/handlers/prompts.js:471`
- nit: Several test `it(...)` descriptions still reference "rename" (e.g. "upstream rename 404s") while the operation is now `patchPrompt`. The test bodies were updated correctly but the description strings are stale. - `test/support/serenity/handlers/prompts-subworkspace.test.js:448`
- suggestion: Consider adding one test exercising a partially-populated upstream metadata object (e.g. only `created_at`/`created_by` present, `updated_*` absent) to lock the `??` fallback behavior in `buildPromptDto` against future refactors. - `src/support/serenity/handlers/prompts.js:248`
- suggestion: The merge gate (WP2 PE-client dep bump) is enforced socially. A semver-range assertion in a test file or CI step that fails until the installed client carries the v3 facade methods would prevent an accidental premature merge across team rotation.

</details>

- Accidental `node_modules` symlink removed (force-push a962166)
- Sort/order now spread conditionally so unsorted reads carry no keys at the handler boundary
- Per-item vs batch-instant timestamp difference documented with intentionality comment
- `unknown` sentinel policy documented explicitly at `resolveCallerId`
- `patchPrompt` stub signature updated to match production's combined v3 body shape
- Conditional-spread ternary adopted per prior nit
- Oversized annotation comment trimmed

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 37s | <b>Cost:</b> $8.19 | <b>Commit:</b> <code>a9621661eba06b305ebbaf8e712526579b81245c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/serenity/handlers/prompts.js
```diff
@@ -48,6 +48,124 @@ export const BULK_CREATE_CONCURRENCY = 8;
...
+ *
+ * @param {string} callerId - already resolved + capped by {@link resolveCallerId}.
+ */
+export function buildCreateMetadata(callerId) {
+  const now = new Date().toISOString();
+  // Defensive floor mirroring resolveCallerId's `unknown` sentinel: every
+  // production path already passes a resolved+capped id, but a future direct
+  // caller that skips resolveCallerId must never stamp `created_by: undefined`
+  // (which the upstream metadata column would reject / store as a null author).
+  const id = callerId || 'unknown';
```
aliciadriani, line 118: **Fixed (suggestion #1):** `buildCreateMetadata`/`buildUpdateMetadata` now floor `callerId` to the `unknown` sentinel (`callerId || 'unknown'`) — the same safety net `handleCreateMarketSubworkspace`'s `callerId = 'unknown'` default provides. Every production path still threads a `resolveCallerId`-resolved string, but a future direct caller can no longer silently stamp `created_by`/`updated_by: undefined`. The test matcher helpers (`createItemMatch`/`patchTextMatch`) floor identically, so an omitted-callerId test now asserts the exact `unknown` value real execution stamps — which also closes the earlier round's `undefined`-in-tests nit. Fixed in 7fd696ce.

test/support/serenity/handlers/prompts-subworkspace.test.js
```diff
@@ -330,9 +448,9 @@ describe('prompts-subworkspace handlers', () => {
       expect(result.body.error).to.equal('marketNotFound');
     });
 
-    it('404s promptNotFound when the upstream rename 404s (no tag write)', async () => {
+    it('404s promptNotFound when the upstream patchPrompt 404s (no tag write)', async () => {
```
aliciadriani, line 536: **Fixed (nit #2):** refreshed the stale `rename` wording in the prompt-update `it(...)` descriptions across BOTH twin suites (`prompts-subworkspace.test.js` + `prompts.test.js`) — the operation is the combined v3 `patchPrompt` (name + metadata in one request), not the retired v2 `rename`. Only the description strings were stale; the bodies already exercised `patchPrompt`. Left untouched: the legitimate `renamePrompt` transport-method tests in `rest-transport.test.js` and the tag rename/re-parent descriptions in `tags.test.js`. Fixed in 7fd696ce.

test/support/serenity/handlers/prompts.test.js
```diff
@@ -2030,3 +2066,243 @@ describe('handlers/prompts.js — unified type classification (serenity-docs#31)
...
+      });
+    });
+
+    // Locks buildPromptDto's per-field `?? null` fallback against a PARTIALLY
+    // populated upstream metadata object — a real shape for an un-backfilled
+    // prompt that was created (created_* stamped) but never edited (updated_*
+    // absent). Each field must fall back independently: present keys map, absent
+    // keys become null. A refactor that mapped metadata as a whole (or defaulted
+    // the object) instead of per-key would regress this and go uncaught.
+    it('maps only the populated metadata keys, nulling the absent ones per-field', async () => {
```
aliciadriani, line 2968: **Fixed (suggestion #3):** added a test exercising a PARTIALLY populated upstream `metadata` object (`created_*` present, `updated_*` absent — the real shape of a prompt created before a backfill but never edited). It asserts each authorship field falls back independently: present keys map through, absent keys → `null`. A refactor that mapped `metadata` wholesale (or defaulted the object) instead of the current per-key `?? null` would regress this and get caught. Fixed in 7fd696ce.

test/support/serenity/rest-transport.test.js
```diff
@@ -1670,3 +1700,45 @@ describe('Semrush REST transport', () => {
...
+// moment the bumped client ships the facade.
+describe('WP2 merge gate — Project Engine v3 authorship-metadata facade', () => {
+  const V3_METADATA_METHODS = [
+    'createPromptsWithMetadata',
+    'patchPrompt',
+    'patchPromptMetadata',
+    'patchPromptsMetadataBatch',
+  ];
+
+  it('the installed @adobe/spacecat-shared-project-engine-client exposes every v3 write method (bump before merge)', () => {
```
aliciadriani, line 1747: **Fixed (suggestion #4):** replaced the socially-enforced WP2 merge gate with a structural one. This test constructs the REAL `createSerenityProjectEngineTransport` and asserts it exposes all four v3 facade methods (`createPromptsWithMetadata`, `patchPrompt`, `patchPromptMetadata`, `patchPromptsMetadataBatch`). The installed client (1.15.0) does not carry them yet, so **this test fails today** — CI cannot go green, and this PR cannot merge clean, until `@adobe/spacecat-shared-project-engine-client` is bumped to the WP2 release that vendors them. That is deliberate: deploying WP1 against the current client would throw `<method> is not a function` on every native prompt write — the exact 'green in tests (mocked transport), broken in prod (real client lacks the method)' class the repo guards against (SITES-45260). Self-heals on the dep bump; delete this block + the file-level `ProjectEngineTransportWithMetadata` typedef then. Fixed in 7fd696ce.

aliciadriani:
Addressed all 4 non-blocking items from the last review in 7fd696ce. Inline notes on each fixed line.

- **suggestion #1 — builder `callerId` floor:** `buildCreateMetadata`/`buildUpdateMetadata` now floor to the `unknown` sentinel; matcher helpers floor identically (also retires the older `undefined`-in-tests nit).
- **nit #2 — stale `rename` wording:** prompt-update `it(...)` descriptions in both twin suites now say `patchPrompt`; the real `renamePrompt` transport tests and tag rename descriptions are untouched.
- **suggestion #3 — partial-metadata DTO test:** locks `buildPromptDto`'s per-field `?? null` fallback against a `created_*`-only upstream item.
- **suggestion #4 — structural WP2 merge gate:** a new test asserts the REAL PE-client exposes all four v3 facade methods.

⚠️ **Heads-up: CI is now intentionally RED.** Suggestion #4's gate FAILS against the currently-installed `@adobe/spacecat-shared-project-engine-client@1.15.0`, which does not yet vendor the v3 metadata facade. This is the merge gate made structural rather than social: the red is the guardrail, since deploying WP1 against the current client would throw `<method> is not a function` on every native prompt write (SITES-45260 class). It self-heals the moment the dep is bumped to the WP2 release — then delete the gate block + the file-level `ProjectEngineTransportWithMetadata` typedef. This matches the PR's own "DO NOT MERGE YET — merge window: after WP2 released" banner.

Local verification (node 24): `type-check` ✅ · `lint` ✅ · affected serenity suites 256 passing + the 1 intentional gate failure.

MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - prior findings addressed, no new blocking issues found.
**Complexity: HIGH** - large diff; API surface (OpenAPI + transport layer).
_Changes: Stamps prompt authorship metadata (created/updated at/by) on every write path, surfaces the fields on the list DTO, and adds server-side sort/order on the by_tags read (17 files)._
Note: Recommend a human read before merge - this change modifies a shared contract (OpenAPI schema). The bot review is a complement to, not a replacement for, a human read here.

<details>

- suggestion: Consider adding one test where `callerId` is explicitly `''` (empty string) to exercise the `||` branch in production, since the builders use falsy-coalescing (`||`) but the test matchers use nullish-coalescing (`??`). Low priority - the asymmetry only matters if a future caller passes `''` rather than `undefined`. - `src/support/serenity/handlers/prompts.js:477`
- suggestion: The WP2 merge gate's deletion instruction ("delete this block + the file-level typedef") should live in the WP2 ticket's acceptance criteria so it does not get forgotten across team rotation. Confirm LLMO-6289 WP2's definition-of-done includes removing the gate. - `test/support/serenity/rest-transport.test.js:1703`

</details>

- Builder `callerId` defensive floor added - `buildCreateMetadata`/`buildUpdateMetadata` now floor to `'unknown'` sentinel (commit 7fd696ce)
- Stale "rename" test descriptions updated to "patchPrompt" across both twin suites (commit 7fd696ce)
- Partial-metadata DTO test added - locks per-field `?? null` fallback against a half-populated upstream metadata object (commit 7fd696ce)
- WP2 merge gate structurally enforced via CI-blocking test that introspects the real installed PE client (commit 7fd696ce)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 3s | <b>Cost:</b> $4.45 | <b>Commit:</b> <code>7fd696ce43bf666b38b447dacf9d54df61cef79e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

test/support/serenity/handlers/prompts.test.js
```diff
@@ -2030,3 +2066,254 @@ describe('handlers/prompts.js — unified type classification (serenity-docs#31)
...
+      expect(meta.updated_at).to.match(/^\d{4}-\d{2}-\d{2}T.*Z$/);
+    });
+
+    // The builders floor with `||` (falsy-coalescing), so an EMPTY-STRING callerId
+    // floors to `unknown` just like `undefined`/`null` — a resolveCallerId result
+    // can only be a non-empty string or the sentinel, but a future direct caller
+    // that passes `''` must not stamp a blank author. (The test matchers use `??`,
+    // which would keep `''`; this builder-level assertion covers the `||` branch
+    // that the `??` matchers deliberately do not.)
+    it('floors an empty-string callerId to the unknown sentinel in both builders', () => {
```
aliciadriani, line 2897: **Fixed (suggestion #1):** added a builder-level test asserting an empty-string `callerId` floors to the `unknown` sentinel in both `buildCreateMetadata` and `buildUpdateMetadata`. This exercises the builders' `||` (falsy-coalescing) branch directly — the exact asymmetry you flagged, since the test matchers use `??` (which would keep `''`). A `resolveCallerId` result can only be a non-empty string or the sentinel, so this guards the future-direct-caller case. Fixed in 0151ded4.

test/support/serenity/rest-transport.test.js
```diff
@@ -1670,3 +1700,48 @@ describe('Semrush REST transport', () => {
...
+// the current client, every native prompt write would throw
+// `TypeError: <method> is not a function` in production — the exact "green in
+// tests (mocked transport), broken in prod (real client lacks the method)" class
+// the repo guards against (SITES-45260; see CLAUDE.md "Lambda Bundle Constraints").
+// This test FAILS until the installed client carries all four methods, so CI
+// cannot go green — and this PR cannot merge clean — before the WP2 dependency
+// bump lands. It self-heals (delete this block + the file-level typedef) the
+// moment the bumped client ships the facade. REMOVING THIS GATE (this block + the
+// `ProjectEngineTransportWithMetadata` typedef in rest-transport.js) is part of
+// WP2's definition-of-done (LLMO-6289) — it must be torn down with the dep bump,
```
aliciadriani, line 1737: **Addressed (suggestion #2):** reinforced the in-code pointer so the teardown instruction is impossible to miss at the gate itself — removing this block **and** the file-level `ProjectEngineTransportWithMetadata` typedef is called out as part of WP2's definition-of-done (LLMO-6289). The Jira-side DoD update is an author action I can't make from here; flagging it so it lands on the LLMO-6289 WP2 ticket's acceptance criteria as you suggested. Addressed in 0151ded4.

MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - prior findings addressed, no new blocking issues found.
**Complexity: HIGH** - large diff; API surface (OpenAPI + transport layer).
_Changes: Stamps prompt authorship metadata (created/updated at/by) on every write path, surfaces the fields on the list DTO, and adds server-side sort/order on the by_tags read (17 files)._
Note: Recommend a human read before merge - this change modifies a shared contract (OpenAPI schema). The bot review is a complement to, not a replacement for, a human read here.

<details>

- suggestion: Consider adding one test where `callerId` is explicitly `''` (empty string) to exercise the `||` branch in production, since the builders use falsy-coalescing (`||`) but the test matchers use nullish-coalescing (`??`). Low priority - the asymmetry only matters if a future caller passes `''` rather than `undefined`. - `src/support/serenity/handlers/prompts.js:477`
- suggestion: The WP2 merge gate's deletion instruction ("delete this block + the file-level typedef") should live in the WP2 ticket's acceptance criteria so it does not get forgotten across team rotation. Confirm LLMO-6289 WP2's definition-of-done includes removing the gate. - `test/support/serenity/rest-transport.test.js:1703`

</details>

- Empty-string callerId test added - exercises the `||` branch in both `buildCreateMetadata` and `buildUpdateMetadata` (commit 0151ded)
- WP2 merge gate teardown instruction documented explicitly in the gate comment as part of WP2's definition-of-done (commit 0151ded)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 33s | <b>Cost:</b> $2.53 | <b>Commit:</b> <code>0151ded44da3e58b2f53e71df7af21c4923002c7</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/serenity/rest-transport.js
```diff
@@ -97,6 +97,29 @@ const DEFAULT_TIMEOUT_MS = 15_000;
...
+ *   createPromptsWithMetadata: (init: any) => Promise<any>,
+ *   patchPrompt: (init: any) => Promise<any>,
+ *   patchPromptMetadata: (init: any) => Promise<any>,
+ *   patchPromptsMetadataBatch: (init: any) => Promise<any>,
+ * }} ProjectEngineTransportWithMetadata
```
aliciadriani, line 120: **Conflict resolved (main ↔ LLMO-6289 transport typedefs):** main's transport-typedef refactor introduced a new set — `PeTransport`/`UmSchemas`, the derived `*Body` typedefs, `WorkspaceResources`, `TransportEnv`, `FetchResult`, and `SerenityTransport` (= `ReturnType<typeof createSerenityTransport>`). This branch's `ProjectEngineTransportWithMetadata` (the v3 write-surface intersection / merge-gate stopgap) is additive and orthogonal, so **both are kept** — no typedef dropped. `type-check:base` + `type-check:strict` verified green on the rebased head.

src/support/serenity/rest-transport.js
```diff
@@ -420,18 +450,29 @@ export function createSerenityTransport({ env, imsToken }) {
      * @param {number} [body.limit]
      * @param {string} [body.search]
      * @param {boolean} [body.unassigned]
+     * @param {string} [body.sort] - allow-listed metadata sort field (LLMO-6289).
+     * @param {string} [body.order] - `asc` / `desc`; sent only with `sort`.
```
aliciadriani, line 454: **Conflict resolved (`listPromptsByTags` JSDoc):** kept the LLMO-6289 METADATA prose (sort/order is a real path now — it supersedes main's stale "Semrush rejects sort_field/sort_dir" note) **and** adopted main's added `@param` lines, extending them with `body.sort` / `body.order` so the doc matches the signature that forwards those keys. The downstream `if (body?.sort)` guard is unchanged, so an unsorted read stays byte-for-byte the legacy call.

src/support/serenity/handlers/prompts.js
```diff
@@ -522,18 +655,24 @@ export function normalizePromptInput(input) {
  * — so every id must already be a known-good upstream tag id, resolved by the
  * caller and never guessed.
  *
+ * STAMPS create authorship (LLMO-6289): every create goes through the v3
```
aliciadriani, line 658: **Conflict resolved (`createOnePrompt` JSDoc):** kept this branch's `STAMPS create authorship` paragraph and adopted main's stronger `@param {SerenityTransport} transport` in place of `{object}`. Safe: `SerenityTransport` = `ReturnType<typeof createSerenityTransport>`, and `createSerenityTransport` exposes `createPromptsWithMetadata` at the top level, so the v3 call here still type-checks.

src/support/serenity/handlers/prompts-subworkspace.js
```diff
@@ -364,10 +380,12 @@ export async function handleCreatePromptsSubworkspace(
  * PATCH /serenity/prompts/:semrushPromptId (subworkspace) — in-place edit.
  * Resolves the slice's project from the live listing, then edits the prompt IN
  * PLACE exactly like the flat-mode twin (see handleUpdatePrompt's contract):
- * `rename` first (the one op that can refuse — upstream 404 → promptNotFound,
- * 409 text collision → thrown for the controller's `conflict` mapping), then
- * the replace-mode batch tag write. The prompt id is preserved end to end and
- * echoed unchanged in the response; nothing is deleted on this path.
+ * the combined v3 `PATCH .../{id}` first (text `name` + `updated_*` metadata
```
aliciadriani, line 383: **Conflict resolved (`handleUpdatePromptSubworkspace` JSDoc):** kept this branch's combined-v3-`PATCH` prose (text `name` + `updated_*` stamp in one request, then the metadata-free v2 tag write) and adopted main's `@param {SerenityTransport} transport`. Kept in lockstep with the flat-mode twin.

test/support/serenity/rest-transport.test.js
```diff
@@ -1687,3 +1717,48 @@ describe('Semrush REST transport', () => {
...
+// not left behind to fail forever.
+describe('WP2 merge gate — Project Engine v3 authorship-metadata facade', () => {
```
aliciadriani, line 1739: **Review — Must Fix (by design, not a code change):** the sole failing test is *this* gate, and it is red on purpose. The installed `@adobe/spacecat-shared-project-engine-client@1.15.0` does not expose the v3 methods, so WP1 must not merge until WP2 bumps the dep — that is exactly what this structural gate enforces. "Fixing" it now (bumping to a WP2 release that may not exist yet, or stubbing the methods) would defeat the gate and risk shipping `is not a function` to prod. **Disposition: keep as-is** — the fix is the WP2 dep bump + deleting this block and the `ProjectEngineTransportWithMetadata` typedef at that time (already documented in the gate comment and the LLMO-6289 WP2 DoD).

src/support/serenity/handlers/prompts.js
```diff
@@ -1142,7 +1294,14 @@ export async function handleUpdatePrompt(
   typed = await injectComputedIntent(projectId, typed);
 
   try {
-    await transport.renamePrompt(semrushWorkspaceId, projectId, semrushPromptId, nextText);
+    // Combined v3 write: sets the text (`name`) and stamps the `updated_*`
+    // metadata pair in one request (replaces the v2 `rename`). Same refusal
+    // contract as rename — 404 (unknown id) → promptNotFound, 409 (text
+    // collides with a sibling) → thrown for the controller's `conflict` mapping.
+    await transport.patchPrompt(semrushWorkspaceId, projectId, semrushPromptId, {
```
aliciadriani, line 1301: **Review — Should Fix (by design, won't change server-side):** confirmed intentional. Every edit request stamps `updated_*` even when the text is unchanged, because upstream has no GET-by-id for the handler to diff against, and the contract is "a PATCH is a touch" (the DTO doc defines `updatedAt` as the last in-place edit of text *or* tags). Suppressing a no-op bump can only be decided by the client that knows nothing changed — not here. **Disposition: no change**; behavior matches the documented `handleUpdatePrompt` contract and the OpenAPI `updatedAt` description.

src/support/serenity/handlers/prompts.js
```diff
@@ -53,6 +53,124 @@ export const BULK_CREATE_CONCURRENCY = 8;
...
+ * an otherwise-legitimate write. If a future requirement needs `unknown`-authored
+ * writes rejected, that is a deliberate contract change to make HERE (reject at
+ * this boundary), not a silent behavior to assume.
+ *
+ * @param {object} ctx - the controller request context.
+ * @returns {string} the caller id, `unknown` when unresolved, ≤100 chars.
+ */
+export function resolveCallerId(ctx) {
+  const profile = ctx?.attributes?.authInfo?.getProfile?.();
+  const raw = profile?.user_id ?? profile?.sub;
```
aliciadriani, line 97: **Review — Nit (by design):** `resolveCallerId` floors a non-string `user_id`/`sub` to the `unknown` sentinel via `hasText`. IMS profile claims (`user_id`, `sub`) are always strings, so the numeric-id path is theoretical. Hardening it to coerce non-strings would add a branch for an input the auth layer never produces and slightly widen this CORRECTNESS-CRITICAL function's surface. **Disposition: no change** — documented here for the record.

src/support/serenity/handlers/prompts.js
```diff
@@ -53,6 +53,124 @@ export const BULK_CREATE_CONCURRENCY = 8;
...
+ * requested (order defaults to `desc` — newest first for "Last modified");
+ * throws 400 for an unknown sort field or order.
+ *
+ * @param {object} query
+ * @returns {{ sort?: string, order?: string }}
+ */
+export function resolveSort(query) {
+  const sort = query?.sort;
+  const order = query?.order;
+  if (sort === undefined || sort === null || sort === '') {
```
aliciadriani, line 156: **Review — Nit (by design):** `order` supplied without `sort` returns `{}` (order silently dropped). This matches the OpenAPI contract for the `order` param ("Ignored when `sort` is omitted") and the `resolveSort` doc — an order with nothing to order by is a no-op, not a 400. **Disposition: no change** — behavior is intentional and spec-consistent.

src/support/serenity/rest-transport.js
```diff
@@ -470,6 +511,101 @@ export function createSerenityTransport({ env, imsToken }) {
...
+     * `updated_by` longer than 100 chars) rolls the WHOLE batch back and answers
+     * 400 — so callers must isolate a deterministic offender rather than retry the
+     * batch whole. Nothing else in this WP calls it; it exposes the batch write
+     * surface the ADR pins for bulk stampers.
+     *
+     * @param {string} semrushWorkspaceId
+     * @param {string} projectId
+     * @param {Array<{ id: string, metadata: object }>} items
+     */
+    async patchPromptsMetadataBatch(semrushWorkspaceId, projectId, items) {
```
aliciadriani, line 600: **Review — Nit (by design):** `patchPromptsMetadataBatch` is intentionally unused in WP1. It exposes the ADR-pinned batch-metadata write surface for future bulk stampers (backfill / WP-N); nothing in this WP calls it yet, as its own JSDoc states. **Disposition: keep** — deliberate forward-looking transport surface, type-checked and covered by the metadata transport tests.

aliciadriani:
## Review (self-review pass, head `ae3cb284`)

Re-reviewed after the rebase onto latest `main`. **No actionable defects found.** `npx tsc` (base + strict), `eslint .`, and the serenity suites (**1527 passing**) all pass; the single failing test is the intentional WP2 merge gate.

Every item surfaced is documented in-code as intentional-by-design or is theoretical — each now has an inline disposition note:

| Item | Line | Disposition |
|---|---|---|
| WP2 dep-bump gate | `rest-transport.test.js` | By design — the gate *is* the mechanism; delete on the WP2 dep bump |
| No-op edit bumps `updated_*` | `prompts.js` `handleUpdatePrompt` | By design — a PATCH is a touch; no GET-by-id to diff; matches OpenAPI `updatedAt` |
| `resolveCallerId` non-string id | `prompts.js` | Theoretical — IMS claims are strings; no change |
| `order` without `sort` dropped | `prompts.js` `resolveSort` | Spec-consistent (OpenAPI: "Ignored when `sort` is omitted") |
| `patchPromptsMetadataBatch` unused | `rest-transport.js` | Deliberate forward-looking surface for bulk stampers |

**Only merge-blocker: the intentional WP2 dependency bump** (structurally enforced by the gate test). Security, contract/impl/test alignment, and end-to-end sort/order wiring all verified sound. Converged — nothing else to address.

jjenscodee:
Post-merge nit (non-blocking) — nice work; the stamping/caller-id logic reads exactly to spec (single `resolveCallerId` point, `getProfile().user_id ?? sub`, 100-char cap, never the forwarded bearer; create/update asymmetry so `created_*` is never rewritten; sort allow-list as the injection guard).

One cleanup to verify: the temporary **`ProjectEngineTransportWithMetadata` typedef shim** is still in `rest-transport.js` (`:120/:395/:474`) on `main`, even though the client bumped to `1.17.0`. The PR description said this shim was temporary — *"WP2's generated types supersede it on the dep bump."* So either 1.17.0's generated types now cover the v3 facade methods (→ the shim is redundant and can be dropped), or they still don't (→ it's load-bearing and the "supersede" note was optimistic). Worth a quick check + a small cleanup if redundant. No action needed on the merge itself.

PR #2882, opened 2026-07-22
fix(serenity): check retryOnQuota's shared deadline before every attempt

byteclimber:
## Summary
Self-review follow-up to #2874. Found that the shared per-request deadline was only checked inside the catch block, after a poll attempt had already failed — so the first poll attempt always ran regardless of how long `ensure()` itself took (e.g. queued behind other same-child recoveries on `withResourceLock`'s own up-to-10s safety valve).

This undermines the deadline's whole purpose: a request where multiple stacked wrap sites (e.g. `createProject` + `createPromptsByIds`, both hit in the same `generateTopics: true` create-market call on a freshly-activated brand) each independently hit the disguised-405 lag could still compound latency past the intended budget — a narrower recurrence of the exact stacked-call-site risk the shared deadline was built to close (round-2 SRE review on #2874).

- Moves the deadline check to run before every attempt, including the first, not just between retries.
- Tracks the most recent metered-quota error (`lastError`) so a pre-attempt bail still throws something meaningful instead of needing a fresh failure to report.
- `attempt` now starts at 0 and increments only once an attempt is actually made, so a pre-attempt deadline bail reports `exhausted` at the attempt count actually reached (possibly 0, meaning `ensure()` alone consumed the whole budget).

## Test plan
- [x] `npm run lint && npm run type-check && npm test` — full suite green (15424 passing)
- [x] Updated the two deadline-specific tests to reflect the corrected pre-attempt semantics
- [x] Added a new test proving a budget already blown by `ensure()` skips the poll attempt entirely (the exact gap this fix closes)

### DIFF
--- a/src/support/serenity/allocation-metrics.js
+++ b/src/support/serenity/allocation-metrics.js
@@ -60,8 +60,8 @@
  *   `recovered + exhausted + abandoned` as "total recovery cycles" doesn't silently undercount,
  *   per Alicia Adriani's review). `Attempt` (the 1-based poll attempt the cycle ended on, or `0`
  *   if the shared deadline was already exhausted before any poll attempt could even be made — see
- *   `retryOnQuota`'s pre-attempt deadline check, api-service#2882) and `CallSite` (a small closed
- *   set — `publishProject`, `createProject`, `createOnePrompt`,
+ *   `retryOnQuota`'s pre-attempt deadline check, adobe/spacecat-api-service#2882) and `CallSite`
+ *   (a small closed set — `publishProject`, `createProject`, `createOnePrompt`,
  *   `createPromptsByIds`, ...) together give the attempt-to-recovery distribution per write path,
  *   without which a rising `exhausted` rate or a shift toward later attempts is invisible until it
  *   becomes an incident (round-2 SRE review).
@@ -173,7 +173,7 @@ export function recordMeteredQuotaClassifier(matched) {
  * @param {{ attempt: number, callSite: string }} dims - `attempt` is the 1-based poll attempt the
  *   cycle resolved (recovered), gave up on (exhausted), or was cut short on by a non-quota error
  *   (abandoned) — OR `0` for `exhausted` specifically when the shared deadline was already blown
- *   before any poll attempt could be made (api-service#2882). `callSite` is a short
+ *   before any poll attempt could be made (adobe/spacecat-api-service#2882). `callSite` is a short
  *   closed-vocabulary label identifying which wrapped write/publish this recovery was for.
  *   `attempt` is a raw number, not pre-bucketed — it stays low-cardinality only because
  *   `retryOnQuota`'s `maxAttempts` is small (3 today); if `maxAttempts` is ever raised
--- a/src/support/serenity/allocation-metrics.js
+++ b/src/support/serenity/allocation-metrics.js
@@ -58,8 +58,10 @@
  *   shared deadline), or `abandoned` (a genuinely non-quota error cut the cycle short before
  *   either of the above — distinct from `exhausted` so a dashboard summing
  *   `recovered + exhausted + abandoned` as "total recovery cycles" doesn't silently undercount,
- *   per Alicia Adriani's review). `Attempt` (the 1-based attempt the cycle ended on) and
- *   `CallSite` (a small closed set — `publishProject`, `createProject`, `createOnePrompt`,
+ *   per Alicia Adriani's review). `Attempt` (the 1-based poll attempt the cycle ended on, or `0`
+ *   if the shared deadline was already exhausted before any poll attempt could even be made — see
+ *   `retryOnQuota`'s pre-attempt deadline check, api-service#2882) and `CallSite` (a small closed
+ *   set — `publishProject`, `createProject`, `createOnePrompt`,
  *   `createPromptsByIds`, ...) together give the attempt-to-recovery distribution per write path,
  *   without which a rising `exhausted` rate or a shift toward later attempts is invisible until it
  *   becomes an incident (round-2 SRE review).
@@ -168,13 +170,15 @@ export function recordMeteredQuotaClassifier(matched) {
 
 /**
  * @param {'recovered'|'exhausted'|'abandoned'} outcome
- * @param {{ attempt: number, callSite: string }} dims - `attempt` is the 1-based attempt the
- *   poll-retry cycle resolved (recovered), gave up on (exhausted), or was cut short on by a
- *   non-quota error (abandoned); `callSite` is a short closed-vocabulary label identifying which
- *   wrapped write/publish this recovery was for. `attempt` is a raw number, not pre-bucketed — it
- *   stays low-cardinality only because `retryOnQuota`'s `maxAttempts` is small (3 today); if
- *   `maxAttempts` is ever raised significantly, revisit whether `Attempt` should be capped/bucketed
- *   to hold the module's low-cardinality dimension contract (see the module doc above).
+ * @param {{ attempt: number, callSite: string }} dims - `attempt` is the 1-based poll attempt the
+ *   cycle resolved (recovered), gave up on (exhausted), or was cut short on by a non-quota error
+ *   (abandoned) — OR `0` for `exhausted` specifically when the shared deadline was already blown
+ *   before any poll attempt could be made (api-service#2882). `callSite` is a short
+ *   closed-vocabulary label identifying which wrapped write/publish this recovery was for.
+ *   `attempt` is a raw number, not pre-bucketed — it stays low-cardinality only because
+ *   `retryOnQuota`'s `maxAttempts` is small (3 today); if `maxAttempts` is ever raised
+ *   significantly, revisit whether `Attempt` should be capped/bucketed to hold the module's
+ *   low-cardinality dimension contract (see the module doc above).
  * @returns {void}
  */
 export function recordQuotaRetryOutcome(outcome, { attempt, callSite }) {
--- a/src/support/serenity/dynamic-allocation-active.js
+++ b/src/support/serenity/dynamic-allocation-active.js
@@ -181,15 +181,9 @@ export function createHeadroomGuard(transport, {
         let attempt = 0;
         let lastError = e;
         for (;;) {
-          // Checked BEFORE every attempt, including the first (not just between retries): if
-          // `ensure()` itself already consumed the whole shared budget (e.g. it queued behind
-          // other same-child recoveries on `withResourceLock`'s own up-to-10s safety valve), don't
-          // spend a further `fn()` call on top of an already-blown deadline. This closes a real
-          // gap found on self-review: checking the deadline only inside the catch (i.e. only after
-          // a poll attempt had already failed) meant the FIRST attempt always ran regardless of
-          // how long `ensure()` took, so the shared per-request budget this deadline exists to
-          // enforce (see the comment above) wasn't actually a hard cap — exactly the
-          // stacked-call-site compounding-latency risk the deadline was built to close.
+          // Checked BEFORE every attempt, including the first — an already-blown deadline (e.g.
+          // `ensure()` itself ate the whole budget queued behind another same-child recovery) must
+          // not spend a further `fn()` call on top of it. See the header comment above for why.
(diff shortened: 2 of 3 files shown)


MysticatBot (approved):
Hey @byteclimber,

**Verdict: Approve** - clean, well-scoped fix that closes a real gap in deadline enforcement; no blocking findings.
**Complexity: LOW** - small single-service bug fix.
_Changes: Moves the shared per-request deadline check to run before every poll attempt (including the first), so a slow `ensure()` call can no longer silently blow past the budget (2 files)._

<details>

- nit: The 8-line inline comment block (lines 181-192 in the new code) restates the design rationale already present in the header comment and includes process artifacts ("self-review finding", "found on self-review") that will age poorly. Consider trimming to 2-3 lines that state the invariant, not its discovery story - `src/support/serenity/dynamic-allocation-active.js:181`
- suggestion: When the pre-attempt deadline check fires at `attempt=0`, `recordQuotaRetryOutcome('exhausted', { attempt: 0, callSite })` introduces a new metric dimension value that did not exist before (minimum was previously 1). Confirm downstream dashboards/alerts that consume this metric handle `attempt=0` gracefully, or consider a distinct outcome label like `'budget_exceeded'` - `src/support/serenity/dynamic-allocation-active.js:190`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 7m 59s | <b>Cost:</b> $3.35 | <b>Commit:</b> <code>de25ef5a75da64003bc84d5ba950c513053e2589</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

aliciadriani:
## Code review — `fix(serenity): check retryOnQuota's shared deadline before every attempt`

Follow-up to #2874; 2-file diff on `fix/retryonquota-deadline-preattempt`._

### Summary
Tight, correct fix. Moving the `requestDeadline` check to the top of the loop (before every attempt, not just between retries) genuinely closes the gap where a slow `ensure()` — queued behind other same-child recoveries on `withResourceLock`'s up-to-10s valve — could let the first poll run on an already-blown budget. The `lastError` tracking is the right complement: a pre-attempt bail now throws something meaningful (the original triggering error, or the most recent metered failure) instead of needing a fresh failure. The `attempt`-starts-at-0-and-increments-before-each-poll rewrite preserves the existing metric semantics for the common paths (recovered/exhausted still report the poll attempt number they ended on) while letting a pre-attempt bail honestly report `attempt: 0`. Tests are updated to match the corrected pre-attempt semantics and the new "budget already blown by `ensure()`" case pins exactly the gap being closed. I traced the recovered / exhausted-by-maxAttempts / abandoned / deadline-before-first-poll / deadline-mid-loop paths and they all hold.

### Should Fix / Consider

**1. The trailing `sleep(backoffMs)` can still overshoot the shared deadline by up to `backoffMs`, doing nothing useful — which slightly undercuts this PR's own "hard cap" intent.**

`dynamic-allocation-active.js` — after a failed metered attempt, when `attempt < maxAttempts`, the code sleeps `backoffMs` **unconditionally**, and only the *next* loop-top check consults the deadline. So if an attempt starts just inside the budget and fails, the loop sleeps a full `backoffMs` and *then* bails at the top without another `fn()` call:

```
attempt starts at t=8000 (deadline=9000) → fails fast at t≈8050
attempt < maxAttempts → sleep(3000) → t≈11050
loop top: 11050 >= 9000 → exhausted, throw   (overshoot ≈ 2050ms, zero useful work)
```

The moved-up pre-attempt check makes the deadline a gate on *starting* an attempt, but the request can still run ~`backoffMs` past the deadline sleeping into a bail. This is precisely the stacked-call-site scenario the deadline exists to bound (a late call site whose budget is nearly spent), so it's on-thesis feedback rather than incidental. A one-line guard before the sleep — e.g. bail if `now() + backoffMs >= requestDeadline` (or just re-check the deadline before sleeping) — makes the cap tight without a wasted wait. Minor (bounded by `backoffMs`, and with the default `maxAttempts=3`/`backoffMs=3000` the attempt cap usually binds first for an early call site), but it's a real, easy tightening.

### Nits (optional)

**2. `attempt: 0` now contradicts the metric's JSDoc, which wasn't updated.** This PR can emit `recordQuotaRetryOutcome('exhausted', { attempt: 0, ... })` when `ensure()` alone blows the budget (correct, and called out in the PR body). But `allocation-metrics.js` — untouched here — still documents `attempt` as *"the **1-based** attempt the poll-retry cycle resolved / gave up on"*. That's now stale: 0 is a valid value meaning "no poll attempt was made." Worth a one-line JSDoc tweak on `recordQuotaRetryOutcome` (and the module `QuotaRetryOutcome` doc block) so a dashboard author isn't surprised by an `Attempt=0` bucket. Cardinality contract is unaffected (adds one bounded value).

### What's Good
- **Root cause correctly identified and minimally addressed** — the deadline was only ever consulted after a failure, so it was never a true pre-attempt cap; moving it up is the right fix, not a workaround.
- **`lastError` seeding from the initial triggering error `e`** means a bail at `attempt: 0` still throws a real, relevant error rather than fabricating one or throwing `undefined`.
- **Metric attempt-number semantics preserved** across the increment-timing change — the `recovered`/`exhausted` attempt values still line up with the poll attempt that ended the cycle, so no dashboard reinterpretation is needed for the existing buckets.
- **The new test is exactly the missing case** (`ensure()` alone consumes the budget → zero poll attempts, original error propagates), and the shared-deadline test was correctly updated to the new pre-attempt semantics (call site B now bails on its first pre-attempt check, `calledOnce`) while still guarding the recompute-per-call regression.

MysticatBot (approved):
Hey @byteclimber,

**Verdict: Approve** - clean, well-scoped fix that closes the pre-attempt deadline gap and addresses all prior review findings.
**Complexity: MEDIUM** - medium diff, single-service bug fix.
_Changes: Moves the shared per-request deadline check to run before every poll attempt (including the first) and adds a pre-sleep check so sleeping does not overshoot the budget (3 files)._

<details>

- suggestion: The "budget blown by ensure()" test pins the core new behavior but does not assert `recordQuotaRetryOutcome('exhausted', { attempt: 0, callSite: 'unknown' })` was called - the one new metric dimension value this PR documents. Adding an esmock stub + assertion (like the sleep-overshoot test does) would pin the contract cheaply - `test/support/serenity/dynamic-allocation-active.test.js:393`
- nit: PR references in comments use the short form `api-service#2882` where the workspace convention is full-qualified `adobe/spacecat-api-service#2882` - `src/support/serenity/allocation-metrics.js:62`
- nit: The "deadline stops retrying" test name implies it tests pre-attempt semantics, but it actually only proves deadline-over-attempt-cap priority (the "budget blown by ensure()" test is the real regression guard for pre-attempt behavior) - `test/support/serenity/dynamic-allocation-active.test.js:354`

</details>

- Sleep overshoot: pre-sleep check added (`now() + backoffMs >= requestDeadline`)
- JSDoc for attempt=0: both the module doc and `recordQuotaRetryOutcome` JSDoc updated
- Comment verbosity: header comment trimmed
- Shared-deadline test: updated to reflect pre-attempt semantics (call site B now bails before its first poll attempt)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 30s | <b>Cost:</b> $4.69 | <b>Commit:</b> <code>ea7634be244b12c0f894024b24c6e0d391f625a9</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

aliciadriani:
## Re-review — both prior findings addressed ✅

Confirmed the two points from my earlier review are fully resolved, and I re-traced the loop for regressions — none found.

### Summary
Clean resolution. The deadline is now a genuine hard cap on both ends: gated *before* every attempt (the original fix) and *before* the inter-attempt sleep, so a request can no longer overshoot the shared budget by up to `backoffMs` sleeping into a guaranteed bail. The metric doc now matches the code. Correctness of the recovered / exhausted-by-attempts / exhausted-by-deadline / abandoned / pre-attempt-bail / pre-sleep-bail paths all check out.

### Prior findings — status

**1. Trailing `sleep(backoffMs)` overshoot → FIXED.** The exhaustion guard is now `attempt >= maxAttempts || now() + backoffMs >= requestDeadline`, so if the next sleep would cross the deadline the loop bails immediately instead of sleeping ~`backoffMs` into a bail it already knows is coming. Importantly this drops **no** useful retries: any attempt after that skipped sleep would start at `≥ requestDeadline` and be rejected by the loop-top check anyway — so the change is a pure latency tightening, not a behavior regression. The new test (`a sleep that WOULD overshoot the deadline is skipped entirely`) pins it precisely: `now()=950`, `backoffMs=3000`, `deadline=1000` → bails at `attempt: 1`, `sleep` asserted **not** called, original error propagated.

**2. `attempt: 0` doc drift → FIXED.** `allocation-metrics.js` (both the module-level `QuotaRetryOutcome` block and the `recordQuotaRetryOutcome` JSDoc) now documents `attempt` as the 1-based poll attempt *or* `0` when the shared deadline was already exhausted before any poll ran, cross-referencing #2882. The `attempt: 0` case is also now asserted in the metric via the updated "budget already blown by `ensure()`" test.

### What's Good
- Both fixes came with a **dedicated regression test that fails without the change** (skipped `sleep`, `attempt: 0` metric), not just a green suite.
- The pre-sleep guard is the minimal correct form — reusing the existing `now()`/`backoffMs` rather than introducing new state — and the inline comment explains the "overshoot by a full `backoffMs` doing nothing useful" rationale exactly.
- Metric semantics stayed stable for every existing bucket; only the genuinely-new `attempt: 0` value was added, and it's documented rather than left for a dashboard author to discover.

LGTM. (Note: I reviewed the diff only — I did not run the suite locally, as this repo isn't checked out in my workspace; the PR reports 15,424 passing.)

MysticatBot (approved):
Hey @byteclimber,

**Verdict: Approve** - clean, well-scoped fix that closes the pre-attempt deadline gap; all prior findings addressed, no blocking issues remain.
**Complexity: MEDIUM** - medium diff, single-service bug fix.
_Changes: Moves the shared per-request deadline check to run before every poll attempt (including the first), adds a pre-sleep overshoot guard, and pins the metric assertion for the attempt=0 case (3 files)._

<details>

- nit: The deadline-priority test comment says "3rd = the deadline check before poll attempt 2 (now past budget, so it never runs)" but the 3rd `now()` call actually fires in the pre-sleep check (`now() + backoffMs >= requestDeadline`) inside the catch after poll attempt 1 fails, not at the loop-top before poll attempt 2. The test logic is correct but the comment could mislead a reader debugging a failure - `test/support/serenity/dynamic-allocation-active.test.js:366`
- suggestion: The pre-attempt check (`now() >= requestDeadline`) and the pre-sleep check (`now() + backoffMs >= requestDeadline`) use different comparison semantics for correct reasons (current-time vs projected-time). A one-line inline note on the pre-sleep check making the distinction explicit ("bail if sleeping would land at or past the deadline, not just if already past") would help a future maintainer who sees two `>= requestDeadline` checks and wonders why one adds `backoffMs` - `src/support/serenity/dynamic-allocation-active.js:191`

</details>

- Sleep overshoot: pre-sleep check added (`now() + backoffMs >= requestDeadline`)
- JSDoc for attempt=0: both the module doc and `recordQuotaRetryOutcome` JSDoc updated
- Comment verbosity: header comment trimmed
- Shared-deadline test: updated to reflect pre-attempt semantics
- PR reference format: now uses fully-qualified `adobe/spacecat-api-service#2882`
- Metric assertion: "budget blown by ensure()" test now asserts `recordQuotaRetryOutcome('exhausted', { attempt: 0, callSite: 'unknown' })`

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 14s | <b>Cost:</b> $4.27 | <b>Commit:</b> <code>9a6e1b6438bcee2abc56a9f2280358bbcf8a8333</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

aliciadriani (approved):

PR #2883, opened 2026-07-22
fix(serenity): finish LLMO-6190 headroom wiring (unbreak type-check + build)

byteclimber:
## Summary

The `feat/serenity-intent-classification` branch shipped the LLMO-6190 dynamic-allocation headroom guard **half-wired**, which is why the branch's own CI (PR #2785) and the stacked edit-path PR #2875 are both red. This fixes the base breakage. Base is `feat/serenity-intent-classification`, not `main`.

### What was broken (reproduced on the pristine base)
- `type-check`: `markets-subworkspace.js` — `generateAndAttachPrompts`'s JSDoc documents a required `headroom` param and the caller passes it, but the function never declared/used it (TS8024 + TS2554: 6 args to a 5-param fn).
- `ci / build` (eslint): `prompts-subworkspace.js:251` — a second `headroom` guard shadowed the one at line 171 (`no-shadow`).
- test: `handleCreatePromptsSubworkspace` dynamic-allocation test failed — its options object was passed in the `env` positional slot instead of the 8th, so the guard was never enabled.

### Fixes
1. **`markets-subworkspace.js`** — add `headroom` as the 6th param to `generateAndAttachPrompts` and front `headroom.ensure({ prompts: texts.size }, { includeDrafted: true })` before the metered `createPromptsByIds` write loop (the metered write is the create, not publish). Matches the JSDoc + the `prompts-subworkspace.js` twin. No-op when the flag is OFF.
2. **`prompts-subworkspace.js`** — ⚠️ **behavior change, please scrutinize:** removed the redundant pre-publish `ensure` (which re-created a second guard → the `no-shadow`, and issued a second `getWorkspaceResources` read). The LLMO-6190 design meters **once before the write** (test title: "front headroom before the write, *not just before publish*"; assertion: `getWorkspaceResources` called **once**, before `createPromptsByIds`). The pre-write `ensure` at line 176 already reserves the whole batch, so the pre-publish read was a pre-LLMO-6190 leftover. If a second pre-publish top-up is intentional, this is the line to push back on.
3. **`prompts-subworkspace.test.js`** — fixed the stale positional call (options → 8th slot) to match the controller + signature.

### Verification
Type-check clean, lint clean. Combined with PR #2875, the three serenity handler test suites are **223 passing / 0 failing**.

### ⚠️ Interdependency with #2875
This PR and #2875 are complementary — the feature branch is broken in two places and each PR fixes one, so **neither is green alone**; both must merge for `feat/serenity-intent-classification` (and PR #2785) to go green. Verified locally that edit-path + headroom together are fully green.

### DIFF
--- a/src/support/serenity/handlers/markets-subworkspace.js
+++ b/src/support/serenity/handlers/markets-subworkspace.js
@@ -224,7 +224,7 @@ function validateCreateBody(body) {
 async function generateAndAttachPrompts(transport, workspaceId, projectId, {
   domain, country, topicCap = 0, brandNames = [], provisioned, env,
   writeDeadline = computeWriteDeadline(),
-}, log) {
+}, log, headroom) {
   const raw = await transport.getBrandTopics(workspaceId, { domain, country });
   let topics = [];
   if (Array.isArray(raw)) {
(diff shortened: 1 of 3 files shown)


byteclimber:
Consolidated into #2875. These two fixes are a *mutual* dependency (each PR's CI ran the full suite and was red on the other's unfixed domain), so two separate green PRs was technically impossible. The headroom/LLMO-6190 commit is now cherry-picked onto #2875; that branch is green (type-check + lint clean, 223 serenity handler tests passing). Closing this one.

PR #2887, opened 2026-07-22
fix(llmo): 404 brand-claims when the S3 object is missing

davidaurelio:
## Problem

`GET /sites/{siteId}/llmo/brand-claims` returns a presigned S3 URL for the brand-claims data file. It was **always** returning a `200` + presigned URL, even when no data file exists for the site — the client only found out when it fetched the URL and got a `404` from S3.

Root cause: presigning a `GetObjectCommand` is an **offline** operation (pure local signature computation via `s3-request-presigner`) — it never contacts S3. The controller's existing `catch (NoSuchKey)` branch was therefore **dead code** and could never fire. So the endpoint could not be used to tell whether data exists.

## Change

- `HeadObject` the key before signing, and return a clean `404` when the object is missing — mirroring `getFanoutReport` (`controllers/llmo/fanout-report.js`), which already does exactly this.
- Error handling matches the sibling: `NotFound` name **or** `$metadata.httpStatusCode === 404` → `404`; `NoSuchBucket` and other errors keep their existing `400` mapping.
- Cost: one extra in-region `HeadObject` round-trip per call (metadata-only, ~single-digit ms). The Lambda role already `HeadObject`s this bucket in `fanout-report`/`demo`, so no new IAM is needed.

This makes the endpoint usable as a cheap **availability probe** — the motivating use case is an "all brands" view in the UI, where most brands don't have brand-claims data and we want to know which do without downloading anything.

## Related Issues

No linked issue — problem described above (surfaced while investigating an "all brands" brand-claims view in the Elmo UI).

---

### PR checklist

- [x] Problem described above (no separate issue to link).
- [x] **Changing the API implementation:** the OpenAPI spec (`docs/openapi/llmo-api.yaml`) **already** documents the `404` response for this endpoint, so no spec change is required — this PR makes the implementation finally honor the documented contract. (`docs/index.html` is regenerated from that spec and is likewise already accurate.)
- [x] Not a new audit type; not an API-specification (schema/path) change.

## Testing

- `npx eslint src/controllers/llmo/brand-claims.js test/controllers/llmo/brand-claims.test.js` — clean.
- Updated `test/controllers/llmo/brand-claims.test.js` (the old cases mocked `getSignedUrl` rejecting, which exercised the unreachable path). New/updated cases cover: happy path asserts `HeadObject` runs with the right bucket/key before signing; `404` on `HeadObject` `NotFound`; `404` on `$metadata.httpStatusCode === 404` (and asserts we never sign a URL for a missing object); `400` on `NoSuchBucket`; `400` on generic errors. `12 passing`.

### DIFF
--- a/src/controllers/llmo/brand-claims.js
+++ b/src/controllers/llmo/brand-claims.js
@@ -13,6 +13,7 @@
 import {
   badRequest, notFound,
 } from '@adobe/spacecat-shared-http-utils';
+import { HeadObjectCommand } from '@aws-sdk/client-s3';
 import { cachedOk } from '../../support/cached-response.js';
 
 /**
@@ -45,6 +46,14 @@ export async function handleBrandClaims(context) {
 
   try {
     const { getSignedUrl, GetObjectCommand } = s3;
+
+    // Presigning a GetObject URL is an offline operation and never checks that
+    // the object exists, so without this HeadObject the endpoint would happily
+    // hand out a URL that 404s on fetch. Verify existence first and return a
+    // clean 404 otherwise (mirrors getFanoutReport). This also lets callers use
+    // the endpoint as a cheap availability probe (e.g. an "all brands" view).
+    await s3.s3Client.send(new HeadObjectCommand({ Bucket: bucketName, Key: s3Key }));
+
     const command = new GetObjectCommand({
       Bucket: bucketName,
       Key: s3Key,
@@ -60,7 +69,7 @@ export async function handleBrandClaims(context) {
       expiresAt: new Date(Date.now() + expiresIn * 1000).toISOString(),
     });
   } catch (s3Error) {
-    if (s3Error.name === 'NoSuchKey') {
+    if (s3Error.name === 'NotFound' || s3Error.$metadata?.httpStatusCode === 404) {
       log.warn(`Brand claims file not found for site ${siteId} at ${s3Key}`);
       return notFound(`Brand claims data not found for site ${siteId}`);
     }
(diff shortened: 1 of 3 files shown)


MysticatBot (approved):
Hey @davidaurelio,

**Verdict: Approve** - clean, well-scoped fix that makes the implementation honor the already-documented 404 contract.
_Changes: adds a HeadObject existence check before presigning the S3 URL in the brand-claims endpoint, returning a clean 404 when the object is missing (2 files)._

<details>

- nit: `HeadObjectCommand` is imported directly from `@aws-sdk/client-s3` while `GetObjectCommand` is destructured from the injected `s3` context object - a DI asymmetry within the same function. Follows the established `fanout-report.js` pattern so not a divergence, but worth noting if the team later consolidates these into a shared helper. - `src/controllers/llmo/brand-claims.js:16`
- suggestion: the `httpStatusCode === 404` test case does not assert on the `log.warn` call (the `NotFound`-by-name test does). Adding the assertion keeps observability expectations symmetric across both 404 paths. - `test/controllers/llmo/brand-claims.test.js:188`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 50s | <b>Cost:</b> $3.49 | <b>Commit:</b> <code>e05c46d035b9c4d13e4acae4b44784f8985ff90b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2889, opened 2026-07-22
fix(serenity): classify prompt-publish quota rejections, not just market-create

claudiaboldis:
## What & why

Full coverage of adobe/serenity-docs#72 ("Prompt Quota Exhaustion — Customer Experience and Alerting") across every prompt-affecting write path, built up over this PR's commit history. Started as a follow-up to #2854 (which only covered the market-create publish leg) and grew to close every remaining gap identified in a full re-audit against the spec.

## What changed

**§4.1 quota classification** — every metered write and publish leg now surfaces a residual disguised-405 quota rejection as the stable `{ status: 409, error: 'quotaExceeded' }` token instead of a generic `publish: <message>` 502 record or a raw upstream status:
- Prompt bulk-create's per-item metered write and `publishAffected`'s shared publish-leg fan-out (flat + sub-workspace twins, `prompts.js` / `prompts-subworkspace.js`).
- Bulk-delete's publish leg (both twins).
- Model-set-change publish (`syncModelsForProject`, shared core in `markets.js`, used by both flat `handleUpdateModels` and `handleUpdateModelsSubworkspace`).
- Market-create (already covered by #2854, kept in lockstep with the other call sites).

**§4.1 atomicity** — `reconcilePublishErrors` rolls back (deletes) any prompts this request staged in a project whose publish was quota-rejected, moving them from `created` into `failed` as a 409 — never left as unpublished drafts. A rollback that itself fails is a distinct, alertable engineering-defect signal (`alertRollbackFailure`), not just a log line.

**§5 Slack alerting** — a new `quota-alerts.js` module (`alertQuotaRejection`, `alertRollbackFailure`, `alertPoolFreeThreshold`), opt-in via a `SERENITY_QUOTA_ALERTS_ENABLED` kill-switch (default OFF) and a configurable Slack channel, wired into every classified rejection site above, plus the JIT allocator's hard-exhaustion throws (`orgPoolExhausted` / `brandAiLimit` in `resource-manager.js`) and its advisory pool-free early-warning read. Deduplicated per org/brand/case/dimension within a 15-minute window, with a bounded eviction sweep so the in-memory dedup map can't grow unbounded across a long-lived warm container.

## Explicitly NOT in this PR (tracked as follow-up in LLMO-6470)

- Pre-flight draft-market project-unit capacity check (§4.1) — only the post-hoc rollback exists; there's no check BEFORE a draft project is created that verifies headroom, for either the allocator-ON or allocator-OFF path.
- Computing the `carved` field (sum of child sub-workspace carves) for the org-pool alert payload — the field exists in the payload shape but is never populated; feasibility of enumerating child sub-workspaces is still an open investigation.
- Wiring the `SERENITY_QUOTA_ALERTS_ENABLED` / `SERENITY_QUOTA_ALERTS_SLACK_CHANNEL_ID` env vars into any CDK stack or Vault config — the alerting code is live but currently inert everywhere (including prod) until ops configures and flips it on.

## Test plan

- New/updated tests across `test/support/serenity/handlers/{prompts,prompts-subworkspace}.test.js`, `test/support/serenity/quota-alerts.test.js`, `test/support/serenity/resource-manager.test.js`, and `test/controllers/serenity.test.js` covering every classification/rollback/alerting path added.
- eslint clean on all touched files.
- `tsc --noEmit` clean (this repo's opt-in `// @ts-check` coverage over `src/support/serenity/**`).
- Full serenity suite (`test/support/serenity/**/*.test.js` + `test/controllers/serenity.test.js`) passing, no regressions.

Introduced by: N/A


MysticatBot (changes requested):
Hey @claudiaboldis,

**Verdict: Request changes** - two actionable issues: an unbounded dedup Map and an unredacted error message leaking to Slack.
**Complexity: HIGH** - large diff (13 files); API surface signal.
_Changes: Extends quota-rejection classification from market-create to all prompt-affecting writes (bulk create, bulk delete, model-set change), adds rollback atomicity for staged prompts, and introduces a kill-switch-gated Slack alerting module (13 files)._

## Must fix before merge
1. [Important] Dedup Map in `quota-alerts.js` grows unbounded - expired entries are never evicted - `src/support/serenity/quota-alerts.js` (the module-scoped `seen` Map, line 15)

   The `seen` Map accumulates one entry per unique dedup key (`orgId:brandId:caseType:dimension`) and never evicts expired entries. Keys that fire once and never recur persist forever in the warm Lambda container. The fix is trivial: add a lazy eviction sweep (e.g. `if (seen.size > 500) { for (const [k, v] of seen) { if (now - v.firstAt >= DEDUP_WINDOW_MS) seen.delete(k); } }`) or delete the stale entry in the existing check when the window has elapsed.

2. [Important] Unredacted upstream error message leaked to Slack - `src/support/serenity/handlers/prompts.js`, `reconcilePublishErrors` rollback catch block

   `rollbackErr?.message` is passed directly to `alertRollbackFailure` without going through `redactUpstreamMessage()`. The JSDoc on `alertRollbackFailure` claims the caller has "already redacted" the error, but this call site does not. The raw upstream message from `transport.deletePromptsByIds` could contain internal service URLs or stack traces. Fix: `rollbackError: redactUpstreamMessage(rollbackErr)` - the function is already imported in this file.

<details>

- suggestion: Consider moving `await alertQuotaRejection` after the `Promise.all` loop in `publishAffected` to avoid adding Slack-post latency to the publish fan-out critical path (dedup guarantees only one post anyway) - `src/support/serenity/handlers/prompts.js`
- nit: `brandId` passed both positionally and inside `alertContext` in `handleUpdateModels` creates a redundant double-source of truth - `src/support/serenity/handlers/markets.js`
- suggestion: Truncate `semrushPromptIds` array in `alertRollbackFailure` message (e.g. first 10 + "... and N more") to prevent oversized Slack messages from being silently dropped - `src/support/serenity/quota-alerts.js`
- nit: Positional `undefined` for `wrapPublish` in bulk-delete callers of `publishAffected` is fragile; an options bag for trailing params would be more resilient - `src/support/serenity/handlers/prompts-subworkspace.js`
</details>

aenascut:
Great work closing the quota-rejection gaps across all four prompt call sites. CI's clean, scope is solid. A few catches before merge:

## Must fix

**1. `reconcilePublishErrors`: quota project with zero staged prompts silently drops**

When a quota-rejected project has no matching `rollbackProjectId` in `created`, the error vanishes — never appended to `failed`, no 409 emitted.

```js
for (const r of publishErrors) {
  if (pubErr.code === ERROR_CODES.QUOTA_EXCEEDED) {
    const staged = [];
    // ...splice logic...
    if (staged.length > 0) {
      // rollback
    }
    // ← if staged.length === 0, quota error is lost
    for (const item of staged) {
      failed.push(...); // never runs
    }
  }
}
```

Add fallback in `prompts.js` + `prompts-subworkspace.js`:
```js
if (staged.length === 0) {
  failed.push({ text: '', status: 409, error: ERROR_CODES.QUOTA_EXCEEDED, message: pubErr.message });
}
```

**2. PR description conflicts with code**

> "Model-set change (handleUpdateModelsSubworkspace) publish leg — a separate call site, not touched here."

But `syncModelsForProject` in `markets.js` gained the full `isMeteredQuota` catch + throw, and `handleUpdateModelsSubworkspace` is threading `orgId`/`brandId`/`env`. Both ARE touched. Reword "Explicitly NOT in this PR" to match the actual changes.

**3. `syncModelsForProject` publish 405→409 path untested**

The new quota catch in `markets.js` has no test exercising it. `prompts.test.js` covers both `handleCreatePrompts` create and publish paths — the model-update path should too.

## Worth a follow-up

**4. `prompts-subworkspace.test.js` missing rollback-failure case**

`prompts.test.js` has it. The twin invariant (flat + subworkspace stay in lockstep) suggests this should be here too — could be a quick add before merge or a follow-up.

**5. Dedup counter increments but never re-posts**

`entry.count` tracks collapsed attempts but `formatMessage` always gets `count: 1`. The spec's "attempt counter" is tracked internally but never surfaced to the customer. Either surface it on re-post (requires timer/threshold logic) or document this as a TODO. As-is it's slightly misleading about what the dedup mechanism does.

---

Everything else looks clean: classification logic is solid across all sites, return-shape extension (`code` field) is backward-compat, alert dedup keys are well-scoped, and the backwards-splice loop in `reconcilePublishErrors` is correct. Type-check, bundle, and IT suite all pass.

MysticatBot (approved):
Hey @claudiaboldis,

**Verdict: Approve** - all prior blocking findings addressed; remaining items are non-blocking.
**Complexity: HIGH** - large diff (14 files, 2136 lines); API surface + new module.
_Changes: Addresses prior review feedback - adds bounded dedup eviction, redacts rollback errors before Slack, adds zero-staged fallback, adds syncModelsForProject quota test, and adds subworkspace rollback-failure test (14 files)._

- Dedup Map unbounded growth - `EVICT_SWEEP_THRESHOLD` + lazy `evictExpired()` sweep added
- Unredacted rollback error leaked to Slack - `redactUpstreamMessage(rollbackErr)` now used
- Zero-staged project silently drops quota signal - else-branch fallback added in `reconcilePublishErrors`
- syncModelsForProject publish 405-to-409 path untested - test added in `markets.test.js`
- Subworkspace missing rollback-failure test case - test added in `prompts-subworkspace.test.js`

<details>

- nit: `alertQuotaRejection` in `publishAffected` is the only call site that does not await - other sites (market-create, per-item metered write) await for Lambda-freeze safety; consider consistency - `src/support/serenity/handlers/prompts.js:293`
- nit: Dedup `count` is tallied but never surfaced to the Slack message (acknowledged via TODO) - the `formatMessage` function supports it (`count > 1` branch exists) but is always called with `count: 1` - `src/support/serenity/quota-alerts.js:147`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 5m 49s | <b>Cost:</b> $1.58 | <b>Commit:</b> <code>430263073eff481596f767c72bef3a663551a1ad</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2891, opened 2026-07-23
feat(serenity): make the per-brand AI ceiling bindable + tunable via Vault (LLMO-6190 gate)

byteclimber:
## Summary

Closes the mechanism half of the §8 flag-flip gate *"a per-brand ceiling with a BINDING value"* (serenity-docs `brand-semrush-dynamic-resource-allocation.md`). The enforcement mechanism was already merged (`ensureAiHeadroom` throws `brandAiLimit`/409 past the cap), but the shipped `DEFAULT_BRAND_AI_CEILING` is astronomical and **no call site passed a ceiling** — so the gate was definitionally open. This wires the ceiling so it can actually bind, resolved from Vault.

- **`resolveBrandAiCeiling(env)`** (`dynamic-allocation-active.js`, next to `isDynamicAllocationEnabled`) reads `SERENITY_BRAND_AI_CEILING_PROJECTS` / `_PROMPTS` off the same per-request env as the kill-switch. **Fail-safe** (mirrors the kill-switch's "anything non-canonical → safe default"): unset, or a malformed / non-positive value (a Vault typo like `5o00`, `-1`, `0`), is left unenforced and never breaks a customer write. Digit-only parse (`parseInt` would wrongly accept `5o00` as `5`, a dangerously low cap). Returns `undefined` when neither dimension is set → the guard keeps its non-binding default.
- **Controller** threads `ceiling: brandAiCeiling(ctx)` alongside `dynamicAllocation` into the three guard-building subworkspace handlers (create-prompts, create-market ×2, update-models); handlers pass it to `createHeadroomGuard`. Partial ceilings supported (a dimension may be omitted → unenforced).

### Design (settled in discussion)
A **single global backstop, the same for every brand** — "the largest plausible *legitimate* single-brand demand", a runaway circuit-breaker. Deliberately **not** a per-org pool fraction (a fraction false-fails a legitimate `800 + 200`-in-a-`1000`-pool split; `orgPoolExhausted` stays the demand-aware contention gate) and **not** per-customer config (that data model is deferred by the spec). The number itself is a product decision set in Vault later.

**Merging is a no-op**: unset → byte-for-byte today's behavior. The gate fully closes when a value is set in Vault (`dx_mysticat/<env>/api-service`), gated on the usage p99 estimate — a separate, product-owned step, deliberately not in this PR.

## Test plan
- [x] `npm run lint && npm run type-check` clean
- [x] `npm test` — full unit suite green (15843 passing)
- [x] Unit: `resolveBrandAiCeiling` (unset/valid/partial/malformed/mixed) + guard-override (binding cap → `brandAiLimit`, under-cap succeeds, partial ceiling)
- [x] Controller: threads ceiling from env; unset → `undefined`; malformed → `undefined` (never throws)
- [x] Handler→guard fronting: a binding ceiling throws `brandAiLimit` (409) before any write/transfer
- [x] Added an IT case (`test/it/shared/tests/serenity.js`, dynamic-allocation-ON block): low `SERENITY_BRAND_AI_CEILING_PROMPTS` → prompt write 409s over the wire. **Not run locally** (no ECR creds for the data-service image pull); runs in CI's `it-postgres` job.

### DIFF
--- a/src/controllers/serenity.js
+++ b/src/controllers/serenity.js
@@ -64,7 +64,7 @@ import {
 } from '../support/serenity/handlers/tags.js';
 import { ensureSubworkspace, decommissionBrandWorkspace } from '../support/serenity/workspace-lifecycle.js';
 import { isSerenityActiveForOrg } from '../support/serenity/serenity-active.js';
-import { isDynamicAllocationEnabled } from '../support/serenity/dynamic-allocation-active.js';
+import { isDynamicAllocationEnabled, resolveBrandAiCeiling } from '../support/serenity/dynamic-allocation-active.js';
 import { MAX_TOPICS_ON_CREATE } from '../support/serenity/brand-provisioning.js';
 import { marketForGeoTargetId } from '../support/serenity/locations.js';
 import { brandNeedles, classifyBrandedTag } from '../support/serenity/branded-classifier.js';
@@ -448,6 +448,12 @@ function SerenityController(context, log, env) {
   // handlers front through a no-op guard (byte-for-byte pre-PR behavior).
   const dynamicAllocationEnabled = (ctx) => isDynamicAllocationEnabled(ctx?.env || env);
 
+  // Global per-brand AI ceiling (LLMO-6190 flag-flip gate) — a runaway backstop, the SAME value
+  // for every brand, resolved from Vault off the same per-request env as the kill-switch above.
+  // `undefined` when unset → the guard keeps its non-binding default (byte-for-byte today). Passed
+  // as a plain object alongside `dynamicAllocation` at each guard-building handler call site.
+  const brandAiCeiling = (ctx) => resolveBrandAiCeiling(ctx?.env || env, log);
+
   /** Loads the Brand model instance (for subworkspace-mode write/lifecycle flows). */
   async function loadBrand(ctx, brandUuid) {
     const Brand = ctx?.dataAccess?.Brand;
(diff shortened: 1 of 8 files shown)


MysticatBot (approved):
Hey @byteclimber,

**Verdict: Approve** - clean, well-tested feature gate with no blocking issues.
**Complexity: HIGH** - medium diff; API surface signal.
_Changes: Wires the per-brand AI ceiling (LLMO-6190 gate) so it can bind from Vault, adding fail-safe env parsing via `resolveBrandAiCeiling` and threading the resolved ceiling through the controller to the subworkspace handler guard sites (8 files)._

<details>

- nit: `resolveBrandAiCeiling` is called (and may warn on a Vault typo) even when `dynamicAllocation` is OFF - consider guarding behind the same kill-switch or suppressing the warn when disabled - `src/support/serenity/dynamic-allocation-active.js:100`
- suggestion: controller-level tests assert `ceiling` threading for `createPrompts` but not for `updateModels` or the single-market `createMarket` path - handler-level tests cover enforcement thoroughly, but parity would pin the wiring - `test/controllers/serenity.test.js`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 39s | <b>Cost:</b> $5.86 | <b>Commit:</b> <code>afd9f4117ad55497a92b75c551f906da7b633bac</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2892, opened 2026-07-23
feat(configuration): add GET /configurations/versions endpoint

alinarublea:
## Why

The `spacecat-configuration` admin skill needed **direct S3/AWS-console access** for one remaining task: discovering the config **VersionIds** consumed by `GET /configurations/:version`, `POST /configurations/:version/restore`, and offline exports. Config history lives as S3 object versions with no REST way to list them.

This adds the missing list endpoint so the skill (and any admin) can discover versions through the API — no AWS access required.

## What

`GET /configurations/versions` — lists the config's S3 object-version history, newest first.

- **Static route**, matched before the dynamic `GET /configurations/:version` (added a route-precedence test so "versions" can never be captured as a VersionId).
- Gated by `authorizeConfigRead` (admin OR S2S `configuration:read`), mapped to `CAP_CONFIGURATION_READ`, added to the FACS admin route list (no new `:param`).
- Query params: `limit` (clamped 1–100), `keyMarker`/`versionIdMarker` (pagination), `detail` (default true → `updatedBy`/`updatedAt` enrichment).
- `ConfigurationDto.versionsToJSON`; OpenAPI path + response schema; controller/DTO/route tests.

## ⚠️ Depends on

- adobe/spacecat-shared#1839 (adds `ConfigurationCollection.listVersions`). **Merge/release that first**, then this PR bumps `@adobe/spacecat-shared-data-access` to the released version. Until then CI resolves the dep via the lockfile; the controller tests mock the collection so they don't require the new method at unit-test time.

### DIFF
--- a/docs/openapi/configuration-api.yaml
+++ b/docs/openapi/configuration-api.yaml
@@ -186,6 +186,119 @@ latest-configuration:
         description: Configuration not found.
       '500':
         $ref: './responses.yaml#/500'
+configuration-versions:
+  get:
+    tags:
+      - configuration
+    summary: List configuration versions (version history)
+    description: |
+      Lists the version history of the global configuration, newest first. Each
+      version corresponds to an S3 object version. Use the returned `versionId`
+      values with `GET /configurations/{version}` or
+      `POST /configurations/{version}/restore`. This removes the need for direct
+      S3/AWS access to discover VersionIds.
+
+      When `detail` is true (default) each version is enriched with
+      `updatedBy`/`updatedAt` read from the object's S3 user-metadata; versions
+      written before that metadata was introduced return `null` for those fields.
+    operationId: listConfigurationVersions
+    security:
+      - admin_key: [ ]
+    parameters:
+      - name: limit
+        in: query
+        required: false
+        description: Maximum number of versions to return (1–100).
+        schema:
+          type: integer
+          minimum: 1
+          maximum: 100
+          default: 25
+      - name: keyMarker
+        in: query
+        required: false
+        description: S3 KeyMarker for pagination (from a previous response's nextKeyMarker).
+        schema:
+          type: string
+      - name: versionIdMarker
+        in: query
+        required: false
+        description: S3 VersionIdMarker for pagination (from a previous response's nextVersionIdMarker).
+        schema:
+          type: string
+      - name: detail
+        in: query
+        required: false
+        description: When false, skips updatedBy/updatedAt enrichment for a faster listing.
+        schema:
+          type: boolean
+          default: true
+    responses:
+      '200':
+        description: A page of configuration versions, newest first.
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                versions:
+                  type: array
+                  items:
+                    type: object
+                    properties:
+                      versionId:
+                        type: string
+                        example: "EqZSCYxbPdKJ0Xso9QkbS0h0ZfDbhXWe"
+                      lastModified:
+                        type: string
+                        format: date-time
+                        example: "2026-07-23T10:00:00.000Z"
+                      isLatest:
+                        type: boolean
+                        example: true
+                      size:
+                        type: integer
+                        example: 3120487
+                      updatedBy:
+                        type: [string, 'null']
+                        example: "admin@adobe.com"
+                      updatedAt:
+                        type: [string, 'null']
+                        format: date-time
+                        example: "2026-07-23T10:00:00.000Z"
+                isTruncated:
+                  type: boolean
+                  example: false
+                nextKeyMarker:
+                  type: [string, 'null']
+                nextVersionIdMarker:
+                  type: [string, 'null']
+            examples:
+              versions-list:
+                summary: A page of versions
+                value:
+                  versions:
+                    - versionId: "EqZSCYxbPdKJ0Xso9QkbS0h0ZfDbhXWe"
+                      lastModified: "2026-07-23T10:00:00.000Z"
+                      isLatest: true
+                      size: 3120487
+                      updatedBy: "admin@adobe.com"
+                      updatedAt: "2026-07-23T10:00:00.000Z"
+                    - versionId: "SzrMxECGfivkXWmCK2zlmzHrc0BVoSdR"
+                      lastModified: "2026-07-22T09:00:00.000Z"
+                      isLatest: false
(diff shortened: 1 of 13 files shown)


src/controllers/configuration.js
```diff
@@ -159,6 +159,49 @@ function ConfigurationController(ctx) {
...
+    const rawLimit = searchParams.get('limit');
+    const parsedLimit = Number.parseInt(rawLimit, 10);
+    // Clamp to [1, 100]; fall back to the collection default when unparseable.
+    const limit = Number.isInteger(parsedLimit)
+      ? Math.min(Math.max(parsedLimit, 1), 100)
+      : undefined;
+
+    const keyMarker = searchParams.get('keyMarker') || undefined;
+    const versionIdMarker = searchParams.get('versionIdMarker') || undefined;
+    // Enrichment (updatedBy/updatedAt) is on by default; opt out with detail=false.
```
MysticatBot, line 201: issue (blocking): **Inconsistent conditional-spread for pagination markers.**

`limit` uses a conditional spread (`...(limit !== undefined ? { limit } : {})`) to avoid sending an explicit `undefined` key to the data-access layer. But `keyMarker` and `versionIdMarker` are passed as bare properties, always present in the object even when `undefined`:

```js
const page = await Configuration.listVersions({
  ...(limit !== undefined ? { limit } : {}),
  keyMarker,       // always present, even when undefined
  versionIdMarker, // always present, even when undefined
  detail,
});
```

If `ConfigurationCollection.listVersions` uses `'keyMarker' in opts` (a common pattern for 'was this option provided?'), the explicit `undefined` key satisfies that check and could change behavior.

**Fix:** Apply the same conditional-spread pattern:
```js
...(keyMarker !== undefined ? { keyMarker } : {}),
...(versionIdMarker !== undefined ? { versionIdMarker } : {}),
```
Or filter undefined values before passing.

docs/openapi/configuration-api.yaml
```diff
@@ -186,6 +186,119 @@ latest-configuration:
...
+      - name: limit
+        in: query
+        required: false
+        description: Maximum number of versions to return (1–100).
+        schema:
+          type: integer
+          minimum: 1
+          maximum: 100
+          default: 25
+      - name: keyMarker
```
MysticatBot, line 217: issue (blocking): **OpenAPI `default: 25` not enforced in controller code.**

The spec declares `default: 25` for the `limit` parameter, but the controller passes `undefined` to the data-access layer when no limit is provided, delegating to whatever default the collection method uses internally.

If that collection default is not 25, consumers relying on the documented spec will get unexpected page sizes.

**Fix (pick one):**
1. Enforce in controller: change the fallback to `const limit = ... : 25` so code and spec agree.
2. Remove the `default: 25` from the spec and document it as 'server-defined default' (less consumer-friendly but avoids the coupling).

MysticatBot (changes requested):
Hey @alinarublea,

**Verdict: Request changes** - two issues to address before merge (one code consistency, one spec/code mismatch).
**Complexity: MEDIUM** - medium diff; new API endpoint with FACS/route-precedence concerns.
_Changes: Adds `GET /configurations/versions` endpoint to list S3 object-version history for admin use, removing the need for direct AWS access (12 files)._

## Must fix before merge
1. [Important] Inconsistent conditional-spread for pagination markers - `src/controllers/configuration.js:341` (details inline)
2. [Important] OpenAPI `default: 25` for limit is not enforced in controller code - `docs/openapi/configuration-api.yaml:217` (details inline)

<details>

- nit: `facs-capabilities.js` places the static route after the dynamic `:version` pattern, inconsistent with ordering in `index.js` and `required-capabilities.js` - `src/routes/facs-capabilities.js:241`
- nit: OpenAPI response schema does not declare `required` properties, so generated types cannot distinguish guaranteed from optional fields - `docs/openapi/configuration-api.yaml:244`
- suggestion: Add a test with `Configuration.listVersions.rejects(...)` to prove the error path surfaces a clean 500, not a raw stack trace - `test/controllers/configurations.test.js`
- suggestion: Add a test with `{ request: { url: 'not-a-valid-url' } }` to prove the URL-parse fallback explicitly covers malformed URLs (not just missing context) - `test/controllers/configurations.test.js`
- suggestion: Add a reasonable length check on `keyMarker`/`versionIdMarker` (e.g., reject > 1024 chars) as defense-in-depth, even though the endpoint is admin-gated - `src/controllers/configuration.js:341`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 5m 22s | <b>Cost:</b> $6.62 | <b>Commit:</b> <code>bcdd90eb92661ae845242f48bfdf01780d649db8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @alinarublea,

**Verdict: Approve** - all prior blocking findings addressed, no new issues.
**Complexity: MEDIUM** - medium diff; API surface + FACS route concerns.
_Changes: Adds `GET /configurations/versions` endpoint to list S3 object-version history for admin use, with pagination, input validation, and full test coverage (12 files)._

Note: CI checks are currently pending - verify they pass before merge.

- Inconsistent conditional-spread for pagination markers - now uses conditional spread for both `keyMarker` and `versionIdMarker`
- OpenAPI `default: 25` not enforced in controller - limit now defaults to 25 in code, matching the spec
- OpenAPI response schema missing `required` properties - now declared on both envelope and item objects
- FACS route ordering (static after dynamic) - `GET /configurations/versions` now listed before `GET /configurations/:version`
- Marker length validation added (defense-in-depth, 1024 char cap)
- Tests added for error propagation path and malformed URL fallback

<details>

- suggestion: Consider documenting the `detail` query param default in the OpenAPI `description` field (it says enrichment is on by default in the summary but the param schema has no `default: true`) - `docs/openapi/configuration-api.yaml`

</details>

---

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 58s | <b>Cost:</b> $9.54 | <b>Commit:</b> <code>5b9dd1a0369130f49492bafaf71e5b0f6468f5be</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2894, opened 2026-07-23
fix: validate temporalCondition strictly to close SQL injection in paid traffic endpoints | VULN-37317

Sihalnic-Alin:
## What & why
[VULN-37317](https://jira.corp.adobe.com/browse/VULN-37317) — Authenticated SQL injection on `/api/v1/sites/{siteId}/traffic/paid/impact-by-page` via `temporalCondition`.

`fetchTop3PagesTrafficData` (and every endpoint routed through it) built its Athena SQL query by interpolating the `temporalCondition` query param directly into the `WHERE` clause, validated only by checking the decoded string contained the substrings `"week"` and `"year"`. HackerOne report #3880600 showed this could be bypassed with a boolean-oracle payload (`IF(1=1, CAST('n4z8c6q2' AS INTEGER), 1) = 1`) to blind-inject SQL, and the same shape (closing paren + `OR`) could in principle escape the site-scoped filter for cross-tenant disclosure. This PR closes that path by parsing the parameter into validated integers and rebuilding the SQL fragment server-side, so no attacker-controlled text ever reaches the query.

## Scope
- `parseTemporalCondition` (`src/controllers/paid/traffic.js`) replaces the substring check with a regex-anchored parse of `(week=N AND year=N)` clauses (either order, flexible whitespace, digits-only operands), returning `null` for anything that doesn't match exactly — including the reported exploit payload and a closing-paren/`OR` escape attempt.
- The Athena `WHERE` fragment is rebuilt as `(week=${week} AND year=${year})` joined by `OR`, from the parsed integers only — the raw client string is no longer used in the query, the log line, or the query description.
- `MAX_TEMPORAL_CLAUSES` (8) caps the number of `OR`-joined clauses, closing a secondary "query resource abuse" angle the same report called out — the documented use case (`docs/openapi/site-paid.yaml`) is a 4-week window, so 8 leaves headroom without allowing an unbounded chain.
- Applies to all five endpoints that route through `fetchTop3PagesTrafficData`: `getImpactByPage`, `getImpactByPageTrafficType`, `getImpactByPageDevice`, `getImpactByPageTrafficTypeDevice`, `getTrafficLossByDevices`.

## Deferred / out of scope
- `traffic-tools.js`'s `getPredominantTraffic` builds its own `temporalCondition` server-side (`generateTemporalCondition()`) and never accepts client input for it — unaffected by the report, not touched here.
- No OpenAPI/contract change — the parameter's shape and semantics for legitimate callers are unchanged; only invalid or malicious input is now rejected (`400 Invalid temporal condition`) instead of reaching Athena.

## Tests
Extended `test/controllers/paid/traffic.test.js` with four new cases: rejects the reported exploit payload (asserts `400` and that Athena is never called), rejects a closing-paren/`OR` escape attempt, rejects a `temporalCondition` over the 8-clause cap, and accepts one exactly at the cap. Full suite: **86 passing**. Lint clean on both changed files (aside from pre-existing CRLF/header-year noise unrelated to this diff).

### DIFF
--- a/src/controllers/paid/traffic.js
+++ b/src/controllers/paid/traffic.js
@@ -329,6 +329,7 @@ function TrafficController(context, log, env) {
 
     const parsedTemporalCondition = parseTemporalCondition(decodedTemporalCondition);
     if (!parsedTemporalCondition) {
+      log.info(`Rejected request with invalid temporalCondition | siteId: ${siteId} | reason: failed strict parse/shape validation | requestId: ${requestId}`);
       return badRequest('Invalid temporal condition');
     }
 
--- a/src/controllers/paid/traffic.js
+++ b/src/controllers/paid/traffic.js
@@ -88,6 +88,51 @@ function validateTemporalParams({ year, week, month }) {
 
 const isTrue = (value) => value === true || value === 'true' || value === '1' || value === 1;
 
+// Matches a single "(week=W AND year=Y)" clause (week/year order and internal
+// whitespace both flexible) with numeric-only operands.
+const TEMPORAL_CLAUSE_PATTERN = /^\(\s*(?:week\s*=\s*(\d{1,2})\s*AND\s*year\s*=\s*(\d{4})|year\s*=\s*(\d{4})\s*AND\s*week\s*=\s*(\d{1,2}))\s*\)$/i;
+
+// The documented use case is a 4-week window (see docs/openapi/site-paid.yaml); this
+// caps clause count well above that to block resource-abuse via a huge OR-chain
+// while still allowing legitimate callers plenty of headroom.
+const MAX_TEMPORAL_CLAUSES = 8;
+
+/**
+ * Strictly parses a client-supplied temporal condition string into a list of
+ * {week, year} integer pairs, rejecting anything that doesn't match the exact
+ * expected shape. This is the only path by which the value is later used to
+ * build the Athena SQL WHERE clause, so no attacker-controlled text ever
+ * reaches the query - only integers we validated and re-serialize ourselves.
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @Sihalnic-Alin,

**Verdict: Approve** - solid parse-then-reconstruct fix that definitively closes the injection vector.
**Complexity: LOW** - small, focused security fix in a single controller.
_Changes: Replaces unsafe substring-check validation of `temporalCondition` with strict regex parsing and integer-only re-serialization, preventing SQL injection in all five paid traffic endpoints (2 files)._

<details>

- suggestion: Add an assertion on `mockAthenaQuery`'s call argument in the acceptance-path test to verify the rebuilt safe string is what reaches Athena (guards against future regressions that silently re-introduce the raw input) - `test/controllers/paid/traffic.test.js:1268`
- suggestion: Consider a `log.info` when validation rejects a request (with a generic reason tag, not the raw input) to give operators visibility into whether legitimate callers hit the stricter validation post-deploy - `src/controllers/paid/traffic.js:332`
- nit: The 8-clause acceptance test only exercises `week-first` ordering; adding one `year-first` clause would cover the regex's alternate capture group path - `test/controllers/paid/traffic.test.js:1268`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 4s | <b>Cost:</b> $6.50 | <b>Commit:</b> <code>2b1c0c1f8606ded9f653f3a9674e72deeb7130b5</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

vivesing (approved):

PR #2895, opened 2026-07-23
fix: reject single-quote characters in pageTypes pattern to close stored SQL injection | VULN-37318

Sihalnic-Alin:
## What & why
[VULN-37318](https://jira.corp.adobe.com/browse/VULN-37318) — stored SQL injection via `pageTypes[].pattern`.

A HackerOne report showed that an authenticated user could store a malicious Athena SQL expression in a site's `pageTypes[].pattern` via `PATCH /api/v1/sites/{siteId}`. The field was only validated as a syntactically-valid JS `RegExp`, with no check that it was also safe to splice into a SQL string literal. The stored pattern is later inserted unescaped into `REGEXP_LIKE(path, '<pattern>')` when a paid-traffic report with the `page_type` dimension is generated (`GET /sites/{siteId}/traffic/paid/page-type`), letting an attacker break out of the string literal and inject arbitrary SQL — confirmed via `INVALID_CAST_ARGUMENT`, `COLUMN_NOT_FOUND`, and `FUNCTION_NOT_FOUND` Athena error oracles. This is the same class of issue as the `temporalCondition` injection fixed in VULN-37317, but a different stored-config source and query sink.

## Scope
- `validatePageTypes` (`src/controllers/sites.js`) now rejects any `pageTypes[].pattern` containing a single quote character before it can ever be persisted via `site.setPageTypes()` — Trino/Presto string literals have no backslash-escape mode, so `'` is the only character capable of terminating the literal early and injecting SQL.
- New `SQL_UNSAFE_PATTERN_CHAR` check runs before the existing `new RegExp()` validity check, returning `400` with a clear error (`pageTypes[N] pattern must not contain a single quote character`) instead of silently accepting the value.
- Added a regression test in `test/controllers/sites.test.js` that reproduces the exact reported PoC payload (`x' || CAST(CAST(current_schema AS INTEGER) AS VARCHAR) || 'x`) and asserts it's rejected with `400`.

## Deferred / out of scope
- Escaping `pattern` at the actual query-building sink (`buildPageTypeCase` in `@adobe/spacecat-shared-athena-client`'s `queries.js`, which already escapes the sibling `name` field via `.replace(/'/g, "''")` but not `pattern`) is not done here — that code lives in the `spacecat-shared` repo, a separate vendored dependency. Recommend a follow-up there for defense-in-depth, so the query builder itself is safe even if a future write path bypasses `validatePageTypes`.
- No retroactive audit/cleanup of already-stored `pageTypes` records that may still contain a quote in `pattern` from before this fix (e.g. the site referenced in the report). That's an operational/data task against production, not a code change, and out of scope for this PR.

## ⚠️ One review item
Rejecting any `'` outright (rather than escaping it) means a legitimate pattern that needed to match a literal apostrophe in a URL path would now be turned away with `400`. This seems like an acceptable trade-off for a regex matched against URL paths, but flagging it as a deliberate behavior change worth a second opinion.

## Tests
- Added 1 new test to the `pageTypes validation` suite in `test/controllers/sites.test.js` (now 11 tests in that block, all passing).
- Full `sites.test.js` suite: 409 passing, 1 pre-existing failure (`triggerBrandProfile` — an unrelated `esmock`/`X_PRODUCT_HEADER` loading issue, confirmed present on unmodified `main` as well, not caused by this change).
- Lint: no new issues introduced by this diff (pre-existing CRLF/header lint noise from the Windows checkout is present identically before and after this change).

### DIFF
--- a/src/controllers/sites.js
+++ b/src/controllers/sites.js
@@ -409,6 +409,12 @@ function SitesController(ctx, log, env) {
     if (productCodeError) {
       return badRequest(productCodeError);
     }
+    if (isArray(context.data?.pageTypes)) {
+      const validation = validatePageTypes(context.data.pageTypes);
+      if (!validation.isValid) {
+        return badRequest(validation.error);
+      }
+    }
     let site;
     let status;
     try {
--- a/src/controllers/sites.js
+++ b/src/controllers/sites.js
@@ -306,6 +306,13 @@ const applyTopOrganicPagesFilter = async (metricsData, limit, options) => {
   return result;
 };
 
+// pageTypes[].pattern is later spliced, unescaped, into an Athena SQL string literal
+// (REGEXP_LIKE(column, '<pattern>')) by @adobe/spacecat-shared-athena-client. Trino/Presto
+// string literals have no backslash-escape mode, so a bare single quote is the only
+// character that can terminate the literal early and inject SQL - reject it outright
+// rather than trying to escape/re-quote a value we don't control the eventual sink for.
+const SQL_UNSAFE_PATTERN_CHAR = /'/;
+
 /**
  * Validates that pageTypes array contains valid regex patterns
  * @param {Array} pageTypes - Array of page type objects with name and pattern
@@ -323,6 +330,13 @@ const validatePageTypes = (pageTypes) => {
 
     if (!hasText(pageType.pattern)) {
       return { isValid: false, error: `pageTypes[${index}] must have a pattern` };
+    }
+
+    if (SQL_UNSAFE_PATTERN_CHAR.test(pageType.pattern)) {
+      return {
+        isValid: false,
+        error: `pageTypes[${index}] pattern must not contain a single quote character`,
+      };
     }
 
     try {
--- a/test/controllers/sites.test.js
+++ b/test/controllers/sites.test.js
@@ -298,6 +298,38 @@ describe('Sites Controller', () => {
     expect(site).to.have.property('baseURL', 'https://site1.com');
   });
 
+  it('returns bad request when creating a site with an unsafe pageTypes pattern', async () => {
+    mockDataAccess.Site.findByBaseURL.resolves(null);
+
+    const response = await sitesController.createSite({
+      data: {
+        baseURL: 'https://site1.com',
+        pageTypes: [
+          { name: 'Schema probe', pattern: "x' || CAST(CAST(current_schema AS INTEGER) AS VARCHAR) || 'x" },
+        ],
+      },
+    });
+
+    expect(mockDataAccess.Site.create).to.have.not.been.called;
+    expect(response.status).to.equal(400);
+    const error = await response.json();
+    expect(error).to.have.property('message', 'pageTypes[0] pattern must not contain a single quote character');
+  });
+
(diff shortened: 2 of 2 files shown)


src/controllers/sites.js
```diff
@@ -325,6 +332,13 @@ const validatePageTypes = (pageTypes) => {
       return { isValid: false, error: `pageTypes[${index}] must have a pattern` };
     }
 
+    if (SQL_UNSAFE_PATTERN_CHAR.test(pageType.pattern)) {
```
MysticatBot, line 335: issue (blocking): `validatePageTypes` is called in `updateSite`, but `createSite` (POST /sites, line ~410 in this file) passes `context.data` directly to `Site.create()` without calling `validatePageTypes`. An attacker with site-create privileges (admin or S2S with `CAP_SITE_CREATE`) can bypass this fix by creating a new site with a malicious `pageTypes[].pattern` containing single quotes.

Fix: Add the same validation in `createSite` before `Site.create()`:
```js
if (isArray(context.data?.pageTypes)) {
  const validation = validatePageTypes(context.data.pageTypes);
  if (!validation.isValid) {
    return badRequest(validation.error);
  }
}
```
Add a corresponding test for the create path.

MysticatBot (changes requested):
Hey @Sihalnic-Alin,

**Verdict: Request changes** - one validation gap leaves the injection vector partially open through the create path.
_Changes: Adds input validation to reject single-quote characters in `pageTypes[].pattern` to close a stored SQL injection via VULN-37318 (2 files)._

## Must fix before merge
1. [Important] `createSite` bypasses the new `SQL_UNSAFE_PATTERN_CHAR` validation - `src/controllers/sites.js:~410` (details inline)

<details>

- suggestion: Add minimal boundary-case tests (a pattern of just `"'"` and `"foo'bar"`) alongside the PoC payload test to document the exact rejection surface and guard against accidental regex regressions - `test/controllers/sites.test.js:5277`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 10s | <b>Cost:</b> $4.41 | <b>Commit:</b> <code>c288f96fd602d15b89b844fec9a9689a32b3e4d2</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @Sihalnic-Alin,

**Verdict: Approve** - prior findings addressed, fix is complete and well-tested.
_Changes: Closes the createSite validation gap and adds boundary-case tests for the SQL injection fix in pageTypes pattern (2 files)._

- createSite path now validates pageTypes before persistence, matching the updateSite path
- Boundary-case tests added (quote-only, mid-string quote) alongside the injection-payload test

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 2s | <b>Cost:</b> $5.53 | <b>Commit:</b> <code>a3432508b09a759139a11bbd090da295bdd47e8d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

vivesing (approved):

PR #2903, opened 2026-07-24
fix(ci): skip type-check job on push events

iuliag:
## 1. Abstract
Guard the `type-check` job so it only runs on `pull_request` events, matching the pattern already used by `protect-nyc-config` and `validate-pr-title`.

## 2. Reasoning
Investigating why several CI checks (`it-postgres`, `type-check`, `build`) show up twice on PRs: `ci.yaml` triggers on both an unrestricted `push` and `pull_request`. When a branch already has an open, same-repo PR, pushing to it fires both events for the same commit, and any job with no event-name guard runs in both resulting workflow runs. `type-check` has no downstream consumers — nothing depends on its output — so its duplicate run is pure waste with no trade-off, unlike `build`/`it-postgres`, which are also required by push-triggered deploy jobs (`branch-deploy`, `deploy-stage`, `semantic-release`) and therefore can't be guarded the same way without losing the dev-branch-deploy-on-push guarantee. This PR only fixes the no-trade-off case.

Before: `type-check` ran once per `push` event and once per `pull_request` event for the same commit whenever the branch had an open PR, doubling its CI time/compute for identical output.

After: `type-check` only runs under `pull_request`, exactly like `protect-nyc-config`/`validate-pr-title` in the shared `service-ci.yaml`. No change to the check's own behavior — every commit is still type-checked before merge, just once instead of twice.

Confirmed the fix mirrors an already-proven pattern: `protect-nyc-config` and `validate-pr-title` in `adobe/mysticat-ci`'s `service-ci.yaml` use the identical `if: github.event_name == 'pull_request'` guard and are the only two checks in the screenshot that did *not* duplicate. Also found and fixed a stale local `node_modules` (missing `tsc` binary) via `npm install` while validating the pre-commit hook chain — unrelated to this change but needed to get `type-check`/`docs:lint` pre-commit hooks passing locally.

## 7. Test plan
(a) Ran `npm run type-check` locally — passes. GitHub Actions trigger semantics can't be exercised locally; confidence instead comes from the guard syntax being identical to the already-working `protect-nyc-config`/`validate-pr-title` jobs in the same reusable workflow.
(b) Verification on merge: push a commit to a branch with an open PR against `spacecat-api-service` and confirm the PR's checks list shows `type-check` exactly once (under `pull_request`), with no accompanying `push`-triggered duplicate. Separately confirm `type-check` still runs and gates merge on `opened`/`synchronize`.


rainer-friederich (approved):
LGTM

MysticatBot (approved):
Hey @iuliag,

**Verdict: Approve** - clean, well-scoped change that eliminates redundant CI compute.
**Complexity: TRIVIAL** - 1 line, single CI config file.
_Changes: Guards the type-check job to skip push events, eliminating duplicate runs when a branch has an open PR (1 file)._

<details>

- suggestion: ADR `docs/decisions/005-opt-in-type-checking.md` describes type-check as "enforced in CI (blocking)" without specifying trigger scope - consider adding a parenthetical clarifying that enforcement is via PR required-status-check, not all push events - `docs/decisions/005-opt-in-type-checking.md`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 10s | <b>Cost:</b> $2.50 | <b>Commit:</b> <code>15a78d70d15b62e670947a2146d25726c2f34c09</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2906, opened 2026-07-27
fix: bypass auto_fix subService gate for non-ASO products (LLMO-6553)

ravverma:
## Problem

The auto-fix endpoint (`PATCH /sites/{siteId}/opportunities/{opportunityId}/suggestions/auto-fix`) returns a **403** for users who authenticate via **LLMO**:

```json
{ "message": "User does not belong to the organization or does not have sufficient permissions" }
```

Reported against the Adobe Showcase Org — see [LLMO-6553](https://jira.corp.adobe.com/browse/LLMO-6553).

## Root cause

`autofixSuggestions` in `src/controllers/suggestions.js` gated access with:

```js
accessControlUtil.hasAccess(site, 'auto_fix')
```

When a `subService` is supplied, `AccessControlUtil.hasAccess()` requires the `dx_aem_perf_auto_fix` **user scope**:

```js
return hasOrgAccess && authInfo.hasScope('user', `${SERVICE_CODE}_${subService}`);
```

That scope (and its `auto_fix` profile) is only minted when the user logs in through **ASO** (`dx_aem_perf`). LLMO logins never carry it, so the check fails even for users who legitimately belong to the org.

## Fix

Make the `auto_fix` subService gate ASO-specific. Read the `x-product` header and only pass the `'auto_fix'` subService when the product is `ASO`; for any other product (LLMO) fall back to plain org-membership access via `hasAccess(site)`.

```js
const xProduct = context.pathInfo?.headers?.[X_PRODUCT_HEADER];
const autoFixSubService = xProduct === 'ASO' ? 'auto_fix' : '';
...
} else if (!await accessControlUtil.hasAccess(site, autoFixSubService)) {
```

Org membership is still enforced in every case; only the ASO-specific extra scope requirement is scoped to ASO. The S2S capability path is unchanged.

## Testing

- Added two unit tests in `test/controllers/suggestions.test.js`:
  - non-ASO product (LLMO) → `hasAccess` called without a subService (bypass)
  - ASO product → `hasAccess` still called with `'auto_fix'` (enforcement preserved)
- `npx mocha test/controllers/suggestions.test.js -g "auto-fix suggestions"` → 66 passing
- ESLint clean on changed files

### DIFF
--- a/src/controllers/suggestions.js
+++ b/src/controllers/suggestions.js
@@ -58,7 +58,7 @@ import {
   getIsSummitPlgEnabled,
   isViewAsTrialRequest,
 } from '../support/utils.js';
-import AccessControlUtil from '../support/access-control-util.js';
+import AccessControlUtil, { X_PRODUCT_HEADER } from '../support/access-control-util.js';
 import { redactFeedbackContent } from '../support/feedback-redaction.js';
 import { CAP_FIX_ENTITY_CREATE, CAP_SUGGESTION_WRITE } from '../routes/capability-constants.js';
 import { grantSuggestionsForOpportunity } from '../support/grant-suggestions-handler.js';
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @ravverma,

**Verdict: Approve** - well-scoped fix with clear root cause, correct implementation, and adequate test coverage.
**Complexity: MEDIUM** - small diff; auth-gated controller logic with API surface risk.
_Changes: makes the `auto_fix` subService gate ASO-specific so LLMO logins fall back to org-membership access instead of being incorrectly 403'd (2 files)._

No blocking issues found. The `x-product` header is infrastructure-injected by Fastly (overwriting any client-supplied value), so the trust decision is sound. Passing an empty string to `hasAccess` is consistent with the method's default parameter design and correctly degrades to org-membership-only access.

<details>

- suggestion: Consider adding a test for the missing-header case (`x-product` absent/undefined) to document that the default behavior is org-membership-only access - `src/controllers/suggestions.js:1317`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 8m 56s | <b>Cost:</b> $5.42 | <b>Commit:</b> <code>684ac188e575d5e4441368498325f1cff2de23fd</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2907, opened 2026-07-27
fix(serenity): soft-fail Semrush re-sync on brand URL edit (LLMO-6545)

aenascut:
## Summary

`PATCH /brands/:id` returned an error when the Semrush re-sync failed after the brand row had already been committed, so an edit that persisted in LLMO was reported to the customer as a failure. This blocked KDDI from saving brand data.

The re-sync is now best-effort, matching the create path. The committed row is returned with `semrushSyncPending: true`, the failure is logged with the context needed to enumerate and recover the drift, and the divergence is reconciled later.

## Root cause

The create path treats Semrush propagation as best-effort so a benchmark hiccup cannot strand a half-provisioned brand. The edit path hard-failed instead, on the reasoning that an already-live brand must not silently diverge from Semrush. That reasoning inverts once the row is written: the DB commit happens before the re-sync block, so the only thing a hard fail changes is what the customer is told about a write that already succeeded.

Three call sites sat outside the re-sync's error handling and produced the same asymmetry:

- the pre-write self-referential-competitor guard, which lists Semrush projects to build the reserved-domain set
- Semrush IMS token resolution for the post-commit re-sync
- transport construction for that re-sync

## Who this affects

The driving failure is **permanent, not transient**. Semrush binds a forwarded IMS token to a Semrush user record by `authId` or email before executing the call, and JIT user provisioning from IMS claims was explicitly declined — users are added to Semrush manually at onboarding. Migrated orgs therefore have LLMO users who edit brand content in LLMO but have no Semrush user record at all, and the observed upstream status for such an identity is 403.

For those users every edit drifts, on every request, indefinitely. Retrying cannot clear it; someone has to provision the user. Transient upstream failures (5xx, timeouts, network) are a separate and much smaller class.

## Changes

**`src/controllers/brands.js`**

- The post-commit re-sync returns `ok({ ...updated, semrushSyncPending: true })` instead of rethrowing. Any `semrushRejectedAliases` collected before the throw are preserved, so a partial result is not discarded.
- Token resolution and transport construction moved inside the re-sync `try`. A 401 there (no or expired promise token, non-IMS bearer) is a re-sync failure like any other and no longer reports an auth error for an edit that persisted. A non-IMS bearer is still refused before any transport is built or forwarded upstream.
- The pre-write competitor guard degrades instead of failing the request: on a listing error it falls back to a brand-URL-only reserved set and logs a warning. The guard is a data-quality nicety, and a Semrush outage should not reject the write outright. The trade-off is explicit — a competitor equal to one of the brand's own market domains can be stored while the listing is unavailable, which the next successful edit re-checks.
- Failures are classified by class. 401/403 logs at `warn` with `permanent: true`; everything else logs at `error`. Both carry `brandId`, `semrushSubWorkspaceId`, which syncs were touched, `status`, `message` and `stack`.

**`docs/openapi/schemas.yaml`, `docs/openapi/brands-v2-api.yaml`**

`semrushSyncPending` documented on `V2BrandUpdateResponse` and in the 200 description, alongside the existing `semrushRejectedAliases`. The semantic is load-bearing: the row persisted and Semrush does not reflect it yet.

## Observability

Recovery reads from **CloudWatch, not Splunk**. Measured on 2026-07-27, api-service prod ships about 7 events per 24h to `index=dx_aem_engineering sourcetype=dx_aem_sites_spacecat_backend_prod`, against 6,920 for content-scraper and 3,599 for audit-worker over the same window, and a search for this log line across `sourcetype=dx_aem_sites_spacecat_*` returns nothing. A Splunk-based enumeration of drifted brands would silently return an empty set. Use CloudWatch Logs Insights against the api-service prod Lambda log group:

```
fields @timestamp, brandId, semrushSubWorkspaceId, urlsTouched, competitorsTouched, aliasesTouched, status, permanent
| filter @message like "serenity: brand-edit Semrush re-sync failed after row commit"
| sort @timestamp desc
```

The permanent class warrants a standing alert: it is not self-healing, it means a named user needs provisioning in Semrush, and it does not carry the noise that would make alerting on transient upstream errors impractical. Filter on `permanent: true` to separate the two. `semrushSyncPending` has no consumer yet — no client currently reads the flag, so today the divergence is visible only in the logs.

## Recovery

Extend the `--reconcile` migration in `mysticat-data-service` to diff and reconcile `brand_urls`, `competitors` and `aliases` per Semrush project. The Semrush client methods already exist (`list_brand_urls`, `create_brand_urls`, and siblings). Tracked on LLMO-6545.

Enqueuing the re-sync for retry alongside the 200, per the repo's queue-based async pattern, would let transient failures self-heal and make `--reconcile` a backstop rather than the only recovery. Tracked as a follow-up on LLMO-6545.

## Testing

- `npx mocha test/controllers/brands.test.js` — 412 passing
- `npx eslint src/controllers/brands.js test/controllers/brands.test.js` — clean
- `npm run type-check` — clean

Covered:

- Semrush re-sync throws after commit gives 200 with `semrushSyncPending: true`
- a partial `semrushRejectedAliases` set collected before the throw survives into the response
- a non-IMS caller never has a bearer built into a transport or forwarded upstream, and the refusal soft-fails at the permanent class rather than returning 401 for a committed edit
- a pre-write listing failure degrades to the brand-URL-only reserved set and the write proceeds
- a Semrush error on the create path is mapped to a generic message, with the internal gateway URL kept out of both the response body and the `x-error` header

## What does not change

- The feature-flag gate (`isSerenityActiveForOrg`) still returns 403 on PATCH when the flag is off.
- The POST create path is untouched; it was already best-effort.
- Serenity activate/deactivate paths are already graceful and self-healing, and are out of scope.

### DIFF
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -2071,8 +2071,16 @@ function BrandsController(ctx, log, env) {
             competitorsTouched,
             aliasesTouched,
             status: syncError?.status,
+            message: syncError?.message,
+            stack: syncError?.stack,
           });
-          return ok({ ...updated, semrushSyncPending: true });
+          // Preserve any partial rejectedAliases collected before the throw so
+          // the UI can still warn about aliases that were refused by Semrush.
+          return ok({
+            ...updated,
+            semrushSyncPending: true,
+            ...(rejectedAliases.length > 0 && { semrushRejectedAliases: rejectedAliases }),
+          });
         }
       }
 
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -2061,10 +2061,9 @@ function BrandsController(ctx, log, env) {
             rejectedAliases.push(...(aliasResult?.rejected ?? []));
           }
         } catch (syncError) {
-          // The brand row is already committed; re-sync hard-fails (the brand
-          // must not silently drift out of sync with Semrush). Log the upstream
-          // context (workspace + which sync) so the DB/Semrush divergence is
-          // diagnosable, then rethrow to the handler's catch.
+          // LLMO-6545: Accept drift — DB is already committed so hard-failing returns
+          // a 500 while the edit actually succeeded. Log all context so the divergence
+          // is diagnosable via Splunk and recoverable via --reconcile migration later.
           log.error('serenity: brand-edit Semrush re-sync failed after row commit', {
             brandId,
             semrushSubWorkspaceId: updated.semrushSubWorkspaceId,
@@ -2073,7 +2072,7 @@ function BrandsController(ctx, log, env) {
             aliasesTouched,
             status: syncError?.status,
           });
-          throw syncError;
+          return ok({ ...updated, semrushSyncPending: true });
         }
       }
 
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -1945,11 +1945,23 @@ function BrandsController(ctx, log, env) {
         if (hasText(brandState?.semrushSubWorkspaceId)) {
           // Semrush brand: market/project domains come from the project listing.
           // List once and stash for the post-commit re-sync (see prefetchedProjects).
-          const imsToken = await resolveSemrushImsToken(context, log, 'brands');
-          const transport = createSerenityTransport({ env: context.env, imsToken });
-          prefetchedProjects = await resolveProjects(transport, brandState.semrushSubWorkspaceId);
+          // Best-effort: if the listing fails (e.g. user not provisioned in Semrush),
+          // degrade to brand-URL-only reserved set and log — the self-reference guard
+          // is a data-quality nicety and must not block the competitor write entirely.
+          try {
+            const imsToken = await resolveSemrushImsToken(context, log, 'brands');
+            const transport = createSerenityTransport({ env: context.env, imsToken });
+            prefetchedProjects = await resolveProjects(transport, brandState.semrushSubWorkspaceId);
+          } catch (guardError) {
+            log.warn('serenity: competitor-guard project listing failed; degrading to brand-URL-only reserved set', {
+              brandId,
+              semrushSubWorkspaceId: brandState.semrushSubWorkspaceId,
+              status: guardError?.status,
+              message: guardError?.message,
+            });
+          }
           reservedDomains = buildReservedDomains(
-            prefetchedProjects.map((p) => p?.domain),
+            (prefetchedProjects ?? []).map((p) => p?.domain),
             brandOwnUrls,
           );
         } else {
@@ -1979,17 +1991,16 @@ function BrandsController(ctx, log, env) {
         return notFound(`Brand not found: ${brandId}`);
       }
 
-      // Brand-level Semrush re-sync: when an edit changes URL sources or
-      // competitors and the brand is in sub-workspace mode, propagate the change
-      // onto every market/project (region-filtered per market). Skipped for
-      // flat-mode brands and unrelated edits. Hard-fail so the brand never drifts
-      // out of sync silently. One transport for both syncs.
+      // Brand-level Semrush re-sync: when an edit changes URL sources, competitors
+      // or aliases and the brand is in sub-workspace mode, propagate the change onto
+      // every market/project (region-filtered per market). Skipped for flat-mode
+      // brands and unrelated edits. One transport shared across all three syncs.
       //
-      // NOTE (intentional asymmetry vs create): the SAME URL/competitor
-      // propagation is BEST-EFFORT on the create path (handleCreateMarket-
-      // Subworkspace swallows a benchmark hiccup so it cannot strand a
-      // half-provisioned brand) but HARD-FAIL here on edit — an already-live
-      // brand must not silently diverge from Semrush after a row commit.
+      // Best-effort, matching the create path: the brand row is already committed
+      // when this block runs, so a re-sync failure must not surface as a 5xx on an
+      // edit that did persist. The response carries semrushSyncPending:true so the
+      // divergence is visible to the caller, and the error log below carries the
+      // context needed to enumerate and recover drifted brands.
       const urlsTouched = updates.urls !== undefined
         || updates.socialAccounts !== undefined
         || updates.earnedContent !== undefined;
@@ -2063,8 +2074,12 @@ function BrandsController(ctx, log, env) {
         } catch (syncError) {
           // LLMO-6545: Accept drift — DB is already committed so hard-failing returns
           // a 500 while the edit actually succeeded. Log all context so the divergence
-          // is diagnosable via Splunk and recoverable via --reconcile migration later.
-          log.error('serenity: brand-edit Semrush re-sync failed after row commit', {
+          // is diagnosable and recoverable via --reconcile migration later.
+          // 401/403 = permanent (user not provisioned in Semrush — needs manual action).
+          // Other status / no status = transient (network, timeout, upstream hiccup).
+          const isPermanent = syncError?.status === 401 || syncError?.status === 403;
+          const logFn = isPermanent ? log.warn : log.error;
+          logFn('serenity: brand-edit Semrush re-sync failed after row commit', {
             brandId,
             semrushSubWorkspaceId: updated.semrushSubWorkspaceId,
             urlsTouched,
@@ -2073,6 +2088,7 @@ function BrandsController(ctx, log, env) {
             status: syncError?.status,
             message: syncError?.message,
             stack: syncError?.stack,
+            permanent: isPermanent,
           });
           // Preserve any partial rejectedAliases collected before the throw so
           // the UI can still warn about aliases that were refused by Semrush.
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -2009,13 +2009,16 @@ function BrandsController(ctx, log, env) {
       const rejectedAliases = [];
       if ((urlsTouched || competitorsTouched || aliasesTouched)
         && hasText(updated.semrushSubWorkspaceId)) {
-        // Forward only an IMS user token upstream (matches the create path +
-        // the rest of /serenity/*): PATCH /brands is organization:write and thus
-        // S2S-reachable, so prefer an x-promise-token exchange and otherwise
-        // refuse a non-IMS bearer rather than proxy it.
-        const imsToken = await resolveSemrushImsToken(context, log, 'brands');
-        const transport = createSerenityTransport({ env: context.env, imsToken });
         try {
+          // Forward only an IMS user token upstream (matches the create path +
+          // the rest of /serenity/*): PATCH /brands is organization:write and thus
+          // S2S-reachable, so prefer an x-promise-token exchange and otherwise
+          // refuse a non-IMS bearer rather than proxy it. Inside the try because
+          // the row is already committed: a token-resolution 401 here is still a
+          // re-sync failure, and must soft-fail like any other rather than report
+          // an edit that did persist as an auth error.
+          const imsToken = await resolveSemrushImsToken(context, log, 'brands');
+          const transport = createSerenityTransport({ env: context.env, imsToken });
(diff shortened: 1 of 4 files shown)


src/controllers/brands.js
```diff
@@ -2073,7 +2072,7 @@ function BrandsController(ctx, log, env) {
             aliasesTouched,
             status: syncError?.status,
           });
-          throw syncError;
+          return ok({ ...updated, semrushSyncPending: true });
```
MysticatBot: issue (blocking): Early return drops `rejectedAliases` accumulated before the throw.

If `syncBrandUrlsAcrossMarkets` succeeds and pushes items into `rejectedAliases`, then a subsequent sync call (`syncCompetitorBenchmarksAcrossMarkets` or `syncBrandAliasesAcrossMarkets`) throws, this return discards the partial-success info. The UI uses `semrushRejectedAliases` to warn operators - that signal is now silently lost on the degraded path.

**Fix:**
```js
return ok({
  ...updated,
  semrushSyncPending: true,
  ...(rejectedAliases.length > 0 && { semrushRejectedAliases: rejectedAliases }),
});
```

src/controllers/brands.js
```diff
@@ -2073,7 +2072,7 @@ function BrandsController(ctx, log, env) {
             aliasesTouched,
             status: syncError?.status,
```
MysticatBot, line 2093: issue (blocking): The log captures `status` but not the error message or stack trace.

Now that this path is a production-visible soft-fail (not a hard 500 that propagates to the outer catch), the structured log is the *only* observability signal. For non-HTTP errors (network timeouts, JSON parse failures, unexpected TypeErrors), `status` will be `undefined` and the Splunk entry will be insufficient to triage.

**Fix:** Add `message` and optionally `stack` to the log payload:
```js
message: syncError?.message,
stack: syncError?.stack,
```

MysticatBot (changes requested):
Hey @aenascut,

**Verdict: Request changes** - one correctness gap and one observability gap on the new soft-fail path.
**Complexity: MEDIUM** - small diff; API surface risk.
_Changes: Converts brand PATCH Semrush re-sync from hard-fail (500) to soft-fail (200 with `semrushSyncPending` flag) when sync throws after DB commit (2 files)._

## Must fix before merge

1. [Important] Early return drops `rejectedAliases` collected before throw - `src/controllers/brands.js:2075` (details inline)
2. [Important] Missing `syncError.message`/stack in log payload degrades Splunk diagnosability - `src/controllers/brands.js:2073` (details inline)

<details>

- nit: `semrushSyncPending` is not part of the documented API contract - consumers that validate response shape may silently ignore it. Consider documenting it in the schema or linking a UI follow-up ticket. - `src/controllers/brands.js:2075`
- suggestion: Add a test case where URL sync succeeds (populating `rejectedAliases`) and then competitor/alias sync throws - verifies both flags appear in the response. - `test/controllers/brands.test.js:6722`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 50s | <b>Cost:</b> $3.35 | <b>Commit:</b> <code>268ab4f38e705a289130dac52a20ad4c248536b4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @aenascut,

**Verdict: Approve** - both prior blocking findings are resolved cleanly, CI is green.
**Complexity: MEDIUM** - small production fix; API surface signal (existing endpoint behavior change).
_Changes: Converts brand PATCH Semrush re-sync from hard-fail (500) to soft-fail (200 with `semrushSyncPending` flag), preserving partial `rejectedAliases` and logging error message/stack for Splunk diagnosability (2 files)._

<details>

- nit: `semrushSyncPending` is still not part of the documented API contract (OpenAPI schema). Consumers validating response shape may ignore it silently. Consider a follow-up to document it in `docs/openapi/schemas.yaml` or link the UI ticket that consumes this flag. - `src/controllers/brands.js:2080`

</details>

- rejectedAliases now preserved in the soft-fail return via conditional spread
- syncError.message and syncError.stack now included in the structured log payload

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 8m 9s | <b>Cost:</b> $1.78 | <b>Commit:</b> <code>914c465e6bb605ed1c590043a0df8c756a5b3c72</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich (changes requested):
## Summary

The motivating problem is real and I verified it independently: migrated orgs have LLMO users who edit brand content in LLMO but have no Semrush user record, and today those edits fail. Soft-failing the post-commit re-sync is the right direction. The DB-commit-then-throw asymmetry genuinely was a bug.

Three things block merge for me, and the most important one is not in the diff — it is that the driving failure class is **permanent**, not transient, which invalidates the observability and recovery plan the PR relies on.

For context on why this differs from the bot approval: these findings come from the surrounding code and from the state of the prod log pipeline, neither of which a diff-scoped review would surface.

---

## Blocking

### B1 — Competitor edits still hard-fail, so the change does not cover the case it is for

`src/controllers/brands.js` runs a self-referential-competitor guard **before** the DB write, and it calls Semrush:

```js
// brands.js:1948-1950
const imsToken = await resolveSemrushImsToken(context, log, 'brands');
const transport = createSerenityTransport({ env: context.env, imsToken });
prefetchedProjects = await resolveProjects(transport, brandState.semrushSubWorkspaceId);
```

`resolveProjects` hits `transport.listProjects` (`src/support/serenity/resolve-projects.js:35`). That block sits outside the new try/catch, so a 403 there lands in the outer catch at `brands.js:2100` and returns 403 to the client — with no row written at all.

Net effect: a PATCH touching `urls`, `socialAccounts`, `earnedContent` or `brandAliases` now succeeds, but a PATCH touching `competitors` fails exactly as before, for exactly the users this change is meant to unblock. `resolveSemrushImsToken` at `brands.js:2005` is also outside the `try {` on line 2007.

The fix needs a design call rather than just moving a brace: degrading the guard to "no reserved project domains" on a listing failure weakens the self-reference protection (a competitor equal to one of the brand's own market domains could be stored). Worth deciding explicitly — my preference is to degrade and log, since the guard is a data-quality nicety while the write being rejected is a hard block.

### B2 — The failure class is permanent, and the Phase 2/3 recovery path cannot run

The PR body concludes "Semrush timeouts are transient; alerting would be noisy". For the migrated-org case that is the wrong model:

- Semrush binds the forwarded IMS token to a Semrush **user record** by `authId` or email before executing the call — `mysticat-architecture/platform/proposal-semrush-ims-integration.md`
- JIT user provisioning from IMS claims was explicitly declined — `serenity-docs/docs/prompt-management/prompts-and-bp-proposal.md`, ask 4: *"No JIT for now. Users added manually to Semrush on onboarding."*
- The observed status for an unprovisioned Adobe identity is **403**, confirmed by live debugging on 2026-05-29 against the `/serenity/*` proxy

So this is not a hiccup that the next edit heals. Every edit by an unprovisioned user drifts, on every request, indefinitely.

That matters because the recovery story does not currently work. Phase 2 is an on-demand Splunk query on `"serenity: brand-edit Semrush re-sync failed after row commit"`, and Phase 3 `--reconcile` depends on enumerating drifted brands from it. **api-service prod does not meaningfully ship to Splunk.** Measured 2026-07-27 on `index=dx_aem_engineering sourcetype=dx_aem_sites_spacecat_backend_prod`, last 24h:

| service | events / 24h |
|---|---|
| content-scraper | 6,920 |
| audit-worker | 3,599 |
| import-worker | 65 |
| auth-service | 11 |
| **api-service** | **7** |
| mysticat-projector | 3 |

Those 7 events are unrelated `info` lines from two isolated invocations. Mis-routing ruled out: `sourcetype=dx_aem_sites_spacecat_*` over 24h for `"serenity:"` OR `"brand-edit Semrush"` returns zero. Search authorization is index-scoped, so that zero is authoritative about indexing.

This log-delivery gap is shared infrastructure, not something this PR introduced. But as written the change ships permanent silent drift whose affected-brand set cannot be enumerated by the mechanism nominated to enumerate it. Combined with no standing alert and no consumer for `semrushSyncPending` (`gh search code` across `adobe` and `adobe-rnd` finds none), the customer sees an unqualified success and nobody downstream learns anything.

Concretely, I would want at least the first of these before merge:

1. **Enqueue the re-sync for retry** alongside the 200, per the repo's queue-based async pattern, so transient failures self-heal and permanent ones surface as repeated failures rather than as nothing. This also makes `--reconcile` a backstop instead of the only recovery.
2. **Split the log/alert by class** — 401/403 means "user not provisioned in Semrush, needs action", 5xx/timeout means "retry". A standing alert on the permanent class is not noisy, because it is not transient.
3. **Base recovery on CloudWatch**, not Splunk, until api-service delivery is fixed.

Related prior art worth noting: the `/serenity/*` prompts proxy already converts a Semrush 403 into `200 {items:[], errors:[...]}` that the elmo UI ignores, so unauthorized users see a silent empty list rather than access-denied. A 200 carrying a flag no client reads reproduces that shape on a second surface.

### B3 — The comments now state the opposite of what the code does

`brands.js:1982-1992` still reads:

> Hard-fail so the brand never drifts out of sync silently.
> NOTE (intentional asymmetry vs create): ... HARD-FAIL here on edit — an already-live brand must not silently diverge from Semrush after a row commit.

That is now false, and it is the block a future reader hits first. It needs rewriting to describe the behaviour that exists, e.g.:

```js
// Brand-level Semrush re-sync: when an edit changes URL sources, competitors or
// aliases and the brand is in sub-workspace mode, propagate the change onto every
// market/project (region-filtered per market). Skipped for flat-mode brands and
// unrelated edits. One transport for both syncs.
//
// Best-effort, matching the create path: the brand row is already committed when
// this runs, so a re-sync failure must not surface as a 5xx on an edit that did
// persist. The response carries semrushSyncPending so the divergence is visible,
// and the error log below carries the context needed to recover it.
```

Same for the tests — `brands.test.js` currently narrates the change ("Before LLMO-6545 this re-threw the error (500). After the fix it soft-fails"). Once merged, that reads as history a fresh reader cannot verify. State the invariant instead.

---

## Should fix

### S1 — The gateway-URL redaction assertions are gone, and that path is still reachable

On `main`, `test/controllers/brands.test.js` asserts `Upstream request failed` / `Upstream authorization failed` four times. On this branch it asserts them zero times: the four tests were repurposed to assert `200 + semrushSyncPending`, where the no-leak checks are now tautological — the body is `{...updated, semrushSyncPending}` and structurally cannot contain the error.

`createErrorResponse`'s Semrush branch (`brands.js:239-243`) is still live on this route, reachable via the pre-write guard in B1. The primitives stay covered in `test/support/serenity/errors.test.js`, so this is a coverage regression rather than a logic hole — but re-point one of those tests at the pre-commit path rather than deleting the assertion. `codecov/patch` is green because it only measures changed lines.

### S2 — `semrushSyncPending` is missing from the OpenAPI spec

`docs/openapi/schemas.yaml:7797-7813` documents the sibling flag `semrushRejectedAliases` on `V2BrandUpdateResponse`, and `docs/openapi/brands-v2-api.yaml:172-176` describes it in the 200. The new field is added to the same response body and documented in neither. `CLAUDE.md` makes "OpenAPI First / Specification Sync" a must-follow, so this is a convention gap rather than a nit — and the schema to extend is already sitting next to it.

Worth documenting the semantic too, since it is load-bearing: `semrushSyncPending: true` means the row persisted and Semrush does not reflect it yet.

---

## What I agree with

- Returning the committed row rather than a 5xx is correct — the previous behaviour reported failure for a write that succeeded.
- Preserving partial `rejectedAliases` through the catch is the right call.
- Adding `message` and `stack` to the error log is right, and will matter more once the log is actually reachable.

rainer-friederich (approved):
LGTM

PR #806, opened 2025-03-05
feat: introduce Bulk scrape and audit

dzehnder:
introduce two new slack commands to bulk run scrapes and audits:

`@spacecat run scrape <CSV-FIle>` and `@spacecat run audits all <CSV-FIle>`

PR #2909, opened 2026-07-27
feat(brand-claims): enable/disable Slack command + brandClaimsEnabled flag (LLMO-5741)

akshaymagapu:
## What

Two Slack keywords to turn Brand Claims on/off for a brand, keyed on brand ID:

```
enable-brand-claims {brandId}
disable-brand-claims {brandId}
```

They flip `brands.brand_claims_enabled`. Also surfaces that flag as `brandClaimsEnabled` on the brand read API (default `false`, ops-set).

## Why

`brand_claims_enabled` is the gate the mystique Brand Claims consumer reads back before turning a brand-presence sheet-ready event into a claims run. Until now there was no way to set it except direct SQL — this gives operators a one-line Slack command to opt a brand in or out.

Epic LLMO-5741. Requires mysticat-data-service #806 (the `brand_claims_enabled` column) applied to the DB.


src/support/slack/commands/toggle-brand-claims.js
```diff
@@ -0,0 +1,87 @@
...
+    name: 'Brand Claims',
+    description: 'Enables or disables Brand Claims scheduling for a brand (by brand ID).',
+    phrases: [ENABLE_PHRASE, DISABLE_PHRASE],
+    usageText: `${ENABLE_PHRASE} {brandId} | ${DISABLE_PHRASE} {brandId}`,
+  });
+
+  const { dataAccess, log } = context;
+
+  const execute = async (message, slackContext) => {
+    const { say, user } = slackContext;
```
MysticatBot, line 42: issue (blocking): The extracted `brandId` is passed to PostgREST without UUID format validation. While PostgREST parameterizes the query (no injection risk), an arbitrary string quietly round-trips to the database and returns a misleading "No brand found" message. The repo's CLAUDE.md documents UUID validation as a standard practice (`isValidUUIDV4()`).

Add a format check before calling `setBrandClaimsEnabled`:

```js
import { isValidUUIDV4 } from '../../../utils/common.js';

if (!isValidUUIDV4(brandId)) {
  await say(`:warning: '${brandId}' is not a valid brand ID (expected UUID). ${baseCommand.usage()}`);
  return;
}
```

This gives operators immediate feedback and avoids unnecessary DB calls for obviously invalid input.

test/support/slack/commands/toggle-brand-claims.test.js
```diff
@@ -0,0 +1,130 @@
...
+  describe('Execute', () => {
+    it('enables brand claims when the enable keyword is used', async () => {
+      setBrandClaimsEnabledStub.resolves(BRAND);
+
+      const command = BrandClaimsCommand(context);
+      await command.execute(`enable-brand-claims ${BRAND_ID}`, slackContext);
+
+      const args = setBrandClaimsEnabledStub.firstCall.args[0];
+      expect(args.brandId).to.equal(BRAND_ID);
+      expect(args.enabled).to.equal(true);
```
MysticatBot, line 68: issue (blocking): The enable and disable tests verify the arguments passed to `setBrandClaimsEnabled` but never assert that `slackContext.say` was called with the expected confirmation message. The command could silently drop the success `say()` call and these tests would still pass.

Add an assertion on the response content:

```js
expect(slackContext.say.calledWithMatch(/Brand claims \*enabled\*/)).to.be.true;
```

(and `/\*disabled\*/` in the disable test.)

MysticatBot (changes requested):
Hey @akshaymagapu,

**Verdict: Request changes** - a few items to tighten up before this ships.
**Complexity: MEDIUM** - medium diff; API surface addition.
_Changes: Adds enable/disable-brand-claims Slack commands and surfaces `brandClaimsEnabled` as a read-only field on the brand API (7 files)._

## Must fix before merge

1. [Important] Missing UUID format validation on `brandId` - `src/support/slack/commands/toggle-brand-claims.js:41` (details inline)
2. [Important] Enable/disable tests do not assert on the success Slack message - `test/support/slack/commands/toggle-brand-claims.test.js:62` (details inline)
3. [Important] Stale command count in CLAUDE.md - The Slack Integration section says "36 commands" in `src/support/slack/commands/`; this PR adds the 37th. Update the count.

<details>

- nit: `brand.name` from DB is interpolated into Slack mrkdwn without escaping - if a brand name contains `*`, `~`, or `<url|label>` syntax it will render unexpectedly. Consider using Block Kit with `plain_text` type for the success response. - `src/support/slack/commands/toggle-brand-claims.js:75`
- suggestion: Log successful state transitions at info level (brand ID, new value, actor) for operational audit trail beyond the `updated_by` column. - `src/support/slack/commands/toggle-brand-claims.js:73`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 4m 10s | <b>Cost:</b> $4.64 | <b>Commit:</b> <code>fa951b8c36fbdddc15b8403b5ab492d12a397dc7</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

akshaymagapu:
Thanks for the review — addressed in 45fab947:

**Must-fix**
1. ✅ **UUID validation** — added `isValidUUID(brandId)` (from `@adobe/spacecat-shared-utils`, matching the delegate commands; the repo's actual validator, not `isValidUUIDV4`) before the DB call, with a clear warning on bad input.
2. **say() assertion** — this was already present: the enable/disable tests assert `slackContext.say.calledWithMatch(/Brand claims \*enabled\* .../)` (and `\*disabled\*`) at test lines 70/81. No change needed.
3. ✅ **CLAUDE.md count** — bumped 36 → 37 and listed the new keywords.

**Non-blocking**
- Added `log.info` of each enable/disable transition (brand id, new state, actor) for the audit trail.
- Brand-name escaping in mrkdwn: leaving as-is for now — this is an internal ops command over trusted brand-table names; Block Kit conversion isn't worth the surface here. Noted as a follow-up if we ever expose it more broadly.

The failing CI checks (branch-deploy / deploy-stage / semantic-release) look environment/permission related rather than code — will confirm.

src/support/brands-storage.js
```diff
@@ -833,6 +838,45 @@ export async function getBrandBySite(organizationId, siteId, postgrestClient, lo
...
+export async function setBrandClaimsEnabled({
+  brandId,
+  enabled,
+  postgrestClient,
+  updatedBy = 'system',
+}) {
+  if (!postgrestClient?.from) {
+    throw new Error('PostgREST client is required');
+  }
+  if (!hasText(brandId)) {
```
MysticatBot, line 866: issue (blocking): The UPDATE query filters only on `id` (`.eq("id", brandId)`) without excluding soft-deleted brands. The sibling function `setBrandStatus` uses `.neq("status", "deleted")` to prevent mutating deleted brands - this new write path should follow the same pattern.

Without this guard, an operator could unknowingly flip the flag on a soft-deleted brand. If the downstream mystique Brand Claims consumer reads this flag without also checking brand status, it could schedule claims runs for a non-existent brand.

Fix: add `.neq("status", "deleted")` to the query chain:

```js
  .from('brands')
  .update({ brand_claims_enabled: enabled, updated_by: updatedBy })
  .eq('id', brandId)
  .neq('status', 'deleted')
  .select('id, name')
  .maybeSingle();
```

Add a test case confirming that a deleted brand returns null.

MysticatBot (changes requested):
Hey @akshaymagapu,

**Verdict: Request changes** - one gap in the write path to address before merge.
**Complexity: HIGH** - medium diff; API surface risk flag (OpenAPI schema change).
_Changes: Adds enable/disable-brand-claims Slack commands and surfaces `brandClaimsEnabled` as a read-only field on the brand API (8 files)._
Note: Recommend a human read before merge - this change modifies an OpenAPI contract (shared-contract trigger) and the root CLAUDE.md (architectural-path trigger). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge

1. [Important] `setBrandClaimsEnabled` does not exclude soft-deleted brands - `src/support/brands-storage.js:863` (details inline)

<details>

- nit: `setBrandClaimsEnabled` does not validate that `enabled` is a boolean before passing it to PostgREST. Current caller always passes `true`/`false`, but as a public export a one-line `if (typeof enabled !== 'boolean') throw` guard keeps the contract explicit. - `src/support/brands-storage.js:852`
- nit: postgrestClient availability check is duplicated between the command (friendly Slack message) and the storage function (throws). Harmless layered defense, but the storage function's throw can never fire from this call site. - `src/support/slack/commands/toggle-brand-claims.js:53`

</details>

- UUID validation on brandId now uses `isValidUUID` from `@adobe/spacecat-shared-utils`.
- CLAUDE.md command count bumped from 36 to 37.
- Audit `log.info` for enable/disable transitions added.

akshaymagapu:
Addressed in ea33d236:

- **Must-fix — soft-deleted brands:** `setBrandClaimsEnabled` now adds `.neq('status', 'deleted')`, so a deleted brand returns no row → the command reports "No brand found" instead of flipping the flag. Matches the guard used elsewhere in `brands-storage.js`.
- **nit — boolean guard:** added `if (typeof enabled !== 'boolean') throw` on the public export.
- **nit — duplicated postgrestClient check:** left as intentional layered defense (command gives a friendly Slack message; storage fn throws as a contract guard for other callers).

198 tests pass, 100% coverage on the changed files.

MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - all prior blocking findings resolved; code is clean and well-tested.
**Complexity: HIGH** - medium diff; API surface + shared-contract (OpenAPI schema) + architectural-path (CLAUDE.md) signals.
_Changes: Adds enable/disable-brand-claims Slack keyword commands and surfaces `brandClaimsEnabled` as a read-only field on the brand API (8 files)._
Note: Recommend a human read before merge - this change modifies an OpenAPI contract (shared-contract trigger) and the root CLAUDE.md (architectural-path trigger). The bot review is a complement to, not a replacement for, a human read here.
Note: CI checks are currently pending (build, it-postgres) - confirm they pass before merge.

<details>

- nit: `setBrandClaimsEnabled` does not explicitly set `updated_at` in the update payload - other mutations in `brands-storage.js` (e.g. `setBrandStatus`) pass `updated_at: new Date().toISOString()`. If the table relies on application-set timestamps rather than a DB trigger, this could leave a stale `updated_at`. Verify the table has an `updated_at` trigger; if not, add it for audit consistency. - `src/support/brands-storage.js:864`
- nit: Command test sets `context.log = console` - the error-path test will print a noisy stack trace to test runner output. A sinon spy (e.g. `{ info: sinon.spy(), error: sinon.spy() }`) would keep output clean and allow asserting on the log call. - `test/support/slack/commands/toggle-brand-claims.test.js:549`

</details>

- UUID format validation on brandId now uses `isValidUUID` from `@adobe/spacecat-shared-utils`.
- CLAUDE.md command count bumped from 36 to 37, new keywords listed.
- Test assertions on the Slack response content confirmed present.
- `setBrandClaimsEnabled` now excludes soft-deleted brands via `.neq('status', 'deleted')`.
- Boolean type guard added on the `enabled` parameter.
- Audit `log.info` added for enable/disable transitions with actor tracking.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 20s | <b>Cost:</b> $4.01 | <b>Commit:</b> <code>3946ecc60b8dfa98a3aed772ba16746a6fc46aa1</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2912, opened 2026-07-27
fix: send category without prefix appended in the elements

vivesing:
## Summary

Fixes a double-prefixing bug in the `category`/`categoryId` filter across `/serenity/brand-presence/*` endpoints. Callers already send the fully-namespaced Semrush tag value (e.g. `category=category__Living+Room+Furniture+Retail`), but the payload builders were unconditionally re-adding the `category__` prefix before sending it to the Semrush Elements API — producing a mismatched, never-matching tag (`category__category__...`) that silently returned empty/incorrect results instead of erroring.

## Root Cause

Six code paths built the Semrush `CBF_tags` filter as `` `category__${category}` `` regardless of whether the incoming value already carried the prefix. Since the query param is passed straight through from the caller with no stripping anywhere upstream, any already-prefixed input got double-prefixed.

## Fix

Stop re-adding the `category__` prefix — pass the `category` value through to Semrush as-is.

## Endpoints Fixed

| Endpoint | Builder / call site |
|---|---|
| `GET /serenity/brand-presence/url-inspector/stats` | `buildOwnedUrlsStatsPayload` |
| `GET /serenity/brand-presence/url-inspector/owned-urls` | `buildOwnedUrlsStatsPayload` + `buildOwnedUrlsTrendPayload` |
| `GET /serenity/brand-presence/url-inspector/domain-urls` | `buildDomainUrlsPayload` |
| `GET /serenity/brand-presence/url-inspector/cited-domainsoad` |
| `GET /serenity/brand-presence/url-inspector/prompts/count` | inline `tags` array in `getUrlInspectorPromptsCount` |
| `GET /serenity/brand-presence/sentiment-overview` | `buildSentimentOverviewPayload` |               

                                   
**Not affected:**
- `GET /serenity/brand-presence/stats` — accepts `categoryId`/`category` for parity with the Postgres-RPC reference endpoint, but it's a documented no-op (never forwarded to Semrush).
- `GET /serenity/brand-presence/kpi-headlines`, `source-visibility-headline` — no `category` param at all.
- All other serenity endpoints (`weeks`, `prompts`, `topics`, `topics/:topicId/prompts`, `market-tracking-trends`, `competitor-summary`,
`filter-dimensions`) don't accept a `category` filter.

## Files Changed

- `src/support/elements/definitions/owned-urls.js`
- `src/support/elements/definitions/domain-urls.js`
- `src/support/elements/definitions/cited-domains.js`
- `src/support/elements/definitions/sentiment-overview.js`
- `src/controllers/elements.js`
- `docs/openapi/serenity-api.yaml` — updated `category`/`cans to reflect the new pass-through contract
- `test/support/elements/definitions/sentiment-overview.test.js`
- `test/controllers/elements.test.js`

## Behavior Change (Breaking for Un-prefixed Callers)

Callers must now send the **full** `category__<label>` value. A bare label with no prefix will no longer be tagged/matched at all (previously it was auto-prefixed; now it's sent verbatim). This matches how the UI/consuming clients already construct the value (confirmed via a real production
request).

## Test Plan

- [x] `npx mocha 'test/support/elements/**/*.test.js'` — 326 passing
- [x] `npx mocha test/support/elements/definitions/sentimenntrollers/elements.test.js` — 155 passing
- [x] `npx eslint` clean on all changed files

### DIFF
--- a/docs/openapi/serenity-api.yaml
+++ b/docs/openapi/serenity-api.yaml
@@ -1242,8 +1242,8 @@ v2-serenity-brand-presence-sentiment-overview:
         in: query
         required: false
         description: |
-          Category label sent to Semrush as a `category__<label>` tag. `category`
-          is accepted as an alias.
+          Full `category__<label>` tag value (caller includes the `category__`
+          prefix), sent to Semrush as-is. `category` is accepted as an alias.
         schema: { type: string }
     responses:
       '200':
@@ -1726,7 +1726,7 @@ v2-serenity-url-inspector-stats:
       - name: categoryId
         in: query
         required: false
-        description: Category label, mapped to a Semrush tag (`category__<label>`).
+        description: Full `category__<label>` tag value (caller includes the prefix), sent as-is.
         schema: { type: string }
       - name: category
         in: query
@@ -1839,7 +1839,7 @@ v2-serenity-url-inspector-prompts-count:
       - name: categoryId
         in: query
         required: false
-        description: Category label, mapped to a Semrush tag (`category__<label>`).
+        description: Full `category__<label>` tag value (caller includes the prefix), sent as-is.
         schema: { type: string }
       - name: category
         in: query
--- a/src/controllers/elements.js
+++ b/src/controllers/elements.js
@@ -1447,11 +1447,13 @@ export default function ElementsController(context, log, env) {
         workspaceId, query, service, projectIds,
       } = scope;
 
+      // category already carries the `category__<label>` prefix from the caller;
+      // sent through as-is, not re-prefixed.
       const category = query.categoryId || query.category;
       const { count: totalPrompts } = await service.getPrompts(workspaceId, {
         model: query.model,
         platform: query.platform,
-        tags: category ? [`category__${category}`] : [],
+        tags: category ? [category] : [],
         projectIds,
       });
 
--- a/src/support/elements/definitions/cited-domains.js
+++ b/src/support/elements/definitions/cited-domains.js
@@ -47,7 +47,8 @@ function defaultDateRange() {
  * @param {string} [params.platform] - Legacy alias for `model`; `model` takes precedence.
  * @param {string} [params.startDate] - ISO date (YYYY-MM-DD). Defaults to 28 days ago.
  * @param {string} [params.endDate] - ISO date (YYYY-MM-DD). Defaults to today.
- * @param {string} [params.category] - Category label, pushed as the tag `category__<label>`.
+ * @param {string} [params.category] - Full `category__<label>` tag value, sent
+ *   as-is (callers already include the `category__` prefix).
  * @param {string} [params.projectId] - Semrush project id for region scoping (top-level).
  */
 export function buildCitedDomainsPayload({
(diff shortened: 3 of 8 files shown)


MysticatBot (approved):
Hey @vivesing,

**Verdict: Approve** - clean bug fix, well-scoped and well-documented.
**Complexity: MEDIUM** - small diff; API surface change.
_Changes: removes redundant `category__` prefix addition across six Semrush filter builders, fixing a double-prefixing bug that produced non-matching tags (8 files)._
Note: Recommend a human read before merge - this change modifies an OpenAPI contract file (`docs/openapi/serenity-api.yaml`). The bot review is a complement to, not a replacement for, a human read here.
Note: CI checks are partially failing (branch-deploy, deploy-stage, semantic-release) - verify these are unrelated before merge.

<details>

- nit: Stale endpoint-level description in OpenAPI spec - the summary for `listSerenityBrandPresenceSentimentOverview` (~line 1187) still reads "`category` becomes a `category__<label>` tag", contradicting the new pass-through contract. The per-parameter descriptions were updated but this top-level prose was missed. - `docs/openapi/serenity-api.yaml:1187`
- nit: The `getUrlInspectorStats` controller test (~line 1185) still passes `categoryId=Firefly` (bare, unprefixed). While the test is technically valid (it stubs the service and only checks passthrough), the input no longer matches the documented API contract where callers provide the full `category__<label>` value. - `test/controllers/elements.test.js:1185`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 54s | <b>Cost:</b> $7.64 | <b>Commit:</b> <code>446b9bd7191b0841b3d1fb224db7e4d150d1e0b9</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

calvarezg:
## Multi-Persona Review — PR #2912

Correct root-cause fix — the `category__category__…` double-prefix is real and eliminated at all 6 sites, consistently, with JSDoc/comments/OpenAPI updated in lockstep. Verified against head `446b9bd`. Two asks before merge:

**[Major] Test coverage gap — 4 of 6 changed paths have no assertion on the new tag value.** Only `buildSentimentOverviewPayload` (`sentiment-overview.test.js`) and the prompts-count controller (`elements.test.js:1445`) assert the resulting `CBF_tags`/`tags` value. `buildCitedDomainsPayload`, `buildDomainUrlsPayload`, and `buildOwnedUrls{Stats,Trend}Payload` have **no test file** and no assertion anywhere on their category tag — their `if (category)` branch is line-covered (why Codecov is green), but the exact "sent as-is" behavior this PR changes is unasserted. A future revert of the prefix in any of those 4 would not be caught. Add one assertion per builder mirroring `sentiment-overview.test.js`: input `category: 'category__travel'` → assert `{ op:'eq', val:'category__travel', col:'CBF_tags' }`.

**[Question] Silent-empty failure mode for un-prefixed callers.** The old bug double-prefixed a bare label → matched nothing. Now a caller still sending a bare label (`category=Firefly`) produces `CBF_tags eq 'Firefly'` which — because `op:'eq'` is exact-match — also matches nothing: silent empty 200, no error. Your PR body flags this as intentional, and it's well-founded (the `filter-dimensions` endpoint, the canonical source of category values, already emits prefixed ids like `{ id: 'category__Firefly' }`). So the intended client flow is safe. Confirming you're deliberately accepting the silent-empty behavior for hand-crafted/legacy callers rather than normalizing defensively:
```js
const tag = category.startsWith('category__') ? category : `category__${category}`;
```
That would fix the double-prefix bug, tolerate both forms, and handle over-prefixing — no downside. Or return **400** on a missing prefix so misuse fails loudly instead of returning empty data. Author's call.

**[Minor]** `getUrlInspectorStats` controller test (`elements.test.js:1185`) still uses `?categoryId=Firefly` while its prompts-count sibling was updated to `category__Firefly`. Update for consistency with the new contract.

**[Minor]** Branch name `feat/ctaegory-filters` — typo, and `feat/` prefix on a `fix:` commit. Cosmetic.

**Security: no finding.** The `category__` prefix is a namespacing convention, not a tenant boundary — isolation is enforced by `workspaceId`/`projectIds` resolved server-side, and the tag filter is AND-combined inside that already-scoped query. `op:'eq'` on a structured JSON field means no injection vector.

_Recommendation: Approve with two asks — add the 4 builder assertions, and confirm the intended silent-empty contract._

<sub>🤖 Multi-persona review (Architect · DBA · Security · Tester · Junior Dev)</sub>

calvarezg (approved):

PR #2916, opened 2026-07-27
fix(serenity): skip sub-workspace settle poll on bare brand create (LLMO-6569)

igor-grubic:
## What & why

In Serenity/Semrush mode, sub-workspace-only provisioning frequently times out at the ~15s Fastly first-byte edge budget. The cause: after creating a sub-workspace we wait for it to settle to `status === 'created'` before continuing — a poll of up to **30 × 1s ≈ 30s** (`pollUntilCreated` in `workspace-lifecycle.js`).

For **sub-workspace-only** provisioning (no project, no prompts created in the request), that wait is pure dead weight — nothing downstream needs a settled workspace — and the poll sits between the upstream sub-workspace create and the DB pointer persist, so a timeout there strands the sub-workspace. The exact symptom differs by path:

- **Bare brand create** (`provisionBrandSubworkspaceBare`): the brand row doesn't exist yet, and `capturedWorkspaceId` was only set *after* the poll — so a pre-fix poll timeout **leaked the sub-workspace upstream** (holding its full `CREATE_ALLOCATION`) with no brand row referencing it, and the failure cleanup no-op'd on the null id.
- **Sub-workspace-only activate** (pending→active, bare reactivation): the brand row already exists and `ensureSubworkspace` owns the pointer write — so a poll timeout left the row present with its sub-workspace pointer **never written** ("created upstream, never linked"). Self-heals on retry, but the user still eats the 504.

## Change

- `ensureSubworkspace` gains a `createReadiness: 'poll' | 'skip'` option (default `'poll'` — every caller that creates a project/prompts in the same request is byte-for-byte unchanged).
- `'skip'` bypasses the settle poll on the fresh-create path and persists the pointer immediately; the workspace settles asynchronously (a later market-add settles it via the existing-sub-workspace branch before creating a project).
- Passed by the **three sub-workspace-only callers**: `provisionBrandSubworkspaceBare`, and the two activate branches in `serenity.js` (pending→active, bare reactivation).

## Scope

Only sub-workspace-only paths use `'skip'`. The **create-with-market** path (and the market/project-creating activation branch) keep `'poll'` on purpose — they create a project/prompts immediately, so a fail-fast there would be unsafe: with no persisted brand row yet, a retry can't re-adopt a still-settling workspace (`findAdoptableFamilyMatch` only adopts `created` children) and would leak orphaned sub-workspaces. Those paths are handled separately.

## Known limitation

Persisting the pointer before the workspace settles opens a narrow race: a deactivate issued seconds after a bare create → `decommissionBrandWorkspace` → `releaseFullAllocation` → `transferWorkspaceResources` has no `isWorkspaceNotReady` retry (unlike `resource-manager.js`'s `transferAndSettle`). The window is seconds and the failure is a clean retryable error, not corruption — noted for the record, not blocking.

## Testing

- `workspace-lifecycle.test.js`: new `createReadiness: 'skip'` case (no `getWorkspaceStatus` poll, pointer persisted immediately); the default `'poll'` create+poll+persist path is still pinned.
- `serenity.test.js`: the two sub-workspace-only activate tests now assert `createReadiness: 'skip'`.
- `brand-provisioning.test.js`: bare-provisioning options-bag assertions updated.
- `npm run type-check` ✅ · `npm run lint` ✅ · `npm test` ✅ (16,034 passing, 100% coverage gate).

## Links

- Jira: https://jira.corp.adobe.com/browse/LLMO-6569

### DIFF
--- a/src/controllers/serenity.js
+++ b/src/controllers/serenity.js
@@ -1230,6 +1230,12 @@ function SerenityController(context, log, env) {
           brandPointerReloader(ctx, auth.brandUuid),
           {
             dynamicAllocation: dynamicAllocationEnabled(ctx),
+            // createReadiness 'skip' (LLMO-6569): pending→active is sub-workspace-only — no project
+            // or prompts are created here — so skip the up-to-30s settle poll. This is the path
+            // where a poll timeout otherwise leaves the brand row present but its sub-workspace
+            // pointer unwritten (the "created upstream, never linked" orphan); persisting the
+            // pointer immediately closes that window. Self-heals on retry, but the user ate a 504.
+            createReadiness: 'skip',
           },
         );
         let pendingActivateSucceeded = true;
@@ -1298,6 +1304,10 @@ function SerenityController(context, log, env) {
           brandPointerReloader(ctx, auth.brandUuid),
           {
             dynamicAllocation: dynamicAllocationEnabled(ctx),
+            // createReadiness 'skip' (LLMO-6569): bare reactivation of an already-active brand is
+            // sub-workspace-only (no project/prompts), so skip the settle poll — same safety and
+            // orphan-window rationale as the pending→active branch above.
+            createReadiness: 'skip',
           },
         );
         let bareSucceeded = true;
--- a/src/support/serenity/brand-provisioning.js
+++ b/src/support/serenity/brand-provisioning.js
@@ -348,11 +348,14 @@ export async function provisionBrandSubworkspaceBare(context, {
     // marketCount = 1: the bare sub-workspace is carved for a single future project
     // (the first market the user adds). ensureSubworkspace returns the new id.
     // createReadiness: 'skip' (LLMO-6569) — this path creates NO project/prompts, so nothing
-    // downstream needs a settled workspace. Skipping the up-to-30s settle poll persists the brand's
-    // sub-workspace pointer immediately (closing the orphan window that left brands unlinked when
-    // the poll blew the ~15s Fastly edge budget) and lets the workspace settle asynchronously; the
-    // first market-add later settles it via the existing-sub-workspace branch before creating a
-    // project. This is Rainer's point #1 ("we don't create markets in that").
+    // downstream needs a settled workspace. Skipping the up-to-30s settle poll lets the
+    // sub-workspace pointer be captured/persisted immediately (instead of after the poll) and lets
+    // the workspace settle asynchronously; the first market-add later settles it before creating a
+    // project. Pre-fix, a poll timeout here fired BEFORE the id was captured, so the failure
+    // cleanup saw a null id and no-op'd — leaking the sub-workspace upstream (holding its full
+    // CREATE_ALLOCATION) with no brand row yet referencing it. (The "created upstream but never
+    // linked" orphan is the ACTIVATE variant, where the brand row already exists — see
+    // serenity.js.) This is Rainer's point #1 ("we don't create markets in that").
     const subWorkspaceId = await ensureSubworkspace(
       transport,
       brandStub,
--- a/src/support/serenity/brand-provisioning.js
+++ b/src/support/serenity/brand-provisioning.js
@@ -347,6 +347,12 @@ export async function provisionBrandSubworkspaceBare(context, {
   try {
     // marketCount = 1: the bare sub-workspace is carved for a single future project
     // (the first market the user adds). ensureSubworkspace returns the new id.
+    // createReadiness: 'skip' (LLMO-6569) — this path creates NO project/prompts, so nothing
+    // downstream needs a settled workspace. Skipping the up-to-30s settle poll persists the brand's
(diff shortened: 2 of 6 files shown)


rainer-friederich (changes requested):
The approach is right and the safety argument holds where it matters: the readiness gate exists for exactly one pinned reason — `rest-transport.js` on `getWorkspaceStatus`, *"poll until `created` after a subworkspace create (creating projects against `not ready` can 500)"* — and the bare path creates no project, so skipping it does not violate that constraint. The follow-on claim also checks out: `handleCreateMarketSubworkspace` routes the single-market POST through `ensureSubworkspace`, whose existing-pointer branch polls unconditionally before any `createProject`, so the first market-add does settle the workspace. The default stays `'poll'` and `workspace-lifecycle.test.js` still pins create+poll+persist, so a future default flip breaks a test.

Two things to change before this merges.

## 1. Extend `createReadiness: 'skip'` to the two sub-workspace-only activate branches

`src/controllers/serenity.js:1223` (pending → active) and `src/controllers/serenity.js:1291` (bare reactivation of an already-active brand) are both explicitly sub-workspace-only — no project, no prompts, nothing created against the workspace in that request. They meet the exact criterion this PR's `@param` doc sets for `'skip'`, yet both still pass an options bag without it and pay the same up-to-30s settle poll on the same fresh-create path.

They are also the paths where the orphan symptom this PR cites as motivation genuinely occurs. On the bare **create** path the brand row cannot exist yet — `upsertBrand` in `src/controllers/brands.js` runs after provisioning returns — so a poll timeout there leaves no brand row at all. On the **activate** paths the brand row already exists and `ensureSubworkspace` owns the pointer write, so a poll timeout leaves precisely the "workspace created upstream, brand row present, the two never linked" state. It self-heals on retry (a null pointer re-enters the create path and `findAdoptableFamilyMatch` adopts the same-title child), but the user still eats the 504 that the fix is meant to eliminate.

Both branches persist the pointer inside `ensureSubworkspace` and then create nothing, so `'skip'` is safe there by the same argument that makes it safe for bare create, and it closes the orphan window instead of leaving it open on the two paths that actually have one.

While updating the `@param` text: it currently lists "activate" among the callers that create a project or prompts in the same request, which is only true of the project-activation branch further down, not of these two.

## 2. Stale comment at `src/support/serenity/workspace-lifecycle.js:519`

Three lines below the new conditional, the persist is still annotated:

```js
// Persist AFTER the workspace reads back `created` — flips the brand to subworkspace mode.
```

That is now false on the `'skip'` path, where the whole point is that the pointer is persisted without the workspace having read back `created`. Reword to state what the code does in both modes.

---

While looking at this — two observations for the description rather than the code, since they change nothing in the diff:

- The "brand row exists and the two are never linked" mechanism does not apply to the bare create path (see above). What a pre-fix poll timeout actually produced there is worse and worth stating: `capturedWorkspaceId` is only set when `ensureSubworkspace` calls `setSemrushSubWorkspaceId`, which happens *after* the poll — so on a timeout both the local catch in `provisionBrandSubworkspaceBare` and the compensation in `createBrand` see a null id and no-op, and the sub-workspace leaks upstream holding its full `CREATE_ALLOCATION` with no brand row referencing it. This PR closes that leak as a side effect, because the capture now happens.
- Persisting the pointer before the workspace settles opens a narrow new race for a deactivate issued seconds after a bare create: `decommissionBrandWorkspace` → `releaseFullAllocation` → `transferWorkspaceResources` has no `isWorkspaceNotReady` retry, unlike `resource-manager.js`'s `transferAndSettle`. The window is seconds and the failure is a clean retryable error rather than corruption, so it does not need to block this change — worth a line in the description so it is on the record.

igor-grubic:
Thanks for the thorough review — all addressed in `795836368`.

**① Extended `createReadiness: 'skip'` to the two sub-workspace-only activate branches** (pending→active and bare reactivation). Both now skip the settle poll, closing the "created upstream, never linked" window on exactly the paths that actually have it. Added `'skip'` assertions to both activate tests, and fixed the `@param` doc — the bare activate branches use `'skip'`; only the market/project-creating activation branch needs `'poll'`.

**② Fixed the stale persist comment** at `workspace-lifecycle.js` — it now states the pointer is persisted after `created` on `'poll'`, and *without* waiting for settle on `'skip'`.

**Description observations (no diff):**
- Corrected the bare-create mechanism — you're right, pre-fix it **leaked** the sub-workspace upstream (`capturedWorkspaceId` set only after the poll → both the local catch and `createBrand` compensation saw a null id and no-op'd), rather than "row present but unlinked" (that's the activate variant, where the row already exists). Updated the PR description and the inline comment accordingly.
- Added the deactivate-race note to the PR description as a known, non-blocking limitation (`releaseFullAllocation` → `transferWorkspaceResources` has no `isWorkspaceNotReady` retry; seconds-wide, clean retryable error).

Also rebased onto the latest `main` merge that landed on the branch.

rainer-friederich (approved):
LGTM - needs a rebase and then this is good to go.

PR #2918, opened 2026-07-28
fix: title Serenity sub-workspaces with the bare brand name

### LINKED ISSUE #2915: Drop the [id] suffix from Serenity sub-workspace titles without breaking create-recovery adoption
## Problem

Sub-workspaces created through the UI are titled `<brand name> [<first 8 chars of brand id>]`, e.g. `Lovesac [a1b2c3d4]`. Sub-workspaces created by the migration CLI are titled with the bare brand name. Customers see both conventions side by side in the Semrush UI, and the id suffix is meaningless to them.

The suffix is built in `src/support/serenity/workspace-lifecycle.js`:

```js
const ID_SUFFIX_LEN = 8;

function subworkspaceTitle(brand) {
  // ...
  return hasText(name) ? `${name} [${suffix}]` : `brand-${suffix}`;
}
```

The desired end state is the bare brand name on both paths.

## Why this is not a one-line deletion

The title is not decoration — it is the only key two recovery paths have, both in `ensureSubworkspace`:

1. **Ambiguous-create recovery.** `transport.createSubworkspace` has no idempotency key. On a 504 the create is ambiguous, so `adoptFromFamily` lists the parent's family and matches the exact title to find the workspace that may or may not have been created.
2. **Idempotent create-or-adopt** (issue 2718). Before creating, `findAdoptableFamilyMatch` looks for an existing `created`, project-empty child with the same title and reuses it, so a retry after a partial provisioning failure does not spawn a duplicate stub.

Brand display names are not unique within an org. The block comment above `subworkspaceTitle` spells out the failure mode a bare title reopens: one brand adopts a different, same-named brand's sub-workspace after a create timeout, and a later deactivate then decommissions the wrong brand's live markets.

This is not hypothetical right now. Create timeouts are being hit in practice on the brand-create path (observed while testing https://github.com/adobe/project-elmo-ui/pull/2493 — the brand and sub-workspace exist while the request times out), which is exactly when the title match runs.

## The migration precedent

The migration CLI (`mysticat-data-service`, `scripts/serenity_migration/executor.py`) runs the same pattern — `_child_workspaces_by_title` plus an adopt-on-not-ready wait — against bare titles. It accepts the collision risk because it works from a planned batch with known, deduplicated brand names. The UI path has no such plan and creates brands one at a time from free-text user input, so it cannot inherit that assumption for free.

## What a fix needs

Drop the suffix from the title and give the two recovery paths a key that does not depend on it. Options worth weighing:

1. Ask Semrush for an idempotency key or a client-supplied correlation id on `createSubworkspace`. This removes the need for title matching altogether and is the only option that fully closes the ambiguity.
2. Match on a non-title attribute of the child workspace if one is stable and readable back from the family listing.
3. Keep a title-based match but narrow it with a pre-create reservation, for example persisting the intended title and creation timestamp on the brand row and refusing to adopt a family entry that predates it.
4. Accept the collision risk on the UI path the way the migration does, but only with an explicit guard: refuse to adopt when the family listing returns more than one same-title candidate, and alert instead of guessing.

Option 4 is the cheapest and is close to what `adoptFromFamily` already does for multiple matches. It should be verified rather than assumed, and paired with a decision on what happens when a genuine duplicate-name brand hits a timeout.

## Also worth doing here

`subworkspaceTitle` throws a 500 when the brand id is missing, purely to protect the suffix. If the suffix goes, that guard should go with it rather than being left as dead defensive code.

## Scope

- `spacecat-api-service` — `src/support/serenity/workspace-lifecycle.js` (title construction, `findAdoptableFamilyMatch`, `adoptFromFamily`) and their tests.
- Existing sub-workspaces keep their current titles. No retroactive rename is implied by this issue; if one is wanted it is a separate operational task, and it interacts with the fact that the title is the adoption key.

## Related

- https://github.com/adobe/project-elmo-ui/pull/2493 — brand and market split UI, where the inconsistency was noticed
- Issue 2718 — the duplicate-stub problem that introduced the create-or-adopt reuse

rainer-friederich:
## 1. Abstract
Serenity sub-workspaces are now titled with the brand's bare display name on every provisioning path, and adoption during create-recovery keys on the brand's persisted workspace claim instead of on the title.

## 2. Reasoning
Sub-workspaces created through the UI were titled `<brand name> [<first 8 chars of brand id>]`, while ones created by the migration CLI were titled with the bare brand name. Customers saw both conventions side by side in the Semrush UI, and the id suffix means nothing to them.

The suffix was not decoration. The title is the only key two recovery paths in `ensureSubworkspace` have: ambiguous-create recovery after a 504 (`createSubworkspace` has no idempotency key, so a timed-out create is ambiguous and the parent's family listing is matched by title), and the idempotent create-or-adopt reuse from issue 2718 (an existing `created`, project-empty same-title child is reused so a retry after a partial provisioning failure does not spawn a duplicate stub). Brand display names are not unique within an org, so simply deleting the suffix would let one brand adopt a different, same-named brand's sub-workspace — and a later deactivate would then decommission the wrong brand's live markets.

Create timeouts are being hit in practice on the brand-create path, which is exactly when the title match runs.

**Titles.** Sub-workspace titles are the brand's bare display name, matching what the migration CLI already produces. Existing sub-workspaces keep their current titles — no retroactive rename is implied. The construction guard moves with the title: a missing brand *name* is now the hard error, replacing the old missing-brand-*id* error, because an untitled workspace is the thing adoption could never disambiguate.

**Adoption no longer trusts the title alone.** Before applying the match, ambiguity and emptiness rules, `findAdoptableFamilyMatch` drops every family candidate whose id is already persisted as another brand's `semrush_sub_workspace_id`. Candidates are dropped rather than escalated to an error, which keeps two existing behaviours intact: the proactive path still creates a fresh workspace when the only same-title candidate belongs to a sibling, and a genuine lone match is still adoptable when a claimed twin sits beside it in the listing. A candidate claimed by the *same* brand stays adoptable — that is a concurrent request for that brand, which the existing `reloadPointer` guard settles. The lookup is required whenever a same-title candidate exists and fails closed when it is unavailable, since without it the bare title would be the only key left.

**Failure compensation now gates on provenance.** Because a bare title can resolve to a workspace this request did not create, `ensureSubworkspace` reports a *freshly created* sub-workspace through a new `onWorkspaceCreated` signal, and every teardown path keys off that rather than off the resolved id — the concurrency-loser release, both brand-provisioning cleanups, and the brand-create compensation in the brands controller.

This closes a destructive path that dropping the suffix would otherwise have opened: two same-named brands onboarding concurrently can resolve the same workspace, the first persists its claim, and the second's brand-row write then fails on the UNIQUE constraint on `brands.semrush_sub_workspace_id`. The old compensation treated the workspace as its own orphan and would delete its projects and strip its allocation — on a live workspace correctly owned by the other brand. The unique-constraint failure is precisely the signal that someone else owns it.

Operationally visible: ignored claimed candidates are logged with their ids, and the concurrency-loser log now states whether it is releasing or standing down.

## 4. Required information
- Jira / issue: https://github.com/adobe/spacecat-api-service/issues/2915 — Closes #2915
- Implementation plan: `docs/specs/2026-06-15-serenity-subworkspace-dual-mode-implementation-plan.md` (hardening-pass note updated in this PR to describe the adoption key)
- Other: https://github.com/adobe/project-elmo-ui/pull/2493 — brand and market split UI, where the inconsistency was noticed

- `mysticat-data-service` — contract, unchanged. Its migration CLI (`scripts/serenity_migration/executor.py`) already matches child workspaces by bare title (`_child_workspaces_by_title`) and accepts the collision risk because it works from a planned batch of known, deduplicated brand names. This PR brings the API-service path onto the same titles, so both sides now produce and match one convention. The CLI's own matching is untouched.

Verified against the production database rather than assumed:

- **Same-named brands within one org are real.** Of 5767 brands, 12 org+name duplicate pairs exist today. In three of them one twin already holds a sub-workspace (`behr`, `test`, `toyota gazoo racing`) — the exact shape the claim filter has to handle.
- **The race the provenance gate closes is real, not theoretical.** Creation-timestamp deltas within those duplicate pairs include two at **0.17 seconds** (`bitdefender`, `under armour`) and two more within 11–25 seconds, so concurrent onboarding of identically-named brands in the same org demonstrably happens.
- **The UNIQUE constraint the design leans on exists.** `brands_semrush_sub_workspace_id_key`, `CREATE UNIQUE INDEX … USING btree (semrush_sub_workspace_id)` — confirmed live, so at most one brand can ever claim a workspace and the claim lookup resolves unambiguously.

## 7. Test plan

**Locally.** The PostgreSQL integration suite (`test/it/postgres/serenity.test.js`), which drives the Semrush vendor mock containers over real HTTPS and exercises activate end to end, **could not be run in this session** — the ECR image pull returns 403 because the local AWS token had expired, and refreshing it needs an interactive browser login. That suite runs in CI (`it-postgres: true`), so it gates this PR there; it is called out here rather than left implied.

**Reviewer focus.** The adoption rules are the risky surface: whether dropping (rather than erroring on) claimed candidates is right for both the proactive and the 504-recovery paths, and whether gating teardown on created-vs-adopted leaves any workspace genuinely leaked.

**On dev.**
1. Create a brand in an org that already has an active brand with the same display name. Confirm the new sub-workspace appears in the Semrush UI titled with the bare brand name, and that it is a *new* workspace — not the existing twin's.
2. Check the API-service logs for `ensureSubworkspace: ignoring same-title family entr(ies) already claimed by another brand`, carrying the skipped workspace id.
3. Confirm the pre-existing twin's workspace is untouched: its projects and its allocation are unchanged.
4. Re-run a brand create that fails after provisioning (for example by forcing the brand-row write to fail) and confirm that a workspace the request *created* is still released back to the parent pool, so the compensation has not been disabled wholesale.

**Known residual.** Two same-named brands provisioning concurrently before either persists its pointer can still resolve the same workspace. The unique constraint makes that a clean failure for the losing request instead of two brands sharing a workspace, and the provenance gate ensures the loser leaves the winner's workspace intact. Fully closing it needs an idempotency key or a client-supplied correlation id on `createSubworkspace` from Semrush, which is outside this repo.

### DIFF
--- a/docs/specs/2026-06-15-serenity-subworkspace-dual-mode-implementation-plan.md
+++ b/docs/specs/2026-06-15-serenity-subworkspace-dual-mode-implementation-plan.md
@@ -12,10 +12,17 @@
 > `decommissionBrandWorkspace`).
 >
 > **Hardening pass (post-review, gap audit):** five further safety/correctness
-> gaps were closed. (1) The sub-workspace **title now embeds the brand id**
-> (`"<name> [<uuid>]"`) so ambiguous-create recovery cannot adopt a *same-named*
-> brand's workspace; adoption additionally verifies the candidate is
-> **project-empty** before adopting. (2) `ensureSubworkspace` takes an optional
+> gaps were closed. (1) The sub-workspace **title is the brand's bare display
+> name**, which is *not* unique within an org, so ambiguous-create recovery does
+> not key on it alone: `findAdoptableFamilyMatch` **drops every candidate already
+> bound to a different brand** (`brands.semrush_sub_workspace_id`, a UNIQUE
+> column) before applying the match/ambiguity rules, and additionally verifies the
+> candidate is **project-empty** before adopting — together these keep a timed-out
+> create from adopting a *same-named* brand's workspace. Because a bare title can
+> resolve to a workspace this request did not create, `ensureSubworkspace` reports
+> a **freshly-created** workspace via `onWorkspaceCreated` and every failure-
+> compensation path releases only on that signal, never on a merely-adopted
+> workspace. (2) `ensureSubworkspace` takes an optional
 > `reloadPointer` and, on the create path, **re-reads the brand pointer before
 > persisting** — if a concurrent activation already won, it releases its own
 > freshly-created workspace's allocation and adopts the winner (residual race
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -1495,6 +1495,13 @@ function BrandsController(ctx, log, env) {
     // sub-workspace was provisioned but the brand row failed to persist, the
     // catch releases the orphaned allocation (see below).
     let provisionedWorkspaceId = null;
+    // True only when provisioning CREATED the sub-workspace upstream. Sub-workspace titles are
+    // bare brand display names, which are not unique within an org, so provisioning may instead
+    // ADOPT an existing same-title workspace — possibly one belonging to a same-named sibling
+    // brand whose own create is still in flight and has not persisted its claim yet. The
+    // compensation in the catch below tears the workspace down, so it must fire only for a
+    // workspace this request actually created.
+    let provisionedWorkspaceWasCreated = false;
 
     try {
       if (!hasText(spaceCatId)) {
@@ -1683,6 +1690,7 @@ function BrandsController(ctx, log, env) {
             writeDeadline,
           }, log);
           provisionedWorkspaceId = provisioned.semrushSubWorkspaceId;
+          provisionedWorkspaceWasCreated = provisioned.createdByThisRequest === true;
           provisionedInitialMarket = {
             projectId: provisioned.projectId,
             geoTargetId: provisioned.geoTargetId,
@@ -1700,6 +1708,7 @@ function BrandsController(ctx, log, env) {
             brandName: brandData.name,
           }, log);
           provisionedWorkspaceId = bare.semrushSubWorkspaceId;
+          provisionedWorkspaceWasCreated = bare.createdByThisRequest === true;
         }
       }
 
@@ -1775,11 +1784,18 @@ function BrandsController(ctx, log, env) {
         emitBrandDemotionBlocked(context, 'createBrand');
       }
       log.error(`Error creating brand for organization ${spaceCatId}:`, error);
-      // Compensation: a sub-workspace was provisioned upstream but the brand row
-      // failed to persist (e.g. a unique-constraint 409 or transient PostgREST
-      // error). Nothing references that workspace, so release its allocation back
-      // to the parent pool (best-effort) rather than leaking it.
-      if (provisionedWorkspaceId && hasText(provisionedWorkspaceId)) {
+      // Compensation: a sub-workspace was CREATED upstream but the brand row failed to
+      // persist (e.g. a unique-constraint 409 or transient PostgREST error). Nothing
+      // references that workspace, so release its allocation back to the parent pool
+      // (best-effort) rather than leaking it.
+      //
+      // Gated on having created it. A workspace that provisioning ADOPTED is not ours to
+      // tear down: titles are bare brand display names, so an adopted workspace can be a
+      // same-named sibling brand's, and the unique-constraint failure that lands us here is
+      // itself the signal that the sibling won the race and legitimately owns it. Releasing
+      // it would delete that live brand's projects and strip its allocation.
+      if (provisionedWorkspaceId && hasText(provisionedWorkspaceId)
+        && provisionedWorkspaceWasCreated) {
         log.error('serenity: brand-create failed after subworkspace provision; releasing orphaned allocation', {
           semrushWorkspaceId: provisionedWorkspaceId,
         });
--- a/src/controllers/serenity.js
+++ b/src/controllers/serenity.js
@@ -794,6 +794,10 @@ function SerenityController(context, log, env) {
             // positionally above (auth.parentWorkspaceId) — not duplicated in this options bag.
             dynamicAllocation: dynamicAllocationEnabled(ctx),
             ceiling: brandAiCeiling(ctx),
+            // Sub-workspace titles are bare brand names, so ensureSubworkspace needs the Brand
+            // collection to tell this brand's own interrupted create from a same-named sibling
+            // brand's workspace. Only consulted when this brand has no sub-workspace yet.
+            brandCollection: ctx.dataAccess.Brand,
           },
         );
         // Mirror this market as a SpaceCat Site (+ brand_sites link), once its
@@ -1230,6 +1234,7 @@ function SerenityController(context, log, env) {
           brandPointerReloader(ctx, auth.brandUuid),
           {
             dynamicAllocation: dynamicAllocationEnabled(ctx),
+            brandCollection: ctx?.dataAccess?.Brand,
           },
         );
         let pendingActivateSucceeded = true;
@@ -1298,6 +1303,7 @@ function SerenityController(context, log, env) {
           brandPointerReloader(ctx, auth.brandUuid),
           {
             dynamicAllocation: dynamicAllocationEnabled(ctx),
+            brandCollection: ctx?.dataAccess?.Brand,
           },
         );
         let bareSucceeded = true;
@@ -1372,6 +1378,7 @@ function SerenityController(context, log, env) {
         brandPointerReloader(ctx, auth.brandUuid),
         {
           dynamicAllocation: dynamicAllocationEnabled(ctx),
+          brandCollection: ctx?.dataAccess?.Brand,
         },
       );
       // LLMO-6554: resolved ONCE for the whole batch (same brand, so every market
--- a/src/support/serenity/brand-provisioning.js
+++ b/src/support/serenity/brand-provisioning.js
@@ -52,15 +52,17 @@ export function initialMarketProjectName(market, languageCode) {
  * quota 405 from publishing surfaces as a 409 "Quota exceeded".
  *
  * The brand row is written by the caller AFTER this resolves, so provisioning is
- * driven through a lightweight brand stub: it supplies the brand's name + the
- * pre-generated id for the "Name [id]" sub-workspace title (the adoption key) and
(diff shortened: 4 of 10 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-designed safety mechanism for a real production race condition; no blocking issues found.
**Complexity: HIGH** - large diff; API surface (controllers + handlers).
_Changes: Titles Serenity sub-workspaces with the bare brand display name, adds a claim filter to prevent cross-brand workspace adoption, and gates failure compensation on provenance so only freshly-created workspaces are released (10 files)._
Note: Recommend a human read before merge - this change amends a documented spec in `docs/specs/`. The bot review is a complement to, not a replacement for, a human read here.
Note: CI checks are partially failing (branch-deploy/deploy-stage/semantic-release) and build/it-postgres/type-check are pending - resolve before merge.

<details>

- nit: Log message uses `"entr(ies)"` which is difficult to grep in production - use `"entries"` (always plural since it only fires when count > 0) - `src/support/serenity/workspace-lifecycle.js:468`
- suggestion: `claimedIds` in the INFO log exposes workspace IDs belonging to other brands (cross-tenant identifiers in shared logs); consider logging only the count at INFO and the full IDs at DEBUG - `src/support/serenity/workspace-lifecycle.js:469`
- suggestion: The `Promise.all` claim lookup fans out one DB query per same-title candidate with no upper bound; consider a defensive cap (e.g. 10) with a 500 if exceeded, to protect against zombie accumulation from a Semrush behavior change - `src/support/serenity/workspace-lifecycle.js:460`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 25s | <b>Cost:</b> $6.67 | <b>Commit:</b> <code>ebb615a326b8a7c68c8cb20f7863be1f8ef59447</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the review in `0a1cf04bd`.

MysticatBot (all 3 non-blocking):

- nit: `entr(ies)` is hard to grep — **done**. The line only fires when at least one candidate was skipped, so the count is always plural; it now reads `ignoring same-title family entries already claimed by another brand`.
- suggestion: `claimedIds` exposes cross-tenant identifiers; log ids at DEBUG — **declined**. `GET /v1/workspaces/{parent}/family` returns only that parent's children and grandchildren, so every id in the list belongs to the same org — it is intra-org, not cross-tenant. The line already carries `title`, the brand display name, which is more identifying than a workspace UUID. Moving the ids to DEBUG would also hide them in production, where this line exists precisely so an operator can tell *which* workspace was skipped without running a manual family query.
- suggestion: cap the `Promise.all` claim fan-out, to protect against zombie accumulation — **declined**. Zombies cannot feed this fan-out: the candidate set is filtered to `status === 'created'` before the lookups, and the ack-then-fail stubs from issue 2718 are stuck at `not ready`, so they are excluded by the status filter that precedes it. What remains is bounded by the number of same-named brands in one org (production maximum today: 2), and each lookup is a point read on a unique index. A hard 500 above a cap would convert a legitimate — if unusual — customer configuration into a failed brand create, which costs more than the indexed reads it saves.

On the review's CI note: that was a snapshot taken mid-run. `it-postgres`, `type-check`, `build` and `codecov/patch` have since passed; `branch-deploy`, `deploy-stage` and `semantic-release` are `skipping` because they only run post-merge on `main`, not on a PR.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - nit addressed, pushbacks are reasonable; no new concerns.
**Complexity: HIGH** - large diff; API surface + controllers.
_Changes: Titles Serenity sub-workspaces with the bare brand display name, adds a claim filter to prevent cross-brand workspace adoption, and gates failure compensation on provenance so only freshly-created workspaces are released (10 files)._
Note: Recommend a human read before merge - this change amends a documented spec in `docs/specs/`. The bot review is a complement to, not a replacement for, a human read here.

- Log message now uses grep-friendly `"entries"` instead of `"entr(ies)"` - `src/support/serenity/workspace-lifecycle.js:220`

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 50s | <b>Cost:</b> $1.01 | <b>Commit:</b> <code>0a1cf04bd6be5b6ae610de1cf82472c3a1697e05</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Runtime validation, on commit `0a1cf04bd`.

Exercised by `ci / it-postgres` (pass): `test/it/postgres/serenity.test.js` drives `POST /serenity/activate` end to end against the Semrush vendor mock containers over real HTTPS — not in-process stubs — so the create path runs for real:

- `POST /serenity/activate` → provisions + publishes, then `deactivate` decommissions → pass. This goes through `ensureSubworkspace`, so the bare-name title is what is actually sent to `POST /v2/workspaces/{parent}/child`, and the proactive `GET /v1/workspaces/{parent}/family` check runs against the mock rather than a stub.
- That also covers the new `brandCollection` wiring on the controller path: with a same-title candidate present and no collection threaded through, `findAdoptableFamilyMatch` throws by design, so a mis-wired call site would surface here rather than silently degrade.

Not covered live, and deliberately so:

- The claim filter's actual filtering decision needs a same-title `created` candidate sitting in the parent family. Reproducing that on dev means creating a second brand with a duplicate display name, which provisions a Semrush sub-workspace that cannot be deleted (production and dev never delete one — only Semrush CS reclaims a shell). Not worth a permanent dev artifact; it is covered by unit tests instead: sibling-claimed candidate dropped, lone unclaimed match still adopted beside a claimed twin, self-claimed candidate still adopted, 502 on the 504-recovery path when the only match is a sibling's, and fail-closed when the lookup is unavailable or errors.
- The provenance gate on failure compensation is likewise unit-tested on both sides: a created workspace is still released, an adopted one is left untouched.

CloudWatch log-line confirmation was not done — that needs an interactive AWS login this session did not have.

igor-grubic (approved):

Issue #2922, opened 2026-07-28
serenity: drop the flat Semrush resource carve in ensureSubworkspace (or gate it on SERENITY_DYNAMIC_ALLOCATION)


rainer-friederich:
## Summary

`ensureSubworkspace` carves a Semrush AI resource allocation out of the org's parent
pool before a brand's sub-workspace can be used. That carve is no longer needed:
every Semrush workspace we provision against now runs with `limits_enabled: false`
and a pool far larger than any brand consumes. The carve is now pure downside — it
is a mandatory upstream write in front of every market create, it can fail the whole
request with a 502, and its failure mode is unmapped.

Turn the flat allocation off, or put the remaining un-gated allocation writes behind
`SERENITY_DYNAMIC_ALLOCATION` the same way the flat re-grant already is.

## What broke in prod

Brand `019fa7e8-dbe6-73db-9a9e-e00224bfe3fc` ("Example website 4", org
`cb9df376-a5d8-47e3-94f5-d9687a376eed`) could not add its first market from the UI.
Two attempts, 13:30:57 and 13:33:51 Zurich on 2026-07-28, both returned

```json
{"error":"serenityUpstreamError","message":"Upstream request failed"}
```

CloudWatch `/aws/lambda/spacecat-services--api-service`, RequestId
`a2322bcc-e807-45fc-8af4-f00aa40f33a3`:

```
ERROR Serenity upstream error SerenityTransportError:
  Semrush POST .../enterprise/users/api/v2/workspaces/{ws}/resources/transfer failed: 422
  at transferWorkspaceResources → ensureSubworkspace
     → handleCreateMarketSubworkspace → createMarket
  status: 422, body: { message: 'insufficient available units in subscription' }
```

The request never reached `createProject`. It died on the allocation step that runs
first.

The sizing is the reason it died. `resourceAllocation(marketCount)` in
`src/support/serenity/workspace-lifecycle.js` returns
`{projects: marketCount + 2, prompts: 500 * projects}`, and the transfer sets
**absolute** totals — so a brand's *first* market demands the child's project total be
raised to **3**, requiring 2 free units in the parent. The parent
(`bb0f4e1c-8bb1-402e-88f2-f68618ea7397`) had exactly 1 free. A one-project market
was refused because of a two-project headroom reservation it never needed.

Creating the same project directly against Semrush, with the identical body the
handler builds, returns 201 and publishes fine — the upstream product does not gate
it. Only our own pre-create transfer does.

[...]

PR #2923, opened 2026-07-28
fix(serenity): create sub-workspaces with no resource allocation

rainer-friederich:
## 1. Abstract

A brand's Semrush sub-workspace is now created with no AI resource allocation, and no lifecycle path transfers resource units onto or off it. The just-in-time allocator is retained, dormant behind `SERENITY_DYNAMIC_ALLOCATION`, as the only mechanism that can size a child.

## 2. Reasoning

The flat resource carve could only ever fail a request — it could never enable one.

Product metering and unit accounting are separate systems upstream. Our parent workspaces run with `limits_enabled: false` on the `ai` product, which disables metering of project creates and prompt writes. It does **not** stop `POST .../resources/transfer` from validating requested totals against the subscription's units, which it does strictly, returning a terminal `422 insufficient available units in subscription`.

The transfer also sets *absolute* totals rather than deltas. `resourceAllocation(marketCount)` returned `marketCount + 2` projects, so a brand's **first** market demanded the child's project total be raised to 3.

On 2026-07-28 that combination refused a one-project market for brand `019fa7e8-dbe6-73db-9a9e-e00224bfe3fc` against a parent pool holding 2 project units, and surfaced as a generic `502 serenityUpstreamError` before the request ever reached project creation. Creating the same project directly against Semrush, with the identical body the handler builds, succeeded — only our own pre-create transfer refused it.

**Before:** `ensureSubworkspace` carved an allocation out of the org's parent pool — a fixed `{projects: 1, prompts: 500}` at create, and a `{projects: marketCount + 2}` re-grant on every activation of an already-bound workspace. Deactivate and the provisioning-failure cleanup paths transferred it back down to a `{projects: 1, prompts: 1}` floor.

**After:** the sub-workspace is created with no `resources` payload and never carries an allocation. Decommission and every failure-cleanup path empty the workspace's projects and leave the shell in place, as before.

Behaviour deltas:

- A market create no longer performs a mandatory upstream write before doing its actual work, and can no longer fail for a capacity reason that does not apply to it. The activation path loses two upstream round trips (the transfer plus its settle poll); the create path loses one.
- A capacity refusal can no longer surface from the lifecycle as an opaque `502`. The JIT path, which maps the same upstream condition to a typed `409 orgPoolExhausted`, is the only remaining producer of it.
- `ensureSubworkspace` loses its `marketCount` positional parameter and its `dynamicAllocation` option — both existed solely to size the carve.
- `deleteAllProjects` now takes the org parent id and asserts against it. That guard previously lived in the release step, which ran *after* the projects were already deleted on the brand-provisioning paths, so the parent was never actually protected there. This makes it effective for the first time.

Operationally visible:

- **Children provisioned before this change keep their existing carve, and nothing reclaims it.** With the release transfers gone, deactivating such a brand no longer returns its units to the parent pool. Accepted deliberately: the pools are unmetered and effectively unbounded, so stranded units cost nothing, and `scripts/serenity-rightsizing-sweep.mjs` can reclaim them if a pool ever does matter. That script's header was corrected — it previously scoped itself to children predating the JIT allocator, but the flag was off throughout, so children created after it also carry a carve.
- The JIT allocator had no documented trigger for turning it on. `docs/serenity.md` now records one: `quotaExceeded` appearing while the allocator is off is the signal, since with no allocation to exhaust it can only mean the upstream refused on its own terms.

Deliberately out of scope, and left as follow-ups: folding `ensureSubworkspace`'s test-only `timing` seam and `reloadPointer` into its options bag; honouring `createReadiness: 'skip'` on the bound-workspace branch as well as the create branch (behaviour-affecting, and it would change reactivation latency); and parallelising `deleteAllProjects`, which is now the whole cost of a deactivate and is serial across up to 50 markets.

## 4. Required information

- Jira / issue: https://github.com/adobe/spacecat-api-service/issues/2922
- ADR(s): `docs/decisions/008-no-subworkspace-resource-carve.md`, added in this PR

**Live probe against Semrush — the load-bearing assumption.** The issue recorded, from `workspace-lifecycle.js` and `docs/serenity.md`, that a child with zero metered quota rejects prompt writes and live publish with a disguised `405`. That was observed while product limits were enforced and had never been re-probed. It is the assumption the whole change rests on, so it was probed directly against the LLMO-Dev-2 parent (`bb0f4e1c-8bb1-402e-88f2-f68618ea7397`) on `adobe-hackathon.semrush.com`, twice, on 2026-07-28.

A throwaway child created with **no `resources` body**:

- settled to `created` and reported `projects 0/0  prompts 0/0`;
- accepted `createProject`, a two-prompt draft, and `publishProject`, all without error;
- finished at `projects.used: 1` against `total: 0` — upstream permitted it to exceed a zero total outright;
- returned both prompts from the LIVE (published) prompt view, confirming the publish genuinely landed rather than succeeding as a silent no-op (a failure mode this codebase has hit before, so it was checked explicitly rather than trusting the 2xx).

Both throwaway workspaces were deleted; the parent family was re-listed afterwards to confirm no probe leftovers, and the pool read back at `projects 0/2, prompts 0/1913`.

**Vault — correcting a premise in the issue.** The issue inferred that `SERENITY_DYNAMIC_ALLOCATION` must be ON in prod, reasoning that a request which 422'd at 13:33 succeeded at 13:49 on the same Lambda version. Read directly from `dx_mysticat/prod/api-service`, it is **`false`**, and unset (so off) in stage and dev. The JIT allocator is not running in production at all. The parent's project *total* is 2, so a child asking for a total of 3 could never have been satisfied regardless of what was free — which independently rules out the flag having been on. That inference should not be carried forward; the live probe, not the flag state, is what makes this change safe.

**Pool inventory.** Every real customer parent read that day carried `limits_enabled: false` with a pool on the order of 200 projects against zero used. The only workspace that can starve is LLMO-Dev-2, our own test parent — which is exactly where the production failure fired.

## 7. Test plan

**Done locally this session:** the live end-to-end probe described in section 6 — a real sub-workspace created against the real Semrush gateway with no allocation, driven through the full metered surface (project create, prompt drafts, publish) and verified through the published read path, then torn down. This exercises precisely the path the change alters.

**dev:** create a serenity brand, then add its first market. Under the old code this was the failing request. Confirm the market goes live, and confirm via `GET /enterprise/users/api/v1/workspaces/{childId}/resources` that the child sits at `total: 0` on both AI dimensions while holding a live project. Then deactivate the brand and confirm the projects are gone and the shell remains.

**prod:** re-run the market-add for brand `019fa7e8-dbe6-73db-9a9e-e00224bfe3fc` (org `cb9df376-a5d8-47e3-94f5-d9687a376eed`), the brand from the incident, and confirm it completes. Watch `/aws/lambda/spacecat-services--api-service` for absence of `SerenityTransportError` on `resources/transfer` — after this change the lifecycle issues no transfers at all, so any surviving transfer error can only come from the JIT path, which is off.

**Rollback:** revert the commit. There is no flag to flip — the carve is removed rather than gated, which was the explicit choice over the flag-gated alternative, since the flag was off in production and gating would have left the failing behaviour switched on by default.


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-evidenced fix for a production outage, with thorough ADR, test coverage, and consistent application of the positional argument shift across all call sites.
**Complexity: HIGH** - large diff; production incident fix touching lifecycle core with API surface changes.
_Changes: removes the flat resource allocation (carve) from Semrush sub-workspace provisioning, simplifying all lifecycle paths to project-deletion-only cleanup and retaining the JIT allocator dormant behind `SERENITY_DYNAMIC_ALLOCATION` (20 files)._
Note: Recommend a human read before merge - this change adds an architectural decision record (`docs/decisions/008-no-subworkspace-resource-carve.md`). The bot review is a complement to, not a replacement for, a human read here.
Note: CI checks are partially pending (build and integration tests still running) - confirm green before merge.

<details>

- nit: `src/support/serenity/resource-manager.js` module header comment has an awkward line break from reformatting ("ABSOLUTE (sets `total`) + / idempotent; a carve" split across lines reads as a sentence fragment)
- nit: `test/support/serenity/brand-provisioning.test.js` inline comment still says "nothing to release" after the test description was renamed to "does NOT attempt a cleanup" - stale wording
- nit: `test/support/serenity/brand-provisioning.test.js` "defensive branch coverage" section still asserts `meta.error` equals `'release network error'` - terminology drift from the other renamed test that uses `'cleanup network error'`
- suggestion: add a direct unit test for the new `assertNotParent` guard inside `deleteAllProjects` (e.g. `deleteAllProjects(transport, PARENT_WS, PARENT_WS)` throws) - `test/support/serenity/workspace-lifecycle.test.js`. The guard is tested indirectly through `decommissionBrandWorkspace` and `emptyProvisionedWorkspace`, but a direct test pins the contract at the exported-function level.
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 11s | <b>Cost:</b> $7.51 | <b>Commit:</b> <code>8160f90650ce5f355f00f764b37d88d7aafac0c6</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Pushed a follow-up after the integration suite caught a real defect in the first commit.

**What broke.** Dropping the carve, I stopped passing `resources` to `createSubworkspace` altogether. `handlers.createWorkspaceV2Form` marks `resources` as **required**, while every field inside `handlers.createWorkspaceV2Resources` is optional — so omitting the key is contract-violating, and `{}` is the schema-valid way to express "no allocation". The live gateway happens to tolerate the omission (my probe against LLMO-Dev-2 succeeded that way, which is exactly why I did not catch it), but our own Counterfact vendor mock is generated from that schema and correctly refused it. Brand create returned 502 in `it-postgres`.

Worth flagging for anyone touching this transport: **`type-check` does not catch a dropped required field here.** The generated client's request body typing did not reject `resources: undefined`, so the blocking tsc gate passed on a body the schema forbids. The integration suite against the mock was the only thing that caught it — a good argument for the mock being spec-generated rather than hand-written.

**Fix.** `createSubworkspace` now always sends `resources: {}`. Re-probed live against LLMO-Dev-2 to confirm this is not merely schema-legal but behaviourally identical to omitting it: the child settles at `projects 0/0  prompts 0/0`, accepts `createProject`, prompt drafts and `publishProject`, ends at `projects.used: 1` against `total: 0`, and both published prompts read back from the LIVE prompt view. Throwaway workspace deleted.

The transport unit test now pins the exact wire body, so the required field cannot be silently dropped again at the layer where it actually matters.

One correction to the issue, repeated here since it changes how the change should be read: the issue inferred `SERENITY_DYNAMIC_ALLOCATION` was ON in prod. Read from Vault, it is `false` (and unset in stage/dev). The JIT allocator is not running in production, so nothing sizes a new sub-workspace at all — that is safe because of the live probe, not because of the flag.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - the new commit correctly fixes a schema compliance defect, and the overall change is well-evidenced and clean.
**Complexity: HIGH** - large diff; production incident fix touching lifecycle core with API surface changes.
_Changes: fixes `createSubworkspace` to send `resources: {}` (schema-required) instead of omitting the key, after the integration suite's spec-generated mock caught the contract violation (21 files total across both commits)._
Note: Recommend a human read before merge - this change adds an architectural decision record (`docs/decisions/008-no-subworkspace-resource-carve.md`). The bot review is a complement to, not a replacement for, a human read here.
Note: CI checks are partially pending (build and integration tests still running) - confirm green before merge.

<details>

- nit: `src/support/serenity/resource-manager.js` module header comment has an awkward line break from reformatting ("ABSOLUTE (sets `total`) + / idempotent; a carve" split across lines reads as a sentence fragment)
- nit: `test/support/serenity/brand-provisioning.test.js` "defensive branch coverage" section still asserts `meta.error` equals `'release network error'` while the equivalent test in provisionBrandSubworkspaceBare was renamed to `'cleanup network error'` - terminology drift between the two failure paths
- nit: `test/support/serenity/brand-provisioning.test.js` inline comment still says "nothing to release" after the test description was renamed to "does NOT attempt a cleanup" - stale wording
- suggestion: add a direct unit test for the `assertNotParent` guard inside `deleteAllProjects` (e.g. `deleteAllProjects(transport, PARENT_WS, PARENT_WS)` throws) - `test/support/serenity/workspace-lifecycle.test.js`. The guard was correctly moved into the primitive in this PR, which is the important structural fix; a direct test pins the contract at the exported-function level.
</details>

- Schema compliance: `createSubworkspace` now sends `resources: {}` (schema-required by `createWorkspaceV2Form`) instead of omitting the key entirely - caught by integration suite, fixed in commit abb6bf43.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 9s | <b>Cost:</b> $5.81 | <b>Commit:</b> <code>abb6bf43b6b64e5bd8b065eb49a50c31342c8f6a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed all four non-blocking items in `fd7e37990`.

- **`resource-manager.js` header** — re-wrapped the transfer-semantics paragraph; the sentence no longer breaks mid-clause. Wording unchanged apart from `+` reading as `and`.
- **`brand-provisioning.test.js` "nothing to release"** — now "nothing to empty", matching the renamed test.
- **`brand-provisioning.test.js` terminology drift** — the defensive-branch test now uses `'cleanup network error'`, aligning both failure paths.
- **`assertNotParent` direct test** — added to `workspace-lifecycle.test.js`. It also asserts the guard fires *before* `listProjects`, which is the property that was actually broken before this PR: the guard used to sit in the release step, which ran after the projects were already deleted.

Two further stale references in the same family, not flagged but part of the same drift: the enclosing `describe` named `releaseCapturedOnFailure`, a function this PR removed (the current name is `emptyCapturedOnFailure`), and it carried a brittle `(lines 141-145)` reference that no longer pointed at the catch block. Both corrected, and the line-number reference dropped rather than updated. The `makeReleaseTransport` test helper is now `makeCleanupTransport`.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all four non-blocking items from the prior review are cleanly addressed; no new concerns from the follow-up commit.
**Complexity: HIGH** - large diff; production incident fix touching lifecycle core with API surface changes.
_Changes: addresses all four non-blocking items from the prior review - comment rewrap, terminology alignment, and a direct unit test for the assertNotParent guard in deleteAllProjects (3 files)._
Note: Recommend a human read before merge - this change adds an architectural decision record (`docs/decisions/008-no-subworkspace-resource-carve.md`). The bot review is a complement to, not a replacement for, a human read here.

- Module header comment in resource-manager.js no longer breaks mid-sentence
- brand-provisioning.test.js inline comment now says "nothing to empty" matching the renamed test
- Defensive branch coverage test now uses 'cleanup network error' matching the other failure path
- Direct unit test added for assertNotParent guard inside deleteAllProjects, verifying it fires before the listing

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 51s | <b>Cost:</b> $2.85 | <b>Commit:</b> <code>fd7e37990a0cb38ec34f6e1fcd07911dac774186</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2560, opened 2026-06-05
feat(backfill-llmo): simplify cdn-logs-report backfill to per-day DB imports

akshaymagapu:
## Summary

Simplifies the `cdn-logs-report` backfill to a single model — **a backfill is a set of traffic days**. The command enumerates the UTC days to backfill (oldest first) and queues one date-based daily-import message per day. The audit worker imports each day's agentic + referral traffic into PostgreSQL; the projector refreshes the daily serving table on every import and the weekly rollup when a week's closing Sunday lands.

## Commands (`audit=cdn-logs-report`)

| Arg | Backfills |
|-----|-----------|
| `weeks=N` | last N completed ISO weeks (default 2, max 4) |
| `weeks=0` | current week to date (Mon → yesterday) |
| `days=M` | last M days ending yesterday (max 31) |
| `date=YYYY-MM-DD` | a single traffic day |
| `mode=weekly-db date=YYYY-MM-DD` | force the weekly rollup refresh for that completed ISO week (`wrpc_refresh_agentic_traffic_weekly`) |

Days are queued oldest→newest, staggered 5s, as `{ auditContext: { date: trafficDay + 1 } }` (the worker exports the day before `date`). Each day import is idempotent and independently retryable.

## Removed / replaced

- The old `weekOffset` fan-out path (API emitted week-offset messages that the worker re-fanned into 7 daily messages) — the API now enumerates days directly.
- The `mode=db` daily flag (plain `date=` does it).

## Testing

`npm run lint` clean; backfill-llmo command tests passing; command-file coverage ~100% (gate 90%).

PR #2926, opened 2026-07-28
feat: introduce projectId CSV parameters in the serenity apis

vivesing:
# Serenity Elements API: `region` → `projectId` migration

## Summary

Replaced the `region`/`regionCode`/`region_code` query parameter with a `projectId`/`project_id` query parameter (comma-separated for multi-value) across all Serenity Elements endpoints. Callers now pass Semrush project UUID(s) directly instead of a market/region code that had to be resolved server-side via a Markets-element lookup.

Corresponding UI PR https://github.com/adobe/project-elmo-ui/pull/2560 

## Endpoints affected

All in `src/controllers/elements.js`:

- `listCitedDomains`, `listSentimentOverview`, `listTopics`, `listTopicPrompts`
- `listOwnedUrls`, `listDomainUrls`
- `getUrlInspectorStats`, `getUrlInspectorPromptsCount` (via shared `resolveUrlInspectorScope`)
- `getMarketTrackingTrends`, `getStats`, `getCompetitorSummary`, `getKpiHeadlines`, `getSourceVisibilityHeadline`
- `listUrlInspectorFilterDimensions` (added afterward — its `TOPICS` element also accepted a `region`-like `projectId`)

## Behavior changes

| Aspect | Before | After |
|---|---|---|
| Param | `region` / `regionCode` / `region_codctId` / `project_id` (CSV, e.g.`projectId=uuid1,uuid2`) |
| Resolution | Region code → Semrush project ID via a live Markets-element lookup | None — caller-supplied ID(s) used
directly |
| Absent param | Aggregate across all the brand's projects | Same (unchanged fallback) |
| `region=all` | Special-cased as "aggregate" | No special-casing — `projectId=all` is now a literal (bogus) ID, not a keyword |
| Multi-value | Not supported (single region on, via `CBF_project`/`CBF_projects` OR-filterswhere the element allows it, or per-project fan-out where it doesn't |

## Element-level payload changes

- **`CBF_project`/`CBF_projects` OR-filter elements** (`SENTIMENT`, `PROMPTS_BY_TOPIC`, `TOTAL_EXECUTIONS`, `MENTIONS`, `VISIBILITY`, `CITATIONS_KPI`, `TRENDS_MV`, `MARKET_CITATIONS_TREND`, `KPI_SHARE_OF_VOICE`, `KPI_BRAND_VISIBILITY`,
`KPI_SOURCE_VISIBILITY`, and now `TOPICS`): all call**, OR'd together.
- **Top-level `project_id` elements** (`CITED_DOMAINS`, `STATS_PER_URL`, `URL_TRENDS`): only accept one ID per call, so multiple IDs trigger a **fan-out** — one call per ID, results merged afterward.
  - New: `CITED_DOMAINS` fan-out/merge logic ins.js` (`transformCitedDomainsResponses`) — sumsnumeric fields for domains cited under more than one project, backfills `contentType`, sorts/paginates the merged set once.
  - `STATS_PER_URL`/`URL_TRENDS` (owned-urls/dos fan-out pattern; just fed caller-supplied IDsinstead of Markets-resolved ones.
- **`TOPICS`** (filter-dimensions): moved from top-level `project_id` to a `CBF_project` OR-filter, per explicit
instruction, since the element supports the fil

## Removed

- `resolveRegionProjectId` (dead code) from `src/support/elements/elements-service.js`.
- All region-resolution branches, 404-on-unresolvable-region guards, and related `brandId`/`brand` variables that became
unused once resolution was removed.

## Known side effect

- `listOwnedUrls`'s Postgres traffic-join RPC nn` filter (no Semrush-project-id equivalent), soagentic/referral traffic now aggregates across all regions instead of filtering to one when a single project is selected.

## Tests

- Updated `test/controllers/elements.test.js` and `test/support/elements/definitions/topics.test.js` to match the new `projectId`-based contract (removed obsolete `region`/`resolveRegionProjectId` assertions).
- Added new coverage: `test/support/elements/dest.js` (new file — payload builder,single-response transform, fan-out/merge), plus multi-`projectIds` tests in `sentiment-overview.test.js`, `topic-prompts.test.js`, `topics.test.js`, and a new `getCitedDomains` describe block in `elements-service.test.js`.
- Full suite: 16,072 passing, 0 failing; lint clean.


src/controllers/elements.js
```diff
@@ -158,6 +158,21 @@ function splitCsv(value) {
...
+/**
+ * Extracts caller-supplied Semrush project ids from the `projectId`/`project_id`
+ * query param (CSV, e.g. `projectId=uuid1,uuid2`). Replaces the old single-value
+ * `region`/`regionCode` param — the caller now supplies the Semrush project id(s)
+ * to scope to directly, instead of a market/region code that had to be resolved
+ * via the Markets element. Absent/empty → caller wants the aggregate view across
+ * every project the brand owns.
+ *
+ * @param {object} query - Parsed query params (from `extractQuery`).
+ * @returns {string[]} Requested project ids, or [] for the aggregate view.
```
MysticatBot: issue (blocking): **No input validation or count cap on caller-supplied projectIds.**

`extractProjectIds` splits on commas and trims, but performs no validation on the resulting values:
- No UUID format check
- No upper bound on CSV length

Without a cap, a single request with `projectId=id1,id2,...,id200` triggers 200 upstream Semrush API calls via the cited-domains fan-out (concurrency is bounded at 8, but total work is unbounded). The OR-filter builders (sentiment, topics, topic-prompts) would generate filter clauses with hundreds of entries.

The old `resolveRegionProjectId` provided implicit validation (the Markets element lookup rejected unknown regions). The new path trusts the caller entirely.

**Fix:** Add a UUID format regex and a reasonable max-count cap (e.g., 20) in `extractProjectIds`. Return 400 if exceeded. Document the limit in the OpenAPI spec.

src/controllers/elements.js
```diff
@@ -706,27 +697,12 @@ export default function ElementsController(context, log, env) {
 
       const service = await buildService(ctx);
```
MysticatBot, line 785: issue (blocking): **Missing authorization cross-check on caller-supplied project IDs.**

Previously, `resolveRegionProjectId` resolved the region code against the brand's own `BrandSemrushProject` rows via the Markets element, implicitly enforcing that a caller could only scope to projects their brand owns. With this change, the caller supplies project UUIDs directly and they flow to Semrush with no ownership check.

In this endpoint (and `listTopics`, `listTopicPrompts`, `listCitedDomains`), when projectIds are caller-supplied, the code does NOT fetch or check `brandSemrushProjects` at all. An authenticated caller who knows another brand's project UUID (within the same workspace) could query that brand's data.

In endpoints like `getStats`, `getMarketTrackingTrends`, `listKpiHeadlines` etc., `brandSemrushProjects` IS fetched for the aggregate fallback but not used to validate the caller-supplied IDs.

**Fix:** After extracting `requestedProjectIds`, validate they are a subset of the brand's `brandSemrushProjects` (fetch if not already available). Return 403 for IDs not owned by the brand. This restores the access-control invariant the Markets resolution previously enforced.

MysticatBot (changes requested):
Hey @vivesing,

**Verdict: Request changes** - two material gaps in input validation and authorization need to be addressed before merge.
**Complexity: HIGH** - large diff; API surface change across all Serenity endpoints.
_Changes: Replaces `region`/`regionCode` query parameters with caller-supplied `projectId`/`project_id` (CSV) across all Serenity Elements endpoints, removing server-side Markets-element resolution and adding fan-out/merge for multi-project cited-domains calls (14 files)._
Note: Recommend a human read before merge - this change modifies a shared API contract (OpenAPI spec) and touches project documentation. The bot review is a complement to, not a replacement for, a human read here.
Note: CI checks are still pending (build and it-postgres running).

## Must fix before merge

1. [Important] No input validation or count cap on caller-supplied projectIds - `src/controllers/elements.js:170` (details inline)
2. [Important] Missing authorization cross-check on caller-supplied project IDs - `src/controllers/elements.js:699` (details inline)

<details>

- nit: OpenAPI spec drops `format: uuid` constraint on `projectId` without replacement (now `type: string` for the CSV, but a `pattern` for comma-separated UUIDs would help gateway validation) - `docs/openapi/serenity-api.yaml`
- suggestion: Hoist `CITED_DOMAINS_PROJECT_CONCURRENCY = 8` to module scope with the other constants and add a one-line rationale comment - `src/support/elements/elements-service.js`
- suggestion: The Postgres traffic regression in `listOwnedUrls` (no `region` filter passed to RPC) is acknowledged in-code but should be explicitly called out as a product-accepted trade-off in the PR description for future archaeology - `src/controllers/elements.js`
- question: The Topics element switched from top-level `project_id` to `CBF_project` filter (unlike Cited Domains which explicitly ignores CBF filters). The PR body says "per explicit instruction, since the element supports the filter" - can you confirm this was validated with Semrush? A brief inline comment noting the verification would help future readers. - `src/support/elements/definitions/topics.js`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 18s | <b>Cost:</b> $6.41 | <b>Commit:</b> <code>becce701bd83348ac7a40d4ae501c0d7bdd24f34</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

calvarezg:
## Multi-persona review — additional findings

Ran Architect / DBA / Security / Tester+Junior personas over the full diff. The two **blocking** items I found (missing ownership check on caller-supplied `projectId`s, and no cap/validation on the CSV) are the same as MysticatBot's — not repeating them here. Below are only the findings **not** already in that review.

### Major
**`region`/`regionCode` removed from the contract, not deprecated.** All `region` params are deleted from the OpenAPI spec and no longer read anywhere. A client still sending `region=US` is silently ignored and receives the aggregate view across all the brand's markets, with a `200`. The old code deliberately `404`'d on an unresolvable region to avoid returning a superset. Silent scope-widening is worse than a hard error.
*Fix:* confirm the UI (#2560) is the only caller and deploys in lockstep. If not, reject known-legacy params (`region`/`regionCode`) with a `400` for one release.

### Minor
- **Duplicate ids double-count in the cited-domains fan-out.** `getCitedDomains` does `projectIds.filter(hasText)` with no dedup, then fans out one call per id and `mergeDomainRows` sums. `projectId=X,X` doubles `totalCitations`/`totalUrls`/`promptsCited` for X. The OR-filter builders are immune (a duplicated OR term is a no-op) — only the fan-out path corrupts counts. Reachable because the CSV is unvalidated. *Fix:* `[...new Set(ids)]` before fan-out; add a test. (A validation fix for the CSV-cap blocker naturally covers this.)
- **Aggregate-fallback + OR-filter logic duplicated across ~6 endpoints / 3 builders.** Four OR-filter endpoints fall back via `brandSemrushProjects.map(p => p.semrushProjectId)`; the URL-inspector/owned/domain-urls endpoints fall back via `service.getOwnedUrlProjects`. The projectIds-extraction + empty-scope branch is copy-pasted ~6 times; the OR-filter push block is copy-pasted across `sentiment-overview.js`, `topic-prompts.js`, `topics.js`. The ownership fix (the open blocker) will need to touch every copy — they will drift. *Fix:* extract one `resolveProjectScope(...)` controller helper and one `buildProjectFilter(projectId, projectIds)` payload helper.
- **Fan-out error-propagation test only exercises the single-call path.** The "propagates a rejected upstream call" test in `elements-service.test.js` passes no `projectIds`, so it hits the single `fetchElement`, not the `mapWithConcurrency` fan-out. The fan-out's rejection behavior is unasserted. *Fix:* add a rejection test with `projectIds: ['a','b']`.

### Questions
- **Does the UI still send legacy `region`/`all`?** `projectId=all` is now a literal (bogus) id with no aggregate special-case — a caller sending `all` gets empty/garbage results, not the aggregate view. Confirm #2560 fully migrated.
- **Does the owned/domain-urls transform still depend on `p.region`?** The scoped path now builds `[{ projectId }]` (no `region`), while the aggregate path (`getOwnedUrlProjects`) still yields `{ region, projectId }`. If the transform tags URLs by `p.region`, scoped results lose region tagging. (Low confidence — the transform is not in this diff.)

### Working well
Cited-domains merge logic (dedup by domain, sum, `contentType` backfill, single sort/paginate) is correct and well covered by the new `cited-domains.test.js`. Project ids go into structured JSON `val` fields, so no injection.

src/controllers/elements.js
vivesing: fixed

src/controllers/elements.js
vivesing: fixed

MysticatBot (approved):
Hey @vivesing,

**Verdict: Approve** - both prior blocking findings are properly addressed; no new blocking issues.
**Complexity: HIGH** - large diff; API surface change across all Serenity endpoints.
_Changes: Replaces `region`/`regionCode` query parameters with caller-supplied `projectId`/`project_id` (CSV) across all Serenity Elements endpoints, adding UUID validation, count cap, deduplication, and brand-ownership authorization (14 files)._

<details>

- nit: `extractProjectIds({ projectId: '  , , ' })` returns `[]` (aggregate view) when the param is present but empty after cleaning - a 400 would be more explicit for callers who accidentally sent a malformed value - `src/controllers/elements.js:474`
- suggestion: Hoist `CITED_DOMAINS_PROJECT_CONCURRENCY = 8` to module scope alongside the other constants and add a one-line rationale - `src/support/elements/elements-service.js:260`
- suggestion: The repeated projectIds-extraction + empty-scope + ownership-check sequence is copy-pasted across ~8 endpoints - extracting a shared `resolveProjectScope(ctx, brand)` helper would reduce the surface for a future ownership-check omission - `src/controllers/elements.js`

</details>

- Input validation and count cap on caller-supplied projectIds now enforced via `extractProjectIds` (UUID format check, dedup, cap at 8, returns 400)
- Authorization cross-check on caller-supplied project IDs now enforced via `checkProjectIdsOwnership` (checks against brand's BrandSemrushProject rows, returns 403)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 59s | <b>Cost:</b> $4.00 | <b>Commit:</b> <code>02a2c5a491c6190c1c15c2cb07591aedd29f4617</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

calvarezg:
## Re-review after `02a2c5a4` ("address all review comments") — blockers resolved

Verified each fix against the head file, not just the diff.

### Both blocking findings fixed ✅
- **Authorization / IDOR** — `checkProjectIdsOwnership` now intersects caller-supplied ids against the brand's own `BrandSemrushProject` rows and returns 403 on any unowned id, in all ~10 endpoints. Confirmed the ownership list is scoped to the caller's own brand everywhere — including `listUrlInspectorFilterDimensions` (`spacecatBrands = [auth.brand]`), so the cross-brand-workspace hole is genuinely closed. Per-endpoint 403 tests added.
- **No cap / no validation** — `extractProjectIds` dedups, caps at `MAX_PROJECT_IDS = 8`, and rejects non-UUIDs with a 400 (confirmed `mapError` maps `ErrorWithStatusCode` to a real 400, not a masked 500). OpenAPI got a UUID + max-8 `pattern` on every param. Over-cap and bad-UUID tests added.

### My earlier net-new findings
- **Duplicate double-count** — fixed (dedup in both `extractProjectIds` and `getCitedDomains`; test added).
- **Fan-out error-path test** — fixed (new test asserts rejection propagates through the `mapWithConcurrency` fan-out).
- **`projectId=all`** — now returns 400 instead of silently scoping to a bogus id. Fails loud. 👍

### Remaining (non-blocking)
- **`region` removed, not deprecated** — a stale client sending `region=US` is still silently ignored → aggregate view. Only matters if a non-UI caller still sends legacy `region`. Worth confirming #2560 is the only caller.
- **Traffic RPC over-counts under a scoped view** — acknowledged product trade-off in the PR body; not a code defect.
- **Boilerplate** — the fetch-extract-check-return block is copy-pasted across ~10 endpoints (the check helper itself was extracted). Cosmetic. `listTopics`/`listTopicPrompts` also fetch `brandSemrushProjects` on the aggregate path just for the check — a wasted read; gate behind `requestedProjectIds.length > 0` if you like.

CI all green (build, it-postgres, type-check, Mysticat pr-review).

**Recommendation: Approve.**

calvarezg (approved):
Approving — multi-persona re-review confirms both blocking findings (authorization cross-check on caller-supplied projectIds, and CSV validation/cap) are resolved and verified against the head file. Remaining items are non-blocking (acknowledged traffic trade-off, legacy `region` migration note). CI green.

PR #2928, opened 2026-07-28
fix(serenity): make the Semrush transport a real type-check gate

### LINKED ISSUE #2925: type-check does not reach calls made through `@param {object}` values, so a contract-violating Semrush body shipped past a green gate
## Summary

The opt-in `type-check` gate (ADR-005) does not check calls made through the serenity transport, or through any other value that arrives as `@param {object}`. Such a value is `any`, so member existence, argument count and argument types are all unenforced on it. A contract-violating request body reached the wire past a green `type-check`, and only the spec-generated vendor mock in `it-postgres` caught it.

This is about the gate's reach, not about the generated Semrush client types, which are correct and do enforce the contract when a typed value actually reaches them.

## What happened

In the sub-workspace no-allocation change, `workspace-lifecycle.js` stopped passing an allocation and called:

```
transport.createSubworkspace(parentWorkspaceId, title)
```

against a transport method still declared as `createSubworkspace(parentWorkspaceId, title, resources)`. Inside the transport, `resources` was therefore `undefined` and the body `{ title, resources }` serialized to `{ title }`.

`resources` is **required** by `handlers.createWorkspaceV2Form` (every field *inside* `createWorkspaceV2Resources` is optional, so `{}` is the schema-valid way to express "no allocation"). The live Semrush gateway accepts the omission, so a live end-to-end probe against LLMO-Dev-2 passed. `type-check` passed. `it-postgres` failed with 502-instead-of-201, because the Counterfact mock is generated from the spec and correctly refuses the body.

## Root cause

Three behaviours compose. Each was verified with `tsc` against this repo's `tsconfig.json`, not inferred.

**1. The generated client types are sound.** Calling the typed User Manager client directly with a malformed body is rejected. All of the following error today: `resources` omitted (TS2322 "Property 'resources' is missing"), `resources: undefined` (TS2322), body omitted (TS2345), unknown property (TS2353), wrong primitive (TS2322), missing path param (TS2345), unknown route (TS2345). So the contract is enforced — at the point where a typed value meets the client.

**2. `@param {object}` is `any`, and unknown-member access is silent.** `tsconfig.json` sets `noImplicitAny: false`. Under that setting, in JS files, TS suppresses TS2339 entirely. A function declared `@param {object} transport` accepts `transport.anyNameAtAll()` with no error; flipping `noImplicitAny` to `true` makes the same line report TS2339. Every serenity lifecycle function receives its transport this way, so no call through `transport.*` is checked at all.

There are roughly 140 `@param {object}` / `@param {Object}` annotations across the files in the `include` set — 19 in `workspace-lifecycle.js`, 11 in `brand-provisioning.js`, 8 in `rest-transport.js` itself.

**3. Undocumented parameters are optional, so arity is not enforced either.** This is the part that makes "just annotate the transport" insufficient. Typing the transport as `ReturnType<typeof createSerenityTransport>` yields a correct 34-method object type, but because the transport methods' own parameters carry no JSDoc, they are implicitly `any` and TS treats them as optional in a JS file. Against that inferred type, `createSubworkspace('a')` and even `createSubworkspace()` both pass — TS reports the signature as "0-2 arguments".

The contrast within the same object is the clearest evidence: `createPromptsByIds` *is* JSDoc-typed `(string, string, string[], string[])`, and against the identical inferred transport type it correctly reports "Expected 4 arguments, but got 2" and rejects a number passed for a string.

So the chain is: an `any` transport erases the call site, and undocumented method parameters erase arity even after the transport is typed. The sound generated types are simply never reached.

## Why the other gates did not cover it

- **Live probe** — the real gateway tolerates the omitted key, so exercising the real API cannot detect it.
- **Unit tests** — the transport is stubbed, so the wire body is not validated against the spec.
- **`it-postgres`** — the only gate that caught it, because the Counterfact mock is generated from the OpenAPI spec and validates the body. That makes the mock-backed integration suite, not `type-check`, the de-facto contract gate against Semrush today.

## Proposed direction

ADR-005 already records `noImplicitAny: false` as a deliberate, temporary relaxation to be ratcheted to `true` later. This is a concrete argument for doing that, and a measurement of the cost.

Flipping `noImplicitAny: true` on the current `include` set yields **815 errors**:

| code | count | nature |
|---|---|---|
| TS7006 / TS7031 / TS7034 / TS7005 / TS7053 | 431 | missing parameter and variable annotations |
| TS2339 | 320 | member access on a value annotated `{object}` |
| TS2345 / TS2322 / TS18047 | 19 | assignability and possibly-null |
| TS7016 | 1 | `@adobe/helix-shared-utils` ships no declarations |

Concentrated in `controllers/brands.js` (151), `rest-transport.js` (76), `handlers/markets.js` (75), `handlers/markets-subworkspace.js` (65), `handlers/prompts.js` (59).

Most of the 19 assignability errors look like fallout from the same loose annotations (`object` narrowing where a `string` is wanted) rather than latent defects, but two possibly-null reports in `brands.js` warrant a look. That set needs triage; nothing here asserts they are bugs.

Suggested sequencing, smallest useful step first:

1. **Give the transport a name and annotate its methods.** Export a `SerenityTransport` typedef from `rest-transport.js`, replace `@param {object} transport` with it across the serenity module, and JSDoc the transport methods' parameters. Both halves are required — the typedef alone does not restore arity checking, as measured above. This is what would have caught this specific defect, and it covers the highest-value surface (every outbound Semrush call) without touching the other files.
2. **Then ratchet `noImplicitAny` to `true`**, per-directory or via a baseline, rather than in one sweep.

Independently worth deciding: whether the spec-generated mocks should be acknowledged as the contract gate. They currently are, in effect, and they only run in `it-postgres`.

## Reproduction

Add a file under `src/support/serenity/` with `// @ts-check` and run `npx tsc -p tsconfig.json`:

```js
/** @typedef {ReturnType<typeof createSerenityTransport>} SerenityTransport */

/** @param {object} loose */
export async function a(loose) {
  await loose.noSuchMethodAnywhere();          // silent; TS2339 under noImplicitAny
}

/** @param {SerenityTransport} t */
export async function b(t) {
  await t.createSubworkspace();                 // silent: params undocumented -> optional
  await t.createPromptsByIds('a', 'b');         // TS2554: params are JSDoc-typed
}
```

## Context

Surfaced by https://github.com/adobe/spacecat-api-service/pull/2923 (issue https://github.com/adobe/spacecat-api-service/issues/2922). The defect itself is already fixed there; this issue is only about the gate that did not catch it.

rainer-friederich:
## 1. Abstract

Makes the Semrush transport a value the type-checker can actually see: `rest-transport.js` exports a `SerenityTransport` typedef and documents every transport method's parameters, and every parameter across the serenity module that receives the transport is annotated with it — the fifty-nine previously annotated `{object}` / `{any}` and the twenty-four that carried no `@param` tag for the transport at all.

## 2. Reasoning

The opt-in `type-check` gate did not reach calls made through the serenity transport, or through any other value annotated `@param {object}`. Such a value is `any`, so member existence, argument count and argument types were all unenforced on it.

The consequence was not theoretical. A sub-workspace create dropped its `resources` argument against a method then declared with three parameters; `resources` was therefore `undefined` and the body serialized without the key. `resources` is required by the vendor contract — every field *inside* it is optional, so `{}` is how the spec expresses "no allocation". That body reached the wire past a green gate. The live Semrush gateway tolerates the omission, so a live end-to-end probe passed too; only the spec-generated vendor mock refused it.

Two properties compose to produce this, and fixing either alone enforces nothing:

- An `{object}`-annotated value is `any`, which erases the call site entirely.
- An undocumented parameter is implicitly `any`, and TypeScript treats it as *optional*. So even against a correctly inferred 35-method transport object, a method with no `@param` tags accepts any argument count — the transport's own `createPromptsByIds`, which was already JSDoc-typed, correctly rejected a short call while its neighbours did not.

The second property also applies to the *receiving* side, which is why annotating the `{object}` parameters was not sufficient on its own. Twenty-four serenity functions — `handleCreateMarket`, `pollUntilCreated`, `resolveLanguageId` and others — took the transport with no `@param` tag whatsoever, so they were never `{object}`-annotated and were invisible to a search for that pattern while being just as unchecked.

This is primarily a build-time gate change. Four runtime touches fell out of it, described at the end; one of them fixes a real defect.

**The transport is now a typed seam,** on both sides. `rest-transport.js` exports a `SerenityTransport` typedef and gives all 35 methods `@param` tags, and every function in the module that receives a transport names that type. Request shapes are *derived from the generated vendor contracts* rather than restated, so a vendor spec change surfaces at the call site instead of on the wire. Two derivation styles are used deliberately: the Project Engine facade's request bodies are reached through its own method signatures, because the generator emits `requestBody` inline for some operations and via a shared `requestBodies` map for others, and only the signature normalises both; the User Manager allocation object is named directly from its schema, because it is a nested field shared by two differently-wrapped call sites rather than a whole method body.

Before, a lifecycle function receiving the transport had no checking on anything it called through it. After, wrong argument counts, wrong argument types and wrong request-body shapes are all build failures.

**A strict tier.** `noImplicitAny: false` suppresses unknown-member reporting in JS files *outright*, even against a fully typed receiver — so a method-name typo, and any wrong assumption about a response's shape, stays invisible to the base tier no matter how well the value is typed. `tsconfig.strict.json` turns that on over an explicit file list, starting with `rest-transport.js` (the module that defines the outbound contract for the whole surface). `npm run type-check` runs both tiers, so the CI job and the pre-commit hook are untouched and the strict pass cannot be forgotten. The file list is enumerated rather than globbed so an unfixed file cannot join by accident.

**Regression guards, one per tier.** `test/types/` holds type-level fixtures that are checked by `tsc` and never executed. `base/` asserts that the arity / argument-type / body-shape errors actually fire — drop a transport method's `@param` tags, or widen the parameter back to `{object}`, and the directives go unused and the build fails. `strict/` asserts that an unknown member is reported, which is the strict tier's whole reason for existing; removing `noImplicitAny` from its config was otherwise a silent no-op.

The per-tier split is load-bearing, not tidiness: a TS2339 assertion is suppressed in the base tier, so a single shared fixture would fail it with an unused-directive error.

Those fixtures are enforced by being *compiled*, which makes them silently removable — drop the directory from a tsconfig `include`, or delete the file, and `type-check` still exits 0 because nothing is left to complain. `test/types/fixtures-are-wired.test.js` closes that: it is a normal mocha test asserting each fixture is still listed in its tier's `include` and still carries assertions. The guards pin the types; it pins the guards.

**Runtime touches.** Four, three of which are behaviour-identical.

`handlers/markets.js` had a genuine typing defect: `.filter(Boolean)` does not narrow away the `undefined` a `Map.get()` carries; it is now an explicit `!== undefined`, identical in effect because the values were guarded by `hasText` on insertion. `usersBaseUrl()` binds its env value before guarding it, because `hasText` is typed to take a `string` and so cannot accept a `string | undefined` — this is the narrowing idiom this directory already documents. And once the previously-undocumented functions became checked, the `createProject` body in `handleCreateMarket` no longer type-checked: as a plain object literal its `type: 'ai'` inferred as `string`, widening away from the enum the generated contract declares, so the body is now typed from `ProjectCreateBody` where it is built.

The fourth **does** change behaviour, and is the second real defect this gate surfaced. The sub-workspace market delete keys off a listed project's `id` twice — the upstream `DELETE` addresses the project by id, and so does the mapping-row tombstone. A matched project carrying no id was previously carried into both, tombstoning against a blank id and reporting a silent success. It now returns `204` with a warning naming the contract break, and both are skipped.

Worth being precise about why that was reachable, because it is the sharpest illustration of what this gate does *not* yet cover: the generated contract declares `id: string`, required and non-nullable, so on paper the read was safe. But `listProjects`'s response resolves to `any` at every call site — the vendor facade's own return type degrades — so nothing enforced the declaration, and the live gateway is documented to return bodies the spec forbids. Naming the transport fixed the *request* side; the response side is still unchecked, which is what let this through.

**Residual gap, deliberately left.** 167 `@param {object}` and 49 `@param {any}` annotations remain in the checked set, on `log`, `context` and DTO-ish payloads. Closing them is the same job as ratcheting `noImplicitAny` — bare `object` has no members, so under that flag a member access on an `{object}` value is an error and the annotation has to become a real type. ADR-005 records the measured cost of the whole remaining step and the two cross-cutting shortcuts (`log` and `context` typedefs) worth taking first.

## 4. Required information

- Jira / issue: https://github.com/adobe/spacecat-api-service/issues/2925 — Closes #2925
- ADR(s): `docs/decisions/005-opt-in-type-checking.md` (updated here — the reach of the gate, the strict tier, and the ratchet cost)
- Other: surfaced by https://github.com/adobe/spacecat-api-service/pull/2923 for https://github.com/adobe/spacecat-api-service/issues/2922 — that PR fixed the defect itself; this one closes the gate that missed it

- `spacecat-shared` — contract dependency, not changed here. The new typedefs derive their request shapes from `@adobe/spacecat-shared-project-engine-client` and `@adobe/spacecat-shared-user-manager-client`. That is the point: a released change to either generated contract now surfaces as a build failure in this repo rather than as a malformed request. It also means a breaking regeneration upstream will fail this repo's gate at bump time, which is the intended early-warning.

- **`hasText` semantics verified by execution**, since two comments in this change rest on them: it does not trim (`'   '` is truthy to it) and it rejects non-strings. That confirms the `usersBaseUrl` rewrite is behaviour-identical for unset, empty, whitespace, non-string and valid-URL inputs — the added truthiness check exists only to narrow for TypeScript.
- **Transport responses resolve to `any`, verified by probe.** `listProjects` is declared to return `model.ProjectListResponse`, but assigning its result to `never` reports `any` under both tiers — the degradation is in the vendor facade's `TransportData` helper, not in this repo's wrapper. That is why the nullable-looking `project.id` in section 3 was reachable despite a required, non-nullable declaration, and it is the boundary of what this PR closes.
- **A pre-existing test-order dependency was found and left alone.** `intent-classification.test.js`'s "caps the number of logged soft-failure reasoning samples at 10" fails when the serenity suite runs *serially*, and it fails identically on a clean `main`. It passes in isolation, and CI runs the suite in parallel mode, where it does not occur. Not introduced here and out of scope, but worth knowing it exists.

## 7. Test plan

**Verified locally beyond the unit suites — each guard was deliberately broken to confirm it fails for the right reason:**

- Removing `createSubworkspace`'s `@param` tags makes the short call against it legal again, and the now-unused directive fails the gate.
- Widening the fixture's own parameter back to `@param {object} transport` makes *all four* assertions legal, and all four directives fail the gate.
- Removing `noImplicitAny` from the strict config makes the unknown-member assertion legal, and that directive fails the gate.
- Dropping either fixture from its tier's `include`, or deleting it outright, fails `test/types/fixtures-are-wired.test.js`.

The first two demonstrate the two halves are both load-bearing, which is the claim this PR rests on. The third demonstrates the strict tier is doing something the base tier cannot. The fourth demonstrates the guards cannot be removed without a red test. Restoring each returned the gate to green.

The gate was also confirmed to reject the defect classes it exists for: a short call reports an arity error, a malformed allocation object is rejected on shape, and a numeric workspace id is rejected on type.

**Per-environment verification:** none required. The one behavioural change — the sub-workspace market delete skipping an id-less project — is reachable only when Semrush's project listing breaks its own contract, which no environment can be made to do on demand. It is covered by a unit test asserting the `204`, the warning, and that neither the upstream delete nor the tombstone runs. The three behaviour-identical touches are covered by the existing serenity suites.

**For reviewers:** `npm run type-check` now runs two tiers; `npm run type-check:base` and `npm run type-check:strict` run them separately.


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - Well-designed type-safety infrastructure with no runtime regressions.
**Complexity: HIGH** - medium diff; dependency signal from package.json script changes.
_Changes: Exports a SerenityTransport typedef, annotates all 54 transport parameters with it, adds a strict tier (tsconfig.strict.json), and pins the contract with a type-level regression guard (22 files)._
Note: Recommend a human read before merge - this change amends an architectural decision record (docs/decisions/005-opt-in-type-checking.md). The bot review is a complement to, not a replacement for, a human read here.
Note: CI checks are currently failing (deploy-stage, semantic-release, branch-deploy) - verify these are unrelated before merge.

<details>

- suggestion: Add `test/types/serenity-transport.types.js` to `tsconfig.strict.json`'s include list so TS2339 (unknown member) assertions can be added to the regression guard without a future config change - `tsconfig.strict.json:14`
- suggestion: Consider adding a test case for `{ SEMRUSH_USERS_BASE_URL: '' }` confirming the projects-URL fallback, since the `usersBaseUrl` function was refactored for type narrowing - `src/support/serenity/rest-transport.js:232`
- nit: The 4-line comment on `.filter((id) => id !== undefined)` is disproportionate for a standard TS narrowing idiom; a single line suffices - `src/support/serenity/handlers/markets.js:1134`
- suggestion: Document the strict-tier merge-back trigger (when the strict file list equals the base include set, fold into one config) in package.json or nearby, so future maintainers know the two-tier split has a planned end-of-life - `package.json:19`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 31s | <b>Cost:</b> $6.88 | <b>Commit:</b> <code>2f103e1e3c0351222934535172315ce4034250e7</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed all four non-blocking items in `596a927`.

- **suggestion: add the type fixture to `tsconfig.strict.json`'s include so TS2339 assertions can be added** — done, but not as literally stated, because that form does not work: the fixture is also in the base config's include, and a TS2339 assertion is *suppressed* there, so the directive reads as unused and fails the base tier (TS2578 — verified). Implemented instead by splitting the fixtures per tier: `test/types/base/` is checked only by the base config, `test/types/strict/` only by the strict one. That made room for the guard the strict tier was actually missing — it now asserts an unknown member IS reported. Before this, deleting `noImplicitAny` from the strict config was a silent no-op; now it fails the build (verified by removing it).

- **suggestion: test `SEMRUSH_USERS_BASE_URL: ''` confirming the projects-URL fallback** — done. Good catch: an empty string is precisely where the bound-then-guard rewrite short-circuits before reaching `hasText`, so it is the edge worth pinning.

- **nit: the comment on `.filter((id) => id !== undefined)` is disproportionate** — agreed, cut from four lines to one. The invariant it was explaining (map values are non-empty by construction) is not what the reader needs there; that only the explicit form narrows is.

- **suggestion: document the strict tier's merge-back trigger** — done, in `tsconfig.strict.json` next to the file list rather than in `package.json`, so it sits where someone adding a file will actually read it. It records the end state (fold the two configs into one when the list covers the base include set) and points at ADR-005's ratchet path for the remaining cost and ordering.

On the CI note in the review — `deploy-stage`, `semantic-release` and `branch-deploy` were **cancelled, not failed**: two runs raced on the same SHA (push-triggered and PR-open-triggered) and the concurrency group cancelled the `pull_request` one. `gh pr checks` renders cancelled as `fail`. The authoritative push run completed successfully, with `codecov/patch` passing. Unrelated to this diff.

Agreed on the recommendation for a human read — this amends ADR-005, and the judgement worth checking is the scoping one: the strict tier covers `rest-transport.js` only, leaving the same `@param {object}` blind spot reachable elsewhere. ADR-005 now records the measured cost of closing the rest.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - Clean fix commit addressing all prior suggestions; one stale doc path to fix at your convenience.
**Complexity: HIGH** - large diff; dependency signal.
_Changes: Addresses all 4 non-blocking suggestions from the prior review - splits type fixtures per tier, adds empty-URL fallback test, trims a comment, and documents the strict tier merge-back trigger (7 files)._
Note: Recommend a human read before merge - this change amends an architectural decision record (docs/decisions/005-opt-in-type-checking.md). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: Two docs reference the pre-split path `test/types/serenity-transport.types.js` which no longer exists (now at `test/types/base/serenity-transport.types.js`) - `src/support/serenity/CLAUDE.md:46` and `docs/decisions/005-opt-in-type-checking.md:229`

</details>

- Strict-tier type fixture now included in tsconfig.strict.json (implemented as per-directory split rather than literal include - correctly avoids TS2578 unused-directive conflict)
- Empty SEMRUSH_USERS_BASE_URL fallback test added
- Disproportionate 4-line comment trimmed to 1 line
- Merge-back trigger documented in tsconfig.strict.json header comment

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 37s | <b>Cost:</b> $3.66 | <b>Commit:</b> <code>596a9270fcd565292c7eb00eca63aacd0f10747e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Round 2 — the one non-blocking item is fixed in `2b52531`.

- **Stale fixture paths** — valid, both confirmed. `docs/decisions/005-opt-in-type-checking.md:229` and `src/support/serenity/CLAUDE.md:46` still named the pre-split `test/types/serenity-transport.types.js`. Both now point at `test/types/base/serenity-transport.types.js`, and the ADR additionally names the strict-tier counterpart and says *why* the two cannot share a directory — a `TS2339` directive is unused, and therefore itself an error, under the base tier. That reason is the part a reader needs in order not to "helpfully" merge them again.

On the note about a human read before merge: agreed, and it is the right call. This PR amends an ADR, and the judgement it encodes — that the strict tier starts at `rest-transport.js` alone rather than covering the whole checked set — is a sequencing decision, not a fact the diff can justify on its own.

One correction to the record, since it changes how the remaining work is sized. The ADR's 745-error table and the residual `@param {object}` annotations are **not** two separate backlogs. Bare `object` has no members, so under `noImplicitAny` a member access on an `{object}`-annotated value is a `TS2339` — verified on a throwaway file: it reports `Property 'noSuchMember' does not exist on type 'object'` under the strict tier and reports nothing at all under the base tier. Closing the 745 therefore *is* closing the `{object}` blind spot across the checked set; there is no second pass hiding behind it.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - Prior nit resolved; doc paths now point at the correct per-tier fixture locations.
**Complexity: HIGH** - large diff; dependency signal + ADR change.
_Changes: Fixes two stale references to `test/types/serenity-transport.types.js` in the ADR and the serenity CLAUDE.md, updating them to the post-split path `test/types/base/serenity-transport.types.js` and documenting the strict-tier counterpart (2 files)._

- Stale doc path `test/types/serenity-transport.types.js` in `docs/decisions/005-opt-in-type-checking.md` and `src/support/serenity/CLAUDE.md` now corrected to `test/types/base/serenity-transport.types.js`, with the strict-tier path documented alongside it.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 8m 31s | <b>Cost:</b> $1.66 | <b>Commit:</b> <code>2b5253126ead1af5469a43cc572885c88a2a5b29</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
`4064078` closes a gap this PR had left open, found while measuring the remaining ratchet work rather than by review.

Annotating the `@param {object} transport` parameters covered only the parameters that *had* an annotation. **Twenty-three serenity functions took the transport with no `@param` tag at all** — `handleCreateMarket`, `pollUntilCreated`, `resolveLanguageId`, `listMarkets`, `republishBestEffort` and others. An undocumented parameter is implicitly `any` for exactly the same reason an `{object}` one is, so every transport call inside those functions stayed unchecked. They were invisible to a search for `{object}`, which is why the first pass missed them: the pattern being fixed was the annotation, and these had none.

`handleCreateMarket` is the pointed case — it is the same *kind* of function as the one whose dropped `resources` argument motivated this issue.

All 23 now name `SerenityTransport`, and the strict tier confirms none remain. Typing them surfaced one contract mismatch: `handleCreateMarket` builds the `createProject` body as a plain object literal, so `type: 'ai'` inferred as `string` and no longer matched the enum the generated contract declares. The body is now typed from `ProjectCreateBody` where it is built, which also brings `location_id` and `language_id` under the contract. Behaviour is unchanged — the value sent was always `'ai'`.

Verification worth stating plainly, since three serenity specs fail here: `async-intent-classification`, `async-job-runner` and `classify-prompts-job` time out at 2000 ms under `--parallel`. They fail **identically on a clean tree** (checked by stashing this change and re-running the same command: 1240 passing / 3 failing either way) and all pass in isolation. Pre-existing parallel-mode flakiness, unrelated to this PR, and noted here only so the next person does not re-diagnose it.

Two figures in the description were wrong and are corrected there: the residual annotations are 151 `@param {object}` plus 42 `@param {any}`, not "roughly 136".

rainer-friederich:
Re-requesting: the last review landed mid-push and analysed the wrong commit.

GitHub attributes review `4799307653` to `4064078`, but the review's own footer records the commit it actually read as `2b52531`, and its summary describes only that commit's two doc files. The substantive change on this branch — 23 transport parameters that carried no JSDoc at all, and the `ProjectCreateBody` typing that fell out of it — has not been reviewed by anything yet.

Flagging it rather than treating the approval as covering the branch, since the API's `commit_id` and the analysed commit disagree here and only the footer is truthful about which one was read.

CI has run against `4064078` (`head_sha` on run 30375654161) and is green, including `codecov/patch`.

rainer-friederich:
MysticatBot's round-3 review approved with nothing outstanding, but its pipeline then declined to re-review (it dedups on GitHub's attributed `commit_id`, which differed from the commit its own footer says it read). The review for the rounds after that was therefore produced locally with `/pr-review` — five reviewers: staff-engineer, architect, project-conventions, security, qa. Nine findings, every one verified against the code before being accepted. Resolutions below, plus the `main` merge.

**Important — all fixed**

- Six transport parameters were still `{object}` / `{any}`, while the docs claimed every one was typed. Typed; the claim is now true. The original sweep missed them because it used `grep`, and `handlers/markets-subworkspace.js` contains a NUL byte (a composite map-key separator, pre-existing on `main`) that makes `grep` treat the file as binary and return nothing. Every subsequent sweep was run through the compiler or Python instead.
- A tsconfig `include` lists the program's **roots**, not the set of files that get checked — `noImplicitAny` is program-wide, so every `// @ts-check` file reachable by import is reported too. Raised independently by two reviewers. Both `tsconfig.strict.json` and ADR-005 said otherwise; both corrected, and the claim was verified with a probe.
- Following from that, the ADR's recommended ratchet order was the most expensive one available: `src/controllers/brands.js` owns 151 errors but roots a program reporting 502 across 23 files, while a leaf costs 1. The ADR now says leaf-first and gives the recipe for measuring a candidate before adding it.
- Nothing pinned the type fixtures themselves — they are enforced by being compiled, so dropping either from its tier's `include` or deleting it left `type-check` green. `test/types/fixtures-are-wired.test.js` now asserts both are wired and still carry assertions; verified by performing both deletions.
- The sub-workspace market delete used a nullable `project.id` unguarded for both the upstream `DELETE` and the mapping-row tombstone. Guarded, warned, and unit-tested.

**Minor — all fixed**

- A redundant `hasText` on the project id was worse than redundant: it made the guard fire only for truthy non-strings, so a numeric `id` would have returned `204` without deleting anything. Dropped.
- An id-less project reported a silent success; it now names the contract break in a warning.
- `ProjectCreateBody` was reached through an inline `import(...)` in one file and an alias in another; consistent now.
- ADR-005 conflated `context` with the `log` shortcut, left a dangling "At that point", and understated the runtime touches as one when there are four. All corrected.
- A fixture comment named TS2345 where the compiler reports TS2559.
- `usersBaseUrl` narrows with `typeof` rather than leaning on `hasText`, which is not a TS type guard. Behaviour-identical, and proved so by differential probe across the unset / empty / whitespace / non-string / valid-URL inputs.
- A `subWorkspaceId` local read as a second, different id; renamed.
- The serenity `CLAUDE.md` rule on `@param` ordering contradicted the code it governs. Rewritten to say what a partial signature actually costs.

**Declined (1)**

- Commit `4064078`'s subject overstates "carried no JSDoc at all" — most of those functions had docblocks missing only the `@param` tag. Accurate finding, but force-pushing is prohibited org-wide and the commit body states it correctly. The PR description now uses the precise wording.

**Merged `main`**

`main` had advanced past this branch, including https://github.com/adobe/spacecat-api-service/pull/2923 which reduced `createSubworkspace` to two parameters. That silently invalidated two of the four base-fixture assertions — the arity directive would have gone unused and the clean call would have become an arity error. Arity is now pinned by under-calling, which is the assertion that actually proves the `@param` tags restore arity checking, and the contract-derived body-shape assertion moved to `transferWorkspaceResources`. Verified by stripping the tags again and confirming the build fails.

The measured `noImplicitAny` figures in ADR-005 and `tsconfig.strict.json` were re-measured against the merged tree rather than carried over.

**Out of scope, flagged not fixed**

`src/support/elements/elements-service.js` takes `@param {object} transport` — a second transport surface with the identical weakness, outside this gate's `include` set.

**Known remaining hole, disclosed rather than closed**

Transport *responses* still resolve to `any`, so property reads off them are unchecked — precisely the class that produced the nullable-`project.id` defect above. Closing it is ratchet work, not this PR.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - Well-structured type-safety ratchet; runtime changes are minimal, defensive, and test-covered.
**Complexity: HIGH** - large diff (28 files); dependency signal from package.json script changes + ADR amendment.
_Changes: Exports a SerenityTransport typedef, annotates all 54 transport parameters, adds a strict tier (tsconfig.strict.json with noImplicitAny), type-level regression guards, and handles nullable project.id in delete (28 files)._
Note: Recommend a human read before merge - this change amends an architectural decision record (docs/decisions/005-opt-in-type-checking.md). The bot review is a complement to, not a replacement for, a human read here.

<details>

- suggestion: The JSONC comment-strip in `fixtures-are-wired.test.js` (line 33) uses a naive line filter (`.filter(line => !line.trim().startsWith('//'))`) that would silently misparse inline `//` comments or block comments if someone adds them to the tsconfigs. Low-probability given these are controlled files, but a JSONC-aware parser (or at minimum a note that only line-leading comments are handled) would make the meta-test more robust - `test/types/fixtures-are-wired.test.js:33`
- suggestion: The `resolvedWorkspaceId` type assertion (`/** @type {string} */ (workspaceId)`) is correct given `resolveProject` succeeded, but a one-line guard (`if (!workspaceId) return { status: 204, deletedSiteId: null }`) before the cast would make the narrowing structural rather than relying on control-flow inference across function boundaries - `src/support/serenity/handlers/markets-subworkspace.js:953`
- nit: The nullable-id test covers `id: null` but the guard (`!projectId`) also catches `id: ''`. An `id: ''` case would pin the contract more precisely, though the truthiness check is trivially correct for all falsy values - `test/support/serenity/handlers/markets-subworkspace.test.js:1052`

</details>

- Strict-tier type fixture included in tsconfig.strict.json (per-directory split avoids TS2578 conflict)
- Empty SEMRUSH_USERS_BASE_URL fallback test added
- Disproportionate 4-line comment trimmed to 1 line
- Merge-back trigger documented in tsconfig.strict.json header
- Stale doc paths corrected to per-tier locations

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 18s | <b>Cost:</b> $7.51 | <b>Commit:</b> <code>64de3cc984d9eb155d0ecb230e502b0ba41d3c2e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the round-4 review in `e83ecc08d`. All three were non-blocking; one is fixed, two are declined, and chasing the third turned up an error of mine worth calling out.

**suggestion — naive JSONC strip in `fixtures-are-wired.test.js`: fixed, with a correction.**

The helper's contract is now stated in its docblock. The finding's premise is wrong, though, and the note says so rather than repeating it: the strip cannot *silently* misparse. A trailing `// …` or a `/* … */` block survives the whole-line filter and reaches `JSON.parse`, which throws — verified in both Node and Python. The failure mode is a red test, not a wrong answer, so a JSONC parser buys nothing here. Documenting the limit is still worth it; hardening against it is not.

**suggestion — structural guard instead of the `resolvedWorkspaceId` cast: declined.**

The proposed `if (!workspaceId) return { status: 204, deletedSiteId: null }` is unreachable. Reaching that line requires `resolveProject` to have returned a matching project, and it only returns non-null when `transport.listProjects(workspaceId)` came back with items for that id — a blank id cannot produce that against the real gateway. Adding it would be dead code, and on this repo that is not free: `codecov/patch` requires 100% on changed lines, so an uncoverable branch turns the check red, and covering it would mean a test for a state the upstream cannot reach.

**nit — add an `id: ''` case alongside `id: null`: declined, but it found something.**

Both are the same falsy branch, so the case adds no coverage. Chasing *why* the field could be absent is what mattered: I checked the generated schema, and `model.ProjectResponse` declares **`id: string`** — required and non-nullable. My guard's comment, its test's comment, and ADR-005 all called the field "contract-nullable". That was wrong, and this commit corrects all three.

The guard itself stands, for a better reason than the one I gave it. `listProjects` is *declared* to return `model.ProjectListResponse`, but a probe assigning the result to `never` reports `any` under both tiers — the degradation is in the vendor facade's own `TransportData` helper, not in this repo's wrapper. So nothing enforces the declaration at any call site, and the live gateway is documented to return bodies the spec forbids. An id-less entry reaches the delete path unchecked regardless of what the schema says.

That also sharpens the boundary of this PR, which the description now states: naming the transport closed the **request** side. The **response** side is still `any`, and this defect is what that costs.

Superseding one line from the previous comment: this defect is not an instance of a nullable field being read unguarded — it is an instance of a *non-nullable* field being unenforceable because the response is untyped.

rainer-friederich:
Validated on `a9d16b14e` — locally in-process, plus CI's own environments. There is no deployable surface to drive here beyond that: no per-PR ephemeral stack exists on this repo, and the one behavioural change is only reachable when Semrush breaks its own contract, which no live environment can be made to do on demand. So the changed code was executed directly rather than through a deployment.

**The behavioural change, driven through the real handler** — `handleDeleteMarketSubworkspace` invoked in-process with a stub transport, asserting the transport calls and the mapping-row lookups actually made:

- normal project, `id: 'p-1'` → `deleteProject('p-1')` called, mapping-row lookup made, no warning. Unchanged path still works.
- matched project, `id: null` → `204`, `deleteProject` **not** called, **no** mapping-row lookup, one warning naming the contract break. This is the guard, doing what it claims.
- matched project, `id: ''` → identical to `id: null`. This is the case the review suggested testing separately; executing it confirms it is the same branch, which is why a second test case was declined rather than added.
- matched project, `id: 12345` (numeric) → `deleteProject(12345)` **is** called. This is the case the removed redundant `hasText` would have swallowed: it would have returned `204` without deleting anything upstream. Confirms that removal fixed a real defect rather than tidying a no-op.
- no matching project → `204`, nothing called, no warning.

**The two behaviour-identical touches, proved by differential execution** rather than asserted:

- `usersBaseUrl` old guard vs the shipped `typeof`-narrowed one, 19 inputs (unset, absent key, `undefined`, `null`, `''`, whitespace, bare host, valid URL, `0`, `1`, `false`, `true`, `{}`, `[]`, `['x']`, `NaN`, `Symbol`, function, boxed `String`): **0 divergences**.
- `.filter(Boolean)` vs `.filter(id => id !== undefined)` on `Map.get()` output: 1 divergence, on an empty-string assignment id — `Boolean` drops it, `!== undefined` keeps it. It is unreachable: values enter that map only past `hasText(it.id)`, and the keys are the map's own, so `get` never misses. Worth stating precisely rather than claiming an unconditional equivalence — the change is identical *given the insertion guard*, not in general.

**CI environments**, on this commit: `ci / it-postgres` green twice, which is the mock-backed serenity suite driving the Counterfact vendor mock over real HTTPS — the contract gate this ADR names. `ci / build` green twice, covering the Lambda bundle (`hedy --test-bundle` invokes the bundled handler), which is this repo's documented "green tests, broken production" seam and matters here because the PR touches `package.json` scripts.

**Not run, and why:** the local `it-postgres` suite could not start — its `mysticat-data-service` image needs ECR auth and the AWS session is expired (`klam login` is interactive). CI ran the same suite on this commit instead. The local `npm run build` also could not run, for missing VPC env values; CI's `ci / build` covers it.

**Observation, out of scope and not changed:** `tombstoneMappingRow` guards its own input with `hasText`, so a numeric project id would delete upstream without tombstoning the mapping row. Pre-existing in a helper this PR does not touch, and unreachable while the contract declares `id: string`.

PR #2934, opened 2026-07-29
fix(test): create sinon fakes per test, not at module scope

### LINKED ISSUE #2932: Serial test runs fail in intent-classification.test.js: sinon.restore() in other specs disables its afterEach reset
## Summary

`test/support/serenity/intent-classification.test.js` passes under `npm test` and fails whenever the serenity suite is run **serially**, which is what a developer does when scoping a run to the files they touched. The failure is not in the file that fails, and not in the change that triggered it — it is cross-file state leaking through sinon's default sandbox.

```
1) intent-classification.js — observability (serenity-docs#32)
     caps the number of logged soft-failure reasoning samples at 10:
   AssertionError: expected [ …(2) ] to have a length of 10 but got 2
```

## Reproduce

Two files are enough. Neither is broken on its own.

```bash
npx mocha test/support/serenity/mapping-rows.test.js \
          test/support/serenity/intent-classification.test.js
# 45 passing, 1 failing

npx mocha test/support/serenity/intent-classification.test.js
# 22 passing        <- passes alone
```

## Root cause

Three facts compose.

- `intent-classification.test.js` creates its logger stubs **at module scope, on sinon's default sandbox**:

  ```js
  const log = {
    info: sinon.stub(), warn: sinon.stub(), error: sinon.stub(), debug: sinon.stub(),
  };
  ```

  and clears them with `afterEach(() => { sinon.reset(); })`.

- **`sinon.restore()` empties the default sandbox's fake collection.** Any stub registered before that call is no longer tracked, so every later `sinon.reset()` is a silent no-op for it. Demonstrated standalone:

  ```js
  const log = { info: sinon.stub() };
  log.info('a'); sinon.reset();
  // callCount = 0   (reset works)
  log.info('b'); sinon.restore(); log.info('c'); sinon.reset();
  // callCount = 2   (reset is now a no-op — history survives)
  ```

- **61 test files in this repo call `sinon.restore()`.** When any of them runs in the same process before `intent-classification.test.js`, that file's `afterEach` stops clearing `log`, and call history accumulates across its own tests.

The assertion then reads the wrong call. It takes the **first** matching entry:

```js
const softLog = log.info.getCalls().find((c) => /soft failures/.test(c.args[0]));
expect(softLog.args[1].samples).to.have.length(10);
```

With history retained, the first match is the soft-failure log from the earlier test in the same file — the one that produces `low_confidence: 2`, hence `samples.length === 2`. The failing test's own log, with 10 samples, is further down the list and never inspected.

## Why CI is green

`npm test` runs `mocha --parallel`, which executes each spec file in its own worker process. Every file therefore gets a fresh module registry and a fresh default sandbox, and no `sinon.restore()` from another file can reach it. The bug is invisible to CI by construction and only appears in the serial runs developers actually use — including the scoped single-directory invocations documented in `CLAUDE.md`.

## Blast radius

A scan of the non-IT suites found **exactly one** file with the vulnerable shape today (module-scope sinon fakes plus reliance on `sinon.reset()`): `intent-classification.test.js`. So this is one failing test, not a systemic outage.

The hazard is latent rather than contained, though. The pattern is easy to write, reads as correct, and is silently disarmed by unrelated files. There is nothing in the suite that would catch a second instance.

An empirical correlation confirms the mechanism — pairing `intent-classification.test.js` with each of the other 35 serenity spec files, all 11 that fail call `sinon.restore()` and none of the 25 that pass do. No false positives, no false negatives.

<details>
<summary>The 11 files that trigger it</summary>

- `dynamic-allocation-active.test.js`
- `dynamic-allocation-fronting.test.js`
- `handlers/markets-subworkspace.test.js`
- `handlers/markets.test.js`
- `handlers/prompts-subworkspace.test.js`
- `handlers/tags.test.js`
- `mapping-rows.test.js`
- `serenity-active.test.js`
- `site-linkage.test.js`
- `subworkspace-projects.test.js`
- `workspace-lifecycle.test.js`

</details>

## Suggested fix

Give the file its own sandbox so it is immune to what other files do to the default one. This is also the pattern `CLAUDE.md` documents as the standard test shape.

```js
const sandbox = sinon.createSandbox();
const log = {
  info: sandbox.stub(), warn: sandbox.stub(), error: sandbox.stub(), debug: sandbox.stub(),
};
// in both describe blocks:
afterEach(() => {
  sandbox.reset();
  sinon.reset();
});
```

Verified: with that change the failing pair is `46 passing, 0 failing`, and the full serial serenity glob is `1243 passing, 0 failing`.

Two alternatives were considered and are worse. Reading the **last** matching call instead of the first makes the assertion pass without fixing the leak, so the next test written this way fails the same way. Resetting each stub by hand in `afterEach` works but re-states the sandbox's job at every call site.

Worth considering alongside the fix: a lint rule or a review convention against creating fakes on the default sandbox at module scope, since the failure mode gives no signal at the point the mistake is made.

## Not part of this issue

Three async serenity specs (`async-intent-classification`, `async-job-runner`, `classify-prompts-job`) were separately observed timing out under `--parallel` during unrelated work. That is a different failure mode — nondeterministic timeouts, not state leakage — and is not re-verified against the current tree here. Flagged only so the two are not conflated.

## Context

Found while working on https://github.com/adobe/spacecat-api-service/pull/2928 . Not introduced by it: the failure reproduces on a clean checkout with that branch's changes reverted, and neither `intent-classification.js` nor its spec is touched there.

rainer-friederich:
## 1. Abstract
Test-suite fix: seven spec files built their sinon fakes at module scope, where the fakes are shared across every test in the file. They now build them per test on a dedicated sandbox, and a lint rule keeps the pattern from returning.

## 2. Reasoning
`test/support/serenity/intent-classification.test.js` passed under `npm test` and failed whenever the serenity suite ran serially — which is what a developer does when scoping a run to the files they touched.

Three facts compose. A sinon fake created at module scope is registered on sinon's default sandbox. `sinon.restore()` empties that sandbox's fake collection, so any fake registered before that call stops being tracked and every later `sinon.reset()` silently becomes a no-op for it. Sixty-one spec files in this suite call `sinon.restore()`. When any of them runs first in the same process, the failing file's `afterEach` quietly stops clearing its logger stubs and call history accumulates across its own tests.

The assertion then reads the wrong call: it scans for the *first* log matching a soft-failure pattern, and with history retained that is an earlier test's log carrying two samples rather than its own carrying ten.

CI never saw this. `npm test` runs `mocha --parallel`, giving each spec file its own process and its own default sandbox, so no other file's `sinon.restore()` can reach it. The bug is invisible to CI by construction and appears only in serial runs.

No production source is touched; this changes test setup, a lint rule, and the contributor documentation.

**Fake lifetime.** Before, the affected files created stubs once at module load and relied on `afterEach` to clear them — a contract that another file could silently void. After, each file creates its fakes inside `beforeEach` on a sandbox from `sinon.createSandbox()` and restores that sandbox in `afterEach`. Nothing survives a test boundary, so no other file's cleanup can reach them.

**Blast radius was wider than the issue reported.** The issue found one file by scanning for module-scope fakes *plus* reliance on `sinon.reset()`. Scanning for module-load fakes generally found six more carrying the same latent shape — including `async-intent-classification.test.js` with a byte-identical logger-stub construction, and `llmo-sheet-write.test.js`, which calls `sinon.restore()` in its own `afterEach` while holding module-scope stubs. None were failing yet; all could fail the same way once assertions or file ordering shifted.

Two of those files needed more than a mechanical move:

- `llmo-sheet-write.test.js` built a context object in a `describe` body that closed over the logger. Mocha evaluates suite callbacks during collection, before any hook runs, so that object captured the fake before it existed. It is now assembled per test.
- `project.test.js` shared two model instances across tests that individual specs assign onto. They are now built per test from factory functions, so one test's assignment cannot reach the next.

`run-audit.test.js` keeps its mock object's identity stable, because esmock injects that exact reference into the module under test; only the fake on it is replaced per test.

**Guard rail.** A `no-restricted-syntax` rule in the test override rejects sinon fakes evaluated at module load, covering both the `sinon.stub()` and the bare `stub()` call shape this suite uses via a named import. The rule deliberately does not fire on fakes built inside factory functions, which are created per call and are not the hazard.

**Documentation.** `CLAUDE.md` now states the sandbox rule and the mechanism behind it in the Standard Test Pattern section, and the troubleshooting section gains the symptom that identifies this class: a test that passes alone and fails in a serial run.

## 4. Required information
- Jira / issue: https://github.com/adobe/spacecat-api-service/issues/2932

Closes #2932

Verification done in-session, beyond running the suites:

- **The mechanism was reproduced in isolation** rather than taken from the issue. A standalone script confirmed that `sinon.reset()` clears a fake normally, and that after an intervening `sinon.restore()` the same `reset()` leaves call history intact.
- **The same script showed the issue's suggested fix keeps the trap one level in.** A dedicated sandbox is immune to *other* files' `sinon.restore()`, but calling `sandbox.restore()` on that sandbox disarms its own later `sandbox.reset()` in exactly the same way. The per-test sandbox used here avoids this, since the sandbox is never reused after being restored.
- **The lint selector was measured, not assumed.** An early, broader form flagged sites inside module-scope factory functions — fakes created per call, not the hazard. The selector was narrowed and re-measured against the whole test tree until every remaining hit was a genuine module-load fake, then widened to the bare-`stub()` call shape after finding a file that imports it that way.
- **A per-file before/after comparison** was taken by stashing the change, so each converted file's result could be checked against its own pre-change state rather than against expectation.
- **One scoping decision worth flagging.** A fake declared directly in a `describe` body carries the same hazard, since suite callbacks also run once at collection. Extending the rule to cover that form was implemented and measured: it flags several hundred existing sites across the suite. That form is long-established style here, so enforcing it is a migration in its own right and is out of scope for this fix; the rule covers the module-scope form that actually broke, and the limitation is recorded in both the rule's comment and `CLAUDE.md` rather than left silent.

Unrelated pre-existing condition, noted because it shows up in a full local run: `npm test` reports one uncaught error from a missing `@aws-sdk/client-dynamodb`, an undeclared electrodb peer dependency absent from local `node_modules`. It is present identically before this change and is untouched here.

## 7. Test plan
This change has no runtime surface — it alters when test doubles are constructed, plus a lint rule and documentation. There is nothing to exercise on a deployed environment, and no eph/dev/stage/prod verification applies.

To verify locally:

- Reproduce the original failure on `main` by running the two files together serially, mapping-rows first, then intent-classification. The soft-failure sample assertion fails there and passes on this branch.
- Run the serenity suite serially over its full glob. This is the run the issue describes and the one CI's parallel mode cannot represent.
- Run each of the seven changed spec files on its own, and then together, to confirm neither isolation nor co-execution changes their results.
- Confirm the guard rail bites: add a module-scope `sinon.stub()` to any file under `test/` and check that lint rejects it, then move it into `beforeEach` and check that it passes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -424,6 +424,14 @@ describe('Sites Controller', () => {
 });
 ```
 
+**Build every fake inside `beforeEach`, on a per-test sandbox** — never at module scope. A fake created at module scope is registered on sinon's *default* sandbox, and `sinon.restore()` in **any** other spec file empties that sandbox's fake collection. From then on `sinon.reset()` silently stops clearing the fake and its call history accumulates across the tests in your file, so an assertion like `getCalls().find(...)` starts reading an earlier test's call. Cleanup in your own `afterEach` does not protect you — the fake is created in the wrong place, not cleaned up in the wrong place.
+
+`mocha --parallel` (what `npm test` and CI run) gives each spec file its own process, so no other file's `sinon.restore()` can reach it and the whole failure mode is invisible. It appears only in serial runs — including the scoped single-file and single-directory runs under **Single Test Execution** above.
+
+A fake declared directly in a `describe` body is shared the same way, because mocha evaluates suite callbacks during collection, before any test runs. Assertions that scan call history — `getCalls().find(...)`, `calledOnce` — are the ones that break under sharing.
+
+A `no-restricted-syntax` rule in `eslint.config.js` fails the build on `sinon.stub/spy/fake/mock/createStubInstance/useFakeTimers()` evaluated at module load in `test/**/*.js`, in both the `sinon.stub()` and bare `stub()` call shapes. It catches the module-scope form only. The `describe`-body form is long-established style in this suite and is not enforced; nor can a syntactic rule see a fake built by a helper that is itself *called* at module load. Build fakes in `beforeEach` and none of these distinctions arise.
+
 **Tools**:
 - **Mocha**: Test runner
 - **Chai**: Assertions (`expect`, `chai-as-promised`)
@@ -634,6 +642,8 @@ For this repo:
 - Verify stubs are restored in `afterEach`
 - Use `esmock` for ES module mocking
 - Check test fixtures match current schema
+
+**A test that passes alone and fails in a serial run** (or passes under `npm test` and fails under `npx mocha <dir>/*.test.js`): look for a fake built at module scope rather than in `beforeEach`. Its call history survives across the tests in the file once another spec calls `sinon.restore()` — see the sandbox rules under **Standard Test Pattern**. The reverse pairing, green in serial and red in parallel, is a different problem: cross-file order dependence or a shared external resource.
 
 ### Debugging
 
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -13,6 +13,33 @@
 import { defineConfig, globalIgnores } from '@eslint/config-helpers'
 import {recommended, source, test} from '@adobe/eslint-config-helix';
 
+// A sinon fake created at module scope lands on sinon's default sandbox and
+// outlives every test in the file. Any other spec calling `sinon.restore()`
+// drops it from that sandbox's collection, after which `sinon.reset()` no
+// longer clears it and call history accumulates across tests — a failure that
+// surfaces only in serial runs, never under `mocha --parallel` (issue 2932).
+const FAKE_FACTORY = '/^(stub|spy|fake|mock|createStubInstance|useFakeTimers)$/';
+// Both call shapes the repo uses: `sinon.stub()` and, via
+// `import sinon, { stub } from 'sinon'`, a bare `stub()`.
+const SINON_FAKE = `CallExpression[callee.object.name='sinon'][callee.property.name=${FAKE_FACTORY}]`;
+const BARE_FAKE = `CallExpression[callee.type='Identifier'][callee.name=${FAKE_FACTORY}]`;
+// `:not(:function *)` narrows these to fakes evaluated at module load. A fake
+// built inside a factory function is created per call and is not the hazard,
+// so it must not be flagged. This is a lexical test, not an execution-timing
+// one, so it cannot see a fake built by a helper that is itself called at
+// module load — build fakes in `beforeEach` and the distinction never arises.
+//
+// A fake declared directly in a `describe` body shares the hazard — mocha
+// evaluates suite callbacks during collection, so it too is created once for
+// the whole suite. That form is not covered here: it is long-established style
+// in this suite (hundreds of call sites), so enforcing it is a migration of its
+// own rather than something this rule can turn on. What is covered is the shape
+// that actually broke, and the one a new file is most likely to reach for.
+const SINON_SHARED_FAKE = `${SINON_FAKE}:not(:function *), ${BARE_FAKE}:not(:function *)`;
+const SINON_SHARED_FAKE_MESSAGE = 'Create sinon fakes per test, inside beforeEach, on a '
+  + 'sandbox from sinon.createSandbox() that afterEach restores. A fake created at module '
+  + 'scope is shared by every test in the file.';
+
 export default defineConfig([
   globalIgnores([
     '.vscode/*',
@@ -43,6 +70,16 @@ export default defineConfig([
     rules: {
       'no-console': 'off',
       'func-names': 'off',
+      'no-restricted-syntax': [
+        'error',
+        'ForInStatement',
+        'LabeledStatement',
+        'WithStatement',
+        {
+          selector: SINON_SHARED_FAKE,
+          message: SINON_SHARED_FAKE_MESSAGE,
+        },
+      ],
     },
   },
 ]);
--- a/test/controllers/llmo/llmo-agentic-traffic.test.js
+++ b/test/controllers/llmo/llmo-agentic-traffic.test.js
@@ -2363,10 +2363,9 @@ describe('llmo-agentic-traffic — rotation (demo sites)', () => {
   let clock;
   beforeEach(() => {
     setUpAuthStubs();
-    clock = sinon.useFakeTimers({ now: Date.UTC(2026, 6, 6), toFake: ['Date'] });
+    clock = sandbox.useFakeTimers({ now: Date.UTC(2026, 6, 6), toFake: ['Date'] });
   });
   afterEach(() => {
-    clock.restore();
     sandbox.restore();
   });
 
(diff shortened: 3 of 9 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-scoped fix for a subtle serial-run test failure, with a lint guard to prevent recurrence.
**Complexity: MEDIUM** - medium diff; test infrastructure across 8 files with lint rule and documentation.
_Changes: Moves sinon fakes from module scope into per-test sandboxes in seven test files, adds an ESLint no-restricted-syntax rule to prevent the pattern from returning, and documents the mechanism in CLAUDE.md (8 files)._
Note: Recommend a human read before merge - this change amends root CLAUDE.md (project conventions). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: `sinon.useFakeTimers()` in the rotation describe blocks (`llmo-agentic-traffic.test.js:2365`, `llmo-referral-traffic.test.js:1468`) uses the global sinon default sandbox rather than the per-test `sandbox`. Not an active vulnerability (the clock is created in `beforeEach`, so the cross-file race window does not apply), but inconsistent with the PR's goal of consolidating all fakes onto the per-test sandbox. If someone later removes `clock.restore()` thinking `sandbox.restore()` covers it, the timer would leak. Fix: `clock = sandbox.useFakeTimers(...)`.
- nit: `llmo-sheet-write.test.js` retains `sinon.restore()` alongside `sandbox.restore()` in afterEach for in-test stubs that still use the default sandbox. The file is half-converted; a follow-up to migrate those inline `sinon.stub(obj, method)` calls to `sandbox.stub(obj, method)` would eliminate the dual-restore.
- suggestion: Helper functions `metricsMock()` and `loadWithClassifier()` in `intent-classification.test.js` close over the module-level `sandbox` variable and will throw a cryptic TypeError if called before `beforeEach` runs. A one-line comment at the declaration noting the precondition would prevent future misuse.
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 5s | <b>Cost:</b> $4.03 | <b>Commit:</b> <code>05e8c3ee3057a0c27cc1784247fe730f13d54756</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the review in `be2f770`.

MysticatBot — all three non-blocking items:

- **nit: `sinon.useFakeTimers()` in the rotation blocks uses the default sandbox** — fixed. Both rotation suites now install their timers via `sandbox.useFakeTimers(...)`, which makes `sandbox.restore()` the single owner of teardown; the explicit `clock.restore()` in those `afterEach` hooks is gone as a result. `llmo-agentic-traffic` keeps a local `clock` because one test re-arms the timer to a different date mid-test; `llmo-referral-traffic` does not, so the variable went away entirely there.
- **nit: `llmo-sheet-write.test.js` keeps `sinon.restore()` alongside `sandbox.restore()`** — declined as a change, documented instead. Those inline stubs are built per test inside `it` bodies, so they are not the shared-fake hazard this PR fixes and the lint rule correctly does not flag them. Migrating roughly thirty call sites in one file would widen the diff well past the issue without changing behaviour. The `afterEach` now carries a comment saying why both calls are there and that the default-sandbox ones were deliberately left.
- **suggestion: note the precondition on `metricsMock()` / `loadWithClassifier()`** — done. The `sandbox` declaration in `intent-classification.test.js` now records that the loader helpers build their fakes on it and are therefore callable only from inside a test or hook.

On the recommendation for a human read before merge: agreed, and worth flagging what to look at. The `CLAUDE.md` change states the sandbox rule, the mechanism behind it, and one deliberate limitation — the lint rule covers fakes at module scope but not fakes declared in a `describe` body, which carry the same hazard because mocha evaluates suite callbacks during collection. Extending the rule to that form was implemented and measured during this work: it flags several hundred existing sites, so it is a migration of its own rather than something this fix can turn on. That trade-off is the main thing worth a second opinion.

PR #2948, opened 2026-07-29
fix(serenity): gate the Semrush error allowance behind the serenity_ui org flag | LLMO-6565

rainer-friederich:
## 1. Abstract

Puts the Semrush-failure allowance on brand edits behind a new org-level `LLMO/serenity_ui` feature flag. An org still inside the Semrush migration keeps today's behaviour (the failure is absorbed); an org pinned to the Serenity UI gets the failure reported.

## 2. Reasoning

`PATCH /v2/orgs/{org}/brands/{brandId}` currently absorbs Semrush failures: the brand row commits, the response comes back `200` with `semrushSyncPending: true`, and the Semrush side quietly diverges. That was introduced to unblock brand edits for migrated orgs whose LLMO users have no Semrush user record at all — Semrush binds the forwarded IMS token to a Semrush user before executing, and JIT provisioning from IMS claims was declined, so for those users Semrush answers 4xx on every call, permanently. Reporting that would tell the customer their edit failed when the row did persist.

That is a migration-window allowance, not a permanent contract. Once an org is fully migrated its users are provisioned in Semrush, and from that point a Semrush failure is a real fault that the customer needs to see rather than expected migration noise to swallow. The `serenity_ui` flag is exactly the marker for "this org is past the migration" — it is what pins the org's UI to Serenity and hides the classic/Serenity switcher — so it is the right switch to end the allowance.

The org-wide flag reader in `src/support/serenity/serenity-active.js` becomes generic over the flag name (its cache key was already composite, built for exactly this) and gains a second predicate for `LLMO/serenity_ui`, sharing the same cache, TTLs and default-OFF failure behaviour as the existing activation flag: a missing row, a `false` row, an absent PostgREST client and a transient read error all resolve to off.

Behaviour delta on the brand-edit path, given `LLMO/serenity` is on (it gates the route entirely, unchanged):

| `LLMO/serenity_ui` | Semrush failure on a brand edit |
| --- | --- |
| off (default, every org today) | absorbed — `200` with `semrushSyncPending: true` |
| on | reported — `401` / `403` / `502` |

Both places that swallow a Semrush failure are re-armed together, off a single answer pinned for the whole request, so an edit can never end up half-absorbed:

- the **pre-write** self-referential-competitor guard, which lists the brand's Semrush projects to build the reserved-domain set. Absorbed, it degrades to a reserved set built from the brand's own URLs alone and the write proceeds. Reported, it rejects the write before the row is touched.
- the **post-commit** re-sync. Absorbed, it returns the committed row with `semrushSyncPending: true`. Reported, it rethrows into the controller's error mapper — `401`/`403` pass through, everything else maps to `502`, and the internal gateway URL stays out of both the body and the `x-error` header on this path as it does on the others.

The row stays committed either way. The flag changes only what the caller is told, never whether the write happened — worth being explicit about, because a reported failure now means "your edit persisted but Semrush does not reflect it", not "nothing happened". The drift breadcrumb (`serenity: brand-edit Semrush re-sync failed after row commit`, warn with `permanent: true` for 401/403, error otherwise) is emitted on both paths, and the competitor guard's own log now carries whether it rejected the write or degraded — without it the outer handler would log only a generic brand-update error with no Semrush context.

The flag is resolved lazily, only once a Semrush call has actually failed. Failures are rare and most brand edits are flat-mode or non-Serenity, so the common path pays no flag read at all.

No behaviour changes for any org today: nothing sets `serenity_ui` yet, and it defaults off.

## 4. Required information

- Jira / issue: LLMO-6565 — https://jira.corp.adobe.com/browse/LLMO-6565
- Design of record: there is no separate spec document for this change — the design was agreed on the Jira ticket and the linked Slack thread. The resulting contract is written down in `docs/serenity.md` (section "Migration-window flag (`LLMO/serenity_ui`)"), which this PR adds: the flag's backend meaning, the two-flag truth table, both re-armed call sites, and the rollout procedure.
- Other: frontend half — https://github.com/adobe/project-elmo-ui/pull/2565 ; the change being gated — https://github.com/adobe/spacecat-api-service/pull/2907

- **project-elmo-ui** — contract. It is the other consumer of this flag and must match the spelling: its open PR introduces the constant as `serenity-ui`, which cannot work (see section 6). It needs to read `serenity_ui`.
- **mysticat-data-service** — contract, unchanged. Owns the `feature_flags` table and the CHECK constraint that fixes the flag-name spelling.

Queried mysticat-prod directly to settle the flag name, because LLMO-6565 specifies `serenity-ui` and states the backend already emits it. Neither holds.

The table rejects the hyphenated name outright. `pg_constraint` for `feature_flags_flag_name_valid` in prod:

`CHECK ((char_length(flag_name) <= 255) AND (flag_name <> '') AND (flag_name ~ '^[a-z][a-z0-9_]*$'))`

Snake_case only — a row named `serenity-ui` can never be inserted, and `isValidFeatureFlagName` in this repo mirrors the same regex so the admin write endpoint rejects it before the database ever sees it. A client matching on the hyphenated name would read `false` forever with no error surfaced anywhere, which reads as "the flag is simply off".

Nor does the flag exist yet under any spelling. Grouping every row in prod `feature_flags` returns exactly three names, all under product LLMO: `brandalf` (1096 orgs, 1092 on), `brandalf_migration` (27 orgs, 1 on), `serenity` (83 orgs, 83 on). Enabling this for an org therefore needs a flag row created first — it is not already being emitted.

This PR uses `serenity_ui` on both counts.

## 7. Test plan

Not verified against a live environment this session — the reported path needs an org carrying a flag that does not exist in any environment yet, and a Semrush call that actually fails for the calling user. Both halves were exercised against the controller instead: the absorbed path (unchanged), the reported path for an authorization failure, a non-auth upstream failure and a failed token exchange, the pre-write guard rejecting rather than degrading, that the flag is not read when nothing fails, and that one answer is pinned when both call sites fail.

The flag name itself is covered against a real database rather than a mock, because that is where it can actually go wrong: nothing on the read path validates a flag name, so a wrong spelling resolves to `false` forever and looks exactly like the flag being off. The integration test writes `serenity_ui` through the admin endpoint, reads it back through the same storage helper the controller's gate uses, and asserts the hyphenated spelling is refused. An endpoint-level integration test for the 401/502 responses is not reachable today — the brands integration suite uses flat-mode brands only, and the Semrush vendor mock has no failure-injection control, so there is no way to make the re-sync fail from inside the harness.

Verifying on dev, once deployed:

1. Pick a Serenity-active org and a sub-workspace brand under it. Confirm the current behaviour first: edit the brand's URLs as a user who is not provisioned in Semrush and observe `200` with `semrushSyncPending: true`.
2. Turn the flag on: `PUT /organizations/{orgId}/feature-flags/llmo/serenity_ui` with `{"value": true}` and an admin key.
3. Allow up to the negative cache TTL (30s) for warm containers to pick up the flip.
4. Repeat the same edit. Expect `403` rather than `200`, no `semrushSyncPending` in the body, and no internal gateway URL in the body or the `x-error` header. Confirm the row still shows the edit — the write persists on both paths.
5. Edit `competitors` on the same brand. Expect the write to be rejected before the row is touched (the pre-write guard), i.e. the brand row is unchanged afterwards.
6. `DELETE` the same flag path and confirm the org falls back to the absorbed behaviour.

Prod rollout is the same sequence, but only for an org that is genuinely past the migration — turning it on for an org that still has unprovisioned Semrush users would start failing their brand edits, which is exactly what the allowance exists to prevent.

- https://github.com/adobe/project-elmo-ui/pull/2565 — related. Introduces the frontend half of the same flag, currently spelled `serenity-ui`. It needs to move to `serenity_ui` or it will never match a stored row.
- https://github.com/adobe/spacecat-api-service/pull/2907 — already merged. This PR gates the behaviour that one introduced.

Ordering is unconstrained: the flag defaults off, so this can merge and deploy on its own with no behaviour change for any org. The real sequence is operational — the elmo PR must agree on the spelling, and a `serenity_ui` flag row must be created for an org before either side changes behaviour for it.

### DIFF
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -1917,12 +1917,19 @@ function BrandsController(ctx, log, env) {
       // memo also pins one answer for the whole request, so the pre-write
       // competitor guard and the post-commit re-sync below cannot disagree and
       // half-absorb an edit.
-      let serenityUiFlag;
-      const shouldSurfaceSemrushErrors = () => {
-        if (serenityUiFlag === undefined) {
-          serenityUiFlag = isSerenityUiActiveForOrg(context, spaceCatId, log);
+      let serenityUiFlagPromise;
+      /**
+       * MUST be awaited — this returns the memoised PROMISE, not a boolean, so a
+       * bare `if (resolveSurfaceSemrushErrors())` is always truthy and would
+       * silently surface every Semrush error regardless of the flag.
+       *
+       * @returns {Promise<boolean>} `true` when Semrush failures must be reported.
+       */
+      const resolveSurfaceSemrushErrors = () => {
+        if (serenityUiFlagPromise === undefined) {
+          serenityUiFlagPromise = isSerenityUiActiveForOrg(context, spaceCatId, log);
         }
-        return serenityUiFlag;
+        return serenityUiFlagPromise;
       };
 
       // LLMO-5645: a region must not be removed from a brand while prompts still
@@ -1996,7 +2003,7 @@ function BrandsController(ctx, log, env) {
             // Logged before the rethrow either way: the outer catch only sees a
             // generic "Error updating brand", so this is the sole record of which
             // Semrush workspace and which pre-write step failed.
-            const rejectingWrite = await shouldSurfaceSemrushErrors();
+            const rejectingWrite = await resolveSurfaceSemrushErrors();
             log.warn('serenity: competitor-guard project listing failed', {
               brandId,
               semrushSubWorkspaceId: brandState.semrushSubWorkspaceId,
@@ -2051,7 +2058,7 @@ function BrandsController(ctx, log, env) {
       // surface as a 5xx on an edit that did persist. The response carries
       // semrushSyncPending:true so the divergence is visible to the caller, and the
       // error log below carries the context needed to enumerate and recover drifted
-      // brands. For an org on the Serenity UI (see shouldSurfaceSemrushErrors) the
+      // brands. For an org on the Serenity UI (see resolveSurfaceSemrushErrors) the
       // failure is reported instead — that org's users are provisioned in Semrush, so
       // a failure is a real fault rather than expected migration noise.
       const urlsTouched = updates.urls !== undefined
@@ -2148,7 +2155,7 @@ function BrandsController(ctx, log, env) {
             stack: syncError?.stack,
             permanent: isPermanent,
           });
-          if (await shouldSurfaceSemrushErrors()) {
+          if (await resolveSurfaceSemrushErrors()) {
             // The org is on the Serenity UI: its users are provisioned in Semrush, so
             // this is a genuine upstream failure and the caller is told. The row stays
             // committed either way — the difference is whether the divergence is
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -72,7 +72,7 @@ import { isSemrushTransportError, unwrapTransportCause } from '../support/sereni
 import { syncBrandUrlsAcrossMarkets } from '../support/serenity/brand-urls.js';
 import { syncBrandAliasesAcrossMarkets } from '../support/serenity/brand-aliases.js';
 import { resolveProjects } from '../support/serenity/resolve-projects.js';
-import { isSerenityActiveForOrg } from '../support/serenity/serenity-active.js';
+import { isSerenityActiveForOrg, isSerenityUiActiveForOrg } from '../support/serenity/serenity-active.js';
 import {
   buildReservedDomains,
   dropReservedCompetitors,
@@ -1903,6 +1903,28 @@ function BrandsController(ctx, log, env) {
         }
       }
 
+      // LLMO-6565 migration-window allowance. A Semrush failure on this edit is
+      // absorbed — the row is committed and the response carries
+      // semrushSyncPending — only while the org is still migrating: its LLMO users
+      // may have no Semrush user record yet, so Semrush answers 4xx on every call
+      // made on their behalf and no retry the caller can make will clear it. An org
+      // pinned to the Serenity UI is past that window, its users are provisioned, so
+      // a Semrush failure there is a real error and propagates to the caller.
+      //
+      // Resolved lazily, and at most once per request: only an actual Semrush
+      // failure needs the answer, so the happy path — and every flat-mode brand
+      // edit, which never reaches a Semrush call at all — pays no flag read. The
+      // memo also pins one answer for the whole request, so the pre-write
+      // competitor guard and the post-commit re-sync below cannot disagree and
+      // half-absorb an edit.
+      let serenityUiFlag;
+      const shouldSurfaceSemrushErrors = () => {
+        if (serenityUiFlag === undefined) {
+          serenityUiFlag = isSerenityUiActiveForOrg(context, spaceCatId, log);
+        }
+        return serenityUiFlag;
+      };
+
       // LLMO-5645: a region must not be removed from a brand while prompts still
       // use it — DRS schedules off each prompt's `regions`, so dropping a brand
       // region would orphan those prompts on a market the brand no longer
@@ -1961,20 +1983,32 @@ function BrandsController(ctx, log, env) {
         if (hasText(brandState?.semrushSubWorkspaceId)) {
           // Semrush brand: market/project domains come from the project listing.
           // List once and stash for the post-commit re-sync (see prefetchedProjects).
-          // Best-effort: if the listing fails (e.g. user not provisioned in Semrush),
-          // degrade to brand-URL-only reserved set and log — the self-reference guard
-          // is a data-quality nicety and must not block the competitor write entirely.
+          // Best-effort for a migrating org: if the listing fails (e.g. user not
+          // provisioned in Semrush), degrade to brand-URL-only reserved set and log —
+          // the self-reference guard is a data-quality nicety and must not block the
+          // competitor write entirely. For an org on the Serenity UI the failure is
+          // real and rejects the write before the row is touched.
           try {
             const imsToken = await resolveSemrushImsToken(context, log, 'brands');
             const transport = createSerenityTransport({ env: context.env, imsToken });
             prefetchedProjects = await resolveProjects(transport, brandState.semrushSubWorkspaceId);
           } catch (guardError) {
-            log.warn('serenity: competitor-guard project listing failed; degrading to brand-URL-only reserved set', {
+            // Logged before the rethrow either way: the outer catch only sees a
+            // generic "Error updating brand", so this is the sole record of which
+            // Semrush workspace and which pre-write step failed.
+            const rejectingWrite = await shouldSurfaceSemrushErrors();
+            log.warn('serenity: competitor-guard project listing failed', {
               brandId,
               semrushSubWorkspaceId: brandState.semrushSubWorkspaceId,
               status: guardError?.status,
               message: guardError?.message,
+              // true → the write is rejected; false → degraded to the reserved set
+              // built from the brand's own URLs alone, and the write proceeds.
+              rejectingWrite,
             });
+            if (rejectingWrite) {
+              throw guardError;
+            }
           }
           reservedDomains = buildReservedDomains(
             (prefetchedProjects ?? []).map((p) => p?.domain),
@@ -2012,11 +2046,14 @@ function BrandsController(ctx, log, env) {
       // every market/project (region-filtered per market). Skipped for flat-mode
       // brands and unrelated edits. One transport shared across all three syncs.
       //
-      // Best-effort, matching the create path: the brand row is already committed
-      // when this block runs, so a re-sync failure must not surface as a 5xx on an
-      // edit that did persist. The response carries semrushSyncPending:true so the
-      // divergence is visible to the caller, and the error log below carries the
-      // context needed to enumerate and recover drifted brands.
+      // For a migrating org this is best-effort, matching the create path: the brand
+      // row is already committed when this block runs, so a re-sync failure must not
+      // surface as a 5xx on an edit that did persist. The response carries
+      // semrushSyncPending:true so the divergence is visible to the caller, and the
+      // error log below carries the context needed to enumerate and recover drifted
+      // brands. For an org on the Serenity UI (see shouldSurfaceSemrushErrors) the
+      // failure is reported instead — that org's users are provisioned in Semrush, so
+      // a failure is a real fault rather than expected migration noise.
       const urlsTouched = updates.urls !== undefined
         || updates.socialAccounts !== undefined
         || updates.earnedContent !== undefined;
@@ -2111,6 +2148,15 @@ function BrandsController(ctx, log, env) {
             stack: syncError?.stack,
             permanent: isPermanent,
           });
+          if (await shouldSurfaceSemrushErrors()) {
+            // The org is on the Serenity UI: its users are provisioned in Semrush, so
+            // this is a genuine upstream failure and the caller is told. The row stays
+            // committed either way — the difference is whether the divergence is
+            // reported or absorbed. createErrorResponse maps the status and keeps the
+            // internal gateway URL out of the response. Any partial rejectedAliases
+            // are dropped with the body; the log above carries the same context.
+            throw syncError;
+          }
           // Preserve any partial rejectedAliases collected before the throw so
           // the UI can still warn about aliases that were refused by Semrush.
           return ok({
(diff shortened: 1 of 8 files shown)


src/controllers/brands.js
```diff
@@ -1903,6 +1903,28 @@ function BrandsController(ctx, log, env) {
         }
       }
 
+      // LLMO-6565 migration-window allowance. A Semrush failure on this edit is
```
MysticatBot, line 1906: issue (blocking): The project CLAUDE.md states: "New or modified endpoints must include integration tests in `test/it/` - add shared test logic in `shared/tests/`, seed data in `postgres/seed-data/`, and a wiring file in `postgres/`."

This PR modifies the behavior of PATCH `/v2/orgs/{org}/brands/{brandId}` by adding conditional error propagation (new 401/502 response codes) gated by the `serenity_ui` flag. While unit tests comprehensively cover the controller logic, no integration tests in `test/it/` verify the end-to-end behavior.

Additionally, the 401 path is documented in the OpenAPI spec but untested in the unit tests. Add a test case where `syncBrandUrlsAcrossMarkets` throws a `SerenityTransportError(401, ...)` with `isSerenityUiActiveForOrg: true` and assert `response.status === 401`.

If the IT infrastructure cannot support the full integration test yet (mock Semrush containers needed), document the gap with a tracking comment or issue reference so it does not get lost.

MysticatBot (changes requested):
Hey @rainer-friederich,

**Verdict: Request changes** - missing integration test coverage per project rules, and the documented 401 response path has no test.
**Complexity: HIGH** - large diff (496 additions across 7 files), API surface change.
_Changes: Gates the Semrush-failure allowance on brand edits behind the new org-level `LLMO/serenity_ui` feature flag so fully-migrated orgs get real error reporting instead of silent absorption (7 files)._
Note: Recommend a human read before merge - this change modifies documented platform behavior (docs/serenity.md) and shared API contracts (OpenAPI specs). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Important] Missing integration tests for modified endpoint behavior per CLAUDE.md rule: "New or modified endpoints must include integration tests in `test/it/`" - the PATCH brands endpoint now returns new status codes (401/502) gated by a PostgREST-backed flag. If the IT harness can support this (seed a `feature_flags` row, stub Semrush to fail), add coverage; if not, document the gap with a tracking reference so it does not get lost. - `test/it/` (absent)
2. [Important] The OpenAPI spec documents a 401 response and the production code maps `err.status === 401` distinctly, but no test exercises this path. The 403 and 502 paths are both tested; 401 (expired/missing promise token) is not. - `test/controllers/brands.test.js`

<details>

- nit: `shouldSurfaceSemrushErrors` stores and returns a Promise but naming suggests boolean - a future maintainer could write `if (shouldSurfaceSemrushErrors())` (always truthy) and get a silent always-true gate. Consider renaming the variable to `serenityUiFlagPromise` or adding a one-line comment. - `src/controllers/brands.js:1919`
- suggestion: Replace "promise token" with "IMS token" or "authentication token" in the 401 OpenAPI description - internal jargon that external spec readers won't understand. - `docs/openapi/brands-v2-api.yaml:188`
- suggestion: In the "reads the flag once when both fail" test, add `expect(syncCompetitorBenchmarksAcrossMarkets).to.have.been.calledOnce` - without it the test passes even if the controller short-circuits before reaching the re-sync. - `test/controllers/brands.test.js`
- suggestion: The 502 test should include the same gateway-URL-leak negative assertions as the 403 test (`expect(JSON.stringify(body)).to.not.contain('gw.internal')` and on the `x-error` header). - `test/controllers/brands.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 4m 59s | <b>Cost:</b> $8.39 | <b>Commit:</b> <code>e8d71d44e92e3dc1f80d5fc772ef8305f0b1b543</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the review in `8dadb6a0e`. All six items, both blocking and non-blocking.

**Must fix**

- **[Important] Missing integration test** — done, and it found a better target than the endpoint itself. The IT harness cannot induce a Semrush failure on a sub-workspace brand (the existing brands IT deliberately uses flat-mode brands, and the vendor mock has no failure-injection control), so an endpoint-level IT for the 401/502 paths is not reachable today. What *is* reachable is the part that genuinely needs a real database: the flag name. `feature_flags` carries `CHECK (flag_name ~ '^[a-z][a-z0-9_]*$')` and nothing on the read path validates a name, so a wrong spelling resolves to `false` forever with no error surfaced anywhere — indistinguishable from "the flag is off". The new IT writes `serenity_ui` through the admin endpoint against real Postgres, reads it back through the same storage helper the controller's gate uses, and asserts the hyphenated `serenity-ui` is refused with 400. That is exactly the defect class this PR had to correct against the ticket, and only an IT can catch it.
- **[Important] 401 path untested** — done. `resolveSemrushImsToken` throws `ErrorWithStatusCode(401)` for an expired promise-token exchange or a non-IMS bearer, and it sits inside the re-sync try, so mid-migration it soft-fails like any other failure while the flag maps it through to 401. Now covered, including that the row still committed.

**Non-blocking**

- **nit: `shouldSurfaceSemrushErrors` returns a Promise but reads as a boolean** — valid and the sharpest catch here, since `if (fn())` would have silently gated every Semrush error open. Renamed to `resolveSurfaceSemrushErrors` (matching the `resolve*` async idiom already used in this file) with JSDoc stating it MUST be awaited and returns `Promise<boolean>`.
- **suggestion: "promise token" jargon in the 401 spec description** — done, reworded to IMS user token in terms a spec reader can act on.
- **suggestion: assert the competitor sync ran in the memo test** — the intent was right and it exposed a real weakness, but the specific stub is never reached on that path: the guard's listing failure leaves nothing prefetched, so the re-sync lists projects again and fails *there*, before any per-entity sync. Asserting on `syncCompetitorBenchmarksAcrossMarkets` failed outright. Asserted the equivalent correct signal instead — the transport listing called exactly twice, proving both catch blocks ran and the single flag read really does mean the memo held.
- **suggestion: gateway-URL-leak assertions on the 502 test** — done, same two assertions as the 403 test.

**On the two notes in the verdict**

- CI was green at review time and is green again after this push.
- Agreed on wanting a human read: this changes documented platform behaviour and a shared API contract. Worth flagging for that reader — the ticket specifies the flag as `serenity-ui` and states the backend already emits it. Neither holds. Verified against mysticat-prod: the CHECK constraint rejects hyphens, and prod `feature_flags` contains only `brandalf`, `brandalf_migration` and `serenity`. This PR uses `serenity_ui`, and the frontend PR needs the same spelling or it will never match a stored row.

src/controllers/brands.js
rainer-friederich: Fixed in `8dadb6a0e`. Renamed to `resolveSurfaceSemrushErrors` (matching the `resolve*` async idiom already used in this file) and added JSDoc stating it MUST be awaited and returns `Promise<boolean>` — a bare truthiness check would have silently surfaced every Semrush error regardless of the flag.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all prior findings addressed, no new issues found.
**Complexity: HIGH** - large diff (500+ lines, 8 files); API surface change + docs update.
_Changes: Gates the Semrush-failure allowance on brand edits behind the new `LLMO/serenity_ui` org-level feature flag so fully-migrated orgs get real error reporting (8 files)._
Note: Recommend a human read before merge - this change modifies documented platform behavior (docs/serenity.md) and shared API contracts (OpenAPI specs). The bot review is a complement to, not a replacement for, a human read here.

- Integration test for flag name against real Postgres added
- 401 path now tested (IMS token failure with flag ON)
- Renamed to `resolveSurfaceSemrushErrors` with JSDoc await warning
- "promise token" reworded to IMS user token in OpenAPI spec
- Memo test now asserts `listProjectsStub.calledTwice` proving both catch blocks ran
- Gateway-URL-leak negative assertions added to 502 test

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 33s | <b>Cost:</b> $4.36 | <b>Commit:</b> <code>8dadb6a0e1a336eb2b21a7066afd8f36580e868e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
### Runtime validation — not performed, and why

Recording this rather than leaving it implied. This PR has a runtime surface, so it would normally be exercised on dev before merge. It was not, and I could not get there:

- **The branch is not deployed anywhere.** Standing up a personal dev Lambda alias needs AWS credentials; `spacecat-dev` returns `ExpiredToken` and `klam login` is an interactive browser flow.
- **Dev API auth failed both ways.** The `ci` alias is up (`GET /api/ci/slack/events` → 200, `GET /api/ci/organizations` → 401 unauthenticated, as expected), but neither an IMS user token nor the dev `ADMIN_API_KEY` from Vault authenticates against it — both 401. So even the reduced check (write `serenity_ui` through the admin endpoint, read it back) was not reachable.
- **The reported branch needs a real upstream 4xx.** It fires when the calling user has no Semrush user record. That is an identity deliberately absent upstream; it cannot be manufactured on demand, and the vendor mock has no failure-injection control either.

Note that the `ci` alias serves `main`, so a flag write there would have validated the environment, not this change.

**What stands in its place.** The one genuinely environment-dependent fact here is the flag name, and it is verified twice against real PostgreSQL rather than a mock:

- Against **mysticat-prod**: `pg_constraint` for `feature_flags_flag_name_valid` is `CHECK ((char_length(flag_name) <= 255) AND (flag_name <> '') AND (flag_name ~ '^[a-z][a-z0-9_]*$'))`, and the full flag inventory is three names — `brandalf`, `brandalf_migration`, `serenity`.
- In **CI's `it-postgres`** on this PR's HEAD, against real Postgres: `✔ accepts serenity_ui and rejects the hyphenated serenity-ui (LLMO-6565)`.

Both behavioural branches — absorbed and reported, across both re-armed call sites — are covered at the controller level.

**Blast radius if this merges unvalidated.** Bounded to nothing today: the flag defaults OFF, and no environment has a `serenity_ui` row. Every org keeps the current absorbed behaviour until someone creates that row. The first real runtime exercise will be the rollout itself, which is why the test plan in the description spells out the enable/verify/disable sequence rather than assuming a green e2e here.

rainer-friederich:
### Correction to the runtime-validation note above

Two statements in the previous comment are wrong and worth correcting on the record, because they understate what was available.

**This branch was deployed.** CI deployed HEAD `8dadb6a0` to the shared `dev-branches` environment at 14:15 UTC (deployment `5659050631`, state `success`) — visible as the "temporarily deployed to dev-branches" entry on the timeline. So the claim that the branch is not deployed anywhere is false, and the follow-on claim that the `ci` alias serves `main` is also false: that alias serves whichever branch most recently pushed a green build, which for roughly thirteen minutes was this one.

**What was actually blocking is narrower than stated: credentials, not availability.** Neither an IMS user token nor the dev `ADMIN_API_KEY` authenticated against the alias — both 401 — so the reduced check could not be run even though the environment was up and serving this exact commit. That is a different and much smaller gap than "nothing to test against".

**Why it is still not validated now.** The `dev-branches` slot is shared and single-writer: an unrelated branch claimed it at 14:28 UTC and holds it currently. Any request sent there now would exercise that branch's code, not this one, so a result gathered now would be meaningless rather than reassuring.

Everything in the previous comment under "What stands in its place" and "Blast radius" is unaffected and still holds — the flag-name verification against prod `pg_constraint` and against real Postgres in `it-postgres`, and the fact that the flag defaults OFF with no `serenity_ui` row in any environment.

igor-grubic:
Cross-posting a question from the frontend half (project-elmo-ui#2565, LLMO-6565), raised in review there.

On the UI side, `serenity_ui` is a **full-Semrush override**: when set for an org it forces Serenity mode **and** the Semrush data source (`isSemrushEnabled`) **above** the `FT_LLMO-5004` / `showSerenityDashboards` LaunchDarkly gates — i.e. FF trumps LD, independent of the LD entitlement. The client treats this as UI/data-source selection and is fail-closed, not as an authorization boundary.

Question for the backend: do the Serenity / `-sr` (Semrush) data endpoints **re-enforce entitlement server-side**, independent of `serenity_ui`? We want to confirm that a `serenity_ui` org that is *not* otherwise Semrush-entitled can't reach Semrush data purely because the client flag is on — the FE is relying on the server as the actual authz gate. (Same concern noted as "make server-side enforcement explicit" in the elmo review.)

Not blocking this PR — just want the enforcement boundary documented before `serenity_ui` is enabled for any real org.

PR #2953, opened 2026-07-29
fix(llmo): tolerate null opportunity tags in brand aggregate filter | LLMO-6639

baranovskyalexandr:
## Summary
- Guard `opportunity.getTags()` with `?? []` in LLMO brand-aggregated opportunities filter and Slack LLMO usage command so `tags: null` no longer throws `TypeError` and wipes an entire site’s opportunities from `GET /org/{orgId}/brands/all/opportunities`.
- Add regression tests for null-tags on per-site aggregation and `brandId=all`.

## Context (LLMO-6639)
Mystique meta-tags V2 can create opportunities with `tags: null`. Spreading null crashed the filter inside a per-site `try/catch`, so Halliburton (and any site with one bad opp) returned **0** opportunities on org chrome while the site API still had data.

## Test plan
- [x] Unit tests: null-tags site kept; sibling LLMO opps retained for `brandId=all`
- [ ] Deploy and verify `GET /org/{orgId}/brands/all/opportunities` for Halliburton org returns non-empty when a null-tags opportunity exists (or after recreating one)

Jira: https://jira.corp.adobe.com/browse/LLMO-6639

### DIFF
--- a/src/controllers/llmo/opportunities/llmo-opportunities-controller.js
+++ b/src/controllers/llmo/opportunities/llmo-opportunities-controller.js
@@ -27,7 +27,10 @@ const VALID_STATUSES = new Set(['NEW', 'IN_PROGRESS']);
  * Matches the filtering logic used in get-llmo-opportunity-usage.js.
  */
 function isLlmoOpportunity(opportunity) {
-  const tags = [...(opportunity.getTags())];
+  // getTags() can be null/undefined when the DynamoDB set attribute is unset
+  // (e.g. a freshly created meta-tags opportunity). Spreading null throws and
+  // used to wipe the entire site's aggregate via the surrounding try/catch.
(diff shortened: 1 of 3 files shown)


BaDos (approved):
LGTM

Non-blocking):

1. **Slack command has no null-tags test.** The same `?? []` fix was applied to `get-llmo-opportunity-usage.js`, but its inline filter is a verbatim copy of `isLlmoOpportunity` and has no regression test for the null-tags case. A one-line case in the existing Slack test would close the gap.

2. **Duplicated filter logic.** `get-llmo-opportunity-usage.js` duplicates `isLlmoOpportunity` (the controller already exports it — the comment even says "Matches the filtering logic used in get-llmo-opportunity-usage.js"). Two copies drifting is how this class of bug recurs; consider extracting to a shared util so the rule + its guard live in one place.

Issue #2954, opened 2026-07-29
Cover the prompt authorship metadata read path in the integration suite (LLMO-6289)


rainer-friederich:
Prompt authorship metadata (LLMO-6289) stamps `created_at/by` and `updated_at/by` on every native
prompt write. The write path is covered end to end against the Project Engine mock; the **read**
path is not, and it is broken in production in two independent ways. This issue covers the
remaining integration coverage and the two source fixes it pins down.

Write-path coverage — the four-key stamp reaching upstream on create, in-place edit, dedupe,
409 and 404 — is asserted through the mock's store snapshot and lands separately. What remains is
everything downstream of the read.

## Part 1 — the list read never opts into metadata (actionable now)

`listPromptsByTags` in `src/support/serenity/rest-transport.js` never sends `include_metadata`;
the string appears nowhere in `src/`. Semrush therefore omits the `metadata` key from every item,
and `buildPromptDto`'s `metadata?.created_at ?? null` yields `null` for all four fields — so
prompts that ARE stamped surface as empty values to the UI.

The flag belongs on the **query string**. Passing it in the body makes the key read back `null`,
which looks like "the prompt is unstamped" and is the more misleading of the two failures
(verified live 2026-07-29 against LLMO-Dev-2, Brand-A, project `997e17c3…`).

The mock is faithful here — it gates inline `metadata` behind the `include_metadata=true` query
param and omits the key entirely when the flag is absent — so integration tests genuinely fail
before the fix and pass after it.

Coverage to add:

- A stamped prompt read back through `GET /serenity/prompts` carries non-null `createdAt`,
  `createdBy`, `updatedAt`, `updatedBy` matching what was stamped upstream.
- After an in-place text edit, the read shows the bumped `updatedAt` / `updatedBy` and the
  preserved `createdAt` / `createdBy`.
- A prompt with no metadata reads back four nulls rather than erroring.

Because `npm test` excludes `test/it/**`, the fix needs unit coverage as well to satisfy the
codecov patch gate — the integration suite alone will not carry it.

## Part 2 — the sorted read sends the wrong keys (blocked)

> **Blocked by https://github.com/adobe/spacecat-shared/issues/1856** — do not start Part 2 until
> that ships and the client is released.

`listPromptsByTags` sends `sort` / `order` in the request body. The wire contract is
`sort_field` / `sort_dir`, so the server-side "Last modified" sort is a silent no-op: a 200 in
[...]

PR #2956, opened 2026-07-29
test(serenity): cover prompt authorship metadata stamping in the IT suite (LLMO-6289)

rainer-friederich:
## 1. Abstract

Integration coverage for the prompt authorship metadata that every native prompt write stamps upstream — `created_at` / `created_by` and `updated_at` / `updated_by` — driven end to end against the Semrush Project Engine mock. Test-only; no production source changes.

## 2. Reasoning

The authorship stamping feature shipped with unit coverage only. Its correctness depends entirely on what actually reaches the vendor — which caller is attributed, which of the two timestamp pairs moves on an edit, and whether a refused write leaves the stored stamp alone — and a stubbed transport cannot demonstrate any of that. The integration suite already drives the real Project Engine mock over HTTPS, so the write contract can be pinned there instead of being assumed.

A new `prompt authorship metadata (live mock)` block in the serenity integration factory exercises the stamping contract through the public API and then inspects what the vendor actually persisted:

- A create stamps all four keys, with `created_*` and `updated_*` carrying the same instant and the same caller — a create is its own first edit.
- An edit advances only the updated pair; the created pair is immutable. Asserted across two personas, so the attribution assertion also proves the author is taken from the requesting token rather than the bearer forwarded upstream, and does not depend on millisecond timestamp granularity.
- Re-categorising a prompt without changing its text still advances the updated pair, because the edit body is the full next state and the combined upstream write always carries both name and metadata.
- A repeated text folds into the existing prompt and preserves the original stamp, so a dedupe hit never re-attributes an existing prompt to whoever re-submitted it.
- A refused write leaves authorship untouched: a `409` text collision, a `404` unknown id, and a bulk delete all leave stored metadata byte-identical.
- The `sort` / `order` query allow-list refusals are asserted directly, including the asymmetry that `order` is only validated once a `sort` is present.

**Why the assertions read the mock's store rather than the API's own list response.** The upstream `by_tags` list gates the `metadata` block behind an `include_metadata=true` query parameter, and the service does not currently send it. Reading the list DTO therefore cannot distinguish "the write never stamped" from "the read never asked" — both surface as empty values. These tests read the mock's store snapshot through a new `dumpPeMock()` control-route helper, which separates the two and scopes this PR to the write contract alone. Surfacing the four values on the read DTO is deliberately out of scope here and tracked separately.

No behaviour change is visible to users or operators: the diff touches only the integration suite and its documentation.

## 4. Required information

- Jira / issue: LLMO-6289 (parent feature; this PR is follow-up test coverage for it)
- Spec: https://github.com/adobe/serenity-docs/pull/35
- Other: https://github.com/adobe/spacecat-api-service/issues/2954 (read-path follow-up) · https://github.com/adobe/spacecat-shared/issues/1856 (mock sort fidelity, blocks the sorted-read coverage)

- `spacecat-shared` — contract dependency, not changed here. The integration harness resolves the Project Engine mock image tag from the *installed* `@adobe/spacecat-shared-project-engine-client` version, so this coverage only holds against a mock built from the client version this repo ships. It exercises the v3 metadata write surface that landed in client 1.17.0, which `package.json` already pins.

Verification was done against the live integration stack (Docker: PostgreSQL, PostgREST, MinIO, and both Semrush vendor mock containers), not against stubs.

The suite was mutation-tested to confirm the new assertions can actually fail, since a test asserting vendor-side state can easily pass vacuously:

- Removing the create-path stamp fails every new test except the `sort` / `order` allow-list case, which correctly does not depend on stamping.
- Removing only the sub-workspace edit-path stamp fails exactly the two edit cases and nothing else. That precision also established that this API surface exercises the sub-workspace prompt twin rather than the flat one.

That second result exposed a gap worth recording: because both mutations delete the metadata write wholesale, neither could catch a narrower regression that skipped the merge-patch while leaving a stale-but-valid timestamp. The tag-only edit case now asserts on the timestamp's value rather than merely its presence, closing that hole.

One environment note for anyone reproducing this locally: the harness derives the mock image tag from the installed client version rather than from `package.json`, so a stale `node_modules` silently runs the suite against an older mock without the v3 metadata surface. Running `npm install` first is required.

### Review suggestions considered and deliberately not taken

Recorded so they are not re-raised — each was evaluated against the file as it exists rather than in the abstract:

- **Hoisting the per-block `base` / `US_GEO` / market / category helpers into shared scope.** Self-contained describe blocks are this file's established convention, not something this PR introduces: on `main`, `base` is already declared five times, `US_GEO` three, and the market helper twice, each re-declared per block. Hoisting would mean editing five pre-existing blocks to depart from the prevailing pattern.
- **Collapsing the repeated store-snapshot fetches.** Roughly a fifth of this block's snapshot reads are redundant, but they cost a negligible fraction of its runtime, and the suggested "fetch once, look up directly" form bypasses the helper that asserts a metadata block is actually populated — reintroducing the vacuous-comparison hole that helper exists to close. Negligible speed is not worth real assurance.
- **Extracting the mock-controls skip guard into a shared helper.** Two call sites, and the case for extracting rested on a third arriving later. It would also mean editing a pre-existing block, against the same convention as the first item.

One suggestion was accepted and applied: the two store-snapshot helpers now share a private implementation instead of duplicating an identical fetch body.

## 7. Test plan

**Verified locally.** The full serenity integration suite was run against the live mock containers and is green, including the two pre-existing dynamic-allocation tests that consume the `dumpUmMock` helper refactored here. The mutation runs described in section 6 confirm the new tests fail when the behaviour they cover is removed.

```
npx mocha --require test/it/postgres/harness.js --timeout 60000 test/it/postgres/serenity.test.js
```

The `60000` timeout is required: this suite drives the vendor mocks over real HTTPS and the default is too tight. Docker plus an ECR login is needed for the private PostgREST image; see `test/it/README.md`.

**Per-environment verification.** None applies. This PR adds no runtime code, so there is nothing to observe on eph, dev, stage or prod — CI running the integration suite is the complete signal.

- https://github.com/adobe/spacecat-api-service/issues/2954 — related follow-up, not a dependency. It carries the read-path fix (sending `include_metadata` as a query parameter) and, once unblocked, the `sort_field` / `sort_dir` wire-key correction. This PR is intentionally independent of it and can merge on its own.
- https://github.com/adobe/spacecat-shared/issues/1856 — blocks the *sorted-read* portion of that follow-up, not this PR. The mock implements no sort logic today, so an ordering assertion would pass regardless of which sort keys the service sends; the mock has to distinguish them before that coverage is meaningful.

No ordering constraint applies to merging this PR.


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-structured test-only PR with sound assertion strategy.
**Complexity: MEDIUM** - medium diff; tests-heavy.
_Changes: Adds integration tests for prompt authorship metadata stamping against the PE mock's store snapshot, plus a refactored dumpMock helper and README documentation (4 files)._

<details>

- nit: `createUsMarket` does not assert its response status unlike the sibling helpers (`createCategory`, `createPrompt`) - a setup failure would surface as a confusing downstream error rather than a clear "market creation failed" message - `test/it/shared/tests/serenity.js:935`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 45s | <b>Cost:</b> $4.66 | <b>Commit:</b> <code>82b6c3630d31b0e179adb109468d8e020c30cbcd</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Thanks — the one non-blocking item is addressed in `0849a76ed`.

**nit: `createUsMarket` does not assert its response status — `test/it/shared/tests/serenity.js:935`** — fixed.

Worth recording the nuance, because the comparison in the finding points at the wrong siblings. `createUsMarket` is not new in this PR: it already exists twice on `main`, and both of those definitions are the same non-asserting arrow form. That shape is deliberate there — those blocks contain tests where the market response *is* the subject (`const created = await createUsMarket(); expect(created.status).to.equal(201);`), so the helper hands the response back rather than asserting on it.

The concern underneath the finding is still real, though, and not merely stylistic: market creation genuinely can be refused with a 422 on insufficient units, so a silently-failed setup is a failure mode this system actually has, not a hypothetical one. In this block nothing asserts on the market response — all nine call sites discard it — so the assertion now lives inside the helper. A refused market create fails at the market call, rather than surfacing as a puzzling status mismatch on the tag or prompt request that depended on it.

Scoped to this block only: the two pre-existing definitions keep their current shape, since changing them would mean rewriting the tests that assert on the market status directly.

Verified against the live integration stack — 56 passing, 0 failing. The new assertion is exercised on every test in the block, which also confirms 201 is the correct expectation; a wrong one would have turned the run red rather than passing silently.

PR #2957, opened 2026-07-29
fix(serenity): request prompt metadata and use the real sort keys on by_tags (LLMO-6289)

rainer-friederich:
## 1. Abstract

Corrects three request-shape defects on the Semrush `by_tags` prompt read and the batch metadata PATCH, so prompt authorship metadata is actually returned, the "Last modified" sort actually sorts, and the batch stamp surface addresses prompts by the key upstream defines.

## 2. Reasoning

Prompt authorship metadata is written correctly — a prompt created through the UI carries all four `created_*` / `updated_*` values upstream — but the Prompt Library renders an em dash in every authorship column. The read side never asks for the data.

`include_metadata` is a **query** parameter on `by_tags`. It was never sent. Upstream's response to a request that omits it is a 200 with the `metadata` key absent from every item, so `buildPromptDto`'s `metadata?.created_at ?? null` mapped all four DTO fields to `null`. The UI's em dash — its correct rendering for an un-backfilled prompt — became indistinguishable from a stamped one.

Two more defects of the same kind sit next to it. The body's sort keys are `sort_field` / `sort_dir` (`model.AIOPromptsListRequest`); we sent `sort` / `order`, and upstream ignores unknown body keys with a 200, so the server-side sort was a silent no-op. The batch metadata PATCH keys its items `prompt_id` (`model.PatchAIOPromptsBatchItem`); we sent `id`, which carries no prompt reference at all.

All three shared one enabler: the request bodies at this boundary were passed through an `any` cast, so nothing compared them against the generated contract. Now that the vendored client ships the v3 facade natively, the cast has no remaining purpose.

**Prompt authorship metadata is returned.** The `by_tags` read sends `include_metadata=true` as a query parameter on every call. Before: `metadata` absent from every item, all four DTO authorship fields `null`, em dashes in the Prompt Library for prompts that are correctly stamped. After: stamped prompts show their real author and timestamps; genuinely un-backfilled prompts still show an em dash, which is now the only thing an em dash means.

The flag is sent unconditionally rather than exposed as a per-call option. A caller that forgets to opt in receives data that cannot be told apart from unstamped data — the failure has no signal — and the cost of always asking is four short strings per item. This also covers the tag-walk readers (`listTagsForProject`, `countPublishedPrompts`), which ignore the field.

**Server-side prompt sorting works.** The transport keeps `sort` / `order` as its own parameter names — the public API query parameters and the `resolveSort` allow-list are unchanged — and maps them onto `sort_field` / `sort_dir` at the wire boundary. Before: "Last modified" returned the default order. After: it sorts. No change to any caller or to the OpenAPI contract.

**The batch metadata PATCH addresses prompts correctly.** Items are keyed `prompt_id`. The method takes `{ promptId, metadata }` in camelCase, consistent with the rest of the transport's JS-facing API, and maps to the upstream key internally. This surface has no caller yet, so the defect was latent rather than live.

**The boundary is type-enforced instead of cast away.** The `ProjectEngineTransportWithMetadata` intersection typedef and its `any` cast are gone — the vendored client's generated types now cover the v3 facade — so request shapes at this boundary derive from the generated contract, matching how the rest of this file already treats Semrush shapes. The list body is annotated with the generated type and built without an object spread, because a spread suppresses excess-property checking and is exactly what lets a misspelled key through. A wrong key name or value type at this call site is now a compile error.

Consequently the WP2 dependency gate is retired. It asserted at runtime that the installed client exposes the four v3 write methods; with the cast removed, the strict type-check tier reports a missing method directly against the real client type, which is a stronger and earlier check than the introspection test it replaces.

## 4. Required information

- Jira / issue: LLMO-6289 — https://jira.corp.adobe.com/browse/LLMO-6289
- GitHub issue: https://github.com/adobe/spacecat-api-service/issues/2954 — this PR delivers both source fixes that issue pins down; its remaining integration coverage is not in this diff (see section 8), so the issue stays open.
- ADR(s): `docs/decisions/006-serenity-v1-v2-read-drift.md` (the `by_tags` live-layer read this fix extends), `docs/decisions/005-opt-in-type-checking.md` (the type-checking scope the boundary now leans on)

- **project-elmo-ui** — consumer, no change needed. Its Prompt Library authorship columns and CSV export read the four DTO fields this PR starts populating; they render an em dash today purely because the values arrive `null`. The columns begin showing real data as soon as this deploys, with no UI change.
- **spacecat-shared** — contract. The Project Engine mock's `by_tags` handler implements no sort logic, so it cannot distinguish the corrected sort keys from the ignored ones; teaching it that contract is tracked in https://github.com/adobe/spacecat-shared/issues/1856. The same mock IS faithful on `include_metadata`, which is why the metadata half of this change is coverable today and the sort half is not.
- **mysticat-data-service** — contract only. Its migration CLI speaks the same upstream surface and already sends `include_metadata` and `prompt_id` correctly; no change is required there, and nothing in this PR alters what the CLI writes.

All three request shapes were confirmed against the live upstream API on an internal test workspace, before and after the fix:

- **The write path was already correct; the read was not.** A prompt created through the UI carries all four `created_*` / `updated_*` values upstream. Reading that same prompt with `include_metadata` on the query string returns the metadata block; reading it the way this service did returns the item with no `metadata` key at all.
- **Passing the flag in the body does not work.** The item then comes back with `metadata: null` — which reads as "this prompt is unstamped" rather than as a rejected request. Worth knowing, because it is the more misleading of the two failure modes.
- **Sort keys.** With `sort_field` / `sort_dir` the prompts come back ordered by `metadata.updated_at` as intended. With `sort` / `order` the response is byte-identical to an unsorted read.
- **Batch item key.** `model.PatchAIOPromptsBatchItem` in the vendored `/v3` contract requires `prompt_id`; `id` is not a member.

Probing the create path required writing one throwaway prompt to a test project; it was deleted afterwards and the project re-verified back at its original contents.

Operationally worth knowing, independent of this PR: prompts migrated before the authorship stamp shipped carry no metadata upstream, so they will keep reading as four nulls until a backfill runs. This fix makes that state visible rather than hiding it behind the same em dash a stamped prompt produced.

## 7. Test plan

Verified beyond unit tests: the live upstream probes in section 6 establish the corrected request shapes against the real API, which is the part unit tests cannot prove — a mocked transport is happy with either spelling. The type-level claims were checked by deliberately reintroducing each defect and confirming the strict tier now reports it, rather than accepting it as it did behind the `any` cast.

Integration coverage: the metadata half is covered by the read-path assertions landing against the Project Engine mock, which gates the inline `metadata` block on the query parameter and omits the key entirely when it is absent — so those assertions genuinely fail without this fix. The sort half is deliberately not asserted in the integration suite: the mock returns store order regardless of the keys sent, so an ordering assertion would pass with the wrong keys too and prove nothing. Ordering coverage waits on the mock work in https://github.com/adobe/spacecat-shared/issues/1856; until then the outgoing request shape is guarded by unit assertion.

**dev** — this branch was deployed to the shared `dev-branches` environment and exercised end to end there: creating a prompt through `POST /serenity/prompts` and reading it back through the list endpoint returns populated `createdAt` / `createdBy` / `updatedAt` / `updatedBy`, where the same read previously returned four nulls for a stamped prompt. The probe prompt was deleted and the slice re-verified afterwards.

**prod** — open the Prompt Library for a Serenity brand with prompts created after the authorship stamp shipped; the authorship columns should show a real user and timestamp instead of em dashes. Prompts created before it legitimately stay em-dashed. The read is additive: a prompt with no metadata upstream still yields `null` fields and the same em dash shown today, so a brand whose prompts all predate the stamp looks unchanged.

- https://github.com/adobe/spacecat-api-service/pull/2956 — related, no ordering constraint. It adds the authorship-metadata IT coverage and touches only `test/it/**`; this PR touches only `src/` and `test/support/**`. Disjoint files, both based on `main`, so they merge in either order. Between them they satisfy the metadata half of issue 2954.
- https://github.com/adobe/spacecat-shared/issues/1856 — follow-up, does not block this PR. Once the mock honours `sort_field` / `sort_dir` and ships, this repo bumps the client and the ordering coverage that issue 2954 asks for becomes writable. Shipping the transport fix ahead of that is deliberate: it corrects production behaviour now, and the mock's sort blindness means no integration test regresses either way.

No deploy sequencing is required — the change is additive on the read path and needs no coordinated release with elmo-ui.

### DIFF
--- a/src/support/serenity/rest-transport.js
+++ b/src/support/serenity/rest-transport.js
@@ -58,6 +58,9 @@ const DEFAULT_TIMEOUT_MS = 15_000;
 /** @typedef {Parameters<PeTransport['updateCompetitors']>[0]['body']} CiCompetitorsBody */
 /** @typedef {CiCompetitorsBody['ci_competitors']} CiCompetitors */
 /** @typedef {Parameters<PeTransport['updatePromptTags']>[0]['body']['items']} PromptTagUpdates */
+/** @typedef {Parameters<PeTransport['listPromptsByTagIds']>[0]['body']} PromptsListBody */
+/** @typedef {Parameters<PeTransport['patchPromptsMetadataBatch']>[0]['body']} MetadataBatchBody */
+/** @typedef {MetadataBatchBody['items'][number]['metadata']} PromptMetadataPatch */
 
 /**
  * The `resources` allocation object shared by the v2 child-create and the v2 resources
@@ -95,29 +98,6 @@ const DEFAULT_TIMEOUT_MS = 15_000;
  * argument count and argument types on every call made through it.
  *
  * @typedef {ReturnType<typeof createSerenityTransport>} SerenityTransport
- */
-
-/**
- * The Project Engine facade EXTENDED with the v3 prompt-authorship-metadata
- * write surface (LLMO-6289). WP2 re-vendors the client so the generated
- * `SerenityProjectEngineTransport` carries these operations natively; until that
- * release lands (this WP's merge gate — the dependency is bumped to the
- * WP2-released version before merge), they are declared here as a local
- * intersection so the single transport boundary type-checks under `// @ts-check`
- * against the currently-installed client. Method names/shapes track the routes
- * the revised ADR pins (spec §13 item 1):
- *   - createPromptsWithMetadata → POST  /v3 prompts        { items:[{name,metadata}], tag_ids }
- *   - patchPrompt               → PATCH /v3 prompts/{id}          { name?, metadata? }
- *   - patchPromptMetadata       → PATCH /v3 prompts/{id}/metadata { <merge-patch> }
- *   - patchPromptsMetadataBatch → PATCH /v3 prompts/metadata      { items:[{id,metadata}] }
- * WP2's generated types supersede this typedef verbatim on the dep bump.
- *
- * @typedef {ReturnType<typeof createSerenityProjectEngineTransport> & {
- *   createPromptsWithMetadata: (init: any) => Promise<any>,
- *   patchPrompt: (init: any) => Promise<any>,
- *   patchPromptMetadata: (init: any) => Promise<any>,
- *   patchPromptsMetadataBatch: (init: any) => Promise<any>,
- * }} ProjectEngineTransportWithMetadata
  */
 
 /**
@@ -392,9 +372,7 @@ export function createSerenityTransport({ env, imsToken }) {
   // client's internal.js (library defaults: maxRetries 2, retryBaseDelayMs 200;
   // per-attempt timeout via the injected `createTimeoutFetch`, since the retry
   // layer wraps it and calls it once per attempt).
-  const projects = /** @type {ProjectEngineTransportWithMetadata} */ (
-    createSerenityProjectEngineTransport(projectEngineOptions)
-  );
+  const projects = createSerenityProjectEngineTransport(projectEngineOptions);
 
   // Raw Project Engine client, ONLY for GET /v1/workspaces/{id}/brand-topics: the
   // facade does not expose a brand-topics method yet (it is in-spec — the future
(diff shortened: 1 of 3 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-scoped bug fix with thorough test coverage and correct wire-key corrections.
**Complexity: MEDIUM** - medium diff, API surface signal.
_Changes: Fixes three request-shape defects on the Semrush by_tags read and batch metadata PATCH by sending include_metadata as a query parameter, mapping sort/order onto sort_field/sort_dir, and keying batch items as prompt_id; removes the temporary ProjectEngineTransportWithMetadata typedef and WP2 merge gate (3 files)._

<details>

- nit: test name "ignores caller-supplied wire keys - only the transport builds the body" slightly oversells; it only checks three specific keys, not arbitrary injection - `test/support/serenity/rest-transport.test.js:723`
- suggestion: consider a guard test proving that passing the old `{ id }` shape to `patchPromptsMetadataBatch` does not silently forward a prompt reference (the type system now catches this at build time with zero existing callers, but a runtime assertion at the boundary would be cheap insurance against silent data loss if a future caller cargo-cults the old contract) - `test/support/serenity/rest-transport-metadata.test.js:109`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 7s | <b>Cost:</b> $5.37 | <b>Commit:</b> <code>3cead903712ddf8802c3cf60b18b40c7ea0859ed</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed both non-blocking items from the review in `a251247`.

MysticatBot:
- nit: test name oversells what it checks — done. Renamed to `does not forward caller-supplied sort_field / sort_dir — only \`sort\` maps to them`, which is the one property it actually asserts. No assertion change.
- suggestion: guard test for the old `{ id }` batch-item shape — declined. The shape cannot reach runtime: `patchPromptsMetadataBatch` is annotated `Array<{ promptId, metadata }>`, and a caller passing `{ id }` fails the **base** type-check tier, which covers all of `src/support/serenity/**` and is blocking in CI and in the pre-commit hook. Verified by compiling exactly that caller — `TS2353: Object literal may only specify known properties, and 'id' does not exist in type '{ promptId: string; metadata: ... }'`. A runtime assertion would only fire in a state the build already refuses, so per this repo's "only test cases that can actually happen" rule it would be dead weight rather than insurance. The compile-time gate is the stronger guard, and it is the one this PR added by removing the `any` cast.

rainer-friederich:
Runtime validation, on the shared `dev-branches` environment with this branch holding the slot at commit `3cead903` (slot ownership confirmed against PR HEAD before and after the checks). Auth was a stage IMS user token — the hosted dev api-service validates stage IMS.

- List read on a slice whose prompts predate the authorship stamp: 200, and all four authorship fields are present on every item rather than absent. They read `null`, which is correct there — a direct upstream read of that project shows none of its prompts carry metadata, so that slice alone cannot tell a working read from a broken one.
- Full round-trip, to obtain a genuinely stamped prompt: created one prompt through `POST /serenity/prompts`, then read it back through the list endpoint — `createdAt` / `updatedAt` populated with the creation instant, `createdBy` / `updatedBy` populated. Before this change the same read returned four nulls for a stamped prompt, which is the defect. This exercises the write and the fixed read together against the deployed branch.
- `createdBy` resolved to the `unknown` sentinel, which is expected for this auth channel and is not a finding: `resolveCallerId` falls back to `unknown` when the bearer carries no `user_id`/`sub`. Through the browser session-JWT plus promise-token path a real user id is stamped, confirmed independently by reading a UI-created prompt straight from upstream.
- Cleanup: the probe prompt was deleted via `POST /serenity/prompts/bulk-delete` (`deleted: 1`) and the slice re-verified back at its original prompt count with no probe rows.

Sort behaviour was not asserted here — the slice had no stamped prompts of its own to order. The corrected keys were verified directly against the live upstream instead: `sort_field`/`sort_dir` orders by `metadata.updated_at`, while `sort`/`order` returns an unsorted response.

The commit pushed after this run (`a251247`) is a test-name change only, with no runtime delta.

PR #2958, opened 2026-07-29
test(serenity): cover the prompt authorship metadata read path in the IT suite (LLMO-6289)

rainer-friederich:
## 1. Abstract

Integration coverage for the prompt authorship **read** path: the four values a prompt write stamps upstream are asserted to come back out through `GET /serenity/prompts`. Test-only.

## 2. Reasoning

The read path pins the outgoing request shape at the unit level — `include_metadata` on the query string, `sort_field` / `sort_dir` in the body. That proves what the service *sends*. It does not prove the four values reach the caller, which is the behaviour the feature exists for and the thing that was silently broken: the endpoint answered 200 with every authorship field null for prompts that were in fact stamped, so nothing anywhere reported an error and the UI simply rendered blanks.

The Project Engine mock gates the metadata block on the same query parameter and omits the key entirely without it, so this round trip is genuinely observable end to end rather than being a mock convenience.

A new `prompt authorship metadata on the list read (live mock)` block in the serenity integration factory:

- A stamped prompt read back through the list carries all four values populated, with `created_*` and `updated_*` matching on a fresh create and an RFC 3339 UTC timestamp shape.
- After an edit by a different caller, the read shows the editor in `updatedBy` while `createdBy` and `createdAt` still carry the original author's write — asserted across two personas, so authorship is pinned on identity rather than on clock granularity.
- A sorted read still carries the four values, covered once per sortable metadata field. This is what an unsorted read cannot cover: the metadata opt-in travels on the query string while the sort keys travel in the body, so a regression that moved the opt-in into the body alongside the sort keys would strip authorship here while leaving the plain read green.

Ordering itself is deliberately not asserted. The mock implements no sort logic, so it returns store order whichever keys are sent — an order assertion would pass unconditionally and prove nothing. That coverage stays blocked on the mock gaining sort support.

One case is knowingly left uncovered, and the test file says so in place of a silent omission: a prompt with no upstream metadata, whose read must degrade to four nulls. It is unreachable through this surface — the mock's only unstamped prompt lives in its own seeded workspace and project, while a brand's slice resolves to a project created fresh per run, so no unstamped row is ever visible to this endpoint. Reaching it would mean injecting one through the mock's `__seed` control route into the resolved project, which is harness work beyond this change; the null mapping is carried at the unit level.

## 4. Required information

- Jira / issue: LLMO-6289
- Spec: https://github.com/adobe/serenity-docs/pull/35
- Other: https://github.com/adobe/spacecat-api-service/issues/2954 (this closes its integration-coverage half) · https://github.com/adobe/spacecat-shared/issues/1856 (mock sort fidelity, blocks the ordering coverage)

- `spacecat-shared` — contract dependency, not changed here. The harness pins the Project Engine mock image to the *installed* client version, so this coverage holds only against a mock built from the client version this repo ships. A stale `node_modules` silently runs the suite against an older mock, so run `npm install` before reproducing.

Verified against the live integration stack (PostgreSQL, PostgREST, MinIO and both Semrush vendor mock containers over real HTTPS).

The new cases were checked for sensitivity rather than merely observed passing, by reverting `src/support/serenity/rest-transport.js` to its pre-fix state and re-running:

- **With the read-path fix in place** — they pass.
- **Without it** — they fail, on the exact production symptom: `expected null to be a string` for the timestamps and `expected null to equal 'test-admin@adobe.com'` for the author.

The source file is byte-identical to `main`, so the diff here is test-only.

## 7. Test plan

**Verified locally.** The serenity integration suite is green on this branch, and the new cases were additionally run in isolation both with and against the fix as described above.

```
npx mocha --require test/it/postgres/harness.js --timeout 60000 test/it/postgres/serenity.test.js
```

The `60000` timeout is required — this suite drives the vendor mocks over real HTTPS. Docker plus an ECR login is needed for the private PostgREST image; see `test/it/README.md`.

**Per-environment verification.** None applies: this PR adds no runtime code, so there is nothing to observe on eph, dev, stage or prod. CI running the integration suite is the complete signal.

- https://github.com/adobe/spacecat-api-service/pull/2956 — **related**, no dependency. It covers the write path against the mock's store snapshot and touches the same test file, so whichever of the two lands second needs a trivial conflict resolution: both append a describe block at the end of `test/it/shared/tests/serenity.js`.

No other ordering constraint applies — this PR can merge on its own.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,45 @@
+## [1.708.2](https://github.com/adobe/spacecat-api-service/compare/v1.708.1...v1.708.2) (2026-07-30)
+
+
+### Bug Fixes
+
+* **sites:** restrict provisioning-managed LLMO config fields on site update ([#2961](https://github.com/adobe/spacecat-api-service/issues/2961)) ([be57915](https://github.com/adobe/spacecat-api-service/commit/be57915e37e9c7152315906be72cf1181c7de7cb))
+
+## [1.708.1](https://github.com/adobe/spacecat-api-service/compare/v1.708.0...v1.708.1) (2026-07-30)
+
+
+### Bug Fixes
+
+* **user-details:** resolve the caller's own profile without the admin gate (LLMO-6290) ([a059943](https://github.com/adobe/spacecat-api-service/commit/a059943d7a5c555175db99353480cc790c421dbb))
+
+# [1.708.0](https://github.com/adobe/spacecat-api-service/compare/v1.707.0...v1.708.0) (2026-07-30)
+
+
+### Features
+
+* **llmo:** FACS-capability auth for edge CDN routing; edge-optimize-config under can_configure ([#2947](https://github.com/adobe/spacecat-api-service/issues/2947)) ([b4e6c32](https://github.com/adobe/spacecat-api-service/commit/b4e6c329d27743b52f628345bf3018a7261511fc))
+
+# [1.707.0](https://github.com/adobe/spacecat-api-service/compare/v1.706.0...v1.707.0) (2026-07-30)
+
+
+### Features
+
+* **serenity:** server-owned source dimension + canonicalizeSource (LLMO-6282) [WP-S2] ([#2867](https://github.com/adobe/spacecat-api-service/issues/2867)) ([7a58e4f](https://github.com/adobe/spacecat-api-service/commit/7a58e4f101c132c332cd22ac92667fd5c52a828b))
+
+# [1.706.0](https://github.com/adobe/spacecat-api-service/compare/v1.705.1...v1.706.0) (2026-07-30)
+
+
+### Features
+
+* **serenity:** strict `origin` root resolution — drop tolerant `source` fallback (LLMO-6280) [WP-O6] ([#2917](https://github.com/adobe/spacecat-api-service/issues/2917)) ([5bdb7b0](https://github.com/adobe/spacecat-api-service/commit/5bdb7b0e740fd382e3f8a998e78937c1095e727f)), closes [#2867](https://github.com/adobe/spacecat-api-service/issues/2867) [#2867](https://github.com/adobe/spacecat-api-service/issues/2867)
+
+## [1.705.1](https://github.com/adobe/spacecat-api-service/compare/v1.705.0...v1.705.1) (2026-07-29)
+
+
+### Bug Fixes
+
+* **serenity:** request prompt metadata and use the real sort keys on by_tags (LLMO-6289) ([#2957](https://github.com/adobe/spacecat-api-service/issues/2957)) ([4cc7026](https://github.com/adobe/spacecat-api-service/commit/4cc7026a2d09e3aba310a4f0c95ecd514d6ea476)), closes [Hi#level](https://github.com/Hi/issues/level)
+
 # [1.705.0](https://github.com/adobe/spacecat-api-service/compare/v1.704.2...v1.705.0) (2026-07-29)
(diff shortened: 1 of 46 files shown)


MysticatBot (approved):
@/tmp/mysticat-workspace-p8ajfsh_/review-body-2958.txt

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-scoped integration tests that pin the authorship metadata read path.
**Complexity: MEDIUM** - 133 lines, single test file.
_Changes: Adds three integration tests verifying that prompt authorship metadata (createdAt, updatedAt, createdBy, updatedBy) flows correctly through GET /serenity/prompts, covering basic read, multi-persona edit attribution, and sorted read with metadata opt-in (1 file)._

Note: CI build is still pending - verify it passes before merge.

<details>

- suggestion: `createUsMarket` does not assert its response status - if market creation silently fails (e.g. returns 409), downstream tests pass or fail for the wrong reason. Adding `expect(res.status).to.equal(201)` like the other helpers would catch this early. - `test/it/shared/tests/serenity.js:930`
- nit: The `for...of` loop in the sorted-read test masks the second iteration if the first fails. Consider unrolling into two explicit assertion blocks (also removes the eslint-disable comment). - `test/it/shared/tests/serenity.js:1037`
- suggestion: `readPromptById` could include the actual list contents in its failure message for easier debugging (e.g. `expected prompt ${id} in list [${items.map(...)}]`). - `test/it/shared/tests/serenity.js:955`
- nit: `US_GEO = 2840` is a magic number with no comment explaining its origin (UN M49 numeric code for the United States). A brief inline note would help future readers. - `test/it/shared/tests/serenity.js:923`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 7s | <b>Cost:</b> $3.24 | <b>Commit:</b> <code>f71d838de8b016af46124514b3072b1ae7295876</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed all four non-blocking items in `2d7f24745`.

- **suggestion: `createUsMarket` does not assert its response status** — done. Every test in the block needs a live US market and none asserts on the market response, so the check now lives inside the helper. Worth noting the concrete failure mode is a 422 on insufficient units rather than a 409: a refused create now fails at the market call instead of surfacing as a puzzling status mismatch on the tag or prompt request that depended on it.

- **nit: the `for...of` loop masks the second iteration if the first fails** — done, though not by unrolling. Two assertion blocks inside one test would not have fixed it: mocha aborts a test at the first failed assertion either way, so the second block would still never run. Each sortable metadata field now gets its own case, which genuinely reports a regression that reaches only one of them, and drops the `eslint-disable` along the way.

- **suggestion: `readPromptById` could include the actual list contents in its failure message** — done. A miss now names the ids the read did return, which separates "the list came back empty" from "the ids drifted"; a bare presence check reported only that `undefined` does not exist.

- **nit: `US_GEO = 2840` is a magic number with no comment** — done, with a correction to the stated origin. 2840 is not the UN M49 numeric code for the United States (that is 840) — it is Google's geo target id, which is what the upstream API expects. The comment records it as such, matching the wording the other blocks in this file already use.

Re-verified against the live integration stack after the changes: the serenity integration suite is green, and all four read-path cases pass. Lint and type-check clean; the diff remains test-only.

On the note about CI: checks are re-running on this commit; the cycle is not closed until they land green.

PR #2960, opened 2026-07-30
fix(user-details): resolve the caller's own profile without the admin gate (LLMO-6290)

rainer-friederich:
## 1. Abstract
`/organizations/{organizationId}/userDetails` (single and bulk) now resolves the **caller's own** details from their auth profile, without an IMS call and without the admin gate. Everyone else's details stay admin-gated, and the claim that identifies "who is calling" is now read in exactly one place, shared with the server-owned authorship stamp.

## 2. Reasoning
A prompt created through the Serenity prompt library in prod rendered its author as an unresolved member ("Unknown member", empty email) minutes after the authorship-metadata feature went live. The stamp itself was correct: the create was JWT-authenticated and `created_by` / `updated_by` carried the author's real IMS user id — confirmed in prod CloudWatch, where the table's `userDetails` lookup asked for exactly that id and answered `User is not admin, returning system defaults`.

The gap is on the read side. `userDetails` resolves a display name only for a `TrialUser` row or an admin requestor; every other caller gets the `system` placeholder, which the UI renders as an unresolved member. So a non-admin enterprise user cannot see their own name — not even on content they authored themselves seconds earlier, where the stamped id *is* their own id.

Returning a caller their own claims discloses nothing new: it is the identity they authenticated with. Resolving *other* users stays admin-only, which is deliberate — the id is caller-supplied, so an ungated lookup would let any authenticated caller resolve arbitrary IMS ids to names and emails.

Behaviour delta on `GET /organizations/{organizationId}/userDetails/{externalUserId}` and `POST /organizations/{organizationId}/userDetails`:

- **Before:** a requested id with no `TrialUser` row resolved to a real profile only for an admin requestor (via the IMS admin profile lookup). Everyone else — including a caller asking about themselves — received `{ firstName: 'system', lastName: '-', email: '' }`.
- **After:** if the requested id is the caller's own, the response is built from their own auth-profile claims — no IMS call, no admin gate. Any other id keeps the previous behaviour exactly: `TrialUser` row, else admin IMS lookup, else the `system` placeholder.

Response shape is unchanged, so the UI's existing resolution path needs no change: a freshly authored record now shows the author's name and email instead of an unresolved member.

Three details worth reviewing deliberately:

- **One reader for the caller identity.** `resolveCallerImsUserId` (`src/support/utils.js`) is now the single definition of the claim precedence: `user_id`, else `sub`, else `email`. The authorship stamp resolves the id it *writes* through it, and this endpoint resolves the caller's own id through it as well. Previously the two sides each read the claims independently — which is precisely how a stamp can keep working while only the display degrades, silently.
- **`email` is an id here, not an address** — and both facts matter in the same file. The auth service deliberately puts the IMS user id (not the real address) in the JWT `email` claim, and `AdobeImsHandler` deletes `user_id` from the profile it builds, leaving `email` the only carrier for an IMS-authenticated caller; this is the same claim the pre-existing writers of `prompts.updated_by` read, and prod rows carry ids of exactly that shape. So `email` participates in resolving the **id**, while the **email field** of the response is sourced only from `trial_email` / `preferred_username` and returns empty rather than putting an id in an email column.
- **The IMS-call counter now counts IMS calls.** The bulk path's counter — which feeds the "Fetched user details from IMS N times" log used to reason about IMS volume — incremented for every non-TrialUser id, including the non-admin placeholder path that never calls IMS (pre-existing) and now the self path. `fetchNonTrialUserDetails` reports which path answered, and only a real IMS call is counted (a call that threw still counts; it consumed capacity). An existing test had frozen the inflated behaviour in place while simultaneously asserting IMS was never called — that assertion was corrected rather than worked around.

## 4. Required information
- Jira / issue: LLMO-6290 (prompt authorship metadata — display/resolution side), follow-up to LLMO-6289 (server-owned stamp + list read)
- Spec: https://github.com/adobe/serenity-docs/pull/35 — `docs/prompt-management/prompt-authorship-metadata.md` (author = one opaque id per side, resolved to a display name at read time via `userDetails`)

- **project-elmo-ui** — consumer, unchanged. Its `userUtils` resolution and the Serenity prompt-library authorship columns are the caller of this endpoint; they benefit without a code change (the response shape is identical). Note the label distinction it makes: the literal `unknown` sentinel renders as "Unknown", whereas an unresolved real enterprise-org id renders as "Unknown member" — the symptom here was the latter.
- **spacecat-auth-service** — contract, unchanged. It mints the SpaceCat session JWT whose `sub` and `email` claims both carry the IMS user id this endpoint now matches against; a change to those claims would break both the stamp and this resolution together.

Verified in prod (CloudWatch `/aws/lambda/spacecat-services--api-service`, profile `spacecat-prod`, `us-east-1`) rather than inferred:

- The prompt create at `2026-07-29T20:36:07Z` on org `cb9df376-a5d8-47e3-94f5-d9687a376eed` / brand `019f37d0-dfda-79e1-9e6c-e932f0f805c5` logged `Authenticated with jwt` — so authorship resolved from the session JWT, and the promise token was used only for the upstream Semrush hop.
- The prompt-library render at `2026-07-30T04:59:04Z` issued `POST /organizations/cb9df376…/userDetails` for exactly one id — a real enterprise-org IMS user id — and logged `User is not admin, returning system defaults for <id>`. That is the failure this PR removes, and it also rules out the `unknown` sentinel as the cause: the UI filters that sentinel out before the lookup, so a sentinel would never have reached the endpoint.

Useful for future triage: DEBUG lines are emitted in prod, so `filter-log-events` on `returning system defaults` shows precisely which ids a lookup could not resolve.

The claim shape this endpoint depends on was confirmed against a live production session token, not only against the minting code: the payload carries no `user_id` at all, the IMS user id in both `sub` and `email`, the real address in `trial_email` / `preferred_username`, and populated `first_name` / `last_name`. That is what makes the fallback order safe and the response's email field correct.

The claim precedence was also checked against what the rest of the service already writes, rather than assumed: prod `prompts.updated_by` holds IMS user ids in the same `<hex>@<hex>.e` shape (written from `profile.email` by the v2 prompts path), while `prompts.created_by` has never carried a user id at all — every recent row is null, the only non-null values being batch markers. So Serenity authorship writes into an identifier namespace that already exists in prod, which is what makes `userDetails` the right resolver for it.

## 7. Test plan
- Locally: exercised through the controller, including the round trip that matters in production — the id the authorship stamp resolves for a given caller is fed straight into the `userDetails` read path for the same caller and must come back as that caller's name and email. Also covered: bulk requests mixing the caller's own id with another user's (own resolved, other still gated), the IMS-profile claim shape, and the case where the only address-looking claim is the IMS GUID (email must stay empty).
- Integration: both endpoints now run against the real auth stack and PostgreSQL, so the claim precedence is proven against a signed-and-validated token instead of a hand-built profile object. This required a new `enterpriseUser` persona — every pre-existing persona uses an address as `sub`, so none of them could exercise the id-vs-address distinction the self-resolution path depends on. Writing it surfaced two things a unit test structurally cannot see: the router does not URL-decode path params (so a client that percent-encodes the `@` in an IMS id silently gets the placeholder — the UI is unaffected, it calls the bulk endpoint), and `TRIAL_USER_1` was seeded without the `external_user_id` the lookup matches on, so the trial branch could never resolve it and every request fell through to the non-trial path.
- dev: open the Serenity prompt library as a **non-admin** user on a sub-workspace brand, create a prompt, and confirm the Updated by / Email columns show that user's own name and email instead of an unresolved member. Then confirm a row authored by a *different* org member without a trial-user record still shows the unresolved label for that same non-admin viewer — the gate for other users is intentionally unchanged.
- prod: after deploy, repeat the create-and-read check on the LLMO-Dev-2 brand used above. Absence of `returning system defaults` for the viewer's own id in the api-service log group is the signal.

### DIFF
--- a/src/controllers/user-details.js
+++ b/src/controllers/user-details.js
@@ -71,11 +71,13 @@ function UserDetailsController(ctx) {
    * Returning the caller's own claims discloses nothing new: it is the identity
    * they authenticated with. Other users' details stay admin-gated.
    *
-   * `profile.email` is deliberately NOT a source for the email field — on both
-   * the JWT and the IMS profile that claim carries the IMS user GUID rather than
-   * an RFC-5322 address (same reason llmo-akamai.js `getCallerEmail` prefers
-   * `trial_email` / `preferred_username`), and a GUID in an email column is worse
-   * than an empty one.
+   * `profile.email` is deliberately NOT a source for the EMAIL field — on both
+   * the JWT and the IMS profile that claim carries the IMS user id rather than an
+   * RFC-5322 address (same reason llmo-akamai.js `getCallerEmail` prefers
+   * `trial_email` / `preferred_username`), and an id in an email column is worse
+   * than an empty one. It IS a source for the id itself, inside
+   * {@link resolveCallerImsUserId} — the two uses pull opposite ways on the same
+   * claim, which is exactly why the id resolution lives in one shared place.
    *
    * @param {string} externalUserId - The requested external user ID.
    * @param {string} organizationId - The organization ID the request addresses.
@@ -88,6 +90,9 @@ function UserDetailsController(ctx) {
       return null;
     }
 
+    // Invariant: a non-null id means the profile chain resolved, so this second
+    // traversal cannot throw — no optional chaining needed (unlike the defensive
+    // shared helper above, which is reached with an arbitrary context).
     const profile = ctx.attributes.authInfo.getProfile();
     const email = [profile.trial_email, profile.preferred_username].find((v) => hasText(v));
     return {
@@ -102,25 +107,37 @@ function UserDetailsController(ctx) {
    * Resolves the details of a user who has no TrialUser row: the caller's own
    * identity comes from their auth profile, anyone else's from IMS and only for
    * an admin requestor.
+   *
+   * Reports which of the three paths answered alongside the details, so the bulk
+   * caller can count IMS calls it actually made. `source: 'ims'` means the
+   * upstream call was issued — including when it threw, since a failed call still
+   * consumed IMS capacity; the self and placeholder paths issue none. That count
+   * is the operational signal for IMS volume, so it must not absorb the two paths
+   * that never leave the process.
+   *
    * @param {string} externalUserId - The external user ID to resolve.
    * @param {string} organizationId - The organization ID for fallback.
-   * @returns {Promise<Object>} User details object.
+   * @returns {Promise<{details: Object, source: 'self'|'placeholder'|'ims'}>}
+   *   The user details and which path produced them.
    */
   const fetchNonTrialUserDetails = async (externalUserId, organizationId) => {
     const own = resolveCallerOwnDetails(externalUserId, organizationId);
     if (own) {
       log.debug(`Resolved the caller's own details from the auth profile for ${externalUserId}`);
-      return own;
+      return { details: own, source: 'self' };
     }
 
     // Check if requestor has admin access
     if (!accessControlUtil.hasAdminReadAccess()) {
       log.debug(`User is not admin, returning system defaults for ${externalUserId}`);
       return {
-        firstName: 'system',
-        lastName: '-',
-        email: '',
-        organizationId,
+        details: {
+          firstName: 'system',
+          lastName: '-',
+          email: '',
+          organizationId,
+        },
+        source: 'placeholder',
       };
     }
 
@@ -129,16 +146,22 @@ function UserDetailsController(ctx) {
       log.debug(`Admin user requesting details for ${externalUserId}, attempting IMS fallback`);
       const imsProfile = await imsClient.getImsAdminProfile(externalUserId);
       return {
-        ...toProfileShape(imsProfile),
-        organizationId,
+        details: {
+          ...toProfileShape(imsProfile),
+          organizationId,
+        },
+        source: 'ims',
       };
     } catch (error) {
       log.warn(`Failed to fetch user details from IMS for ${externalUserId}: ${error.message}`);
       return {
-        firstName: '-',
-        lastName: '-',
-        email: '',
-        organizationId,
+        details: {
+          firstName: '-',
+          lastName: '-',
+          email: '',
+          organizationId,
+        },
+        source: 'ims',
       };
     }
   };
@@ -186,7 +209,10 @@ function UserDetailsController(ctx) {
         };
       } else {
         // User not found in trial users - own profile, else IMS if admin
-        userDetails = await fetchNonTrialUserDetails(externalUserId, organizationId);
+        ({ details: userDetails } = await fetchNonTrialUserDetails(
+          externalUserId,
+          organizationId,
+        ));
       }
 
       return ok(userDetails);
@@ -245,9 +271,14 @@ function UserDetailsController(ctx) {
           };
         } else {
           // User not found in trial users - own profile, else IMS if admin
-          imsCallCount += 1;
           // eslint-disable-next-line no-await-in-loop
-          const details = await fetchNonTrialUserDetails(externalUserId, organizationId);
+          const { details, source } = await fetchNonTrialUserDetails(
+            externalUserId,
+            organizationId,
+          );
+          if (source === 'ims') {
+            imsCallCount += 1;
+          }
           userDetailsMap[externalUserId] = details;
         }
       }
--- a/src/controllers/user-details.js
+++ b/src/controllers/user-details.js
@@ -24,6 +24,7 @@ import {
 } from '@adobe/spacecat-shared-utils';
 
 import AccessControlUtil from '../support/access-control-util.js';
+import { resolveCallerImsUserId } from '../support/utils.js';
 
 // IMS GUID format: <hex>@<alphanumeric-with-dots>
 const IMS_USER_ID_RE = /^[A-Za-z0-9]+@[A-Za-z0-9.]+$/;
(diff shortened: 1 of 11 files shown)


src/controllers/user-details.js
```diff
@@ -194,10 +244,10 @@ function UserDetailsController(ctx) {
             organizationId: trialUser.getOrganizationId(),
           };
         } else {
-          // User not found in trial users - try IMS if admin
+          // User not found in trial users - own profile, else IMS if admin
           imsCallCount += 1;
           // eslint-disable-next-line no-await-in-loop
```
MysticatBot, line 274: issue (blocking): `imsCallCount` is incremented unconditionally when a user has no TrialUser row, but `fetchNonTrialUserDetails` now short-circuits via `resolveCallerOwnDetails` without making any IMS call. The subsequent log message `Fetched user details from IMS ${imsCallCount} times` overstates actual IMS usage, which degrades the diagnostic value of this counter for capacity reasoning and incident triage.

**Why it matters:** This counter is the operational signal for IMS call volume in the bulk path. Inflated counts make it harder to correlate rate-limit events or latency spikes with real IMS traffic.

**Fix:** Either rename to `nonTrialLookupCount` (with a matching log message), or move the increment inside `fetchNonTrialUserDetails` conditioned on the IMS branch actually executing. The simplest option: increment only when `resolveCallerOwnDetails` returns null (i.e. the IMS path will run). Or more cleanly, have `fetchNonTrialUserDetails` return a `{ details, source }` tuple and increment only when `source === "ims"`.

MysticatBot (changes requested):
Hey @rainer-friederich,

**Verdict: Request changes** - one observability regression in the bulk path.
**Complexity: HIGH** - medium diff touching an API surface (controllers).
_Changes: Adds caller self-resolution to the userDetails endpoint so non-admin users can see their own name/email from their auth profile, fixing the unresolved-member display bug (5 files)._

## Must fix before merge
1. [Important] `imsCallCount` increments for self-resolved profiles that never call IMS - `src/controllers/user-details.js:249` (details inline)

<details>

- nit: `resolveCallerOwnDetails` calls `ctx.attributes.authInfo.getProfile()` a second time after `resolveCallerImsUserId` already traversed the same chain - safe (profile is proven to exist at that point), but the asymmetry with the defensive style in the utility is worth a one-line comment noting the invariant - `src/controllers/user-details.js:72`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 52s | <b>Cost:</b> $8.67 | <b>Commit:</b> <code>c19a504c1ad034777a42b59d095a4a1e775f6811</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the review in `d021a1682`.

MysticatBot:
- [Important] `imsCallCount` inflated by self-resolved profiles — done. Valid, and it predates this PR: the counter also inflated on the non-admin placeholder path, and an existing test had frozen that in place (it asserted the log claimed one IMS fetch while asserting IMS was never called). `fetchNonTrialUserDetails` now returns `{ details, source }` and the bulk path counts only `source === 'ims'` — including a call that threw, since a failed call still consumed IMS capacity. The stale assertion was corrected to expect silence when nothing called IMS.
- nit: second `getProfile()` traversal without the defensive chaining — done. Comment added stating the invariant (a non-null id proves the chain resolved), so the asymmetry with the shared helper reads as deliberate.

One change beyond the review, from verifying the approach against what the rest of the service already does:

- `resolveCallerImsUserId` now falls back to the `email` claim after `user_id` / `sub`. The three are carriers of the same value, and which ones are populated depends on the auth handler: `AdobeImsHandler` deletes `user_id` from the profile it builds and leaves the id only on `email`, so an IMS-authenticated caller previously resolved to nothing. `email` is not a human address on this platform — the auth service deliberately puts the IMS user id in that claim (the ASO-607 rationale in `support/api-key-ims-handler.js`), which is why the pre-existing writers of `prompts.updated_by` read it directly. Prod bears that out: `prompts.updated_by` holds ids of exactly this shape, written via `profile.email`. For a JWT caller all three claims are the same string, so nothing changes for the UI path; it only stops raw-IMS callers from degrading to an unresolved author.

src/controllers/user-details.js
rainer-friederich: Fixed in `d021a1682` — `fetchNonTrialUserDetails` now returns `{ details, source }` and the bulk path counts only `source === "ims"` (a call that threw still counts; the self and placeholder paths never leave the process). Worth noting the counter already inflated on the non-admin placeholder path before this PR, and an existing test had frozen that behaviour — that assertion was corrected too.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - prior finding addressed, no new blocking issues.
_Changes: Adds caller self-resolution to the userDetails endpoint so non-admin users can see their own name/email from their auth profile, and centralizes caller-identity claim precedence into a single shared helper (5 files)._

<details>

- nit: The `===` comparison between `callerUserId` and `externalUserId` is case-sensitive. IMS ids are stored canonically so this is safe today, but a normalizing `.toLowerCase()` on both sides would harden against a future transport that changes case - `src/controllers/user-details.js:64`
- suggestion: If a future auth handler ever populates `profile.email` with a real RFC-5322 address rather than an IMS GUID, the authorship stamp would silently start writing email addresses into `createdBy`/`updatedBy`. A runtime guard (e.g. asserting the resolved id matches `IMS_USER_ID_RE` at the stamp site) would catch such a regression early - `src/support/utils.js:815`
</details>

- IMS-call counter now increments only when `source === 'ims'`, fixing the inflated volume log.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 4m 22s | <b>Commit:</b> <code>d021a16828876c9c153c690fec96bb6826531b6c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Round 3 — both non-blocking items triaged, plus integration coverage that was previously called out as missing.

**nit: case-insensitive id comparison** — declined. Normalizing both sides widens what counts as "you" on a path that decides whether to hand back a profile without the admin gate. IMS ids are canonical, so a `.toLowerCase()` cannot fix a real mismatch today; it can only make two ids that are *not* equal compare equal. On an authorization-adjacent comparison the conservative direction is exact match, so the hardening would point the wrong way.

**suggestion: assert the resolved id matches `IMS_USER_ID_RE` at the stamp site** — declined, because the guard would misfire on ids that are legitimately not IMS GUIDs. `LegacyApiKeyHandler` builds a profile of `{ user_id: 'admin' | 'legacy-user' }`, and `ScopedApiKeyHandler`'s profile is an ApiKey entity — neither matches the pattern. A regex gate would convert those real, working callers into the `unknown` sentinel, losing attribution the stamp records correctly today. The premise is also narrower than it looks: `email` carrying the user id is a deliberate, documented contract of the auth service (ASO-607), not an accident waiting to drift — and a change to it would break `sub` in the same commit, which the round-trip test already covers.

**Integration coverage added** (was disclosed as a gap in the previous round; Docker became available):

- Both endpoints now run against the real auth stack and PostgreSQL, so claim precedence is proven against a signed-and-validated token rather than a hand-built profile object.
- Required a new `enterpriseUser` persona: every pre-existing persona uses an address as `sub`, so none could exercise the id-vs-address distinction this path turns on.
- Two things a unit test structurally cannot see, both found by running it: the router does not URL-decode path params, so a client percent-encoding the `@` in an IMS id silently receives the placeholder (the UI is unaffected — it calls the bulk endpoint); and `TRIAL_USER_1` was seeded without the `external_user_id` the lookup matches on, so the trial branch could never resolve it and every request fell through to the non-trial path. The seed is fixed here; the pre-existing trial-users suite still passes against it.

The claim shape was additionally confirmed against a live production session token rather than only the minting code: no `user_id` present, the IMS id in both `sub` and `email`, the real address in `trial_email` / `preferred_username`, and populated name claims.

The earlier `ci / build` red was an external TLS reset (`ECONNRESET` to `bundles.aem.page`) in a `SerenityController` before-each hook — 16346 passing, 1 failing, unrelated to this diff. Superseded by the run on the new commit rather than papered over.

rainer-friederich:
## Runtime validation — dev

Executed against the `dev-branches` deployment of `d5734f651` (this branch's head after merging main), log group `/aws/lambda/spacecat-services--api-service`, account 682033462621.

The same request was issued twice against the same caller and org, straddling the deploy — so the two builds are directly comparable rather than merely described:

- `06:54:34Z` — previous build: `Admin user requesting details for 1AC57EC8…@9bcd5ee5…, attempting IMS fallback`
- `07:02:56Z` — this build: `Resolved the caller's own details from the auth profile for 1AC57EC8…@9bcd5ee5…`, with **no** IMS-fallback line for that id

Response body on the new build, via the bulk endpoint the prompt library actually calls:

```
{"1AC57EC8…@9bcd5ee5…":{"firstName":"Rainer","lastName":"Friederich",
                        "email":"rfriederich@adobe.com","organizationId":"f2c31686-…"}}
```

Two details worth stating precisely, because each was initially misread here:

- **The response alone does not prove the fix.** The dev caller is `is_admin: true`, and the previous build already resolved a name for an admin through the IMS lookup. An identical-looking body was in fact served by the old path at 06:54. Only the absence of the IMS-fallback line together with the presence of the self-resolution line distinguishes them — which is why the log, not the payload, is the evidence cited above.
- **`dev-branches` is a single shared slot.** The first probe ran while the slot still held the previous build; the deployment record advanced to `d5734f651` at `06:54:57Z` and the Lambda alias followed at `06:56:23Z`. The validating probe is the `07:02:56Z` one, after both. Deployment SHA was re-confirmed either side of it.

The claim shape the endpoint depends on was separately confirmed against a live production session token: no `user_id` present, the IMS user id in both `sub` and `email`, the real address only in `trial_email` / `preferred_username`, and populated name claims — so `email` correctly participates in resolving the id while the response's email field carries the address.

Not covered by this run: a **non-admin** caller on dev, since no non-admin identity was available for that environment. That path is covered by the integration suite, which drives an `is_admin: false` persona through the real auth stack and PostgreSQL and is green in CI.

PR #2961, opened 2026-07-30
fix(sites): restrict provisioning-managed LLMO config fields on site update

akshaymagapu:
## What

Adds an access-control check on the full-site update endpoint (`PATCH /sites/:siteId`) so that only privileged callers may set or change the provisioning-managed LLMO configuration fields. Non-privileged callers keep whatever is currently stored (or none, if unset); privileged callers are unaffected.

## Why

These fields are owned by the provisioning flow and should not be writable by ordinary callers through the general site-update path. This aligns that path with the dedicated LLMO config endpoints, which already enforce the same privilege level.

## Changes

- New helper `src/support/llmo-config-guards.js` that returns the sanitized config for a caller.
- Wired into `updateSite` after the existing config merge.
- Unit tests for the helper and controller tests covering the privileged and non-privileged paths.

## Testing

- `npm test` (unit + controller); lint clean.

### DIFF
--- a/src/controllers/sites.js
+++ b/src/controllers/sites.js
@@ -65,6 +65,7 @@ import { listViewableResourceIds } from '../support/state-access-mapping-utils.j
 import { requirePostgrestForFacsMappings } from '../support/postgrest-availability.js';
 import { isFacsRebacResource } from '../routes/facs-capabilities.js';
 import { ASO_PRODUCT_CODE, STATUSES as PLG_STATUSES } from './plg/plg-onboarding/constants.js';
+import { guardProvisioningLlmoFields } from '../support/llmo-config-guards.js';
 
 const VALIDATION_ERROR_NAME = 'ValidationError';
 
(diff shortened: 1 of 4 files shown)


MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - clean, well-scoped access-control guard with solid test coverage.
_Changes: Adds a guard that prevents non-privileged callers from writing provisioning-managed LLMO config fields (cdnlogsFilter, cdnBucketConfig) on the general site-update endpoint (4 files)._

<details>

- suggestion: Verify what happens when a non-privileged caller sends `config: { llmo: null }` - the `if (merged.llmo)` check is falsy for null, so the guard is skipped. If the downstream Config model accepts null, provisioning fields could be wiped via this pre-existing path. Consider adding a test to document the intended behavior. - `src/controllers/sites.js:1135`
- suggestion: Consider defensive cloning when restoring stored values (`sanitized[field] = structuredClone(stored)`) since this is an access-control boundary and the stored objects are assigned by reference. Low risk given the object goes straight to setConfig, but a cheap hardening. - `src/support/llmo-config-guards.js:39`
- nit: Controller integration tests only exercise `cdnlogsFilter` for non-privileged callers. Adding a parallel test for `cdnBucketConfig` would verify wiring for both guarded fields end-to-end. - `test/controllers/sites.test.js`
- suggestion: A debug-level log at the call site when the guard strips a provisioning field would help integrators diagnose "my update didn't stick" without requiring log-level forensics. - `src/controllers/sites.js:1135`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 4s | <b>Cost:</b> $4.00 | <b>Commit:</b> <code>c3d273ef4b06b8d4290d2b0dd9956a5aa78cb590</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2924, opened 2026-07-28
feat(brand-claims): add run-brand-claims Slack command (LLMO-6143)

akshaymagapu:
## What

New Slack command `run-brand-claims {baseURL|siteId}` — forces an on-demand Brand Claims run for a site:

1. Resolves the site (base URL or UUID) and its brand.
2. Refuses if `brandClaimsEnabled` is off (points at `enable-brand-claims {brandId}`).
3. Looks up the site's actual latest Brand Presence sheet in the DRS-owned `drs-v2-prod-bp` bucket.
4. Publishes the DRS-shaped `BRAND_PRESENCE_SHEET_WRITTEN` event onto `mysticat-bp-sheet-ready`.
5. Replies with confirmation naming the exact `s3_key`, week/year, and cadence it referenced.

## Why a real S3 lookup, not a constructed key

A hand-built event with a guessed/constructed `s3_key` would be silently rejected — Mystique's Brand Claims consumer re-lists S3 itself as a freshness guard and requires the event's key to be the true latest object for that site+platform. So this does a real listing under `{siteId}/{brandSlug}/analytics/chatgpt_free/` and picks the same "latest" the consumer's own `discover_latest_bp_object` would (max by S3 date partition, then `LastModified`), rather than guessing today's date.

`sanitizePathComponent` is a byte-for-byte port of DRS's own `sanitize_path_component` — the consumer's freshness re-list uses the same `brand` value to find the sheet, so this has to match exactly (verified against DRS's actual regex, not its docstring, which has a stale example).

## Dependencies

- **`spacecat-infrastructure#709`** — grants `spacecat-role-lambda-generic` (this service's execution role) `s3:ListBucket`/`s3:GetObject` on `drs-v2-prod-bp`. Without it, this command's S3 lookup will fail with access denied.
- **New env var `SQS_BP_SHEET_READY_QUEUE_URL`** needs to be added to this service's deployment config, pointing at the `mysticat-bp-sheet-ready` queue (created by `spacecat-infrastructure#674`, already merged/deployed). The command checks for this and errors clearly if unset, rather than failing deep in execution.

## Testing

18 new unit tests (site resolution by URL/UUID, enablement gate, S3 pagination, weekly/daily cadence detection from filename, ignoring non-sheet keys like experiment runs, sanitization edge cases including the non-ASCII hash fallback, and all error paths). Full existing Slack command suite (908 tests) still passes. Lint clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2962, opened 2026-07-30
fix(claims): run-brand-claims sends the spacecat org UUID, not the IMS org (LLMO-6143)

akshaymagapu:
## What
`run-brand-claims` (#2924) publishes the `BRAND_PRESENCE_SHEET_WRITTEN` event with `organization_id: organization.getImsOrgId()`. The mystique Brand Claims consumer resolves the brand via `GET /v2/orgs/{spaceCatId}/sites/{siteId}/brand`, whose org path param is **validated as a UUID** (`brands.js` `getBrandForOrgSite` → `isValidUUID`). An IMS org id (`…@AdobeOrg`) is not a UUID → **400** → the consumer's enablement gate classifies the 4xx as resolved-negative and **silently drops the event** (`enablement_dropped`). Net: the command reports success but **no claims scan ever runs**.

## Fix
Send `site.getOrganizationId()` (the spacecat org UUID) instead, and drop the now-unused `Organization.findById` lookup + IMS-org guard.

## Validation
- Verified empirically: a manual event with the spacecat org UUID triggers a claims scan end-to-end (Changi, dev, scan `cd087d57`), while the IMS-org variant was dropped.
- `mocha test/support/slack/commands/run-brand-claims.test.js` — 21 passing (updated the published-event assertion to `ORG_ID`, removed the obsolete IMS-org test).
- `eslint` clean.

### DIFF
--- a/src/support/slack/commands/run-brand-claims.js
+++ b/src/support/slack/commands/run-brand-claims.js
@@ -211,11 +211,6 @@ function RunBrandClaimsCommand(context) {
       const event = {
         event_type: 'BRAND_PRESENCE_SHEET_WRITTEN',
         schema_version: 1,
-        // The Brand Claims consumer resolves the brand via
-        // `GET /v2/orgs/{spaceCatId}/sites/{siteId}/brand`, whose path param is the
-        // SpaceCat org UUID (validated `isValidUUID`) — NOT the IMS org id. Sending
-        // `getImsOrgId()` (…@AdobeOrg) 400s there and the event is silently
-        // enablement-dropped. Use the spacecat org UUID. (LLMO-6143)
         organization_id: site.getOrganizationId(),
         brand_id: brand.id,
         brand: brandSlug,
--- a/src/support/slack/commands/run-brand-claims.js
+++ b/src/support/slack/commands/run-brand-claims.js
@@ -133,7 +133,7 @@ function RunBrandClaimsCommand(context) {
   });
 
   const { dataAccess, log, sqs } = context;
-  const { Site, Organization } = dataAccess;
+  const { Site } = dataAccess;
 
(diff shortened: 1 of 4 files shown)


MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - clean bug fix, correctly targeted, well-tested.
**Complexity: LOW** - small single-service fix, 4 files.
_Changes: Fixes the org-id format mismatch in `run-brand-claims` (sends SpaceCat UUID instead of IMS org string) and adds URL-based brand resolution to `toggle-brand-claims` (4 files)._

<details>

- nit: `toggle-brand-claims.test.js` uses `calledWith(ORG_ID, SITE_ID)` (partial match) on the `getBrandBySite` stub - the production call passes 4 args (`orgId, siteId, postgrestClient, log`) but only the first two are asserted. A future wiring regression (dropped `postgrestClient`) would not be caught. Consider `calledWithExactly` or asserting all four positional args.
- nit: `target` (raw Slack user input) is interpolated into `say()` responses in the new else-branch of `toggle-brand-claims.js`. Slack mrkdwn supports mention syntax (`<@U..>`, `<!here>`) that a user could trigger via the bot. This is the same pattern used elsewhere in the codebase (`run-brand-claims` echoes `siteArg` identically), so not a regression - but worth a future cleanup to escape or use `plain_text` blocks.

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 17s | <b>Cost:</b> $5.19 | <b>Commit:</b> <code>e1d8fb25dc0e109eb2f73f6d132065013e5014d3</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2964, opened 2026-07-30
feat(sites): default missing x-product to ASO on /sites-resolve under SKIP_AUTH

solaris007:
## 1. Abstract
`/sites-resolve` defaults a missing `x-product` header to ASO, but only when `SKIP_AUTH === 'true'` (local dev). Deployed environments are unchanged.

## 2. Reasoning
Running the local UI against a local api-service surfaced this: a locally-served ASO UI build can predate the `x-product` requirement on `/sites-resolve` and omit the header, so the endpoint fails with "Product code required" and the UI can never resolve a site locally. Rather than force every local UI checkout to carry the header, default the product locally where auth is already skipped.

- In `resolveSite` (`src/controllers/sites.js`), when `x-product` is absent AND `env.SKIP_AUTH === 'true'`, default `productCode` to `ASO_PRODUCT_CODE` and log an info line, instead of returning `badRequest('Product code required in x-product header')`.
- The guard is strict equality on `SKIP_AUTH === 'true'`. `SKIP_AUTH` is only ever set locally (never in a deployed env), so this path cannot activate in dev/stage/prod - the header contract there is byte-for-byte unchanged.
- Behaviour delta: local-only. `POST /sites-resolve` without `x-product` under `SKIP_AUTH` now proceeds to org/site resolution (as ASO) instead of 400ing on the missing header.

## 7. Additional information outside the code
Verified this session end-to-end: with this default in place, the locally-served ASO UI (webpack bundle in the exc-app shell) resolved a site against the local api-service and rendered the opportunity list. Without it, `/sites-resolve` returned "Product code required" and the UI fell back to onboarding.

## 8. Test plan
- Unit (`test/controllers/sites.test.js`, `resolveSite` block): a missing `x-product` under `SKIP_AUTH='true'` no longer returns "Product code required" (resolution proceeds past the product guard); a missing `x-product` when `SKIP_AUTH` is not exactly `'true'` still returns "Product code required". Run: `npx mocha test/controllers/sites.test.js -g "resolveSite"`.
- dev/stage/prod: no behavioural change to verify - `SKIP_AUTH` is unset there, so the pre-existing "Product code required" contract holds. Confirm via any `POST /sites-resolve` without `x-product` still returning 400.

### DIFF
--- a/src/controllers/sites.js
+++ b/src/controllers/sites.js
@@ -1780,7 +1780,16 @@ function SitesController(ctx, log, env) {
     } = context.data;
     const { pathInfo } = context;
     const X_PRODUCT_HEADER = 'x-product';
-    const productCode = pathInfo.headers[X_PRODUCT_HEADER];
+    let productCode = pathInfo.headers[X_PRODUCT_HEADER];
+    // Local-dev affordance (SKIP_AUTH only): the local UI harness may run a UI
+    // build that predates the x-product requirement on /sites-resolve and so omits
+    // the header. When auth is skipped (local only; SKIP_AUTH is never 'true' in a
+    // deployed env), default the product to ASO so the local UI ⇄ local api-service
+    // e2e can resolve a site. Prod keeps enforcing the header contract below.
+    if (!hasText(productCode) && env.SKIP_AUTH === 'true') {
+      productCode = ASO_PRODUCT_CODE;
+      log.info('[resolveSite] SKIP_AUTH local-dev: defaulting missing x-product to ASO');
+    }
     if (!hasText(productCode)) {
       return badRequest('Product code required in x-product header');
     }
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @solaris007,

**Verdict: Approve** - clean local-dev affordance, correctly guarded and tested.
**Complexity: MEDIUM** - API surface change in controller.
_Changes: Defaults a missing `x-product` header to ASO on `/sites-resolve` when `SKIP_AUTH === 'true'` (local dev only), with two new test cases (2 files)._

<details>

- suggestion: First new test could positively assert the SKIP_AUTH code path was taken (e.g. verify `loggerStub.info` was called with the SKIP_AUTH default message) rather than inferring it from reaching the next validation guard - `test/controllers/sites.test.js:6731`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 19s | <b>Cost:</b> $3.13 | <b>Commit:</b> <code>b839adbf2bd66873613e0ef0edc0ebbfe2b67170</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2965, opened 2026-07-30
fix(serenity): name a new market after its market, not the brand plus a random hash

rainer-friederich:
## 1. Abstract
A market created in serenity mode is named after the market it is — `US-en`, `CH-de` — instead of the brand display name plus a random 6-hex suffix.

## 2. Reasoning
Creating markets for a brand from the Elmo UI produced three markets named `Example website 5-0a8d59`, `Example website 5-e4cdd8` and `Example website 5-ee9d11`. The customer cannot tell them apart, and LLMO-managed sub-workspaces are locked but still **visible** to the customer in the Semrush navigation, so those names are customer-facing.

The rest of the system already assumes the market-shaped name. `transformMarketsToFilterDimensions` derives a filter region id by splitting the project name on `-` and taking the first segment, and its own type doc gives `AU-en` as the example label; the Elmo filter then localizes any 2-character id to a country name. Fed `Example website 5-0a8d59`, that split yields `Example website 5`, the localization never fires, and the raw name is rendered verbatim. The Semrush migration has meanwhile written `{REGION}-{language}` since the readable-names change, so UI-created and migrated markets carried two different conventions.

The server-side default name for a new market becomes `<REGION>-<language>` — uppercase ISO-2 country code plus the normalized language code. An explicitly supplied `name` still wins.

Before → after, for a market created without a `name`:

- Sub-workspace mode (the Markets tab): `Example website 5-0a8d59` → `US-en`.
- Flat mode: same default, applied at the same seam.
- Brand creation with an initial market: `US - EN` → `US-en`. Brand creation no longer passes a name of its own; it takes the handler's default, so a brand's first market and every market added later from the Markets tab are named identically.

Two behaviours follow from the name now being market-shaped:

- The dashboard Markets filter resolves a real 2-character region id from the label and renders the localized country name instead of the raw project name.
- A later migration pass, which adopts an existing project by matching its exact name within the sub-workspace, adopts a UI-created market rather than creating a duplicate beside it.

The dropped random suffix carried no identity: a market is addressed by its `(geoTargetId, languageCode)` slice, which is unique per brand, so the names within a sub-workspace remain collision-free. No lookup in this service keys on a project name. The deterministic name is also a usable correlation marker for the ambiguous-create recovery the provisioning proposal describes — a random suffix never was, since a retry generated a fresh one.

Language codes keep any regional subtag (`BR-pt-br`), so two slices that differ only in that subtag stay distinguishable; every code the migration emits is a bare primary subtag and renders exactly as the migration writes it.

## 4. Required information
- Other: naming convention source — adobe/mysticat-data-service `scripts/serenity_migration/planner.py` (`project_name=f"{region}-{language}"`) and `scripts/serenity_migration/rename.py`, the one-shot pass that renamed already-provisioned prod entities to this scheme.

- **mysticat-data-service** — contract. Its `serenity_migration` planner defines the `{REGION}-{language}` convention this change adopts, and its rename/adopt pass matches live projects by exact name.
- **project-elmo-ui** — consumed, no change needed. It renders the market label from `regions[].label` in the filter-dimensions payload and localizes it only when the derived region id is a 2-character code; the corrected name is what makes that path resolve.

- Verified the target convention against the migration source rather than assuming it: `planner.py` builds `project_name` as `f"{region}-{language}"` with an uppercase ISO-2 region and a lowercase primary language subtag, and `rename.py` documents the same scheme (`Lovesac` / `US-en`) as what already-provisioned prod entities were renamed to.
- Traced the display chain end to end to confirm the reported names originate here and not from a Site or brand name: `defaultMarketName` → Semrush project name → Markets element → `transformMarketsToFilterDimensions` `label` → Elmo `FilterDropdown` row. The market DTO the UI binds carries no `name` field, so the label can only be the project name.
- Confirmed no code in this service resolves a project by name; project resolution is by `(geoTargetId, languageCode)` slice, and the only name-keyed adoption is the sub-workspace title, which is untouched.

## 7. Test plan
Local: the unit suites for the two market handlers, brand provisioning and the brands controller were exercised against the new default, including the mixed-case (`ch`/`DE` → `CH-de`) and regional-subtag cases. No local stack run — the create path needs live Semrush.

Dev:
1. In the Elmo UI, open a serenity-mode brand's Markets tab and add a market (e.g. United States / English).
2. The market appears in the Semrush navigation as `US-en`, not `<brand>-<hex>`.
3. Add a second market in a different country and confirm the two are distinguishable by name.
4. On a dashboard with the Markets filter (URL inspector / brand presence), confirm the filter now lists the localized country name rather than the raw project name.
5. Create a brand with an initial market and confirm that market is also named `<REGION>-<language>`.

Existing markets are untouched — this changes the create default only. Markets already carrying a hashed name keep it until renamed.


src/support/serenity/handlers/markets.js
```diff
@@ -236,14 +235,29 @@ function validateCreateBody(body) {
...
+ * the Semrush navigation, so a project's name has to read as the market it is.
+ * This is the same convention the Semrush migration writes (adobe/
+ * mysticat-data-service `scripts/serenity_migration/planner.py`), so a market
+ * added from the Markets tab is indistinguishable from a migrated one — and a
+ * later migration pass, which adopts a project by matching its exact name
+ * within the sub-workspace, adopts it instead of creating a duplicate beside it.
+ *
+ * The name carries no identity: a market is addressed by its
+ * (geoTargetId, languageCode) slice, which is unique per brand, so the whole
+ * set of names within a sub-workspace is collision-free.
```
MysticatBot, line 251: issue (blocking): `defaultMarketName` is exported with a public JSDoc contract (`@param {string} market`, `@param {string|null} languageCode`), but `defaultMarketName(null, "en")` returns `"-en"` and `defaultMarketName("us", "")` returns `"US-"`. The tests explicitly assert this tolerance, so the behavior is intentional - but the output is customer-visible in the Semrush navigation.

Current callers validate before reaching this point (both handlers return 400 on invalid market/language), so the degenerate path is unreachable today. But the function is exported and its contract does not communicate this precondition. A future caller that skips the validation gauntlet would silently create a project named `-en` in a customer workspace.

Suggested fix (pick one):
- Throw when `region` or `lang` is empty after normalization (fail-fast at the source).
- Add `@precondition` / `@throws` to the JSDoc documenting that callers must pre-validate.
- If the tolerance is truly intentional and safe (workspace isolation guarantees uniqueness), add a code comment explaining why degenerate names are acceptable.

The test `tolerates a missing market or language` would then either expect a throw or carry the explanatory comment.

MysticatBot (changes requested):
Hey @rainer-friederich,

**Verdict: Request changes** - one design concern on the exported naming function.
**Complexity: HIGH** - medium diff; API surface risk flag (openapi schema change).
_Changes: replaces the random-hash market name default with a deterministic `<REGION>-<language>` convention matching the migration scripts (8 files)._
Note: Recommend a human read before merge - this change modifies a shared API contract (`docs/openapi/schemas.yaml`). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Important] Exported `defaultMarketName` produces degenerate names for empty inputs - `src/support/serenity/handlers/markets.js:248` (details inline)

<details>

- nit: `buildCreateProjectBody` now takes 5 positional parameters with two semantically similar strings (`languageId`, `languageCode`) that are easy to swap silently - consider an options object in a follow-up - `src/support/serenity/handlers/markets-subworkspace.js:235`
- suggestion: The retry cleanup comment (line 408) still frames the concern as if the name varies per retry; now that naming is deterministic, acknowledge that retries produce identical bodies and the cleanup is for orphan-project hygiene only - `src/support/serenity/handlers/markets.js:408`
- nit: `brandDisplayName` schema description change is correct but represents a behavioral contract change for existing API consumers who relied on it influencing the project name; a changelog note would help - `docs/openapi/schemas.yaml`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 40s | <b>Cost:</b> $5.11 | <b>Commit:</b> <code>ab0870e34f3a577a998591a001db7cb2cf31b026</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the review in `13f1a08f8`.

MysticatBot:
- [Important] `defaultMarketName` degenerate names for empty inputs — done. It now throws a 400 instead of returning `-en` / `US-`. You're right that the path is unreachable today (both create handlers 400 on an unparseable market or language before naming anything), but the export carried no such precondition and the output is customer-visible in the Semrush navigation, so fail-fast at the source beats documenting a landmine. The tolerance test became a throws test.
- nit: `buildCreateProjectBody` 5 positional params, `languageId` / `languageCode` transposable — done, and without deferring it. The parameter is gone rather than wrapped in an options object: `normalizeLanguageCode` is pure and already imported there, so deriving the code from `body` in place yields the identical value and restores the 4-arg signature. The two similar strings are never adjacent at a call site now.
- suggestion: retry-cleanup comment still framed around a varying name — done. It now says what actually holds: the retry body is byte-identical, and the duplicate happens because Semrush accepts no idempotency key.
- nit: `brandDisplayName` contract change deserves a changelog note — declined on the mechanism, addressed on the intent. This repo's CHANGELOG.md is generated by semantic-release from commit messages, so there is no hand-maintained file to annotate; the `fix(serenity):` commit is the changelog entry, and the PR body carries the explicit before/after. `brandDisplayName` itself is unchanged — it still sets `brand_name_display` and the own-brand benchmark; it just no longer seeds the default name. I did add the actionable half to the schema: the `name` description now states that supplying it overrides the default.

On the degraded-review note: there is no spec document for this one — it's a defect fix, and the design-of-record is the migration's own convention, linked in section 4 of the body (`scripts/serenity_migration/planner.py` builds `project_name` as `f"{region}-{language}"`, and `rename.py` is the pass that renamed live prod entities to it). I'd rather point at that than invent a spec after the fact.

src/support/serenity/handlers/markets.js
rainer-friederich: Fixed in `13f1a08f8` — took the fail-fast option. `defaultMarketName` now throws a 400 when either part is empty, so a `-en` / `US-` name can no longer reach a customer workspace, and the JSDoc documents the precondition with `@throws`. The `tolerates a missing market or language` test now asserts the throw.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all prior findings addressed cleanly; no new blocking issues.
**Complexity: HIGH** - large diff (generated docs inflate line count); shared API contract change.
_Changes: replaces the random-hash market name default with a deterministic `<REGION>-<language>` convention matching the migration scripts, and adds fail-fast validation on empty inputs (8 files)._
Note: Recommend a human read before merge - this change modifies a shared API contract (`docs/openapi/schemas.yaml`). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: `docs/serenity.md` line 452 still says `default is <brand>-<6 hex>` - now inaccurate; should read `<REGION>-<language>` (e.g. `US-en`) - `docs/serenity.md:452`
</details>

- `defaultMarketName` degenerate names on empty inputs - now throws 400 with clear message
- `buildCreateProjectBody` positional parameter confusion - extra param removed, language derived in place
- Retry-cleanup comment framing around varying names - reworded to reflect deterministic bodies
- `brandDisplayName` schema description - updated to document actual role (brand_name_display + benchmark)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 34s | <b>Cost:</b> $5.43 | <b>Commit:</b> <code>13f1a08f87b487b796dd2b81bf39b5b4fc9fc249</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the remaining nit in `78ee86109`.

MysticatBot:
- nit: `docs/serenity.md:452` still documents `<brand>-<6 hex>` — done, and thanks, that one was a genuine miss. The walkthrough now reads `<REGION>-<language>` (e.g. `US-en`). I also swept the rest of the repo for the old wording; the only other hit is noted below.

Deliberately left alone: `docs/specs/2026-05-28-prompts-api-abstraction.md:147` still shows the old default and, just above it, the rationale for the random suffix. That file is a dated design record (`Status: Draft (2026-05-28)`) rather than living documentation — the living contract is the OpenAPI schema and `docs/serenity.md`, both now correct. Rewriting the historical spec to today's behaviour would erase why the suffix existed, which is the more useful thing for it to carry. Happy to update it if you'd rather the specs track current behaviour.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - prior nit addressed; no new issues.
**Complexity: HIGH** - large diff (generated docs inflate line count); shared API contract change.
_Changes: replaces the random-hash market name default with a deterministic `<REGION>-<language>` convention matching the migration scripts (9 files)._
Note: Recommend a human read before merge - this change modifies a shared API contract (`docs/openapi/schemas.yaml`). The bot review is a complement to, not a replacement for, a human read here.
Note: CI checks are pending - confirm they pass before merge.

- `docs/serenity.md:452` walkthrough comment now reads `<REGION>-<language>` instead of `<brand>-<6 hex>`

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 4m 26s | <b>Cost:</b> $1.44 | <b>Commit:</b> <code>78ee86109c5cb7bdee425dacf08ce0995ebbd5fa</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Runtime validation — **not performed**, environment unavailable.

`ci / branch-deploy` succeeded for `78ee86109` and claimed the shared `dev-branches` slot at 11:12:16Z, but an unrelated branch took it 2m32s later:

- 11:12:16Z `78ee86109` `feat/project_name`
- 11:14:48Z `386aaae9c` `feature/LLMO-5566-cloudfront-log-delivery-assume-role` (current)

`dev-branches` is one shared slot on this repo, last green push wins, so exercising the dev API now would be testing that branch's code, not this PR's. Recording that rather than a result observed against someone else's deployment.

What has been verified instead:
- All required checks green on `78ee86109`, including `codecov/patch` at 100.00% of diff.
- The convention this PR adopts was confirmed against live Semrush data: across the LLMO-Dev and LLMO-Dev-2 parents, 14 AIO projects were renamed to `<REGION>-<language>` and read back live (`US-en`, `CH-de`, `AU-da`, `BO-es`, `US-fr`), so the target format is what the upstream accepts and displays.

The three markets from the original report (`Example website 5-0a8d59` / `-e4cdd8` / `-ee9d11`) are now `US-fr`, `BO-es` and `AU-da` — which is what this PR's default would have produced for those slices at create time.

Re-running the runtime check needs the slot back; happy to re-claim and validate if that is wanted before merge.

PR #2966, opened 2026-07-30
fix(brands): guard full-replace child syncs against silent wipes (LLMO-6591)

andreeastroe96:
## What & why

[LLMO-6591](https://jira.corp.adobe.com/browse/LLMO-6591) — `replaceChildRows` deletes every row for a brand in a child table (`brand_aliases`, `competitors`, `brand_social_accounts`, `brand_earned_sources`) and only *then* checks whether there's anything to reinsert — an empty input is a complete wipe, not a no-op. This hit DNP's trial on 2026-07-28 (7 aliases added live in a customer meeting, gone a few hours later) and, per the ticket's current-prod-exposure data, every one of these tables is populated at meaningful scale in production. Same failure class as LLMO-5533/5587 (the `status` full-replace fix), never extended to these collections.

Two routes reached it:
- **Route 1 (PATCH, guarded but defeated by the client):** `updateBrand` already skipped a child sync when the field was `undefined`, but elmo-ui's save path spreads the entire in-memory brand into every save, so a stale/reloaded tab sends `brandAliases: []` and wipes real data. This is the route that hit DNP (fixed in a companion project-elmo-ui PR).
- **Route 2 (upsert, unguarded):** `upsertBrand` called all four sync functions with no `!== undefined` guard at all — reachable with zero browser involvement (onboarding's upsert, Brandalf's async job).

## What changed

- **Centralized the `undefined`-skip guard** inside `syncAliases`/`syncCompetitors`/`syncSocialAccounts`/`syncEarnedSources` themselves, not at each call site. Protects both `upsertBrand` and `updateBrand` from one place, and can't be forgotten by a future third caller. `updateBrand`'s now-redundant per-field `!== undefined` wrappers are removed.
- **Optimistic concurrency on `updateBrand`**: an `expectedUpdatedAt` field on the update payload 409s (`brand_stale_write`) when it no longer matches the persisted row — another tab/request wrote in between. This is the mechanism that actually closes Route 1: an explicit empty array (not just an omitted field) is the genuinely destructive case, and "omit the field" alone (the `status` fix's approach) doesn't work here because an empty list is a legitimate way to delete the last alias. Backward compatible — omitting `expectedUpdatedAt` is a no-op, safe to deploy before any client sends it.
- **`BrandStaleWriteRejected` and `BrandCollectionWiped` EMF metrics**, mirroring the existing `BrandDemotionBlocked` pattern (`Mysticat/Brands` namespace) — the former on every stale-write rejection, the latter as a best-effort detection signal when a *successful* write takes a collection from non-empty to empty (often legitimate, but worth being able to see).
- **`createErrorResponse` now exposes `code` in the JSON body** (previously only `message`) — lets a client distinguish typed error cases without regex-matching message text. Also benefits the pre-existing demotion guard and the `uq_brand_name_per_org` name-conflict detection in elmo-ui, which had a TODO comment waiting for exactly this.

## Explicitly out of scope (flagged for follow-up, not blocking)

- Soft-delete/history for `brand_aliases` — real gap (no way to retroactively identify other wiped customers), but a schema change.
- Brandalf's own async writer — can't see its code from this repo; protected by the centralized guard *if* it already omits the field correctly, needs confirming with its owner.
- A related risk in `llmo-onboarding.js`: its onboarding upsert always sends a single stub alias, never `undefined`/`[]`, so on a re-onboard of an already-populated brand it would replace real aliases with just the stub. Not the empty-wipe pattern this ticket describes — worth its own follow-up ticket.

## Testing

- Storage unit tests: each sync function no-ops on `undefined` without touching the DB; still replaces on an explicit empty array; `updateBrand` 409s on a mismatched `expectedUpdatedAt`; an intentional empty-array write still succeeds when the timestamp matches; omitting `expectedUpdatedAt` still succeeds.
- Controller tests: 409 + `code: 'brand_stale_write'` body; `BrandStaleWriteRejected` EMF metric emission.
- 616 tests passing (195 storage + 421 controller). Full non-integration suite otherwise clean — 4 unrelated `Audits Controller` failures confirmed as pre-existing test-order flakes (reproduced in isolation on plain `main` before this change, unrelated files).
- Lint clean, type-check (base + strict) clean, docs:lint clean.

Companion project-elmo-ui PR threads `expectedUpdatedAt` through the save path and surfaces the 409 as a "reload and reapply" message.


MysticatBot (approved):
Hey @andreeastroe96,

**Verdict: Approve** - well-scoped bug fix with comprehensive test coverage and sound defensive design.
**Complexity: HIGH** - large diff; API surface changes (controller + storage layer).
_Changes: Guards four brand child-table sync functions against silent wipes on undefined input, adds optimistic concurrency via expectedUpdatedAt, adds observability metrics for stale writes and collection wipes, exposes typed error codes in API responses (4 files)._

<details>

- suggestion: Add a storage-layer test sending `expectedUpdatedAt: "not-a-date"` to document that malformed timestamps fail closed (rejected as stale, not skipped). The `Number.isNaN(expected)` branch is covered by the broader inequality check today, but an explicit test documents the design decision. - `src/support/brands-storage.js:1280`
- nit: The optimistic concurrency check reads `updated_at` via SELECT then issues a separate UPDATE without a WHERE clause on the timestamp. A concurrent write landing in between could pass the check but overwrite a newer version. Low risk (millisecond window, consistent with existing baseSiteId/status guards in this file), but a future `WHERE updated_at = $expected` conditional UPDATE would eliminate it entirely. - `src/support/brands-storage.js:1273`

</details>


---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 13m 34s | <b>Cost:</b> $6.54 | <b>Commit:</b> <code>018d2e6f62d1a20ea033b93227a77c29125ff368</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/controllers/brands.js
```diff
@@ -2039,10 +2100,32 @@ function BrandsController(ctx, log, env) {
...
 
+      // LLMO-6591: best-effort detection when this write takes a collection
+      // from non-empty to empty. Only pays for the pre-write read when at
+      // least one collection is being explicitly emptied — the common edit
+      // touches none of them, or replaces them with real values.
+      const emptiedCollections = ['brandAliases', 'competitors', 'socialAccounts', 'earnedContent']
+        .filter((field) => Array.isArray(updates[field]) && updates[field].length === 0);
+      const beforeForWipeCheck = emptiedCollections.length > 0
+        ? await getBrandById(spaceCatId, brandUuid, postgrestClient)
+        : null;
```
MysticatBot: issue (blocking): `getBrandById` throws on PostgREST errors (no `.status` on the thrown Error), which propagates to the outer catch and returns a 500 via `internalServerError()`. This converts a best-effort observability read into a request-blocking failure - a customer's brand edit fails with 500 when only the wipe-detection read hits a transient error, even though the actual write would succeed.

The PR documents this as "a detection signal for retroactive review, not a rejection," but the code does not match that intent.

Fix: wrap the `getBrandById` call in a try/catch that logs and falls back to `null`:
```javascript
let beforeForWipeCheck = null;
if (emptiedCollections.length > 0) {
  try {
    beforeForWipeCheck = await getBrandById(spaceCatId, brandUuid, postgrestClient);
  } catch (wipeCheckErr) {
    log.warn('brands: wipe-check pre-read failed, skipping detection', { brandId, error: wipeCheckErr.message });
  }
}
```

MysticatBot (changes requested):
Hey @andreeastroe96,

**Verdict: Request changes** - one new failure mode introduced by the wipe-detection read.
**Complexity: HIGH** - large diff; API surface changes (controller + storage layer).
_Changes: Closes the TOCTOU race in optimistic concurrency, extends null guards to all four sync functions, adds self-referential competitor rejection, and emits BrandCollectionWiped metrics (6 files)._

## Must fix before merge
1. [Important] Best-effort wipe-detection read can throw and 500 the request - `src/controllers/brands.js:~2125` (details inline)

<details>

- suggestion: Add a storage-layer test sending `expectedUpdatedAt: "not-a-date"` to document that malformed timestamps fail closed (rejects with 409, not silently accepted). Pins the `Number.isNaN` guard against future refactors. - `src/support/brands-storage.js:1299`
- nit: Consider normalizing `expectedUpdatedAt` to ISO 8601 (`new Date(value).toISOString()`) before passing to PostgREST `.eq()`, or document that clients must echo the exact `updatedAt` string from the read response. Postgres does semantic timestamptz comparison so this is unlikely to bite in practice, but the contract is implicit today. - `src/support/brands-storage.js:1397`
</details>

- TOCTOU race in optimistic concurrency check now closed with atomic `.eq('updated_at', expectedUpdatedAt)` predicate on the UPDATE statement.


---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 59s | <b>Cost:</b> $7.34 | <b>Commit:</b> <code>76596765d8e8d01284558703e5e778d4bb2b659d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @andreeastroe96,

**Verdict: Approve** - prior blocking finding addressed; no new issues.
**Complexity: HIGH** - large diff; API surface changes (controller + storage layer).
_Changes: Wraps the wipe-detection pre-read in try-catch so a failed getBrandById no longer 500s the write path, addressing the prior review's blocking finding (6 files)._

<details>

- suggestion: Normalize `expectedUpdatedAt` to ISO 8601 before passing to the PostgREST `.eq('updated_at', ...)` predicate (e.g. `new Date(updates.expectedUpdatedAt).toISOString()`). The JS validation uses millisecond-precision `Date.getTime()` comparison while the query sends the raw client string for Postgres to cast. This works today but the implicit contract is fragile if the transport layer ever changes. - `src/support/brands-storage.js:1397`
- nit: A one-line comment on the `.eq('updated_at', ...)` line clarifying it is the authoritative concurrency fence (with the JS pre-check being an early-exit optimization) would prevent a future maintainer from removing the DB-level filter thinking the JS check is sufficient. - `src/support/brands-storage.js:1395`

</details>

- Best-effort wipe-detection read now wrapped in try-catch with warn log; a failed pre-read skips detection instead of 500ing the request.
- Storage-layer test for malformed expectedUpdatedAt exists (fails closed with 409).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 41s | <b>Cost:</b> $5.44 | <b>Commit:</b> <code>5209e1b2da0ea17e1dc27933003d6c63b4abefaa</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2972, opened 2026-07-30
fix: dedupe fix_entity creation for aso mark-as-deployed (SITES-48951)

FentPams:
The ASO-UI "mark as deployed" path (POST /fixes, origin 'aso') could create a duplicate fix_entity for a suggestion already deployed via Mystique /v1/apply, surfacing a documentPath-less fix that disabled "Open in AEM Editor".

createFixes now returns the existing active (DEPLOYED/PUBLISHED) fix in the same opportunity instead of creating a duplicate. The dedupe is scoped to origin 'aso' (so the Mystique SQS worker's create_fixes, origin 'spacecat', keeps its 201) and to the request's opportunityId (so a caller can't have another tenant's fix echoed back by passing a foreign suggestionId). Preserves the manual-deploy case (no active fix => create).

## Related Issues

Thanks for contributing!

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Jeddie Chuang", "Dominique Jäggi"]
```


jeddiec:
## Code review (SITES-48951)

Server-side counterpart to the ESS-UI PR (#2154) — defense-in-depth on the backend.

### Overview
`createFixes` (`POST /fixes` batch) now short-circuits the **ASO-UI "mark as deployed"** path: if a target suggestion already has an active (`DEPLOYED`/`PUBLISHED`) fix in the same opportunity — e.g. a Mystique `/v1/apply` fix that raced it — it returns that existing fix as a per-item `200` instead of creating a documentPath-less duplicate. Scoped to `origin: 'aso'` (the Mystique SQS worker's `origin: 'spacecat'` keeps its `201`) and to the request's `opportunityId` (no cross-tenant echo). Two private helpers + five tests.

### Verifications
- **Linchpin accessor is real.** `getFixEntitiesBySuggestionId` is a genuine Suggestion-collection method, already used at `src/controllers/suggestions.js:992`, returning fix model instances with `getStatus()`/`getOpportunityId()`/`getId()`. (Note: the new tests *stub* it, so a non-existent method wouldn't be caught by tests — but it does exist, so this risk is closed.)
- **`countSucceeded` (`statusCode < 400`) counts the deduped `200` as success**, matching the tests' `success: 1` metadata.
- **Authz ordering is safe** — `checkOwnership(null, opportunityId, siteId, …)` runs before the batch map, and the echoed fix belongs to that same authorized `opportunityId`, so no privilege escalation via the surfaced documentPath.

### Code quality — strong
- Clean extraction (`#dedupeAsoFix` / `#findExistingActiveFix`), precise JSDoc on the *why* (race, origin scoping, cross-tenant reasoning); the `no-await-in-loop` disable is justified by the short-circuit.
- Uses the `FixEntityModel.ORIGINS.ASO` constant (better than the adjacent existing `#prepareDocumentPathEnrichment`, which still checks the `'aso'` literal — see nit).
- Good negative-space coverage: terminal `ROLLED_BACK` still creates, non-aso worker not deduped *and lookup not consulted*, cross-opportunity no-leak, non-active `PENDING` still creates.

### Findings / risks

- **(Low–Med, semantic) All-or-nothing dedupe keyed on *any* suggestion in the group.** `#findExistingActiveFix` returns on the first `suggestionId` with an active fix, and `#dedupeAsoFix` then short-circuits the *entire* `fixData`. If a group `suggestionIds: [A, B]` has an active fix for A but not B, no fix is created/linked for B, yet the UI (#2154) still patches **both** A and B to `FIXED` — leaving B `FIXED` with no backing fix of its own. Safe under the page-relationship group-homogeneity assumption (Mystique's apply fix would already be linked to the whole group), and #2154 suppresses aso fixes for V2-direct suggestions anyway — but the asymmetry is worth a one-line comment, or a stricter check (require all `suggestionIds` to resolve to the *same* active fix before deduping).

- **(Low) Intra-batch duplicates still slip through.** Two aso `fixData` in one request targeting the same currently-inactive suggestion both create (parallel `Promise.all`, neither persisted before the other's lookup). Pre-existing and outside this fix's target (the race against an *already-active* fix), just noting the dedupe only defends against already-live fixes — not full idempotency.

- **(Low, perf) Up to `suggestionIds.length` sequential DB reads per aso fixData.** Fine for typical small batches (and it short-circuits); flagging only that many suggestionIds × many fixData multiplies reads — a bulk accessor would be cheaper if that grows.

- **(Test gap, minor) `PUBLISHED` is in `ACTIVE_FIX_STATUSES` but only `DEPLOYED` is exercised.** Add a `PUBLISHED` case (and ideally the partial-group `[A,B]` case above).

- **(Nit, out of scope) Origin-check inconsistency.** New code uses `FixEntityModel.ORIGINS.ASO`; the adjacent `#prepareDocumentPathEnrichment` still uses `fixData.origin === 'aso'`. Migrating the literal keeps them from drifting.

### Security
No concerns — the `opportunityId` filter is the correct control against a caller probing another tenant's fix via a foreign `suggestionId` (explicitly tested), sitting behind the existing ownership check.

*Automated review — findings only; approve/merge is the reviewer's call.*

jeddiec (approved):
Approving. Re-reviewed at HEAD — my earlier findings are addressed: group-homogeneity is now documented in `#dedupeAsoFix`, the PUBLISHED and partial-group [A,B] cases have tests, and the origin literal is migrated to `FixEntityModel.ORIGINS.ASO`. Dedupe logic + `opportunityId` cross-tenant scoping verified; remaining intra-batch/perf notes are pre-existing and out of scope. Nothing blocking.

PR #2973, opened 2026-07-30
fix(sites): populate top-level code from hlxConfig on EDS site discovery/approval

rpapani:
## Problem

EDS / `aem_edge` sites discovered via the CDN/RUM site-detection hooks resolve their repo coordinates into **`hlxConfig`** — `hlxConfig.code` from the Franklin aggregated config API (`admin.hlx.page/config/{owner}/aggregated/{site}.json`) and `hlxConfig.rso` from the EDS hostname (`‹ref›--‹site›--‹owner›.aem.live`). But the only attribute read by downstream consumers is the **top-level `code`**:

- import-worker's code import (`codeConfigProvider` → `site.getCode()`)
- autofix-worker's code-PR flow (`CodeRepoManager.createFrom` → `site.getCode()`)

Neither reads `hlxConfig`. So sites created via site-candidate **approval** ended up with a populated `hlxConfig` but an empty top-level `code`, and their code import silently no-op'd (`"Code configuration is not set for site …"`). Example: World Bank `60c9fbc2` (`wbgextapp/externalweb-ibrd`) had `code: null` despite a full `hlxConfig`.

## Fix

Add `deriveCodeFromHlxConfig(hlxConfig)` (`src/support/utils.js`) and set the top-level `code` at the two places a **Site** is actually written with an hlxConfig:

- **`approve-site-candidate.js`** — on `Site.create` from an approved candidate, and on the existing-site update branch (backfill only when `code` is unset).
- **`hooks.js`** — when an existing EDS site's `hlxConfig` is refreshed by a CDN re-discovery, backfill `code` if unset.

Mapping (mirrors the import-worker's derivation):

| top-level `code` | source |
|---|---|
| `owner` | `hlxConfig.code.owner` → `hlxConfig.rso.owner` |
| `repo` | `hlxConfig.code.repo` → `hlxConfig.rso.site` |
| `ref` | `hlxConfig.rso.ref` (default `main`) |
| `type` | `hlxConfig.code.source.type` (default `github`) |
| `url` | `hlxConfig.code.source.url` (default `https://github.com/‹owner›/‹repo›`) |

An explicitly-set `code` is **never overwritten** (update/backfill paths only act when `code` is empty).

## Why only api-service (not audit-worker)

The audit-worker `site-detection/handler.js` and the api-service CDN/RUM hooks both only create **`SiteCandidate`** records, and `SiteCandidate` has no `code` attribute. The **Site** is created/updated in exactly these api-service paths, so that's the only correct place to set the top-level `code`.

## Scope / notes

- This is a forward fix — new/re-discovered EDS sites get a working `code`. Existing sites (like World Bank) were backfilled separately via a direct API patch.
- Makes the top-level `code` the single source of truth for repo coordinates, so `hlxConfig.code` can be retired later. Companion cleanup: adobe/spacecat-autofix-worker#684 (route AEMY/CM by repo coordinates, drop `installationId` from routing).

## Trace (verified)

Approval of a World-Bank-shaped candidate → `deriveCodeFromHlxConfig({rso:{owner:'wbgextapp',site:'externalweb-ibrd',ref:'main'}, code:{…}})` → `{type:'github', owner:'wbgextapp', repo:'externalweb-ibrd', ref:'main', url}` → `Site.create({…, code})`. The import-worker then reads `site.getCode()` (non-numeric → AEMY) and imports; autofix's `CodeRepoManager` routes the same coordinates to AEMY.

## Tests

- `deriveCodeFromHlxConfig` unit tests (derive from `code`, `rso` fallback, defaults, null cases).
- `approve-site-candidate` / `approve-friends-family`: create sets `code`; create omits `code` when non-derivable; update backfills; update does not clobber an existing `code`.
- `hooks`: existing-site hlxConfig refresh backfills `code`; does not backfill when a `code` already exists.
- Full suite in a clean run: **16417 passing**, coverage 99.84% lines / 98.59% branches (gate is 90%), lint + strict type-check clean. (Any local full-suite hook-timeout flakiness is CPU contention from running multiple large suites at once — the affected files don't import these changes.)

### DIFF
--- a/src/controllers/hooks.js
+++ b/src/controllers/hooks.js
@@ -376,7 +376,10 @@ function HooksController(lambdaContext) {
           // Backfill the top-level `code` attribute (source of truth for the
           // import-worker / autofix-worker) from the resolved hlxConfig when it
           // is not already set, so re-discovered EDS sites get a working code
-          // config without clobbering an explicit one.
+          // config without clobbering an explicit one. Note: this sits inside the
+          // `hlxConfigChanged` branch, so a site whose hlxConfig is unchanged does
+          // not self-heal here — new sites get code via the approve path, and
+          // pre-existing sites were backfilled separately (forward-fix scope).
           const existingCode = site.getCode() || {};
           if (Object.keys(existingCode).length === 0) {
             const derivedCode = deriveCodeFromHlxConfig(updatedHlxConfig);
--- a/src/controllers/hooks.js
+++ b/src/controllers/hooks.js
@@ -29,7 +29,7 @@ import yaml from 'js-yaml';
 
 import { BaseSlackClient, SLACK_TARGETS } from '@adobe/spacecat-shared-slack-client';
 import { Site as SiteModel, SiteCandidate as SiteCandidateModel } from '@adobe/spacecat-shared-data-access';
-import { isHelixSite } from '../support/utils.js';
+import { deriveCodeFromHlxConfig, isHelixSite } from '../support/utils.js';
 import { getHlxConfigMessagePart } from '../utils/slack/base.js';
 
 const CDN_HOOK_SECRET_NAME = 'INCOMING_WEBHOOK_SECRET_CDN';
@@ -373,6 +373,17 @@ function HooksController(lambdaContext) {
 
         if (hlxConfigChanged) {
           site.setHlxConfig(updatedHlxConfig);
+          // Backfill the top-level `code` attribute (source of truth for the
+          // import-worker / autofix-worker) from the resolved hlxConfig when it
+          // is not already set, so re-discovered EDS sites get a working code
+          // config without clobbering an explicit one.
+          const existingCode = site.getCode() || {};
+          if (Object.keys(existingCode).length === 0) {
+            const derivedCode = deriveCodeFromHlxConfig(updatedHlxConfig);
+            if (derivedCode) {
+              site.setCode(derivedCode);
+            }
+          }
           await site.save();
 
           const action = siteHasHlxConfig && hlxConfigChanged ? 'updated' : 'added';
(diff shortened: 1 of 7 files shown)


FentPams (approved):
**APPROVE** — clean, well-scoped, well-tested. Fact-checked against the import-worker (the reference this cites):

- The derived shape `{ type, owner, repo, ref, url }` is exactly what downstream consumes — `import-worker/src/handler/code-handler.js` reads the top-level `code` via `site.getCode()` and routes on `owner/repo`; `installationId` is data, not routing. ✅
- `repo = rso.site` is validated by the World Bank fixture (`rso.site` === `code.repo`), and the derivation prefers `hlxConfig.code.repo` and only falls back to `rso.site`. ✅
- `deriveCodeFromHlxConfig` uses `isObject`/`hasText`, both imported in `utils.js` — no undefined-reference risk. ✅

Two minor, non-blocking notes:

1. **Doc wording:** the comment says it "mirrors the import-worker's derivation (`code-handler.js`)," but that file *consumes* the top-level `code` — it doesn't derive from `hlxConfig`. This PR is the one introducing the derivation; the output shape is correct. Consider "produces the shape the import-worker consumes."
2. **`hooks.js` self-heal gap:** the backfill sits inside `if (hlxConfigChanged)`, so an existing EDS site with empty `code` but an unchanged `hlxConfig` won't self-heal on re-discovery (only via the approve path or an actual config change). Accepted forward-fix given existing sites were backfilled via API — worth a one-line comment noting it.

(Negligible: `Object.keys(existingCode).length === 0` treats a partially-populated `code` as "set" and skips backfill — extremely unlikely in practice.)

rpapani:
Thanks for the thorough review! Both notes addressed in 0418ef5:

1. **Doc wording** — you're right, and it's worth being precise: `code-handler.js` *consumes* the top-level `code` (via `site.getCode()`) and does not derive from `hlxConfig`. There was briefly an import-worker derivation PR, but it was reverted in favor of exactly this approach — so this helper is now the sole place the derivation happens. Reworded the comment to "produces the top-level `code` shape the import-worker and autofix-worker consume via `site.getCode()`."

2. **`hooks.js` self-heal gap** — added a comment noting the backfill sits inside the `hlxConfigChanged` branch, so an existing site with unchanged `hlxConfig` doesn't self-heal there (new sites get `code` via the approve path; pre-existing sites were backfilled via API). Kept as the forward-fix per your call rather than adding a `save()` on every unchanged re-discovery.

On the negligible partial-`code` note: agreed it's an edge case — our writers always set the full `{type, owner, repo, ref, url}` shape, so a partially-populated `code` shouldn't occur in practice; left the simple emptiness check as-is.

PR #2974, opened 2026-07-31
feat(onboard-status): show scraping stats and data-source status on demand

tkotthakota-adobe:
## Problem

The `onboard status <url>` Slack command renders **only** the "Opportunity Statuses" section. Unlike the onboarding auto-report (produced by `spacecat-task-processor`), it has **no scraping / data-source view** — so an operator running `onboard status` has no way to see that a site scraped **0 pages** (e.g. a bot-protected, Fastly-fronted site). That gap surfaced while investigating `blog.clover.com`, which produced 0 successful scrapes across multiple onboarding runs.

This is a follow-up to the task-processor fix that made the onboarding auto-report truthful (spacecat-task-processor#340 / SITES-48954). The command itself was already correct for opportunity statuses (DB-based, no CloudWatch); this PR only **adds** the missing scraping visibility.

## Change

Add a best-effort **scraping section** to `onboard status`, rendered before Opportunity Statuses:

- **Scraping Statistics** — Completed / Failed / **In progress** / Total.
- **Data Sources** — a `Scraping` line with a **tri-state** status: ✅ available / ⏳ in-progress / ❌ failed. A still-running scrape reads as in-progress (⏳), not failed.

Logic lives in a small helper module `onboard-status-scraping.js`:
- `deriveScrapingStatus(stats)` — pure tri-state (mirrors the task-processor helper).
- `getScrapingStats(baseURL, lastStartTime, context)` — aggregates `ScrapeUrl` statuses across the current onboarding's scrape jobs (filtered by `lastStartTime`), counting `PENDING`/`RUNNING` separately.
- `buildScrapingSection(...)` — builds the Slack lines; **best-effort**: a scrape-client failure returns `null` and the command still renders opportunity statuses.

## Notes
- The scraping helpers duplicate a small amount of logic from `spacecat-task-processor`'s opportunity-status report. Consolidating both into `@adobe/spacecat-shared-utils` is a reasonable follow-up (called out in code comments).
- This makes `onboard status` consistent in spirit with the onboarding auto-report while remaining a synchronous, DB-first command.

## Tests
- New `onboard-status-scraping.test.js` — 9 cases covering `deriveScrapingStatus`, `getScrapingStats` (aggregation + timestamp filtering + null), and `buildScrapingSection` (in-progress render, no-results render, best-effort error → null).
- New command-wiring tests in `onboard-status.test.js` — section rendered when scrape data exists; omitted when unavailable.
- Full command suite: **37 passing**; lint clean.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Dominique Jäggi", "FNU Tathagat"]
```


tathagat2241 (commented):
## Review: feat(onboard-status): show scraping stats and data-source status on demand

Additive follow-up to spacecat-task-processor#340 (+339/−0). Adds a best-effort scraping section (stats + tri-state data-source line) to the `onboard status` Slack command, so an operator can see a site scraped 0 pages / is still scraping — the gap that surfaced on `blog.clover.com`. New helper module + 10 lines of wiring + tests. I reviewed the tri-state logic, the onboarding-session scoping, and the best-effort path.

### Blockers
- None.

### Should-fix
- **The "no data yet" state renders `Scraping ❌`, reintroducing the running-vs-failed conflation this PR is meant to fix.** When there are no scrape jobs for this onboarding session (or no URL results yet), `getScrapingStats` returns `null` → `deriveScrapingStatus(null)` → `'unknown'` → the emoji `|| ':x:'` fallback → ❌. The null branch then pairs that ❌ with the text *"Scraping is in progress or no results available yet"* — a contradictory signal, and it's pinned by the `renders the "no results yet" message` test asserting `dataSourceLine === 'Scraping :x:'`. For an operator running `onboard status` early, a not-yet-scraped site reads as **failed** — exactly the confusion #340 and this PR set out to remove. Recommend mapping `unknown` → ⏳ (or an ℹ️ neutral) in the null-stats branch (and update that test). Note the *genuinely-terminal-failed* case (0 completed, 0 pending, >0 failed) still correctly resolves to `failed` → ❌ via the tri-state, so real failures stay ❌ — only the empty/unknown case needs to stop masquerading as a failure.

### Nits
- **Same status-enum assumption as #340** — `deriveScrapingStatus`/`getScrapingStats` count only `{COMPLETE, PENDING, RUNNING}`; any other non-terminal `ScrapeUrl` status falls through to `failed` when nothing has completed. Worth confirming the enum is exactly those four (shared concern with the task-processor helper).
- **Duplication drift risk** — the `deriveScrapingStatus` copy here mirrors the task-processor one (acknowledged; consolidation to `@adobe/spacecat-shared-utils` flagged as follow-up). Until that lands, remember the two copies must move together — e.g. if the should-fix above changes the `unknown` mapping, mirror it in #340 (or vice-versa).

### Strengths
- **Correctly scoped to the current onboarding** — filters scrape jobs by `onboardConfig.lastStartTime`, the *same* anchor the already-working `computeAuditCompletion` in this command uses, so stale prior-onboarding jobs don't bleed in and the timestamp type is consistent with an existing consumer (verified: tests treat it as ms, matching that usage).
- **Genuinely best-effort** — a scrape-client failure is caught, logs a warning, returns `null`, and the command still renders opportunity statuses; zero behavior change to the existing (DB-based) opportunity-status path. Pinned by an error→null test.
- **Consistent with the auto-report** — the tri-state mirrors the task-processor fix, so `onboard status` and the onboarding auto-report now tell the same running-vs-failed story.
- **Good targeted tests** — `deriveScrapingStatus` states, `getScrapingStats` aggregation + timestamp filtering + null, `buildScrapingSection` in-progress/no-results/error-null, plus command-wiring tests (section rendered when data exists, omitted when unavailable).

### Tests
Strong for the change — I read them and they cover the helper states, the session scoping, and the best-effort path. I could not execute the suite here, so `37 passing` is trusted. Heads-up: the `renders the "no results yet" message` test currently asserts `Scraping :x:`, so adopting the should-fix means flipping that expectation.

### Verdict
**Approve with one should-fix** — solid, well-scoped, well-tested additive visibility that reuses the right onboarding anchor and stays best-effort. The one change I'd make before merge is mapping the empty/`unknown` scraping state to ⏳/ℹ️ instead of ❌, so the on-demand command doesn't reintroduce the "still running looks failed" confusion it's meant to cure.

tathagat2241 (approved):

tkotthakota-adobe:
Thanks for the careful review 🙏 — all three addressed:

**Should-fix — empty/`unknown` state was rendering ❌.** ✅ Fixed in 758e22ea. The `unknown` scraping state now maps to a neutral **ℹ️** in the data-source line instead of ❌, so an early snapshot of a not-yet-scraped site no longer masquerades as a failure. (Went with ℹ️ over ⏳: honest about "no data yet", stays distinct from true `in_progress` (⏳), and matches the existing ℹ️ = "ran, nothing to show" convention in this report.) A genuinely terminal-failed scrape (0 completed, 0 pending, >0 failed) still resolves to ❌ via the tri-state, and the pinned `no results yet` test now asserts `Scraping :information_source:`.

**Nit 1 — status-enum assumption.** Confirmed against source: `scrape_url_status` is exactly `{ PENDING, RUNNING, REDIRECT, COMPLETE, FAILED, STOPPED }` (`@mysticat/data-service-types`). `PENDING`/`RUNNING` are the only non-terminal states, so the predicate is complete — no `QUEUED`/`PROCESSING`. `REDIRECT`/`FAILED`/`STOPPED` are terminal non-successes and correctly read as `failed` when nothing has completed.

**Nit 2 — duplication drift.** Mirrored this exact `unknown → ℹ️` change in the task-processor copy so the two `deriveScrapingStatus` consumers stay in sync: adobe/spacecat-task-processor#343. (Shared-utils consolidation remains the longer-term follow-up.)

PR #2976, opened 2026-07-31
fix(llmo): echo detectedCdn from the site-only onboarding endpoint

andreeastroe96:
## Summary
- `onboardSiteOnly` (`POST /v2/orgs/:orgId/llmo/onboard-site`) already runs CDN detection via `performLlmoOnboarding` (the same orchestrator the legacy `/llmo/onboard` handler uses) and persists it to the site's LLMO config, but never included `detectedCdn` in its own response — unlike the legacy handler one line above it.
- This silently dropped CDN auto-select pre-fill for every client using the site-only endpoint. Currently that's project-elmo-ui's self-serve domain-onboarding wizard step; a companion elmo-ui PR (reusing that same wizard for first-run org onboarding) depends on this fix to avoid regressing CDN pre-fill for that new surface too.
- Also documents the field on `LlmoOnboardSiteResponse` in the OpenAPI spec, reusing the existing CDN-provider enum already defined for the sibling `/llmo/onboard` endpoint.

## Test plan
- [x] Updated the existing happy-path test (`test/controllers/llmo/onboard-site-only.test.js`) which previously asserted the exact response body and would have failed once `detectedCdn` appeared
- [x] Added a new test asserting a non-null `detectedCdn` is echoed through
- [x] Full unit suite passes (`npm test`)
- [x] `redocly lint` validates the updated OpenAPI spec


test/controllers/llmo/onboard-site-only.test.js
```diff
@@ -174,10 +174,27 @@ describe('LlmoController — onboardSiteOnly (LLMO-5606)', () => {
         organizationId: 'org123',
         baseURL: 'https://example.com',
         dataFolder: 'dev/example-com',
+        detectedCdn: null,
         status: 'processing',
       });
     });
 
+    it('echoes a non-null detectedCdn from performLlmoOnboarding (LLMO client CDN pre-fill)', async () => {
+      performLlmoOnboardingStub.resolves({
```
MysticatBot, line 183: issue (blocking): The test uses `'cloudflare'` as the `detectedCdn` value, but the canonical `CDN_TYPES` vocabulary (defined in `src/controllers/llmo/llmo-utils.js`) and the OpenAPI enum you are documenting in this same PR list only prefixed compound values (`byocdn-cloudflare`, `aem-cs-fastly`, etc.). `'cloudflare'` is not among them and would never be returned by `performLlmoOnboarding` in production.

The pass-through logic is still exercised correctly, but the fixture is misleading: a reader will assume `'cloudflare'` is a real production value, and it contradicts the contract formalized one file away in this same commit.

**Fix:** Replace `'cloudflare'` with a real enum member, e.g. `'byocdn-cloudflare'`.

MysticatBot (changes requested):
Hey @andreeastroe96,

**Verdict: Request changes** - test fixture uses a CDN value that contradicts the OpenAPI enum documented in this same PR.
_Changes: Adds the missing `detectedCdn` field to the site-only LLMO onboarding endpoint response and documents it in the OpenAPI spec (3 files)._

## Must fix before merge
1. [Important] Test uses non-canonical CDN value `'cloudflare'` which is not in the `CDN_TYPES` vocabulary - `test/controllers/llmo/onboard-site-only.test.js:183` (details inline)

<details>

- suggestion: OpenAPI nullable enum should include `null` in the enum list for strict JSON Schema 2020-12 compliance (without it, strict validators will reject `{"detectedCdn": null}`) - `docs/openapi/schemas.yaml:4676`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 24s | <b>Cost:</b> $4.35 | <b>Commit:</b> <code>fa95b872bf757c49430b5f6750eb774000e97a74</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2977, opened 2026-07-31
fix: read geo-experiment insights from the Mystique bucket (S3_MYSTIQUE_BUCKET)

ssilare-adobe:
## Problem

`GET /sites/{siteId}/geo-experiments/{id}?includeInsights=true` always returned `insights: null`, even when the insights blob existed in S3.

The insights `GetObject` used `context.s3.s3Bucket` (the **default services bucket**), but the impact-measurement artifacts — `insights.json` **and** the per-analysis detail blobs its `rawDataUrl`s point at — are written by Mystique/the engine to the **Mystique assets bucket** (`S3_MYSTIQUE_BUCKET`, e.g. `spacecat-dev-mystique-assets`). So the object was never in the bucket api-service looked in → `NoSuchKey`, swallowed and logged as "Could not fetch insights".

### Evidence (dev)
- `s3://spacecat-dev-mystique-assets/geo-experiments/4741ad90-…/insights.json` exists (LastModified `2026-07-31T07:59:10Z`).
- api-service Lambda log, **after** that write:
  `[geo-experiment] Could not fetch insights for 4741ad90-…: The specified key does not exist.`
- The prompts that *do* resolve (`geo-experiments/{siteId}/…-prompts.json`) are **404 in the Mystique bucket** — confirming api-service reads a different (default) bucket for them.

## Fix

Read the insights JSON from `context.env.S3_MYSTIQUE_BUCKET` instead of the default `s3Bucket`. Prompts stay on the default bucket (that's where they are), and the `rawDataUrl` presigning is unchanged — it already extracts the bucket from each `s3://…` URI.

`S3_MYSTIQUE_BUCKET` is an existing env var (already used by the reports controller; present in `.env.example`), so no new configuration is introduced.

## What does NOT change
- Prompts fetch (default `s3Bucket`).
- `presignInsightsRawData` (bucket parsed from each `rawDataUrl`).
- Response shape / the `includeInsights` gate.

## Test plan
- [x] `getGeoExperiment` suite green (18 tests), incl. new assertions that insights `GetObject` uses the Mystique bucket while prompts use the default bucket, and a "no `S3_MYSTIQUE_BUCKET` configured → insights null, no fetch attempted" case.
- [x] Full `suggestions.test.js` green (489 passing).
- [x] `eslint` clean.

Relates to LLMO-6042 (serve insights via geo-experiments API).

### DIFF
--- a/src/controllers/suggestions.js
+++ b/src/controllers/suggestions.js
@@ -2444,16 +2444,20 @@ function SuggestionsController(ctx, sqs, env) {
     // Fetch impact-measurement insights from S3 only when explicitly requested.
     // Insights exist once impact measurement completes; the S3 key is stored on the
     // experiment as insightsLocation (see spacecat-shared GeoExperiment model).
+    // The insights JSON (and the per-analysis detail blobs its rawDataUrls point at) are
+    // written by Mystique/the engine to the Mystique assets bucket (S3_MYSTIQUE_BUCKET),
+    // NOT the default services bucket — read it from there.
     let insights;
     const includeInsights = context.data?.includeInsights === 'true';
     if (includeInsights) {
       insights = null;
       const insightsS3Key = geoExperiment.getInsightsLocation?.();
-      if (insightsS3Key) {
+      const { S3_MYSTIQUE_BUCKET: mystiqueBucket } = context.env;
+      if (insightsS3Key && mystiqueBucket) {
         try {
-          const { s3Client, s3Bucket, GetObjectCommand } = context.s3;
+          const { s3Client, GetObjectCommand } = context.s3;
           const response = await s3Client.send(
-            new GetObjectCommand({ Bucket: s3Bucket, Key: insightsS3Key }),
+            new GetObjectCommand({ Bucket: mystiqueBucket, Key: insightsS3Key }),
           );
           const body = await response.Body.transformToString();
           insights = JSON.parse(body);
--- a/test/controllers/suggestions.test.js
+++ b/test/controllers/suggestions.test.js
@@ -10170,6 +10170,8 @@ describe('Suggestions Controller', () => {
 
   describe('getGeoExperiment', () => {
     const GEO_EXP_ID = 'b1b2c3d4-e5f6-7890-abcd-ef1234567890';
+    // Insights (and their detail blobs) live in the Mystique assets bucket, not the default one.
+    const MYSTIQUE_BUCKET = 'test-mystique-bucket';
 
     let mockGeoExperiment;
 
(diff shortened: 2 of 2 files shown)


khushboo5723 (approved):

src/controllers/suggestions.js
```diff
@@ -2444,16 +2444,20 @@ function SuggestionsController(ctx, sqs, env) {
...
+    // written by Mystique/the engine to the Mystique assets bucket (S3_MYSTIQUE_BUCKET),
+    // NOT the default services bucket — read it from there.
     let insights;
     const includeInsights = context.data?.includeInsights === 'true';
     if (includeInsights) {
       insights = null;
       const insightsS3Key = geoExperiment.getInsightsLocation?.();
-      if (insightsS3Key) {
+      const { S3_MYSTIQUE_BUCKET: mystiqueBucket } = context.env;
+      if (insightsS3Key && mystiqueBucket) {
```
MysticatBot, line 2456: issue (blocking): The new guard `if (insightsS3Key && mystiqueBucket)` introduces a silent no-op path. Before this PR, the code always attempted the fetch and logged on failure (via the catch block). Now, when `insightsS3Key` is truthy but `mystiqueBucket` is falsy, no fetch is attempted and no log is emitted. The caller sees `insights: null` with no way to distinguish "insights not yet generated" from "deployment misconfiguration."

This matters because the reports controller in the same codebase treats `S3_MYSTIQUE_BUCKET` as required (throws at startup if missing). The suggestions controller now introduces a softer contract for the same env var, but without any observability signal when the soft path fires.

Fix - add a warning before or as an else to the guard:
```js
if (insightsS3Key && !mystiqueBucket) {
  context.log.warn('[geo-experiment] S3_MYSTIQUE_BUCKET not configured; cannot fetch insights');
}
```
One line, makes misconfiguration visible in operational logs without changing the response shape.

MysticatBot (changes requested):
Hey @ssilare-adobe,

**Verdict: Request changes** - one observability gap in the new configuration guard.
_Changes: Fixes geo-experiment insights fetch to read from the Mystique assets bucket (S3_MYSTIQUE_BUCKET) instead of the default services bucket (2 files)._

## Must fix before merge
1. [Important] Silent skip when bucket env var is missing - no log emitted - `src/controllers/suggestions.js:2456` (details inline)

<details>

- suggestion: Extract `S3_MYSTIQUE_BUCKET` at the controller factory level (alongside other env-driven config) rather than per-request inside the conditional. This is how `reports.js` surfaces the same dependency and makes the env requirement visible at the module boundary. Not a blocker for this PR.
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 54s | <b>Cost:</b> $4.39 | <b>Commit:</b> <code>620328d97251adc36c7b9ac06cd326b72d80eee0</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2982, opened 2026-08-02
feat(llmo): add on-demand opportunity validation API

jindaliiita:
## Summary
- Adds `POST /sites/:siteId/opportunities/:opportunityId/validate`, letting a caller trigger the same S3-vs-live-edge prerender validation the hourly `optimize-at-edge-enabled-marking` job runs, instead of waiting for the automatic per-site run or going through the Slack-only `validate-only` command. Request body is one of two mutually-exclusive modes:
  - `{ suggestionIds: string[] }` (up to 200) — an explicit, ad hoc list of suggestions to re-check. Result lands in `opportunity.data.validation`.
  - `{ geoExperimentId: string }` — used by the deploy-to-edge experimentation flow to re-check a whole `GeoExperiment`'s own suggestions in one call, without the caller needing to already know every suggestionId. The controller stays thin — it only validates the ID's shape and forwards it; import-worker resolves the `GeoExperiment` and its suggestions itself. Result lands on `GeoExperiment.metadata.validation` + each covered suggestion's own `data.validation`, not `opportunity.data.validation` (a scoped-subset result there would corrupt the bulk job's site-wide validation cache).
- Fire-and-forget: enqueues an SQS message to import-worker's existing `imports` queue (reusing the `optimize-at-edge-enabled-marking` message type) and responds `202` immediately.
- Only `prerender`-type opportunities are accepted today (400 otherwise) — the route and message contract are deliberately generic (`opportunityId`, not `prerenderOpportunityId`) so a future opportunity type can be added later without reshaping the contract.
- Adds/updates OpenAPI docs (path, request/response schemas) per this repo's OpenAPI-first convention.

## Import-worker side
Both message shapes are now fully implemented on the import-worker side, in [spacecat-import-worker#836](https://github.com/adobe/spacecat-import-worker/pull/836) — `suggestionIds` mode was already read; `geoExperimentId` mode is new there too (bypasses the automatic candidate-selection path entirely, resolves the experiment's own `suggestionIds` directly, skips the S3-staleness check since those suggestions were just deployed). Full design at that PR's `docs/specs/opportunity-validation-api/spec.md`.

Also depends on [spacecat-shared#1863](https://github.com/adobe/spacecat-shared/pull/1863) (adds the `validation` field to the `PRERENDER` suggestion data schema and `GeoExperiment.METADATA_KEYS.VALIDATION`) — not a hard runtime dependency, since both target fields are pre-existing freeform JSON, but should land first for the schema documentation to be accurate.

## Test plan
- [x] Unit tests for the controller — both modes: happy path, 400/404 for every validation branch, mutual-exclusion check, 200/201-URL cap boundary
- [x] `getRouteHandlers` route-segregation test updated for the new route/mock
- [x] `required-capabilities`/`facs-capabilities`/`capability-constants` consistency tests pass unmodified (no unclassified route/param)
- [x] 100% coverage on the changed controller file
- [x] `npm run docs:lint` / `npm run docs:build` pass (pre-existing warning classes only, not new)
- [ ] Integration tests (`test/it/`) and e2e coverage not added in this PR — flagged as a follow-up given the scope of Postgres seed-data wiring involved

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```


src/controllers/llmo/opportunities/opportunity-validation-controller.js
```diff
@@ -0,0 +1,113 @@
...
+  accepted, badRequest, notFound,
+} from '@adobe/spacecat-shared-http-utils';
+import { isNonEmptyObject, isValidUUID } from '@adobe/spacecat-shared-utils';
+import { Audit } from '@adobe/spacecat-shared-data-access';
+
+// The route/request contract is generic (opportunityId, not prerenderOpportunityId) so a
+// future opportunity type can be added later without reshaping it. This map is the one place
+// that gates which opportunity types are actually supported today, and which import-worker
+// message type knows how to validate each one — adding a second type means adding an entry
+// here (and building that type's own resolver/comparator on the import-worker side); no other
```
MysticatBot, line 23: issue (blocking): The error message `Site ID required` is misleading when the parameter IS present but fails UUID validation. The `geoExperimentId` check on line 54 uses the clearer pattern (`geoExperimentId must be a valid UUID`). Consistency within the same controller matters for debuggability.

**What is wrong:** Lines 23 and 29 return "Site ID required" / "Opportunity ID required" when the actual problem is format validation, not a missing param.

**Why it matters:** A caller debugging a malformed UUID gets a confusing hint that the field is missing rather than malformed.

**Fix:** Use `Site ID must be a valid UUID` and `Opportunity ID must be a valid UUID`, matching the pattern already established for `geoExperimentId` on line 54.

src/controllers/llmo/opportunities/opportunity-validation-controller.js
```diff
@@ -0,0 +1,113 @@
...
+ */
+function OpportunityValidationController() {
+  /**
+   * POST /sites/:siteId/opportunities/:opportunityId/validate
+   * Body: { geoExperimentId: string }.
+   */
+  const triggerValidation = async (context) => {
+    const { dataAccess, sqs, log } = context;
+    const { Site, Opportunity, Configuration } = dataAccess;
+    const siteId = context.params?.siteId;
```
MysticatBot, line 56: issue (blocking): `geoExperimentId` is a known ReBAC entity in this system (it has its own `:geoExperimentId` path param on GeoExperiment routes elsewhere). Here it is accepted as an opaque UUID in the request body with no ownership verification - a caller with `opportunity:write` on Site A could reference a GeoExperiment belonging to Site B.

**Why it matters:** The FACS wrapper cannot enforce a relationship check on body params today, and the controller does not verify the experiment belongs to the referenced site/opportunity. This creates an authorization gap between format validation and actual ownership.

**Fix (pick one):**
1. If GeoExperiment is queryable from this service's data layer, add a lookup confirming it belongs to the same site or opportunity before enqueuing.
2. If ownership validation is intentionally deferred to import-worker (as the PR description states), add an explicit code comment documenting this trust boundary decision, e.g.: `// Ownership check deferred to import-worker - this controller only validates UUID format per design in spacecat-import-worker#836`

Option 2 is acceptable since the caller must already hold a privileged capability (`llmo/can_configure`) and the import-worker resolves the experiment in context.

MysticatBot (dismissed):
Hey @jindaliiita,

**Verdict: Request changes** - two issues to address before merge.
**Complexity: MEDIUM** - medium diff; API surface + FACS/ReBAC signal.
_Changes: Adds a fire-and-forget POST endpoint for on-demand opportunity validation, wiring it through FACS capabilities, required capabilities, and OpenAPI docs (11 files)._

## Must fix before merge
1. [Important] Inconsistent error messages for UUID validation - `opportunity-validation-controller.js:23,29` (details inline)
2. [Important] Body param `geoExperimentId` references a known ReBAC entity without ownership verification or documented trust boundary - `opportunity-validation-controller.js:54` (details inline)

<details>

- suggestion: Include `opportunityId` in the SQS message payload for downstream observability and future-proofing - `opportunity-validation-controller.js:60-65`
- nit: Body validation happens after the opportunity-type check, so a caller sending `{}` to a non-prerender opportunity gets "type not supported" rather than "no data provided" - ordering is unconventional but not a bug - `opportunity-validation-controller.js:47-54`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 14m 0s | <b>Cost:</b> $7.16 | <b>Commit:</b> <code>71d4c64d521b1560ccadfa2407cde791be58e423</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @jindaliiita,

**Verdict: Approve** - prior findings addressed; remaining items are non-blocking suggestions.
**Complexity: HIGH** - large diff (11 files); API surface + FACS/ReBAC signal.
_Changes: Adds a fire-and-forget POST endpoint for on-demand GeoExperiment validation, with ownership verification, FACS/capability wiring, and OpenAPI docs (11 files)._

<details>

- suggestion: Wrap `Configuration.findLatest()` / `sqs.sendMessage` in try/catch for explicit error logging - currently an SQS outage or missing config produces a generic 500 with no controller-level log context. Other controllers follow the same unwrapped pattern, so this is not a regression, but this endpoint is user-facing (callers get 202 on success then a raw 500 on infra failure with no retry signal). A try/catch + `log.error` + `internalServerError()` would improve debuggability. - `opportunity-validation-controller.js:73-78`
- nit: Body-validation branches (`No data provided`, `geoExperimentId must be a valid UUID`) omit `log.warn`, unlike every other validation branch in the same function - makes 400s from body issues invisible in operational logs. - `opportunity-validation-controller.js:57-61`
- suggestion: Include `opportunityId` in the SQS message payload for downstream correlation and DLQ debugging - the import-worker currently has to resolve it from the GeoExperiment's link, which is null for site-scoped experiments. - `opportunity-validation-controller.js:73-78`
</details>

- Error messages now correctly say "must be a valid UUID" instead of the misleading "required" (commit 1e726c31)
- GeoExperiment ownership verification added - controller checks siteId match and opportunityId link before enqueuing (commit f157c5ca)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 25s | <b>Cost:</b> $7.25 | <b>Commit:</b> <code>b73e4c220c19c1d17ec9ac88539f38e10caa2921</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2986, opened 2026-08-03
feat: allow insightsLocation in the geo-experiment PATCH endpoint

ssilare-adobe:
## What
Adds `insightsLocation` to the `PATCHABLE_FIELDS` allow-list of `PATCH /sites/:siteId/geo-experiments/:geoExperimentId` (`patchGeoExperiment`).

## Why
The DTO already exposes `insightsLocation` on read, and the `GeoExperiment` model has `setInsightsLocation`, but the PATCH allow-list omitted it — so it could never be set via the API. A PATCH that included `insightsLocation` returned `200` while silently dropping the field (and `400 "No valid fields to update"` if it was the only field).

This is the field the llmo-experimentation-engine writes at the end of the impact-measurement lifecycle (the `geo-experiments/{id}/insights.json` pointer). Allowing it here gives an authenticated, access-controlled way to set/correct it manually — direct PostgREST writes to `geo_experiments` are service-role-only, so the API is the only user-facing write path.

## Changes
- `src/controllers/suggestions.js` — one entry added to `PATCHABLE_FIELDS`: `{ key: 'insightsLocation', setter: 'setInsightsLocation' }`.
- `test/controllers/suggestions.test.js` — added `setInsightsLocation` to the mock, a dedicated `patches insightsLocation and saves` test, and extended the all-fields test to cover it.

## Test plan
- [x] `npx eslint` on both files — clean
- [x] `patchGeoExperiment` suite passes (16 passing, incl. the new cases)

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```

### DIFF
--- a/src/controllers/suggestions.js
+++ b/src/controllers/suggestions.js
@@ -2521,6 +2521,7 @@ function SuggestionsController(ctx, sqs, env) {
       { key: 'suggestionIds', setter: 'setSuggestionIds' },
       { key: 'promptsCount', setter: 'setPromptsCount' },
       { key: 'promptsLocation', setter: 'setPromptsLocation' },
+      { key: 'insightsLocation', setter: 'setInsightsLocation' },
       { key: 'startTime', setter: 'setStartTime' },
       { key: 'endTime', setter: 'setEndTime' },
       { key: 'metadata', setter: 'setMetadata' },
--- a/test/controllers/suggestions.test.js
+++ b/test/controllers/suggestions.test.js
@@ -10546,6 +10546,7 @@ describe('Suggestions Controller', () => {
         setSuggestionIds: sandbox.stub(),
         setPromptsCount: sandbox.stub(),
         setPromptsLocation: sandbox.stub(),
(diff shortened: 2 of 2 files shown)


MysticatBot (approved):
Hey @ssilare-adobe,

**Verdict: Approve** - clean, well-scoped addition of a missing field to an existing allow-list.
**Complexity: LOW** - trivial diff; single API surface signal.
_Changes: Adds `insightsLocation` to the PATCHABLE_FIELDS allow-list for the geo-experiment PATCH endpoint, with corresponding unit tests (2 files)._

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 12s | <b>Cost:</b> $2.56 | <b>Commit:</b> <code>8d2de0b7bf4f72b7d145cadeed3de184e3a0655b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2987, opened 2026-08-03
fix(suggestions): reject illegal per-item status transitions in bulk PATCH (SITES-49063)

alinarublea:
## What

The bulk `PATCH /sites/{siteId}/opportunities/{oppId}/suggestions/status` handler (`patchSuggestionsStatus`) applied a caller-supplied target status across the whole batch **without checking per-item legality**, so it silently promoted **OUTDATED → FIXED**. An OUTDATED suggestion's issue is no longer detected, so it must not become FIXED.

This adds a per-item transition-legality gate using `isAllowedSuggestionTransition(currentStatus, status)` from the shared status-transition table (SITES-47091), returning `400 Illegal status transition: <from> -> <to>` **before** `setStatus`. Same-status writes, the SKIPPED skip-field-clearing path, and `updatedBy` stamping are unaffected (the gate sits inside the existing `currentStatus !== status` branch). The pre-existing REJECTED hard-rule (REJECTED only from PENDING_VALIDATION) is a stricter subset kept ahead for its specific message/ACL.

Also bumps `@adobe/spacecat-shared-data-access` **4.16.0 → 4.16.1** (the reconciled transition table from adobe/spacecat-shared#1860).

## Why

Surfaced by the SITES-47091 guard running in **warn** mode: **159 `OUTDATED → FIXED` violations** over 25 days in prod, all from this endpoint. This is the **last caller-side item** gating a safe flip of `STATUS_TRANSITION_ENFORCEMENT=warn → enforce` (SITES-47286) — once the endpoint stops doing this, `OUTDATED → FIXED` stays correctly disallowed in the shared table and `enforce` won't throw on legitimate traffic.

## Rollout-risk check

The shared table was reconciled from 25 days of warn-mode logs, which surfaced exactly **4** violation types: `OUTDATED→FIXED`, `REJECTED→OUTDATED`, `FAILED→ROLLED_BACK`, `NEW→REJECTED`. `SKIPPED→APPROVED` (a plausible "unskip & approve" one-hop) was **not** observed — since the guard wraps `setStatus`, it would have logged a violation had it occurred — so no legitimate current traffic is newly rejected by this gate.

## Tests

- New `status transition legality (SITES-49063)` block: `OUTDATED→FIXED` rejected with an explicit **save-not-called** assertion; `IN_PROGRESS→FIXED` allowed (saves once).
- Adjusted 9 pre-existing bulk-patch tests that had used synthetic placeholder statuses to use real legal transitions (added a `throwOnSetStatus` mock flag to preserve the setStatus-error path).
- **492 passing**, lint clean.

## Tickets
- Fixes SITES-49063
- Part of SITES-47286 (guard warn → enforce rollout) / SITES-47076 (lifecycle parent)

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```

### DIFF
--- a/src/controllers/suggestions.js
+++ b/src/controllers/suggestions.js
@@ -1126,16 +1126,13 @@ function SuggestionsController(ctx, sqs, env) {
             }
           }
 
-          // Reject per-item transitions the suggestion lifecycle does not allow
-          // (SITES-49063): the bulk endpoint applies a caller-supplied status across a
-          // whole batch, which previously flipped OUTDATED -> FIXED (an OUTDATED
-          // suggestion's issue is no longer detected, so it must not become FIXED).
-          // The shared transition table (SITES-47091) is the single source of truth;
-          // this also gates a safe flip of STATUS_TRANSITION_ENFORCEMENT=enforce (SITES-47286).
-          // NOTE: the REJECTED hard-rule above (REJECTED only from PENDING_VALIDATION) is a
-          // stricter subset of this same invariant kept for its specific message/ACL; it
-          // fires first, so keep it ahead of this general gate.
+          // Per-item transition-legality gate (SITES-49063; part of the SITES-47286
+          // warn->enforce rollout). NOTE: the REJECTED hard-rule above is a stricter
+          // subset of this and fires first — keep it ahead of this general gate.
           if (!isAllowedSuggestionTransition(currentStatus, status)) {
+            // logged so unexpected 400s can be triaged from Splunk when
+            // STATUS_TRANSITION_ENFORCEMENT is flipped to enforce (SITES-47286).
+            context.log.info(`[patchSuggestionsStatus] rejected illegal status transition suggestionId=${id} ${currentStatus} -> ${status}`);
             return {
               index,
               uuid: id,
--- a/src/controllers/suggestions.js
+++ b/src/controllers/suggestions.js
@@ -39,6 +39,7 @@ import {
   FEEDBACK_TIERS,
   verdictToSignal,
   toReviewView,
+  isAllowedSuggestionTransition,
 } from '@adobe/spacecat-shared-data-access';
 import TokowakaClient from '@adobe/spacecat-shared-tokowaka-client';
 import { SuggestionDto, SUGGESTION_VIEWS, SUGGESTION_SKIP_REASONS } from '../dto/suggestion.js';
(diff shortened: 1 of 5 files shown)


src/controllers/suggestions.js
```diff
@@ -1125,6 +1126,24 @@ function SuggestionsController(ctx, sqs, env) {
...
+              index,
+              uuid: id,
+              message: `Illegal status transition: ${currentStatus} -> ${status}`,
+              statusCode: 400,
+            };
+          }
+
           suggestion.setStatus(status);
           if (status === SuggestionModel.STATUSES.SKIPPED) {
             isNewSkipTransition = true;
```
MysticatBot, line 1146: issue (blocking): **Missing integration test for modified endpoint behavior.**

The project CLAUDE.md rule states: "New or modified endpoints must include integration tests in `test/it/`". This PR materially modifies the bulk PATCH endpoint (previously-allowed transitions now return 400), but no integration test exercises the full-stack rejection path.

**Why it matters:** Unit tests import the real `isAllowedSuggestionTransition` from the bumped package, but the DB layer is fully mocked. An IT test against PostgreSQL would confirm the guard prevents the write end-to-end, covering any model-layer side effects the mock cannot.

**How to fix:** Add a case in `test/it/` (likely the existing suggestions IT file) that attempts an illegal transition (e.g. OUTDATED to FIXED) via the HTTP endpoint and asserts: (1) the per-item response is 400 with the expected message, and (2) the suggestion row in the DB retains its original status.

MysticatBot (changes requested):
Hey @alinarublea,

**Verdict: Request changes** - one conventions gap to close before merge.
**Complexity: HIGH** - medium diff; API surface + dependency signal.
_Changes: Adds a per-item transition-legality gate to the bulk PATCH suggestions/status handler, rejecting illegal status transitions (e.g. OUTDATED to FIXED) with a 400 before setStatus is called (4 files)._

## Must fix before merge
1. [Important] Missing integration test for the new rejection behavior - `src/controllers/suggestions.js:1136` (details inline)

<details>

- nit: The 12-line comment block (lines 1126-1137) re-explains what the ticket and function name already say. Consider trimming to just the NOTE about REJECTED ordering and the SITES ticket references - `src/controllers/suggestions.js:1126`
- suggestion: Log the rejected transition (suggestion ID + illegal pair) at info/warn level so the SITES-47286 enforce-mode flip can be triaged from Splunk if unexpected 400s appear - `src/controllers/suggestions.js:1136`
</details>

Note: CI checks are still pending - verify before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 28s | <b>Cost:</b> $5.98 | <b>Commit:</b> <code>943ee607f11ca1be6270199370810939774c53ef</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @alinarublea,

**Verdict: Approve** - all prior findings addressed, no new issues.
**Complexity: HIGH** - medium diff; dependency signal.
_Changes: Adds a per-item transition-legality gate to the bulk PATCH suggestions/status handler, rejecting illegal transitions (e.g. OUTDATED to FIXED) with a 400 before setStatus is called, and adds integration test coverage (5 files)._

<details>

- suggestion: Consider structured log fields (`context.log.info('rejected illegal status transition', { suggestionId: id, from: currentStatus, to: status })`) instead of template-literal interpolation, so Splunk queries can filter on discrete fields rather than regex against the message string - `src/controllers/suggestions.js:1132`
</details>

- Integration test for the new rejection behavior now exercises OUTDATED to FIXED end-to-end against PostgreSQL and confirms the DB row retains OUTDATED.
- Comment block trimmed from 12 lines to 3 lines.
- Rejected transitions are now logged at info level with suggestion ID and transition pair for Splunk traceability.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 59s | <b>Cost:</b> $5.49 | <b>Commit:</b> <code>3f53b3fd69bad43eb33cde6fd5d989d64fa6ebc3</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2989, opened 2026-08-03
fix: expose changed-field diff per audit-policy revision (SITES-49159)

sandsinh:
listRevisions now diffs each revision row against its predecessor (version - 1) at read time and returns it as changedFields, so callers don't have to reconstruct diffs from full-state snapshots client-side
- a reconstruction that also breaks across cursor-pagination boundaries. Predecessors are contiguous by version, so only the last row on a page needs one extra indexed lookup; the rest are already in the fetched page.

## Related Issues

Thanks for contributing!


sandsinh:
Deferring backend implementation - https://github.com/adobe/spacecat-api-service/pull/2989

Included changes in the PR - https://github.com/OneAdobe/experience-success-studio-ui/pull/2148/changes/a5f40a3e120e3d359a66482f39aecc8f5379dcae

PR #2992, opened 2026-08-04
fix(opportunities): revoke stale suggestion grants on resolve (SITES-49175)

sandsinh:

## Related Issues

[SITES-49175](https://jira.corp.adobe.com/browse/SITES-49175)

## Summary
- Old opportunities superseded by a newer opportunity of the same type (e.g. `broken-backlinks`) never had their outstanding suggestion grants cleaned up, so reopening the stale opportunity later created new grants on top of the un-revoked old ones, leaking PLG token-bucket capacity.
- Added `revokeExistingGrants(dataAccess, opportunity)` (`src/support/grant-suggestions-handler.js`): fetches only `NEW` suggestions for the opportunity, resolves their `grantId`s via `SuggestionGrant.splitSuggestionsByGrantStatus`, and revokes them via the existing `revokeGrants` helper — suggestions in any other status (APPROVED, FIXED, etc.) are never touched.
- Wired it into `patchOpportunity` (`src/controllers/opportunities.js`): fires when an opportunity genuinely transitions to `RESOLVED` on an `isSummitPlgEnabled` site, mirroring the existing `getByID` grant-handler pattern (non-fatal, logged via `ctx.log.warn` on failure).

## Test plan
- [x] Unit tests for `revokeExistingGrants` covering NEW-only fetch, no-grant no-op, and revoke-failure propagation (`test/support/grant-suggestions-handler.test.js`)
- [x] Unit tests for `patchOpportunity`'s revoke-on-resolve behavior: resolves+PLG (revokes), non-RESOLVED status change (no-op), already-RESOLVED re-save (no-op), non-PLG site (no-op), revoke failure handled gracefully (`test/controllers/opportunities.test.js`)
- [x] Full unit suite passes locally (`c8 mocha --spec=test/**/*.test.js --ignore=test/it/**`)
- [x] `eslint` clean on touched files

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```


src/controllers/opportunities.js
```diff
@@ -319,6 +325,16 @@ function OpportunitiesController(ctx) {
...
         opportunity.setUpdatedBy(profile.email || 'system');
         const updatedOppty = await opportunity.save(opportunity);
+
+        if (isResolving && await getIsSummitPlgEnabled(site, ctx, context)) {
+          try {
+            await revokeExistingGrants(dataAccess, updatedOppty);
+          /* c8 ignore next 3 */
+          } catch (err) {
+            ctx.log?.warn?.('Revoke existing grants handler failed', err?.message ?? err);
+          }
```
MysticatBot, line 339: issue (blocking): `getIsSummitPlgEnabled` is evaluated outside the inner try-catch that wraps `revokeExistingGrants`. If the entitlement lookup throws (transient network failure, service unavailable), the exception falls through to the outer catch which returns a 500 to the caller - even though `opportunity.save()` already succeeded and the status transition committed.

**Why it matters:** A transient failure in an unrelated service (entitlement) turns a successful mutation into a 500 response. The caller may retry, but `status !== opportunity.getStatus()` will be false on retry (status was already persisted as RESOLVED), so the opportunity is silently stuck with no revocation attempted.

**Fix:** Move the `getIsSummitPlgEnabled` call inside the inner try-catch so failures are caught and logged alongside revoke failures:
```js
try {
  if (isResolving && await getIsSummitPlgEnabled(site, ctx, context)) {
    await revokeExistingGrants(dataAccess, updatedOppty);
  }
} catch (err) {
  ctx.log?.warn?.('Revoke existing grants handler failed', err?.message ?? err);
}
```

src/controllers/opportunities.js
```diff
@@ -319,6 +325,16 @@ function OpportunitiesController(ctx) {
       if (hasUpdates) {
         opportunity.setUpdatedBy(profile.email || 'system');
         const updatedOppty = await opportunity.save(opportunity);
+
```
MysticatBot, line 328: issue (blocking): `getIsSummitPlgEnabled` requires `x-client-type: sites-optimizer-ui` in the request headers (or a ViewAsTrial override) - it returns false for other callers. This means opportunities resolved by backend processes, automation, or non-UI clients will skip grant revocation entirely, leaving the capacity leak this PR is designed to fix.

The PR description says it mirrors the existing `getByID` grant-handler pattern, which is consistent. But the granting path fires from the UI, so restricting revocation to the UI is only safe if no other code path can resolve an opportunity.

**Why it matters:** If any backend job, scheduler, or API consumer resolves opportunities without the UI header, stale grants persist.

**Fix (if backend resolves exist or are planned):** Replace with a simpler entitlement-only check that verifies the site is PLG-tier without requiring a client-type header. If all resolves today come exclusively from the UI, add a code comment documenting that assumption so future maintainers know the coupling is intentional.

MysticatBot (changes requested):
Hey @sandsinh,

**Verdict: Request changes** - two structural issues in the revoke-on-resolve wiring.
_Changes: Adds grant revocation when opportunities transition to RESOLVED on PLG-enabled sites, preventing stale grants from leaking token-bucket capacity (4 files)._

## Must fix before merge
1. [Important] `getIsSummitPlgEnabled` outside inner try-catch - entitlement lookup failure turns successful PATCH into 500 - `src/controllers/opportunities.js:335` (details inline)
2. [Important] Revocation gated on UI client-type header - backend-initiated resolves still leak grants - `src/controllers/opportunities.js:328` (details inline)

<details>

- nit: `splitSuggestionsByGrantStatus` result destructured without defensive default; if it returns unexpected shape, `revokeGrants` receives `undefined` - `src/support/grant-suggestions-handler.js:252`
- nit: tautological assertion `expect(true).to.be.true` in early-return test proves nothing; assert no downstream call was made instead - `test/support/grant-suggestions-handler.test.js:1403`
- suggestion: include opportunity ID and site ID in the warn log for operational debugging - `src/controllers/opportunities.js:333`
- nit: test mutates shared `opptys[0].status` without restoring it in afterEach - `test/controllers/opportunities.test.js:741`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 2s | <b>Cost:</b> $5.89 | <b>Commit:</b> <code>a49b231afa5c7e79cf9675287f209386528b282a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

sandsinh (commented):
Addressed all review feedback:
- Wrapped the entitlement lookup (`getIsSummitPlgEnabled`) in its own try/catch on resolve so a lookup failure can't turn a successful PATCH into a 500 (`src/controllers/opportunities.js`)
- Dropped the UI-only `requestContext` gate on the revocation call so backend-initiated resolves also revoke stale grants, not just ones going through the sites-optimizer-ui client (`src/controllers/opportunities.js`)
- Added a defensive default when destructuring the `splitSuggestionsByGrantStatus` result (`src/support/grant-suggestions-handler.js`)
- Included opportunity/site IDs in the revoke-failure warn log for operational debugging
- Replaced the tautological `expect(true).to.be.true` assertion with a real check on the early-return value (`test/support/grant-suggestions-handler.test.js`)
- Fixed the test that mutated the shared `opptys[0].status` fixture without restoring it (`test/controllers/opportunities.test.js`)
- Added tests covering the backend-initiated-resolve fix and the entitlement-lookup-failure fix

All 137 tests in the affected files pass; lint is clean.

MysticatBot (approved):
Hey @sandsinh,

**Verdict: Approve** - all prior findings addressed, no new blocking issues.
_Changes: Adds grant revocation when opportunities transition to RESOLVED on PLG-enabled sites, preventing stale grants from leaking token-bucket capacity (4 files)._

<details>

- nit: `revokeExistingGrants` guard clause returns silently when DAL entities are missing (a wiring-bug scenario); consider logging or throwing so the outer catch can surface it for diagnostics - `src/support/grant-suggestions-handler.js:243`
</details>

- `getIsSummitPlgEnabled` now inside inner try-catch; entitlement lookup failure no longer turns a successful PATCH into 500
- requestContext gate dropped; backend-initiated resolves now revoke stale grants
- Defensive default added for `splitSuggestionsByGrantStatus` result
- Tautological assertion replaced with meaningful check on early-return value
- Opportunity/site IDs now included in the revoke-failure warn log
- Shared fixture mutation guarded with try/finally

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 14s | <b>Cost:</b> $2.82 | <b>Commit:</b> <code>55f57b468a801f3f7f12c9ef59c95101b40baf5a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2993, opened 2026-08-04
feat: retire legacy un-paginated GET /sites response

### LINKED ISSUE #2591: Sunset un-paginated GET /sites: migrate remaining legacy-shape consumers
## Background

`GET /sites` recently gained optional pagination (`limit`/`cursor`) via PR adobe/spacecat-api-service#2497 (commit `ac0dab47`, SITES-41367). When called **without** `limit`/`cursor`, the endpoint still returns the legacy flat array (full `fetchAllPages` scan) for backwards compatibility, and emits a `[sites][legacy-shape]` log line recording the caller's `clientId`.

We want to sunset the un-paginated path. This issue tracks the remaining consumers that still call `GET /sites` without pagination params so we can migrate them before removing the legacy shape.

## Method

Coralogix, prod (`spacecat-services-prod`, account `aws-640168421876`, function `/spacecat-services/api-service/latest`), aggregating the `[sites][legacy-shape]` marker by `clientId`.

- Window: **2026-06-08 → 2026-06-11** (last 3 days)
- The legacy-shape log line carries only `clientId` — no `User-Agent`, headers, or source IP (all `null`).
- S2S `clientId` values resolved to consumer names via the prod `consumers` Postgres table (`idx_consumers_client_id`).

### Named S2S consumers (migration targets)

| Calls | Consumer name | clientId | Notes |
|------:|---------------|----------|-------|
| 169 | **s2s-llmo-data-retrieval-service** | `6a342b9f35a84f4ab60e4f7f82dfeda0` | DRS — BP monitor, site-name maps, schedulers |
| 71 | **AMS Sites Optimizer ESE Validator Script and Bot** | `a3c925e9aaeb43e7b96c9f7b5c1d7f3c` | Registered 2026-06-03 |
| 34 | **S2S LLMO Usage Report Generator** | `500bb9199427463f9c7f6dd022d2248b` | LLMO usage/MAU reporting |
| 8 | **S2S AEM-Trustcenter** | `2f61fa914f324902b9b5bc0c96bfcb4f` | AEM Trustcenter integration |
| 1 | **AEM Engineering Operations DaaS ETL** | `1980219afe204b13b84116f118bcd33f` | Registered 2026-06-10 — already on legacy shape |

All five are `ACTIVE`, under IMS org `908936ED5D35CC220A495CD4@AdobeOrg`, and hold `site:readAll`.

### Admin path — `admin-bypass` (4,539 calls)

The vast majority of un-paginated calls came through the **shared admin credential** (admin x-api-key / admin IMS JWT). These all log `clientId=admin-bypass`, and because the legacy-shape log line carries no user-agent/source, **they cannot be attributed to a specific consumer from these logs alone**. This bucket is internal tooling / UI / scripts / monitors sharing the admin key.

## Open items

- [x] **Attribute the `admin-bypass` bucket.** It is ~95% of legacy-shape volume and is currently opaque. Break it down via **API Gateway access logs** (they carry `User-Agent` + source IP keyed by the same `requestId`). Consider also enriching the `[sites][legacy-shape]` log line with `User-Agent` so future attribution is possible directly.
- [x] Migrate the 5 named S2S consumers to pass `?limit=&cursor=` (paginated envelope):
  - [x] s2s-llmo-data-retrieval-service (DRS)
  - [x] AMS Sites Optimizer ESE Validator Script and Bot
  - [x] S2S LLMO Usage Report Generator
  - [x] S2S AEM-Trustcenter
  - [x] AEM Engineering Operations DaaS ETL
- [x] Flag to the **DaaS ETL** owners that their newly registered consumer (2026-06-10) is already integrating against the legacy shape — fix before sunset.
- [x] Run the same breakdown for **stage/dev** before removal.
- [x] Define the deprecation/sunset timeline and remove the legacy flat-array path once consumers are migrated.

## Notes

- This analysis is **prod only**.
- The legacy path performs a full `fetchAllPages` scan (filtering internal orgs to stay under the 6MB Lambda limit), which is the cost motivating the sunset.

martinst06:
## What & why

Sunsets the legacy un-paginated branch of `GET /sites` (issue #2591 — a SITES-46094). The legacy path did a full `fetchAllPages` scan, filtered internal orgs to stay under the 6MB Lambda limit, returned a top-level array, and emitted a `[sites][legacy-shape]` marker used as the sunset gate. All tracked S2S consumers have been migrated.

A call with no `limit`/`cursor` now flows through the existing cursor-paginated path and returns the **first page** (`limit` defaults to 100, no cursor) as the `{ sites, pagination }` envelope — the same shape as an explicit paginated request.

## Changes

- **`src/controllers/sites.js`** — remove the `paginated` gate and the legacy `else` branch (`fetchAllPages` scan, `DEFAULT_ORGANIZATION_ID`/`ORGANIZATION_ID_FRIENDS_FAMILY` exclusion, `[sites][legacy-shape]` log). No params ⇒ first page envelope. `baseUrlContains` search path is unchanged.
- **`docs/openapi/sites-api.yaml`** — drop the deprecated `SiteList` array branch from the `GET /sites` `oneOf`; update the endpoint / `limit` descriptions. `SiteList` itself stays (still used by other endpoints). Rebuilt `docs/index.html`.
- **`test/controllers/sites.test.js`** — delete the two `[sites][legacy-shape]` tests and the legacy flat-array test; update no-param tests to assert the envelope.
- **`test/it/shared/tests/sites.js`** — the no-param admin / S2S `readAll` tests now assert the `{ sites, pagination }` envelope with **no** exclusion (all 7 seeded sites on the first page).

## Behavioral / compatibility note

⚠️ Breaking for any remaining no-param caller: the response is now `{ sites, pagination }` instead of a top-level array, and the first page is **no longer filtered** (internal/friends-family orgs now appear). The 6MB motivation for the exclusion is moot on a bounded page. Per #2591 the named S2S consumers were migrated; the opaque `admin-bypass` bucket should be watched post-deploy.

## Testing

- ✅ `npx mocha test/controllers/sites.test.js` — 435 passing
- ✅ `npm run lint` (changed files) — clean
- ✅ `npm run docs:lint` — valid; `docs:build` regenerated `docs/index.html`
- ⏳ PostgreSQL IT suite — deferred to CI (local run blocked on interactive ECR/KLAM auth)

Refs 
#2591
SITES-46094

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```


MysticatBot (approved):
Hey @martinst06,

**Verdict: Approve** - clean sunset of a legacy code path with thorough test coverage.
**Complexity: HIGH** - medium diff; breaking API surface change (response shape).
_Changes: Removes the legacy un-paginated GET /sites branch and unifies all calls through the cursor-paginated path, returning a `{ sites, pagination }` envelope by default (7 files)._
Note: Recommend a human read before merge - this change modifies a shared API contract (`docs/openapi/sites-api.yaml`) and an architectural doc path. The bot review is a complement to, not a replacement for, a human read here.

<details>

- suggestion: Consider a short-lived `log.info` when `!hasText(limitParam) && !hasText(cursor)` to track remaining no-param callers during the post-deploy watch period mentioned in the PR description - `src/controllers/sites.js:~617`
- nit: The "gets all sites with slim DTO (default first page)" test asserts `body` has keys `sites` and `pagination` but does not assert pagination values; the later "returns the first-page envelope" test does. Either remove the overlap or add `expect(body.pagination).to.deep.equal(...)` to the earlier test - `test/controllers/sites.test.js:~975`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 42s | <b>Cost:</b> $4.48 | <b>Commit:</b> <code>1c3130d2d1e09e1e41e625f424c13c97dc8916e8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

martinst06:
Thanks @MysticatBot — both non-blocking items noted; declining both, with reasoning:

1. **No-param `log.info` to track remaining callers** — this was a deliberate design decision on this PR: we *dropped* the no-param observability log rather than keep a marker. The point of the sunset is that a no-param `GET /sites` is no longer a special "legacy" case — it's just a normal paginated request (first page), so re-adding a marker would reintroduce a variant of the `[sites][legacy-shape]` line we just removed. It also wouldn't serve the stated goal: a server-side "no-param call happened" log doesn't reveal whether a client actually broke on the shape change (that's a client-side contract concern), and caller attribution for the sunset was already completed in #2591 via the legacy-shape log + API Gateway access logs. Not re-adding a log.

2. **Test overlap on the slim-DTO test** — intentional separation of concerns rather than redundancy. `gets all sites with slim DTO (default first page)` owns the field-projection assertions (list items exclude heavy fields) and only checks the envelope *keys* incidentally to reach `body.sites`; `returns the first-page envelope when no limit or cursor is provided` owns the pagination *contract* (`deep.equal` on `{ limit, cursor, hasMore }` plus the `Site.all(..., { returnCursor: true })` routing). Duplicating the pagination-value assertion into the DTO test would couple two unrelated intents. Leaving as-is.

PR #2996, opened 2026-08-04
feat(ai-seo): AI SEO API revision 

dmitriigulkin-semrush:
Branch AISEO-2011 adds AI Visibility v1 reporting APIs:

- 20 new GET endpoints for cited pages/sources, competitor stats, and FTS-based topic, prompt, brand, and source-domain reporting.
- Includes data, totals, and export variants.
- Adds report:read and llmo/can_view authorization mappings.
- Updates OpenAPI schemas, examples, and generated docs.
- Upgrades generated ai-seo-ts SDK from v0.101.3 to v0.102.1.

This is the first step of sunsetting initial POC API, PR for the frontend part is coming soon

After both FE and BE get merged, I'll prepare another with removing initial API, removing tests and open api documentation

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```


calvarezg:
## Multi-persona review — AISEO-2011 (v1 reporting APIs)

Reviewed the non-generated surface (handlers, controller, routes, OpenAPI, tests); `third-party/ai-seo-ts` SDK + `docs/index.html` treated as codegen. I also deep-checked the untested paths for real defects rather than just flagging coverage.

**Recommendation: changes recommended** — 0 critical, 1 major. Not a blocker; the one Major has a one-line fix using a helper already in the repo.

### Major

**`topicId` interpolated into the QL string without validation** — `src/support/ai-visibility/handlers/v1/prompt/brand-prompts.js` (`buildBrandPromptsDimensionFilterQl`)

```js
if (topicId) {
  filters.push(`topic_hash = ${topicId}`);        // raw, unquoted, unescaped
}
if (targetUrl) {
  filters.push(`target_url = "${escapeQlString(targetUrl)}"`);  // escaped
}
```

`topicId` comes straight from `sp.get('topicId')` and is interpolated raw into `dimension_filter_ql` (list **and** totals requests), while the sibling `targetUrl` on the next line is escaped. There is no runtime request validation on this gateway (no OpenAPI validator middleware is wired; validation is hand-rolled per controller), so the OpenAPI `^\d+$` pattern is documentation only. `?topicId=0 OR mentions>0` produces `topic_hash = 0 OR mentions>0`, and combined with the `AND target_url=…` clause the injected `OR` defeats both filters.

- **Impact is bounded**, not a breach: authenticated (`report:read` + `llmo/can_view`), read-only, and the target `domain` is already caller-controlled — so this is upstream-QL filter manipulation, not cross-tenant data access.
- **The safe pattern already exists and is the team's own.** `grpc-utils.js` ships `resolveTopicIdsDimensionFilter` / `resolveTopicIds` (`TOPIC_HASH_ID_PATTERN = /^\d+$/`, manual 400s, doc comment: *"injection-safe numeric ids only"*). The legacy POC handler being sunset (`handlers/brands.js`) uses it and is unit-tested; the new v1 handler skips it. The raw line is pre-existing within v1, but this PR refactors it into the reused helper and adds an escaped field beside it — the right moment to close it.

**Fix:** route `topicId` through `resolveTopicIds` (it reads the singular `topicId` too), or guard with `TOPIC_HASH_ID_PATTERN` and return 400 on non-numeric before interpolating.

### Minor

- **Filter-building + reshaping logic has no test (regression risk, not a live bug).** I traced the two paths where the coverage gap could hide a real defect and both are correct today: (1) every `body: { data: json.X ?? [] }` key matches the proto's emitted field (incl. `source_domains` → `sourceDomains`); (2) v1 brand-prompts reads `topicId` singular and the OpenAPI param is `topicId` singular — no silent drop. The risk is forward-looking: the escaping logic (which gates the Major above) and these field-name wirings would break silently on a future SDK bump/edit. Most `build*FilterQl` helpers sit inside `/* c8 ignore */`, so such a break would be invisible to coverage. Move each `export function build*` above its `c8 ignore start` and add a small unit table (topicId-only / targetUrl-only / both / neither / quote+backslash escaping).
- **gap-\* self-as-competitor fallback** duplicated verbatim in 6 files; the fallback uses raw `domain` while the populated path normalizes via `brandTarget` (trim+lowercase). Extract a helper + comment.
- **`competitors-stats.js`** omits the `|| LLM_ENUM.ALL` engine default its siblings use.
- **`parseLimitOffset`** has no upper bound — clamp `limit` (e.g. 500) to cap upstream load.
- **`prompts-by-topic-ids`** bypasses `resolveTopicIds`; the OpenAPI `maxItems: 50` isn't enforced (no injection — typed `repeated uint64`).
- **400 responses** return the raw `fromJson` error `message` (leaks proto field names); prefer a static message + server-side log.
- **Unknown gRPC errors** are rethrown with no `log.error` across the new endpoints — hard to diagnose in production.

### Question

- gap-\* now sends `[{ domain, name: domain }]` (target as its own competitor) instead of `[]` when no competitors are supplied. Is that the intended upstream semantics? A "gap vs yourself" query may return degenerate results. Worth confirming + a test for the empty branch.

### Verified clean

AuthZ complete — every new route is in **both** `required-capabilities.js` (`report:read`) and `facs-capabilities.js` (`llmo/can_view`). No SSRF (nothing fetched server-side). `searchQuery` / `textFilter` / `targetUrl` / `domain` are quoted+escaped. Two suspected bugs checked false: `cited-pages-totals` `filter_ql` and FTS `dimensionFilterQl` both match real proto fields (no silent filter drop); ROUTE_MAP reorder is cosmetic (exact-key object map). Response reshaping field names all match the schema.

---
*Reviewed by parallel Architect / DBA / Security / Tester + Junior-Dev agents; every Major/Minor verified against the diff and the generated proto schema.*

dmitriigulkin-semrush:
hey @calvarezg , thanks a lot for the review!

i've adressed the major [here](https://github.com/adobe/spacecat-api-service/pull/2996/changes/d0ff510c0b3938ca6f8c8ac9effc1e4cb661c538)

> Filter-building + reshaping logic has no test (regression risk, not a live bug). I traced the two paths where the coverage gap could hide a real defect and both are correct today: (1) every body: { data: json.X ?? [] } key matches the proto's emitted field (incl. source_domains → sourceDomains); (2) v1 brand-prompts reads topicId singular and the OpenAPI param is topicId singular — no silent drop. The risk is forward-looking: the escaping logic (which gates the Major above) and these field-name wirings would break silently on a future SDK bump/edit. Most build*FilterQl helpers sit inside /* c8 ignore */, so such a break would be invisible to coverage. Move each export function build* above its c8 ignore start and add a small unit table (topicId-only / targetUrl-only / both / neither / quote+backslash escaping).
> gap-* self-as-competitor fallback duplicated verbatim in 6 files; the fallback uses raw domain while the populated path normalizes via brandTarget (trim+lowercase). Extract a helper + comment.
> competitors-stats.js omits the || LLM_ENUM.ALL engine default its siblings use.
> parseLimitOffset has no upper bound — clamp limit (e.g. 500) to cap upstream load.
> prompts-by-topic-ids bypasses resolveTopicIds; the OpenAPI maxItems: 50 isn't enforced (no injection — typed repeated uint64).
> 400 responses return the raw fromJson error message (leaks proto field names); prefer a static message + server-side log.
> Unknown gRPC errors are rethrown with no log.error across the new endpoints — hard to diagnose in production.

all fair, i'd really like to adress them all right after initial PoC is removed, there are quite a few things that could be easily absracted/simplified while keeping 1/1 mapping with SR API, so we don't spend as much resources supporting all the changes in downstream API. would you mind if i park as additional jira task?


> gap-* now sends [{ domain, name: domain }] (target as its own competitor) instead of [] when no competitors are supplied. Is that the intended upstream semantics? A "gap vs yourself" query may return degenerate results. Worth confirming + a test for the empty branch.

yep, that's just a quirk of the api, i'll document it later. no competitors for gap methods is a valid case, but passing them as `[]` will always trigger grpc validation 😭

calvarezg:
Thanks @dmitriigulkin-semrush — two of the three are settled, one still open.

**gap-\* fallback:** makes sense, the empty-list → gRPC-validation issue explains it. Happy to treat the self-as-competitor as an intended quirk; a doc comment later is enough. Resolved.

**Minors → follow-up Jira:** fine by me. None is blocking, and they overlap with the PoC-removal cleanup where the abstraction will be cheaper. Park them.

**Major — still exploitable, the fix is a no-op here.** The change was:

```js
filters.push(`topic_hash = ${escapeQlString(topicId)}`);
```

`escapeQlString` only escapes `"` and `\` (to keep a value inside a double-quoted string literal). But `topic_hash = <id>` is **unquoted** — a numeric comparison, not a string literal — and the injection payload has no quotes or backslashes to escape:

- `?topicId=0 OR mentions>0`
- `escapeQlString("0 OR mentions>0")` → `"0 OR mentions>0"` (unchanged)
- result: `topic_hash = 0 OR mentions>0` → the `OR` still lands.

Escaping is the wrong category of fix for an unquoted numeric field. It needs **validation** that `topicId` is digits-only, which the repo already has:

```js
// grpc-utils.js: TOPIC_HASH_ID_PATTERN = /^\d+$/
if (topicId) {
  if (!TOPIC_HASH_ID_PATTERN.test(topicId)) {
    return { status: 400, body: { error: 'invalid_request', message: 'topicId must be a non-negative integer' } };
  }
  filters.push(`topic_hash = ${topicId}`);
}
```

or route `topicId` through the existing `resolveTopicIds` helper (it reads the singular `topicId` and enforces `^\d+$` + the 50-cap, returning a 400 body on bad input). Either closes it; the string-escape does not.

calvarezg (approved):
Approving — the Major is genuinely resolved.

**Injection fixed.** `buildBrandPromptsDimensionFilterQl` now validates `topicId` against `/^\d+$/` and returns a 400 on non-numeric input, so only a digits-only value reaches `topic_hash = ${topicId}`. `?topicId=0 OR mentions>0` now fails the regex → 400. `handleBrandPrompts` guards the new `string | {status,body}` return type. This is the right fix (validation, not the earlier escape).

**One non-blocking follow-up** — `handleBrandPromptsExport` (`brand-prompts-export.js:70`) calls the same builder but doesn't guard the new return type:

```js
dimension_filter_ql: buildBrandPromptsDimensionFilterQl(sp),   // can now be a {status,body} object
```

On an invalid `topicId` at the export endpoint it still returns 400, but via the `fromJson` throw — so the caller gets the raw bufbuild error message instead of the clean `"topicId must be a non-negative integer"`, and the two sibling endpoints answer the same bad input differently. Not a security or data issue (nothing malicious reaches upstream). Worth the same 3-line guard the data handler got:

```js
const dimensionFilterQl = buildBrandPromptsDimensionFilterQl(sp);
if (typeof dimensionFilterQl !== 'string') {
  return dimensionFilterQl;
}
```

Fine to fold into the follow-up Jira with the other Minors if you'd rather. The gap-* self-competitor quirk (validation workaround) and the parked Minors are all acknowledged — no objection to landing this.

PR #2850, opened 2026-07-18
feat(llmo-akamai): CUSTOM-default OAE onboarding (PUT deploy + caching/scope gates)

ABHA61:
## What
Reworks the Akamai "Optimize at Edge" (BYOCDN) onboarding for **CUSTOM-default** properties.

**Rule builder (`llmo-akamai-utils.js`)**
- Wildcard the UA match values (`*GPTBot*`) so they match real agent strings, not only an exact `GPTBot`.
- Add a Caching behavior (Honor origin Cache-Control + Expires) to the OAE rule **only when the property's default rule has none**; `cacheId` is always present.
- Append the managed "Optimize at Edge" wrapper **last** so its origin + cacheId win (Akamai evaluates siblings top-down, last match wins).
- OAE origin host comes from `env.EDGE_OPTIMIZE_EDGE_DOMAIN` (dev/stage/live.edgeoptimize.net; defaults to live); `matchSan` (`*.edgeoptimize.net`) covers all three.

**Controller (`llmo-akamai.js`)**
- deploy/plan: **JSON-Patch → full-tree PUT** (`updateRuleTree`, pinned `ruleFormat`; plan uses the client's `dryRun`).
- **Scope gate**: reject a non-CUSTOM default origin (`400`) *before* creating a version (no orphan version).
- Optional `baseVersion` (copy from a specific version; rejects `0`); caller email in the activation note.
- **Activation resilience**: the PAPI activation POST frequently exceeds the client request timeout on large trees (and a retry then returns `422 already-activated`) — in both cases Akamai has already queued the activation, so `activate` recovers by reporting the in-flight/active activation for the same version+network instead of a false failure.

## Why
Validated end-to-end on `tokowaka-aem.stage.adobe.com`: an OAE-rule Caching behavior on a property that (correctly) defaults HTML to `no-store` overrides it and makes the optimized path cacheable, so Akamai replays a stale/passthrough copy to bots (v40 = 0/20 optimized at IAD). Conditional caching + full-tree PUT + the CUSTOM scope gate + append-last ordering fix it — AI bots receive `x-edgeoptimize-request-id` + `no-store`, humans pass through.

## Design
High-Level Design: [Optimize at Edge — Akamai BYOCDN Automation (HLD)](https://wiki.corp.adobe.com/spaces/AEMSites/pages/3973412135/Optimize+at+Edge+-+Akamai+BYOCDN+Automation+High-Level+Design)

## Depends on
[...]

PR #2820, opened 2026-07-15
fix(llmo-akamai): apply Optimize-at-Edge rules via JSON Patch, not full-tree PUT

ABHA61:
> [!NOTE]
> **Unblocked.** Depends on adobe/spacecat-shared#1813, which is **merged and published as `@adobe/spacecat-shared-akamai-client@1.1.0`** (adds `patchRuleTree`). The dependency is bumped `1.0.2 → 1.1.0` in this PR, so `deploy`/`plan` resolve the real `patchRuleTree` at runtime. Ready for review.

## What
Switches the Akamai Optimize-at-Edge **deploy** from a full-tree PUT to a server-side **JSON Patch**, and makes **plan** dry-run that same patch.

## Why
Deploy did `GET rules → merge → PUT (validateRules=true)`. Under rule format `latest`, the GET re-expands untouched `origin` behaviors (SSL/TLS fields under "Use Platform Settings") and the PUT re-stores them, so PAPI rejects the tree with `incompatible_features` on behaviors we never touched. Reproduced on the POC property: the base version validated **0 errors** in Property Manager, yet the wizard's deploy returned **2 errors** on the default origin + a child origin. Rebasing off the clean *active* version still failed — confirming it's the GET→PUT round-trip, not the base version. See LLMO-6252.

A server-side PATCH applies only our deltas to the *stored* tree, so existing behaviors are never re-serialized and the whole class of false rejection disappears (also robust to advanced behaviors / per-rule UUIDs / frozen-vs-latest formats).

## Changes
- **deploy**: `createVersion` → `buildRuleTreePatch` → `patchRuleTree` (guarded by the new version's `etag` via If-Match). No more full-tree PUT.
- **plan**: dry-runs the same patch so Review reflects the real deploy validation. Previously plan only echoed the base version's GET-time warnings — a merged tree could look clean here yet be rejected at deploy. Degrades gracefully (`validated: false`) if the dry-run itself can't run.
- **buildRuleTreePatch** (`llmo-akamai-utils.js`): builds the JSON Patch; idempotent by **trimmed** rule name — also removes a legacy `"Optimize at Edge "` (trailing space) that the old merge left duplicated.
- **`mergeIntoTree`** now matches managed rules by trimmed name too, so the plan preview agrees with deploy.
- **OpenAPI**: the plan response's `validated` / `errors` / `warnings` fields are declared; `newVersion` is surfaced on both deploy error paths.

## Tests
[...]

PR #3000, opened 2026-08-05
fix(llmo-akamai): add per-call timeout headroom, retry-resume, and fuller PAPI error logging

ABHA61:
## What
Two independent fixes found while investigating an Akamai OAE onboarding incident on a property
with many rules (`prp_729288`):

1. **Re-wire `plan()`/`deploy()` from the full-tree PUT back to a JSON-Patch-based merge.**
   `buildRuleTreePatch` + `AkamaiClient#patchRuleTree` already existed in this codebase — built in
   #2820 specifically to avoid PAPI rejecting untouched, GET-expanded behaviors, then reverted 4
   days later in #2850 (bundled into an unrelated caching fix, with no stated reason). A patch is a
   delta: existing rules/behaviors are never re-serialized. All of #2850's CUSTOM-SSL scope gate,
   conditional-caching decision, and append-last ordering are preserved unchanged — only the "how
   do we ship it to Akamai" call changes.
2. **`deploy()` accepts an optional `retryVersion`.** When a prior deploy attempt already created a
   property version but failed to write its rules into it (e.g. a timeout on a large property), the
   caller can resume writing into THAT version directly instead of minting a fresh one on every
   retry. Previously `baseVersion` was always re-cloned, so clicking Retry repeatedly left a trail
   of permanently-empty orphan versions on the customer's property.

Also logs the full (bounded) `papiErrors` content on a real PAPI rejection, not just a count —
previously undiagnosable from logs alone.

## Why
Splunk logs for `prp_729288` showed every `plan`/`deploy` call dying with the identical
`Request timeout after 10000ms` — confirmed root cause is a missing per-call timeout override in
`AkamaiClient` (fixed in adobe/spacecat-shared#1864). Independently, reading the actual code paths
surfaced these two additional issues on top of that: the round-trip class of PAPI rejection that
JSON Patch was built to prevent, and the wasteful version-churn on retry.

## Depends on
adobe/spacecat-shared#1864 (rule-tree call timeout) for the root-cause fix. This PR is
independently backward-compatible either way (`retryVersion` is optional), but should have the
dependency bumped once that merges and publishes.

## Testing
- Updated `llmo-akamai.test.js` for the patch-based path (etag-per-target-version, ops instead of a
  full merged tree) — 91/91 in this file, 136/136 across `llmo-akamai*.test.js`.
- New tests for `retryVersion` (skips `createVersion`, resumes into the given version, rejects
  non-integer/zero).
- New test asserting full `papiErrors` detail is logged, not just a count.
- Full repo suite: `npm test` — 16,232 passing, 0 failing, coverage thresholds met.
- `npm run lint`, `npm run docs:lint`, `npm run docs:build` all clean; `docs/index.html` rebuilt.

## Follow-up (separate PR, project-elmo-ui)
Wire `retryVersion` through the Akamai onboarding wizard's Deploy step so a failed attempt's
created version is tracked and resent on Retry.


ABHA61:
⚠️ **Temporary commit added for testing** (`659b11f1`): `@adobe/spacecat-shared-akamai-client` is pointed at a gist-hosted tarball built from adobe/spacecat-shared#1864's branch, since that PR hasn't merged/published yet. This lets the timeout fix be exercised end-to-end here before the real npm release exists.

**This must be reverted before merging this PR** — once #1864 merges and publishes, swap `package.json`/`package-lock.json` back to the real published version.

ABHA61:
**Reverted the JSON-Patch adoption** (`bafe5853`), back to the full-tree PUT.

Tested the patch-based merge against real Akamai on stage: a first-time onboard (add-only patch) applied cleanly, but every re-onboard (remove-then-add) was rejected outright with a generic `400 "The system cannot understand your request"` — 100% reproducible, not intermittent. This is almost certainly the real reason #2820 (which introduced the same patch approach) was quietly reverted 4 days later in #2850 with no stated reason at the time.

`retryVersion` and full `papiErrors` logging are unaffected — both still work, just via `updateRuleTree` (PUT) instead of `patchRuleTree`.

ABHA61:
Superseded by #3013. That branch's earlier JSON-Patch attempt was reverted to a full-tree PUT here; #3013 rebuilds a working PATCH-delta deploy and also carries the retry-resume, per-call timeout, and fuller PAPI error logging from this PR, validated end-to-end on dev. Closing in favor of #3013.

PR #3006, opened 2026-08-06
feat(llmo): serve brand-claims history by ISO week (LLMO-6808)

akshaymagapu:
## What

Teaches `GET /sites/{siteId}/llmo/brand-claims` to read the new per-ISO-week Brand Claims layout produced by the delivery side (mystique):

```
brand_claims/llmo/{siteId}/{YYYY-Www}/data.json.gz
```

- **No `date`** → `ListObjectsV2` on the site prefix, presign the lexically-greatest `YYYY-Www` week. If no week folder exists yet, fall back to the legacy flat `brand_claims/llmo/{siteId}/data.json.gz`.
- **`date=<any day in the week>`** (e.g. `2026-04-22`) → presign that week directly, no list. Invalid date → 400.
- **`model`** → unchanged legacy flat `{model}.json.gz` (mystique never writes model-partitioned week folders).

## Why

Once claims runs weekly, the delivery side writes one object per run under an ISO-week folder instead of overwriting a single `data.json.gz`, so history is preserved. This is the read half: the endpoint serves the latest run by default and lets a caller select an older week by date.

ISO week is the unit because it mirrors the run-lock key (`site_id + platform + week + year`) and the one-run-per-week cadence.

## Backward compatibility

- Default fetch (no params) returns the latest run — **same shape and behaviour as today**; no client change required.
- The legacy flat key is still served via fallback until a site has been written under the new layout, so nothing 404s during migration.
- Response envelope (`{ siteId, model, presignedUrl, expiresAt }`) and the HeadObject existence check are unchanged.

## Deploy ordering

This is the **read** half and is safe to deploy first: before any week folder exists it falls back to the flat key (current behaviour). The delivery-side change (mystique) that stops writing the flat key must land **after** this. Companion PR: mystique `feat/llmo-6808-claims-history-s3`.

## Tests

`test/controllers/llmo/brand-claims.test.js` — latest-week selection, non-week-folder filtering, flat-key fallback, `date` → week, invalid date 400, model flat path, existing 400/404 paths, and `toIsoWeek` unit cases (ISO week, year boundary, zero-pad). **19 passing.** OpenAPI `date` param added; `npm run docs:lint` valid.

Ticket: LLMO-6808 (epic LLMO-5741 Brand Claims: GA Readiness).

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["David Aurelio", "Dominique Jäggi"]
```

### DIFF
--- a/src/controllers/llmo/brand-claims.js
+++ b/src/controllers/llmo/brand-claims.js
@@ -20,6 +20,10 @@ import { dateToIsoWeek } from '../../support/elements/week-utils.js';
 const CLAIMS_PREFIX = 'brand_claims/llmo';
 const WEEK_RE = /^\d{4}-W\d{2}$/;
 const DATE_RE = /^\d{4}-\d{2}-\d{2}$/;
+// `model` is interpolated into the S3 key, so constrain it to alphanumerics,
+// dots, hyphens, underscores — no `/` — to prevent using HeadObject as an
+// object-existence probe across arbitrary key paths.
+const MODEL_RE = /^[\w.-]+$/;
 
 /**
  * Latest run key for a site: the lexically-greatest `YYYY-Www` folder under the
@@ -81,17 +85,21 @@ export async function handleBrandClaims(context) {
     return badRequest('S3 bucket is not configured for this environment');
   }
 
-  if (date !== undefined && (!DATE_RE.test(date) || Number.isNaN(new Date(date).getTime()))) {
-    return badRequest('Invalid date parameter: expected YYYY-MM-DD format');
+  if (model !== undefined && !MODEL_RE.test(model)) {
+    return badRequest('Invalid model parameter');
   }
 
-  // Model files are managed flat (not week-partitioned); `date` resolves directly
-  // to its week; otherwise default to the legacy flat key and upgrade it to the
-  // latest week (via a list) inside the try below.
+  // Model files are managed flat (not week-partitioned) and take precedence;
+  // `date` resolves directly to its week (validated only here, where it is
+  // actually used); otherwise default to the legacy flat key and upgrade it to
+  // the latest week (via a list) inside the try below.
   let s3Key;
   if (model) {
     s3Key = `${CLAIMS_PREFIX}/${siteId}/${model}.json.gz`;
   } else if (date) {
+    if (!DATE_RE.test(date) || Number.isNaN(new Date(date).getTime())) {
+      return badRequest('Invalid date parameter: expected YYYY-MM-DD format');
+    }
     s3Key = `${CLAIMS_PREFIX}/${siteId}/${dateToIsoWeek(date)}/data.json.gz`;
   } else {
     s3Key = `${CLAIMS_PREFIX}/${siteId}/data.json.gz`;
--- a/src/controllers/llmo/brand-claims.js
+++ b/src/controllers/llmo/brand-claims.js
@@ -15,26 +15,10 @@ import {
 } from '@adobe/spacecat-shared-http-utils';
 import { HeadObjectCommand, ListObjectsV2Command } from '@aws-sdk/client-s3';
 import { cachedOk } from '../../support/cached-response.js';
+import { dateToIsoWeek } from '../../support/elements/week-utils.js';
 
 const CLAIMS_PREFIX = 'brand_claims/llmo';
 const WEEK_RE = /^\d{4}-W\d{2}$/;
-
-/**
- * ISO-week key segment (`YYYY-Www`) for a date, matching the delivery side
- * (mystique) that writes one export per run under that folder.
- *
- * @param {Date} d - a valid Date
- * @returns {string} e.g. '2026-W17'
- */
-export function toIsoWeek(d) {
-  const date = new Date(Date.UTC(d.getUTCFullYear(), d.getUTCMonth(), d.getUTCDate()));
-  date.setUTCDate(date.getUTCDate() - ((date.getUTCDay() + 6) % 7) + 3); // Thursday of this week
-  const isoYear = date.getUTCFullYear();
-  const firstThursday = new Date(Date.UTC(isoYear, 0, 4));
-  firstThursday.setUTCDate(firstThursday.getUTCDate() - ((firstThursday.getUTCDay() + 6) % 7) + 3);
-  const week = 1 + Math.round((date.getTime() - firstThursday.getTime()) / (7 * 24 * 3600 * 1000));
-  return `${isoYear}-W${String(week).padStart(2, '0')}`;
-}
 
 /**
  * Latest run key for a site: the lexically-greatest `YYYY-Www` folder under the
@@ -95,7 +79,7 @@ export async function handleBrandClaims(context) {
   if (model) {
     s3Key = `${CLAIMS_PREFIX}/${siteId}/${model}.json.gz`;
   } else if (date) {
-    s3Key = `${CLAIMS_PREFIX}/${siteId}/${toIsoWeek(new Date(date))}/data.json.gz`;
+    s3Key = `${CLAIMS_PREFIX}/${siteId}/${dateToIsoWeek(date.slice(0, 10))}/data.json.gz`;
   }
 
   log.info(`Getting brand claims for site ${siteId}, model: ${model || 'default'}${date ? `, date: ${date}` : ''}`);
--- a/src/controllers/llmo/brand-claims.js
+++ b/src/controllers/llmo/brand-claims.js
@@ -13,13 +13,59 @@
 import {
   badRequest, notFound,
 } from '@adobe/spacecat-shared-http-utils';
-import { HeadObjectCommand } from '@aws-sdk/client-s3';
+import { HeadObjectCommand, ListObjectsV2Command } from '@aws-sdk/client-s3';
 import { cachedOk } from '../../support/cached-response.js';
+
+const CLAIMS_PREFIX = 'brand_claims/llmo';
+const WEEK_RE = /^\d{4}-W\d{2}$/;
+
+/**
+ * ISO-week key segment (`YYYY-Www`) for a date, matching the delivery side
+ * (mystique) that writes one export per run under that folder.
+ *
+ * @param {Date} d - a valid Date
+ * @returns {string} e.g. '2026-W17'
+ */
+export function toIsoWeek(d) {
+  const date = new Date(Date.UTC(d.getUTCFullYear(), d.getUTCMonth(), d.getUTCDate()));
+  date.setUTCDate(date.getUTCDate() - ((date.getUTCDay() + 6) % 7) + 3); // Thursday of this week
+  const isoYear = date.getUTCFullYear();
+  const firstThursday = new Date(Date.UTC(isoYear, 0, 4));
+  firstThursday.setUTCDate(firstThursday.getUTCDate() - ((firstThursday.getUTCDay() + 6) % 7) + 3);
+  const week = 1 + Math.round((date.getTime() - firstThursday.getTime()) / (7 * 24 * 3600 * 1000));
+  return `${isoYear}-W${String(week).padStart(2, '0')}`;
+}
+
+/**
+ * Latest run key for a site: the lexically-greatest `YYYY-Www` folder under the
+ * site prefix. Returns null when no week folder exists yet (caller falls back to
+ * the legacy flat key).
+ */
+async function latestWeekKey(s3, bucketName, siteId) {
+  const prefix = `${CLAIMS_PREFIX}/${siteId}/`;
+  const res = await s3.s3Client.send(new ListObjectsV2Command({
+    Bucket: bucketName,
+    Prefix: prefix,
+    Delimiter: '/',
+  }));
+  const latest = (res.CommonPrefixes || [])
+    .map((p) => p.Prefix.slice(prefix.length).replace(/\/$/, ''))
+    .filter((seg) => WEEK_RE.test(seg))
+    .sort()
+    .pop();
+  return latest ? `${prefix}${latest}/data.json.gz` : null;
+}
 
 /**
  * Handles the brand claims retrieval by generating a presigned S3 URL.
  * Data files are .json.gz and can exceed Lambda's 6MB response limit,
  * so this endpoint returns a presigned URL rather than the data directly.
+ *
+ * Runs are stored per ISO week (`{siteId}/{YYYY-Www}/data.json.gz`). With no
+ * `date`, the latest week is served (falling back to the legacy flat
+ * `{siteId}/data.json.gz` for sites not yet migrated); with `date`, the run for
+ * that date's ISO week is served. A `model` selects a legacy flat
+ * `{model}.json.gz` file, unchanged.
  *
  * @param {object} context - The request context containing log, s3, env, and params
  * @returns {Promise<Response>} The brand claims presigned URL response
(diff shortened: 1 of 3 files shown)


src/controllers/llmo/brand-claims.js
```diff
@@ -38,15 +68,30 @@ export async function handleBrandClaims(context) {
...
   }
 
-  const s3Key = model
-    ? `brand_claims/llmo/${siteId}/${model}.json.gz`
-    : `brand_claims/llmo/${siteId}/data.json.gz`;
+  if (date !== undefined && Number.isNaN(new Date(date).getTime())) {
+    return badRequest(`Invalid date parameter: ${date}`);
+  }
+
+  // Model files are managed flat (not week-partitioned); resolved synchronously.
```
MysticatBot: issue (blocking): **Date validation accepts non-YYYY-MM-DD inputs.**

The check uses `new Date(date).getTime()` which happily parses strings like `"2026"`, `"2026-04"`, or `"2026-04-22T23:59:59Z"`. The OpenAPI spec declares `format: date` (YYYY-MM-DD), but the code does not enforce it. A partial input like `date=2026` passes validation, then `date.slice(0, 10)` yields `"2026"` and `dateToIsoWeek("2026")` produces an unexpected week key that will 404 with a confusing error.

**Fix:** Add a format gate before the Date parse:
```js
const DATE_RE = /^\d{4}-\d{2}-\d{2}$/;
if (date !== undefined && (!DATE_RE.test(date) || Number.isNaN(new Date(date).getTime()))) {
  return badRequest('Invalid date parameter: expected YYYY-MM-DD format');
}
```

src/controllers/llmo/brand-claims.js
```diff
@@ -13,21 +13,51 @@
...
+const WEEK_RE = /^\d{4}-W\d{2}$/;
+
+/**
+ * Latest run key for a site: the lexically-greatest `YYYY-Www` folder under the
+ * site prefix. Returns null when no week folder exists yet (caller falls back to
+ * the legacy flat key).
+ */
+async function latestWeekKey(s3, bucketName, siteId) {
+  const prefix = `${CLAIMS_PREFIX}/${siteId}/`;
+  const res = await s3.s3Client.send(new ListObjectsV2Command({
```
MysticatBot: question: `ListObjectsV2` returns at most 1000 `CommonPrefixes` per page, and this function does not paginate or check `IsTruncated`. At one folder per week that is ~19 years away, so this is safe for the foreseeable future. Worth adding a one-line comment documenting the assumption (and optionally a log warning if `res.IsTruncated`) so a future maintainer knows pagination was intentionally omitted.

MysticatBot (changes requested):
Hey @akshaymagapu,

**Verdict: Request changes** - two issues to resolve before merge.
**Complexity: HIGH** - medium diff; API surface change (OpenAPI contract modified).
_Changes: Adds ISO-week-partitioned S3 key resolution to the brand-claims endpoint, with a `date` query parameter and a fallback to the legacy flat key (3 files)._

## Must fix before merge
1. [Important] Permissive date validation accepts non-YYYY-MM-DD strings - `src/controllers/llmo/brand-claims.js:75` (details inline)
2. [Important] Missing integration tests for modified endpoint behavior per CLAUDE.md convention - `test/it/` (details inline)

<details>

- nit: `latestWeekKey` has no pagination and no comment explaining why it is safe to omit - `src/controllers/llmo/brand-claims.js:30`
- nit: Raw user input reflected in the 400 error message; truncate or use a generic message - `src/controllers/llmo/brand-claims.js:76`
- nit: OpenAPI says `date` is "Ignored when model is set" but validation still rejects invalid `date` when `model` is present - `src/controllers/llmo/brand-claims.js:75`
- suggestion: If `latestWeekKey` throws, the catch block logs `s3Key` which is still undefined at that point; consider assigning the fallback before the try or handling the list error separately - `src/controllers/llmo/brand-claims.js:92`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 5m 11s | <b>Cost:</b> $6.81 | <b>Commit:</b> <code>79c173444fa632657e41fadefea869610734db20</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

akshaymagapu:
Thanks @MysticatBot — all points addressed in c708cbf.

**Must-fix**
- **Date validation** — added a `^\d{4}-\d{2}-\d{2}$` gate before the `Date` parse. Partial inputs (`2026`, `2026-04`) and datetimes now return 400 with a generic message that no longer reflects raw input.
- **Integration tests** — deferring to a follow-up. This endpoint had no `test/it` coverage before this PR (pre-existing gap), and IT for a presign endpoint is a separate scaffold (seed MinIO week folders + shared factory/seed/wiring). All new branching logic is fully unit-covered.

**Nits**
- `latestWeekKey` — documented the no-pagination assumption (~19y under the 1000-prefix page bound) and added an `IsTruncated` warning. Also wrapped the listing in try/catch so a listing failure falls back to the legacy flat key instead of failing the request (a genuinely missing object still 404s at HEAD).
- Catch no longer logs an undefined key — `s3Key` now defaults to the flat key before the `try`; the latest-week list only upgrades it.
- OpenAPI — `date` documented as `YYYY-MM-DD`; reworded so `model` "takes precedence" rather than "ignored when model is set".

src/controllers/llmo/brand-claims.js
```diff
@@ -38,15 +81,34 @@ export async function handleBrandClaims(context) {
...
-    : `brand_claims/llmo/${siteId}/data.json.gz`;
+  if (date !== undefined && (!DATE_RE.test(date) || Number.isNaN(new Date(date).getTime()))) {
+    return badRequest('Invalid date parameter: expected YYYY-MM-DD format');
+  }
 
-  log.info(`Getting brand claims for site ${siteId}, model: ${model || 'default'}`);
+  // Model files are managed flat (not week-partitioned); `date` resolves directly
+  // to its week; otherwise default to the legacy flat key and upgrade it to the
+  // latest week (via a list) inside the try below.
+  let s3Key;
```
MysticatBot, line 99: issue (blocking): **`model` parameter is interpolated into the S3 key without validation.**

The `model` value goes directly into the key: `${CLAIMS_PREFIX}/${siteId}/${model}.json.gz`. While S3 does not resolve `..` as a filesystem would (the key is literal), an authenticated caller can craft values like `model=foo/bar/baz` to probe arbitrary key paths within the bucket via the HeadObject existence check (200 vs 404 leaks whether an object exists at that path).

The `date` parameter received a regex gate in this commit; `model` should get the same treatment.

**Fix:**
```js
const MODEL_RE = /^[\w.\-]+$/;
if (model !== undefined && !MODEL_RE.test(model)) {
  return badRequest('Invalid model parameter');
}
```

This closes the probing vector and aligns with the OpenAPI example (`gpt-4.1`) which is alphanumeric with dots and hyphens.

MysticatBot (changes requested):
Hey @akshaymagapu,

**Verdict: Request changes** - one remaining input validation gap to close.
**Complexity: HIGH** - large diff; API surface change (OpenAPI contract modified).
_Changes: Adds ISO-week-partitioned S3 key resolution to the brand-claims endpoint with a `date` query parameter, graceful legacy fallback, and comprehensive unit tests (3 files)._

## Must fix before merge
1. [Important] `model` query parameter has no input validation - enables S3 key probing - `src/controllers/llmo/brand-claims.js:91` (details inline)

<details>

- nit: `date` is validated (and rejected if malformed) even when `model` is also supplied and takes precedence for key selection; a valid request with a valid model gets a 400 if a stale `date` rides along - `src/controllers/llmo/brand-claims.js:83`
</details>

- Date validation now uses `^\d{4}-\d{2}-\d{2}$` regex gate before Date parse; generic error message no longer reflects raw input
- ListObjectsV2 pagination assumption documented with IsTruncated warning log
- `s3Key` defaults to the flat key before the try block so the catch no longer logs undefined
- OpenAPI reworded so `model` "takes precedence" rather than date being "ignored"

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 48s | <b>Cost:</b> $5.40 | <b>Commit:</b> <code>c708cbfc2e78e4e040dc4e4dcdf3300c5efb08b0</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

akshaymagapu:
Thanks @MysticatBot — both items addressed in 279f7ecb.

- **`model` validation (blocking)** — added a `^[\w.-]+$` gate; a `model` containing `/` now returns `400 Invalid model parameter` before any S3 call, closing the HeadObject key-probing vector. New test covers it.
- **`date` validated even when `model` wins (nit)** — date validation moved into the branch that actually uses it, so a request with a valid `model` (which takes precedence) no longer 400s on a stale/malformed `date` riding along. New test covers it.

src/controllers/llmo/brand-claims.js
```diff
@@ -38,15 +85,38 @@ export async function handleBrandClaims(context) {
...
+  if (model !== undefined && !MODEL_RE.test(model)) {
+    return badRequest('Invalid model parameter');
+  }
+
+  // Model files are managed flat (not week-partitioned) and take precedence;
+  // `date` resolves directly to its week (validated only here, where it is
+  // actually used); otherwise default to the legacy flat key and upgrade it to
+  // the latest week (via a list) inside the try below.
+  let s3Key;
+  if (model) {
```
MysticatBot, line 100: issue (blocking): **Date validation accepts semantically invalid dates that silently roll over.**

Dates like `2026-02-30` or `2026-04-31` pass both the `DATE_RE` regex and the `Number.isNaN(new Date(date).getTime())` check because JavaScript's `Date` constructor silently rolls invalid day-of-month values forward (Feb 30 becomes Mar 2). The user receives data for the wrong ISO week with no error.

Fix: verify the parsed date round-trips back to the original string:

```js
const parsed = new Date(`${date}T00:00:00Z`);
if (Number.isNaN(parsed.getTime()) || parsed.toISOString().slice(0, 10) !== date) {
  return badRequest('Invalid date parameter: expected YYYY-MM-DD format');
}
```

This replaces the current two-check pattern with a single round-trip assertion that catches both unparseable dates and rollover dates.

MysticatBot (changes requested):
Hey @akshaymagapu,

**Verdict: Request changes** - one remaining input validation gap to close.
**Complexity: HIGH** - large diff; API surface change (OpenAPI contract modified).
_Changes: Adds ISO-week-partitioned S3 key resolution to the brand-claims endpoint with a `date` query parameter, graceful legacy fallback, and comprehensive unit tests (3 files)._

## Must fix before merge
1. [Important] Date validation accepts semantically invalid dates (e.g. `2026-02-30`) that silently roll over to the wrong ISO week - `src/controllers/llmo/brand-claims.js:97` (details inline)

<details>

- nit: No length cap on `model` parameter; a ~900-char value passes MODEL_RE and produces a HeadObject call to a nonexistent key (hardening, not exploitable) - `src/controllers/llmo/brand-claims.js:85`
- nit: IsTruncated warning says resolution "may be incomplete" but the actual failure mode is stale-week selection (S3 returns earliest keys first, so `.sort().pop()` on a truncated page picks an old week, not the latest) - `src/controllers/llmo/brand-claims.js:37`
- suggestion: Validate `dateToIsoWeek()` output against WEEK_RE before key interpolation as defense-in-depth - `src/controllers/llmo/brand-claims.js:100`
</details>

- Model parameter now gated by `MODEL_RE = /^[\w.-]+$/`; path-separator injection returns 400
- Date validation scoped to the branch that uses it; model+malformed-date no longer 400s
- Generic error messages no longer reflect raw user input
- ListObjectsV2 pagination assumption documented with IsTruncated warning log


---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 11m 53s | <b>Cost:</b> $4.54 | <b>Commit:</b> <code>279f7ecb1bd5a110b178404a2bef8897f934c742</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/controllers/llmo/brand-claims.js
akshaymagapu: Fixed in 1314353 — replaced the regex+NaN pair with a single UTC round-trip assertion: `new Date(`${date}T00:00:00Z`)` and reject unless `toISOString().slice(0,10) === date`. Catches both unparseable and rollover dates (2026-02-30 → 400). Added a test for the rollover case.

MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - all blocking issues from prior reviews resolved; remaining items are non-blocking suggestions.
**Complexity: HIGH** - medium diff; API surface change (OpenAPI contract modified).
_Changes: Adds ISO-week-partitioned S3 key resolution to the brand-claims endpoint with a `date` query parameter, UTC round-trip date validation, and graceful legacy fallback (3 files)._

<details>

- suggestion: Collapse the dual-condition key assignment into a single else branch (s3Key is assigned the legacy flat key then potentially overwritten 12 lines later inside try; moving latestWeekKey into the else eliminates the double-assignment and makes the fallback explicit in one place) - `src/controllers/llmo/brand-claims.js:95`
- suggestion: Validate dateToIsoWeek() output against WEEK_RE before key interpolation as defense-in-depth - `src/controllers/llmo/brand-claims.js:100`
- suggestion: Add a test for the date path where HeadObject returns 404 (verifies the correct week key appears in the log and no presigned URL is generated for a requested week that does not exist yet) - `test/controllers/llmo/brand-claims.test.js`
- suggestion: Pin empty-string edge cases (model: empty and date: empty) with explicit 400 assertions to guard against future refactors of the validation order - `test/controllers/llmo/brand-claims.test.js`
</details>

- Date validation now uses a single UTC round-trip assertion; catches both unparseable dates and calendar-invalid rollover dates (e.g. 2026-02-30)
- Model parameter gated by MODEL_RE; path-separator injection returns 400
- Date validation scoped to the branch that uses it; model+malformed-date no longer 400s
- ListObjectsV2 pagination assumption documented with IsTruncated warning
- s3Key defaults before try block; catch no longer logs undefined
- OpenAPI reworded so model takes precedence


---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 53s | <b>Cost:</b> $4.17 | <b>Commit:</b> <code>1314353c0022f9a63957663566318e1630a97d8b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/controllers/llmo/brand-claims.js
```diff
@@ -13,21 +13,67 @@
...
+    // One folder per ISO week keeps this well under the 1000-prefix page limit
+    // (~19 years), so pagination is intentionally omitted; warn if that changes.
+    if (res.IsTruncated) {
+      log.warn(`Brand claims week listing truncated for site ${siteId}; latest-week resolution may be incomplete`);
+    }
+    const latest = (res.CommonPrefixes || [])
+      .map((p) => p.Prefix.slice(prefix.length).replace(/\/$/, ''))
+      .filter((seg) => WEEK_RE.test(seg))
+      .sort()
+      .pop();
```
davidaurelio: Unproblematic here, but this should really be a find-max algorithm rather than a full sort. With two years of data, this will do ~6.7x the necessary work.

davidaurelio (approved):
go for it

src/controllers/llmo/brand-claims.js
akshaymagapu: yeah made a fix now

PR #3010, opened 2026-08-06
feat(llmo-6818): Add option all to platform param

amitruluimadalina:

## Related Issues

No issue created.

## Problem

The Serenity URL Inspector `domain-urls` endpoint defaults to `search-gpt` when no `platform` is provided. However, the UI also needs to request URLs across all platforms by sending `platform=all`.

Before this change, `platform=all` was resolved through the default model resolver and effectively behaved like `search-gpt`.

## Solution

Update the domain URLs Semrush Elements payload builder so that:

- no `platform` / `model` still defaults to `search-gpt`
- specific platform values, such as `google-ai-mode`, continue to send a `CBF_model` filter
- `platform=all` omits the `CBF_model` filter, allowing Semrush to return URLs across all platforms
- existing sorting by citations and pagination behavior remain unchanged

## Testing

```bash
npx mocha domain-urls.test.js
npx eslint domain-urls.js domain-urls.test.js

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Andrei Paraschiv", "Dominique Jäggi", "TB ADBE"]
```


src/support/elements/definitions/domain-urls.js
```diff
@@ -10,7 +10,7 @@
  * governing permissions and limitations under the License.
  */
 
-import { resolveElementModel } from '../constants.js';
+import { isAllPlatforms, resolveElementModel } from '../constants.js';
```
MysticatBot, line 13: issue (blocking): The CLAUDE.md rule states: "If the PR is changing the API implementation or an entity exposed through the API: make sure you update the API specification and the examples to reflect the changes."

This PR adds `platform=all` as a sentinel value with distinct behavior (omits the `CBF_model` filter entirely). The OpenAPI spec for this endpoint should document `all` as a valid `platform` value and describe its semantics. Without this, consumers have no way to discover the feature from the API contract.

test/support/elements/definitions/domain-urls.test.js
```diff
@@ -0,0 +1,43 @@
...
+      expect(modelFilter(payload)).to.be.undefined;
+    });
+
+    it('keeps the existing platform translation for non-all values', () => {
+      const payload = buildDomainUrlsPayload({ platform: 'openai' });
+
+      expect(modelFilter(payload).filters[0].val).to.equal('chatgpt-paid');
+    });
+  });
+});
```
MysticatBot, line 183: issue (blocking): The JSDoc for `buildDomainUrlsPayload` documents that `model` wins over `platform` ("Legacy alias for `model`; `model` wins"). No test exercises this precedence. If a future refactor changes `model || platform` to `platform || model`, the existing tests would still pass.

Suggested addition:
```js
it('uses model over platform when both are provided', () => {
  const payload = buildDomainUrlsPayload({ model: 'all', platform: 'openai' });
  expect(modelFilter(payload)).to.be.undefined;
});
```
This also documents that `model=all` omits the filter (same semantics as `platform=all`).

MysticatBot (changes requested):
Hey @amitruluimadalina,

**Verdict: Request changes** - two documentation/spec requirements from the project's CLAUDE.md are unmet.
**Complexity: LOW** - small diff, single-service.
_Changes: Adds `platform=all` support to the domain-urls Semrush Elements payload builder (omits the CBF_model filter) and extracts SerenityTransportError into a leaf module to break a circular import (8 files)._

## Must fix before merge
1. [Important] Missing OpenAPI spec update for `platform=all` sentinel value - `docs/openapi/` (details inline)
2. [Important] Missing test for `model`-over-`platform` precedence contract - `test/support/elements/definitions/domain-urls.test.js` (details inline)

<details>

- nit: `/* c8 ignore start -- LLMO-6160 POC endpoint; unit tests intentionally deferred */` is now stale since this PR adds unit tests for `buildDomainUrlsPayload` - narrow the pragma or remove it - `src/support/elements/definitions/domain-urls.js:15`
- suggestion: The `platform=all` test asserts the model filter is absent but does not assert the date filters are still present - a regression that drops date filters would pass silently - `test/support/elements/definitions/domain-urls.test.js:31`
- suggestion: Add `@deprecated` JSDoc on the re-export in rest-transport.js so future consumers import from the leaf module directly - `src/support/serenity/rest-transport.js:23`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 6m 41s | <b>Cost:</b> $5.10 | <b>Commit:</b> <code>e02ebe2a4f311ef861bc9760b93552bb861e5a26</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

test/support/elements/definitions/domain-urls.test.js
```diff
@@ -0,0 +1,43 @@
...
+
+      expect(modelFilter(payload)).to.deep.equal({
+        op: 'or', filters: [{ op: 'eq', val: DEFAULT_ELEMENT_MODEL, col: 'CBF_model' }],
+      });
+    });
+
+    it('omits the CBF_model filter for platform=all', () => {
+      const payload = buildDomainUrlsPayload({ platform: 'all' });
+
+      expect(modelFilter(payload)).to.be.undefined;
```
tb-adbe, line 45: The `platform=all` test only asserts that the `CBF_model` filter is absent; it doesn't assert that the `CBF_date__start` / `CBF_date__end` filters are still present. This PR restructured `advancedFilters` from a fixed literal into a mutable array that conditionally `unshift`s the model filter, so a future regression that also dropped the date filters on the `all` path would pass this test green - and the practical effect would be an unbounded-date-range query instead of the requested window. Add a positive assertion that both date filters are present in the `platform=all` case (or assert the full expected `advanced.filters` array).

src/support/elements/definitions/domain-urls.js
```diff
@@ -10,7 +10,7 @@
...
 
 /* c8 ignore start -- LLMO-6160 POC endpoint; unit tests intentionally deferred */
```
tb-adbe: The `/* c8 ignore start -- LLMO-6160 POC endpoint; unit tests intentionally deferred */` pragma wraps `buildDomainUrlsPayload`, but this PR adds real unit tests for exactly that function. The comment is now factually wrong, and more importantly the coverage gate is silently not protecting these lines - the modified branches are marked non-coverable, so a future PR could delete a branch of this resolver and coverage tooling wouldn't flag it. Narrow or remove the `c8 ignore` around this function so the new tests actually count toward the gate.

tb-adbe (commented):
Hey @amitruluimadalina,

## Strengths

The core fix in `src/support/elements/definitions/domain-urls.js:41-49` is correct and minimal: gating the `CBF_model` filter behind `!isAllPlatforms(requestedModel)` and using `unshift` to preserve the original filter ordering. The no-platform default (falls through to `search-gpt`) is preserved and locked in by a test. This also mirrors the identical, already-established `isAllPlatforms` pattern in the sibling `url-prompts.js` builder, so it's consistent with how this module already handles the sentinel rather than inventing a new approach.

The bundled `SerenityTransportError` extraction into its own leaf module (`src/support/serenity/serenity-transport-error.js`) is a clean, backward-compatible fix for a real problem: it breaks the `errors.js` <-> `rest-transport.js` import cycle that an old comment explicitly called out. All four importers were migrated, the re-export in `rest-transport.js` keeps external callers working, and `instanceof`-based error classification (`isSemrushTransportError`) still resolves correctly since class identity is preserved.

From a security standpoint, the `platform=all` change actually reduces exposure rather than increasing it: it omits the filter entirely instead of passing any user-controlled value through, and the existing allowlist in `resolveElementModel` (against the frozen `ELEMENT_MODELS` set) is untouched. The outbound payload is a structured JSON body, not a concatenated query string, so there's no filter-injection surface here.

## Issues

### Important (Should Fix)

1. **`test/support/elements/definitions/domain-urls.test.js:31-35`** - The `platform=all` test only asserts that the `CBF_model` filter is absent; it doesn't assert that the `CBF_date__start` / `CBF_date__end` filters are still present. This PR restructured `advancedFilters` from a fixed literal into a mutable array that conditionally `unshift`s the model filter, so a future regression that also dropped the date filters on the `all` path would pass this test green - and the practical effect would be an unbounded-date-range query instead of the requested window. Add a positive assertion that both date filters are present in the `platform=all` case (or assert the full expected `advanced.filters` array).

2. **`src/support/elements/definitions/domain-urls.js:15`** - The `/* c8 ignore start -- LLMO-6160 POC endpoint; unit tests intentionally deferred */` pragma wraps `buildDomainUrlsPayload`, but this PR adds real unit tests for exactly that function. The comment is now factually wrong, and more importantly the coverage gate is silently not protecting these lines - the modified branches are marked non-coverable, so a future PR could delete a branch of this resolver and coverage tooling wouldn't flag it. Narrow or remove the `c8 ignore` around this function so the new tests actually count toward the gate.

### Minor (Nice to Have)

1. **`src/support/serenity/rest-transport.js:125-126`** - The comment on `redactUpstreamMessage` used to explain *why* the `errors.js` <-> `rest-transport.js` cycle existed and noted "extracting the shared helper to a leaf module is a follow-up." This PR created that leaf module but trimmed the comment down to just "would create a cycle. Keep in sync" - which now asserts a constraint your own change removed (`errors.js` no longer imports from `rest-transport.js`). Either collapse the `unwrapTransportCause` duplication now that the leaf module exists, or restore a note pointing future maintainers at it as the follow-up destination.

2. **`src/support/serenity/rest-transport.js:23-25`** - Now that every internal caller imports `SerenityTransportError` directly from the new leaf module, the re-export here only serves external consumers (if any exist) or is dead surface reintroducing "which path do I import from" ambiguity. Worth a one-line comment stating whether this is an intentional compatibility shim.

3. **`src/support/elements/definitions/domain-urls.js:41`** / test file - The JSDoc states `model` wins over `platform` when both are given, but no test exercises that precedence (e.g. `{ model: 'all', platform: 'openai' }`). Low risk since the logic (`model || platform`) is trivially correct and no current caller supplies both, but a one-line test would guard against a future `platform || model` typo.

## Recommendations

- Add coverage for `isAllPlatforms`'s case/whitespace normalization (`'ALL'`, `' all '`) and the unknown-platform fallback (e.g. `platform='bogus'` resolving to `DEFAULT_ELEMENT_MODEL` rather than omitting the filter) - both are one-line additions and the boundaries most likely to drift.
- As the `isAllPlatforms` sentinel pattern spreads to more builders in this module (currently `domain-urls.js` and `url-prompts.js`), consider a single shared `resolveRequestedModel({ model, platform })` helper that returns the effective value plus an `isAll` flag, rather than repeating the guard in each definition file.

An earlier review on this PR flagged a missing OpenAPI spec update for `platform=all` as a blocking issue. After checking, the `listDomainUrls` endpoint this PR touches has no OpenAPI documentation anywhere in `docs/openapi/` - not even for its base functionality - consistent with the file's own `c8 ignore -- LLMO-6160 POC endpoint` marker. Asking this bug-fix PR to author the endpoint's first-ever spec is scope well beyond its stated intent; the real gap (the endpoint being undocumented at all) is worth a tracking ticket against LLMO-6160, not a blocker here.

Separately, only 2 of the ~13 sibling `CBF_model`-filter builders in this module currently honor `platform=all`. If "all platforms" is meant to be a global concept, the divergence will show up as inconsistent behavior across elements - a product-scope follow-up, not a defect in this PR.

## Assessment

Ready to merge? With fixes.
Reasoning: The fix itself is correct, minimal, secure, and consistent with an established in-repo pattern. The two Important findings are both about test/coverage rigor rather than the runtime logic being wrong - they're small, test-and-comment-only changes and shouldn't take long to land.

## Next Steps

1. Add the missing date-filter assertion to the `platform=all` test, and narrow the `c8 ignore` so coverage actually gates this function.
2. The three Minor items (stale cycle comment, re-export intent, precedence test) are optional cleanup.

MysticatBot (approved):
Hey @amitruluimadalina,

**Verdict: Approve** - all prior blocking findings are addressed; no new issues found.
**Complexity: LOW** - small diff, single-service.
_Changes: Adds `platform=all` support to the domain-urls Semrush Elements payload builder, documents the new sentinel in OpenAPI, and extracts SerenityTransportError into a leaf module to break a circular import (9 files)._

- OpenAPI spec now documents `platform=all` with description and example
- Model-over-platform precedence test added
- Date-filter presence assertion added to the `platform=all` test case
- `c8 ignore` pragma repositioned to cover only untested functions below `buildDomainUrlsPayload`

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 23s | <b>Cost:</b> $4.39 | <b>Commit:</b> <code>4d600c513f73b141d06aa38e3ce59b842f49c09d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

tb-adbe (commented):
Hey @amitruluimadalina,

Re-reviewing the fix commit (`4d600c51`) against my earlier findings.

## Strengths

All five previously-flagged items are correctly and completely fixed:

- `test/support/elements/definitions/domain-urls.test.js:21-49` - the new `dateFilters()` helper plus a `deep.equal` assertion now pins that both `CBF_date__start`/`CBF_date__end` survive on the `platform=all` path. Traced it against the builder: this closes the original gap, a regression dropping either date filter would now fail the test.
- `src/support/elements/definitions/domain-urls.js` - the `c8 ignore start` pragma was correctly repositioned to after `buildDomainUrlsPayload`'s closing brace. Confirmed via the full file that `buildDomainUrlsPayload` is now under the coverage gate, while the range still excluded (`hostOf`, `hostMatches`, `parsePagination`, `transformDomainUrlsResponse`, down to the pre-existing `c8 ignore stop` at EOF) is exactly the set that was already untested before this PR - no coverage laundering, no newly-hidden regression risk.
- `src/support/serenity/rest-transport.js:125-129` - the stale "would create a cycle" comment is now an honest, checkable explanation (the `@ts-check` strict-closure cost of importing `unwrapTransportCause`, not a real cycle).
- `src/support/serenity/rest-transport.js:23-25` - the re-export now has a clear "compatibility shim for external consumers" comment.
- `test/support/elements/definitions/domain-urls.test.js:52-56` - the new `model`-over-`platform` precedence test genuinely discriminates: `resolveElementModel('openai')` alone resolves to a defined `chatgpt-paid` filter (per the sibling test), so `{ model: 'all', platform: 'openai' }` returning `undefined` actually proves `model` won, not a coincidence.

## Issues

### Important (Should Fix)

1. **`docs/openapi/llmo-api.yaml:4762-4770`** - This fix commit also added a `platform` parameter description to the `url-inspector-domain-urls` operation, responding to an earlier request to document `platform=all`. Two problems with it:
   - **Wrong endpoint.** That operation is bound (via `docs/openapi/api.yaml:669-670`) to path `/org/{spaceCatId}/brands/{brandId}/brand-presence/url-inspector/domain-urls`, which `src/routes/index.js:711-712` routes to `llmoMysticatController.getUrlInspectorDomainUrls` - a controller this PR never touches. The endpoint this PR actually changes is a different route entirely: `GET /v2/orgs/:spaceCatId/brands/:brandId/serenity/brand-presence/url-inspector/domain-urls` -> `elementsController.listDomainUrls` (`src/routes/index.js:274`), the caller of the modified `buildDomainUrlsPayload`. There is no OpenAPI operation for that v2/serenity path anywhere in `docs/openapi/`, before or after this commit - the endpoint that actually changed remains undocumented.
   - **Factually wrong for the endpoint it landed on.** The new text says "Defaults to `search-gpt` when omitted" - that's the Elements v2 default. `llmoMysticatController`'s own `resolveUrlInspectorPlatform` (`src/controllers/llmo/llmo-url-inspector.js:80-92`) does the opposite: "Returns null when absent (no default model)", and the file's own header comment is explicit that this "differs from brand-presence endpoints which default to `chatgpt-free`". So the spec now asserts a default that contradicts the operation's actual implementation.

   Why it matters: this isn't just a scope gap, it's a documentation regression this fix commit introduced - and the approval that just landed on this PR explicitly cited "OpenAPI spec now documents `platform=all`" as a resolved item, which is incorrect for the endpoint whose behavior this PR changed.

   How to fix: revert the `llmo-api.yaml` edit (it makes the mysticat operation's spec less accurate, not more), and if documenting the v2 serenity endpoint's `platform=all` semantics is wanted, add it to the actual `listDomainUrls` operation - creating that operation's spec entry if none exists yet.

## Assessment

Ready to merge? With fixes.
Reasoning: the four code-level fixes are solid and fully close the gaps they claim to. The one remaining item is a documentation regression from this same fix commit - small to correct, but real, since it currently misinforms consumers about a sibling endpoint's behavior and the basis for the existing approval.

## Next Steps

1. Revert or redirect the `docs/openapi/llmo-api.yaml` change so it doesn't misdocument `llmoMysticatController`'s endpoint.
2. If desired, document the `platform=all` behavior on the actual endpoint this PR changed (`listDomainUrls`), which currently has no OpenAPI entry at all - can be a fast follow rather than blocking this PR, given it was already undocumented before this change.

docs/openapi/llmo-api.yaml
```diff
@@ -4762,7 +4762,12 @@ url-inspector-domain-urls:
       - name: platform
         in: query
         required: false
-        schema: { type: string }
+        description: |
+          AI platform/model to filter by. Defaults to `search-gpt` when omitted.
```
tb-adbe: Two problems with this addition:

**Wrong endpoint.** This operation (`url-inspector-domain-urls`) is bound, via `docs/openapi/api.yaml:669-670`, to path `/org/{spaceCatId}/brands/{brandId}/brand-presence/url-inspector/domain-urls`, which `src/routes/index.js:711-712` routes to `llmoMysticatController.getUrlInspectorDomainUrls` - a controller this PR never touches. The endpoint this PR actually changes is `GET /v2/orgs/:spaceCatId/brands/:brandId/serenity/brand-presence/url-inspector/domain-urls` -> `elementsController.listDomainUrls` (`src/routes/index.js:274`), which is what calls the modified `buildDomainUrlsPayload`. There is no OpenAPI operation for that v2/serenity path anywhere in `docs/openapi/` - the endpoint that actually changed remains undocumented.

**Factually wrong for the endpoint it landed on.** "Defaults to `search-gpt` when omitted" is the Elements v2 default. `llmoMysticatController`'s own `resolveUrlInspectorPlatform` (`src/controllers/llmo/llmo-url-inspector.js:80-92`) does the opposite - "Returns null when absent (no default model)" - and that file's header comment explicitly says this "differs from brand-presence endpoints which default to `chatgpt-free`". The spec now asserts a default that contradicts this operation's actual implementation.

Suggest reverting this edit (it makes the mysticat operation's spec less accurate) and, if wanted, documenting `platform=all` on the actual `listDomainUrls` operation instead.

AndreiAlexandruParaschiv (approved):

tb-adbe (approved):
Hey @amitruluimadalina,

Re-checked at the current head. The two commits since my last review are just merges of `origin/main` into this branch - no new changes to any file this PR actually modifies (`domain-urls.js`, its test file, `rest-transport.js`, `elements.js`, `errors.js`, `resource-manager.js`, `serenity-transport-error.js` are all identical to what I already verified).

The one outstanding item from my last review - the OpenAPI edit that landed on the wrong sibling endpoint (`llmoMysticatController`'s `url-inspector-domain-urls`) with a description that contradicted that endpoint's actual behavior - is now resolved: that file is no longer part of this PR's diff at all. The `platform` parameter on that operation is back to its original, undescribed `schema: { type: string }`, so the incorrect text is gone and nothing new replaced it.

All five original findings remain correctly fixed, CI is green, and there's nothing outstanding from my side.

Approving.

PR #3015, opened 2026-08-07
feat(ai-seo): v1 for top-brands and brand competitors

dmitriigulkin-semrush:
## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```


calvarezg:
## Multi-Persona Review — Changes recommended

Thin, well-scoped PR: two Semrush v2 gRPC passthrough endpoints (`competitors`, `top-brands`) plus OpenAPI, route/capability wiring, and wiring tests. Structure matches the existing v1/brand handlers. One Major to confirm before merge; the rest are minor / consistent-with-siblings.

### Major

**Domain normalization differs between the two new handlers** (confirmed)
- `top-brands.js:L390` sends `domain.replace(/^www\./i,'').toLowerCase()` as `brand_domain`.
- `competitors.js:L310` sends the raw trimmed domain as `target: { domain, name: domain }` — no `www.`-strip, no lowercase.
- So `www.Example.COM` reaches the two endpoints as two different keys. If Semrush canonicalizes the brand domain (lowercase / `www.`-stripped, which top-brands assumes), competitors can return wrong/empty results for mixed-case or `www.`-prefixed input.
- Note: `grpc-utils.js:183` already has an unused `brandTarget(domain)` = `trim().toLowerCase()` → `{domain,name}` — the exact shape competitors hand-builds, lowercased. Caveat: the other `target`-shaped siblings (`competitors-stats`, `stats-by-llm`, `stats-by-country`) also pass the raw domain, so top-brands is the outlier that *added* normalization.
- **Ask:** is the asymmetry intentional? If canonicalization is right, route competitors (ideally the whole target family) through `brandTarget()`; if not, drop it from top-brands. Either way make the pair consistent + add a comment.

### Minor
- **Handler logic is `c8`-ignored and untested** (`competitors.js:L295-L338`, `top-brands.js:L371-L418`). A proto field-name typo or broken 400 path would ship undetected — codecov shows 100% only because the ignore suppresses these lines. Matches every sibling, so team-wide, not a blocker. A thin unit test per handler (400 paths + correct mapped request) would close it.
- **`count` / `limit` unvalidated** — OpenAPI says `count` `minimum: 1` but `count=0` passes through; `top-brands` ignores the repo's `parseLimitOffset` helper. Validate as int >= 1 with a cap before `fromJson`.
- **`limit='10'` default is undocumented** and asymmetric with `count` (which defers to Semrush's default of 4). Document it in the OpenAPI `limitParam` or omit `limit` when unset.
- **gRPC / protobuf error text returned verbatim** (`competitors.js:L317-L320` + shared error mapper) leaks internal field / upstream detail. Low impact behind auth, matches siblings. Prefer a generic client message + server-side log.

### Question
- On an unmapped gRPC error the handler rethrows with no request context. `wrapHandler` does `log.error` the error, but without domain/engine/limit — is that enough for prod diagnosis?

### Working well
- Explicit `if (!domain) return 400` guard that siblings lack.
- `srFilters` correctly injected by `wrapHandler`; OpenAPI (incl. top-brands pagination note) accurate.
- All wiring (route / capability / required-capability / ROUTE_MAP / tests) consistent with siblings.
- No SSRF/injection — params land in typed protobuf fields over a fixed gRPC client.

**Summary:** 0 critical / 1 major / 4 minor / 1 question. Recommendation: **Changes recommended** — confirm the normalization intent before merge.

dmitriigulkin-semrush:
hey @calvarezg 

> Domain normalization differs between the two new handlers (confirmed)
> 
> top-brands.js:L390 sends domain.replace(/^www\./i,'').toLowerCase() as brand_domain.
> competitors.js:L310 sends the raw trimmed domain as target: { domain, name: domain } — no www.-strip, no lowercase.
> So www.Example.COM reaches the two endpoints as two different keys. If Semrush canonicalizes the brand domain (lowercase / www.-stripped, which top-brands assumes), competitors can return wrong/empty results for mixed-case or www.-prefixed input.
> Note: grpc-utils.js:183 already has an unused brandTarget(domain) = trim().toLowerCase() → {domain,name} — the exact shape competitors hand-builds, lowercased. Caveat: the other target-shaped siblings (competitors-stats, stats-by-llm, stats-by-country) also pass the raw domain, so top-brands is the outlier that added normalization.
> Ask: is the asymmetry intentional? If canonicalization is right, route competitors (ideally the whole target family) through brandTarget(); if not, drop it from top-brands. Either way make the pair consistent + add a comment.

i've removed .www clean-up and domain stripping, downstream backend does it for us so we shouldn't worry about that on our end, it's a consistent pattern through v1 endpoint

calvarezg (approved):
Approving — multi-persona review found no blocking issues. The domain-normalization concern is resolved (both handlers now pass the raw domain consistently, downstream canonicalizes per author). Remaining items are minor and consistent with the sibling handlers: c8-ignored/untested handler logic, count/limit not validated against the OpenAPI minimum, undocumented limit default, and verbatim gRPC error text. One small follow-up worth a look: dropping .trim() lets a whitespace-only domain bypass the 400 guard.

PR #3016, opened 2026-08-07
feat(llmo-6844): Make hostname param optional

amitruluimadalina:

## Related Issues

No related issue.

This PR updates the Serenity `domain-urls` endpoint so `hostname` / `domain` is optional. When it is omitted, the endpoint returns URLs across all source hosts for the requested page, preserving the existing citation-descending ordering and pagination behavior. When `hostname` is provided, the existing host-or-subdomain filtering behavior is unchanged.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Andrei Paraschiv", "Dominique Jäggi", "TB ADBE"]
```

### DIFF
--- a/src/controllers/elements.js
+++ b/src/controllers/elements.js
@@ -1236,10 +1236,10 @@ export default function ElementsController(context, log, env) {
   /**
    * GET /v2/orgs/:spaceCatId/brands/:brandId/serenity/brand-presence
    *     /url-inspector/domain-urls
-   * Phase 2 of the Cited Third-Party tree: expand a cited domain → its URLs.
-   * Same Semrush element as owned-urls (Stats-per-URL 9af5ed83) minus the trend
-   * element and the Postgres traffic hybrid, filtered to a single domain
-   * (required `hostname`) client-side instead of `domain_type='Owned'`.
+    * Phase 2 of the Cited Third-Party tree: expand a cited domain → its URLs.
+    * Same Semrush element as owned-urls (Stats-per-URL 9af5ed83) minus the trend
+    * element and the Postgres traffic hybrid, optionally filtered to a single
+    * domain (`hostname`) client-side instead of `domain_type='Owned'`.
    */
   /* c8 ignore start -- LLMO-6160 POC endpoint; unit tests intentionally deferred */
   const listDomainUrls = async (ctx) => {
@@ -1273,13 +1273,6 @@ export default function ElementsController(context, log, env) {
         return badRequest(`Date range must not exceed ${MAX_RANGE_DAYS} days`);
       }
 
-      // hostname (aka domain) is the domain to drill into — required (the UI only
-      // calls this after a domain row is expanded).
-      const hostname = query.hostname || query.domain;
-      if (!hasText(hostname)) {
-        return badRequest('hostname is required for domain URL drilldown');
-      }
-
       const service = await buildService(ctx);
 
       const { BrandSemrushProject } = ctx?.dataAccess ?? {};
@@ -1301,9 +1294,11 @@ export default function ElementsController(context, log, env) {
       } else {
         projects = await service.getOwnedUrlProjects(workspaceId, { brandSemrushProjects });
       }
+      const hostname = query.hostname || query.domain;
 
-      // The transform host-filters, sorts by citations desc, and slices client-side
-      // (Semrush has no server-side pagination); totalCount is the full post-filter count.
+      // The transform optionally host-filters, sorts by citations desc, and slices
+      // client-side (Semrush has no server-side pagination); totalCount is the full
+      // post-filter count.
       const result = await service.getDomainUrls(workspaceId, {
         projects,
         hostname,
--- a/src/controllers/elements.js
+++ b/src/controllers/elements.js
@@ -1236,12 +1236,11 @@ export default function ElementsController(context, log, env) {
   /**
    * GET /v2/orgs/:spaceCatId/brands/:brandId/serenity/brand-presence
    *     /url-inspector/domain-urls
-    * Phase 2 of the Cited Third-Party tree: expand a cited domain → its URLs.
-    * Same Semrush element as owned-urls (Stats-per-URL 9af5ed83) minus the trend
-    * element and the Postgres traffic hybrid, optionally filtered to a single
-    * domain (`hostname`) client-side instead of `domain_type='Owned'`.
+   * Phase 2 of the Cited Third-Party tree: expand a cited domain → its URLs.
+   * Same Semrush element as owned-urls (Stats-per-URL 9af5ed83) minus the trend
+   * element and the Postgres traffic hybrid, optionally filtered to a single
+   * domain (`hostname`) client-side instead of `domain_type='Owned'`.
    */
-  /* c8 ignore start -- LLMO-6160 POC endpoint; unit tests intentionally deferred */
   const listDomainUrls = async (ctx) => {
     try {
       const auth = await authorizeOrg(ctx);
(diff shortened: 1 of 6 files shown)


MysticatBot (approved):
Hey @amitruluimadalina,

**Verdict: Approve** - clean, backward-compatible change with good test coverage.
**Complexity: HIGH** - medium diff; API surface change (risk flag).
_Changes: Makes the `hostname` query parameter optional on the `domain-urls` endpoint, returning all source hosts when omitted while preserving existing filtering behavior when provided (5 files)._

<details>

- nit: JSDoc comment indentation uses 4-space where surrounding lines use 3-space - `src/controllers/elements.js:1239`
- suggestion: Normalize whitespace-only hostname in the controller (`(query.hostname || query.domain || '').trim() || undefined`) so intent is explicit at the API boundary rather than relying on downstream `.trim()` - `src/controllers/elements.js:1297`
- suggestion: Add a test for the `channel` filter combined with hostname-omitted mode in `domain-urls.test.js` to cover the primary new usage pattern

</details>

**Out of scope, worth tracking:** When `hostname` is omitted the in-memory aggregation in `transformDomainUrlsResponse` processes all rows from all projects before pagination. This is pre-existing architecture (Semrush has no server-side pagination) now reachable via the API for the first time. Existing guards (ReBAC auth, date-range cap, Semrush 50k-row cap, pageSize 1000) mitigate the risk, but a follow-up adding a row-count circuit-breaker or observability hook for large unfiltered result sets would be prudent.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 45s | <b>Cost:</b> $5.20 | <b>Commit:</b> <code>ccff26a1a2a23ac15a35187cb36ec46bd1649d20</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/elements/definitions/domain-urls.js
```diff
@@ -114,20 +114,22 @@ function parsePagination({ page, pageSize } = {}) {
...
  *
  * @param {Array<{region?: string, stats: object}>} projectResults
- * @param {object} params - { hostname (required), channel, page, pageSize }.
+ * @param {object} params - { hostname, channel, page, pageSize }.
  * @returns {{ urls: Array<object>, totalCount: number }}
  */
 export function transformDomainUrlsResponse(projectResults = [], params = {}) {
   const { page, pageSize } = parsePagination(params);
-  const hostname = String(params.hostname ?? '').replace(/^www\./, '').toLowerCase();
+  const hostname = String(params.hostname ?? '').trim().replace(/^www\./, '').toLowerCase();
```
tb-adbe, line 131: Important: this `.trim()` call has no test exercising the case it exists to handle. A whitespace-only `hostname` (e.g. `?hostname=%20%20`) is truthy at the controller (`query.hostname || query.domain` passes it through), so without `.trim()` it would reach here as a non-empty value, set `filterByHostname = true`, and silently match zero rows for a caller who effectively asked for no filter. The two new transform tests cover only "hostname key absent" and "one valid hostname" - neither exercises the whitespace-collapses-to-no-filter path this line was added for. Add one assertion with `hostname: '   '` asserting all hosts are returned, to lock in the behavior this line depends on.

src/controllers/elements.js
```diff
@@ -1236,10 +1236,10 @@ export default function ElementsController(context, log, env) {
   /**
    * GET /v2/orgs/:spaceCatId/brands/:brandId/serenity/brand-presence
    *     /url-inspector/domain-urls
-   * Phase 2 of the Cited Third-Party tree: expand a cited domain → its URLs.
-   * Same Semrush element as owned-urls (Stats-per-URL 9af5ed83) minus the trend
-   * element and the Postgres traffic hybrid, filtered to a single domain
-   * (required `hostname`) client-side instead of `domain_type='Owned'`.
+    * Phase 2 of the Cited Third-Party tree: expand a cited domain → its URLs.
```
tb-adbe: Minor: these four replaced/added JSDoc lines are indented one space further than the untouched opening `/**` and the lines directly above them, so the comment block is now visibly ragged. Cosmetic, ESLint doesn't catch it, but worth a one-character fix while the block is already being edited.

tb-adbe (commented):
Hey @amitruluimadalina,

## Strengths

- The filter gating in `transformDomainUrlsResponse` is done via a single `filterByHostname` guard on the existing host-match branch rather than a forked code path, so the two behaviors (filtered / unfiltered) can't drift apart.
- The added `.trim()` normalization is a genuine hardening: a whitespace-only hostname now collapses to "no filter" instead of silently matching zero rows.
- Authorization ordering is untouched: `authorizeOrg` still runs first and gates the endpoint on brand access. `hostname` never participates in scoping, query construction, URL fetches, or output reflection, so relaxing this validation opens no new attack surface.
- Test coverage is well-targeted where it matters most: the subdomain-matching regression path (`adobe.com` matching `help.adobe.com`, `www.` stripping, non-matching exclusion) is proven, and the "hostname omitted" path is verified to preserve citation-descending sort and pagination.
- Docs (`docs/llmo-semrush-apis/filter-dimensions-apis.md`) were updated in the same PR to match the new optional contract.

## Issues

### Important (Should Fix)
- `src/support/elements/definitions/domain-urls.js:131` - the new `.trim()` call has no test exercising the case it exists to handle. A whitespace-only `hostname` (e.g. `?hostname=%20%20`) is truthy at the controller (`query.hostname || query.domain` passes it through), so without `.trim()` it would reach the transform as a non-empty value, set `filterByHostname = true`, and silently match zero rows for a caller who effectively asked for no filter. The two new transform tests cover only "hostname key absent" and "one valid hostname" - neither exercises the whitespace-collapses-to-no-filter path this line was added for. Add one assertion in the `transformDomainUrlsResponse` describe block with `hostname: '   '` asserting all hosts are returned, to lock in the behavior this line depends on.

### Minor (Nice to Have)
- `src/controllers/elements.js:1239-1242` - the four replaced/added JSDoc lines are indented one space further than the untouched opening `/**` and the lines directly above them, so the comment block is now visibly ragged. Cosmetic, ESLint doesn't catch it, but worth a one-character fix while the block is already being edited.

## Recommendations

- The `/* c8 ignore start -- LLMO-6160 POC endpoint; unit tests intentionally deferred */` comment above `listDomainUrls` is now inaccurate: this PR adds the first real unit tests for this handler. Consider removing the ignore now that coverage exists, or updating the comment so it no longer claims tests were deferred.
- Making `hostname` optional means the omitted-hostname path now retains every parseable row across all fanned-out projects in the in-memory Map/sort instead of one host's worth. This mirrors the pre-existing shape of the sibling `owned-urls` endpoint (which already has no hostname-style filter), so it's not a new failure mode, but it's worth a one-line comment or a follow-up ticket noting that "omit hostname" is now the widest in-memory case in this element family, so it doesn't get mistaken for a cheap default later.
- Optional, lower priority: a controller-level test asserting `params.hostname` for `?hostname=` (empty string) would document that the `||` fallback collapses an explicit empty value to `undefined`/`domain`, distinct from the whitespace case handled at the transform layer.

## Assessment
Ready to merge? With fixes.
Reasoning: The behavior change itself is small, correctly scoped, and safe - authorization is untouched, the single-hostname path is provably equivalent to the old behavior, and the omitted-hostname path is additive with sort/pagination verified. The one gap is test coverage for the whitespace-only hostname case, which is quick to close and protects the exact line this PR introduces to handle that input.

## Next Steps
1. Add the one missing test for whitespace-only `hostname` in `transformDomainUrlsResponse`.
2. Optionally tidy the JSDoc indentation and the stale c8-ignore comment while in the area.

AndreiAlexandruParaschiv (approved):

tb-adbe (approved):

Issue #3020, opened 2026-08-10
Resolve the LLMO/serenity flag per brand, so an org migrates to Semrush in waves


rainer-friederich:
## Goal

Serenity activation becomes a **per-brand** decision instead of an org-wide cutover, so Adobe Corp migrates in waves rather than all at once. Wave 1 is Adobe Helpx, Adobe Stock and Community Adobe.

Today one `feature_flags` row — `(organization_id, product='LLMO', flag_name='serenity')` — flips an entire organization onto the Semrush-backed read path, and every consumer reads that single value: the `/serenity/*` surface, the brand write-path fan-outs, the classic-UI edit lock, and the UI's data-source resolver. For a one-brand customer that is right. For Adobe Corp — 16 active brands, 91,671 active prompts — it is one irreversible cutover of everything at once.

After this change, a brand that has not been migrated keeps its legacy read path, its editable classic UI and its Postgres corpus, in the same organization, alongside migrated brands, indefinitely.

## The gate

`feature_flags` carries an optional `brand_id`. A row with `brand_id` NULL is the organization's value for that `(product, flag_name)`; a row naming a brand **overrides** it for that brand alone:

```
resolve(org, brand) := row (org, product, flag_name, brand_id = brand)
                       else row (org, product, flag_name, brand_id IS NULL)
                       else the flag does not exist for that brand
```

The brand row is an override, **not** a second condition ANDed with the organization's, so a brand can be on under an organization that is off — which is what every wave before the last one looks like, and it is why the org-level row is the last step of a rollout rather than the first.

An organization with one org-level row and no brand rows resolves identically for every brand it owns. Every already-migrated customer therefore behaves byte-for-byte as today, with **no backfill at all**, and only a mid-migration org can have the two scopes disagree.

The schema change is `adobe/mysticat-data-service` migration `20260810090000_feature_flags_brand_scope.sql`: a nullable `feature_flags.brand_id` (FK to `brands`, `ON DELETE CASCADE`), the unique key widened to `UNIQUE NULLS NOT DISTINCT (organization_id, product, flag_name, brand_id)`, and an index on `brand_id`.

The org-level `serenity_ui` flag stays org-wide and is **not** overridden per brand: it describes whether the org's *users* are provisioned in Semrush, which no single brand can differ on. It stays **off** while any brand is unmigrated — it is the completion marker of the last wave, not a step in it.

---

## Part 1 — flag storage (ships FIRST, independently, against today's schema)

[...]

PR #3021, opened 2026-08-10
fix(feature-flags): resolve the organization's own flag row independent of the unique key

rainer-friederich:
## 1. Abstract
The three `feature_flags` storage helpers in `src/support/feature-flags-storage.js` now resolve an organization's own flag row by predicate rather than through the table's unique key, so they are correct both against today's schema and against the schema that adds an optional `feature_flags.brand_id`.

## 2. Reasoning
Serenity activation is becoming a per-brand decision so that an organization migrates to Semrush in waves rather than as one irreversible cutover — Adobe Corp carries 16 active brands and 91,671 active prompts behind a single flag today. The gate for that is an optional `feature_flags.brand_id`: a row with `brand_id` NULL is the organization's value, and a row naming a brand overrides it for that brand alone.

Carrying two brands' overrides of one flag means the table's unique key has to widen to `UNIQUE NULLS NOT DISTINCT (organization_id, product, flag_name, brand_id)`, which drops `feature_flags_org_product_flag_name_key`. Any consumer that names that constraint as an `ON CONFLICT` target then fails with *"there is no unique or exclusion constraint matching the ON CONFLICT specification"* the moment the migration deploys, independent of whether a single override row exists. A partial unique index is not an escape: PostgreSQL only infers one for `ON CONFLICT` when the statement repeats its predicate, which a PostgREST upsert never emits.

This change is therefore a precondition for that migration, and it deliberately takes on no dependency in return, so the two can land in either order with no deploy coordination.

The helpers select the organization's row with a predicate that holds under both schemas — `brand_id` is absent from every row before the migration and NULL on the organization's row after it — so none of them names the new column. That is what lets this ship ahead of it.

- `upsertFeatureFlag` previously targeted the three-column key with `ON CONFLICT`. It now reads the matching rows, picks the organization's, and either updates that row keyed on its primary key or inserts a new one. This trades `ON CONFLICT`'s atomicity for a read-then-write: two concurrent creates of the *same* flag now leave one hitting the unique constraint as a retryable duplicate-key error rather than silently merging. Its callers are the admin endpoint, LLMO onboarding, and onboarding-mode remediation, which carry essentially no concurrency. Once the column exists this can move back to a four-column `ON CONFLICT` if the atomicity is wanted.
- `readFeatureFlag` resolved its row through `.maybeSingle()`, which rejects a multi-row result. As soon as any brand overrides a flag, that read would start throwing for the organization. It now selects the whole row set and picks the organization's row.
- `listFeatureFlagsByOrgAndProduct` returned every matching row unfiltered, so overrides would have surfaced as extra entries in `GET /organizations/:id/feature-flags/llmo`. It now returns the organization's rows only.

Nothing user-visible changes. `FeatureFlagDto` does not expose `brand_id`, so the endpoint's response shape, its security tiers, and the enable/disable semantics are all untouched, and project-elmo-ui and the DRS runner (which consume that endpoint rather than the table) need nothing.

For an organization with no brand override rows — every customer today — all three helpers return exactly what they returned before. This is behaviour-preserving on its own; the first observable change comes when a wave writes its first override.

An integration-test helper carried a second copy of the same `ON CONFLICT` upsert and now goes through the production helper, so the test setup path stays valid under whichever shape the unique key has.

## 4. Required information
- Jira / issue: https://github.com/adobe/spacecat-api-service/issues/3020
- Implementation plan: https://github.com/adobe/serenity-docs/issues/317
- Other: https://github.com/adobe/mysticat-data-service/pull/856 (the migration CLI that executes each wave) , https://jira.corp.adobe.com/browse/LLMO-6623 (the Semrush-scoped IMS promise-token pair, which bounds the calendar every wave runs against)

- **mysticat-data-service** — contract, and a blocking prerequisite of its own. It owns `feature_flags` and the pending `brand_id` migration. It also carries two consumers that name the dropped constraint and must be fixed before that migration deploys: `scripts/serenity_migration/persist.py` (lines 238 and 285, the wave-execution CLI, run against prod) and `scripts/seed-semrush-mock.sql` (lines 125 and 176, consumed by the local stack). Several of its reads additionally resolve a flag as `rows[0]` / `fetchone()` over an unordered result, and its two rollback `UPDATE`s carry no brand filter — correct until the first override row exists, wrong afterwards. `docs/feature-flags-schema.md` still documents the three-column key.
- **mystique** — contract only, no code here. `app/services/control/feature_flags.py` (line 552) has the same `ON CONFLICT`, but its write gate needs an opt-in that is off by default in every environment and refuses Prod/Stage regardless, so the path is unreachable in production; only the dev operator console is affected.
- **spacecat-audit-worker** — consumed, no change needed. `isBrandalfEnabled` reads the table with `.maybeSingle()`, which is safe while `brandalf` carries no brand overrides.
- **llmo-data-retrieval-service** — consumed, no change needed. It reads this service's feature-flags endpoint rather than the table, so it is covered by the `listFeatureFlagsByOrgAndProduct` filter here.
- **project-elmo-ui** — contract only, no change needed. Same endpoint, unchanged response shape.

The premise this change was specified against was checked against the live systems rather than taken from the plan:

- **The set of consumers that break was audited across every workspace repo, and it is larger than the issue states.** The issue asserts that `src/support/feature-flags-storage.js` is the only place depending on the three-column key. Five sites name it; the two outside this repo listed in section 5 are prod-relevant. Anyone re-checking this should read the files rather than sweep for a pattern: in `persist.py` the table name sits three to four lines above the `ON CONFLICT` clause inside a multi-line Python string, so a windowed grep for the two together finds nothing and reads as a clean bill of health.
- **Live prod schema and version confirmed** by querying the prod mysticat database: `feature_flags` carries `feature_flags_org_product_flag_name_key UNIQUE (organization_id, product, flag_name)` plus a primary key on `id`, and no `brand_id` column. The primary key is what the new read-then-write targets, which is why it is valid under both schemas. Prod runs PostgreSQL 16.11 and the integration-test image is `postgres:16-alpine`, so the `NULLS NOT DISTINCT` key the migration needs (PostgreSQL 15+) is available in both.
- The migration `20260810090000_feature_flags_brand_scope.sql` does not yet exist on `main` or on the CLI branch of mysticat-data-service, checked through the contents API on both refs. Part 2 of the issue — the per-brand predicate and the four gates — is consequently not in this PR; it cannot function until that column lands.

## 7. Test plan
Locally, beyond the unit suite, the new write path was exercised against real PostgreSQL and PostgREST through the integration harness rather than against fakes, which is the part that matters here: a read-then-write is exactly the kind of change that passes against stubs and fails on the wire.

- `test/it/postgres/feature-flags.test.js` drives both new branches end to end. `PUT /organizations/:id/feature-flags/llmo/brandalf` takes the insert branch against the seeded organization, and the subsequent `DELETE` takes the update branch on the row that PUT created. It also re-reads each write back through `readFeatureFlag`, so the endpoint and the storage helper are shown to agree on the same row.
- `test/it/postgres/brand-for-org-site.test.js` covers the integration helper that was rewritten to use the production helper.
- `test/it/postgres/serenity.test.js` and `test/it/postgres/activate-brand-for-org.test.js` exercise `readFeatureFlag` through `isSerenityActiveForOrg` against a seeded `LLMO/serenity=true` row, confirming the rewritten read resolves the same value the `.maybeSingle()` version did.

Run them with:

    npx mocha --require test/it/postgres/harness.js --timeout 60000 test/it/postgres/feature-flags.test.js test/it/postgres/brand-for-org-site.test.js
    npx mocha --require test/it/postgres/harness.js --timeout 60000 test/it/postgres/serenity.test.js test/it/postgres/activate-brand-for-org.test.js

On **dev**, after deploy:

1. `PUT /organizations/:orgId/feature-flags/llmo/<flag>` on an organization with no such flag, then `GET /organizations/:orgId/feature-flags?product=LLMO` — the flag appears with `flagValue: true`. This is the insert branch.
2. `DELETE` the same flag, then `GET` again — `flagValue: false` and the row is the same `id` as before, not a new one. This is the update branch, and the stable `id` is what shows the read-then-write found the existing row rather than creating a second.
3. `PUT` the same flag a second time — still one row, still the same `id`, `flagValue` back to `true`. Repeating a write must stay idempotent now that `ON CONFLICT` is no longer doing that work.
4. Confirm an organization already carrying `LLMO/serenity=true` still reads as serenity-active on the `/serenity/*` surface, and that a brand edit touching URLs / competitors / aliases still fans out — those paths read the flag through the rewritten helper.

On **prod**, this is behaviour-preserving with no override rows in existence, so the check is negative: after deploy, confirm the admin feature-flag endpoints still write (the operation most likely to regress, since it changed shape), and that no new `Failed to upsert feature flag` or `Failed to read feature flag` appears in the service logs.

This PR is independently deployable and has no prerequisite. It is, however, the first step of a sequence whose later steps depend on it:

- https://github.com/adobe/mysticat-data-service/pull/856 — related. It carries the wave-execution CLI whose `persist.py` names the dropped constraint; that CLI must be corrected before the migration deploys, independently of this PR.
- The `feature_flags.brand_id` migration in mysticat-data-service — depends on this PR being deployed, and equally on the mysticat-data-service consumer fixes described in sections 5 and 6. It is not yet written.
- Part 2 of https://github.com/adobe/spacecat-api-service/issues/3020 (the per-brand predicate and the `/serenity/*`, brand create/edit, and brand-presence gates) — depends on the migration.

Ordered sequence: this PR → the mysticat-data-service consumer fixes (`persist.py`, `seed-semrush-mock.sql`, the unordered reads, the rollback `UPDATE`s, `docs/feature-flags-schema.md`) → the `brand_id` migration → Part 2 in this repo → the per-brand UI in project-elmo-ui → wave 1 execution. Merging this PR alone is safe at any time.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Igor Grubic", "Dominique Jäggi"]
```

### DIFF
--- a/src/support/feature-flags-storage.js
+++ b/src/support/feature-flags-storage.js
@@ -40,7 +40,26 @@ export function isValidFeatureFlagName(flagName) {
 }
 
 /**
- * Upserts a boolean feature flag row (org + product + flag_name unique).
+ * The organization's own row, not a brand's override of it. `brand_id` is absent
+ * from every row before the brand-scope migration and NULL on the organization's
+ * row after it, so this selects correctly under both schemas.
+ *
+ * @param {object} row - Raw PostgREST `feature_flags` row.
+ * @returns {boolean} `true` for the organization-level row.
+ */
+const isOrgRow = (row) => (row.brand_id ?? null) === null;
+
+/**
+ * Writes an organization's boolean feature flag, creating the row when it does
+ * not exist yet and updating it in place when it does. A brand's override of the
+ * same flag is left untouched.
+ *
+ * Keyed on the row's primary key rather than a composite `ON CONFLICT` target,
+ * so it does not depend on the shape of the table's unique key. That costs
+ * atomicity: two concurrent creates of the *same* flag leave one hitting the
+ * unique constraint as a retryable duplicate-key error. The write callers are
+ * the admin endpoint, onboarding and mode remediation, which carry essentially
+ * no concurrency.
  *
  * @param {object} params
  * @param {string} params.organizationId - SpaceCat org id (matches mysticat organizations.id)
@@ -63,19 +82,36 @@ export async function upsertFeatureFlag({
     throw new Error('PostgREST client is required for feature flags');
   }
 
-  const row = {
-    organization_id: organizationId,
-    product,
-    flag_name: flagName,
-    flag_value: value,
-    updated_by: updatedBy,
-  };
-
-  const { data, error } = await postgrestClient
+  const { data: existing, error: readError } = await postgrestClient
     .from('feature_flags')
-    .upsert(row, { onConflict: 'organization_id,product,flag_name' })
-    .select()
-    .single();
+    .select('*')
+    .eq('organization_id', organizationId)
+    .eq('product', product)
+    .eq('flag_name', flagName);
+
+  if (readError) {
+    throw new Error(`Failed to upsert feature flag: ${readError.message}`);
+  }
+
+  const current = (existing ?? []).find(isOrgRow);
+  const { data, error } = current
+    ? await postgrestClient
+      .from('feature_flags')
+      .update({ flag_value: value, updated_by: updatedBy })
+      .eq('id', current.id)
+      .select()
+      .single()
+    : await postgrestClient
+      .from('feature_flags')
+      .insert({
+        organization_id: organizationId,
+        product,
+        flag_name: flagName,
+        flag_value: value,
+        updated_by: updatedBy,
+      })
+      .select()
+      .single();
 
   if (error) {
     throw new Error(`Failed to upsert feature flag: ${error.message}`);
@@ -85,7 +121,8 @@ export async function upsertFeatureFlag({
 }
 
 /**
- * Reads a single feature flag value by org, product, and flag name.
+ * Reads an organization's own value for a feature flag, ignoring any brand's
+ * override of it.
  *
  * @param {object} params
  * @param {string} params.organizationId
@@ -106,20 +143,23 @@ export async function readFeatureFlag({
 
   const { data, error } = await postgrestClient
     .from('feature_flags')
-    .select('flag_value')
+    .select('*')
     .eq('organization_id', organizationId)
     .eq('product', product)
-    .eq('flag_name', flagName)
-    .maybeSingle();
+    .eq('flag_name', flagName);
 
   if (error) {
     throw new Error(`Failed to read feature flag ${flagName}: ${error.message}`);
   }
 
-  return typeof data?.flag_value === 'boolean' ? data.flag_value : null;
+  const row = (data ?? []).find(isOrgRow);
+  return typeof row?.flag_value === 'boolean' ? row.flag_value : null;
 }
 
 /**
+ * Lists an organization's own enabled flags for a product. Brand overrides are
+ * excluded, so the endpoint built on this describes the organization's state.
+ *
  * @param {object} params
  * @param {string} params.organizationId
  * @param {'ASO'|'LLMO'} params.product
@@ -147,5 +187,5 @@ export async function listFeatureFlagsByOrgAndProduct({
     throw new Error(`Failed to list feature flags: ${error.message}`);
   }
 
-  return data ?? [];
+  return (data ?? []).filter(isOrgRow);
 }
--- a/src/support/feature-flags-storage.js
+++ b/src/support/feature-flags-storage.js
@@ -44,6 +44,12 @@ export function isValidFeatureFlagName(flagName) {
  * from every row before the brand-scope migration and NULL on the organization's
  * row after it, so this selects correctly under both schemas.
  *
+ * Every query feeding this predicate selects the full row, and must keep doing
+ * so. Naming `brand_id` in a projection fails against the current schema, where
+ * the column does not exist yet; omitting it once it does exist makes every
+ * override row arrive with `brand_id: undefined` and read as the organization's
+ * own. A wildcard projection is the only shape that is correct under both.
+ *
  * @param {object} row - Raw PostgREST `feature_flags` row.
  * @returns {boolean} `true` for the organization-level row.
  */
@@ -82,6 +88,7 @@ export async function upsertFeatureFlag({
     throw new Error('PostgREST client is required for feature flags');
   }
 
+  // Wildcard projection is required — see `isOrgRow`.
   const { data: existing, error: readError } = await postgrestClient
     .from('feature_flags')
     .select('*')
@@ -141,6 +148,7 @@ export async function readFeatureFlag({
     throw new Error('PostgREST client is required for feature flags');
   }
 
+  // Wildcard projection is required — see `isOrgRow`.
   const { data, error } = await postgrestClient
     .from('feature_flags')
     .select('*')
@@ -175,6 +183,7 @@ export async function listFeatureFlagsByOrgAndProduct({
     throw new Error('PostgREST client is required for feature flags');
   }
 
+  // Wildcard projection is required — see `isOrgRow`.
   const { data, error } = await postgrestClient
     .from('feature_flags')
     .select('*')
--- a/test/controllers/feature-flags.test.js
+++ b/test/controllers/feature-flags.test.js
@@ -33,6 +33,24 @@ describe('FeatureFlagsController', () => {
   let mockHasAdmin;
   let baseCtx;
 
+  /**
+   * Wires `postgrestClient.from` for a flag write: the org-row lookup finds
+   * nothing, so the write lands as an insert that resolves to `row`.
+   *
+   * @param {object} row - The row the write returns.
+   * @returns {object} The `insert` stub, for asserting the written payload.
+   */
+  function stubFlagWrite(row) {
+    const eq3 = sandbox.stub().resolves({ data: [], error: null });
+    const eq2 = sandbox.stub().returns({ eq: eq3 });
+    const eq1 = sandbox.stub().returns({ eq: eq2 });
+    const select = sandbox.stub().returns({ eq: eq1 });
+    const single = sandbox.stub().resolves({ data: row, error: null });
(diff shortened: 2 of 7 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-scoped, behaviour-preserving change with thorough test coverage and an explicitly documented atomicity trade-off.
**Complexity: HIGH** - large diff (733 changed lines across 7 files).
_Changes: Replaces ON CONFLICT upsert with a read-then-write pattern that resolves the organization's own feature-flag row by predicate, decoupling the helpers from the table's unique key shape ahead of a brand-scope migration (7 files)._

<details>

- suggestion: `readFeatureFlag` now uses `.select('*')` but only needs `flag_value` and `brand_id` for the `isOrgRow` predicate - consider `.select('flag_value, brand_id')` to keep the projection tight and express the function's minimal contract - `src/support/feature-flags-storage.js:148`
- suggestion: The documented duplicate-key race on concurrent inserts surfaces as a generic `Failed to upsert feature flag` error with no retry and no distinction from a real DB failure - consider detecting PostgREST error code `23505` to either retry once or surface a more specific error to callers - `src/support/feature-flags-storage.js:117`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 1s | <b>Cost:</b> $4.80 | <b>Commit:</b> <code>d9d200eee8cbd45314025fe0d351dddd5dc560d4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Validated on the shared `dev-branches` environment (`https://spacecat.experiencecloud.live/api/ci/`), against Lambda version 7758 — confirmed from the CloudWatch stream that served each request, not from the deployment record, because another branch claimed the slot at 09:34:00Z while the `latest` alias only moved off 7758 at 09:35:17Z. All three writes below landed inside that window on this branch's build.

Target: the AEM Sites Engineering organization on dev, using a throwaway flag name no consumer reads. Dev has no `brand_id` column yet, so this exercised the pre-migration schema — the half that has to keep working when this ships ahead of the column.

- `PUT /organizations/:orgId/feature-flags/llmo/pr_3021_validation` on an organization with no such flag → 200, new row created with `createdAt` equal to `updatedAt`. This is the insert branch.
- `DELETE` the same flag → 200, `flagValue: false` on the **same row id**, `createdAt` unchanged and `updatedAt` advanced. This is the update branch, and the stable id is what shows the read-then-write resolved the existing row instead of creating a second one.
- `PUT` the same flag again → 200, `flagValue: true`, still the same row id. Repeating a write stays idempotent now that `ON CONFLICT` is no longer providing that.
- `GET /organizations/:orgId/feature-flags?product=LLMO` → 200, returns the organization's rows only.
- Direct read of the table afterwards: exactly one row for that flag, no duplicate. The organization's pre-existing `brandalf` row was untouched, its `updatedAt` unchanged — so the update keyed on the primary key modified only its own row.
- Logs (`/aws/lambda/spacecat-services--api-service`, stream `[7758]`): the `GET`, `PUT` and `DELETE` routes all appear, and no `Failed to read feature flag`, `Failed to upsert feature flag` or `Failed to list feature flags` was emitted. The unrelated site-config validation errors present in the same window are pre-existing and come from other traffic.

The throwaway row was deleted afterwards, leaving the organization with the single `brandalf` row it started with.

Not covered here: the post-migration schema, which cannot be exercised until the `brand_id` column exists. That path is covered by unit tests asserting the organization's row is picked when a brand override row is also present.

rainer-friederich:
Both non-blocking items reviewed. Neither is applied as proposed, and one of them turned out to be actively unsafe — but it pointed at a real gap, which is fixed in `9e6c7b6f2`.

MysticatBot:

- suggestion: narrow `readFeatureFlag` to `.select('flag_value, brand_id')` — **declined, and it must not be applied.** `brand_id` does not exist yet; it arrives with the migration this change is designed to precede. Verified against the dev database, where `select flag_value, brand_id from feature_flags` fails with `column "brand_id" does not exist`. PostgREST surfaces that as a request error, so `readFeatureFlag` would throw on every call, and because `isSerenityActiveForOrg` fails safe to off, every serenity-gated route would silently resolve to off for every organization. That is precisely the regression this change exists to prevent. The narrower projection is also wrong in the other direction: `isOrgRow` reads `row.brand_id ?? null`, so once the column does exist but is left out of the projection, every brand override row arrives with `brand_id: undefined` and is misread as the organization's own row — returning a brand's value as the organization's. A wildcard projection is the only shape correct under both schemas, which makes it load-bearing rather than loose.

  The item was still worth acting on, because nothing in the code said any of that. `9e6c7b6f2` records the constraint in both directions on `isOrgRow` and marks each of the three affected queries, so the next reader reaching for a tighter projection sees why it cannot be tightened. Comments only — no behaviour change.

- suggestion: detect PostgREST `23505` to retry once or surface a more specific error — **declined.** Letting the duplicate key throw is deliberate: the callers are the admin endpoint, LLMO onboarding, and onboarding-mode remediation, none of which run concurrently against the same flag, and a silent retry would paper over a genuine conflict rather than report it. The error is also less opaque than it looks: the thrown message interpolates the underlying PostgREST message, which names the violated constraint, so a duplicate key is already distinguishable from an arbitrary database failure in logs. Once `brand_id` lands, the write can return to a four-column `ON CONFLICT`, which restores atomicity properly instead of simulating it with a retry.

On the degraded-review note: there is no spec document to link. The design of record for this change is issue 3020 itself, and the implementation plan it sits under is already linked in section 4 of the description. Re-requesting would produce the same note rather than a fuller pass, so section 4 is left as it stands.

rainer-friederich:
The runtime validation above was run against `d9d200eee`. The commit that followed it changes comments only — no statement, projection or control flow differs — so the behaviour it recorded is unchanged and was not re-run. All checks are green on the current head, `codecov/patch` included.

igor-grubic (approved):

PR #3024, opened 2026-08-10
feat(serenity): resolve the LLMO/serenity flag per brand, so an org migrates in waves

rainer-friederich:
## 1. Abstract

The `LLMO/serenity` feature flag becomes resolvable per brand: a `feature_flags` row naming a brand overrides its organization's row for that brand alone. This is Part 2 (work items B1–B5) of https://github.com/adobe/spacecat-api-service/issues/3020 .

## 2. Reasoning

Today one flag row flips an entire organization onto the Semrush-backed read path, and every consumer reads that single value. For a one-brand customer that is right. For Adobe Corp — 16 active brands, 91,671 active prompts — it makes migration a single irreversible cutover of everything at once.

Resolving the flag per brand turns that into a wave-by-wave rollout: an unmigrated brand keeps its legacy read path, its editable classic UI and its Postgres corpus, in the same organization, alongside migrated brands, indefinitely. Wave 1 is Adobe Helpx, Adobe Stock and Community Adobe.

It also closes a live rollout hole. Flipping an organization on while its brands are still unbound resolved each of them to flat mode against the organization's parent workspace, so the UI rendered a brand with zero markets instead of falling back to its legacy data.

A row with no `brand_id` is the organization's value for a flag; a row naming a brand overrides it for that brand. Resolution is: the brand's own row, else the organization's row, else off.

The brand row is an **override**, not a second condition ANDed with the organization's. So a brand is active under an organization that is not — which is what every wave before the last one looks like, and why the organization's row is the last step of a rollout rather than the first. A `false` brand row conversely holds one brand back from an organization that is on. An organization carrying only its own row resolves identically for every brand it owns, so an already-migrated customer behaves exactly as before, with no backfill.

Behaviour deltas:

- **The per-brand predicate.** A new `isSerenityActiveForBrand` becomes the central predicate. The flag cache now holds *both* scopes of a flag in one entry per organization and flag, so a single read answers for the organization and for any of its brands and the two views cannot disagree inside a request. The positive cache TTL drops from five minutes to the short brand TTL, because a brand's value flips mid-rollout and the cache is process-local — a flip on one Lambda container otherwise leaves its siblings serving the old answer.
- **The `/serenity/*` surface** gates on the resolved per-brand value after brand resolution, rather than on the organization's flag before it. A brand that does not resolve active answers 404, which is the contract the UI already handles as "no serenity here". This is what closes the zero-markets hole above.
- **Brand edit.** The fan-out gate is now per brand: an edit touching URLs, social accounts, earned content, competitors or aliases is rejected with 403 when *that brand* is bound to a sub-workspace but not yet released, and passes through as ordinary backend data when the brand was never migrated. So an unreleased brand keeps its editable classic UI while a released sibling is locked — where before all 16 brands got one answer.
- **Brand creation stays organization-level**, because a create has no brand to resolve against, and the born-in-Semrush invariant is a property of the organization. A brand created in an active organization needs no override row: with none, it inherits the organization's value, which is the intent. The `serenity_ui` error-surfacing allowance also stays organization-level — it describes whether the organization's *users* are provisioned in Semrush, which no single brand can differ on.
- **Brand presence and the workspace-access probe.** Both authorizers in the elements controller now gate per brand. The org-scoped one was the real gap: it falls back to the organization's parent workspace when a brand has no binding, so the brand-presence reads and the access probe were answering for the organization instead of for an unmigrated brand.
- **The brand payload** carries two derived read-only fields, `serenityActive` and `serenityActivatedAt` (the resolved row's `updated_at`, null while inactive), so project-elmo-ui never re-implements the resolution rule. A mid-migration organization returns both answers in one list, so consumers must read it per brand. The handler returning the payload derives them from a single flag read per response — a 16-brand list resolves from one query — so the shared brand readers stay free of that query and the ~20 internal callers that never surface these fields neither pay for it nor inherit its failure mode. On the write handlers the read is issued before the write, so a flag-read fault cannot turn an edit that already committed into a 500.

Nothing names `brand_id` in a PostgREST filter or projection. Every row arrives through a wildcard projection and `brand_id` is only ever read off it, so against a schema without the column it is absent and resolution degrades to exactly the organization-level answer. That is what makes this safe to deploy independently of the migration — see section 8.

Two deliberate departures from the issue text, both worth a reviewer's attention:

- The issue has the `/serenity/*` gate stay ahead of brand resolution so an inactive organization "never leaks brand existence", and would have both denials share one 404 body. Access control on the organization already runs before that point, and any caller who passes it can enumerate the organization's brands through the brands endpoint — so the precise "brand not found" body hides nothing, and it stays diagnosable mid-wave. The bodies are kept distinct.
- The issue asks the post-commit Semrush re-sync block to gain the active condition. It does not get one. Its condition is exactly the same set of touched fields as the gate earlier in the handler, and that gate already rejects an inactive bound brand outright, so a check there would guard a state that cannot be reached. The intent — a provisioned-but-unreleased brand never fans out — is met by the gate, and the reasoning is recorded at the call site so it does not read as an omission.

Documentation is brought back in line: the serenity activation section described an organization-wide switch, the old two-TTL cache, and a brand read payload that was "deliberately left alone", all of which this change makes wrong. The OpenAPI brand schema documents the two new fields.

## 4. Required information

- Jira / issue: https://github.com/adobe/spacecat-api-service/issues/3020
- Implementation plan: https://github.com/adobe/serenity-docs/issues/317
- Other: Part 1 (the storage change this builds on) https://github.com/adobe/spacecat-api-service/pull/3021 — the schema change https://github.com/adobe/mysticat-data-service/pull/891 — the wave-execution CLI that writes the brand override rows https://github.com/adobe/mysticat-data-service/pull/856

- **mysticat-data-service** — schema dependency. Supplies `feature_flags.brand_id`, the widened `UNIQUE NULLS NOT DISTINCT (organization_id, product, flag_name, brand_id)` key and the composite foreign key. Also the writer: the wave-execution migration CLI is what creates brand override rows, so no endpoint here writes them.
- **project-elmo-ui** — consumes the new `serenityActive` / `serenityActivatedAt` fields. It must read them per brand rather than deriving one answer per organization; the per-brand UI work itself is separate and is inert until an organization has a mixed brand set.
- **llmo-data-retrieval-service** — consumes the organization feature-flags endpoint, whose response shape is unchanged: it continues to report the organization's own rows only, with overrides excluded.

The schema state was verified directly against all three databases on 2026-08-10 rather than inferred from the migration's merge status:

- **dev**, **stage** and **prod** (`mysticat-data-service` v5.90.0 applied) all carry `feature_flags.brand_id`, the unique key `UNIQUE NULLS NOT DISTINCT (organization_id, product, flag_name, brand_id)`, and a **composite** foreign key `(organization_id, brand_id) -> brands(organization_id, id) ON DELETE CASCADE`. The composite shape is stronger than the issue describes — an override cannot name a brand the organization does not own — and the integration tests now assert exactly that. All three are PostgreSQL 16.11.
- `brands.serenity_activated_at`, the column an earlier local plan document specified as the activation stamp, exists in no environment. That design was superseded by `feature_flags.brand_id`, which is what shipped; the resolution rule implemented here follows the issue, not that document.

## 7. Test plan

Verified locally this session:

- The feature-flags integration suite was run against real PostgreSQL 16 with the migration applied, on the bumped image. Beyond the resolution rule, it proves against the live constraint that the widened key admits an organization row and a brand override for the same flag simultaneously (the previous three-column key could not hold both), that a duplicate override for the same brand is still rejected, and that the composite foreign key refuses an override naming another organization's brand. It also asserts the organization-level read and the admin GET endpoint still report only the organization's value when an override exists, and that the brand payload's `serenityActive` flips with the override end to end.
- The integration database pin moves from `mysticat-data-service` v5.81.0 to v5.90.0, the first tag carrying the migration. Ten migrations land in that jump; all are additive. The whole local integration suite was then run against the bumped image to confirm the jump breaks nothing elsewhere, and none of the other suites regressed.
- **Outstanding at the time of opening:** dev runtime validation. It will be reported on this PR before it is considered ready.

To verify on **dev** (which has the column):

- With an organization whose `LLMO/serenity` row is on and no overrides, confirm `GET /v2/orgs/:orgId/brands` reports `serenityActive: true` with a `serenityActivatedAt` timestamp for every brand, and that a `/serenity/*` brand route still serves.
- Insert a `false` override for one of those brands, then confirm that brand reports `serenityActive: false` with a null `serenityActivatedAt`, that its `/serenity/*` routes answer 404, that a URL/competitor/alias edit on it is rejected with 403 while the same edit on a sibling brand still succeeds and fans out, and that the brand-presence access probe answers 404 for it.
- Confirm `GET /organizations/:orgId/feature-flags?product=LLMO` still reports a single `serenity` entry carrying the organization's value, unaffected by the override.

On **prod**, the observable behaviour must not change at all until wave 1 writes its first override row: with no override rows anywhere, every brand resolves to its organization's value, which is what the organization-wide gate answered before.

- **depends-on** https://github.com/adobe/mysticat-data-service/pull/891 — the migration itself. Merged.
- **depends-on** https://github.com/adobe/mysticat-data-service/pull/896 — promotes v5.90.0 to prod. Merged, and applied: the column, the widened unique key and the composite foreign key are confirmed present in production, verified against the database rather than inferred from the merge.
- **related** https://github.com/adobe/mysticat-data-service/pull/856 — the wave-execution CLI that writes the brand override rows. Nothing observable changes until it runs, so it is not a merge blocker for this PR.

Merging to `main` on this repo deploys straight to production, and the schema this depends on is already there, so there is no schema gate left on merging this.

The implementation is schema-agnostic regardless: because `brand_id` is never named in a filter or projection and is only read off a wildcard-selected row, an environment without the column resolves every brand to its organization's value — with no error path — rather than failing. That property is what made this safe to build and review ahead of the promotion; it now also means a rollback of the data-service version would not break this service.

Production behaviour does not change on deploy: with no override rows written yet, every brand resolves to its organization's value, which is what the organization-wide gate answered before.

Ordered sequence: 891 and 896 (both done, prod carries the schema) → this PR → the per-brand UI in project-elmo-ui → 856 and wave 1 execution.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Igor Grubic", "Dominique Jäggi"]
```


rainer-friederich:
## Local review pass, before requesting a bot review

Three independent reviews were run over the diff — correctness, simplicity/DRY, and project conventions. Recording what each found, what was fixed, and what was declined, so none of it gets re-raised.

**Correctness — no findings.** It independently checked the resolution rule at every path, cross-brand isolation, null-deref, fail-closed defaults, and the prod-safety constraint (confirming `brand_id` appears as a literal filter or insert key only in test seeding, never in production code). Two things it verified more rigorously than the original reasoning did, both worth having on the record:

- The claim that the post-commit re-sync needs no serenity check of its own holds. `touchesSemrushSync` and `urlsTouched || competitorsTouched || aliasesTouched` are not merely overlapping, they are the *same* boolean. It then closed the remaining hole in the argument by tracing `updateBrand`'s patch whitelist and confirming it never writes `semrush_sub_workspace_id` — so the pre-write and post-write reads of that column cannot diverge within the handler, making the earlier gate a strict precondition of the re-sync block.
- In `authorizeOrg`, the id handed to the gate is not merely UUID-shaped: `getBrandIdentity` has already confirmed it is a real brand row scoped to that organization.

**Fixed from review:**

- The flag row query was built in two places — `upsertFeatureFlag`'s read-before-write duplicated `readFeatureFlagScopes`. Both now call one private helper that returns the raw result, so each keeps the failure message its callers depend on (the read path names the flag, the write path does not).
- The integration suite's non-existent brand id became a `seed-ids.js` constant, matching the existing `NON_EXISTENT_*` set.
- The brand-override integration suite resets per test rather than once, which no other suite here needs. That is now explained where it happens: the collision it prevents is what one of those tests asserts, and switching to a different brand per test is not possible when the composite foreign key requires a brand the organization owns and this organization owns exactly one.
- The derived `serenityActive` field is read fresh while the gates read a short-lived cache, so for up to one cache TTL after a flip the payload and the gates can disagree. That consequence is now stated at the call site rather than left implicit. It is bounded, the gates fail closed, and no write is half-applied by it.

**Declined:**

- Moving the serenity flag identity constants out of `serenity-active.js` and into `feature-flags-storage.js`, to remove the one edge from `brands-storage.js` into the serenity directory. `feature-flags-storage.js` is deliberately product-agnostic — it serves any ASO or LLMO flag — so placing a serenity-specific identity there trades one layering edge for a worse one. `brands-storage.js` importing it is honest, since what it derives from it is a serenity field. Raised as a preference rather than a defect.

**Coverage:** `feature-flags-storage.js` is at 100% statements, branches, functions and lines. The two new 404 branches in the elements authorizers have dedicated tests.

Still outstanding: dev runtime validation, which will be reported here before this is considered ready.

igor-grubic:
### Concern: making `serenityScopes` a required param of `mapDbBrandToV2` couples every brand read to a `feature_flags` query

Because `serenityScopes` is required and non-defaulted, `getBrandById`, `getBrandBySite`, and `listBrands` now **unconditionally** issue an extra `feature_flags` read and inherit a "throw if that read fails" failure mode. That reaches ~20 call sites that never touch `serenityActive` / `serenityActivatedAt`, including:

- **Unrelated features** — brand-claims Slack commands (`toggle-brand-claims`, `run-brand-claims`), `llmo-opportunities-controller`, `sites.js` attach.
- **The PATCH write path** — `getBrandById` is called up to ~4× (region guard, competitor guard, wipe-check) plus `updateBrand`'s own return, each now doing an extra flag read.

Two effects:

1. **Efficiency** — a single PATCH can issue several extra `feature_flags` reads it doesn't need.
2. **Robustness** — `updateBrand` returns via `getBrandById`, so a `feature_flags` read failure *after* the brand row is committed turns a successful write into a 500. Same DB so low probability, but it cuts against this handler's otherwise-careful "never 5xx after commit" design.

The "fail loud rather than silently default to inactive" intent is right for the **brand-payload endpoints**. Could we make the derivation opt-in — resolve the scopes in the controller and pass them, or gate it behind an explicit option on the reader — so the payload endpoints get the derived fields while the internal reads don't pay for (or fail on) serenity resolution?

Not blocking if you'd rather accept the coupling consciously — mainly want it to be a deliberate call rather than a side effect of the required param.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-designed per-brand resolution with comprehensive test coverage and safe schema-agnostic deployment strategy.
**Complexity: HIGH** - large diff (2539 lines, 18 files).
_Changes: Resolves the LLMO/serenity feature flag per brand instead of per organization, enabling wave-by-wave migration within a single org (18 files)._
Note: Recommend a human read before merge - this change modifies project documentation (docs/serenity.md) that forms part of the service's operational reference, and the conventions reviewer flagged spec divergence from issue 3020 (heuristic, both divergences are explicitly documented and justified in the PR body). The bot review is a complement to, not a replacement for, a human read here.

<details>

- suggestion: Inconsistent 404 response shape between the two `elements.js` authorizers - `authorizeOrg` returns `notFound('...')` (plain message) while `authorizeBrandSubWorkspace` returns `createResponse({ error: 'serenityInactive', message: '...' }, 404)` (structured). Aligning both to the structured shape would let the UI reliably distinguish "brand not found" from "serenity inactive" regardless of code path. - `src/controllers/elements.js:385` vs `:435`
- suggestion: `readFeatureFlag` and `upsertFeatureFlag` now delegate to `readFeatureFlagScopes` which fetches all rows (org + every brand override), then discards everything except the org row. An optional `.is('brand_id', null)` filter on the shared `fetchFeatureFlagRows` for org-only callers would avoid the extra deserialization. Bounded at current scale (<50 brands per org). - `src/support/feature-flags-storage.js:239`
- nit: The cache TTL drop from 5min/30s to a uniform 10s also affects `isSerenityUiActiveForOrg`, which does not flip mid-rollout and would have been fine at the old positive TTL. The extra PostgREST reads are negligible at current QPS but unintentional. - `src/support/serenity/serenity-active.js:236`
- nit: `readSerenityScopes` in `brands-storage.js` documents that the payload and the gate cache "can disagree for up to one cache TTL after a flip" - the disagreement is actually bidirectional (the payload can also be stale while the gate is fresh), so the JSDoc could be tightened. No runtime impact. - `src/support/brands-storage.js:164`
- nit: `brandsOnly` test helper routes all non-`feature_flags` tables to the same chainable query response. A future test adding a second table read could get unexpected data from it. Fine for current coverage. - `test/support/brands-storage.test.js:65`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 43s | <b>Cost:</b> $8.74 | <b>Commit:</b> <code>e99f7429cd56a14524ca87fb6628acd0106f78d4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
## Addressed: @igor-grubic's coupling concern, and the five non-blocking review items

Pushed as `refactor(brands): derive serenity state at the payload boundary`.

### The coupling — fixed, and it was worse than described

Both effects checked out, and the robustness one bites harder than the write-up says. `updateBrand`'s closing `getBrandById` is not inside a `try`, so a `feature_flags` read failure there did not merely turn a committed write into a 500 — it also escaped **before** the post-commit Semrush re-sync block, so the fan-out was skipped and the divergence never logged. In a handler whose LLMO-6545 block absorbs re-sync failures specifically so a committed edit never 5xxes, that was the one path out the back.

Also confirmed the reach: the readers are shared with the brand-claims Slack commands, all five `getBrandBySite` calls in the elements authorizers, site attach, the opportunities controller, and the edit handler's own four pre-write guards — none of which surface these fields.

So the derivation moved out of the readers and onto the handlers that actually return a brand payload:

- `mapDbBrandToV2` is back to taking just a row. `readSerenityFlagScopes` and a pure `withSerenityState` are exported, and the six brand-payload responses — list, by-id, by-site, create, update, status-transition — apply them. One flag read per response, unchanged for the payload endpoints; zero for every internal read.
- A PATCH now issues one, not up to five.
- On the three write handlers the rows are resolved **before** the write, so a flag-read fault is a pre-commit error rather than a 500 on an edit that persisted. That closes the escape above rather than just narrowing it.
- The fields stay on every brand payload, so a consumer merging a PATCH or create response into its store cannot silently drop a brand back to the legacy read path.

A regression test asserts the readers issue no `feature_flags` query and return no `serenityActive`, so the coupling cannot quietly come back.

One thing this made obvious in support: the reason the storage test fixtures needed a table-aware `from` stub at 60 call sites was precisely that the coupling was too broad. Those are plain stubs again and the helper is gone — which also disposes of non-blocking item 5.

### The five non-blocking items

- **Inconsistent 404 shape between the two elements authorizers — declined.** Each gate matches the shape its own function already uses: `authorizeBrandSubWorkspace` returns structured bodies throughout (`subWorkspaceRequired`, `workspaceMisconfigured`), `authorizeOrg` returns plain `notFound` throughout (`Brand not found for this organization`, `Brand has no resolvable Semrush workspace`). Unifying them means changing pre-existing error contracts in one authorizer or the other, which is a separate change from this one.
- **An `.is('brand_id', null)` filter for the org-only callers — declined, and deliberately so.** Not naming `brand_id` in any filter or projection is the property that makes this deployable against production, which does not have the column yet. The filter would turn today's clean degradation into a hard failure on every org-level flag read. The row set it would trim is the same one bounded at under 50 rows per org.
- **The uniform 10s TTL also applies to `isSerenityUiActiveForOrg` — declined.** That is the chosen design, not an oversight: one cache entry per organization-and-flag with one TTL constant. `serenity_ui` gets its own entry under the same constant, so the cost is one indexed read per 10s per container. A second TTL constant to save that on a rarely-hit path is not worth the branch.
- **The staleness note is one-directional — fixed.** It now states both directions: a brand released mid-TTL reports active while a gate on an unexpired entry still refuses it, and a brand rolled back reports inactive while a gate still admits it. The gates remain the authority and fail closed.
- **The `brandsOnly` test helper — removed.** No longer needed once the readers stopped querying `feature_flags`.

### One more, found while verifying coverage

`serenityActivatedAt` had an `?? null` fallback for a missing `updated_at`. That column is `NOT NULL DEFAULT now()` in every environment, so the fallback guarded a state the schema cannot produce and left a branch no honest test could reach. It reads `updated_at` directly now.

### Verification

The brand storage and controller suites, the serenity and elements controller suites, the predicate and flag-storage suites, and the internal-consumer suites (sites, opportunities, both brand-claims commands) are green. Lint, both type-check tiers and `docs:lint` pass. No new uncovered statement or branch on any changed line. The feature-flags integration suite was re-run against real PostgreSQL 16 with the migration applied — 21 passing, including the end-to-end assertion that the brand payload's `serenityActive` flips with an override, which is the path this restructure moved.

`docs/serenity.md` records where the derivation lives and why the write paths resolve before writing.

Still outstanding: dev runtime validation, reported here before this is considered ready. The merge hold on https://github.com/adobe/mysticat-data-service/pull/896 is unchanged.

igor-grubic (approved):

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - clean refactoring that moves serenity-state derivation to the correct layer, with no regressions.
**Complexity: HIGH** - large diff (2094 lines, 18 files).
_Changes: Resolves the LLMO/serenity feature flag per brand instead of per organization, enabling wave-by-wave migration within a single org (18 files)._
Note: Recommend a human read before merge - this change modifies project documentation (docs/serenity.md) and the OpenAPI schema (docs/openapi/schemas.yaml). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: On the write handlers (updateBrand, setBrandStatus), `readSerenityFlagScopes` is resolved before the null check on the write result, so the 404 path pays for one unnecessary DB round-trip. Acceptable given the "resolve before write" design intent and the rarity of the path. - `src/controllers/brands.js:2160`
- suggestion: `listBrandsForOrg` issues the flag-scopes query unconditionally even when `rows` is empty. An early return before the read would skip a round-trip for orgs with zero brands. - `src/controllers/brands.js:1125`
- suggestion: The "resolve before write" ordering invariant is load-bearing (documented three times in comments) but not mechanically enforced by tests. A `sinon.assert.callOrder(readSerenityFlagScopes, updateBrand)` assertion in the update test would lock it in against future refactors. - `test/controllers/brands.test.js`

</details>

- JSDoc on `readSerenityFlagScopes` now documents the bidirectional disagreement window between payload and gates.
- The `brandsOnly` test helper was removed; tests use direct stubs since `mapDbBrandToV2` no longer takes a `serenityScopes` parameter.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 4m 51s | <b>Cost:</b> $9.17 | <b>Commit:</b> <code>4a5f9c0241ab2d176734b05116c32de933c78b58</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
## Second round addressed, and the production schema gate is now cleared

Pushed as `refactor(brands): skip the flag read for an empty brand list, pin the write ordering`.

### The three non-blocking items

- **The list handler queried the flag rows even for an organization with no brands — fixed.** It now skips the read when there is nothing to describe. Declining this one would have been inconsistent: the whole previous round was about not paying for reads a path does not need.
- **The "resolve before write" ordering was documented but not enforced — fixed.** A test now asserts the rollout rows are read before `updateBrand`. That ordering is the entire reason a flag-read fault can no longer surface as a 500 on an edit that already committed, and it is exactly the kind of invariant a later refactor would quietly undo.
- **The 404 path on the write handlers pays one round-trip — declined.** The read has to precede the write and the null check can only follow it, so the two cannot be reordered without giving up the invariant above. The path itself means the brand disappeared between `resolveBrandUuid` and the write, in the same request.

### Production now carries the schema

`mysticat-data-service` https://github.com/adobe/mysticat-data-service/pull/896 merged and has been applied. Verified against the production database rather than inferred from the merge:

- `feature_flags.brand_id` — present, `uuid`, nullable.
- `feature_flags_org_product_flag_name_brand_key` — `UNIQUE NULLS NOT DISTINCT (organization_id, product, flag_name, brand_id)`.
- `feature_flags_organization_id_brand_id_fkey` — `FOREIGN KEY (organization_id, brand_id) REFERENCES brands(organization_id, id) ON DELETE CASCADE`.

That is the same shape as dev and stage, including the composite foreign key that stops an override naming another organization's brand. Sections 6 and 8 of the description reflect the current state.

So the schema gate on merging is gone. Deploying changes no observable production behaviour on its own: with no override rows written anywhere yet, every brand resolves to its organization's value, which is what the organization-wide gate answered before. The first behavioural change comes when the wave-execution CLI writes its first override row.

Still outstanding before this is ready: dev runtime validation, reported here.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - prior concerns addressed, per-brand resolution is clean and well-tested.
**Complexity: HIGH** - large diff (2139 lines, 18 files).
_Changes: Resolves the LLMO/serenity feature flag per brand instead of per organization, enabling wave-by-wave migration within a single org (18 files)._
Note: Recommend a human read before merge - this change modifies the OpenAPI schema (shared contract) and the conventions reviewer flagged spec divergence from issue 3020 (heuristic, both divergences are explicitly documented and justified in the PR body). The bot review is a complement to, not a replacement for, a human read here.

<details>

- suggestion: `authorizeOrg` passes `brandIdParam` directly to `isSerenityActiveForBrand` while `authorizeBrandSubWorkspace` passes the resolved `brandUuid` from `resolveBrandUuid` - aligning the variable names or adding a one-line comment would confirm to future readers that these are the same UUID format. - `src/controllers/elements.js:383` vs `:433`
- nit: `withSerenityState` could guard against a brand with no `id` field (it silently falls through to the org row via `Map.get(undefined)`), though all current callers have already resolved a valid brand from the DB. - `src/support/brands-storage.js:576`

</details>

- listBrandsForOrg skips the flag read when the brand list is empty.
- "Resolve before write" ordering invariant is now mechanically tested with a `calledBefore` assertion.
- JSDoc on readSerenityFlagScopes documents the bidirectional disagreement window between payload and gates.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 40s | <b>Cost:</b> $4.87 | <b>Commit:</b> <code>647e5f9fbbeb2f517182e53627abc82bdb1d4bf6</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #3025, opened 2026-08-10
fix(serenity): guard PATCH on the server-owned source dimension (LLMO-6665)

aliciadriani:
## What

The tag `PATCH` write guard (`buildUpdatePayload`, `src/support/serenity/handlers/tags.js`) keyed on the **closed-vocabulary** axis (`CLOSED_DIMENSIONS`: `intent`, `origin`, `type`) instead of the **ownership** axis, so it did not cover the `source` dimension introduced by spec 47. `source` is registered in `SERVER_OWNED_DIMENSIONS` but, deliberately, **not** in `CLOSED_DIMENSIONS` — its vocabulary is the open set of producing systems, not a frozen enum.

## Why it matters

A `PATCH` on any descendant of the `source` root passed every guard, so a client could **rename or re-parent a server-owned `source` value** even though the server writes it entirely. Every server write resolves a value by its bare name under the dimension root, so renaming `gsc` does not move it — it hides it, and the next prompt write resolve-or-creates a *second* `gsc` under the same root. Prompts then split across two ids for one producing system (the same reasoning the code already gives for refusing a rename of a closed value like `branded`).

Reachability: latent while no live project carried a `source` root; **reachable the moment the migration seeds the root on live projects** (WP-S5 / the LLMO-6802 reconciles) — which is exactly the state the WP-S6 source backfill ([LLMO-6286](https://jira.corp.adobe.com/browse/LLMO-6286)) depends on. This is a pre-backfill safety gate.

## Change

- `buildUpdatePayload`: refuse the edit when `isServerOwnedDimension(target.rootName)` holds (covers `intent`, `origin`, `type`, `source`); message `closed` → `server-owned`.
- The create path (`handleCreateTag`) already used the correct `isServerOwnedDimension` predicate — this brings the update path into line.
- Removed the now-unused `CLOSED_DIMENSIONS` import; repointed its `{@link}`s; rewrote the `buildUpdatePayload` doc-comment (the stale "four roots" / "A CLOSED dimension's value" language flagged in the ticket).
- Tests: updated the origin assertion to the new message; added 2 regression tests — a source-value **rename** and **re-parent**, both `400` with `server-owned "source" …` and the transport never called.

## Verification

- `npx mocha test/support/serenity/handlers/tags.test.js` — 76/76 pass (incl. the 2 new).
- `eslint` clean on both files.
- `npm run type-check` fails only on **pre-existing** `rest-transport.js` errors (spec-35 metadata swagger drift); identical with this change stashed — not introduced here.

Refs: [LLMO-6665](https://jira.corp.adobe.com/browse/LLMO-6665), epic [LLMO-6270](https://jira.corp.adobe.com/browse/LLMO-6270).

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```

### DIFF
--- a/src/support/serenity/handlers/tags.js
+++ b/src/support/serenity/handlers/tags.js
@@ -162,7 +162,7 @@ function parseUpdateParentId(body) {
  * The two flags answer two independent questions. `isClosed`
  * ({@link isClosedDimension}) drives VOCABULARY validation: a closed value's
  * `name` must be one of that dimension's fixed values ({@link closedValuesOf}).
- * `isServerOwned` ({@link SERVER_OWNED_DIMENSIONS}) drives the WRITE GUARD and
+ * `isServerOwned` ({@link isServerOwnedDimension}) drives the WRITE GUARD and
  * CREATE SEMANTICS: it forbids a `parentId` (server-owned values hang directly off
  * their root) and routes the create through resolve-or-create. `source` is
  * server-owned yet open — a `parentId` is refused and the value is resolved-or-
@@ -659,15 +659,14 @@ async function resolveUpdateTargets(
  * Three targets are refused. A DIMENSION ROOT is not editable — the root level is
  * reserved for the registered dimension roots ({@link ALL_DIMENSIONS}), and
  * renaming or moving one would leave its whole subtree without a dimension. A
- * SERVER-OWNED dimension's value ({@link SERVER_OWNED_DIMENSIONS}: `intent`,
- * `origin`, `type`, `source`) is not editable either: its vocabulary is authored by
- * the server, and since every resolve-or-create keys on the bare name under the
- * root, renaming a value — a closed `branded`, or a `source` value such as `gsc` —
- * would not move it, it would hide it, so the next server write mints a second value
+ * SERVER-OWNED dimension's value ({@link SERVER_OWNED_DIMENSIONS}) is not editable
+ * either: its vocabulary is authored by the server, and since every
+ * resolve-or-create keys on the bare name under the root, renaming a value would
+ * not move it, it would hide it — the next server write then mints a second value
  * under the same root and every prompt still carrying the first splits across two
- * ids for one dimension value. This is why the guard keys on the OWNERSHIP axis
- * ({@link isServerOwnedDimension}), not the closed/open vocabulary axis — `source`
- * is open yet server-owned, and was reachable here once its root seeded live
+ * ids for one dimension value. The guard therefore keys on the OWNERSHIP axis
+ * ({@link isServerOwnedDimension}), not the open/closed vocabulary axis: `source`
+ * is open yet server-owned, and became reachable here once its root seeded live
  * projects (LLMO-6665). An UNRESOLVABLE id is refused rather than forwarded: without
  * the target's current parent there is no body that preserves it, and guessing would
  * promote the tag.
--- a/src/support/serenity/handlers/tags.js
+++ b/src/support/serenity/handlers/tags.js
@@ -21,7 +21,7 @@ import {
 } from '../validation.js';
 import { resolveProject } from '../subworkspace-projects.js';
 import {
-  ALL_DIMENSIONS, CLOSED_DIMENSIONS, SERVER_OWNED_DIMENSIONS,
+  ALL_DIMENSIONS, SERVER_OWNED_DIMENSIONS,
   isClosedDimension, isServerOwnedDimension, closedValuesOf, isDimensionRootName,
   MAX_TAG_NAME_LEN,
 } from '../prompt-tags.js';
@@ -160,7 +160,7 @@ function parseUpdateParentId(body) {
  * direct children of its root.
  *
  * The two flags answer two independent questions. `isClosed`
- * ({@link CLOSED_DIMENSIONS}) drives VOCABULARY validation: a closed value's
+ * ({@link isClosedDimension}) drives VOCABULARY validation: a closed value's
  * `name` must be one of that dimension's fixed values ({@link closedValuesOf}).
  * `isServerOwned` ({@link SERVER_OWNED_DIMENSIONS}) drives the WRITE GUARD and
  * CREATE SEMANTICS: it forbids a `parentId` (server-owned values hang directly off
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - well-scoped bug fix with correct predicate swap and solid regression tests.
**Complexity: MEDIUM** - small diff; API surface handler (write guard).
_Changes: Fixes the PATCH write guard in `buildUpdatePayload` to key on the server-owned dimension axis (`isServerOwnedDimension`) instead of the closed-vocabulary axis, closing a gap where `source` values were unprotected (2 files)._

<details>

- nit: JSDoc link at line ~167 references `{@link SERVER_OWNED_DIMENSIONS}` (the constant) where `{@link isServerOwnedDimension}` (the predicate) would be consistent with the `{@link isClosedDimension}` fix two lines above - `src/support/serenity/handlers/tags.js:167`
- nit: Comment at line ~668 hardcodes "intent, origin, type, source" - if a fifth server-owned dimension is added, this list goes stale while the runtime guard remains correct - `src/support/serenity/handlers/tags.js:668`
- suggestion: `makeSourceDescendableTransport` hardcodes `children_count: 2` which happens to match the fixture but is not derived from it - a brief comment or `levels[TAG_IDS.sourceRoot]?.length` would clarify the intent - `test/support/serenity/handlers/tags.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 4s | <b>Cost:</b> $3.90 | <b>Commit:</b> <code>9e587348391a4eb79d2a7f4ef6b11feedbb4810b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/serenity/handlers/tags.js
```diff
@@ -160,9 +160,9 @@ function parseUpdateParentId(body) {
  * direct children of its root.
  *
  * The two flags answer two independent questions. `isClosed`
- * ({@link CLOSED_DIMENSIONS}) drives VOCABULARY validation: a closed value's
+ * ({@link isClosedDimension}) drives VOCABULARY validation: a closed value's
  * `name` must be one of that dimension's fixed values ({@link closedValuesOf}).
- * `isServerOwned` ({@link SERVER_OWNED_DIMENSIONS}) drives the WRITE GUARD and
+ * `isServerOwned` ({@link isServerOwnedDimension}) drives the WRITE GUARD and
```
aliciadriani, line 165: **Fix — MysticatBot nit:** repointed this `{@link}` from the constant `SERVER_OWNED_DIMENSIONS` to the predicate `{@link isServerOwnedDimension}`, matching the `{@link isClosedDimension}` link two lines above.

src/support/serenity/handlers/tags.js
```diff
@@ -657,13 +657,19 @@ async function resolveUpdateTargets(
...
- * reserved for the four roots, and renaming or moving one would leave its whole
- * subtree without a dimension. A CLOSED dimension's value is not editable either:
- * the vocabulary is fixed, and since every resolve-or-create keys on the bare name
- * under the root, renaming `branded` would make the next prompt write mint a
- * second `branded` and silently orphan every prompt still carrying the first. An
- * UNRESOLVABLE id is refused rather than forwarded: without the target's current
- * parent there is no body that preserves it, and guessing would promote the tag.
+ * reserved for the registered dimension roots ({@link ALL_DIMENSIONS}), and
+ * renaming or moving one would leave its whole subtree without a dimension. A
+ * SERVER-OWNED dimension's value ({@link SERVER_OWNED_DIMENSIONS}) is not editable
```
aliciadriani, line 662: **Fix — MysticatBot nit + self-review nit:** dropped the hardcoded `intent, origin, type, source` list that used to be here — it would go stale the moment a fifth server-owned dimension is added, while the runtime guard stayed correct. The `{@link SERVER_OWNED_DIMENSIONS}` reference now carries the authoritative set, and the comment is trimmed a couple of sentences (self-review trim nit).

src/support/serenity/handlers/tags.js
```diff
@@ -684,11 +690,9 @@ function buildUpdatePayload(parsed, target, tagId) {
     err.code = ERROR_CODES.TAG_NOT_FOUND;
     throw err;
   }
-  if ((/** @type {readonly string[]} */ (CLOSED_DIMENSIONS)).includes(
-    /** @type {string} */ (target.rootName),
-  )) {
+  if (isServerOwnedDimension(/** @type {string} */ (target.rootName))) {
```
aliciadriani, line 693: **Core fix (LLMO-6665):** the guard now keys on the ownership axis `isServerOwnedDimension(target.rootName)` (intent/origin/type/**source**) instead of `CLOSED_DIMENSIONS`, so a client can no longer rename or re-parent a server-owned `source` value. Confirmed this does not affect the WP-S6 backfill, which merges tag references via `updatePromptTagsByIds` — a different transport method from the guarded `updateProjectTag`.

test/support/serenity/handlers/tags.test.js
```diff
@@ -57,6 +57,27 @@ function makeEmptyTreeTransport(overrides = {}) {
...
+// this bumps it so the walk descends and resolves a `source` value — the state
+// reachable after WP-S5 / the reconciles seed the root on live projects, and the
+// only state in which the LLMO-6665 guard is exercised. Shared by the flat-mode
+// and subworkspace update describes.
+function makeSourceDescendableTransport(overrides = {}) {
+  const levels = dimensionTreeLevels();
+  // Derive the count from the fixture's own source level rather than hardcoding a
+  // literal, so it can't drift if the fixture gains/loses a seeded `source` value
+  // (mysticatbot suggestion).
+  const sourceChildCount = (levels[TAG_IDS.sourceRoot] ?? []).length;
```
aliciadriani, line 71: **Fix — MysticatBot suggestion:** derive the source-root `children_count` from the fixture's own source level (`levels[TAG_IDS.sourceRoot]?.length`) instead of the hardcoded `2`, so the test can't silently drift if the fixture gains or loses a seeded `source` value.

test/support/serenity/handlers/tags.test.js
```diff
@@ -1562,6 +1633,25 @@ describe('serenity tags handler (POST /serenity/tags)', () => {
       expect(transport.updateProjectTag).to.not.have.been.called;
     });
 
+    // Defense-in-depth for LLMO-6665: the source guard lives in the shared
+    // buildUpdatePayload, so it must refuse a server-owned `source` rename on the
+    // subworkspace route too, not only in flat mode (self-review should-fix).
+    it('400s a rename of a source value — shared server-owned guard (LLMO-6665)', async () => {
```
aliciadriani, line 1639: **Fix — self-review should-fix:** added a subworkspace-mode source-rename regression test. The source guard lives in the shared `buildUpdatePayload`, so both the flat and subworkspace routes must refuse the edit — this proves the subworkspace twin does. The `makeSourceDescendableTransport` helper was hoisted to module scope so both describes reuse it.

aliciadriani (commented):
Addressed all review feedback (self-review + @MysticatBot) in 636c3d4. Inline comments below map each change to its source item. Full serenity suite green (77 tags tests, incl. the new subworkspace case), lint clean. Re-requesting review.

test/support/serenity/handlers/tags.test.js
```diff
@@ -1562,6 +1633,49 @@ describe('serenity tags handler (POST /serenity/tags)', () => {
...
+
+      expect(err.status).to.equal(400);
+      expect(err.message).to.match(/server-owned "source" dimension cannot be renamed or re-parented/);
+      expect(transport.updateProjectTag).to.not.have.been.called;
+    });
+
+    // Completes the flat/subworkspace symmetry for LLMO-6665: the server-owned
+    // guard fires before the placement check in buildUpdatePayload, so a source
+    // RE-PARENT is refused on the subworkspace route too, same as flat mode.
+    it('400s a re-parent of a source value — shared server-owned guard (LLMO-6665)', async () => {
```
aliciadriani, line 1658: **Fix — re-review nit:** completes the flat/subworkspace symmetry for LLMO-6665. The server-owned guard in the shared `buildUpdatePayload` fires before the placement check, so a source **re-parent** is refused on the subworkspace route too — both routes now assert refusal for both a rename and a re-parent.

aliciadriani (commented):
Addressed the re-review nit in 789413c3: added the subworkspace **re-parent** source case so both routes now cover both rename and re-parent of a server-owned `source` value. Full tags suite green (78), lint clean.

PR #3027, opened 2026-08-11
fix(llmo): derive Cloudflare worker targetHost from site base URL

ssilare-adobe:
## Problem

The Cloudflare BYOCDN onboarding `deployWorker` handler (`llmo-cloudflare.js`) took `targetHost` from the client request body and validated it against `site.getBaseURL()`. In practice the site's `baseURL` is the apex address, while the host that actually serves traffic (and where the `x-forwarded-host` / route resolves) is often the `www` host. For sites like `bhhsblakerealtors` — apex `baseURL`, real serving host `www.<domain>` — a client-supplied apex `targetHost` passes validation but sets the worker's `EDGE_OPTIMIZE_TARGET_HOST` binding to the wrong host.

Per the discussion with Nitin and Akash, host derivation should be consistent across CDNs. CloudFront (`llmo-cloudfront.js`) already derives its host via `calculateForwardedHost(site.getBaseURL())` from `@adobe/spacecat-shared-tokowaka-client`; Cloudflare should do the same instead of trusting a client value.

## Change

- `deployWorker` no longer accepts `targetHost` from the request body. It is now **derived server-side** from the site's base URL via `calculateForwardedHost(site.getBaseURL())`, so the worker always forwards to the canonical host for the site (a bare apex like `example.com` is normalized to `www.example.com`; an existing subdomain such as `cdn.example.com` is preserved).
- Request body is now just `{ accountId }`. The `targetHost` hostname/`hostInSiteDomain` boundary checks are removed (the derived value comes from the site's own base URL); `hostInSiteDomain` is still used for route-pattern validation.
- Derivation is wrapped in a defensive `try/catch` (mirroring the CloudFront sibling) that returns a 500 if the base URL can't be resolved.
- OpenAPI spec (`llmo-api.yaml`) updated: `targetHost` dropped from the request body/required list and documented as server-derived; the `200` response still returns the resolved `targetHost`.

### Known follow-up
Sites served from the apex **without** a `www` DNS record still normalize to `www` here — flagged in the thread as a separate enhancement (`calculateForwardedHost` + DNS/redirect follow), tracked as follow-up work, not addressed in this PR.

## Test Plan

- [x] `npm run lint` (changed files) — clean
- [x] Unit: `test/controllers/llmo/llmo-cloudflare.test.js` — 70 passing (added apex→www derivation + subdomain-preservation cases; removed client-supplied `targetHost` validation cases)
- [x] Full llmo controller suite — 2184 passing
- [x] `npm run docs:lint` — OpenAPI valid
- [x] IT shared spec (`llmo-cloudflare-onboarding.js`) updated to drop `targetHost` body validation cases

## Related Issues

_Design discussion: Cloudflare BYOCDN onboarding `targetHost` derivation (Slack thread with @nitigupt, @akbhardwaj)._

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Nitin Gupta", "Dominique Jäggi"]
```


nit23uec (approved):

docs/openapi/llmo-api.yaml
```diff
@@ -5844,20 +5845,12 @@ site-llmo-cloudflare-deploy:
...
             required:
               - accountId
-              - targetHost
             properties:
               accountId:
                 type: string
                 description: Cloudflare account ID (32-character lowercase hex)
                 pattern: '^[0-9a-f]{32}$'
                 example: '0123456789abcdef0123456789abcdef'
-              targetHost:
```
nit23uec, line 5854: can you pls check this for other cdn automations like akamai and cloudfront?

ssilare-adobe: Checked both — Cloudflare was the only outlier here; the other two are already consistent:

**CloudFront** — never took a client `targetHost`. It already derives the origin forwarding host server-side with the same helper (`calculateForwardedHost(baseURL)`, e.g. `llmo-cloudfront.js:515`). So this change actually brings Cloudflare *in line with* CloudFront.

**Akamai** — different integration model: it doesn't set an origin-forwarding binding on an edge worker, it uses the site hostname to find/validate the Akamai property that serves the domain (`findPropertiesByDomain`). The host is still derived server-side (`normalizeDomain(baseURL)` via `siteHostname()`), not client-supplied — so there's no client value to drop. It deliberately uses `normalizeDomain` (literal host) rather than `calculateForwardedHost` on purpose: the apex→www rewrite would break property matching. This distinction is documented in `llmo-utils.js` (the "do not use `calculateForwardedHost()` for an audit host" note).

Net: CloudFront already consistent, Akamai intentionally different-by-design, Cloudflare now aligned. No changes needed on the other two.

MysticatBot (approved):
Hey @ssilare-adobe,

**Verdict: Approve** - well-scoped security fix that aligns Cloudflare host derivation with the established CloudFront pattern.
**Complexity: MEDIUM** - API contract change (OpenAPI schema modification).
_Changes: Removes client-supplied `targetHost` from the Cloudflare deploy endpoint and derives it server-side from the site base URL via `calculateForwardedHost`, consistent with CloudFront (4 files)._
Note: Recommend a human read before merge - this change modifies a shared API contract (OpenAPI schema). The bot review is a complement to, not a replacement for, a human read here.
Note: CI checks are currently failing on codecov/patch (77.27% patch coverage, 5 lines missing).

<details>

- suggestion: Add a test that stubs `calculateForwardedHost` to throw, covering the try/catch error path (the 5 uncovered lines Codecov flags). Validates the structured log and 500 response. - `test/controllers/llmo/llmo-cloudflare.test.js`
- suggestion: Log at INFO when the request body contains a now-ignored `targetHost` field that differs from the derived value (e.g. `log.info("Ignoring client-supplied targetHost; derived: %s", targetHost)`). Helps operators detect mismatches during the transition period. - `src/controllers/llmo/llmo-cloudflare.js:355`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 18s | <b>Cost:</b> $6.77 | <b>Commit:</b> <code>7abca3fa50a7132ed14146784170012f13193029</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #3028, opened 2026-08-11
fix(serenity): classify branded prompts in scripts without word spaces

rainer-friederich:
## 1. Abstract

Makes the Serenity branded / non-branded prompt classifier work in scripts that do not put spaces between words, and adds an operator-run sweep that corrects the `type` tag on prompts already written.

## 2. Reasoning

KDDI reported that the Brand Presence "Tags > type > non-branded" filter returns every prompt rather than the non-branded subset (LLMO-6900). It does, and the filter is innocent: all 762 prompts in their `au` project carry `type__non-branded` and none carries `type__branded`, so filtering on non-branded correctly returns everything.

The classifier is what is wrong. It implements "whole word" as a padded-space token test — normalize both sides, then look for the needle flanked by spaces. Japanese writes no spaces between words, so `auのSIMロック解除` normalizes to a single token and the brand name inside it can never be flanked. 73 of those 762 prompts name the brand and every one of them classified non-branded.

This is not specific to a two-character brand name. Any brand whose prompts are written in Japanese, Chinese or Thai has the same outcome. Measured across production: 30 projects across 25 brands, 61 prompts branded today where 590 should be. Several of the affected brands are Latin-named companies running Japanese markets — Okta, Auth0, PwC Consulting, Toyota Gazoo Racing — so the defect is not confined to CJK-named customers.

A second defect sits underneath the first and had to be fixed with it. Normalization strips `\p{Diacritic}`, and Unicode marks the kana voicing marks and the prolonged sound mark as diacritics even though in Japanese they distinguish one kana from another. NFD decomposes `パ` to `ハ` plus a combining mark, so the strip turns `パナソニック` into `ハナソニック` and shortens `ソニー` to `ソニ`. Harmless while nothing matched at all; the moment the boundary is fixed it would make the needle `ソニー` match Panasonic.

**The matching rule.** Normalization now gives every character of a script that does not delimit words with spaces — Han, Hiragana, Katakana, Thai, Lao, Khmer, Myanmar — its own token. That restores a boundary on both sides of a brand name embedded in running text, and lets a brand named in the script itself (`楽天`) match as a contiguous character run inside `楽天モバイル`. Hangul and Vietnamese are deliberately excluded: Korean and Vietnamese do use word spaces, so `삼성` must keep matching `삼성 전자` and not `삼성전자`.

Two supporting changes make that safe. The kana voicing and prolongation marks are exempted from the diacritic strip, and the result is recomposed with NFC before the punctuation collapse — without the recompose, the surviving combining mark is category `Mn`, which is neither a letter nor a number, so the collapse would turn it into a token gap and undo the exemption.

Before and after, for KDDI's `au`: 0 of 762 prompts branded, becoming 73 of 762. That matches the count the pre-Semrush dashboard shows for the same brand, which is where the customer's expectation comes from.

For text containing no space-less-script characters the new rule is a strict superset of the old one — the only change is inserting token gaps around characters that previously had none, which cannot remove an existing match. No Latin-market brand changes.

**Correcting prompts already written.** Classification is forward-only by design: no write path revisits a prompt it already wrote, so the rule change alone leaves every existing prompt wrong. `scripts/serenity-retype-backfill.mjs` re-derives the value for one brand's prompts and rewrites the `type` tag in place, preserving prompt ids and the Brand Presence history keyed to them. It classifies through the same closure the write paths use rather than a second copy of the rule, which is why `buildPromptTypeClassifier` is now exported.

Operationally visible properties: it is idempotent (a prompt already carrying the right tag is skipped, so a re-run after an interruption is cheap and safe), checkpointed per project and only after publish succeeds (an unpublished patch is invisible, so a patch-then-publish-failure is retried rather than recorded as done), and `--dry-run` is genuinely read-only — it resolves tag ids through plain level listings and refuses rather than creating a missing dimension value.

It cannot be scheduled, and that is a property of the platform rather than a shortcut. Every Semrush call needs a live user's IMS bearer; `ImsPromiseClient` offers only a grant that itself requires a user access token, with no service-identity path; and the non-IMS escape hatch is hard-disabled in production. So the sweep is operator-run, per brand.

## 4. Required information

- Jira / issue: LLMO-6900 — https://jira.corp.adobe.com/browse/LLMO-6900
- Spec: serenity-docs `docs/prompt-management/branded-classification.md` — the design of record for this dimension, already merged

- **mysticat-data-service** — changed in a companion PR. It carries two more implementations of the same rule (the migration CLI's Python classifier, which is what actually seeded KDDI, and the `rpc_brand_prompt_stats` SQL regex). Both had the identical blind spot and are fixed there.
- **serenity-docs** — updated and merged separately. Holds the design of record and the normative cross-implementation conformance table this repo's tests assert against.
- **project-elmo-ui** — consumed, unchanged. Its Semrush-powered Brand Presence surface is where the filter misbehaves, and its prompt-library Branded column and prompt-health tile read the same tag. All of them correct themselves once the tag is right; no UI change is needed. Its classic dashboard computes branded client-side by substring and is deliberately untouched.

Everything below was measured against production this session, not estimated.

- The Elements prompts element for KDDI's workspace returns 762 rows unfiltered, 762 for `tags contains type__non-branded`, and 0 for `type__branded` — confirming the filter is correct and the data uniform. Filtering on the bare `type` root, and AND/OR combinations of root plus child, all return 762, which rules out the tag-selection shape as a cause.
- Running the edited classifier over the 762 live prompt texts, with the brand's real region-clamped aliases, returns 73 branded.
- A harness exercising the same library calls the sweep makes — read-only tag resolution, `listPromptsByTags` pagination, and patch-set construction — against the real workspace: both `type` values already exist so nothing would be created, pagination returns exactly 762, and the sweep would rewrite exactly 73 prompts, refusing none.
- Fleet sizing came from `brand_to_semrush_projects` joined to `brands` on the production reader, then per-project prompt reads: 30 projects, 25 brands, 61 branded now against 590 predicted.

## 7. Test plan

Verified locally beyond the unit suites: the edited classifier reproduces the customer's expected split on their real prompt corpus (73 branded / 689 non-branded), and the sweep's live path was exercised end to end against the real Semrush workspace short of the write itself.

Not covered by any test, and stated so a reviewer does not assume otherwise:

- **Full-width Latin brand names and aliases are matched by none of the three implementations.** Normalization uses NFD, which does not fold full-width forms — only NFKD would. KDDI has an `ａｕ` alias, so that alias is inert. Their prompt corpus contains zero full-width occurrences, so this does not affect the sweep, but the alias does nothing. Not fixed here: NFKD would also change Latin behaviour (ligatures, superscripts) and is a separate decision from this defect.
- **The three implementations agree on the conformance table, not universally.** This repo tokenizes only the space-less scripts; the Python and SQL mirrors use a Latin-alphanumeric lookaround, which treats every non-Latin character as a boundary. So `삼성` matches inside the unspaced `삼성전자는` under the mirrors but not here, and Cyrillic and Greek behave the same way. The practical consequence is that a brand with a Hangul, Cyrillic or Greek name seeded through the migration CLI can carry a `type` tag that permanently disagrees with what the live write path would compute.
- **The conformance table is hand-synced across four places and already drifted once during this change** — the SQL copy was written incomplete and was caught in review. A machine-readable shared fixture was considered and declined: three repos on three release trains with no shared package means a vendored copy goes stale just as easily, and the drift would be less visible in review than a table a reviewer can read.
- **The sweep has no unit tests**, matching the four existing `scripts/*.mjs` precedents, which sit outside the coverage configuration by convention. Its risk is controlled by dry-run, per-project checkpointing after publish only, and idempotency, and it delegates every behaviour-bearing step to already-tested library code.
- Three affected brands stay at zero branded even after the fix — KDDI Biz, UPS Forwarding Hub and The Silk Lakehouse — because their brand names never appear verbatim in their prompts (`kddi biz`, `ups forwarding hub`). That is missing alias data for the brand owners to supply, not something a classifier change can reach.

Verifying on an environment:

- **dev / stage**: create a prompt whose text names the brand immediately followed by kana (no space) and confirm it is stored `branded`; create one that never names the brand and confirm `non-branded`. Then confirm a Latin brand is unaffected — a prompt containing the brand inside a longer word must stay `non-branded`.
- **prod**: after deploy, run the sweep for one brand with `--dry-run` first and check the reported before/after against the predicted numbers; for KDDI that is 0 → 73 of 762. Only then run it for real, and re-read the Elements prompts element with `tags contains type__branded` to confirm the count landed.

- adobe/mysticat-data-service, same branch name — related, independent. Fixes the Python and SQL mirrors of the same rule. Separate release train; no ordering constraint against this PR.
- adobe/serenity-docs — the design of record this PR cites; merged already, so nothing here waits on it.

Ordering that does matter: this PR must be merged **and deployed** before the sweep is run against a brand. The sweep runs from a checkout and therefore always uses the corrected rule, but the live write path does not until the Lambda ships — so a sweep run against a stale deployment leaves a window in which newly created prompts are classified wrong again.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-scoped fix with thorough conformance tests and a carefully designed backfill script.
**Complexity: HIGH** - large diff (511-line new script + classifier changes); API surface signal.
_Changes: Fixes branded/non-branded prompt classifier for scripts without word spaces (Japanese, Chinese, Thai, Lao, Khmer, Myanmar) by tokenizing space-less script characters individually, preserves kana voicing marks during diacritic folding, and adds an operator-run backfill script to correct existing prompts (5 files)._

<details>

- nit: `exit(totals.failed > 0 ? 1 : 0)` exits non-zero even when all failures are expected 403s (authorization scope). An operator running against a fleet where half the projects 403 gets exit code 1 despite a successful sweep. Consider distinguishing expected vs unexpected in the exit code, or documenting that exit 1 does not imply an actionable failure. - `scripts/serenity-retype-backfill.mjs:529`
- suggestion: `--brand-id` is passed to `dataAccess.Brand.findById()` without UUID format validation. A typo produces an unhelpful PostgREST error rather than a clear CLI message. A fast regex check at parse time would improve the operator experience. - `scripts/serenity-retype-backfill.mjs:178`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 20s | <b>Cost:</b> $9.20 | <b>Commit:</b> <code>0125ac62e8eaa9a8f29938b50fa0b951526a1d4b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Both items addressed.

- **Exit code treats expected 403s as failures.** Valid, and it contradicted the script's own model: a 403 is documented as an expected per-project outcome and is already excluded from the consecutive-error abort, yet it still incremented `failed`, which is what drives the exit code. Authorization skips are now counted separately and reported in the summary, with an explicit warning that those workspaces were not swept — an operator does need to see that the run was partial. The exit code is left to mean "something actionable went wrong", and the exit status is documented in the header.
- **`--brand-id` UUID validation.** Added at parse time, matching the existing `parseNumericOption` error style. A malformed id now prints `--brand-id must be a UUID, got "…"` instead of reaching PostgREST.

Verification: eslint and both type-check tiers clean, 1366 serenity tests pass, and the new guard was exercised both ways — a malformed id is rejected, a well-formed one falls through to the environment checks unchanged.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-scoped fix with thorough conformance tests; prior non-blocking items addressed.
**Complexity: HIGH** - large diff (725 lines, 5 files); API surface signal.
_Changes: Fixes branded/non-branded prompt classifier for scripts without word spaces (Japanese, Chinese, Thai, Lao, Khmer, Myanmar) by tokenizing space-less script characters individually, preserves kana voicing marks during diacritic folding, and adds an operator-run backfill script to correct existing prompts (5 files)._

<details>

- nit: When `unresolvableTags` is non-zero, the aggregate count is reported in the summary but individual prompt ids with malformed tags are not logged anywhere. An operator investigating why the count is non-zero has no trail. A `log.debug` with the prompt id in the `unresolvableTags` branch of `sweepProject` would help. - `scripts/serenity-retype-backfill.mjs:375`
</details>

- Exit code no longer treats expected 403s as failures - `authSkipped` counted separately from `failed`
- `--brand-id` UUID format validation confirmed present with explicit error message and early exit

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 2s | <b>Cost:</b> $7.45 | <b>Commit:</b> <code>a1e721ed3e7ce29912eed47fed8d0322ae76ce25</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #3029, opened 2026-08-11
feat(ai-seo): AISEODT-134 added new countries

dmitriigulkin-semrush:
## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```


docs/openapi/schemas.yaml
```diff
@@ -10171,6 +10171,81 @@ AiVisibilitySemrushCountry:
     - TR
     - TW
     - VN
+    - NZ
```
calvarezg, line 10174: issue (blocking): [Important] OpenAPI country enum is not guarded against drift from COUNTRY_ENUM.

This PR makes the runtime list `SR_AI_SEO_SUPPORTED_MARKET_CODES` auto-derive from the vendored `COUNTRY_ENUM` (`src/support/ai-visibility/visibility-filters.js`), but `AiVisibilitySemrushCountry` here is still hand-maintained. The two match exactly today, but nothing keeps them in sync: `docs/openapi/schemas.yaml` is authored by hand (the only build step, `docs:build`, renders `docs/index.html` from this file - it does not generate this file), and no test asserts the two enums are equal.

Why it matters: on the next vendor bump the runtime list updates automatically while this enum stays stale, and the drift is silent. `normalizeMarketToken` coerces any unknown token to `US`, so a country removed from the enum but left here silently maps to US, and one added to the enum but missing here works but is undocumented.

How to fix: add a test that parses the two-letter values of `AiVisibilitySemrushCountry` from this file and asserts set-equality with the `COUNTRY_ENUM`-derived codes (compare as sets - this list is in protobuf declaration order, the runtime list is sorted). `js-yaml` is already a dependency.

Verification: Confirmed - schemas.yaml is hand-maintained (only `docs:build` consumes it), no sync test exists, and unknown tokens coerce to US in `normalizeMarketToken`.

test/support/ai-visibility/visibility-filters.test.js
```diff
@@ -30,11 +31,12 @@ describe('visibility-filters', () => {
...
-    it('contains expected market codes', () => {
-      expect(SR_AI_SEO_SUPPORTED_MARKET_CODES).to.include.members([
-        'US', 'UK', 'DE', 'FR', 'JP',
-        'HK', 'ID', 'KR', 'MY', 'PH', 'SG', 'TH', 'TR', 'TW', 'VN',
-      ]);
+    it('stays synchronized with the vendored COUNTRY enum', () => {
+      const countryEnumCodes = Object.keys(COUNTRY_ENUM)
+        .filter((code) => /^[A-Z]{2}$/.test(code))
+        .sort((a, b) => a.localeCompare(b));
+      expect(SR_AI_SEO_SUPPORTED_MARKET_CODES).to.deep.equal(countryEnumCodes);
```
calvarezg, line 38: issue (blocking): [Important] This sync test is blind to protobuf-side enum regressions.

`countryEnumCodes` is computed with the byte-for-byte same expression as the source in `src/support/ai-visibility/visibility-filters.js` (`Object.keys(COUNTRY_ENUM).filter(/^[A-Z]{2}$/).sort(localeCompare)`), then compared with `deep.equal`. Because both sides re-derive from the same import, the assertion cannot fail when the vendored enum itself changes: a future regen that drops or renames a country (e.g. `UK` -> `GB`) passes green here while `normalizeMarketToken('UK')` would start returning `US`.

This block replaced `it('contains expected market codes')`, which asserted `include.members` of 15 hard-coded codes (including `UK`) - an independent oracle. The new test keeps only `include.members(['NZ', 'AF'])`, so independent coverage dropped from 15 codes to 2.

How to fix: restore an independent hard-coded expected subset (the 15 prior codes incl. `UK`, plus a couple of new ones like `NZ`/`AF`) via `include.members`, and/or assert a concrete `.length`, so a protobuf-side drop actually fails. The `deep.equal` can stay as a secondary check - it still catches someone re-hardcoding the source array.

Verification: Confirmed, with one nuance - the `deep.equal` is not literally "never fails" (it catches a re-hardcoded source array), but it is blind to enum regressions and the independent oracle shrank from 15 codes to 2.

calvarezg (changes requested):
Hey @dmitriigulkin-semrush,

**Verdict: Request changes** - the code is correct and in sync at this SHA; two country-list sync guards are missing or ineffective and are worth closing before merge.
**Complexity: HIGH** - mechanical label: 566-line diff dominated by the regenerated protobuf vendor bundle, a dependency bump, and OpenAPI enum additions; the hand-authored surface is ~90 lines.
_Changes: expands the Semrush AI-SEO supported-country list to the full vendored COUNTRY_ENUM (~117 codes) and mirrors the additions into the OpenAPI enum (15 files)._
Note: Recommend a human read before merge - this PR changes a shared API contract (the OpenAPI schema). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Important] OpenAPI country enum has no drift guard against COUNTRY_ENUM - `docs/openapi/schemas.yaml:10174` (details inline)
2. [Important] The "stays synchronized" test is blind to protobuf-side enum regressions - `test/support/ai-visibility/visibility-filters.test.js:38` (details inline)

<details>

- nit: the PR body is empty; the template asks for intent, the ticket link, and a request/response example for API-spec changes. Not enforced, but this PR changes the public schema. - PR description
- nit: confirm the vendored `third-party/ai-seo-ts` files came from a clean toolchain regen, not hand-edits - the descriptor is one opaque base64 blob. - `third-party/ai-seo-ts/common/types_pb.js:13`
- nit: no documented update process for the vendored `third-party/ai-seo-ts` bundle (no regen script, README, or ADR). Worth a short note since this tree updates repeatedly. - `third-party/ai-seo-ts/`
- suggestion: state the invariant that every 2-letter uppercase COUNTRY_ENUM member is a country, so a future non-country 2-letter member cannot silently leak into the supported list. - `src/support/ai-visibility/visibility-filters.js:17`
</details>

- Generated `v2/topic/enums_pb` now documents `RELEVANCY` order-by as silently reinterpreted as `VISIBILITY` (AISEO-2171). No hand-authored caller here passes `RELEVANCY`; confirm no consumer relies on the old semantics.
- Pre-existing: the OpenAPI `AiVisibilitySemrushCountry` enum lists `UNKNOWN` and `WORLDWIDE`, which the runtime `/^[A-Z]{2}$/` filter excludes. Not introduced here; worth a follow-up.

## Discarded in verification
- [was Important] "Empty PR body violates CONTRIBUTING.md / the PR template" - downgraded to a nit. Both docs use suggestion language ("please follow", "please ensure") with no mandatory wording and no CI enforcement; the API-spec item is an unchecked soft checkbox. Empty body is real, but it is a nit, not a rule violation.
- [was Important] "Deriving from every enum member advertises not-yet-live markets (is_coming_soon)" - false positive. `is_coming_soon` lives on a different schema, `AiVisibilityV1MetaCountry` (`schemas.yaml:10850`), populated at runtime from the Semrush Meta endpoint in `handlers/v1/meta/meta.js`; nothing links it to `COUNTRY_ENUM`. `COUNTRY_ENUM` is the vendored "countries supported by the API" contract, so deriving the full set is correct by its own definition. Which members are "coming soon" is knowable only from Semrush's runtime response, not statically.

calvarezg (approved):
Hey @dmitriigulkin-semrush,

**Verdict: Approve** - both previously flagged items are resolved; nothing blocking.
**Complexity: HIGH** - mechanical label unchanged: diff dominated by the regenerated protobuf vendor bundle; hand-authored surface is small.

- OpenAPI country enum drift guard added - `test/openapi-contract/ai-visibility-api.test.js` parses `AiVisibilitySemrushCountry` from `schemas.yaml` and asserts equality with the `COUNTRY_ENUM`-derived codes, so the doc enum and the vendored enum can no longer drift silently.
- Sync test no longer tautological - `visibility-filters.test.js` restored an independent hard-coded oracle (the original 15 codes incl. `UK`, plus `NZ`/`AF`) alongside the `deep.equal`, so a protobuf-side regen that drops a country now fails the test.

PR #3040, opened 2026-08-13
fix: Rename brands to '{name}_deleted' on delete to free the name for reuse | LLMO-6978

igor-grubic:
## What & why

Soft-deleting a brand kept its original name, so the per-org unique constraint `uq_brand_name_per_org` (which spans deleted rows) kept the name "taken" and blocked recreating a brand with that name.

This rewrites `deleteBrand()` in `src/support/brands-storage.js` to read the brand's name + status and rename any non-deleted brand to `{name}_deleted` in the same UPDATE that sets `status='deleted'`, appending an incrementing index (`_deleted2`, `_deleted3`, …) on a `23505` name collision (retried, race-safe). An already-deleted brand short-circuits to an idempotent 204 without re-renaming (no double-suffix); a NULL status is treated as live via a JS check rather than a SQL filter (which would exclude NULL rows).

The change is contained to `spacecat-api-service`; the shared Brand model is a read/patch surface not in the delete path, so no `spacecat-shared` change was needed. Backfilling pre-existing deleted brands is a separate workstream and out of scope.

## Tests

- Unit (`test/support/brands-storage.test.js`): rename, NULL-status, already-deleted idempotency, `_deleted2` collision index, not-found, read error, non-collision error, retry-exhaustion.
- Integration (`test/it/shared/tests/brands.js`): create → delete → recreate-same-name → delete-again (`_deleted2`) → recreate, asserting freed names against the real constraint. (Runs in CI — Postgres/Docker unavailable locally.)
- Local: eslint clean; `mocha test/support/brands-storage.test.js` (217 passing); `mocha test/controllers/brands.test.js -g deleteBrandForOrg` (10 passing); `npm run type-check` base+strict exit 0; c8 on `brands-storage.js` = 100% stmts/lines, 98.42% branches.

## Links

- Jira: https://jira.corp.adobe.com/browse/LLMO-6978
- Slack thread: https://cq-dev.slack.com/archives/C0BEJH1SNNL/p1786548369491819

---
🤖 Draft opened by the slack-pipeline. Auto-reviewed at implement time; CI + reviewer handled next.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```

### DIFF
--- a/src/support/brands-storage.js
+++ b/src/support/brands-storage.js
@@ -33,6 +33,13 @@ const BRAND_SELECT = [
   'brand_urls(url)',
 ].join(', ');
 
+// LLMO-6978: on delete a brand is renamed to `{name}_deleted` to free its
+// original name for reuse (see deleteBrand). A same-named deleted brand already
+// present bumps the suffix to `_deleted2`, `_deleted3`, ... Each colliding
+// rename costs one round-trip, so this caps a pathological loop (a customer
+// deleting the same name hundreds of times) rather than any expected volume.
+const MAX_DELETED_NAME_ATTEMPTS = 100;
+
 // Re-landed from Igor Grubic's #2504 (LLMO-5183): map the data-layer
 // chk_active_brand_has_site_id CheckViolation to a typed 400 (covers the race
 // where site_id is cleared between our SELECT and this write).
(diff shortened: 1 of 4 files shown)


MysticatBot (changes requested):
Hey @igor-grubic,

**Verdict: Request changes** - one concurrency concern in the rename logic.
**Complexity: MEDIUM** - medium diff (332 lines, 3 files), test-heavy.
_Changes: Rewrites `deleteBrand()` to rename the brand to `{name}_deleted` on soft-delete, freeing the original name for reuse under the spanning unique constraint (3 files)._

## Must fix before merge
1. [Important] TOCTOU race allows double-suffix under concurrent deletion - `src/support/brands-storage.js` (the UPDATE in the retry loop)

   The `status === 'deleted'` guard reads stale data. Between the SELECT and this UPDATE, a concurrent `deleteBrand` call on the same brand can complete its own rename. The second caller read `status != 'deleted'` (first caller had not committed yet), passes the JS guard, and issues its own UPDATE. Since neither UPDATE filters on status, the second one renames the already-renamed row again (e.g. `Acme_deleted` becomes `Acme_deleted2`) - the exact double-suffix scenario the idempotency guard was designed to prevent.

   The window is small and the consequence is cosmetic (a slightly wrong suffix on a deleted row), but it directly contradicts the JSDoc guarantee that a repeated delete never double-suffixes an already-renamed brand.

   **Fix:** Add `.not('status', 'eq', 'deleted')` to the UPDATE filter chain (PostgREST `.not` includes NULLs, so NULL-status rows are still reachable). If the row was already flipped to `deleted` by a concurrent request, the UPDATE matches zero rows, `data` is null, and you can return `true` (another caller completed the delete). This makes the write itself the concurrency guard rather than relying on the stale read. Alternatively, soften the JSDoc claim if the cosmetic race is acceptable.

<details>

- suggestion: Check constraint name in the 23505 handler before retrying - if a future unique constraint fires, the loop burns 100 attempts on an unresolvable violation. Add `!error.message?.includes('uq_brand_name_per_org')` to the throw condition. - `src/support/brands-storage.js`
- nit: Test fixture for "appends an incrementing index" omits the `status` field (`{ name: 'Acme' }`) - other fixtures correctly include `status: 'active'` or `status: null`. Add `status: 'active'` for explicitness. - `test/support/brands-storage.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 5m 42s | <b>Cost:</b> $7.63 | <b>Commit:</b> <code>2acb2a7ad408f1f0daa8cfe6452e5b505253b75b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

igor-grubic:
Thanks @MysticatBot — addressed the review in 09fabd1c8:

**Must fix — TOCTOU double-suffix race (Important):** the soft-delete UPDATE is now filtered with PostgREST `.not('status', 'eq', 'deleted')`, making the write itself the concurrency guard. If a concurrent `deleteBrand` already flipped the row to `deleted` between the SELECT and this UPDATE, we now match zero rows (returned as idempotent success) instead of re-suffixing an already-renamed brand. `.not(...eq...)` still matches NULL-status rows, so NULL-status live brands stay deletable. JSDoc updated to describe the write-level guard (and the stale "returns false → 404" line for the already-deleted case corrected to "returns true → idempotent").

**Non-blocking (both applied):**
- Constraint-name check in the 23505 handler: we only retry when the violation names `uq_brand_name_per_org`; any other unique violation now surfaces immediately instead of burning the 100-attempt budget.
- Test fixture `status` field: added `status: 'active'` to the incrementing-index and retry-budget fixtures.

Added two tests: (1) a concurrent zero-row UPDATE reports idempotent success with a single write, and (2) a 23505 from a different constraint surfaces immediately. All 219 unit tests pass; lint + type-check (base & strict) green.

MysticatBot (approved):
Hey @igor-grubic,

**Verdict: Approve** - all prior findings addressed; no new blocking issues.
**Complexity: MEDIUM** - medium diff (399 lines, 4 files), tests-heavy.
_Changes: Rewrites `deleteBrand()` to rename the brand to `{name}_deleted` on soft-delete, freeing the original name for reuse under the spanning unique constraint (4 files)._

- TOCTOU race closed: UPDATE now uses `.not('status', 'eq', 'deleted')` as the write-level concurrency guard
- Constraint-name check: only retries when the 23505 names `uq_brand_name_per_org`
- Test fixture status field: added `status: 'active'` to relevant fixtures

<details>

- suggestion: Consider logging at DEBUG level when the guarded UPDATE matches zero rows (concurrent no-op path) for future operational visibility - `src/support/brands-storage.js`
- suggestion: If the `name` column has a length limit, appending `_deleted` to a near-max-length name would raise a `22001` error not currently handled. Handle explicitly or document the assumption that the column is unbounded `text` - `src/support/brands-storage.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 54s | <b>Cost:</b> $5.57 | <b>Commit:</b> <code>3d1dd2b07a1547cc9ca119610119cf5aa7d7fea4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #3043, opened 2026-08-13
feat(serenity): read intent tags under the renamed `$abv_tags$intent` root | LLMO-6984

rainer-friederich:
## 1. Abstract

The Semrush `intent` dimension root is being renamed in place to `$abv_tags$intent`, project by project. This makes the API service resolve intent tags under either spelling, on both the Elements read path and the tag-tree write path.

## 2. Reasoning

Semrush hides a tag-tree entry whose name starts with the agreed `$abv_tags$` marker from the customer-facing Brand Presence tag filter. Renaming the intent root is how the prompt-intent dimension gets out of that filter — the same substring-hiding mechanism the dimension-root program already documents. The five values stay bare-named (`Informational`, `Commercial`, `Navigational`, `Transactional`, `Task`); only the root's name changes.

The rename runs from the data-service migration CLI against Semrush, one project at a time, so at any moment some projects carry the new name and some the old. Without this change the failure is silent rather than loud: an Elements tag filter is the `__`-joined tag PATH, so a dimension's prefix is literally its root's name. A filter built from the wrong spelling matches nothing and the endpoint answers HTTP 200 with an empty `userIntent` on every row — which renders the elmo Topic Intent Alignment tile as total misalignment rather than as an error.

**Read path — per-prompt `userIntent`.** The enrichment fires its five intent-filtered calls under `$abv_tags$intent` and retries once under `intent` only when all five succeeded and returned no rows. That empty result is the signature of a project the rename has not reached; a *failed* call is not, so an upstream having a bad day does not get a second round of calls for an answer that would be a guess. A renamed project therefore stays at five calls, not ten, and the retry stops firing by itself as the migration progresses — which is what makes removing the legacy spelling later a no-op rather than a behaviour change. The bounded waste is a project with genuinely no intent-tagged prompts, which pays a second round that also comes back empty.

**Read path — filter dimensions.** `page_intents` is populated from either prefix, and both are registered as known prefixes. That second half matters as much as the first: the unknown-prefix catch-all would otherwise have claimed a renamed project's tags and surfaced them in the URL-inspector filter payload as a dynamic group keyed `$abv_tags$intent` — putting the dimension back in front of customers under its internal name, beside an empty `page_intents`.

**Write path.** `intent` stays the dimension KEY everywhere it is a contract — the `type` a client names on the tag endpoints, the key of the closed vocabularies, the value elmo reads — and is mapped to the upstream root NAME only at the seam that talks to Semrush. Concretely:

- Root resolution provisions the renamed root on a project that has no intent root, and ADOPTS an existing pre-rename root rather than minting a second one beside it. Minting would split the dimension: the new root empty, every already-tagged prompt still under the old one, and no read seeing both.
- The resolved map is now keyed by dimension, so no caller has to know which spelling a given project carries.
- The tree walk folds a tag's raw root name to its dimension key. Two guards depend on this and would otherwise have failed OPEN on exactly the projects the rename reached: the one that refuses a client rename or re-parent of a server-owned value (a customer could have renamed `Informational`, and the next server write would mint a second one beside it), and the parent-placement check. Both spellings also stay reserved at the root level, so a customer value cannot shadow either.

**Observability.** Each of the three tolerances logs when it fires, under its own event key — `intent-rename-legacy-root-adopted` (the write path resolved a pre-rename root), `intent-rename-legacy-retry` (the read enrichment fell back) and `intent-rename-legacy-tags-present` (filter dimensions still saw `intent__` tags). So the follow-up that removes the legacy half is gated on three queries going quiet rather than on a judgement call about which projects the sweep reached.

## 4. Required information

- Jira / issue: LLMO-6984 — https://jira.corp.adobe.com/browse/LLMO-6984
- Spec: dimension-root tag model — https://github.com/adobe/serenity-docs/blob/main/docs/prompt-management/dimension-root-tag-model.md and the substring-hiding mechanism recorded in https://github.com/adobe/serenity-docs/blob/main/docs/prompt-management/origin-dimension.md
- Other: LLMO-6985 (the rename script + migration CLI support) — https://jira.corp.adobe.com/browse/LLMO-6985 ; LLMO-6986 (removes the legacy `intent__` fallback) — https://jira.corp.adobe.com/browse/LLMO-6986

- **mysticat-data-service** — contract. Performs the actual rename against Semrush from the migration CLI (LLMO-6985). This PR is what makes that rename safe to run project by project instead of as a coordinated cutover.
- **project-elmo-ui** — consumed, no change needed. Verified rather than assumed: `intent__` appears nowhere in that repo; `INTENT_DIMENSION` is declared but reaches only a type union, with no helper keying a prompt tag on the intent dimension. The Topic Intent Alignment tile consumes the server-derived `userIntent` string, and the intent filter compares the bare lowercased child values, which the rename does not touch. Its only use of the tag-tree root level is a lookup of the `category` root by name, so the renamed root passes through it unread. A duplicate ticket should not be opened for elmo.

Run against live Semrush (`adobe-hackathon.semrush.com`) with a user IMS bearer, driving the real Elements read path — CI cannot produce this.

**Renamed project** — LLMO-Dev-2 Brand-A CH-de (`997e17c3-1566-47d4-a84f-3db00870bdf7`), whose intent root is already `$abv_tags$intent`: 909 prompts returned, `userIntent` resolved on all 909 (informational 235, commercial 370, task 303, transactional 1), in exactly six Elements calls — one base plus five intent-filtered. No legacy retry fired.

**Un-renamed project with real data** — Adobe Helpx US-en (`cd713e77-e714-4ad5-86b4-6f0a8650977c`), an Adobe-owned brand on prod: 9,690 prompts, `userIntent` resolved on all 9,690, in eleven calls — the five new-prefix calls came back empty, the retry fired and the five legacy calls carried the rows. The dev reference project named in the ticket for this case (Brand-A US-en) turns out to have no prompt executions under any model, so it could not demonstrate it; this prod project could.

**Filter dimensions** — the LLMO-Dev-2 workspace currently holds BOTH shapes at once, which exercises the mixed case directly: `page_intents` came back carrying entries from each prefix with bare labels, no `$abv_tags$intent` group appeared in the payload, and the ungrouped `tags` array stayed empty.

**Unrelated observation, not addressed here.** The same live payload shows the `origins` dimension has the analogous problem from the earlier `source` → `origin` authorship rename: `transformOriginsToFilterDimensions` still extracts `source__` only, so `origins` returns producing-system values mixed with authorship ones (`source__config` beside `source__ai` / `source__human`) while a renamed project's `origin__human` falls through to the catch-all as a stray dynamic `origin` group. Reproduced on both the dev and the prod workspace above. Out of scope for this ticket — flagging it rather than widening the change.

## 7. Test plan

Beyond unit and integration coverage of both tag shapes, the end-to-end verification is the live run recorded in section 6: both root spellings resolved `userIntent` against real Semrush data, the renamed project cost five intent calls and the un-renamed one ten, and the filter-dimension payload was inspected on a workspace holding both shapes.

The write path's no-duplicate-root behaviour is covered against the Semrush vendor mock in the integration suite (a project provisioned through the API now gets `$abv_tags$intent`, so the migration does not have to come back for projects created after this ships). It was deliberately NOT exercised against live Semrush, since that would mutate a real project's tag tree.

To verify on dev after deploy: call the URL-inspector prompts endpoint with `userIntent` enrichment for a project on each side of the rename and confirm the field is populated in both cases; then read the URL-inspector filter dimensions for a workspace holding a renamed project and confirm `page_intents` is populated and no `$abv_tags$intent` key is present. The three transition events (`intent-rename-legacy-root-adopted`, `intent-rename-legacy-retry`, `intent-rename-legacy-tags-present`) are the signal for how far the sweep has got.

Related to LLMO-6985 (the rename run) but deliberately NOT ordered against it — removing that ordering constraint is the point of the dual-spelling read. This can merge and deploy before or after any project is renamed, and the rename can then roll project by project. LLMO-6986 is the cleanup that follows once the three transition log lines have gone quiet everywhere.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```

### DIFF
--- a/docs/openapi/schemas.yaml
+++ b/docs/openapi/schemas.yaml
@@ -11859,7 +11859,10 @@ SerenityPrompt:
         upstream, so a prompt can carry two tags with the same bare name from
         different dimensions (a sub-category `human` and the `source` value
         `human`). Read a tag's dimension from `path[0].name`, never from its
-        own name.
+        own name — bearing in mind that `path[0].name` is the root's UPSTREAM
+        name, which for the intent dimension is `$abv_tags$intent`, or the
+        pre-rename `intent` on a project the rename has not reached. Match both
+        spellings when keying on that dimension.
     tagIds:
       type: array
       items: { type: string }
@@ -11907,7 +11910,10 @@ SerenityPromptTag:
     parentage. What varies is DEPTH, not endpoint: upstream omits `parent_id` and
     `path` on a dimension root and carries both on a descendant. This API
     normalizes that absence to `null`, so a tag whose `path` is `null` is a
-    dimension root and its own `name` is its dimension.
+    dimension root and its own `name` names its dimension upstream. Names are
+    passed through exactly as upstream holds them, so the intent root reads
+    `$abv_tags$intent` — or the pre-rename `intent` on a project the rename has
+    not reached. The dimension is `intent` either way.
   required: [id, name]
   properties:
     id:
@@ -12795,7 +12801,11 @@ SerenityUrlInspectorFilterDimensions:
       items: { $ref: '#/SerenityFilterDimensionItem' }
     page_intents:
       type: array
-      description: Distinct `intent__`-prefixed tags. `id` is the original `intent__`-prefixed tag value.
+      description: >-
+        Distinct intent-prefixed tags. The prefix is the intent root's name on that
+        project — `$abv_tags$intent__`, or `intent__` on a project the rename has not
+        reached yet — and `id` is the original tag value, carrying whichever prefix
+        that project uses.
       items: { $ref: '#/SerenityFilterDimensionItem' }
     origins:
       type: array
--- a/docs/openapi/serenity-api.yaml
+++ b/docs/openapi/serenity-api.yaml
@@ -546,9 +546,9 @@ v2-serenity-tags:
                 description: |
                   A BARE value — no dimension prefix. Must not contain ':' and
                   must not be one of the reserved dimension root names
-                  (`category`, `intent`, `source`, `type`). For a closed
-                  dimension it must exactly match one of that dimension's fixed
-                  enum values.
+                  (`category`, `intent`, `origin`, `type`, `source`,
+                  `$abv_tags$intent`). For a closed dimension it must exactly
+                  match one of that dimension's fixed enum values.
               geoTargetId:
                 type: integer
                 minimum: 1
@@ -687,7 +687,8 @@ v2-serenity-tag-by-id:
                 description: |
                   The tag's new BARE name — no dimension prefix. Must not
                   contain ':' and must not be one of the reserved dimension root
-                  names (`category`, `intent`, `source`, `type`).
+                  names (`category`, `intent`, `origin`, `type`, `source`,
+                  `$abv_tags$intent`).
               parentId:
                 type: string
                 description: |
@@ -1095,7 +1096,10 @@ v2-serenity-url-inspector-filter-dimensions:
 
       `topics`/`categories`/`page_intents`/`origins` are derived from the same
       underlying Topics element response, split apart by their `topic__`/
-      `category__`/`intent__`/`source__` tag prefix. Any other tag prefix Semrush
+      `category__`/`$abv_tags$intent__`/`source__` tag prefix. A tag prefix is the
+      name of the dimension's root tag, so `page_intents` also claims the
+      pre-rename `intent__` prefix on a project the intent rename has not reached.
+      Any other tag prefix Semrush
       introduces later (e.g. `type__branded`) surfaces automatically as an extra
       array property keyed by that prefix. Bare prefix declarations with no
       value (e.g. a lone `category` row) are ignored. `content_types` is
--- a/src/support/elements/constants.js
+++ b/src/support/elements/constants.js
@@ -13,6 +13,19 @@
 export const DEFAULT_ELEMENT_MODEL = 'search-gpt';
 
 /**
+ * Semrush's Elements API tag encoding: `prefix__value`, with `__` also used for
+ * `parent__child` nesting within the value. A tag's wire form is its `__`-joined
+ * PATH, so the leading segment is the name of its ROOT tag — which is why renaming
+ * a dimension root renames every one of its tags on this surface.
+ *
+ * This replaced an earlier `prefix:value` encoding (colon-delimited, `Parent__Child`
+ * nesting only within the value). The cutover was a one-time, atomic migration on
+ * Semrush's side — all workspaces/customers moved to `__` together, so there is no
+ * dual-format transition period and nothing parses both encodings.
+ */
+export const SEP = '__';
+
+/**
  * Sentinel platform/model value meaning "all platforms" — no single-model filter.
  * Mirrors the UI's `PLATFORM_CODES.All` ('all'), the same "no filter" convention the
  * category/region dimensions already use. Elements that support it OMIT the `CBF_model`
--- a/src/support/elements/definitions/prompts.js
+++ b/src/support/elements/definitions/prompts.js
@@ -36,9 +36,11 @@ export const INTENT_ENRICH_CONCURRENCY = 5;
  *   the shape Semrush's own UI sends for this element.
  * - **tags** — each value becomes its own `tags contains <value>` clause; multiple
  *   tags are AND-ed (a prompt must carry all). Callers pass the FULL prefixed tag
- *   value — the element's tag taxonomy varies by workspace (`type__`, `category__`,
- *   `intent__`, `source__`, `topic__`), so no prefix is assumed here. e.g. pass
- *   `type__branded` for the branded count, `category__Brand` for a category filter.
+ *   value — the prefix is the tag's root name, which varies by workspace and, for
+ *   the intent dimension, by how far the `$abv_tags$intent` rename has reached
+ *   (`type__`, `category__`, `$abv_tags$intent__`, `source__`, `topic__`), so no
+ *   prefix is assumed here. e.g. pass `type__branded` for the branded count,
+ *   `category__Brand` for a category filter.
  * - **projectIds** — a `CBF_project` `or` group; omitted → all projects in the
  *   (sub-)workspace. The UI already surfaces these as `semrush_project_id` via the
  *   URL Inspector filter-dimensions `regions`.
--- a/src/support/elements/definitions/topics.js
+++ b/src/support/elements/definitions/topics.js
@@ -10,7 +10,8 @@
  * governing permissions and limitations under the License.
  */
 
-import { resolveElementModel } from '../constants.js';
+import { resolveElementModel, SEP } from '../constants.js';
+import { INTENT_ROOT_NAME, LEGACY_INTENT_ROOT_NAME } from '../../serenity/prompt-tags.js';
 
 /**
  * Builds the payload for the Topics (Tags) filter-dimensions element (row 3).
@@ -44,14 +45,6 @@ export function buildTopicsPayload({
     filters: { advanced: { op: 'and', filters } },
   };
 }
-
-// Semrush's Elements API tag encoding: `prefix__value`, with `__` also used for
-// `parent__child` nesting within the value. This replaced an earlier `prefix:value`
-// encoding (colon-delimited, `Parent__Child` nesting only within the value). The
-// cutover is a one-time, atomic migration on Semrush's side — all workspaces/customers
-// were migrated to `__` together, so there is no dual-format transition period and no
-// need to parse both encodings here.
-const SEP = '__';
 
 /**
  * @typedef {object} FilterDimensionItem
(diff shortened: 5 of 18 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-designed rolling-rename tolerance with clear boundaries and a planned cleanup path.
**Complexity: HIGH** - large diff; API surface + cross-concern read/write/filter paths.
_Changes: Adds dual-spelling support for the intent dimension root rename (`$abv_tags$intent` / `intent`) across the Elements read path, filter-dimension extraction, and tag-tree write path (18 files)._

<details>

- nit: The 400 error message in `parseCreateTagBody` and `parseUpdateTagBody` shows `RESERVED_ROOT_NAMES` which includes `$abv_tags$intent` - an internal Semrush implementation detail that has no meaning to a customer. Consider showing only the domain-relevant dimension names. - `src/support/serenity/handlers/tags.js:207`
- suggestion: The three structured log lines gating LLMO-6986 removal are natural-language strings. Adding a one-word `event` field in the metadata (e.g. `{ event: "intent-rename-legacy-hit", workspaceId }`) would make the Splunk/Datadog query trivial and rewrite-proof. - `src/support/elements/elements-service.js:138`
- suggestion: The `noIntentTag` test matcher uses `.includes("intent__")` rather than a more precise check. A hypothetical future tag value containing the substring `intent__` anywhere in its body would incorrectly match. Unlikely to bite, but subtly different from the production logic. - `test/support/elements/elements-service.test.js:333`
</details>

Note: The retry cost on genuinely intent-less projects (10 calls instead of 5) is explicitly acknowledged in the PR description as bounded waste that disappears with LLMO-6986. If the migration timeline slips, consider a cheaper root-level probe or caching signal to avoid the double-round on every request for those workspaces.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 5s | <b>Cost:</b> $9.30 | <b>Commit:</b> <code>fbdf9be1a163edf095e61298208862548be70d69</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the review in `fd25a0d1c` — all three non-blocking items taken.

MysticatBot:
- nit: reserved-name 400 leaks `$abv_tags$intent` — fixed. The CHECK still covers both intent spellings, but the message enumerates only the dimension names, so the Semrush-internal marker no longer reaches a customer-facing error. Both the create and update reserved-name tests now assert the marker is still refused, so the narrower message cannot quietly become a narrower check. The OpenAPI description keeps naming it — that surface is for integrators, where the complete reserved set is the useful answer.
- suggestion: structured `event` key on the three transition log lines — fixed. They now carry `intent-rename-legacy-root-adopted` (tag-tree resolved the pre-rename root), `intent-rename-legacy-retry` (the Elements enrichment fell back) and `intent-rename-legacy-tags-present` (filter dimensions still saw `intent__` tags), so "has the sweep finished?" is three queries rather than a natural-language grep. The removal criterion is recorded next to `LEGACY_INTENT_ROOT_NAME`, which is what LLMO-6986 deletes.
- suggestion: `noIntentTag` matched on a substring — fixed. It now tests the root prefix under either spelling, the same shape production builds the filter with, so a tag merely containing `intent__` further in no longer satisfies it.

On the closing note about the double round on intent-less projects: leaving it as-is deliberately. A per-project memo of the winning spelling is the obvious cheaper answer, but it puts cross-request cache state in the Lambda for a cost that is bounded and self-cancelling, and the retry is the last of the three tolerances that should be removed rather than the first — a project with genuinely no intent-tagged prompts is indistinguishable from an un-renamed one, so it stays correct either way. Worth revisiting only if the migration timeline actually slips.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all three non-blocking items from the prior review are cleanly addressed.
**Complexity: HIGH** - large diff; API surface + cross-concern read/write/filter paths.
_Changes: Adds dual-spelling support for the intent dimension root rename (`$abv_tags$intent` / `intent`) across the Elements read path, filter-dimension extraction, and tag-tree write path (18 files)._

- Reserved-name 400 no longer leaks `$abv_tags$intent` to customers - error message now enumerates `ALL_DIMENSIONS` while the check still covers every reserved name.
- Transition log lines now carry structured `event` keys (`intent-rename-legacy-root-adopted`, `intent-rename-legacy-retry`, `intent-rename-legacy-tags-present`) for queryable sweep-progress gating.
- `noIntentTag` test matcher uses `startsWith` on both root spellings rather than a loose `.includes("intent__")` substring match.

Note: CI checks are pending (build, integration tests, type-check) - verify green before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 5m 57s | <b>Cost:</b> $1.70 | <b>Commit:</b> <code>fd25a0d1c39ab8bd2286a752d90ede4afa830ac3</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Runtime validation.

Environment: this branch owns the shared `dev-branches` slot — the deployment record shows `fd25a0d1c` / `feat/new_intents` deployed at 13:35:14Z.

The change's runtime surface is what this service sends to, and parses back from, the Semrush Elements API, so it was exercised directly against live Semrush (`adobe-hackathon.semrush.com`) with a user IMS bearer, driving the real read path on both root shapes:

- Renamed project (LLMO-Dev-2 Brand-A CH-de, `997e17c3-1566-47d4-a84f-3db00870bdf7`, root already `$abv_tags$intent`): 909 prompts, `userIntent` resolved on all 909, six Elements calls (one base + five intent-filtered), no legacy retry.
- Un-renamed project with real data (Adobe Helpx US-en, `cd713e77-e714-4ad5-86b4-6f0a8650977c`): 9,690 prompts, `userIntent` resolved on all 9,690, eleven calls — the five new-prefix calls empty, the retry firing, the five legacy calls carrying the rows.
- Filter dimensions on a workspace holding BOTH shapes at once: `page_intents` carried entries from each prefix with bare labels, no `$abv_tags$intent` group in the payload, `tags` empty.

Not exercised: the deployed `/serenity/brand-presence/prompts` route on the dev alias. Serenity routes need an IMS-backed promise token rather than a session token, and the local session token is expired — refreshing it needs an interactive browser login. That is a pre-existing dev-auth limitation, unrelated to this change, and the layer it would have added (routing and auth) is not what this PR touches; the Elements request/response handling that it does touch is what the runs above cover.

rainer-friederich:
Correction to the runtime-validation comment above: the claim that this branch owned the `dev-branches` slot was wrong.

`ci / branch-deploy` reported success and a GitHub deployment record was created for this branch's SHA, but every deploy step inside the job was skipped:

`##[warning]Branch is 2 commit(s) behind main — DEV deploy will be skipped`

The job still creates the deployment record and still goes green, so both `gh pr checks` and the deployments API indicated a deploy that never happened. The alias was in fact serving an unrelated branch — the healthcheck reported a build id belonging to `feat/deploy-suggestion-applystale`. I asserted ownership from the deployment record instead of from the running artifact, which is the one thing that actually answers it.

`origin/main` is now merged in, so the next branch-deploy on this branch should genuinely claim the slot. Nothing in the Semrush-side verification changes — those runs drove the real Elements API directly and never went through the dev alias.

rainer-friederich:
Follow-up on the runtime validation, with the deploy question settled and the auth chain traced.

The slot is now genuinely this branch's: after merging `origin/main` the deploy actually ran, and the dev alias healthcheck reports build `0.0.0+latest31706549024` — the run id of this branch's push. That is the check worth trusting; the deployment record and a green `branch-deploy` both claimed a deploy earlier that had been skipped.

The endpoint itself still could not be exercised, and the reason is authorization rather than anything about this change. Three separate things had to be untangled:

- The dev alias must be reached on the LLMO host. Fastly injects `x-product` server-side per host, and a client-sent header is not honoured, so the same request answers `403 x-product header required for FACS-governed routes` on the spacecat host and `200` on the LLMO one.
- A dev session token is needed, not a prod one.
- With both of those right, the brand-scoped serenity route answers `403 Forbidden`, and the token's own claims say why: the identity holds an ASO free-trial entitlement in a single tenant, with no LLMO entitlement, no admin or LLMO-administrator flag, no FACS permissions and no delegated tenants. `GET .../brands` returns 200 with an empty list on the same identity, which is the same fact seen from the other side.

Exercising this endpoint on dev therefore needs an identity with LLMO entitlement on an org whose brands have mapped Semrush projects. The behaviour this PR changes — which tag prefix goes on the Elements filter and how the response is parsed — is covered by the live runs against Semrush recorded above, which never went through this alias.

One environment note for anyone validating on `dev-branches` later: the slot changed hands twice within about ten minutes while this was being checked. An observation there is only meaningful if the healthcheck build id still matches your own run at the moment of the call.

PR #3044, opened 2026-08-13
fix(serenity): stop benchmark alias writes from destroying what Semrush added

rainer-friederich:
## 1. Abstract
Benchmark alias writes now merge over the list Semrush actually holds instead of replacing it, so the aliases its own brand resolution added survive a brand edit. Alias spellings follow upstream's identity model: one spelling per alias, folded by case.

## 2. Reasoning
A benchmark's `brand_aliases` carries names Semrush added itself, and they are not lowercase copies of anything we hold — benchmark `General Motors` carries `gm`, `Mercedes-Benz` carries `mercedes`, `pixlr` carries the misspelling `pixlar`, `Star Health Insurance` carries `star health & allied insurance`. The only write upstream offers on a benchmark is a PUT that replaces the whole alias list; there is no PATCH and no per-alias route. The sync sent its own values as that whole list, so every alias write dropped Semrush's, and resolution never re-runs to put them back.

Measured on the prod tenant on 2026-08-13 across 307 active brands, 470 projects and 2,738 benchmarks: 493 benchmarks hold 720 such values, and the next alias write to any of them takes them.

The competitor path was worse. A rename sent `{brand_name, domain}` with no `brand_aliases` at all, on the belief — stated in the code comment — that omitting the field left the stored list alone. It clears it. So renaming a competitor we hold no aliases for wiped whatever Semrush had put there.

Before: the own-brand and competitor syncs computed the alias list from the brand row and PUT it as the complete `brand_aliases`, diffing against the published view with a case-insensitive comparison. Anything upstream had added was silently replaced, and a name-only competitor rename omitted the field entirely and cleared the list.

After, the write is a merge over what is live:

- Every alias already on the benchmark is carried forward. Only the aliases this edit removed are subtracted, which needs the brand's previous aliases and competitors — read before the row is written and threaded into the two syncs.
- `brand_aliases` is always sent, even when empty, so an update that changes only a name can no longer clear the list.
- The alias diff reads the draft view (`?draft=true`). Writes act on the draft, so the published listing was a stale snapshot on any project with pending changes.
- Alias spellings are folded by case, one per alias. Upstream identifies an alias by its case-folded value and refuses a list holding two spellings of one alias with a 409 that fails the whole write, so that list can no longer be constructed. A newly created alias takes the lowercase spelling Semrush itself uses when it creates one; an alias already there keeps its stored spelling, because a PUT cannot re-case it.

Operationally visible: a steady-state re-sync now issues no benchmark write and no republish, where previously any alias-touching edit rewrote and republished. The aliases an edit itself removes are also filtered out of what is reported back to the caller as not-applied — upstream lists removed aliases in `rejected_brand_aliases`, so those were being surfaced to the operator as if they were a problem.

Not user-visible: the database keeps the customer's casing. `brand_aliases.alias` and `competitors.aliases` are unchanged, which is what the UI renders and what our own matchers fold anyway.

## 4. Required information
- Jira / issue: SITES-49840 — https://jira.corp.adobe.com/browse/SITES-49840
- Spec: serenity-docs `docs/provisioning/benchmark-brand-alias-casing.md` — https://github.com/adobe/serenity-docs/pull/342
- Other: re-casing the aliases already stored in the customer's casing is SITES-49841 — https://jira.corp.adobe.com/browse/SITES-49841

- `mysticat-data-service` — the `serenity_migration` CLI writes the same benchmarks and carries the matching half of this change (additive merge, folded comparison, lowercase spellings in the plan). Changed, on the same session branch.
- `project-elmo-ui` — consumes the brand aliases as stored and drives the edit through `PATCH /brands`. Contract unchanged: the database keeps the customer's casing, so the rendered alias list is unaffected.

Everything below was run against the live Semrush tenant `https://adobe-hackathon.semrush.com` on 2026-08-13 with a user IMS bearer token, on dev-DB fixture brands. Both fixtures were restored to their original alias lists and republished afterwards.

- **Census** — read-only walk of `brands` (active, `semrush_sub_workspace_id` set) from the prod reader, then the projects and benchmarks per brand. 2,738 benchmark rows, no request errors. It is the source of the 493-benchmark / 720-value exposure, and of the finding that no benchmark anywhere holds two spellings of one alias (0 of 2,738).
- **Contract probe** — `Allow` headers read from the router plus method probes aimed at a nonexistent benchmark id, so nothing could be mutated. The item route answers `Allow: PUT`; v2 and v3 item routes return `404 page not found`; all six spellings of an alias sub-resource 404 while `…/brand_urls` returns 200. Confirms the full-replace PUT is the only write available.
- **Write probes** established the model the code now follows: two spellings of one alias in a single list → `409 duplicate brand name or alias`; a PUT cannot re-case an alias already stored; omitting `brand_aliases` empties the list; a body without `domain` fails with `400 primary_url is required`; and an alias listed in `rejected_brand_aliases` can be re-submitted and lands.
- **End-to-end run of this code** against the tenant, driving `syncBrandAliasesAcrossMarkets` on the "Bark" fixture: adding an alias PUT `["Bark","bark collar"]` — the live spelling kept, the new alias lowercase — an immediate re-run issued zero writes and zero publishes, and removing the alias again restored `["Bark"]` with the project `live`. The first version of this change failed that middle step, rewriting and republishing on every run.

## 7. Test plan
Locally, beyond unit tests: the end-to-end run described in section 6 — add, re-run, remove — against the live tenant on a dev-DB fixture, which is the only place the upstream identity and spelling rules can actually be exercised.

To verify on **dev**: edit a brand's aliases through `PATCH /brands` for a brand with a sub-workspace, then read the benchmark before and after with `GET /enterprise/projects/api/v1/workspaces/{ws}/projects/{pid}/ai_models/benchmarks?draft=true`. Expect the new alias present in lowercase, every previously present alias still there, and the project back at `publish_status: live`. Repeat the same edit with no change and expect no benchmark write in the logs. Rename a competitor and confirm its alias list is unchanged.

On **stage** and **prod**: same check on a brand whose benchmark carries vendor-added values (`gm` on General Motors is the clearest), confirming they are still present after an unrelated alias edit. Re-running the census script is the population-level acceptance check.

- https://github.com/adobe/serenity-docs/pull/342 — related. The design of record. Worth merging first so it is on `main` before this lands, though nothing here depends on it mechanically.
- https://github.com/adobe/mysticat-data-service/pull/856 — related, and deliberately staying unmerged for now. It carries the same alias model for the migration CLI, so until it lands the CLI on `main` keeps replacing benchmark aliases rather than merging them. That is a separate service with separate operators and does not block this change.

This PR is self-contained: it merges and deploys on its own.

SITES-49841 (re-casing existing aliases) must not run until this is deployed, or the population it repairs keeps growing behind it.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```

### DIFF
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -60,6 +60,7 @@ import {
   getBrandById,
   getBrandBySite,
   getBrandCompetitors,
+  getBrandAliases,
   readSerenityFlagScopes,
   withSerenityState,
 } from '../support/brands-storage.js';
@@ -1972,8 +1973,14 @@ function BrandsController(ctx, log, env) {
         ? await getBrandCompetitors(brandUuid, postgrestClient)
         : [];
       // Brand aliases (the extra names the brand is known by) re-sync to every
-      // market's project brand_names + own-brand benchmark on edit.
+      // market's project brand_names + own-brand benchmark on edit. Captured
+      // BEFORE the update for the same reason as the competitors above: the
+      // benchmark alias write removes only the aliases this edit dropped, so that
+      // the values Semrush's brand resolution added there survive it.
       const aliasesTouched = updates.brandAliases !== undefined;
+      const oldAliases = aliasesTouched
+        ? await getBrandAliases(brandUuid, postgrestClient)
+        : [];
 
       // Per-brand serenity rollout gate. An edit that changes URL sources /
       // competitors / aliases re-syncs onto the brand's Semrush projects (the
@@ -2270,6 +2277,7 @@ function BrandsController(ctx, log, env) {
               // of the brand's own properties.
               updated.urls,
               sharedProjects,
+              oldCompetitors,
             );
             rejectedAliases.push(...(competitorResult?.rejected ?? []));
           }
@@ -2281,6 +2289,7 @@ function BrandsController(ctx, log, env) {
               updated.semrushSubWorkspaceId,
               log,
               sharedProjects,
+              oldAliases,
             );
             rejectedAliases.push(...(aliasResult?.rejected ?? []));
           }
--- a/src/support/serenity/aliases.js
+++ b/src/support/serenity/aliases.js
@@ -165,6 +165,45 @@ export function mergeBenchmarkAliases(live, desired, removed = []) {
 }
 
 /**
+ * The alias keys other benchmarks in the same project already own — every OTHER
+ * benchmark's `brand_name` and `brand_aliases`, case-folded.
+ *
+ * Upstream enforces alias uniqueness across the union of every benchmark's name and
+ * aliases WITHIN A PROJECT, case-insensitively, and a duplicate is refused with a
+ * `409 duplicate brand name or alias` that fails the WHOLE write — for a batched
+ * create, every benchmark in the batch, including the ones that would have been
+ * fine (live-verified 2026-08-13). Sending an alias another benchmark owns is
+ * therefore not a small mistake; it takes unrelated work down with it.
+ *
+ * Callers subtract this from what they were going to send. The values are dropped
+ * silently rather than reported: they are ours to derive, so a customer cannot act
+ * on them, and the alternative is a failed edit.
+ *
+ * @param {Array<object>} benchmarks - `aio_benchmarks` from `listBenchmarks`.
+ * @param {string} [ownId] - the benchmark being written, excluded from the set.
+ * @returns {Set<string>} case-folded keys.
+ */
+export function aliasKeysOwnedByOthers(benchmarks, ownId) {
+  const owned = new Set();
+  for (const b of Array.isArray(benchmarks) ? benchmarks : []) {
+    if (ownId !== undefined && ownId !== null && String(b?.id) === String(ownId)) {
+      // eslint-disable-next-line no-continue
+      continue;
+    }
+    const values = [b?.brand_name, ...(Array.isArray(b?.brand_aliases) ? b.brand_aliases : [])];
+    for (const v of values) {
+      // Key on the trimmed value, not the raw one: `hasText` counts whitespace as
+      // text, so a blank name would otherwise add an empty key to the set.
+      const key = typeof v === 'string' ? v.trim().toLowerCase() : '';
+      if (hasText(key)) {
+        owned.add(key);
+      }
+    }
+  }
+  return owned;
+}
+
+/**
  * Whether two alias lists hold the same spellings (order ignored, casing
  * significant). The benchmark write decision uses this rather than
  * {@link sameAliasSet} so that a re-cased alias registers as a difference.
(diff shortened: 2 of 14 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-scoped bug fix with thorough merge semantics, comprehensive tests, and clear documentation of upstream behavior.
**Complexity: HIGH** - large diff; API surface change across multiple sync paths.
_Changes: Benchmark alias writes now merge over the live list instead of replacing it, preserving vendor-added aliases and fixing the omit-clears-list behavior on competitor renames (10 files)._

Note: CI checks are currently failing (build) - resolve before merge.

<details>

- nit: `removedForMarket` in `brand-aliases.js:174` uses `desiredAliases.includes(a)` (case-sensitive) to detect removed aliases, while alias identity is case-insensitive. Both sides come from `collectAliasNames` (same DB rows) so spellings will match in practice, but a case-insensitive check would be more robust against future callers. Same pattern at `competitor-benchmarks.js:248`.
- suggestion: The extra `listBenchmarks` call after a write (to read `rejected_brand_aliases`) doubles the API calls per benchmark update. Consider reading the rejected list from the write response if the PUT returns the updated benchmark, or batching the rejected-alias collection to a single read at the end of the market loop - `brand-aliases.js:189`.
- nit: `sameAliasSetExact` uses `[...new Set(trimmed(a))]` which does not guarantee a stable iteration order across engines for the purpose of the `every` check, but since the comparison is set-membership (`set.has(v)`), not positional, this is fine. The intermediate array from `new Set` is only used for length comparison - `aliases.js:176`.

</details>

rainer-friederich:
Thanks — review addressed, and the build failure with it.

**Fixed**

- The `removedForMarket` / competitor equivalent now compare on the folded value. You were right that it was inconsistent with alias identity, and the consequence was concrete: re-spelling an alias would have read as removing one and adding another.
- The build failure was a real miss of mine, not flakiness: `test/support/serenity/handlers/markets-subworkspace.test.js` asserts the competitor create body, and my local run globbed `test/support/serenity/*.test.js` without the `handlers/` subdirectory. That benchmark now carries the lowercase name like every other create.

**Declined**

- Batching the post-write `listBenchmarks` read: worth doing, but it predates this change and is not caused by it — the read only fires when aliases were actually written. Folding an unrelated call-pattern optimisation into a fix that already spans two sync paths would make it harder to review, not easier. Happy to raise it separately.
- The `sameAliasSetExact` Set-iteration note: agreed with your own conclusion that it is fine — the comparison is membership, not positional, and the array is used only for length.

One thing worth flagging that the review could not see: alias uniqueness turned out to be **project-wide and case-folded**, not scoped to the submitted list, and the 409 fails the whole write — a batched create of two competitors, one carrying an alias a sibling benchmark already held, created neither. Both write paths now subtract the aliases other benchmarks own, verified against the live tenant. A benchmark *name* colliding with another benchmark name is the same 409 and cannot be filtered, since the name is the identity; the migration CLI can reclaim such a holder and api-service still has no equivalent, which is recorded in the design doc rather than widened into this PR.

PR #3042, opened 2026-08-13
feat(suggestions): support applyStale on deploy-to-edge

dhavkuma:
## Summary
- `deploySuggestionToEdge` now reads an optional `applyStale` boolean from the request body and passes it through to `TokowakaClient.deployToEdge(...)` as `metadata: { applyStale }`, defaulting to `false` when omitted.
- Only wired into the synchronous (direct) deploy path — async experiment mode is unaffected.
- Documents the new field in `docs/openapi/site-opportunities.yaml`.

Depends on adobe/spacecat-shared#1878 for the `suggestion.data.applyStale` mirror + patch-level plumbing on the client side; this PR only threads the request body field through the controller, so it's inert until spacecat-shared releases and this repo bumps its `@adobe/spacecat-shared-tokowaka-client` dependency.

## Test plan
- [x] `npx mocha test/controllers/suggestions.test.js` — 519 passing
- [x] `npm run lint` — clean
- [x] `npm run docs:lint` / `npm run docs:build` — valid, no new warnings
- [x] New tests cover: `applyStale: true` passed through as metadata, and defaulting to `false` when omitted from the request body

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3049, opened 2026-08-13
fix: only pass applyStale downstream in deploySuggestionToEdge when set

dhavkuma:
## Summary
- `deploySuggestionToEdge` now reads `applyStale` once from `context.data` and only threads it downstream (into the async GeoExperiment's `metadata` and into the sync `deployToEdge` call's `metadata`) when it's actually defined in the request body — previously each path independently coerced a missing value to a default (`true`-ish for the experiment path via `=== true` forcing an unconditional key, and `false` for the sync path).
- Companion change: adobe-rnd/llmo-experimentation-engine#74 (the engine side reads the same field back off `GeoExperiment.getMetadata()` before calling `deployToEdge`, with the same omit-if-undefined behavior).

This will be merged after PR https://github.com/adobe/spacecat-api-service/pull/3042
## Test plan
- [x] `npx mocha test/controllers/suggestions.test.js` — 521 passing
- [x] `npm run lint` on changed files


docs/openapi/site-opportunities.yaml
```diff
@@ -629,6 +629,15 @@ site-opportunity-suggestions-edge-deploy:
...
                 description: Optional experiment name (defaults to opportunity type). Only used in async experiment mode.
+              applyStale:
+                type: boolean
+                default: false
+                description: |
+                  When true, the deployed patch keeps applying even if the source page's content
+                  goes stale (e.g. a missing/outdated Last-Modified header). Mirrored onto each
+                  deployed suggestion's `data.applyStale` for display, and set on the underlying
+                  edge patch itself. Only synchronous (direct) deployment mode is supported;
+                  ignored in async experiment mode.
```
MysticatBot, line 640: issue (blocking): The description states "Only synchronous (direct) deployment mode is supported; ignored in async experiment mode." However, the code at `src/controllers/suggestions.js:2142` explicitly threads `applyStale` into `metadataBase` in the async GeoExperiment path, and the tests confirm it is persisted onto GeoExperiment metadata. The companion PR (`llmo-experimentation-engine#74`) also reads this field from `GeoExperiment.getMetadata()` before deploying.

This is actively misleading for API consumers. Either:
1. Remove the "ignored in async" claim (if async support is intentional, which the code and companion PR suggest), or
2. Guard the async branch to actually skip `applyStale` if the documentation is the source of truth.

Given the companion engine PR consumes the field from experiment metadata, option 1 (fix the spec) appears correct.

MysticatBot (changes requested):
Hey @dhavkuma,

**Verdict: Request changes** - OpenAPI documentation actively contradicts the implementation.
_Changes: Threads `applyStale` from the request body into downstream metadata only when explicitly set, fixing prior coercion inconsistencies between async and sync deploy paths (4 files)._
Note: Recommend a human read before merge - this change modifies a shared API contract (`docs/openapi/site-opportunities.yaml`) and the conventions reviewer flagged a spec/architecture divergence (heuristic). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Important] OpenAPI spec contradicts implementation for async mode - `docs/openapi/site-opportunities.yaml:639` (details inline)

<details>

- nit: `default: false` in the OpenAPI schema implies the server substitutes `false` when the field is omitted, but the code's intent is to not thread the value at all when absent. If omission and explicit `false` produce different downstream behavior, this annotation is misleading - `docs/openapi/site-opportunities.yaml:634`
- suggestion: No explicit boolean coercion on `applyStale` before propagation. If request-body schema validation middleware enforces types before the controller, this is already mitigated. Otherwise consider a `typeof` guard - `src/controllers/suggestions.js:1918`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 11s | <b>Cost:</b> $5.53 | <b>Commit:</b> <code>c6aeb5fdcceff445231d6595b1e41c65414ba650</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #3052, opened 2026-08-13
feat(suggestions): add PATCH edge-apply-stale endpoint to toggle applyStale without redeploy

dhavkuma:
## Summary
- New `PATCH /sites/:siteId/opportunities/:opportunityId/suggestions/edge-apply-stale` endpoint (`setSuggestionsApplyStale`) toggles `applyStale` for a set of already-deployed suggestions without redeploying them.
- Request body: `{ suggestionIds: string[], applyStale: boolean }`. Same access-control shape as `edge-rollback` (org access → LLMO administrator → site owner), same per-suggestion 207 multi-status response shape as `edge-deploy`/`edge-rollback`.
- Per-suggestion eligibility: 404 if not found, 400 if not yet deployed (`edgeDeployed` unset); suggestions with no matching S3 patch are reported as failed by the underlying client call.
- Wired into `routes/index.js`, both FACS capability maps (`llmo/can_deploy`, `aso/can_deploy`, matching sibling edge-ops routes), and the S2S `INTERNAL_ROUTES` exclusion list (not yet required by S2S, same as edge-deploy/edge-rollback/edge-preview).

Depends on adobe/spacecat-shared#1878 and the new `setApplyStaleForSuggestions` client method (companion PR, to be opened) for the underlying patch/suggestion-data mutation — inert until spacecat-shared releases and this repo bumps its `@adobe/spacecat-shared-tokowaka-client` dependency.

## Test plan
- [x] `npx mocha test/controllers/suggestions.test.js` — 535 passing (14 new tests for `setSuggestionsApplyStale`)
- [x] `npx mocha test/routes/index.test.js test/routes/facs-capabilities.test.js test/routes/required-capabilities.test.js` — 3469 passing
- [x] `npm run lint` on changed files — clean
- [ ] Full `npm test` coverage gate not run locally (90% global threshold) — CI will confirm


MysticatBot (changes requested):
Hey @dhavkuma,

**Verdict: Request changes** - three issues to address before merge.
**Complexity: HIGH** - large diff; new API surface + FACS signal.
_Changes: Adds a new PATCH edge-apply-stale endpoint to toggle the applyStale flag on deployed suggestions without redeploying, and threads applyStale into the existing deploy flow (8 files)._
Note: CI checks are still pending (build + integration tests) - resolve before merge.

## Must fix before merge

1. [Important] Missing OpenAPI path definition for the new PATCH endpoint - `docs/openapi/site-opportunities.yaml` (details inline)
2. [Important] `applyStale` not validated as boolean in the existing `deploySuggestionToEdge` flow - `src/controllers/suggestions.js:1918` (details inline)
3. [Important] OpenAPI description says applyStale is ignored in async experiment mode but the code persists it into GeoExperiment metadata - `docs/openapi/site-opportunities.yaml` / `src/controllers/suggestions.js:2145` (details inline)

<details><summary>Non-blocking (3): minor issues and suggestions</summary>

- nit: Log messages at lines ~377-386 say site: n/a but the site has already been resolved and apexBaseUrl is available - `src/controllers/suggestions.js`
- nit: `context.log.info` serializes the entire 207 response body via JSON.stringify at INFO level; for large batches this produces multi-KB log lines - consider logging only the metadata summary - `src/controllers/suggestions.js:3070`
- nit: `suggestionIds.indexOf` in the ineligibleSuggestions and catch paths is O(n) per item and returns the first occurrence index if duplicates exist; the pre-validation loop uses forEach index correctly but downstream paths do not - `src/controllers/suggestions.js:3049`
</details>

docs/openapi/site-opportunities.yaml
```diff
@@ -629,6 +629,15 @@ site-opportunity-suggestions-edge-deploy:
...
+                default: false
+                description: |
+                  When true, the deployed patch keeps applying even if the source page's content
+                  goes stale (e.g. a missing/outdated Last-Modified header). Mirrored onto each
+                  deployed suggestion's `data.applyStale` for display, and set on the underlying
+                  edge patch itself. Only synchronous (direct) deployment mode is supported;
+                  ignored in async experiment mode.
     responses:
       '207':
         description: |
```
MysticatBot, line 643: issue (blocking): The PR defines applyStale on the existing edge-deploy schema but does not add a path entry for the new PATCH .../edge-apply-stale endpoint itself. API consumers relying on the spec (generated SDKs, docs portals) will not discover this endpoint.

**What:** No OpenAPI path definition for the new PATCH endpoint.
**Why it matters:** Violates the project OpenAPI-first convention and leaves the API contract undocumented.
**How to fix:** Add a site-opportunity-suggestions-edge-apply-stale path entry with the request body schema (suggestionIds: string[], applyStale: boolean) and 207 multi-status response shape, following the pattern of sibling entries.

src/controllers/suggestions.js
```diff
@@ -1915,6 +1916,7 @@ function SuggestionsController(ctx, sqs, env) {
   const deploySuggestionToEdge = async (context) => {
     const siteId = context.params?.siteId;
     const opportunityId = context.params?.opportunityId;
```
MysticatBot, line 1918: issue (blocking): applyStale is extracted from context.data and spread into metadataBase and the deployToEdge call without any type validation. Unlike the new setSuggestionsApplyStale endpoint (which guards with isBoolean(applyStale)), this path accepts any value (string, object, array) and persists it into S3 patch metadata.

**What:** Missing isBoolean guard on applyStale in the deploy flow.
**Why it matters:** A caller with deploy permissions could persist non-boolean values into S3 metadata, violating the schema contract and creating type confusion for downstream consumers.
**How to fix:** Add a validation before using the value:
```js
if (applyStale !== undefined && !isBoolean(applyStale)) {
  return badRequest('applyStale must be a boolean');
}
```

docs/openapi/site-opportunities.yaml
```diff
@@ -629,6 +629,15 @@ site-opportunity-suggestions-edge-deploy:
               name:
                 type: string
                 description: Optional experiment name (defaults to opportunity type). Only used in async experiment mode.
+              applyStale:
+                type: boolean
+                default: false
+                description: |
+                  When true, the deployed patch keeps applying even if the source page's content
+                  goes stale (e.g. a missing/outdated Last-Modified header). Mirrored onto each
```
MysticatBot, line 637: issue (blocking): The description states that applyStale is 'Only synchronous (direct) deployment mode is supported; ignored in async experiment mode.' However, the implementation spreads applyStale into metadataBase unconditionally, and the new tests explicitly verify it persists into GeoExperiment metadata in async mode.

**What:** Documentation and implementation disagree on async-mode behavior.
**Why it matters:** Future maintainers and API consumers will be misled about the flag scope.
**How to fix:** Either (a) guard metadataBase so applyStale is only included in the synchronous path, matching the doc; or (b) update the description to say 'stored in both modes; currently acted upon only in direct deployment mode.'

PR #3053, opened 2026-08-13
feat(ai-visibility): llmo-6585 expose response completeness on prompts/responses

rachel-kittens:
## What & why

First slice of migrating paid-customer Brand Claims ingestion off the BP `shared-all`
sheet onto the native AI Visibility API (LLMO-6585; gap doc LLMO-6586). Consumer:
mystique `app/opportunities/claims_extraction`.

`GET /llmo/ai-visibility/prompts/responses` silently fell back to `briefResponse`
(the excerpt) whenever the per-prompt relation call **failed** or returned no full
response — so a caller could not tell a complete answer from a truncated one. Claims
must never be extracted from an excerpt mistaken for the full answer.

## Change (additive, backwards-compatible)

Per-row fields on `prompts/responses` (legacy `response` value unchanged):

- `responseSource`: `full` | `excerpt` | `none`
- `responseComplete`: boolean
- `relationStatus`: `ok` | `error` | `skipped`  (the `error` case was previously swallowed)
- `date`: execution date passed through from the relation value (was omitted)

Closes the completeness / partial-failure and omitted-date gaps from the LLMO-6586
API-gap doc. The larger whole-brand batch/snapshot feed (the ~500-row scan ceiling,
immutable snapshots, incremental reads) is tracked separately and partly blocked on
the Semrush/gRPC backend.

## Testing

- Unit: `test/support/ai-visibility/handlers/prompts.test.js` — **32 passing** (4 new), eslint clean.
- Live baseline (prod `/api/v1`, domain=adobe.com): confirmed the endpoint returns full
  `response` + `responseExcerpt` today with **no** completeness/date signal, and `total`
  capped at 500 — i.e. exactly the gap these fields fill.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Christopher Wisse", "Dominique Jäggi"]
```


src/support/ai-visibility/handlers/prompts.js
```diff
@@ -19,6 +19,25 @@ import {
   PROMPTS_RESPONSES_PROMPTS_SCAN_LIMIT,
 } from '../grpc-utils.js';
 
+/**
+ * Normalizes a relation `date` into an ISO `YYYY-MM-DD` string. The gRPC relation
+ * value carries `date` as a protobuf Date message (`{ year, month, day }`, plus a
+ * `$typeName` tag), not a scalar — emitting it verbatim leaks that struct to callers.
```
MysticatBot, line 25: issue (blocking): **OpenAPI spec not updated for new response fields.**

The repo CLAUDE.md states under "API Design Principles > Must-Follow Patterns":

> **Specification Sync**: Keep OpenAPI specs and implementation in sync

This PR adds four new fields (`responseSource`, `responseComplete`, `relationStatus`, `date`) to the response shape of `GET /llmo/ai-visibility/prompts/responses` without a corresponding update to `docs/openapi/ai-visibility-api.yaml`. The existing schema (`AiVisibilityPaginatedResponse`) uses `additionalProperties: true` and does not document per-row fields, but this PR extends the gap by introducing new contract surface that consumers (e.g. `claims_extraction`) will depend on.

**Fix:** Add the four fields to the OpenAPI spec (either inline or via a new `AiVisibilityPromptResponseItem` schema), then run `npm run docs:lint` and `npm run docs:build`.

src/support/ai-visibility/handlers/prompts.js
```diff
@@ -39,6 +58,12 @@ export async function handlePromptsResponses(sp, clients) {
   }
   const total = prompts.length;
   const page = prompts.slice(offset, offset + limit);
+  // `attempted[i]` records whether we actually issued the per-prompt relation call.
+  // Without it a skipped prompt (missing identity → Promise.resolve(null)) is
+  // indistinguishable from a relation call that fulfilled with a null value, which
```
MysticatBot: issue (blocking): **Duplicated identity-check logic will diverge and has a subtle bug on serpId=0.**

The new `hasRelationIdentity` predicate:
```js
const hasRelationIdentity = (p) => Boolean(p.promptHash && String(p.serpId ?? '') && p.topicId);
```

duplicates the existing guard inside `page.map`:
```js
if (!promptHash || !serpId || !topicId) { return Promise.resolve(null); }
```

They already disagree on the `serpId=0` edge case: `String(0 ?? '')` = `"0"` (truthy), so `attempted[i]` = `true`, but `!serpId` with serpId=0 is `true`, so the guard short-circuits to `Promise.resolve(null)`. Result: `relationStatus` reports `'ok'` when the call was actually skipped. Protobuf int fields default to 0, so this fires on prompts lacking a serpId.

**Fix:** Extract a single shared predicate and use it in both places:
```js
const canFetchRelation = (p) => Boolean(p.promptHash && p.serpId && p.topicId);
const attempted = page.map(canFetchRelation);
// then in page.map:
if (!canFetchRelation(p)) { return Promise.resolve(null); }
```
This eliminates the string coercion issue and the duplication in one step.

MysticatBot (changes requested):
Hey @rachel-kittens,

**Verdict: Request changes** - two blocking issues around API contract compliance and a logic-divergence bug.
_Changes: Adds per-row response completeness metadata (responseSource, responseComplete, relationStatus, date) to the prompts/responses endpoint so consumers can distinguish full answers from degraded excerpts (2 files)._

## Must fix before merge

1. [Important] OpenAPI spec not updated for new response fields - `docs/openapi/ai-visibility-api.yaml` (details inline)
2. [Important] Duplicated identity-check logic diverges on serpId=0 - `src/support/ai-visibility/handlers/prompts.js:63` (details inline)

<details>

- nit: `responseComplete` name implies content-level quality but only signals source provenance - consider `responseFromRelation` or rely on `responseSource === 'full'` - `src/support/ai-visibility/handlers/prompts.js:101`
- nit: `toIsoDate` passes arbitrary strings without format validation and produces invalid dates from out-of-range protobuf fields (e.g. month=13) - a `/^\d{4}-\d{2}-\d{2}$/` gate on the string branch and a bounds check on the object branch would harden it - `src/support/ai-visibility/handlers/prompts.js:30`
- suggestion: Add a test for the empty-string full response edge case (`relation resolves with response: ''`) to lock in the `responseSource='full', responseComplete=false` behavior - `test/support/ai-visibility/handlers/prompts.test.js`
- suggestion: Add a test for partial protobuf date objects (e.g. `{year: 2026, month: 0, day: 15}`) to confirm `toIsoDate` returns null on zero-valued fields - `test/support/ai-visibility/handlers/prompts.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 53s | <b>Cost:</b> $4.67 | <b>Commit:</b> <code>c6fc44db7c09cc6194f24b025090e227db40f568</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/ai-visibility/handlers/prompts.js
cwjwisse: Fixed in e6836f6c.

Documented the new per-row fields (`responseSource`, `responseComplete`, `relationStatus`, `date`) in the `prompts/responses` OpenAPI operation description, added the ISO-date normalization note to `prompts/responses/latest`, and regenerated `docs/index.html`. `npm run docs:lint` passes.

I kept the shared `AiVisibilityPaginatedResponse` schema instead of introducing a typed item schema: this endpoint intentionally reuses the generic paginated item type (`items: additionalProperties: true`) shared with the sibling AI Visibility list endpoints (topics-research, topics-research-prompts), so I documented the fields at the operation level to stay consistent with that pattern. Happy to switch to a dedicated typed item schema if you'd prefer the stronger contract.

src/support/ai-visibility/handlers/prompts.js
cwjwisse: Deduped in e6836f6c.

The relation-call guard now uses the same `hasRelationIdentity(p)` predicate as `attempted[]`, so there is a single source of truth and the two cannot drift.

On the `serpId=0` point: both the original inline guard and `hasRelationIdentity` coerce via `String(p.serpId ?? '')`, so `serpId=0` becomes `'0'` (truthy) and is treated as present in both - there is no divergence on `0` specifically; only `null`/`undefined`/`''` are treated as missing, identically in both. So the concrete change here is removing the duplication rather than a behavior fix. 33 unit tests pass, including the `skipped`/`ok`/`error` `relationStatus` paths.

cwjwisse:
Pushed `e6836f6c` addressing both blocking findings from the review:

- **OpenAPI spec sync**: documented the four new per-row fields (`responseSource`, `responseComplete`, `relationStatus`, `date`) in the `prompts/responses` operation description, added the ISO-date note to `prompts/responses/latest`, and regenerated `docs/index.html`. `docs:lint` is clean. Kept the shared `AiVisibilityPaginatedResponse` schema (see thread reply for rationale).
- **Duplicated identity check**: unified the relation-call guard and `attempted[]` on a single `hasRelationIdentity` predicate. On the `serpId=0` note — both guards coerce identically via `String(p.serpId ?? '')`, so there is no divergence on `0`; details in the thread reply.

Integration tests for the modified endpoint are deferred (unit tests cover the new field logic; 33 passing), noted in the commit body.

MysticatBot (approved):
Hey @rachel-kittens,

**Verdict: Approve** - prior blocking issues resolved; remaining observations are non-blocking nits.
_Changes: Adds per-row response completeness metadata (responseSource, responseComplete, relationStatus, date) to the prompts/responses endpoint so consumers can distinguish full answers from degraded excerpts (4 files)._

<details>

- nit: toIsoDate passes string values through without format validation - the OpenAPI spec promises ISO YYYY-MM-DD but a malformed upstream string would leak through verbatim. A /^\d{4}-\d{2}-\d{2}$/ regex gate on the string branch would harden the contract. - `src/support/ai-visibility/handlers/prompts.js:30`
- nit: Dense single-line if/else-if/else chain for relationStatus (~120 chars) slows readers down - consider breaking across lines. - `src/support/ai-visibility/handlers/prompts.js:85`
- nit: responseComplete name implies content-level quality but only signals source provenance (full relation response + non-empty). Consumers expecting a completeness guarantee may be surprised. Consider clarifying in docs or renaming to responseFromRelation. - `src/support/ai-visibility/handlers/prompts.js:100`
</details>

- OpenAPI spec now documents all four new fields in the operation description (e6836f6c)
- Identity-check logic deduplicated into shared hasRelationIdentity predicate (e6836f6c)


---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 43s | <b>Cost:</b> $3.93 | <b>Commit:</b> <code>e6836f6c1d34e0750018c9c7366274e1004b5bae</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #3057, opened 2026-08-14
fix(serenity): surface renamed origin__ authorship root in filter dimensions (LLMO-7000)

aliciadriani:
## LLMO-7000

The Origin filter on the Brand Presence / URL Inspector surface offered producing-system values as if they were authorship values, and lost the real authorship values on any project the dimension-root reshape had reached.

### Root cause
`transformOriginsToFilterDimensions` only read the pre-rename `source__` prefix, and `KNOWN_TAG_PREFIXES` omitted `origin__`. After the `source` → `origin` authorship rename:
- real authorship tags (`origin__ai` / `origin__human`) were never surfaced under `origins`, and
- they leaked through the catch-all (`transformOtherTagsForFilterDimensions`) as a stray dynamic `origin` group.

So on a reshaped project the `ai` / `human` options disappeared from the Origin dropdown, while producing-system `source__config` etc. rendered there instead.

### Fix
Mirrors the LLMO-6984 intent-rename fix — both parts required:
1. `transformOriginsToFilterDimensions` dual-reads both roots (`origin` + legacy `source`), preserving the matched prefix when reconstructing `parent_id`.
2. `origin__` added to `KNOWN_TAG_PREFIXES`, so a reshaped project's authorship tags stop resurfacing through the catch-all.

New root-name constants (`ORIGIN_ROOT_NAME`, `LEGACY_ORIGIN_ROOT_NAME`) live in `prompt-tags.js` alongside the intent equivalents.

### Deviation from the ticket: no transition log
The ticket suggested a transition log line "as LLMO-6984 does for intent." Intentionally omitted: the legacy origin spelling `source__` is now **shared** with the new producing-system `source` dimension, so a `source__`-presence log would fire permanently across the entire migrated fleet and never signal when the tolerance can drop — the opposite of intent's clean, self-quieting signal. Documented in the code comments so the follow-up (dropping the `source__` tolerance) is discoverable.

No `project-elmo-ui` change needed — it already carries the tolerant read as `LEGACY_ORIGIN_DIMENSION`. The gap was server-side only.

### Testing
- `topics.test.js` — added dual-root `parent_id`, renamed-root read, mixed authorship+producing-system, no-double-count, and catch-all-claims-`origin` cases
- Full elements sweep (597 tests), `type-check` (base + strict), and `eslint` all pass

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```

### DIFF
--- a/src/support/serenity/prompt-tags.js
+++ b/src/support/serenity/prompt-tags.js
@@ -99,6 +99,33 @@ export const INTENT_ROOT_NAME = '$abv_tags$intent';
 export const LEGACY_INTENT_ROOT_NAME = DIMENSION.INTENT;
 
 /**
+ * The origin (authorship) dimension root name — who authored the prompt. Named
+ * after its dimension key, like the other roots.
+ */
+export const ORIGIN_ROOT_NAME = DIMENSION.ORIGIN;
+
+/**
+ * The pre-rename origin root name. Authorship shipped under a root literally
+ * named `source` before the rename to `origin` (LLMO-6270 §46,
+ * origin-dimension.md). The rename runs project by project from the migration
+ * CLI, so a project can carry either spelling until the sweep completes, and the
+ * Elements read path tolerates both.
+ *
+ * COLLISION — pinned to the literal `source`, NOT to {@link DIMENSION.SOURCE},
+ * even though the two coincide today: this is the *historical* wire spelling of
+ * authorship and must not drift if the producing-system dimension is ever
+ * renamed. The producing-system `source` dimension (LLMO-6270 §47) also emits
+ * `source__` tags, so a `source__` value is ambiguous — pre-rename authorship on
+ * an un-reshaped project, or a producing-system value on a reshaped one — and the
+ * two cannot be told apart at this prefix. The origins read tolerates `source__`
+ * for the un-reshaped case and accepts that producing-system values ride along on
+ * reshaped projects until a dedicated source dimension claims them. Unlike
+ * LLMO-6986's intent tolerance, this one is not cleanly observable from the read
+ * surface (producing-system `source__` masks it), so there is no transition log.
+ */
+export const LEGACY_ORIGIN_ROOT_NAME = 'source';
+
+/**
  * Every name reserved at the root level: the four roots named after their
  * dimension, plus BOTH intent spellings — the upstream `$abv_tags$intent` and
  * the pre-rename `intent`, which stays reserved while any project still carries
--- a/src/support/elements/definitions/topics.js
+++ b/src/support/elements/definitions/topics.js
@@ -11,7 +11,12 @@
  */
 
 import { resolveElementModel, SEP } from '../constants.js';
-import { INTENT_ROOT_NAME, LEGACY_INTENT_ROOT_NAME } from '../../serenity/prompt-tags.js';
+import {
+  INTENT_ROOT_NAME,
+  LEGACY_INTENT_ROOT_NAME,
+  ORIGIN_ROOT_NAME,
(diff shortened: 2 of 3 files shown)


MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - clean, well-tested fix that mirrors the established intent-rename pattern.
_Changes: Surfaces the renamed `origin__` authorship root in filter dimensions alongside the legacy `source__` prefix, fixing the Origin dropdown regression on reshaped projects (3 files)._

Note: CI pipeline checks (branch-deploy, deploy-stage, semantic-release) are failing - these appear to be release-pipeline issues rather than code quality failures. Build and IT checks are still pending.

<details>

- nit: The block comment on `ORIGIN_ROOT_NAMES` (topics.js:168-185) largely duplicates the JSDoc on `LEGACY_ORIGIN_ROOT_NAME` in prompt-tags.js. Consider trimming to a forward reference ("See LEGACY_ORIGIN_ROOT_NAME for the collision semantics.") to avoid two sources of truth for the same design rationale. - `src/support/elements/definitions/topics.js:168`
- suggestion: The intent tolerance has a removal signal via structured log. A similar scoped log in `transformOriginsToFilterDimensions` when the legacy prefix matches would make the eventual cleanup self-evident. The PR body explains why a naive log is permanently noisy here (shared prefix with producing-system), but a conditional signal (e.g. only when a project carries both `origin__` and `source__` authorship values) could still provide actionable signal without the noise. - `src/support/elements/definitions/topics.js:190`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 9m 9s | <b>Cost:</b> $4.89 | <b>Commit:</b> <code>ac356625394c8d59ad0e2e4d1e93c6933c7cbd54</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/serenity/prompt-tags.js
```diff
@@ -98,6 +98,33 @@ export const INTENT_ROOT_NAME = '$abv_tags$intent';
...
+ * renamed. The producing-system `source` dimension (LLMO-6270 §47) also emits
+ * `source__` tags, so a `source__` value is ambiguous — pre-rename authorship on
+ * an un-reshaped project, or a producing-system value on a reshaped one — and the
+ * two cannot be told apart at this prefix. The origins read tolerates `source__`
+ * for the un-reshaped case and accepts that producing-system values ride along on
+ * reshaped projects until a dedicated source dimension claims them. Unlike
+ * LLMO-6986's intent tolerance, this one is not cleanly observable from the read
+ * surface (producing-system `source__` masks it), so there is no transition log.
+ */
+export const LEGACY_ORIGIN_ROOT_NAME = 'source';
```
aliciadriani, line 126: **Nit (review):** `LEGACY_ORIGIN_ROOT_NAME` is a bare literal `'source'`, while its intent sibling uses `DIMENSION.INTENT`. This is deliberate — it's the *historical* authorship wire spelling and must not drift if `DIMENSION.SOURCE` is later renamed. Flagging so a future refactor doesn't "tidy" it into `DIMENSION.SOURCE`.

aliciadriani: Resolved — no code change needed. The constant's own doc block already pins this explicitly: "pinned to the literal `source`, NOT to `DIMENSION.SOURCE`, even though the two coincide today … must not drift if the producing-system dimension is ever renamed." So the guard against a future "tidy" refactor is already in the code.

PR #3059, opened 2026-08-14
refactor(serenity): stop populating brands.pending_semrush_provisioning (SITES-49448)

aliciadriani:
## SITES-49448 — step 1 of a sequenced removal

Retires the `brands.pending_semrush_provisioning` "Save as pending" staging blob, starting with the two api-service write paths.

### Why this is safe now
- **No activation path uses it.** `pending → active` is sub-workspace-only (LLMO-6405) and *ignores* the stashed `primaryUrl`/`markets`; it only reads-then-clears the blob.
- **elmo is already clean.** The draft Markets-tab UI that read/mutated the blob (`PendingMarketsView` / `PendingEditModelsDialog`) is gone from `project-elmo-ui` main — no reference to `pendingSemrushProvisioning` anywhere in its source. api-service *constructs* the blob internally from normal create fields; elmo never POSTs or reads it.
- **Prod is effectively drained.** Live check (2026-08-07): 48 rows carry the blob (10 pending / 14 active / 24 deleted); no row outside `deleted` has a non-empty `markets` array — the survivors are a `{primaryUrl, generatePrompts, markets: []}` shell.

### What this PR changes (api-service only)
- `createBrandForOrg` no longer stashes the blob for a pending draft — it was the last create-time writer. The draft is still saved with no anchor (lands `pending`), exactly as before.
- `updateBrandForOrg` now strips a client-supplied `pendingSemrushProvisioning` **unconditionally** (was pending-only), so no PATCH can (re)populate the column on any brand.

### Deliberately backward-compatible
The activate **read/clear**, the DTO field, and the spacecat-shared attribute are left in place, so the ~24 legacy prod rows drain to `null` as they activate. Nothing here depends on merge ordering with other repos.

### Follow-ups (tracked, not in this PR)
1. api-service: remove the activate read + DTO field once prod shows no live row carries a non-empty blob.
2. spacecat-shared: drop the `pendingSemrushProvisioning` attribute (after #1).
3. mysticat-data-service: drop the `pending_semrush_provisioning` column + CHECK (last).

### Testing
- `brands.test.js` — create/update assertions flipped to the no-stash behavior; dropped 3 tests that only exercised deleted stash-construction internals (primaryURL resolution, modelIds seeding). 431 pass.
- `serenity.test.js` (438) + `brands-storage.test.js` — unchanged read/storage paths still green.
- `type-check` (base + strict) and `eslint` clean.

### DIFF
--- a/src/controllers/brands.js
+++ b/src/controllers/brands.js
@@ -1682,46 +1682,16 @@ function BrandsController(ctx, log, env) {
           return badRequest('market and languageCode are required when generatePrompts is true');
         }
         if (isPendingBrand) {
-          // Defer provisioning: persist the chosen (market, languageCode) AND
-          // the primary URL (if the user entered one before saving as pending)
-          // on the brand, so activation can provision the real sub-workspace +
-          // project later (stored in brands.pending_semrush_provisioning). The primary
-          // URL otherwise lives only on the Semrush side, so a site-less draft
-          // would have nowhere to keep it. The row lands as 'pending' because it
-          // has no anchor (no site_id, no semrush_sub_workspace_id) — see
+          // A pending (draft) brand defers ALL Semrush provisioning — no
+          // sub-workspace, no project. Its markets are added later from the Markets
+          // tab, and pending→active activation is sub-workspace-only (LLMO-6405), so
+          // nothing needs to be stashed at create time. The row lands as 'pending'
+          // because it has no anchor (no site_id, no semrush_sub_workspace_id) — see
           // upsertBrand's anchor check.
-          const primaryUrl = (Array.isArray(brandData.urls) ? brandData.urls : [])
-            .map((u) => (typeof u === 'string' ? u : u?.value))
-            .find(hasText) || null;
-          // AI models (LLMs) the wizard collected. Unlike the direct-provision
-          // path they are NOT required here — a draft can be saved before the
-          // user picks any, and they can be edited per-market later from the
-          // Markets tab. Seed the initial market's modelIds with them when
-          // present so activation applies them; omit the key entirely when none
-          // were chosen (mirrors normalizePendingSemrushProvisioning).
-          const seedModelIds = Array.isArray(brandData.semrushModelIds)
-            ? brandData.semrushModelIds.filter(hasText)
-            : [];
-          // A no-prompt draft may carry NO market at all (location/language are
-          // optional then) — stash only the market actually picked. The activate
-          // flow already handles 0..N stashed markets: an empty list + a primary
-          // URL provisions a single US/EN fallback project, and an empty list +
-          // no URL provisions a sub-workspace-only brand.
-          // TODO: the wizard creates at most one market today; if multi-market
-          // draft creation is added, build this array from all selected markets.
-          const markets = [];
-          if (hasSemrushMarket) {
-            const initialMarket = { market, languageCode };
-            if (seedModelIds.length > 0) {
-              initialMarket.modelIds = seedModelIds;
-            }
-            markets.push(initialMarket);
-          }
-          brandData.pendingSemrushProvisioning = {
-            primaryUrl,
-            markets,
-            generatePrompts,
-          };
+          //
+          // SITES-49448: brands.pending_semrush_provisioning is being retired — this
+          // was its last create-time writer. The staging blob no longer serves any
+          // activation path, so the draft is saved without one.
         } else if (hasSemrushMarket) {
           const brandDomain = brandDomainFromPayload(brandData);
           if (!brandDomain || !hasText(brandDomain)) {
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - clean, well-sequenced first step in retiring the staging blob with no correctness or security concerns.
**Complexity: MEDIUM** - medium diff, single-service, API surface signal.
_Changes: Removes the two write paths (create-time stash and conditional PATCH guard) that populated `brands.pending_semrush_provisioning`, replacing both with unconditional field deletion (2 files)._

<details>

- nit: The `if (isPendingBrand)` block now contains only comments (no executable statements). It functions correctly as a guard to skip the `else if` provisioning path, but a comment-only branch is unusual. Consider collapsing into `if (!isPendingBrand && hasSemrushMarket)` or leaving a one-liner noting the no-op. Defensible as-is if the branch is removed in a near-term follow-up. - `src/controllers/brands.js:1685`
- nit: The PATCH tests ("strips even when brand IS pending" and "strips when NOT pending") still set up `postgrestClient.from('brands').select('status')` mocks that the production code no longer calls. The tests pass but the mocks imply a DB read that no longer happens, which could mislead future readers. Consider removing the dead mock or adding a `.notCalled` assertion on it to document that no status read occurs. - `test/controllers/brands.test.js:~6395`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 35s | <b>Cost:</b> $4.41 | <b>Commit:</b> <code>ace29124bc61ffe653819de488fc652ce5dcd947</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/controllers/brands.js
```diff
@@ -2508,6 +2461,13 @@ function BrandsController(ctx, log, env) {
       // usually baseUrl); a pending brand resolves from its stashed Semrush primary URL.
       // We deliberately do NOT fall back to urls[] — that's the brand's listed URLs,
       // not a declared activation anchor, so guessing one would be wrong.
+      //
+      // SITES-49448: pendingSemrushProvisioning is no longer written at create, so this
+      // stash-primaryUrl fallback now serves ONLY legacy pending rows during drain. A new
```
aliciadriani, line 2466: **Must-confirm (review):** this PR removes the only writer of `pendingSemrushProvisioning`, yet `activateBrandForOrg` still uses `pendingSemrushProvisioning?.primaryUrl` as the site-resolution anchor for a site-less pending brand (and deliberately refuses `urls[]`). Confirm a *new* pending brand cannot reach this fallback and 400 with "Brand has no onboarded primary site".

test/controllers/brands.test.js
```diff
@@ -5620,7 +5620,7 @@ describe('Brands Controller', () => {
         expect(upsertStub.firstCall.args[0].semrushSubWorkspaceId).to.equal(null);
       });
 
-      it('saves a pending draft WITHOUT a primary URL: no provisioning, market stashed (201)', async () => {
+      it('saves a pending draft: defers provisioning, writes no stash, no anchor (201)', async () => {
```
aliciadriani, line 5623: **Should (review):** with the create-time stash gone, add/point to coverage for the *new-model* activation shape — a pending brand anchored by `baseSiteId` with no stash activates cleanly — so the intended post-migration path is pinned, not just the legacy stash-resolve path (test at ~8440).

src/controllers/brands.js
```diff
@@ -1682,46 +1682,16 @@ function BrandsController(ctx, log, env) {
           return badRequest('market and languageCode are required when generatePrompts is true');
         }
         if (isPendingBrand) {
```
aliciadriani, line 1684: **Nit (review):** the `if (isPendingBrand) {}` block is now empty (comment-only). Consider whether it can fold into the `else if` chain for readability.

src/controllers/brands.js
aliciadriani: Addressed in 0650832f — added the clarifying comment right here: the stash fallback is now **legacy-drain-only**. A new pending brand anchors via `baseSiteId` (the caller's site selection) or activates through `/serenity/activate` (sub-workspace-anchored, needs no site); neither reaches this branch. Confirmed not reachable via elmo — its `main` has no `baseSiteId`, no `/activate`, and no brand-create path literals, so the live serenity activation goes through `/serenity/activate`. The anchor-less → 400 is already the codified contract (test 8480). The fallback itself is deleted with the step-2 read-path cleanup once no live row carries a non-empty blob.

test/controllers/brands.test.js
aliciadriani: Already covered — existing activate tests pin the new-model success shape: a pending brand with `baseSiteId: SITE_ID` and **no** stash activates to 200 (e.g. "skips both prompt-gen and schedule when generatePrompts:false and no prompts exist", "reuses an in-flight RUNNING/QUEUED job"). The intentional anchor-less 400 is pinned by test 8480, and the legacy stash-resolve path by test 8440. No new test needed.

src/controllers/brands.js
aliciadriani: Keeping as-is — the empty `if (isPendingBrand)` branch is a **guard**, not dead code. It short-circuits a pending draft before the `else if (hasSemrushMarket)` / `else` provisioning branches, which must never run for a draft. Folding it into the chain would let a pending brand fall through to provisioning, so the empty (comment-documented) branch stays.

MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - clean second pass; prior nits addressed or justified, no new concerns from the 3 comment-only commits.
**Complexity: MEDIUM** - medium diff, single-service, API surface signal.
_Changes: Removes the two write paths that populated `brands.pending_semrush_provisioning` (create-time stash construction and conditional PATCH guard), replacing both with unconditional field deletion (2 files)._

<details>

- nit: Dead `postgrestClient.from('brands').select('status')` mock setup remains in two PATCH tests (~line 6398 "strips even when pending" and ~line 6435 "strips when flipping to active"). The production code now does an unconditional `delete` with no DB call, so the mock is never invoked. Consider removing it or adding a `.to.not.have.been.called` assertion to document the no-read contract. - `test/controllers/brands.test.js:~6398`
- suggestion: Consider adding a date-gated `log.warn` in the activate path if a non-null blob is still encountered during drain, so the "no live row carries a non-empty blob" follow-up condition becomes machine-observable rather than requiring manual prod queries. - `src/controllers/brands.js:~2467`

</details>

- Empty `if (isPendingBrand)` guard block: author justified as preventing fall-through to provisioning branches. Valid - the block is a semantic guard, not dead code.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 26s | <b>Cost:</b> $4.11 | <b>Commit:</b> <code>ad0d4dd7a72d1194ceabc49f772db8ac11ea8ff7</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #3060, opened 2026-08-14
fix(plg): auto-correct OTHER delivery type instead of alerting on mis…

anshulk-public:
## Summary
- During PLG onboarding, when a site's stored delivery type is a placeholder value (`other`) and a subsequent scan detects a real delivery type, the flow now auto-corrects the site's delivery type instead of just posting a Slack mismatch alert.
- The Slack "Delivery Type Mismatch" notification is now only sent when the *stored* delivery type is a confirmed, non-`other` value that disagrees with detection — i.e. a case that actually warrants human review.

## Why
Example real-world case (Hugo Boss AG, site `f5a28dcb-9e08-49a2-add3-92521f4c2825`, `hugoboss.com`):
- Stored delivery type: `other`
- Detected delivery type: `aem_ams`

`other` isn't a confirmed value — it's a placeholder set when we couldn't yet tell what the site was. Alerting on every "other → detected X" case was generating noise for something safe to auto-fix.

## Changes
- `src/controllers/plg/plg-onboarding/onboarding-flow.js`: in Step 5f (post-detection delivery-type comparison), when `currentDeliveryType === SiteModel.DELIVERY_TYPES.OTHER`, call `site.setDeliveryType(detectedDeliveryType)` directly and skip the Slack notification path. All other mismatches (stored is a real type but disagrees with detection) still alert as before.
- `test/controllers/plg/plg-onboarding/site-setup.test.js`: added coverage for auto-correction (non-AMS and AMS detected values) confirming `setDeliveryType` is called and no Slack message is sent, alongside existing mismatch-alert tests (unchanged, still passing).

## Related Issues

Thanks for contributing!

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```

### DIFF
--- a/src/controllers/plg/plg-onboarding/onboarding-flow.js
+++ b/src/controllers/plg/plg-onboarding/onboarding-flow.js
@@ -750,22 +750,29 @@ export async function performAsoPlgOnboarding({
         && detectedDeliveryType !== SiteModel.DELIVERY_TYPES.OTHER
         && detectedDeliveryType !== currentDeliveryType
       ) {
-        log.warn(`Delivery type mismatch for site ${site.getId()} (${baseURL}): stored=${currentDeliveryType} detected=${detectedDeliveryType}`);
-        const channelId = env.SLACK_PLG_ONBOARDING_CHANNEL_ID;
-        const token = env.SLACK_BOT_TOKEN;
-        /* c8 ignore next */
-        if (channelId && token) {
-          const message = ':warning: *PLG Onboarding — Delivery Type Mismatch*\n\n'
-            + `• *Site ID:* \`${site.getId()}\`\n`
-            + `• *Domain:* \`${baseURL}\`\n`
-            + `• *Org ID:* \`${organizationId}\`\n`
-            + `• *Org:* ${organization.getName()} (\`${imsOrgId}\`)\n`
-            + `• *Stored delivery type:* \`${currentDeliveryType}\`\n`
-            + `• *Detected delivery type:* \`${detectedDeliveryType}\``;
-          try {
-            await context.postSlackMessage(channelId, message, token);
-          } catch (err) {
-            log.error(`Failed to post delivery type mismatch alert: ${err.message}`);
+        if (currentDeliveryType === SiteModel.DELIVERY_TYPES.OTHER) {
+          // Stored type was a placeholder ("other") and detection found a confident
+          // signal — safe to auto-correct without a human review step.
+          log.info(`Auto-correcting delivery type for site ${site.getId()} (${baseURL}): stored=${currentDeliveryType} detected=${detectedDeliveryType}`);
+          site.setDeliveryType(detectedDeliveryType);
+        } else {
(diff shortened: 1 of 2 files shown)


src/controllers/plg/plg-onboarding/onboarding-flow.js
```diff
@@ -750,22 +750,29 @@ export async function performAsoPlgOnboarding({
...
-            + `• *Org:* ${organization.getName()} (\`${imsOrgId}\`)\n`
-            + `• *Stored delivery type:* \`${currentDeliveryType}\`\n`
-            + `• *Detected delivery type:* \`${detectedDeliveryType}\``;
-          try {
-            await context.postSlackMessage(channelId, message, token);
-          } catch (err) {
-            log.error(`Failed to post delivery type mismatch alert: ${err.message}`);
+        if (currentDeliveryType === SiteModel.DELIVERY_TYPES.OTHER) {
+          // Stored type was a placeholder ("other") and detection found a confident
+          // signal — safe to auto-correct without a human review step.
```
MysticatBot, line 755: question: `site.setDeliveryType(detectedDeliveryType)` mutates the object in memory. Is there a downstream `site.save()` (or equivalent flush) later in `performAsoPlgOnboarding` that persists this change? The existing flow pattern suggests accumulate-then-save, but the tests only assert the setter call, not persistence. If the save is there, this is fine as-is; if not, the correction is ephemeral and will re-trigger on every subsequent onboarding run.

MysticatBot (approved):
Hey @anshulk-public,

**Verdict: Approve** - clean logic change with good test coverage; one verification question inline.
**Complexity: MEDIUM** - controller path (API surface signal).
_Changes: Auto-corrects sites with placeholder "other" delivery type during PLG onboarding instead of sending a Slack mismatch alert, reserving alerts for confirmed-value conflicts (2 files)._

<details>

- nit: New tests verify `setDeliveryType` is called but do not assert that `site.save()` (or equivalent persistence) is called afterward. Since this path introduces a data write (not just a notification), adding a save assertion would guard against future regressions that remove or move the downstream persist call. - `test/controllers/plg/plg-onboarding/site-setup.test.js:555`
- suggestion: Consider validating `detectedDeliveryType` against `SiteModel.DELIVERY_TYPES` values before writing it to the site record. The outer guard excludes `OTHER` but does not reject unknown strings from the scan result. Low exploitability (internal scan), but good defense-in-depth. - `src/controllers/plg/plg-onboarding/onboarding-flow.js:755`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 19s | <b>Cost:</b> $3.75 | <b>Commit:</b> <code>cc4b82555bb76161c728f3092b9b455fd1197b4d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/controllers/plg/plg-onboarding/onboarding-flow.js
anshulk-public: Yes save is there after this step

PR #3064, opened 2026-08-14
feat(ai-visibility): flag-gated per-brand Semrush auth seam skeleton (inert)

cwjwisse:
## 1. Abstract
Adds a flag-gated, provider-agnostic auth-seam skeleton for resolving the Spacecat → Semrush AI Visibility credential per brand. Inert by default — no behavior change until a later PR injects a real provider.

## 2. Reasoning
Brand Claims and other server-side consumers need to read a customer's Semrush AI Visibility data unattended (no IMS user token), but Semrush authorizes by workspace membership and has no S2S path today — an unattended service token is rejected (LLMO-6836). The eventual fix routes a per-brand technical-account credential, but the exact shape (single shared vs per-org technical account; `workspaceId` vs `ims_org_id` mapping) is still being negotiated. This PR lands the decision-independent seam now, so the contested choice becomes a small provider swap later rather than transport surgery.

- New `semrush-credential-resolver.js`: `resolveSemrushCredential(brand, env)` (returns a credential descriptor or `null`), a pluggable `setSemrushCredentialProvider` (default: none), and `getCachedToken`, a per-credential TTL token cache (bounded, FIFO by mint time, injectable clock).
- `grpc-transport.js` wiring behind `AI_VISIBILITY_PER_BRAND_AUTH_ENABLED` (default **off**):
  - **Flag off (default):** unchanged — the process-wide singleton transport, authenticating with the shared `client_credentials` machine credential minted per request. Byte-identical to today.
  - **Flag on:** resolves the credential per request and keys a bounded per-credential transport pool; when the resolver returns `null` it falls back to the shared credential. Resolved keys are namespaced so a provider key can never collide with the shared sentinel.
- `getGrpcClients` gains an optional `brand` argument (additive; no caller changes).

Behavior delta: **none** in any current environment. The seam is double-gated — no provider is configured, so even with the flag on the resolver returns `null` and the shared credential is still used. Turning the flag on alone is a no-op; PR-3b injects the real provider and threads `brand`.

## 4. Required information
- Jira / issue: [LLMO-7028](https://jira.corp.adobe.com/browse/LLMO-7028) (sub-task of [LLMO-6585](https://jira.corp.adobe.com/browse/LLMO-6585); blocker [LLMO-6836](https://jira.corp.adobe.com/browse/LLMO-6836))
- Spec: adobe/mysticat-architecture#248 — `products/llmo/spec-brand-claims-semrush-ingestion-feed.md` §6.2

## 7. Test plan
(a) Local: unit tests exercise the flag on/off wiring, the null-provider fallback to the shared credential, per-credential token-cache reuse within TTL and re-mint after expiry, the malformed-mint guard, the bounded client-pool eviction + transport rebuild, and confirm the existing `grpc-transport` suite is unchanged. Confirmed flag-off produces the same transport options and client set as before, and never touches the resolver or token cache.

(b) Per environment: nothing to verify — the flag defaults off everywhere and the change is behavior-preserving. The existing `GET /llmo/ai-visibility/*` responses are unchanged. When PR-3b is ready, enabling `AI_VISIBILITY_PER_BRAND_AUTH_ENABLED` in dev (with a provider installed) is the first live check; until then the flag is inert.

- adobe/mysticat-architecture#248 — related (design of record, §6.2); docs-only, no deploy coupling.
- LLMO-7029 (PR-3b, live provider wiring) stacks on this and is gated on the contested technical-account decision. This PR must merge first.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Rachel Dong", "Dominique Jäggi"]
```

### DIFF
--- a/src/support/ai-visibility/grpc-transport.js
+++ b/src/support/ai-visibility/grpc-transport.js
@@ -23,6 +23,10 @@ import {
 } from '@quazar/ai-seo-ts/ai-cr/service_pb.js';
 import { Sources as VoSources } from '@quazar/ai-seo-ts/ai-vo/service_pb.js';
 import { Relations } from '@quazar/ai-seo-ts/ai-pr/service_pb.js';
+import {
+  resolveSemrushCredential,
+  getCachedToken,
+} from './semrush-credential-resolver.js';
 
 const DEFAULT_SCOPES = 'ai-seo.meta ai-seo.topics ai-seo.prompts ai-seo.sources ai-seo.brand-metrics ai-seo.relations ai-seo.competitors-metrics ai-seo.competitor';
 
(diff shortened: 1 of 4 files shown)


MysticatBot (approved):
Hey @cwjwisse,

**Verdict: Approve** - well-structured inert skeleton with conservative flag gating; hardening notes below for when the feature goes live.
**Complexity: HIGH** - large diff (717 added lines across 4 files).
_Changes: Adds a flag-gated, provider-agnostic auth seam for resolving per-brand Semrush credentials, with bounded token and transport caches, inert by default (4 files)._

<details>

- suggestion: `getCachedToken` has a thundering-herd window - concurrent callers for the same key both see a miss, both call `mintFn()`, both write to cache. Add a singleflight/coalescing pattern (in-flight promise map) before the feature goes live in PR-3b. - `src/support/ai-visibility/semrush-credential-resolver.js:112`
- suggestion: Both caches use FIFO eviction (the code explicitly notes "this is not LRU"). Under steady-state with many brands, high-traffic credentials minted early will be evicted first. A promote-on-hit pattern (`delete` + `set` on cache hit) converts to LRU with a one-line change per cache. Worth doing before the pool sees real traffic. - `src/support/ai-visibility/semrush-credential-resolver.js:107`
- suggestion: No single-key invalidation export exists - only `resetSemrushCredentialCache()` which clears everything. When a real provider returns a bad token (401 from Semrush), callers will need `invalidateCachedToken(key)` to surgically clear one entry without flushing the entire cache. Cheap to add now. - `src/support/ai-visibility/semrush-credential-resolver.js:151`
- suggestion: `resolveSemrushCredential` is synchronous. If PR-3b's provider needs an async lookup (e.g., DB or remote config), the contract changes at every call site. Document whether async resolution is anticipated so the follow-up PR author knows the boundary. - `src/support/ai-visibility/semrush-credential-resolver.js:87`
- nit: When `credential` is null on the flag-on path, the interceptor silently falls back to `getAccessToken(env)` - indistinguishable from flag-off. Consider a structured log line in the fallback branch so operators can tell whether per-brand resolution is working or silently falling through. - `src/support/ai-visibility/grpc-transport.js:141`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 42s | <b>Cost:</b> $6.10 | <b>Commit:</b> <code>3707ce0972d79efeecc94d9209c948e751b5e475</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rachel-kittens (approved):

PR #3065, opened 2026-08-17
feat(llmo): resolve canonical host (redirect-follow + PSL) for Cloudflare worker + route

ssilare-adobe:
Cloudflare worker/route onboarding needs the customer's real serving host (apex vs. `www`), and guessing it wrong breaks onboarding (e.g. `gentingsingapore.com` only serves from the apex, so forcing `www` failed). This PR resolves the host properly: start from `calculateForwardedHost`'s guess, correct it for compound-TLD apexes (`.co.uk`, `.com.au`) via the Public Suffix List, then confirm by fetching the candidate and following its redirect chain — landing back on the site's own apex/`www` wins, an unreachable probe or an off-domain landing falls back to the guess. The resolved host is used for both the worker's `targetHost` and, when the client's route pattern targets the site's root host, the route's host (with the pattern's scheme/path preserved exactly).

## Cases tried (verified against the live domains)

| Domain | `calculateForwardedHost` guess | PSL correction applied? | Live probe result | Resolved host |
|---|---|---|---|---|
| `gentingsingapore.com` | `www.gentingsingapore.com` | No (1 dot, correctly classified) | `www.gentingsingapore.com` → 301 → apex | **`gentingsingapore.com`** (the original onboarding bug this PR fixes) |
| `racq.com.au` | `racq.com.au` (misclassified as a subdomain — 2 dots) | Yes → `www.racq.com.au` | `www.racq.com.au` → 200 directly | **`www.racq.com.au`** |
| `google.co.uk` | `google.co.uk` (misclassified as a subdomain — 2 dots) | Yes → `www.google.co.uk` | `www.google.co.uk` → 301 → `www.google.com` (off-domain, rejected) | **`www.google.co.uk`** (candidate kept — never points at a domain we don't own) |

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Akash Bhardwaj", "Dominique Jäggi"]
```


src/support/edge-routing-utils.js
```diff
@@ -70,6 +70,65 @@ export function getHostnameWithoutWww(url, log) {
...
+ * Whether a hostname has any DNS record (CNAME, A, or AAAA) — i.e. it actually resolves.
+ * Never throws: a failed lookup (NXDOMAIN, SERVFAIL, timeout, …) is treated as "does not resolve"
+ * and returns false, mirroring the tolerant DNS handling used elsewhere in this module.
+ *
+ * @param {string} host - Hostname to resolve (e.g. 'www.example.com').
+ * @param {object} [log] - Logger.
+ * @returns {Promise<boolean>} True if any record type resolves, false otherwise.
+ */
+export async function hostResolves(host, log) {
+  const [cnames, ipv4, ipv6] = await Promise.all([
```
MysticatBot, line 83: issue (blocking): `hostResolves` fires three parallel DNS queries with no explicit timeout. Node `dns.promises` inherits the OS resolver config (often 5s per attempt with multiple retries per nameserver), so a fully-unresponsive authoritative NS can stall this call for 30+ seconds. Since this runs in the request path of `POST /sites/:siteId/llmo/cdn-onboard/cloudflare/deploy`, a slow lookup makes the endpoint appear hung to the caller.

Every other external I/O call in this module is timeout-guarded. A `Promise.race` with a ceiling (3-5s total) that treats timeout as "does not resolve" would keep behavior consistent and surface the fallback path faster:

```js
const RESOLVE_TIMEOUT_MS = 5000;
const results = await Promise.race([
  Promise.all([
    dns.resolveCname(host).catch(() => []),
    dns.resolve4(host).catch(() => []),
    dns.resolve6(host).catch(() => []),
  ]),
  new Promise((resolve) => setTimeout(() => resolve([[], [], []]), RESOLVE_TIMEOUT_MS)),
]);
const [cnames, ipv4, ipv6] = results;
```

This keeps the tolerant "treat timeout as no records" semantics the function already uses for rejections.

MysticatBot (changes requested):
Hey @ssilare-adobe,

**Verdict: Request changes** - one blocking concern about unbounded DNS latency in the deploy request path.
_Changes: Adds DNS verification to Cloudflare worker targetHost derivation so apex-only sites fall back to the apex instead of pointing the worker at a nonexistent www host (4 files)._

## Must fix before merge

1. [Important] DNS resolution has no timeout, can stall the deploy endpoint for 30-60s under adverse conditions - `src/support/edge-routing-utils.js:83` (details inline)

<details>

- nit: Return value casing is inconsistent - early-return path preserves original case from calculateForwardedHost, fallback path returns lowercased originalHost - `src/support/edge-routing-utils.js:116`
- nit: `log.info(...)` in resolveCanonicalHost fallback path uses direct access while hostResolves uses `log?.info(...)` - inconsistent optional handling despite log being documented as required - `src/support/edge-routing-utils.js:121`
- suggestion: Default mockResolveCanonicalHost stub is synchronous but the real function is async - use `async (baseURL, l) => calculateForwardedHost(baseURL, l)` for contract fidelity - `test/controllers/llmo/llmo-cloudflare.test.js:96`
- suggestion: Demote the per-host DNS result log in hostResolves from info to debug - in steady state (www resolves fine) this adds noise to every deploy with no actionable signal - `src/support/edge-routing-utils.js:87`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 15s | <b>Cost:</b> $3.60 | <b>Commit:</b> <code>29414be84d2fc9d225985e2baa196725c1cf104f</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (dismissed):
Hey @ssilare-adobe,

**Verdict: Request changes** - the DNS timeout gap flagged in the prior review remains unaddressed.
_Changes: Adds redirect-follow and DNS verification to Cloudflare worker targetHost derivation + route-pattern host normalization so apex-only sites route correctly (7 files)._
Note: Recommend a human read before merge - this change modifies a shared contract (OpenAPI spec). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge

1. [Important] DNS resolution has no timeout, can stall deploy/route endpoints for 30+ seconds under adverse conditions - `src/support/edge-routing-utils.js:82` (details inline)

<details>

- nit: Return value casing inconsistent across resolveCanonicalHost paths - early return preserves original case from calculateForwardedHost, fallback returns lowercased originalHost. Normalize with toLowerCase on the early return. - `src/support/edge-routing-utils.js:156`
- nit: `log.info(...)` without optional chaining in resolveCanonicalHost fallback path, while hostResolves and probeServingHost both use `log?.info(...)` - `src/support/edge-routing-utils.js:165`
- suggestion: Demote per-host DNS result log in hostResolves from info to debug - in steady state this fires on every deploy/route call with no actionable signal - `src/support/edge-routing-utils.js:88`
- suggestion: Use `new URL(location, baseURL)` for Location header parsing to handle protocol-relative URLs per RFC 3986 - `src/support/edge-routing-utils.js:125`
- nit: Default mockResolveCanonicalHost should be async to match production contract fidelity - `test/controllers/llmo/llmo-cloudflare.test.js:98`

</details>

src/support/edge-routing-utils.js
```diff
@@ -70,6 +70,141 @@ export function getHostnameWithoutWww(url, log) {
...
```
MysticatBot, line 83: issue (blocking): hostResolves fires three parallel DNS queries (resolveCname, resolve4, resolve6) with no explicit timeout. Node dns.promises inherits the OS resolver config (often 5s per attempt with multiple retries), so a fully-unresponsive authoritative NS can stall this call for 30+ seconds. Since this runs in the request path of both POST .../deploy and POST .../routes, a slow lookup makes the endpoint appear hung to the caller.

Every other external I/O call in this module is timeout-guarded (probeServingHost uses AbortSignal.timeout(PROBE_TIMEOUT_MS)). A Promise.race with a ceiling that treats timeout as 'does not resolve' would keep behavior consistent and bound the worst-case latency of the endpoint.

src/support/edge-routing-utils.js
```diff
@@ -70,6 +70,141 @@ export function getHostnameWithoutWww(url, log) {
...
+  }
+
+  const location = res.headers.get('location');
+  if (!location) {
+    return null;
+  }
+
+  let redirectHost;
+  try {
+    redirectHost = new URL(location.startsWith('http') ? location : `https://${location}`)
```
ABHA61, line 140: [P1] Resolve relative redirects against the probed URL

Using the actual tracingFetch User-Agent, https://gentingsingapore.com returns 301 with Location: /en/home. This construction parses that value as https:///en/home, reads en as the hostname, and treats it as an off-domain redirect. The resolver then falls through to DNS and selects www.gentingsingapore.com even though www redirects back to the apex. Please use new URL(location, baseURL) so absolute, protocol-relative, and relative redirects resolve correctly.

src/support/edge-routing-utils.js
```diff
@@ -70,6 +70,141 @@ export function getHostnameWithoutWww(url, log) {
...
+ * @returns {Promise<string>} The canonical origin host.
+ * @throws {Error} If the base URL is unparseable (propagated from calculateForwardedHost).
+ */
+export async function resolveCanonicalHost(baseURL, log) {
+  const candidate = calculateForwardedHost(baseURL, log);
+  const originalHost = new URL(baseURL).hostname.toLowerCase();
+
+  // calculateForwardedHost only rewrites a bare apex into www.<apex>. When it returns the original
+  // host unchanged (already www, or a subdomain) the candidate is authoritative — skip all I/O.
+  if (candidate.toLowerCase() === originalHost) {
```
ABHA61, line 185: [P1] Resolve registrable apexes using the Public Suffix List

With the actual stored base URL https://racq.com.au/, calculateForwardedHost returns racq.com.au because its dot-count heuristic mistakes multi-part public suffix apexes such as .com.au for subdomains. This early return skips the live 301 redirect to www.racq.com.au, so both targetHost and normalized root routes use the wrong host. Please identify apex/www root variants using PSL-based registrable-domain logic rather than candidate equality.

src/support/edge-routing-utils.js
```diff
@@ -70,6 +70,141 @@ export function getHostnameWithoutWww(url, log) {
...
+    return candidate;
+  }
+
+  // candidate is a www host synthesized from the bare apex `originalHost`.
+  const served = await probeServingHost(baseURL, originalHost, log);
```
ABHA61, line 190: [P1] Probe and follow the preferred www candidate

The intended flow is to prefer www, request it, and use the apex when www redirects there. This implementation probes only the base apex and, when that result is inconclusive, checks only whether www has DNS; it never requests or follows www. For Genting, www.gentingsingapore.com redirects to the apex. Please probe the preferred candidate and follow same-domain redirects instead of treating DNS existence as canonical-host evidence.

src/support/edge-routing-utils.js
```diff
@@ -70,6 +70,141 @@ export function getHostnameWithoutWww(url, log) {
...
+  if (served) {
+    return served;
+  }
+
+  // Probe undetermined (network error, opaque status, off-domain redirect) — trust the synthesized
+  // www host only if it resolves in DNS, otherwise fall back to the apex.
+  if (await hostResolves(candidate, log)) {
```
ABHA61, line 197: [P2] Do not turn a WAF-blocked probe into a host decision

This probe currently inherits the Spacecat/1.0 User-Agent from tracingFetch, while related Optimize at Edge probes use AdobeEdgeOptimize-Test AdobeEdgeOptimize/1.0. WAFs may handle or block these product User-Agents differently. A 403, 429, 5xx, or network failure from either probe is inconclusive, but this fallback can select www solely because it has DNS. Explicitly choose and document the intended probe User-Agent, test both Spacecat/1.0 and AdobeEdgeOptimize/1.0 where relevant, and probe both root candidates or retain a validated/configured host when HTTP resolution is inconclusive.

src/support/edge-routing-utils.js
```diff
@@ -70,6 +70,141 @@ export function getHostnameWithoutWww(url, log) {
...
+ * @param {object} [log] - Logger.
+ * @returns {Promise<string|null>} The serving host, or null when undetermined.
+ */
+async function probeServingHost(baseURL, apexHost, log) {
+  let res;
+  try {
+    res = await fetch(baseURL, {
+      method: 'GET',
+      redirect: 'manual',
+      signal: AbortSignal.timeout(PROBE_TIMEOUT_MS),
```
ABHA61, line 117: [P2] Pass the timeout through tracingFetch

tracingFetch creates its own signal from options.timeout or its 10-second default and overwrites the supplied AbortSignal before calling Adobe fetch. Therefore AbortSignal.timeout(PROBE_TIMEOUT_MS) does not enforce the intended five-second ceiling here. Please pass timeout: PROBE_TIMEOUT_MS through the supported tracingFetch option.

src/controllers/llmo/llmo-cloudflare.js
```diff
@@ -527,22 +532,43 @@ function LlmoCloudflareController(ctx) {
...
+    // (apex/www chosen by redirect/DNS follow) while preserving the client-supplied path. This
+    // corrects an apex↔www mismatch — e.g. a www pattern for an apex-only site — without touching
+    // the path or specific-subdomain / wildcard patterns.
+    let pattern = clientPattern;
+    try {
+      const siteApex = registrableDomain(new URL(site.getBaseURL()).hostname);
+      const clientHost = routePatternHostGlob(clientPattern);
+      if (clientHost === siteApex || clientHost === `www.${siteApex}`) {
+        const canonicalHost = await resolveCanonicalHost(site.getBaseURL(), log);
+        pattern = `${canonicalHost}${routePatternPath(clientPattern)}`;
```
ABHA61, line 563: [P1] Preserve a validated customer route-host correction path

Both apex and www client patterns are unconditionally replaced with the inferred host. If resolution is wrong, RACQ cannot submit www.racq.com.au/* and Genting cannot submit gentingsingapore.com/* to correct it because the API rewrites the pattern again. If customer correction is required, preserve an explicit validated root-host choice or expose a separate override; the worker upstream targetHost can remain server-controlled.

src/controllers/llmo/llmo-cloudflare-utils.js
```diff
@@ -76,6 +76,18 @@ export const routePatternHostGlob = (pattern) => pattern
...
+/**
+ * The path glob of a route pattern: everything from the first "/" after the host, scheme stripped,
+ * defaulting to "/*" when the pattern carries no path (e.g. "example.com/blog/*" -> "/blog/*",
+ * "https://example.com/*" -> "/*", "example.com" -> "/*"). Lets callers swap a route's host while
+ * preserving the client-supplied path.
+ */
+export const routePatternPath = (pattern) => {
+  const withoutScheme = pattern.replace(/^https?:\/\//i, '');
+  const slashIndex = withoutScheme.indexOf('/');
+  return slashIndex === -1 ? '/*' : withoutScheme.slice(slashIndex);
```
ABHA61, line 88: [P2] Preserve the complete route pattern while replacing its host

Rebuilding the pattern from canonicalHost plus this extracted path removes an explicit https:// scheme and changes a pattern with no path into /*. Cloudflare treats scheme-specific, scheme-less, root-only, and wildcard-path routes differently. Please replace only the hostname in the original pattern so its scheme and exact path scope remain unchanged.

ABHA61:
@jindaliiita could you please have a look at the canonical-host and route-resolution behavior in this PR and see if directions is correct.

I added inline review comments. The two concrete customer cases use these stored base URLs:

- https://racq.com.au/ — the current dot-count/early-return path keeps the apex, although the live apex redirects to www.racq.com.au.
- https://gentingsingapore.com — with the actual Spacecat/1.0 tracingFetch User-Agent, the apex returns a relative 301 to /en/home; the relative Location is misparsed, after which DNS fallback selects www even though www redirects back to the apex.

I also flagged WAF/User-Agent ambiguity, the tracingFetch timeout option, customer route-host correction, and preservation of the complete Cloudflare route pattern.
@sahil9001 we might need to see if we can use calculateForwardedHost if it is getting used at multiple places.

src/support/edge-routing-utils.js
ssilare-adobe: Fixed — removed DNS from the resolve path entirely rather than adding a timeout. `resolveCanonicalHost` now only does redirect-follow (+ a PSL apex check) and falls back to the pre-probe candidate on any failure, so there's no unbounded DNS wait left in the request path.

src/support/edge-routing-utils.js
ssilare-adobe: Same fix as the first review pass — DNS removed from the resolve path entirely (see reply on the other copy of this comment).

src/support/edge-routing-utils.js
ssilare-adobe: Fixed — now uses `redirect: 'follow'` and reads the final `response.url` directly (equivalent to `new URL(location, probeURL)` resolution), so relative/absolute/protocol-relative Location values all resolve correctly. Verified against the live gentingsingapore.com redirect chain (apex → `/en/home`).

src/support/edge-routing-utils.js
ssilare-adobe: Fixed — added a PSL-based (tldts) override for calculateForwardedHost's dot-count miss on compound TLDs. Verified against live domains: racq.com.au → www.racq.com.au, and google.co.uk → www.google.co.uk (correctly does NOT follow through to the off-domain www.google.com redirect).

src/support/edge-routing-utils.js
ssilare-adobe: Fixed — the www candidate is now actually fetched with its full redirect chain followed (`redirect: 'follow'`), so a www→apex redirect is discovered directly instead of inferring it from DNS existence. Verified against live gentingsingapore.com — resolves to the apex as expected.

src/support/edge-routing-utils.js
ssilare-adobe: Fixed — the probe now sends the same documented `EDGE_OPTIMIZE_USER_AGENT` used by `probeSiteAndResolveDomain` elsewhere in this module, instead of tracingFetch's default UA.

src/support/edge-routing-utils.js
ssilare-adobe: Fixed — tracingFetch builds its own AbortSignal from `options.timeout` and was silently discarding the `signal: AbortSignal.timeout(...)` we passed, so the real timeout was always its 10s default. Replaced with `timeout: X` in all three fetch call sites in this file (probe, probeSiteAndResolveDomain, callCdnRoutingApi).

src/controllers/llmo/llmo-cloudflare.js
ssilare-adobe: Discussed with the PR author — decided to keep the auto-normalize-host behavior for POST routes as-is, since the underlying resolution bugs motivating this concern (missing redirect-follow, PSL misclassification) are now fixed upstream in resolveCanonicalHost. Not adding a client-side override for now; can revisit if a real-world resolution mistake surfaces.

src/controllers/llmo/llmo-cloudflare-utils.js
ssilare-adobe: Fixed — added `replacePatternHost`, which swaps only the host and preserves the pattern's scheme and path (or lack of a path) exactly, replacing the old rebuild-from-`routePatternPath` approach that dropped schemes and injected `/*` into root-only patterns.

ABHA61 (commented):
Thanks, Sahil, for working on these updates. I have added a few suggestions around providing the resolved values as defaults, allowing the customer to correct them, and then validating and preserving the customer-confirmed values. I’ll review the updated flow once the changes are ready.

src/controllers/llmo/llmo-cloudflare.js
```diff
@@ -372,14 +372,14 @@ function LlmoCloudflareController(ctx) {
...
-    // origin and keeping host derivation consistent across CDNs (CloudFront uses the same helper).
-    // Known gap: sites served from the apex without a www record still resolve to www here; that
-    // is tracked as a follow-up enhancement.
+    // origin and keeping host derivation consistent across CDNs (CloudFront uses the same helper),
+    // but then confirms that synthesized www host actually resolves in DNS — sites served only from
+    // the apex (no www record) fall back to the apex instead of pointing the worker at a dead host.
     let targetHost;
     try {
-      targetHost = calculateForwardedHost(site.getBaseURL(), log);
+      targetHost = await resolveCanonicalHost(site.getBaseURL(), log);
```
ABHA61, line 382: **Suggestion:** Would it make sense for the BE to provide the resolved `targetHost` and suggested route pattern to the FE as defaults, and then use the customer-confirmed `targetHost` after validating that it belongs to the site domain? For example, for `https://racq.com.au/`, we may suggest `www.racq.com.au`; if the customer corrects it to `racq.com.au`, we could preserve the corrected value.

src/controllers/llmo/llmo-cloudflare.js
```diff
@@ -527,22 +532,43 @@ function LlmoCloudflareController(ctx) {
...
+    // (apex/www chosen by redirect/DNS follow) while preserving the client-supplied path. This
+    // corrects an apex↔www mismatch — e.g. a www pattern for an apex-only site — without touching
+    // the path or specific-subdomain / wildcard patterns.
+    let pattern = clientPattern;
+    try {
+      const siteApex = registrableDomain(new URL(site.getBaseURL()).hostname);
+      const clientHost = routePatternHostGlob(clientPattern);
+      if (clientHost === siteApex || clientHost === `www.${siteApex}`) {
+        const canonicalHost = await resolveCanonicalHost(site.getBaseURL(), log);
+        pattern = replacePatternHost(clientPattern, canonicalHost);
```
ABHA61: **Suggestion:** Once `clientPattern` passes the site-domain validation, could we preserve it instead of resolving the host again? For example, if we suggest `www.racq.com.au/*` and the customer changes it to `racq.com.au/*`, we could create the route using the customer-confirmed `racq.com.au/*` value.

src/support/edge-routing-utils.js
```diff
@@ -70,6 +78,80 @@ export function getHostnameWithoutWww(url, log) {
...
+ * @returns {Promise<string|null>} The final serving host, or null when undetermined.
+ */
+async function followToServingHost(probeURL, rootHost, log) {
+  let res;
+  try {
+    // Same probe UA as probeSiteAndResolveDomain below, for consistent WAF treatment.
+    res = await fetch(probeURL, {
+      method: 'GET',
+      redirect: 'follow',
+      headers: { 'User-Agent': EDGE_OPTIMIZE_USER_AGENT },
```
ABHA61: **Suggestion:** Could we use the combined `Spacecat/1.0 AdobeEdgeOptimize/1.0` identity for the canonical-host probe? This may improve compatibility with WAF-protected domains. For example, `www.indiafirstlife.com` returned HTTP 400 with the current probe identity and HTTP 200 with the combined identity.

src/support/edge-routing-utils.js
```diff
@@ -70,6 +78,80 @@ export function getHostnameWithoutWww(url, log) {
...
+ * @param {object} [log] - Logger.
+ * @returns {Promise<string|null>} The final serving host, or null when undetermined.
+ */
+async function followToServingHost(probeURL, rootHost, log) {
+  let res;
+  try {
+    // Same probe UA as probeSiteAndResolveDomain below, for consistent WAF treatment.
+    res = await fetch(probeURL, {
+      method: 'GET',
+      redirect: 'follow',
```
ABHA61: **Suggestion:** Could we consider following redirects manually with a small hop limit and validating each `Location` before requesting it? With `redirect: 'follow'`, an off-domain destination may be contacted before the final hostname is validated. An invalid or off-domain redirect could continue to use the accepted `www` fallback.

ssilare-adobe: Good catch, fixed — this was a real gap: `redirect: 'follow'` meant the client contacted every hop before we validated where it landed. Switched to `redirect: 'manual'` with a small hop cap (3), validating each `Location`'s host against the site's domain *before* requesting it — an off-domain hop is never contacted at all. Added a regression test asserting exactly that (only one fetch call happens when the first hop redirects off-domain).

src/support/edge-routing-utils.js
ssilare-adobe: Fixed — verified against the live domain: `www.indiafirstlife.com` returns 400 with the bare `AdobeEdgeOptimize-Test AdobeEdgeOptimize/1.0` UA and 200 with `Spacecat/1.0 AdobeEdgeOptimize/1.0`. Switched the canonical-host probe to the combined identity (`CANONICAL_HOST_PROBE_USER_AGENT`). Left `probeSiteAndResolveDomain`'s UA untouched since that one specifically tests for the documented default-UA-routing behavior, not just avoiding WAF blocks.

src/controllers/llmo/llmo-cloudflare.js
ssilare-adobe: Implemented — removed the host-derivation block entirely; `addRoute` now uses `clientPattern` exactly as submitted once it passes the `hostInSiteDomain` check, so a customer can resubmit `racq.com.au/*` (or any root-host variant) and it's honored verbatim. This reverses the earlier decision on the related comment (this thread's sibling) to keep auto-normalizing — re-raising it here made the tradeoff clearer, so we went with preserving client control instead. Also removed `replacePatternHost` (now dead code) and updated the OpenAPI description accordingly.

src/controllers/llmo/llmo-cloudflare.js
ssilare-adobe: Will be taking this up as a follow up.

ABHA61 (approved):

PR #3070, opened 2026-08-17
fix: write \0 instead of a raw NUL byte in composite map keys

### LINKED ISSUE #3067: Raw NUL byte in two source files makes recursive grep/ripgrep silently skip them
Two source files contain a raw `U+0000` byte embedded directly in a template literal, rather than the `\0` escape. Standard search tools classify a file containing a NUL as binary and skip it, so both files are invisible to `grep -r` and to recursive `rg`. The failure is silent: a search still returns hits from other files, so nothing signals that a file was skipped.

## Where

| File | Line | Expression |
|---|---|---|
| `src/support/serenity/handlers/markets-subworkspace.js` | 431 | `` const key = `${typeValue}<U+0000>${intentValue}`; `` |
| `src/support/elements/elements-service.js` | 237 | `` const rowKey = (row) => `${row?.prompt ?? ''}<U+0000>${row?.prompt_topic ?? ''}`; `` |

In both cases the NUL is used as a composite map-key separator — a reasonable choice, since NUL cannot occur in the values being joined. The problem is only that the byte is written literally into the source instead of escaped.

## Why it matters

Searching for a symbol defined in one of these files returns the call sites but never the definition. Verified on `main`, 2026-08-17:

```
$ grep -rl "handleCreateMarketSubworkspace" src/
src/support/serenity/brand-provisioning.js
src/controllers/serenity.js

$ rg -l "handleCreateMarketSubworkspace" src/
src/controllers/serenity.js
src/support/serenity/brand-provisioning.js
```

`markets-subworkspace.js` — the file that defines and exports that function — is absent from both result sets. Naming the file explicitly also fails under plain `grep`:

```
$ grep -c "handleCreateMarketSubworkspace" src/support/serenity/handlers/markets-subworkspace.js
$ echo $?
1
```

`file(1)` reports the file as `data` rather than as source text.

The cost is wasted time and, worse, wrong conclusions: a search that comes back with two hits reads as complete, so it is easy to conclude a symbol has no definition, no other callers, or no test coverage in a file that was never actually read.

## Why this has gone unnoticed

Git is unaffected, which is why nothing in review or CI surfaces it. Git sniffs only the first 8000 bytes of a file for a NUL, and in both files the byte sits past that window (offset 19824 and 11180 respectively). So `git grep` finds these files, `git diff` renders them as text, and the files behave normally in PRs. Only tools that scan the whole file — `grep`, `rg` in recursive mode, `file` — treat them as binary.

## Fix

Replace the literal byte with the `\0` escape in both files. The resulting string is identical at runtime, so there is no behaviour change to the map keys:

```js
const key = `${typeValue}\0${intentValue}`;
```

Verified equivalent: `` `x\0y` === 'x' + String.fromCharCode(0) + 'y' `` is `true`, `charCodeAt(1)` is `0`, length `3`.

After the change both files are ordinary ASCII text and become searchable again.

Worth considering alongside the fix: a guard that rejects a NUL byte in tracked source, so this cannot silently return. A repo-wide scan today finds these two files and no others.

rainer-friederich:
## 1. Abstract

Replaces a raw `U+0000` byte, written literally into a template literal in two source files, with the `\0` escape, and adds a repo-wide test that fails on a raw NUL in any tracked file.

## 2. Reasoning

A file containing a NUL byte is classified as binary by tools that scan the whole file, so `grep -r`, recursive `rg` and `file` skipped `markets-subworkspace.js` and `elements-service.js` entirely. The skip is silent: a search still returns hits from other files, so a result set that is missing a definition reads as complete. Searching for a symbol defined in one of these files returned its call sites but never its definition, which makes it easy to conclude a symbol has no definition, no other callers, or no test coverage in a file that was never read.

Nothing in review or CI reacted to this, and that is the reason it survived. Git sniffs only the first 8000 bytes of a file for a NUL, and both bytes sat past that window, so `git grep`, `git diff` and PR rendering all treated the files as ordinary text. A diff also renders the byte as whitespace, so on screen it is indistinguishable from a space — during this work two independent review agents read the diff and concluded the separator had been a space character, which is worth knowing for anyone reviewing this one.

Both NULs were being used as a composite map-key separator — a sound choice, since neither joined value can contain one. The only problem was that the byte was written literally rather than escaped.

- Before: both files were classified as binary by whole-file scanners and were invisible to recursive search. `file(1)` reported them as `data`.
- After: both are ordinary source text and searchable again. The escaped form produces an identical string, so both map keys are byte-for-byte unchanged and no behaviour changes — a key written with `\0` retrieves a value stored under one built with the literal byte.

Neither key is ever split, parsed back, enumerated or logged; both are opaque lookup keys, so the separator's spelling is the whole of the change.

The new test enumerates tracked regular blobs and fails on a raw NUL, so this cannot silently return. It covers every tracked file type rather than JS template literals alone, and skips a denylist of binary extensions — chosen over a text allowlist deliberately, because an unlisted binary type fails loudly and gets added to the list, whereas an unlisted *text* type would be left silently unguarded, which is the very failure mode being fixed.

## 4. Required information

- Jira / issue: Fixes https://github.com/adobe/spacecat-api-service/issues/3067

Verification that CI does not perform:

- **The guard was mutation-checked, not just observed passing.** Reintroducing the byte at each site makes the test fail and name the exact original coordinates from the issue — `markets-subworkspace.js:431` at byte offset 19824, and `elements-service.js:237` at byte offset 11180. A NUL injected into a Markdown file past the 8000-byte window is also caught, confirming the coverage claim beyond JS.
- **The "no behaviour change" claim was checked against the original bytes**, not the rendered diff: `git show HEAD:<file>` hexdumped at both sites shows `7d 00 24` — `}` then `0x00` then `${`. The separator was a NUL and not a space (`0x20`), so there was no pre-existing key-collision behaviour to change at either site.
- **Runtime equivalence exercised directly:** the escaped and literal-byte forms produce identical strings at both call sites, and a `Map` populated with a literal-byte key is retrieved successfully by an escaped-form key.
- **The bundler preserves the escape.** Running esbuild over the changed module emits `\0` and puts no raw NUL into the output, so the Lambda artifact is searchable too and there is no bundling hazard. (`npm run build` itself needs VPC secrets that are only available in CI, so the check was scoped to the bundler.)
- **Repo-wide scan:** across all tracked files these two were the only offenders, matching the issue's finding — there is no third case to fix.
- A read-only scan of sibling workspace repos found no comparable case in `mysticat-data-service`, `mysticat-projector-service` or `spacecat-shared`. The NUL-bearing files in `mystique`, `project-elmo-ui` and `llmo-data-retrieval-service` are legitimate binaries (images, `.docx`/`.xlsx`, archives, `.ico`), with the exception of two large generated assets under `mystique`'s `app/public/blackboard-visualization/`. Whether those warrant the same treatment is a separate call for that repo and is not addressed here.

## 7. Test plan

Verified locally: the unit suites covering both changed files and their surrounding directories were run, and coverage was confirmed to land on both changed lines specifically rather than merely on the files. The guard test was mutation-checked as described in section 6.

To verify after merge, on `main`:

- `grep -rl "handleCreateMarketSubworkspace" src/` now lists `src/support/serenity/handlers/markets-subworkspace.js` alongside its two call sites; before this change it returned only the call sites. The equivalent recursive `rg` behaves the same way.
- `file src/support/serenity/handlers/markets-subworkspace.js src/support/elements/elements-service.js` reports source text rather than `data`.
- `git ls-files -z | xargs -0 grep -lP '\x00'` returns nothing.

No environment deployment is involved: the change has no runtime surface beyond the two map keys, whose behaviour is unchanged, and the defect being fixed is a property of the source files rather than of a running service.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - clean, well-scoped fix with no runtime behavior change and a solid guardrail test.
**Complexity: LOW** - small diff, single service, no risk signals.
_Changes: Replaces literal NUL bytes with the `\0` escape in two composite map-key separators and adds a repo-wide guard test that prevents reintroduction (4 files)._
Note: Recommend a human read before merge - this change amends `CLAUDE.md`, a documented architectural artifact. The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: `BINARY_EXTENSIONS` denylist will need additions when new binary formats are tracked (e.g. `.wasm`, `.node`) - the fail-loud design handles this by intent, but note the maintenance surface - `test/no-nul-bytes.test.js:29`
- suggestion: Add a one-line inline comment at the separator sites explaining that `\0` is chosen because it cannot appear in field values, making the composite key collision-free - `src/support/elements/elements-service.js:237`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 9m 20s | <b>Cost:</b> $3.52 | <b>Commit:</b> <code>dc31a91fe65b8d3c21d6b87e8a6f787d6a92f4cb</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Addressed the review in `e4da157`.

MysticatBot — both non-blocking items:

- suggestion: document why `\0` is the separator — done, at both sites. Worth more than a note: prompt text is free-form, so a future edit to a visible delimiter would make the composite key genuinely collision-prone. The comment records that constraint and the reason the byte stays escaped, so neither has to be rediscovered.
- nit: the `BINARY_EXTENSIONS` denylist will need additions for new binary formats (`.wasm`, `.node`) — declined, and deliberately not pre-populated. The repo tracks no binary files at all today, so any entry added now is speculative. The design already answers this: an unlisted binary type fails loudly with the offending path named, and the fix is a one-line addition to the list. Pre-seeding guesses trades that loud, self-explanatory failure for a longer list that still cannot anticipate the real case. The inverse choice — a text allowlist — is the one that fails silently, which is the bug this PR exists to close.

On the degraded-review notice: there is no spec document, and none should be invented. The design of record is the linked issue, which contains the analysis, the byte offsets and the fix. Section 4 links it.

Noting the recommendation for a human read: the `CLAUDE.md` change documents the guard this PR adds, in the "Testing Patterns" section.

rainer-friederich:
⚠️ Not validated on dev — access to the route is fine (`…/serenity/brand-presence/prompts` returns 200, `access` reports `hasAccess: true`), but the one brand present on dev has an empty Semrush sub-workspace (0 prompts, 0 topics, 0 weeks), so the changed line never executes for lack of rows.

PR #3069, opened 2026-08-17
feat(serenity): job-status endpoint for async prompt classification (#33)

byteclimber:
## Summary
The async prompt-classification enqueue path (serenity-docs#33) returns `202 {jobId, status}`, but there was **no endpoint to poll that job** — the id was inert, and the worker's `NEEDS_REAUTH` terminal state was never surfaced to any caller. This adds the poll endpoint.

`GET /v2/orgs/:spaceCatId/brands/:brandId/serenity/prompts/jobs/:jobId`:
- Reuses the same `authorize()` brand-access + serenity-active gate as the write path, plus a **defence-in-depth ownership check** (`AsyncJob` `metadata.brandId` must match the resolved brand) and a **job-type guard**, so a caller with access to brand A can't read brand B's job by guessing its id (both report `404`, not `403`).
- Response: `{ jobId, status }` always; `result` (the `{created, skipped, failed}` classify outcome) on `COMPLETED`; `error` on `FAILED`, with **`needsReauth: true`** when the failure was a dead promise token (`code === NEEDS_REAUTH`) so the UI can prompt a re-auth + resubmit rather than showing a generic error.
- Wired into `required-capabilities` (`organization:read`) and the llmo FACS map (`llmo/can_view`). `jobId` was already in `FACS_NON_RESOURCE_PARAMS`.

Also documents the **previously-undocumented `202`** response on the bulk-create path and adds the OpenAPI path + `SerenityPromptClassificationJob` / `SerenityPromptClassificationJobAccepted` schemas.

## Context
Closes the top "still open, blocking" gap from the #33 audit — the classification consumer, runner, and infra are already merged; this makes the returned `jobId` actually usable end-to-end. The `needsReauth`-boolean-only contract (no server-built reauth URL) was chosen deliberately so the API doesn't presume the not-yet-built UI's auth flow.

## Not in this PR (tracked follow-ups)
- **`deferPublish` routing collision** — async routing keys off `deferPublish:true`, which elmo-ui also sets on non-final CSV chunks; needs reconciliation before the UI relies on this path.
- UI 202/poll + re-auth surface (Phase 7).
- Async-path observability/metrics.

## Test plan
- [x] `test/controllers/serenity.test.js` — 9 new cases (in-progress / completed+result / failed+needsReauth / failed-non-reauth / 400 bad jobId / 404 not-found / 404 brand-mismatch / 404 wrong-job-type / 403 no-access)
[...]

PR #3071, opened 2026-08-17
fix(serenity): decouple async prompt-classification routing from deferPublish (#33)

byteclimber:
## Summary
Fixes a live routing collision in the async prompt-classification feature (serenity-docs#33). Async routing was keyed off `deferPublish: true` — but the **synchronous CSV-chunking client already sets `deferPublish: true` on every non-final chunk** for publish-batching. For a flat-mode brand, those chunks silently received a `202 {jobId}` job response the client never handles, breaking bulk import. (Latent today only because serenity brands are currently subworkspace-mode, which stays synchronous — it becomes a live bug the moment a flat-mode brand runs a CSV import.)

`deferPublish` and "go async" are unrelated concerns, so this **splits them**: async is now triggered by a **dedicated, explicit `async: true`** flag. `deferPublish` reverts to being purely a synchronous publish-batching hint. Result: every existing write path — single create/edit, UI multi-add, and the sync CSV-chunking client — stays synchronous and untouched, and the async runner is strictly opt-in for callers that will poll the job.

This is the api-service half of the #2/#3 fix. The elmo-ui half (rewrite CSV import to submit-once with `async:true` + poll the job-status endpoint) follows once the poll endpoint (**PR #3069**) lands.

## Changes
- `handlers/prompts.js`: new `validateAsync(body)` (mirrors `validateDeferPublish`).
- `controllers/serenity.js` `createPrompts`: routes on `validateAsync`; `validateDeferPublish` still governs the sync publish-batching path.
- OpenAPI: documents the new `async` request property, explicitly distinct from `deferPublish`.
- Tests: async-routing suite rekeyed to `async:true`; **added a regression guard that `deferPublish:true` alone stays synchronous** (the collision); `validateAsync` unit tests.

## Coordination with PR #3069
Minimal overlap, no code conflict (different functions/import lines). One doc nit: #3069's POST `202` response prose currently says the trigger is `deferPublish` — after this lands it should read `async`. Flagged to update in #3069.

## Test plan
- [x] `test/controllers/serenity.test.js` + `test/support/serenity/handlers/prompts.test.js` — 358 passing, 0 failing
- [x] eslint clean on all changed files
- [x] OpenAPI: my `async` schema addition introduces zero lint problems

### DIFF
--- a/docs/openapi/schemas.yaml
+++ b/docs/openapi/schemas.yaml
@@ -12027,7 +12027,20 @@ SerenityCreatePromptsRequest:
         carries `published: false`). Used for chunked CSV imports, where the
         client posts several chunks and issues a single publish after the last
         one (serenity-docs#32). Omitted/false publishes once at the end of the
-        call as usual.
+        call as usual. This is a synchronous-path hint only and does NOT trigger
+        async processing — use `async` for that.
+    async:
+      type: boolean
+      default: false
+      description: >
+        When true (flat-mode brands only), the request is processed
+        ASYNCHRONOUSLY: instead of classifying + creating + publishing inline,
+        the whole write is enqueued to a background job and the endpoint returns
+        202 with a job id to poll via
+        `GET .../serenity/prompts/jobs/{jobId}` (serenity-docs#33). Deliberately
+        distinct from `deferPublish` (a synchronous publish-batching hint) so
+        that the synchronous CSV-chunking client is never routed to async
+        unexpectedly. Omitted/false processes synchronously as usual.
 
 SerenityCreatePromptsResponse:
   type: object
--- a/src/controllers/serenity.js
+++ b/src/controllers/serenity.js
@@ -29,7 +29,7 @@ import {
   handleCreatePrompts,
   handleUpdatePrompt,
   handleBulkDeletePrompts,
-  validateDeferPublish,
+  validateAsync,
   BULK_PROMPTS_MAX_ITEMS,
   resolveCallerId,
 } from '../support/serenity/handlers/prompts.js';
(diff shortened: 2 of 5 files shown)


MysticatBot (approved):
Hey @byteclimber,

**Verdict: Approve** - well-scoped fix that correctly decouples two independent concerns.
**Complexity: HIGH** - medium diff; API surface change (OpenAPI schema).
_Changes: Decouples async job routing from `deferPublish` by introducing a dedicated `async: true` flag, keeping the sync CSV-chunking path untouched (5 files)._
Note: Recommend a human read before merge - this change modifies a shared API contract (OpenAPI schema). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: `async` is a JavaScript reserved word - works fine as an object property but cannot be destructured without renaming (`const { async: isAsync } = body`) and may require quoting in some client codegen tools - `src/support/serenity/handlers/prompts.js:205`
- suggestion: Add a test for `{ async: true, deferPublish: true }` sent together, confirming async takes priority and deferPublish is inert on the async path. The behavior is correct (early return on validateAsync means deferPublish is never evaluated) but an explicit assertion documents the interaction for future contributors - `test/controllers/serenity.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 28s | <b>Cost:</b> $7.56 | <b>Commit:</b> <code>18302d76a46b62934584527fa01f60635ad1b773</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

byteclimber:
Closing as not-adopted. The async-classification work landed via #3138 (merged), which deliberately kept `deferPublish` as the routing gate (`serenity.js` still routes on `validateDeferPublish(body)`) rather than decoupling from it as proposed here. Reopen/rebase if we later decide to remove the deferPublish dependency.

PR #3073, opened 2026-08-17
feat(serenity): resolve Sites by full identity and surface each market's tracked url

rainer-friederich:
## 1. Abstract

Surfaces each Semrush market's `primary_url` — the url its project actually analyses — on the markets read model, makes SpaceCat Site resolution use a site's full identity (host **and** path) instead of its bare host, and gives both provisioning call sites one shared create → PATCH → publish sequence.

## 2. Reasoning

A Semrush project carries two URL-ish fields and they are not interchangeable:

- **`domain`** must be a bare FQDN. A path returns a hard 400, the upstream normalizes whatever it receives to the registrable domain, and it is accepted **only at create**.
- **`settings.ai.primary_url`** is the url the project actually analyses. It accepts a subdomain, a subpath, or both.

The markets read model carried only `domain`, so every market in a sub-workspace whose brands share an apex rendered identically: `nba.com` for Kings, Knicks and Lakers alike. The one field that distinguishes them was not in the response at all.

The Site-resolution defect is the same conflation one layer down. `ensureMarketSite` rebuilt a base URL from a bare hostname and resolved it with `Site.findByBaseURL`, on the stated assumption of "at most one Site per domain". Prod contradicts that: `sites.base_url` is unique over the **whole URL including its path**, 440 of 12,926 sites carry a subpath, and `nba.com` coexists with `/kings`, `/knicks`, `/lakers` and `/timberwolves` as five distinct rows. Resolving by host alone returns the root site for every subpath sibling — and that Site becomes `brands.site_id`, which sibling brands on one apex then collide on.

The same host-vs-identity conflation suppressed a brand's own website URL: a subpath brand's `https://nba.com/kings` was skipped from `brand_urls` because its *host* matched a market's primary domain, even though a project's `domain` cannot hold a path and the two strings are visibly different entries.

**Markets carry the url they analyse.** `primaryUrlOf` reads `settings.ai.primary_url` and `projectToSlice` puts it on the market model beside `domain`. The field is asymmetric upstream — sent flat on `model.ProjectUpdateRequest`, read back nested under `settings.ai` — so both spellings are accepted on read, and a whitespace-only value is trimmed before the emptiness test rather than surfacing as a blank url. `domain` stays, and its schema description now says what it is: the registrable domain the project is filed under, not the site it tracks.

**Site resolution by full identity.** `ensureMarketSite` looks Sites up by host **and** path, so `nba.com/kings` resolves to its own Site rather than silently linking the root. Three properties fall out of that and are pinned by tests:

- An existing subpath Site stored *with* a trailing slash is adopted rather than duplicated — the identity strips one, 8 such rows exist in dev, and without the second lookup the create would produce a near-duplicate the uniqueness constraint cannot catch.
- The base URL is composed from host and path separately, because `composeBaseURL` is a hostname helper that lowercases its whole argument and would fold `acme.com/OMES` onto `/omes`. Paths are case-sensitive on most origins.
- A resolved hostname does not imply a resolved identity: WHATWG reads the first path segment of `https:///foo.example.com/bar` as the host, yielding a public hostname where the identity is null. That case is skipped rather than falling back to the hostname, since naming the site `foo.example.com` would be the parser's artifact and not the caller's intent.

The SSRF guard still receives a bare host, never a host and path.

**One provisioning sequence, shared.** `create` ignores `primary_url`, so a non-apex project cannot be provisioned in one call and the PATCH must land before the publish or the corrected value stays in draft. That sequence — create, PATCH, publish, and best-effort orphan cleanup on a failed publish — now lives in one helper both call sites use. Duplicating it fails silently: one path gains the PATCH and the other does not, and the difference only shows up as brands tracking the wrong url months later.

A failed PATCH is best-effort at both call sites. It is logged under a greppable token (`SERENITY_MARKET_PRIMARY_URL_DIVERGENCE`) and the publish proceeds, leaving a market live on its apex — the state every market was in before this work, and one the `mysticat-data-service` reconcile repairs in place. Deleting an otherwise-valid project because a refinement could not be applied would trade a recoverable degradation for no market at all.

**Brand create derives both values from one input.** `brandUrlFromPayload` selects the payload's primary URL once; `domain` and the tracked url are both derived from it, so they cannot describe two different sites. The Site mirror for the initial market is keyed on the tracked url rather than the host, falling back to the host when the payload's spelling yields no identity.

**A brand's own subpath URL is no longer suppressed.** The `brand_urls` skip now compares whole urls, not hosts, so a url carrying a path survives even on a market's primary host.

**`primary_url` is derived, not accepted, on the sub-workspace create-market route.** The controller sets it from the server-side derivation unconditionally, so a caller-supplied value on that request body is discarded rather than forwarded. It is not part of the create-market contract and needs no request-schema change.

The local `siteIdentityFromUrlString` in `src/support/url-utils.js` is replaced by the released `@adobe/spacecat-shared-utils` export, so one implementation of the normalizer exists rather than two that can drift.

## 4. Required information

- Jira / issue: https://github.com/adobe/serenity-docs/issues/348
- Spec: `docs/serenity.md` (updated here — the create-market flow, and the Site-mirroring section that asserted the now-disproven "at most one Site per domain")
- Other: the upstream contract was established live against `adobe-hackathon.semrush.com` and is recorded in the umbrella issue

- **spacecat-shared** — consumed. Provides `siteIdentityFromUrlString`, released in `@adobe/spacecat-shared-utils` 1.126.0, which this repo's exact-version pin is set to.
- **mysticat-data-service** — contract. Writes and reconciles the same field for existing projects. A project provisioned here must read as clean to its drift detection, and vice versa; the two derivations are pinned to one shared corpus for that reason.
- **project-elmo-ui** — consumes `primaryUrl` on the market model to render each market's tracked url. It falls back to `domain` where the field is absent, so it is safe in either order.

- The `ProjectUpdateRequest` shape was checked against the vendored spec rather than assumed: `primary_url` is declared at the top level and the type has no `settings` member, so a nested write spelling would have been dropped silently — the feature would have done nothing while every test passed.
- Prod was queried to confirm the premise behind the Site-resolution change: 440 of 12,926 sites carry a subpath, and `nba.com` plus its four subpath siblings are five distinct rows. Dev was queried for the trailing-slash case: 8 of 1,718 sites end in one, and no `/x` and `/x/` pair coexists.
- No extra lookup is made for the previously-composed all-lowercase spelling. Prod holds 3 sites whose path carries capitals and 6 on a `www.` host, and not one of them backs a brand, a `brand_sites` link, or a market — so there is no row for such a lookup to find, only a round-trip on every create.
- The identity primitive was differential-fuzzed against its Python counterpart in `mysticat-data-service` over 4,000 randomized inputs and 30 real dev site `base_url` values, with zero divergences, after an initial pass found 10 disagreements — including internationalized hosts and non-ASCII paths, which would otherwise have made every reconcile rewrite a correct value forever.

## 7. Test plan

**Locally, beyond unit tests:** the suite was run against the released `@adobe/spacecat-shared-utils` 1.126.0, so the imported primitive is the published one rather than a local stand-in.

**Not verified here:** no project was provisioned against a live Semrush workspace. That needs the change deployed.

**Verification on dev, once deployed:** provision a market for a brand whose site `base_url` carries a subpath and read the project back —

- `domain` should be the apex (the upstream folds it there regardless of what was sent);
- `settings.ai.primary_url` should be the full subpath, e.g. `nba.com/kings`;
- `publish_status` should be `live`, not `live_with_unpublished_updates` — that would mean the PATCH landed but the publish did not.

Then read the markets endpoint for that brand and confirm each market reports its own `primaryUrl` rather than a shared `domain`.

Then confirm the Site mirror: a market created for a subpath brand should link the subpath Site, not the root one. For a brand whose site has no path, the resulting project and Site should be identical to what they are today — that no-op is the safety property for the ~12,486 path-free sites.

- https://github.com/adobe/spacecat-shared/pull/1882 — **depends-on**, satisfied. It added the identity primitive and released it as `@adobe/spacecat-shared-utils` 1.126.0; the pin here is set to that version.
- https://github.com/adobe/spacecat-api-service/pull/3090 — merged. It established the provisioning write and a local copy of the identity helper; this PR consolidates that write into one shared sequence for both call sites and moves the helper onto the released shared implementation.
- https://github.com/adobe/mysticat-data-service/pull/919 — related, no ordering constraint. It repairs existing projects; this PR keeps new ones correct. Either order is safe: a project provisioned here reads as clean to that PR's drift detection.
- https://github.com/adobe/project-elmo-ui/pull/2875 — related, no ordering constraint. It renders the `primaryUrl` this PR adds and degrades to `domain` without it.

No remaining ordering constraint. `mysticat-data-service` and `project-elmo-ui` merge whenever convenient, before or after.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```


MysticatBot (changes requested):
Hey @rainer-friederich,

**Verdict: Request changes** - one fragile error-dispatch pattern that silently degrades the API contract.
**Complexity: HIGH** - large diff; API surface + dependency change.
_Changes: Adds `primary_url` (the URL a Semrush project actually tracks) to the provisioning flow alongside `domain`, and makes Site resolution use full identity (host+path) instead of just bare host (13 files)._

Note: Recommend a human read before merge - this change modifies the project spec (`docs/serenity.md`) which documents the create-market architectural contract. The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Important] String-based error dispatch silently degrades on refactor - `src/support/serenity/handlers/markets.js:435` (details below)

**`src/support/serenity/handlers/markets.js` ~line 435:**

```js
if (e.message === 'Upstream createProject returned no id') {
  return { status: 502, body: { error: 'createNoProjectId', ... } };
}
```

This matches a literal string thrown by `createProvisionAndPublishProject`. If that message is ever reformatted, this branch stops matching silently and the error propagates as an unhandled 500 instead of the documented 502. API consumers relying on the 502 status for retry logic would be affected without any test or linter catching the regression.

**Fix:** Export a typed error class from `project-provisioning.js` and use `instanceof` here:

```js
// project-provisioning.js
export class CreateNoIdError extends Error { constructor() { super('Upstream createProject returned no id'); } }

// markets.js
if (e instanceof CreateNoIdError) { return { status: 502, ... }; }
```

Alternatively, export the message string as a named constant and reference it in both the throw and the catch.

<details>

- suggestion: Extract the PATCH body shape `{ type: 'ai', primary_url }` as a shared builder or constant - it appears in three locations (project-provisioning.js, markets-subworkspace.js, and described in JSDoc) and the mandatory `type` field is easy to forget on a future edit - `src/support/serenity/project-provisioning.js:89`
- suggestion: Remove the redundant `export default { createProvisionAndPublishProject }` at line 122 of project-provisioning.js - the named export is what every caller imports, and the default creates ambiguity about the canonical import style - `src/support/serenity/project-provisioning.js:122`
- nit: The trailing-slash fallback in `ensureMarketSite` encodes knowledge of legacy data shape into query logic - consider tracking a data migration to normalize the 8 affected rows so this branch can eventually be removed - `src/support/serenity/site-linkage.js:258`
- nit: `docs/serenity.md` says "succeeded both upstream calls" but the flow now has three upstream calls (create, PATCH, publish) - should read "all upstream calls" or "all three upstream calls" - `docs/serenity.md:29`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 30s | <b>Cost:</b> $9.17 | <b>Commit:</b> <code>bad8ed436b55970db6c90d2339745d7b4426af7d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Thanks — review addressed. 1 must-fix and 4 non-blocking items, 5 decisions.

**Must fix (1)**

- **Fixed** — string-based error dispatch. `CreateNoProjectIdError` is now exported from `project-provisioning.js` and the handler matches on `instanceof`. Your reasoning is exactly right, and the suite was complicit in it: the existing test asserted `rejectedWith('Upstream createProject returned no id')`, so a reword would have been caught *there* while the handler's dispatch broke silently. There is now a test asserting the type rather than the wording, alongside the message one.

**Non-blocking (4)**

- **Fixed, narrower than proposed** — the PATCH body shape. Two of the three sites you counted are the PATCH shape; the third (`markets-subworkspace.js:249`) is a **create** body, which carries `type: 'ai'` but is a different shape with different required fields, so folding it in would have been wrong. Both PATCH sites now go through `primaryUrlPatchBody(url)`, which captures both easy-to-forget requirements — the mandatory `type` and the flat `primary_url` spelling — in one place. I made it specific rather than a generic `fields` builder: both callers set exactly one field, and a `Record<string, unknown>` return erased the field types against the transport's typed body (`TS2345`), which is precisely the checking this directory exists to keep.

- **Fixed** — removed the redundant `export default`. Confirmed nothing imported it: both the handler and the test use named imports.

- **Declined** — migrating the 8 trailing-slash rows away. The fallback is not legacy-shape knowledge; `sites.base_url` is unique over the whole URL including its path, so `example.com/x` and `example.com/x/` are two legitimately distinct rows that a normalization would have to merge, with whatever references them. That is a data migration with its own risk and its own review, not a cleanup this PR can absorb — and without the fallback here, the lookup misses and creates a near-duplicate Site the uniqueness constraint cannot catch. I have not filed a ticket for it, since filing is opt-in.

- **Fixed** — `docs/serenity.md` now says "all three upstream calls" in both places the old two-call count appeared (the flow preamble and the 409-gate note).

On the CI note in your review: those checks are not failing. `branch-deploy`, `deploy-stage` and `semantic-release` show as **cancelled** runs from the draft phase, superseded when the PR was marked ready. On the reviewed commit `ci / build`, `ci / it-postgres`, `ci / branch-deploy`, `codecov/patch` and `type-check` all passed.

Agreed on a human read before merge — this changes the create-market contract in the spec, and the `docs/serenity.md` diff is the part worth reading closely.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all prior findings addressed; no new blocking issues.
**Complexity: HIGH** - large diff; API surface + dependency change.
_Changes: Adds `primary_url` to Semrush project provisioning and makes Site resolution use full identity (host+path) instead of bare host (13 files)._

Note: Recommend a human read before merge - this change modifies the project spec (`docs/serenity.md`) which documents the create-market architectural contract. The bot review is a complement to, not a replacement for, a human read here.

<details>

- suggestion: The sub-workspace handler reads `body.primaryUrl` directly (line 507 of `markets-subworkspace.js`); today the controller always overwrites it, but re-deriving inside the handler via `siteIdentityFromUrlString` would make it self-protecting regardless of future call sites - `src/support/serenity/handlers/markets-subworkspace.js:507`
- suggestion: The behavioral split (strict PATCH failure in flat handler vs best-effort in sub-workspace) is documented in code comments but not in `docs/serenity.md`; a sentence there would help operators running reconciliation understand when `SERENITY_MARKET_PRIMARY_URL_DIVERGENCE` fires - `docs/serenity.md`

</details>

- String-based error dispatch replaced by `CreateNoProjectIdError` with `instanceof` check
- PATCH body shape extracted to shared `primaryUrlPatchBody()` helper
- Redundant `export default` removed from `project-provisioning.js`
- `docs/serenity.md` corrected from "both upstream calls" to "all three upstream calls"

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 9s | <b>Cost:</b> $9.24 | <b>Commit:</b> <code>895e5b77602bc0e373acf412cfb515332eb4f703</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
2 non-blocking items, 2 decisions.

- **Declined** — re-deriving `primaryUrl` inside the sub-workspace handler. This one would be a regression rather than hardening. `markets-subworkspace.js:507` is inside a JSDoc block; the only read is at line 648, and it already falls back to `siteIdentityFromUrlString(body.brandDomain)` when the body carries nothing.

  The value it would be replacing is the richer one. When the request supplies a `siteId`, the controller derives `primaryUrl` from that Site's `base_url` via `resolveSiteUrls` (`serenity.js:816`) — host **and** path. `brandDomain` is host-only by construction, so re-deriving from it inside the handler would collapse `nba.com/kings` back to `nba.com` for exactly the brands this PR exists to fix. The handler cannot self-derive the way the flat one does, either: its `dataAccess` is narrowed and it has no `Site` access, which is why the derivation lives in the controller. The self-protection you're after is already there and is the reason the controller sets the field unconditionally rather than reading it from the request.

- **Fixed** — documenting the strict-vs-best-effort split. Added to the Observability section, next to the other greppable tokens, since that is where someone meets `SERENITY_MARKET_PRIMARY_URL_DIVERGENCE` for the first time. It now states why the two paths differ, and the part an operator actually needs: a market carrying the token is live and usable but tracking its apex, the `mysticat-data-service` reconcile repairs it in place because it compares against live state, and the signal worth chasing is recurrence for the same market across reconciles rather than a single occurrence.

rainer-friederich (commented):
Cross-checked this PR against the spec ([adobe/serenity-docs#348](https://github.com/adobe/serenity-docs/issues/348)) and the sibling PRs ([spacecat-shared#1882](https://github.com/adobe/spacecat-shared/pull/1882), [mysticat-data-service#919](https://github.com/adobe/mysticat-data-service/pull/919)). One confirmed defect, three scope-level observations.

## What checks out

- **Upstream contract** matches the spec exactly: create → PATCH → publish order, PATCH body sent flat with the mandatory `type: 'ai'`, publish last so the value cannot stay in draft. All pinned by tests.
- **Cross-repo parity with mysticat-data-service#919 is real, not just claimed**: the Python `site_identity()` mirrors the JS rules including the WHATWG edge cases (backslash-as-slash, IDNA punycode, path percent-encoding), the Python PATCH body is the same flat `{primary_url, type: 'ai'}` shape, and the two corpus fixture files are byte-identical (verified by SHA-256).
- **The spec's derivation split** is honored: the `markets.js` and `brands.js` domain derivations stay host-only; the identity flows where workstream 2 says it should.
- The removed `resolveSiteDomain` has no stale references; the dependency pin matches the released `@adobe/spacecat-shared-utils` 1.126.0; every deleted test assertion corresponds to an intentionally changed behavior and has a replacement; the shared helper's orphan-cleanup semantics are logically identical to the old inline code.

## Finding 1 — confirmed: `ensureMarketSite` can throw, violating its never-throw contract

`site-linkage.js` (the comment above `composeBaseURL(/** @type {string} */ (identity))`) asserts the identity cannot be null once the hostname resolved: *"the only input the identity rejects on its own — a leading '/' — yields an empty hostname too."* That is false. Probed against the released 1.126.0:

```
input: /foo.example.com/bar | hostname: foo.example.com | isPublic: true | identity: null
```

WHATWG collapses the slash run in `https:///foo.example.com/bar` and reads the first path segment as the host, so `hostnameFromUrlString` returns a public multi-label hostname while `siteIdentityFromUrlString` (which deliberately rejects a leading `/`) returns null. `composeBaseURL(null)` then throws a TypeError — and it sits **outside** the try/catch, in a function whose documented contract is best-effort/never-throw (`brands.js` relies on it: "a throw here would tear down a live brand's workspace").

Reachability is narrow: the brands.js caller pre-reduces to a hostname, and the market-create callers only run the mirror after a 201 the upstream would refuse for such a domain. The one live route is **activate-retry**: markets already live (all 409) plus a leading-slash `brandDomain` in the body (validation is presence-only) → 500 instead of a graceful skip. Trivial fix: null-guard `identity` (warn + return null, or fall back to `hostname`) instead of the type assertion.

## Finding 2 — latent: `composeBaseURL` mangles the identity it is now fed

The identity preserves path case by design (rule 1: `/OMES` ≠ `/omes`) and keeps `www.` (rule 5), but `composeBaseURL` lowercases the whole string and strips `www.`: `acme.com/UPPER → https://acme.com/upper`, `www.acme.com/uk → https://acme.com/uk`. When a path-carrying value eventually reaches the domain path, the Site lookup will miss any existing row whose stored path has uppercase (or a `www.` host) and create a near-duplicate — the exact class the trailing-slash fallback was added for. The PR queried prod for subpaths and dev for trailing slashes, but not for path case. Latent today, so not a blocker — but the case/www question deserves either a prod query or a comment acknowledging it.

## Finding 3 — spec gap: the Site mirror still discards the path on activation

"Activation stops discarding the path" holds for the Semrush project, but the activate flow still passes host-only `brandDomain` to `ensureMarketSite`, and the linked site id flows on into `linkSiteToLiveRows` and `baseSiteId`. A legacy pending brand drained with a subpath wizard URL (`nba.com/kings`) now gets a project tracking the subpath while its brand/mapping rows link the **root** site. The spec's workstream 2 wanted `ensureMarketSite` on the full identity; the mechanism landed, but no caller feeds it one — and activation is the one caller with the full URL (`brandUrlSource`) in hand. Not a regression (link behavior unchanged from main), but a new internal inconsistency: project and mirror now disagree about which URL the market is about. Worth either passing `brandUrlSource` there or documenting why not.

## Finding 4 — coverage gap: legacy brand-create still provisions apex-tracking projects

The `brands.js` create-with-initial-market path (legacy but still honored) resolves only a host (`brandDomainFromPayload`), and `brand-provisioning.js` passes no `primaryUrl`, so the subworkspace handler falls back to the host identity — a subpath brand created this way tracks its apex until the data-service reconcile repairs it. Consistent with the spec's "stays host-only" listing for that call site, and self-healing, but this path keeps minting projects for the backfill.

Minor: nothing in the test diff asserts the activation path forwards a path-carrying `primaryUrl` into the market bodies — the derivation is covered by line coverage only, with no assertion on the value.

Recommendation: fix Finding 1 before merge (small, and the justifying comment is demonstrably wrong); 2–4 are scope calls — none regress existing behavior.

rainer-friederich:
All four findings addressed, plus the `project.domain` consumers the tracking issue listed. Two of those are not changed, with evidence below.

## Finding 1 — fixed

Reproduced against the released 1.126.0:

```
/foo.example.com/bar  → hostname foo.example.com | isPublic true | identity null | composeBaseURL THROWS
/nba.com              → hostname nba.com         | isPublic true | identity null | composeBaseURL THROWS
```

The comment was wrong and is gone. `ensureMarketSite` now warns and returns null when the hostname resolves but the identity does not, so activate-retry skips gracefully instead of 500ing. Falling back to the hostname was rejected: `/foo.example.com/bar` naming the site `foo.example.com` is the parser reading a path segment as a host, not the caller's intent.

## Finding 2 — fixed, and queried

`composeBaseURL` is a hostname helper — it lowercases the whole string, so `acme.com/OMES → https://acme.com/omes`, and strips `www.`. It now runs on the **host alone**, with the identity's path re-appended verbatim. Host-level normalization (port, trailing dot, `www.`, scheme) is unchanged; path case survives. This matters beyond lookups: paths are case-sensitive on most origins, so the old form also handed downstream fetchers a URL the origin may 404.

Prod, as asked: 12,931 sites, 443 with a path. **3** carry capitals in the path (`bostonscientific.com/en-US/home.html` and two AEM author/publish URLs), **6** are on a `www.` host (1 of those with a path). None of the nine is a brand's primary site, a `brand_sites` link, or a market's site — so there is no pre-existing row a lowercased lookup could miss, and no extra lookup was added for the old spelling. Recorded in the comment.

## Finding 3 — fixed

`ensureMarketSite` on the activate path is fed `brandPrimaryUrl`, not `brandDomain`. That Site becomes `brands.site_id` via `baseSiteId` and the link `linkSiteToLiveRows` writes, so mirroring the host left a brand analysing `nba.com/kings` recorded against the root `nba.com`. It also stops sibling brands on one apex colliding on `brands_base_site_unique`. Nothing moves for a body-supplied `brandDomain` (a bare FQDN whose identity is itself); only a stashed wizard URL carrying a subpath.

## Finding 4 — closed, not accepted

`brandDomainFromPayload` was split so the domain and the tracked url come from one selected payload URL, and `brand-provisioning.js` forwards `primaryUrl` into the create body — the field `handleCreateMarketSubworkspace` already honoured and nothing was passing. The brand-create Site mirror is fed the same identity. That path no longer mints apex-tracking projects for the backfill to repair.

## Minor — closed

The activation test now asserts the value, not just the line: `brandDomain: 'acme.com'`, `primaryUrl: 'acme.com/path'`, mirror on `acme.com/path`. Brand-create has the equivalent.

## From the tracking issue: `project.domain` consumers

**Brand-URL suppression — fixed in both repos.** `collectBrandUrlEntries` skipped any website URL whose *host* was some market's primary domain, which dropped a subpath brand's own `https://nba.com/kings` from `brand_urls` entirely. The skip exists to stop the *same string* being listed twice, so it now compares the whole url: a path-carrying URL is never skipped, a bare `https://acme.com/` still is. The migration CLI carried the identical defect — the `[unverified]` item in the issue, now verified — and has the matching fix and tests.

**Market DTO — fixed.** `primaryUrl` (from `settings.ai.primary_url`, both spellings accepted) added to the market slice, its OpenAPI schema, and the elmo `SerenityMarket` type; the markets overview renders it and falls back to `domain`.

**Own-brand benchmark identity — not changed, and the live data says it is right.** Read the benchmarks on the two repointed NBA projects (`draft=false`, published view):

```
Lakers  US-en   project.domain='nba.com'  primary_url='nba.com/lakers'
        main_brand=True   domain='nba.com'  root_domain='nba.com'  brand_name='Los Angeles Lakers'
Knicks  US-en   project.domain='nba.com'  primary_url='nba.com/knicks'
        main_brand=True   domain='nba.com'  root_domain='nba.com'  brand_name='New York Knicks'
```

The own-brand benchmark's `domain` **is** the project domain, not the tracked url, so `ensureOwnBrandBenchmark` matching on `project.domain` agrees with what is actually stored. Both projects also carry a `main_brand: true` benchmark, so rule 1 resolves them and the domain-match fallback is not even reached. Re-homing to the `primary_url` host would additionally mean delete-and-recreate on live projects, since `main_brand` is create-only (a `PUT` returns 202 and silently ignores the field — confirmed with Semrush eng 2026-07-01).

**Competitor reservation — a real defect, larger than the issue described, and deliberately not fixed here.** The same read refutes the premise I had been working from. Semrush stores **several benchmarks on one domain**, distinguished by `brand_name` — live on the Lakers project right now:

```
main_brand=False  domain='nba.com'  brand_name='Boston Celtics'
main_brand=False  domain='nba.com'  brand_name='Golden State Warriors'
main_brand=False  domain='nba.com'  brand_name='New York Knicks'
main_brand=False  domain='nba.com'  brand_name='San Antonio Spurs'
```

So a same-apex competitor is perfectly expressible upstream, and our competitor sync is domain-keyed end to end against a key that is not unique:

- `buildReservedDomains` on the brand-edit path reserves `projects.map(p => p.domain)` = `nba.com`, so `dropReservedCompetitors` strips every one of those four from an edit. With LLMO-6591's all-stripped handling, editing an NBA brand's competitors discards the lot.
- `collectCompetitorBenchmarks` de-dupes on `seen.add(domain)`, so at most one of four would survive even unreserved.
- `competitorByDomain` is a `Map` keyed by domain: three of the four are invisible to update/delete, and `toUpdate` would `PUT` whichever won the map — writing one competitor's name and aliases onto another's benchmark.
- `toDelete` maps a removed domain to a single id, so removing one competitor targets an arbitrary sibling.

None of this is a regression from this PR — it predates it — but it is exactly the shape this PR makes common. Fixing it means re-keying reservation, dedupe, update **and delete** from `domain` onto `brand_name` (the key upstream actually enforces — it rejects a duplicate name within a project). That is a refactor of live-benchmark delete paths and wants its own PR and its own review, not a bolt-on to this one. The evidence above is what that PR needs.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all human-reviewer findings addressed correctly; no new blocking issues.
**Complexity: HIGH** - large diff; API surface + dependency change.
_Changes: Guards the null-identity edge case, preserves path case in site resolution, mirrors the tracked URL through activation and brand-create, and carries primaryUrl on the markets DTO (13 files)._

Note: Recommend a human read before merge - this change modifies the project spec (`docs/serenity.md`) and the OpenAPI contract (`docs/openapi/schemas.yaml`). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: Comment in `isBarePrimaryDomainUrl` says "Null identity (an unparseable url) is not bare" but the code returns `true` when `identity === null`, treating it as bare (skipped). The behavior is safe but the comment contradicts it - adjust to match the implementation - `src/support/serenity/brand-urls.js:88`

</details>

- ensureMarketSite null-identity guard added (WHATWG leading-slash edge case no longer throws)
- composeBaseURL path-case preserved by splitting host/path composition
- Activation site mirror now uses the tracked URL instead of host-only domain
- Legacy brand-create forwards primaryUrl from payload into provisioning

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 17s | <b>Cost:</b> $7.67 | <b>Commit:</b> <code>e28c16a1e0f6d066337bef2f0eabbac10ee9c10d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
**Comment/code contradiction in `isBarePrimaryDomainUrl` — fixed.** You are right that the code treats a null identity as bare. That is the intended behavior (a url whose host matched a primary but that the identity cannot parse is skipped, exactly as it was before this distinction existed); the comment described the opposite. The comment now matches.

**On the human-read recommendation:** agreed, and the OpenAPI change is the additive `primaryUrl` on `SerenityMarket` — the field the elmo markets overview needs so a project tracking `nba.com/kings` stops displaying as `nba.com`. Its consumer is https://github.com/adobe/project-elmo-ui/pull/2875 .

PR #3076, opened 2026-08-17
fix(serenity): handle tiered daily/weekly prompt resources in metered-405 canary

aliciadriani:
## Summary
- Semrush's AI product resources now report tiered `daily_prompts`/`weekly_prompts` for some workspaces instead of a flat `prompts.{used,total}` — `scripts/serenity-metered-405-canary.mjs` hard-coded the flat shape and could not run at all against a workspace in the new shape.
- The script now reads both shapes via a `resolvePromptDims` helper. For a tiered workspace it only performs the drain when both tiers are already at zero — the `resources/transfer` write API (`handlers.aiProductResources`) still only accepts a flat `prompts` number, with no documented tiered write shape, so the script refuses to guess and exits with an explanation instead of silently doing the wrong thing.
- Found and fixed live, during the SITES-49206 §10.7 per-environment canary run (dev/stage/prod) — see the results comment on [SITES-49206](https://jira.corp.adobe.com/browse/SITES-49206). All three environments confirmed the no-enforcement premise holds; this fix just makes the script itself usable again for the next time it's needed.

## Test plan
- [x] `npx eslint scripts/serenity-metered-405-canary.mjs` — clean
- [x] `node --check scripts/serenity-metered-405-canary.mjs` — valid syntax
- [x] `--dry-run` verified live against the same real (throwaway) sub-workspace used in the SITES-49206 canary run — correctly detects the tiered shape, prints both dims, and reports zero-headroom status
- [x] No existing test coverage for this script (by design — it's outside `tsconfig`/CI, needs a live IMS token + real sub-workspace id per its own header)

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```


aliciadriani (commented):
## Review — round 1

**Must Fix:** none.

**Should Fix (1):**
1. `resolvePromptDims` / `isZeroHeadroom` are pure, I/O-free helpers with zero test coverage — inline comment below.

**Nits (1):**
1. Unguarded `aiResourcesBefore.projects.total` access in the legacy drain branch — inline comment below.

**Suggestion:** none this round.

**What's good:** correctly diagnosed a real live vendor-shape drift, refuses to guess at an undocumented write contract rather than silently doing the wrong thing against a live Semrush workspace, and header docs were updated in the same PR at the same detail level as the rest of the file.

scripts/serenity-metered-405-canary.mjs
```diff
@@ -120,40 +130,83 @@ function printError(label, e) {
...
+const TIERED_PROMPT_KEYS = ['daily_prompts', 'weekly_prompts'];
+
+function isUsedTotalPair(dim) {
+  return dim && typeof dim.used === 'number' && typeof dim.total === 'number';
+}
+
+// Normalizes the two AI-resources read shapes Semrush has been observed to return (see the header's
+// "Read-side shape drift" note): a legacy flat `prompts` dimension, or the current tiered
+// `daily_prompts` + `weekly_prompts` pair. Returns null when neither shape is present.
+function resolvePromptDims(aiResources) {
```
aliciadriani: **Should Fix:** `resolvePromptDims` (and `isZeroHeadroom` below) are pure, I/O-free functions with no test coverage — unlike the rest of this script, which genuinely can't be unit-tested (needs a live IMS token + real sub-workspace). Recommend a small test file covering: legacy shape, tiered-zero, tiered-nonzero, and unrecognized/missing shape (e.g. only one of the two tiers present).

scripts/serenity-metered-405-canary.mjs
```diff
@@ -120,40 +130,83 @@ function printError(label, e) {
...
 
-  console.log(`Draining prompts.total to ${prompts.used} (an absolute transfer — zero prompt headroom left)...`);
-  await transport.transferWorkspaceResources(subWorkspaceId, {
-    ai: { projects: before.product_resources.ai.resources.projects.total, prompts: prompts.used },
-  });
+  if (promptsBefore.shape === 'legacy') {
+    const [prompts] = promptsBefore.dims;
+    console.log(`Draining prompts.total to ${prompts.used} (an absolute transfer — zero prompt headroom left)...`);
+    await transport.transferWorkspaceResources(subWorkspaceId, {
+      ai: { projects: aiResourcesBefore.projects.total, prompts: prompts.used },
```
aliciadriani: **Nit:** `aiResourcesBefore.projects.total` is accessed unguarded here. Pre-existing pattern (not introduced by this PR), but since the prompts side is now validated defensively via `resolvePromptDims`, it'd be consistent to guard `projects` the same way rather than risk a TypeError if a future response omits it.

scripts/serenity-metered-405-canary.mjs
aliciadriani: Addressed: extracted `resolvePromptDims`/`isZeroHeadroom` (and `isUsedTotalPair`) into a new sibling module `scripts/serenity-metered-405-canary-resources.mjs`, since the main script has top-level side effects (argv parsing against real argv, env-var validation that calls `process.exit`) that made it unsafe to `import` directly for testing. Added `test/scripts/serenity-metered-405-canary-resources.test.js` covering legacy shape, tiered shape, legacy-preferred-when-both-present, missing-one-tier, empty/undefined input, and a partial dimension — 8 tests, all passing. Commit 7665de38.

scripts/serenity-metered-405-canary.mjs
aliciadriani: Addressed: guarded `projects.total` the same way the prompts side already is — reads it into `projectsTotal`, checks `typeof projectsTotal !== 'number'`, and exits with a clear message before attempting the drain transfer, instead of accessing `aiResourcesBefore.projects.total` unguarded. Commit 7665de38.

aliciadriani (commented):
## Review — round 2 (post-fix)

Re-reviewed the full current diff (4 files, +181/−31) after the round-1 fixes.

**Must Fix:** none.

**Should Fix:** none — both round-1 items verified resolved: `resolvePromptDims`/`isZeroHeadroom` are now in a separate, importable, tested module (8 passing tests), and `projects.total` is guarded before use in the drain transfer.

**Nits (1):**
1. One test-coverage gap remains — inline comment below.

**Suggestion:** none.

**What's good:** clean extraction into a sibling module solved the actual blocker (top-level side effects made the original script unsafe to import for testing) rather than working around it; lint clean, tests passing, and re-verified live against the real workspace after the fix.

test/scripts/serenity-metered-405-canary-resources.test.js
```diff
@@ -0,0 +1,72 @@
...
+    });
+
+    it('prefers the legacy shape when both a flat prompts field and tiered fields are present', () => {
+      const result = resolvePromptDims({
+        prompts: { used: 1, total: 1 },
+        daily_prompts: { used: 0, total: 0 },
+        weekly_prompts: { used: 0, total: 0 },
+      });
+      expect(result.shape).to.equal('legacy');
+    });
```
aliciadriani, line 47: **Nit:** the existing coverage tests "invalid prompts + no tiered fields" (returns null) and "valid legacy + valid tiered" (prefers legacy), but not the fallback branch in between: a malformed/incomplete `prompts` object (e.g. `{}`) alongside a *fully valid* tiered pair. That's the actual fallback path `resolvePromptDims` is designed for — `isUsedTotalPair(aiResources?.prompts)` is false, so it falls through to the tiered branch and should resolve as `'tiered'`. Worth locking that in explicitly so a future refactor can't silently break the fallback.

aliciadriani: Addressed: added `falls back to the tiered shape when prompts is present but malformed` test case (`prompts: {}` alongside a fully valid tiered pair → resolves as `'tiered'`). 9/9 tests passing. Commit d34784f6.

aliciadriani (commented):
## Review — round 3 (converged)

Re-checked the diff since round 2: the only change is the requested test case (`falls back to the tiered shape when prompts is present but malformed`), correctly implemented and passing. `scripts/serenity-metered-405-canary.mjs` and `scripts/serenity-metered-405-canary-resources.mjs` are unchanged since round 2's clean pass.

**Must Fix:** none.
**Should Fix:** none.
**Nits:** none.
**Suggestion:** none.

Full suite re-run: 9/9 tests passing, eslint clean, `node --check` clean, `--dry-run` re-verified live against the workspace used for the original SITES-49206 canary run. Nothing left to address — good to merge from a review standpoint.

MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - clean bug fix with solid test coverage, correctly refuses to guess at undocumented write shapes.
_Changes: Handles Semrush's new tiered daily/weekly prompt resource shape in the metered-405 canary script, extracting pure helpers into a testable sibling module (4 files)._

<details>

- nit: `isZeroHeadroom` returns `true` on an empty array (vacuous truth of `Array.every`) - not a live bug since `resolvePromptDims` always returns 1-2 element arrays, but a latent foot-gun if the helper is reused. A one-line guard (`if (!dims.length) return false;`) would make the contract explicit. - `scripts/serenity-metered-405-canary-resources.mjs:42`
- nit: The spread `{ key, ...aiResources[key] }` copies all properties from the dimension object (not just `used`/`total`), so extra vendor fields silently ride along in `dims`. Harmless today (nothing downstream reads unexpected props, and extra data is useful for debugging), but worth knowing if `dims` objects are ever serialized for a write call. - `scripts/serenity-metered-405-canary-resources.mjs:22`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 40s | <b>Cost:</b> $3.67 | <b>Commit:</b> <code>acca0bca4b216ec6d6b2a3f6c446b343172f43ed</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

scripts/serenity-metered-405-canary-resources.mjs
```diff
@@ -0,0 +1,59 @@
...
+    return { shape: 'tiered', dims: tiered };
+  }
+  return null;
+}
+
+// `dims` is always the 1-2 element array `resolvePromptDims` returns, never empty — but since this
+// is an exported, standalone helper, make the empty-input contract explicit rather than relying on
+// `Array.prototype.every`'s vacuous-true default (which would otherwise silently read "zero
+// headroom" for "no dimensions given").
+export function isZeroHeadroom(dims) {
```
aliciadriani, line 54: Addressed (MysticatBot nit): `isZeroHeadroom` now explicitly returns `false` for an empty `dims` array instead of relying on `Array.prototype.every`'s vacuous-true default, so the empty-input contract is explicit for any future reuse. Added a test (`is false (not vacuously true) for an empty dims array`) locking this in — 11/11 tests passing.

scripts/serenity-metered-405-canary-resources.mjs
```diff
@@ -0,0 +1,59 @@
...
+
+export function isUsedTotalPair(dim) {
+  return Boolean(dim) && typeof dim.used === 'number' && typeof dim.total === 'number';
+}
+
+// Narrows a raw resource-dimension object down to exactly {key, used, total} — deliberately NOT a
+// spread of the source object, so unrelated vendor fields on the raw response never silently ride
+// along in `dims` (they're not needed downstream today, but `dims` shapes what a future write call
+// could serialize, so keeping it exact avoids that becoming a footgun later).
+function pickDim(key, source) {
```
aliciadriani, line 31: Addressed (MysticatBot nit): replaced the `{ key, ...source }` spread with a `pickDim(key, source)` helper that narrows each dimension down to exactly `{ key, used, total }`, so unrelated vendor fields on the raw response no longer silently ride along in `dims`. Added a test (`does not carry unrelated vendor fields through into dims`) locking this in.

MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - prior nits cleanly addressed, no new issues introduced.
_Changes: Handles Semrush's new tiered daily/weekly prompt resource shape in the metered-405 canary script, extracting pure helpers into a testable sibling module (4 files)._

- `isZeroHeadroom` vacuous-truth on empty array now has explicit `if (!dims.length) return false` guard
- Spread operator no longer copies unrelated vendor fields; `pickDim` narrows to exactly `{key, used, total}`

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 36s | <b>Cost:</b> $2.78 | <b>Commit:</b> <code>67c486f0d762d8a2ab948400d954b9d0094fa301</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #3077, opened 2026-08-17
fix(sites): deep-merge hlxConfig/deliveryConfig on PATCH to preserve content.source (SITES-49362)

tr1eder:
## Summary

`PATCH /sites/:siteId` (`updateSite`) replaced `hlxConfig` and `deliveryConfig` **wholesale** from the request body, so a partial patch (e.g. `hlxConfig` with only `rso`/`code`) silently dropped sibling sub-keys — notably crosswalk `hlxConfig.content.source.type: "markup"`. This deep-merges both against the existing value instead:

- omitted keys keep their existing value
- an explicit `null` deletes a key
- nested objects merge recursively; arrays and scalars replace

`authoringType` stays scalar.

This is the durable **API-level backstop** behind the ASO "Connect to AEM Sites" dialog guard (SITES-49236 / `experience-success-studio-ui#2168`): it stops **any** caller — the dialog, Slack commands, scripts, or other services — from wiping crosswalk config on a partial update, not just the dialog. Relevant now that product is investing in crosswalk (World Bank, auth0, …).

Supersedes draft #3012 — re-applied on current `main` so it doesn't regress features that landed since that branch was cut; the `deepMerge` helper and tests are Miriam's (co-authored).

## Testing

- 4 controller unit tests: partial `hlxConfig` patch preserves `content.source`; `null` deletes a sub-key; omitted `hlxConfig` unchanged; `deliveryConfig` merges. `npx mocha test/controllers/sites.test.js` — green.
- 1 integration test (`test/it/shared/tests/sites.js` → `PATCH /sites/:siteId`): partial patch preserves `content.source`. Not run locally (needs Docker + ECR); CI validates.
- `lint` + `type-check` clean via the pre-commit hook.

No API schema change (fields unchanged; behavior-only), so no OpenAPI/example updates needed.

---

- [x] make sure to link the related issues in this description. Or if there's no issue created, make sure you describe here the problem you're solving.

- [ ] make sure you add a "Not implemented yet" note the endpoint description, if the implementation is not ready yet. Ideally, return a 501 status code with a message explaining the feature is not implemented yet.

## Related Issues

- SITES-49362

Thanks for contributing!

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Miriam Mchome", "Dominique Jäggi"]
```

### DIFF
--- a/src/controllers/sites.js
+++ b/src/controllers/sites.js
@@ -105,6 +105,8 @@ async function buildResolveData(org, site, context, asoEntitlement) {
  * - Omitted keys keep their existing value in `base`.
  * - A `null` value deletes the key.
  * - Plain objects are merged recursively; arrays and non-object values replace.
+ * - Dangerous keys (`__proto__`/`constructor`/`prototype`) are ignored to
+ *   prevent prototype pollution from an untrusted request body.
  *
  * @param {object} base - Existing value.
  * @param {object} patch - Incoming partial patch.
@@ -113,7 +115,9 @@ async function buildResolveData(org, site, context, asoEntitlement) {
 function deepMerge(base, patch) {
   const result = { ...(isObject(base) ? base : {}) };
   for (const [key, value] of Object.entries(patch)) {
-    if (value === null) {
+    if (key === '__proto__' || key === 'constructor' || key === 'prototype') {
+      // Ignore prototype-pollution vectors from the request body.
+    } else if (value === null) {
       delete result[key];
     } else if (isObject(value) && isObject(result[key])) {
       result[key] = deepMerge(result[key], value);
--- a/src/controllers/sites.js
+++ b/src/controllers/sites.js
@@ -99,6 +99,32 @@ async function buildResolveData(org, site, context, asoEntitlement) {
 }
 
 /**
+ * Recursively deep-merges `patch` into `base`, returning a new object.
+ *
+ * Merge semantics (used for PATCH of nested config like hlxConfig/deliveryConfig):
+ * - Omitted keys keep their existing value in `base`.
+ * - A `null` value deletes the key.
+ * - Plain objects are merged recursively; arrays and non-object values replace.
+ *
+ * @param {object} base - Existing value.
+ * @param {object} patch - Incoming partial patch.
+ * @returns {object} Merged result.
+ */
+function deepMerge(base, patch) {
+  const result = { ...(isObject(base) ? base : {}) };
+  for (const [key, value] of Object.entries(patch)) {
+    if (value === null) {
+      delete result[key];
+    } else if (isObject(value) && isObject(result[key])) {
+      result[key] = deepMerge(result[key], value);
+    } else {
+      result[key] = value;
+    }
+  }
+  return result;
+}
+
+/**
  * Resolves the org's per-product default site from config.defaults, validating it belongs
  * to the org and is enrolled. Returns the resolved data object or null to fall through
  * to getFirstEnrollment().
@@ -1331,12 +1357,15 @@ function SitesController(ctx, log, env) {
       ? requestBody.authoringType
       : site.getAuthoringType();
 
+    // Deep-merge `deliveryConfig`/`hlxConfig` so a partial patch (e.g. hlxConfig with only
+    // rso+code) preserves omitted sub-keys like hlxConfig.content.source. An explicit `null`
+    // for a sub-key deletes it; omitting the field entirely keeps the existing value.
     const nextDeliveryConfig = isObject(requestBody.deliveryConfig)
-      ? requestBody.deliveryConfig
+      ? deepMerge(site.getDeliveryConfig(), requestBody.deliveryConfig)
       : site.getDeliveryConfig();
 
     const nextHlxConfig = isObject(requestBody.hlxConfig)
-      ? requestBody.hlxConfig
+      ? deepMerge(site.getHlxConfig(), requestBody.hlxConfig)
       : site.getHlxConfig();
 
     const authoringTypeChanged = nextAuthoringType !== site.getAuthoringType();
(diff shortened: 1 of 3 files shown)


src/controllers/sites.js
```diff
@@ -98,6 +98,32 @@ async function buildResolveData(org, site, context, asoEntitlement) {
...
+ * Merge semantics (used for PATCH of nested config like hlxConfig/deliveryConfig):
+ * - Omitted keys keep their existing value in `base`.
+ * - A `null` value deletes the key.
+ * - Plain objects are merged recursively; arrays and non-object values replace.
+ *
+ * @param {object} base - Existing value.
+ * @param {object} patch - Incoming partial patch.
+ * @returns {object} Merged result.
+ */
+function deepMerge(base, patch) {
```
MysticatBot, line 115: issue (blocking): **Prototype pollution via `__proto__` / `constructor` keys.**

`deepMerge` iterates `Object.entries(patch)` and assigns keys to `result` via bracket notation. When the request body contains `{"__proto__": {"polluted": true}}`, `Object.entries` yields it as an own-property entry. The assignment `result["__proto__"] = value` can modify the result object's internal prototype chain via the built-in setter.

While this does not globally pollute `Object.prototype`, it gives attacker-controlled data influence over the prototype of the merged config object. Downstream code doing `for...in`, prototype-walking, or `hasOwnProperty` checks may observe unexpected state.

**Fix (2 lines):**
```js
const UNSAFE_KEYS = new Set(['__proto__', 'constructor', 'prototype']);

function deepMerge(base, patch) {
  const result = { ...(isObject(base) ? base : {}) };
  for (const [key, value] of Object.entries(patch)) {
    if (UNSAFE_KEYS.has(key)) continue;
    // ... rest unchanged
  }
  return result;
}
```

Alternatively, use `Object.create(null)` as the accumulator to eliminate the setter entirely.

MysticatBot (changes requested):
Hey @tr1eder,

**Verdict: Request changes** - one blocking finding (prototype pollution guard needed on deepMerge).
**Complexity: MEDIUM** - medium diff; API surface risk signal.
_Changes: Replaces wholesale assignment of hlxConfig/deliveryConfig on PATCH /sites/:siteId with a recursive deep-merge that preserves omitted sub-keys, adds null-deletion semantics, and includes 4 unit tests + 1 integration test (3 files)._

## Must fix before merge

1. [Important] Prototype pollution keys not filtered in `deepMerge` - `src/controllers/sites.js:113` (details inline)

<details>

- nit: Unbounded recursion depth in `deepMerge` - a depth cap (e.g. 10) would be trivial defense-in-depth against stack overflow from malicious payloads - `src/controllers/sites.js:113`
- suggestion: Add focused unit tests for `deepMerge` edge cases (undefined base, empty patch, array replacement, deeply nested null deletion) to anchor the merge contract independently of the controller - `test/controllers/sites.test.js`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 10s | <b>Cost:</b> $6.22 | <b>Commit:</b> <code>a4770525320491a33afc87dcd68e6bb90901f66c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @tr1eder,

**Verdict: Approve** - prior blocking finding (prototype pollution) is addressed; no new blocking issues remain.
_Changes: Replaces wholesale assignment of hlxConfig/deliveryConfig on PATCH /sites/:siteId with a recursive deep-merge that preserves omitted sub-keys, adds null-deletion semantics, and guards against prototype pollution (3 files)._
Note: CI checks are currently pending - confirm green before merge.

<details>

- nit: Unbounded recursion depth in `deepMerge` - a crafted request body with thousands of nested levels (~60KB) can exhaust the call stack and crash the Lambda invocation. A one-line depth cap (e.g. `if (depth > 10) return patch`) converts this from "crash" to "defined behavior" with zero impact on real config shapes (2-3 levels deep). Low severity given Lambda auto-restart and auth requirement, but trivially fixable. - `src/controllers/sites.js:124`
- nit: Prototype-pollution test only covers `__proto__` at the top level - add cases for `constructor` and `prototype` keys, and a nested dangerous key (e.g. `{ rso: { __proto__: { x: 1 } } }`) to assert the guard fires at every recursion depth and prevent regression if someone refactors the loop. - `test/controllers/sites.test.js:835`

</details>

- Prototype-pollution guard now filters `__proto__`/`constructor`/`prototype` keys with a dedicated unit test (commit 5dcc51dc).

tr1eder:
### Regression evidence — tests fail on the old code, pass on the fix

To confirm these specs actually catch the bug (and aren't vacuous), I ran them against the pre-fix `sites.js` (wholesale-replace on `origin/main`) and against this branch.

**Old (wholesale-replace) — 4 failing:**
- `deep-merges hlxConfig, preserving content.source on partial patch` → `expected { rso, code } to deeply equal { content: { source: {…} }, code, rso }` — i.e. `content.source` was dropped
- `deletes an hlxConfig sub-key when patched with null` → `expected { content: null } to deeply equal { rso: {…} }`
- `deep-merges deliveryConfig, preserving omitted sub-keys` → `expected { siteId: '1234' } to deeply equal { programId, environmentId, authorURL, siteId }`
- `ignores __proto__ keys (prototype pollution)` → fails

**This branch — all pass (7 passing).**

The first case is the exact reported bug: a partial `PATCH /sites` dropped `hlxConfig.content.source` — the crosswalk `type: "markup"` marker. The integration test (`PATCH /sites/:siteId` against real Postgres in `it-postgres`) exercises the same at the API level and is green in CI.

mimchome (approved):

